PR #387, opened 2026-02-04
Fixes #374

### LINKED ISSUE #374: Type-checking error when there is a function with body
The following program raises a type-checking error: 
```Lean
def smalltest :=
#strata
program Core;

function bool_to_int (b: bool) : int {if b then 1 else 0}
function str_to_bool (s: string) : bool;

procedure test () returns ()
{
  var b: string, i: int;
  i := bool_to_int(str_to_bool(b));
};

#end
```

The error is 
```
(0,(0-0)) ❌ Type checking error.
Impossible to unify string with bool.
```

But the when we rename the variable `b` in `procedure test` or `function bool_to_int` (only 1 of them, just to make the names different), then there is no error.

shigoel:
*Issue #, if available:*
https://github.com/strata-org/Strata/issues/374

*Description of changes:*

Use the new function `TEnv.addInNewestContext` which adds the identifier-`LTy` map to the newest context in favor of the old -- now deleted -- `TEnv.insertInContext`, which could over-write an identifier's entry in an older scope.

CC: @thanhnguyen-aws 

### DIFF
--- a/Strata/DDM/Integration/Categories.lean
+++ b/Strata/DDM/Integration/Categories.lean
@@ -3,12 +3,11 @@
 
   SPDX-License-Identifier: Apache-2.0 OR MIT
 -/
+module
+public import Strata.DDM.AST
 
-import Strata.DDM.AST
-
+public section
 namespace Strata.DDM.Integration
-
-open Strata
 
 /-- Init categories that map to primitive types (no interface/inductive needed) -/
 def primitiveCategories : Std.HashSet QualifiedIdent := Std.HashSet.ofList [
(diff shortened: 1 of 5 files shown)


StrataTest/Languages/Core/Examples/FuncTypeCheckBody.lean
joscoh, line 1: Super minor and non-blocking, but maybe name this `FuncTypeCheckBody` or something? We probably won't remember the bug or the fix in the future.

shigoel: You're right! Thanks -- this is an eye sore.

PR #388, opened 2026-02-05
Add opaque definitions for a simple API

atomb:
These opaque definitions will allow us to discuss and agree on a simple, high-level API that could be used to implement CLI commands and other Strata clients that live at a similar level of abstraction.

Most of these APIs can easily be implemented based on direct calls to existing code. One, at least, will require a little more work, however.


andrewmwells-amazon:
Design makes sense to me.

atomb:
> Looks good in general, but we haven't talked about generating SARIF from `VCResults` here, though the capability exists in Strata.

My thought is that that'd be an additional API, which we could certainly add later, rather than a change to any of these APIs. Does that seem reasonable to you?

PR #390, opened 2026-02-05
Reduce time complexity of type checking ops in Factory, prove termination of LExpr.resolveAux

aqjune-aws:
This patch

- Avoids recursive traversal of body of LFuncs in Factory during type checking. Instead, the bodys are typechecked separately.
- Proves termination of LExpr.resolveAux. The above refactoring enabled this :)
- And adds tiny well-formedness statements about type variables of LFunc. This includes a statement "concreteEval and body cannot coexist", which was briefly discussed in a closed space. Also, well-formedness related data structures are factored out to `FactoryWF` because it was getting larger. 

The `axioms` clause is not removed yet in this pull request, since it seems to require quite a few updates to translation of Map related functions.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Strata/Languages/Python/FunctionSignatures.lean
+++ b/Strata/Languages/Python/FunctionSignatures.lean
@@ -58,19 +58,19 @@ deriving Inhabited
 
 namespace Signatures
 
-def getFuncSigOrder (db : Signatures) (fname: FuncName) : List String :=
+def getFuncSigOrder (db : Signatures) (fname: FuncName) : Except String (List String) :=
   match  db.functions[fname]? with
-  | some decl => decl.args |>.map (·.name) |>.toList
-  | none => panic! s!"Missing function signature : {fname}"
+  | some decl => .ok (decl.args |>.map (·.name) |>.toList)
+  | none => .error s!"Missing function signature : {fname}"
 
 -- We should extract the function signatures from the prelude:
-def getFuncSigType (db : Signatures) (fname: FuncName) (arg: String) : String :=
+def getFuncSigType (db : Signatures) (fname: FuncName) (arg: String) : Except String String :=
   match  db.functions[fname]? with
-  | none => panic! s!"Missing function signature : {fname}"
+  | none => .error s!"Missing function signature : {fname}"
   | some decl =>
     match decl.argIndexMap[arg]? with
-    | none => panic! s!"Unrecognized arg : {arg}"
-    | some idx => decl.args[idx].type
+    | none => .error s!"Unrecognized arg : {arg}"
+    | some idx => .ok decl.args[idx].type
 
 end Signatures
 
--- a/Strata/Languages/Python/PythonToCore.lean
+++ b/Strata/Languages/Python/PythonToCore.lean
@@ -62,6 +62,27 @@ structure PythonClassDecl where
   name : String
 deriving Repr, BEq, Inhabited
 
+inductive ExtrinsicModelingChoice where
+  | havocAll : ExtrinsicModelingChoice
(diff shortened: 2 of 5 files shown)


Strata/DL/Lambda/FactoryWF.lean
```diff
@@ -0,0 +1,187 @@
...
+    ∀ b, f.body = .some b →
+      (LExpr.freeVars b).map (·.1.name) ⊆ f.inputs.map (·.1.name)
+  -- concreteEval does not succeed if the length of args is incorrect.
+  concreteEval_argmatch:
+    ∀ fn md args res, f.concreteEval = .some fn
+      → fn md args = .some res
+      → args.length = f.inputs.length
+  -- body and concreteEval cannot exist at once
+  body_or_concreteEval:
+    ¬ (f.concreteEval.isSome ∧ f.body.isSome)
```
joscoh: I would advocate for enforcing this in the types e.g. replace these two fields with `eval : Option (Either ConcreteEvalFunc LExpr)`, though I know this is a more disruptive change so others may disagree.

Strata/DL/Lambda/LExprT.lean
```diff
@@ -214,154 +214,136 @@ def inferConst (C: LContext T) (Env : TEnv T.IDMeta) (c : LConst) :
                 {c}\n\
                 Known Types: {C.knownTypes}")
 
-mutual
-partial def resolveAux (C: LContext T) (Env : TEnv T.IDMeta) (e : LExpr T.mono) :
+def resolveAux (C: LContext T) (Env : TEnv T.IDMeta) (e : LExpr T.mono) :
```
joscoh, line 217: Yay!

Strata/DL/Lambda/FactoryWF.lean
aqjune-aws: Yeah, wanted to avoid a large change in this patch.

Strata/Languages/Core/Core.lean
```diff
@@ -40,10 +40,15 @@ def typeCheck (options : Options) (program : Program)
   let C := { Lambda.LContext.default with
                 functions := factory,
                 knownTypes := Core.KnownTypes }
-  let (program, _T) ← Program.typeCheck C T program
-  -- dbg_trace f!"[Strata.Core] Annotated program:\n{program}"
-  if options.verbose >= .normal then dbg_trace f!"[Strata.Core] Type checking succeeded.\n"
-  return program
+  match Factory.typeCheck C T with
+  | .error k =>
+    -- TODO: DiagnosticModel for functions defined in Factory?
```
MikaelMayer, line 45: What do you mean by this TODO?

Strata/Languages/Core/Core.lean
```diff
@@ -40,10 +40,15 @@ def typeCheck (options : Options) (program : Program)
...
+    throw (DiagnosticModel.fromFormat k)
+  | .ok T =>
+    let (program, _T) ← Program.typeCheck C T program
+    -- dbg_trace f!"[Strata.Core] Annotated program:\n{program}"
```
MikaelMayer, line 49: You can remove this debug statement comment

andrewmwells-amazon: Personally I don't see an issue keeping these around.

Strata/Languages/Core/Core.lean
aqjune-aws: It means currently the error message does not explicitly state that it is from a Factory function, but if DiagnostmicModel can be extended to refer to the Factory functions then it would have more information. :)

PR #377, opened 2026-02-03
`Transform.runProgram` return an extra `updated` bool result, add analysis cache, and updates to `ProcedureInlining`

aqjune-aws:
*Description of changes:*

1. The definition of `Transform.runProgram` now returns `(updated, new program)`, and receives an extra optional arg `allowProcList : Option (List String)` that makes the transformation only run on specific procedures.

The goal of this change is to allow the user iteratively apply the transformation to the specific procedure (e.g., the "main" function) until there is no change anymore.

2. Add `doInline` fn to Procedure Inlining, let CoreTransformState cache the result of expensive analysis like CallGraph construction.

In this patch, ProcedureInlining.inlineCallCmd takes a more generic lambda function `doInline` that receives the calling procedure name as well as the CallGraph data structure and decide whether to inline or not.
Since building CallGraph every time is expensive, CallGraph is incrementally updated after each inlining.
To keep the latest status of CallGraph, we have a new `CoreTransformState` datatype which tracks analyses cache, and it is the transformation's responsibility that must invalidate or correctly update the cache. This mimics what LLVM's passes do.
From the CallGraph, you can check whether e.g., there is only one caller of some procedure or not.

Also, since we have `CoreTransformState`, more cleaups to the signatures of transformation functions were possible. The transformation passes were originally taking an extra source `Program` argument that was being transformed, but this was awkward because transformations are defined at the Statement level. This removes the extra argument, and makes `CoreTransformState` track the currently modifying Program and Procedure.

3. The `Program.filterProcedures` function is extracted out as a new transformation called `FilterProcedure`.

The new `Program.filterProcedures` introduced in https://github.com/strata-org/Strata/pull/375 was extracted out to `FilterProcedure` because I thought that the correctness of this `FilterProcedure` can be very nicely described. :) To address the comment that I left in the pull request, I made `FilterProcedure` a standalone transform that is not parameteric to other transformations.
Also, there is performance win - combined with the AnalysisCache, now CallGraph is calculated inside `FilterProcedure` once, reused in `ProcedureInlining`, then reused in the second `FilterProcedure`!

Also, the proofs in CallElimCorrect is commentized even more.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

@thanhnguyen-aws

PR #391, opened 2026-02-05
Flesh out Python support and add corresponding tests

andrewmwells-amazon:
Flesh out Python support and add corresponding tests

### DIFF
--- a/.kiro/settings/mcp.json
+++ b/.kiro/settings/mcp.json
@@ -22,7 +22,9 @@
         "lean_loogle",
         "lean_leanfinder",
         "lean_state_search",
-        "lean_hammer_premise"
+        "lean_hammer_premise",
+        "lean_file_outline",
+        "lean_local_search"
       ]
     }
   }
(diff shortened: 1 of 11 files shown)


StrataTest/Languages/Python/expected/test_arithmetic.expected
```diff
@@ -0,0 +1,24 @@
+
+datetime_now_ensures_0: ✅ pass (at byte 7129)
```
shigoel, line 2: I expect that the prelude proofs will go away when https://github.com/strata-org/Strata/pull/377 is merged. Right @aqjune-aws?

shigoel (approved):

StrataTest/Languages/Python/expected/test_arithmetic.expected
aqjune-aws: I think these messages will still be here because the pull request does not update pyAnalyze to only deal with main yet. :) But indeed prelude proof messages will disappear at some point.

shigoel: Awesome, thanks!

PR #393, opened 2026-02-06
Missing extrinsic handling

andrewmwells-amazon:
Add initial implementation for handling unmodeled functions

Suppose we need to analyze Python code that looks like this:
```
def foo(l: List[str]):
    a : int = 0
    b : int = 0
    if(len(l) > 0):
      a = unmodeled_fun(l) # line 5
      return modeled_fun(l[0])
    else:
      b = -1
    return modeled_fun_int(b)
 ```

What we should do for modeled_fun seems clear— we use the model we have. The question is what we should do for unmodeled_fun. 

We propose dividing unmodeled functions into three categories (for now we ignore aliasing):

## Category A (`havocAll`)

An unmodeled function in Category A is assumed to havoc everything. This is the safest option, but will let us prove the least about the code. In our example, at line 5 we would havoc a, l and b.

## Category B (`havocArgsAndRet`)

An unmodeled function in Category B is assumed to havoc only it’s parameters. This strikes a balance between safety and expressive power. In our example, at line 5 we would havoc a and l[0].

## Category C (`havocRet`)

An unmodeled function in Category C is assumed to havoc neither it’s parameters nor other variables. This is the least safe option, but lets us prove the most. In our example, at line 5 we would havoc a.

### DIFF
--- a/Strata/DL/SMT/Solver.lean
+++ b/Strata/DL/SMT/Solver.lean
@@ -4,6 +4,8 @@
   SPDX-License-Identifier: Apache-2.0 OR MIT
 -/
 
+import Strata.Languages.Core.Options
+
 /-!
 Based on Cedar's Term language.
 (https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/SymCC/Solver.lean)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by andrewmwells-amazon** [comment]
> It seems a and b are locals, no? Since variables a and b are local variables with type int, I think they are inaccessible from unmodeled_fun (if they were list they could have been referenced through aliasing). with Category A we still have a theoretical reason to not havoc them.

```
import ctypes
import sys

def unmodeled_fun(l):
    """Modifies the caller's local variable 'b' by mutating the integer object"""
    frame = sys._getframe(1)
    locals_dict = frame.f_locals
    
    if 'b' in locals_dict:
        b_obj = locals_dict['b']
        b_id = id(b_obj)
        
        # Mutate the integer by directly modifying its internal representation
        # For a positive single-digit value like 999, we need ob_size=1 and digit[0]=999
        class PyLongObject(ctypes.Structure):
            _fields_ = [
                ("ob_refcnt", ctypes.c_ssize_t),
                ("ob_type", ctypes.c_void_p),
                ("ob_size", ctypes.c_ssize_t),
                ("ob_digit", ctypes.c_uint32 * 10)
            ]
        
        long_ptr = ctypes.cast(b_id, ctypes.POINTER(PyLongObject))
        
        # Set to 999: positive number with 1 digit
        long_ptr.contents.ob_size = 1
        long_ptr.contents.ob_digit[0] = 0
    
    return 42

def modeled_fun(s):
    return f"String: {s}"

def modeled_fun_int(i):
    return f"Int: {i}"

def foo(l):
    a = 0
    b = 10000  # Use a larger integer not in the small int cache
    if len(l) > 0:
        print(f"Before unmodeled_fun, b = {b}")
        a = unmodeled_fun(l)
        print(f"After unmodeled_fun, b = {b}")
        return modeled_fun(l[0])
    else:
        b = -10000
    return modeled_fun_int(b)

if __name__ == "__main__":
    print("Test 1: Non-empty list (b is modified by unmodeled_fun)")
    result1 = foo(["test"])
    print(f"Result: {result1}\n")
```

Output:
```
Test 1: Non-empty list (b is modified by unmodeled_fun)
Before unmodeled_fun, b = 10000
After unmodeled_fun, b = 0
Result: String: test
```

andrewmwells-amazon:
Agree it would make sense to add a 4th case that havocs all globals instead of everything. In either case, this PR leaves A and the hypothetical A' unimplemented.

aqjune-aws (approved):

PR #395, opened 2026-02-09
Adding string support to Laurel

olivier-aws:
*Description of changes:*

Add support of String to Laurel, with a translation to Strata Core strings.

Added a new test: T_11_String

### DIFF
--- a/Strata/DL/SMT/Solver.lean
+++ b/Strata/DL/SMT/Solver.lean
@@ -4,6 +4,8 @@
   SPDX-License-Identifier: Apache-2.0 OR MIT
 -/
 
+import Strata.Languages.Core.Options
+
 /-!
 Based on Cedar's Term language.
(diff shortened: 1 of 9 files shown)


Strata/Languages/Laurel/Grammar/LaurelGrammar.st
```diff
@@ -4,11 +4,13 @@ dialect Laurel;
 category LaurelType;
 op intType : LaurelType => "int";
 op boolType : LaurelType => "bool";
+op stringType : LaurelType => "string";
 op compositeType (name: Ident): LaurelType => name;
 
 category StmtExpr;
 op literalBool (b: Bool): StmtExpr => b;
 op int(n : Num) : StmtExpr => n;
+op string (s: Init.Str): StmtExpr => s;
```
fabiomadge:
```suggestion
op string (s: Str): StmtExpr => s;
```

keyboardDrummer:
I expect that we'll need different kinds of String in Laurel to support the different types of strings in the front-ends, but I think Lean's String might work well for Python.

PR #396, opened 2026-02-09
Add metadata to more places in Laurel

keyboardDrummer:
### Changes
- Add metadata to many more places in Laurel. For example, preconditions and postconditions are now given a metadata and that is propagated to Core
- Made it an internal error if diagnostics are reported for which no source location is known
- Remove the default empty value for the metadata of `Procedure.Check`

### Testing
- Turned on tests for preconditions and postconditions, although the precondition test does not report errors correctly due to Core reporting them on the precondition instead of on the call, and with an incorrect message.

### DIFF
--- a/Strata/Languages/Laurel/Laurel.lean
+++ b/Strata/Languages/Laurel/Laurel.lean
@@ -201,10 +201,7 @@ abbrev HighTypeMd := WithMetadata HighType
 abbrev StmtExprMd := WithMetadata StmtExpr
 
 theorem WithMetadata.sizeOf_val_lt {t : Type} [SizeOf t] (e : WithMetadata t) : sizeOf e.val < sizeOf e := by
-  cases e
-  rename_i val md
-  show sizeOf val < 1 + sizeOf val + sizeOf md
-  omega
+  cases e; grind
 
 instance : Inhabited StmtExpr where
   default := .Hole


Strata/Languages/Laurel/Laurel.lean
```diff
@@ -63,22 +63,33 @@ instance : Repr (Imperative.MetaData Core.Expression) := inferInstance
 
 
 mutual
+
+structure HighTypeMd where
+  val : HighType
+  md : Imperative.MetaData Core.Expression
+  deriving Repr
+
+structure StmtExprMd where
```
MikaelMayer: I would recommend creating one structure that adds metadata to any type like the DDM does, i.e.
structure WithMetadata (t: T) where
  val: T
  md: Imperative.Metadata Core.Expression
  deriving Repr

It will make it easier to generalize down the road. You can always make StmtExprMd an abbreviation of it.

I would even have you consider reusing DDM's `Ann` that is fully parametric in the metadata and has the same purpose. It has also methods to replace the metadata.

StrataTest/Languages/Laurel/Examples/Objects/T1_MutableFields.lean
MikaelMayer, line 23: Interesting, there are no modifies clauses? Nothing to indicate what the procedure can or cannot change?

keyboardDrummer: in a subsequent PR that adds support for modifies clauses, this generates an error, but only because there is an ensures clause so the procedure is opaque.

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: Applied the change

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -153,114 +171,113 @@ def freshVarName : TransformM Identifier := do
...
+/-- Helper to wrap a StmtExpr into StmtExprMd with empty metadata -/
+def mkMd (e : StmtExpr) : StmtExprMd := ⟨e, #[]⟩
+
 /--
 Transform an expression, adding heap parameters where needed.
 - `heapVar`: the name of the heap variable to use
 - `valueUsed`: whether the result value of this expression is used (affects optimization of heap-writing calls)
 -/
-def heapTransformExpr (heapVar : Identifier) (expr : StmtExpr) (valueUsed : Bool := true) : TransformM StmtExpr :=
+partial def heapTransformExpr (heapVar : Identifier) (expr : StmtExprMd) (valueUsed : Bool := true) : TransformM StmtExprMd :=
```
joscoh: Would prefer not to make things `partial` unless needed. Is this because it is trickier with the nested record structure for metadata?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -239,17 +248,17 @@ structure CompositeType where
...
+  constraint : StmtExprMd
+  witness : StmtExprMd
 
 /-
 Note that there are no explicit 'inductive datatypes'. Typed unions are created by
 creating a CompositeType for each constructor, and a ConstrainedType for their union.
 
 Example 1:
-`composite Some<T> { value: T }`
+`composite SomeT> { value: T }`
```
joscoh: Not sure why this was changed

Strata/Languages/Laurel/LaurelFormat.lean
```diff
@@ -29,7 +28,18 @@ def formatOperation : Operation → Format
   | .Gt => ">"
   | .Geq => ">="
 
-def formatHighType : HighType → Format
+private theorem HighTypeMd.sizeOf_val_lt (e : HighTypeMd) : sizeOf e.val < sizeOf e := by
+  cases e
```
joscoh: Does `cases e<;> simp_wf<;> omega` work? (the current proof is fine, just suggesting a potentially shorter one)

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -22,24 +22,31 @@ open Core (intAddOp intSubOp intMulOp intDivOp intModOp intNegOp intLtOp intLeOp
...
+  rename_i val md
+  show sizeOf val < 1 + sizeOf val + sizeOf md
+  omega
+
 /-
 Translate Laurel HighType to Core Type
 -/
-def translateType (ty : HighType) : LMonoTy :=
-  match ty with
+partial def translateType (ty : HighTypeMd) : LMonoTy :=
```
joscoh: termination should be provable here

Strata/Languages/Laurel/LaurelFormat.lean
keyboardDrummer: It does

PR #397, opened 2026-02-09
Support modifies clauses

keyboardDrummer:
### Changes
1. Change the encoding of the Heap to use `Map Composite (Map Field Box)` where Box is a datatype that can hold different values. This way, the translation of Laurel remains executable. The previous Heap encoding relied on axioms.
1. Support modifies clauses

### Testing 
- Update T1_MutableFields because change 1 fixed a bug.
- Add T4_ModifiesClauses

### DIFF
--- a/Strata/DL/Imperative/Stmt.lean
+++ b/Strata/DL/Imperative/Stmt.lean
@@ -9,6 +9,9 @@
 import Strata.DL.Imperative.Cmd
 
 namespace Imperative
+
+open Std.Format
+
 ---------------------------------------------------------------------
 
 /-! ## Statements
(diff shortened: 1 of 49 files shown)


Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -0,0 +1,99 @@
...
+      match (computeExprType env types target).val with
+      | .UserDefined typeName =>
+          match lookupFieldInTypes types typeName fieldName with
+          | some ty => ty
+          | none => panic s!"Could not find field in type"
+      | _ => panic s!"Selecting from a type that's not a composite"
+  -- Pure field update returns the same type as the target
+  | .PureFieldUpdate target _ _ => computeExprType env types target
+  -- Calls — we don't track return types here, so fall back to TVoid
+  | .StaticCall _ _ => panic "Not supported StaticCall"
```
atomb, line 62: Given that this function will always need the option to fail (if given an ill-typed input), I'd prefer to prepare for that by making this function return `Except ErrorType HighTypeMd`, instead of using `panic`. I'll leave it up to you to choose what `ErrorType` should be. It could be a string or something more structured.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -264,131 +317,115 @@ def translateProcedure (constants : List Constant) (proc : Procedure) : Core.Pro
...
 
-def heapTypeDecl : Core.Decl := .type (.con { name := "Heap", numargs := 0 })
-def fieldTypeDecl : Core.Decl := .type (.con { name := "Field", numargs := 1 })
+def heapTyMono : LMonoTy := Core.mapTy (.tcons "Composite" []) (Core.mapTy (.tcons "Field" []) (.tcons "Box" []))
+
+def fieldTypeDecl : Core.Decl := .type (.con { name := "Field", numargs := 0 })
 def compositeTypeDecl : Core.Decl := .type (.con { name := "Composite", numargs := 0 })
 
+/-- The Box datatype with constructors for each Laurel primitive type + Composite -/
+def boxDatatype : Lambda.LDatatype Core.Visibility :=
```
atomb: It would be great to move this to some sort of prelude file eventually, but that shouldn't block this PR.

atomb (dismissed):

Strata/Languages/Laurel/LaurelTypes.lean
keyboardDrummer: I was planning on having this function take a proof that every identifier is in the environment. Transforming that for the recursive calls is quite a bit of work though, so I'd rather not do that now.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Done

PR #401, opened 2026-02-09
Change default SMT solver from z3 to cvc5

MikaelMayer:
Change default SMT solver from z3 to cvc5

This PR updates the default SMT solver throughout the codebase from z3 to cvc5. The changes include:

- Updated `defaultSolver` in `Strata/Languages/Core/Options.lean`
- Updated default solver parameter in `Strata/Languages/B3/Verifier/State.lean`
- Updated solver references in `StrataMain.lean` for Laurel analysis
- Updated all test files to use cvc5 instead of z3

Files that check for both z3 and cvc5 (like `Strata/DL/Imperative/SMTUtils.lean` and `Strata/Languages/Core/Verifier.lean`) were left unchanged as they handle both solvers.

### DIFF
--- a/Strata/Languages/Core/Options.lean
+++ b/Strata/Languages/Core/Options.lean
@@ -6,7 +6,7 @@
 
 inductive VerboseMode where
   | quiet
-  | results
+  | models
   | normal
   | debug
   deriving Inhabited, Repr, DecidableEq
@@ -14,7 +14,7 @@ inductive VerboseMode where
 def VerboseMode.toNat (v : VerboseMode) : Nat :=
   match v with
   | .quiet => 0
-  | .results => 1
+  | .models => 1
   | .normal => 2
   | .debug => 3
 
@@ -73,8 +73,8 @@ instance : Inhabited Options where
 def Options.quiet : Options :=
   { Options.default with verbose := .quiet }
 
-def Options.results : Options :=
-  { Options.default with verbose := .results }
+def Options.models : Options :=
+  { Options.default with verbose := .models }
 
 def Options.debug : Options :=
   { Options.default with verbose := .debug }
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -258,7 +258,7 @@ instance : ToFormat VCResult where
   format r := f!"Obligation: {r.obligation.label}\n\
                  Property: {r.obligation.property}\n\
                  Result: {r.result}\
-                 {r.smtResult.formatModelIfSat (r.verbose >= .results)}"
+                 {r.smtResult.formatModelIfSat (r.verbose >= .models)}"
 
 def VCResult.isSuccess (vr : VCResult) : Bool :=
   match vr.result with | .pass => true | _ => false
--- a/StrataTest/Languages/Core/Examples/RemoveIrrelevantAxioms.lean
+++ b/StrataTest/Languages/Core/Examples/RemoveIrrelevantAxioms.lean
@@ -173,7 +173,7 @@ Model:
 #guard_msgs in
 #eval do
   let results ← verify "cvc5" irrelevantAxiomsTestPgm
-        (options := {Options.results with removeIrrelevantAxioms := true})
+        (options := {Options.models with removeIrrelevantAxioms := true})
   IO.println (normalizeModelValues (toString results))
 
 ---------------------------------------------------------------------
@@ -230,6 +230,6 @@ Result: 🟡 unknown
 -/
 #guard_msgs in
 #eval verify "cvc5" irrelevantAxiomsTestPgm
-        (options := {Options.results with removeIrrelevantAxioms := false})
+        (options := {Options.models with removeIrrelevantAxioms := false})
 
 ---------------------------------------------------------------------


StrataTest/Languages/Core/Examples/RemoveIrrelevantAxioms.lean
```diff
@@ -107,53 +126,55 @@ Obligation: assert_4
 Property: assert
 Result: ❌ fail
 Model:
-($__x0, 3)
+($__x0, model_not_2)
```
shigoel, line 129: Could we use https://github.com/strata-org/Strata/blob/main/Strata/Languages/Core/Verifier.lean#L110 to not print the model in quiet mode or something?

shigoel (commented):
LGTM, except for one comment.

StrataTest/Languages/Core/Examples/RemoveIrrelevantAxioms.lean
MikaelMayer: I like that we are displaying at least what the model looks like, and that even though the model's value is redacted it is tested to not be 2 (which would be unsound).
It seems that the option to not print the model in quiet mode is already provided with the `(verbose: Bool)` parameter, which is currently set to true by the test suite.

If despite that information above, it's really your preference that we disable printing models in the test suite, I won't object.

shigoel: I think this may be a fix beyond the scope of this PR, but when we have `(options := Options.quiet)` in `Core.verify`, I think we should suppress model printing (but not in verbose/default mode). Upgrading solvers may change the model returned from the solvers, so it makes sense to be able to suppress this in our regressions. Or at least have an option to do so.

Strata/Languages/Core/Options.lean
```diff
@@ -6,15 +6,17 @@
 
 inductive VerboseMode where
   | quiet
+  | results
```
shigoel: Like this new option, thank you! Nit: to me, "results" also includes pass/fail though. Is something like `displayModel` or `displayTrace` more appropriate?

MikaelMayer:
```suggestion
  | models
```

Strata/Languages/Core/Options.lean
```diff
@@ -6,15 +6,17 @@
...
   | normal
   | debug
   deriving Inhabited, Repr, DecidableEq
 
 def VerboseMode.toNat (v : VerboseMode) : Nat :=
   match v with
   | .quiet => 0
-  | .normal => 1
-  | .debug => 2
+  | .results => 1
```
MikaelMayer:
```suggestion
  | .models => 1
```

Strata/Languages/Core/Options.lean
```diff
@@ -71,5 +73,8 @@ instance : Inhabited Options where
 def Options.quiet : Options :=
   { Options.default with verbose := .quiet }
 
+def Options.results : Options :=
```
MikaelMayer:
```suggestion
def Options.models : Options :=
```

Strata/Languages/Core/Options.lean
```diff
@@ -71,5 +73,8 @@ instance : Inhabited Options where
...
+  { Options.default with verbose := .results }
```
MikaelMayer:
```suggestion
  { Options.default with verbose := .models }
```

Strata/Languages/Core/Verifier.lean
```diff
@@ -258,7 +258,7 @@ instance : ToFormat VCResult where
   format r := f!"Obligation: {r.obligation.label}\n\
                  Property: {r.obligation.property}\n\
                  Result: {r.result}\
-                 {r.smtResult.formatModelIfSat true}"
+                 {r.smtResult.formatModelIfSat (r.verbose >= .results)}"
```
MikaelMayer:
```suggestion
                 {r.smtResult.formatModelIfSat (r.verbose >= .models)}"
```

StrataTest/Languages/Core/Examples/RemoveIrrelevantAxioms.lean
```diff
@@ -107,53 +126,55 @@ Obligation: assert_4
...
 Model:
-($__x3, 3)
+($__x3, model_not_2)
 -/
 #guard_msgs in
-#eval verify "z3" irrelevantAxiomsTestPgm
-        (options := {Options.quiet with removeIrrelevantAxioms := true})
+#eval do
+  let results ← verify "cvc5" irrelevantAxiomsTestPgm
+        (options := {Options.results with removeIrrelevantAxioms := true})
```
MikaelMayer:
```suggestion
        (options := {Options.models with removeIrrelevantAxioms := true})
```

StrataTest/Languages/Core/Examples/RemoveIrrelevantAxioms.lean
```diff
@@ -208,7 +229,7 @@ Property: assert
 Result: 🟡 unknown
 -/
 #guard_msgs in
-#eval verify "z3" irrelevantAxiomsTestPgm
-        (options := {Options.quiet with removeIrrelevantAxioms := false})
+#eval verify "cvc5" irrelevantAxiomsTestPgm
+        (options := {Options.results with removeIrrelevantAxioms := false})
```
MikaelMayer:
```suggestion
        (options := {Options.models with removeIrrelevantAxioms := false})
```

Tools/BoogieToStrata/Tests/Quantifiers.expect
```diff
@@ -1,16 +1,16 @@
...
+Quantifiers.core.st(99, 4) [assert_5]: ✅ pass
+Quantifiers.core.st(110, 4) [assert_6]: ✅ pass
 Quantifiers.core.st(123, 4) [assert_7]: ✅ pass
 Quantifiers.core.st(134, 4) [assert_8]: ✅ pass
 Quantifiers.core.st(146, 4) [assert_9]: ✅ pass
 Quantifiers.core.st(158, 4) [assert_10]: ✅ pass
 Quantifiers.core.st(170, 4) [assert_11]: ✅ pass
 Quantifiers.core.st(183, 4) [assert_12]: ✅ pass
 Quantifiers.core.st(195, 4) [assert_13]: ✅ pass
-Finished with 10 goals passed, 4 failed.
```
atomb, line 16: I'm happy to see that this is the only Boogie test that had to change, and that CVC5 is actually able to prove a bit more.

PR #402, opened 2026-02-10
Update Core formatting

keyboardDrummer:
### Changes
Update Core formatting.

```
(procedure coprime :  ((a : int) (b : int)) → ((return : bool)))
modifies: []
preconditions: (pre, ((~Bool.And ((~Int.Gt a) #0)) ((~Int.Gt b) #0)))
postconditions: (post, #true)
body: init (i : int) := init_i
i := a
if ((~Int.Lt b) a) then {i := b}
...
```

Becomes
```
procedure coprime :  ((a : int) (b : int)) → ((return : bool))
  modifies: []
  preconditions: (pre, ((~Bool.And ((~Int.Gt a) #0)) ((~Int.Gt b) #0)))
  postconditions: (post, #true)
{
  init (i : int) := init_i
  i := a
  if ((~Int.Lt b) a) then {i := b}
  ...
}
```
### Testing
Updated tests

### DIFF
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -495,17 +495,30 @@ def toDiagnosticModel (vcr : Core.VCResult) : Option DiagnosticModel := do
   match vcr.result with
   | .pass => none  -- Verification succeeded, no diagnostic
   | result =>
-    let fileRangeElem ← vcr.obligation.metadata.findElem Imperative.MetaData.fileRange
-    match fileRangeElem.value with
-    | .fileRange fileRange =>
-      let message := match result with
-        | .fail => "assertion does not hold"
-        | .unknown => "assertion could not be proved"
-        | .implementationError msg => s!"verification error: {msg}"
-        | _ => panic "impossible"
+    let message := match result with
+      | .fail => "assertion does not hold"
+      | .unknown => "assertion could not be proved"
+      | .implementationError msg => s!"verification error: {msg}"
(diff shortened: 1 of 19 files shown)


Strata/Languages/Core/Verifier.lean
```diff
@@ -527,7 +527,7 @@ structure Diagnostic where
   deriving Repr, BEq
 
 def DiagnosticModel.toDiagnostic (files: Map Strata.Uri Lean.FileMap) (dm: DiagnosticModel): Diagnostic :=
-  let fileMap := (files.find? dm.fileRange.file).get!
+  let fileMap := (files.find? dm.fileRange.file).getD (panic s!"Could not find {repr dm.fileRange.file} in {repr files.keys} when converting model '{dm}' to a diagnostic")
```
aqjune-aws, line 530: It is surprising to know that `panic!` only happens the option value was `.none`, because my mental model of Lean4 was that it does call-by-value reduction.

keyboardDrummer:
> FWIW, I'm working on [a pretty printer for Core](https://github.com/strata-org/Strata/tree/shilpi/core-pretty-printer) that'd render Core programs in the same DDM-defined concrete syntax as is used by the parser. That should land EoW or early next week.

One problem I see with that is that sometimes you want to do some processing before printing. For example, when you have `(app (app a b) c)`, then you want to print that as `paren (group (joinSep ([a,b,c].map format) line))` but you need to de-nest the expression to get `[a,b,c]`.

PR #403, opened 2026-02-10
Add support for SARIF output to pyAnalyze and pyAnalyzeLaurel

tautschnig:
*Description of changes:*

We previously added SARIF tooling and actual use to StrataVerify. This adds the same support to pyAnalyze and pyAnalyzeLaurel, which is enabled when adding `--sarif` to their invocation. Testing added to CI.

### DIFF
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -463,7 +463,6 @@ def pyAnalyzeLaurelCommand : Command where
               takesArg := .repeat "ion_file" },
             { name := "sarif", help := "Write results as SARIF to <file>.sarif." }]
   help := "Verify a Python Ion program via the Laurel pipeline. Translates Python to Laurel to Core, then runs SMT verification."
->>>>>>> origin/main
   callback := fun v pflags => do
     let verbose := pflags.getBool "verbose"
     let outputSarif := pflags.getBool "sarif"


StrataTest/Languages/Python/run_py_analyze_sarif.sh
```diff
@@ -0,0 +1,101 @@
...
+    if [ ! -f "$sarif_abs" ]; then
+        echo "ERROR: SARIF file not created for $base_name (expected $sarif_abs)"
+        failed=1
+        rm -f "$ion_file"
+        continue
+    fi
+
+    # Validate JSON structure and SARIF content
+    validation=$(python3 -c "
+import json, sys
```
MikaelMayer: I would prefer a separate python script instead of an embedded one. That will avoid a tower of inception.
And if python is going to be used, I would ask to consider making the entire analysis run on python.

tautschnig: Thanks, done (the entire analysis is now being run using Python).

StrataTest/Languages/Python/run_py_analyze_laurel_sarif.py
```diff
@@ -0,0 +1,81 @@
...
+import sys
+from pathlib import Path
+
+from validate_sarif import validate
+
+REPO_ROOT = Path(__file__).resolve().parent.parent.parent.parent
+TEST_DIR = Path(__file__).resolve().parent
+TEST_FILES = [
+    "tests/test_arithmetic.py",
+    "tests/test_precondition_verification.py",
```
shigoel: Curious why the test files are restricted to these two in both run_py_analyze_*sarif.py scripts.

tautschnig: I was thinking to keep test execution time small, but actually it doesn't take much longer to do all the feasible ones. Fixed.

tautschnig:
Reverting to draft as #451 should be merged first. Once that is merged I will resolve conflicts and mark this ready-for-review.

StrataTest/Languages/Python/run_py_analyze_laurel_sarif.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
```
andrewmwells-amazon: Can we combine these into 1 script with a `--laurel` argument?

tautschnig: Done.

joehendrix:
**Duplicated SARIF output block**

The SARIF write logic at `StrataMain.lean:389-400` and `StrataMain.lean:481-492` is nearly identical between `pyAnalyzeCommand` and `pyAnalyzeLaurelCommand`, and a third copy exists in `StrataVerify.lean:128-139`. Consider extracting a shared helper to reduce the repetition:

```lean
def writeSarifOutput
    (files : Map Strata.Uri Lean.FileMap)
    (vcResults : Core.VCResults)
    (outputPath : String) : IO Unit := do
  let sarifDoc := Core.Sarif.vcResultsToSarif files vcResults
  let sarifJson := Strata.Sarif.toPrettyJsonString sarifDoc
  try
    IO.FS.writeFile outputPath sarifJson
    IO.println s!"SARIF output written to {outputPath}"
  catch e =>
    IO.eprintln s!"Error writing SARIF: {e.toString}"
```

Each call site would then reduce to building the files map and calling `writeSarifOutput files vcResults (filePath ++ ".sarif")`. Not a blocker, but it would make future additions (e.g. adding `--sarif` to more commands) less error-prone.

joehendrix:
**Nit: `--sarif` flag help could note output location**

`StrataMain.lean:312` and `StrataMain.lean:406`:

```lean
{ name := "sarif", help := "Write results as SARIF to a file." }
```

The help string doesn't mention where the file is written. Something like `"Write results as SARIF to <file>.sarif."` would make the output location discoverable without reading source code. (Same applies to `StrataVerify.lean`, but that's pre-existing.)

joehendrix (approved):
All four review comments have been addressed in follow-up commits. The shared `writeSarifOutput` helper, stderr fix, scoped `warning_results`, and improved help string all look good.

Two minor items remain (not blocking):
- **Shallow test validation** — only `test_precondition_verification` and `test_arithmetic` have semantic checks; other tests use generic structural validation only.
- **Line length** — one hard violation at `run_py_analyze_sarif.py:82` (105 chars), plus several soft-limit warnings.

LGTM overall.

Issue #398, opened 2026-02-09
Concrete instantation of polymorphic procedure with precondition gives SMT encoding error


joscoh:
Given the following program:
```
def polyProcPgm : Program :=
#strata
program Core;
datatype List (a : Type) { Nil(), Cons(head: a, tail: List a) };
procedure Extract<a>(xs : List a) returns (h : a)
spec {
  requires List..isCons(xs);
};
procedure Test() returns () spec { ensures true; }
{
  var xs : List int;
  xs := Cons(1, Nil());
  havoc xs;
 //assume List..isCons(xs);
  var h : int;
  call h := Extract(xs);
};
#end

#eval verify "cvc5" polyProcPgm (options := .quiet)
```
The `requires` check gives 
```
Obligation: (Origin_Extract_Requires)Extract_requires_0
Property: assert
Result: 🚨 Implementation Error! SMT Encoding Error! Unimplemented encoding for type var $__ty0
```
However, if the `assume` line is uncommented, the encoding works correctly. Since the `assume` should not change any types, this seems like a bug.

PR #408, opened 2026-02-11
Transform procedure preconditions into assertions, run before typechecking

joscoh:
Adds new transformation to replace procedure preconditions with assertions 
Ensures polymorphic preconditions instantiated at concrete type are unified before SMT encoding

*Issue #, if available:* #398

*Description of changes:*

### DIFF
--- a/Strata/Transform/PreconditionAssert.lean
+++ b/Strata/Transform/PreconditionAssert.lean
@@ -31,8 +31,7 @@ Given a procedure's preconditions and the formal→actual substitution,
 produce a list of `assert` statements.
 -/
 private def mkPrecondAsserts
-    (proc : Procedure) (args : List Expression.Expr)
-    (md : Imperative.MetaData Expression) : List Statement :=
+    (proc : Procedure) (args : List Expression.Expr) : List Statement :=
   let subst := proc.header.inputs.keys.zip args
   proc.spec.preconditions.flatMap fun (label, check) =>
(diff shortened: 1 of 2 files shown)


Strata/Transform/PreconditionAssert.lean
```diff
@@ -0,0 +1,84 @@
...
+    match Program.Procedure.find? p pname with
+    | some proc => mkPrecondAsserts proc args ++ [.cmd (.call lhs pname args md)]
+    | none => [s]
+  | .block label bss md =>
+    [.block label (transformStmts p bss) md]
+  | .ite cond tss ess md =>
+    [.ite cond (transformStmts p tss) (transformStmts p ess) md]
+  | .loop guard measure invariant bss md =>
+    [.loop guard measure invariant (transformStmts p bss) md]
+  | _ => [s]
```
shigoel, line 63: nit: let's not use a catch-all here, esp. since statements can be in flux.

StrataTest/Languages/Core/ExprEvalTest.lean
```diff
@@ -189,7 +189,7 @@ open Lambda.LTy.Syntax
 
 
 -- This may take a while (~ 5min)
-#eval (checkFactoryOps false)
+-- #eval (checkFactoryOps false)
```
shigoel, line 192: I know this is temp, but leaving this comment as a reminder.

joscoh: Good catch, thanks

PR #414, opened 2026-02-12
Let Core use more generic Imperative.SMTUtils for discharging proof obligation

aqjune-aws:
There was duplicated code in Core's SMT verification condition discharger and more generic Imperative.SMTUtils. The latter one was virtually not being used, except examples in StrataTest. However, since the latter is more general, let Core use it.

### DIFF
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -293,7 +293,7 @@ def verifySingleEnv (pE : Program × Env) (options : Options)
             dbg_trace f!"\n\nResult: {result}\n{prog}"
           if options.stopOnFirstError then break else continue
       -- For `unknown` results, we appeal to the SMT backend below.
-      let maybeTerms := ProofObligation.toSMTTerms E obligation
+      let maybeTerms := ProofObligation.toSMTTerms E obligation SMT.Context.default options.useArrayTheory
       match maybeTerms with
       | .error err =>
         let err := f!"SMT Encoding Error! " ++ err
--- a/Strata/DDM/Util/SourceRange.lean
+++ b/Strata/DDM/Util/SourceRange.lean
@@ -5,6 +5,8 @@
 -/
 module
 
+public import Lean.Data.Position
+
 public section
 namespace Strata
 
@@ -33,6 +35,17 @@ def isNone (loc : SourceRange) : Bool := loc.start = 0 ∧ loc.stop = 0
 
 instance : Std.ToFormat SourceRange where
  format fr := f!"{fr.start}-{fr.stop}"
+
+/-- Format a SourceRange as a string using a FileMap for line:column conversion.
+    Renders location information in a format VSCode understands.
+    Returns "path:line:col-col" if on same line, otherwise "path:line:col". -/
+def format (loc : SourceRange) (path : System.FilePath) (fm : Lean.FileMap) : String :=
+  let spos := fm.toPosition loc.start
+  let epos := fm.toPosition loc.stop
+  if spos.line == epos.line then
+    s!"{path}:{spos.line}:{spos.column+1}-{epos.column+1}"
+  else
+    s!"{path}:{spos.line}:{spos.column+1}"
 
 end Strata.SourceRange
 end
--- a/Strata/DL/SMT/Op.lean
+++ b/Strata/DL/SMT/Op.lean
@@ -158,6 +158,12 @@ inductive Op.Strings : Type where
   | re_index : Nat → Op.Strings
 deriving Repr, DecidableEq, Inhabited, Hashable
 
+inductive Op.Arrays : Type where
+  ---------- SMTLib theory of arrays (`ArraysEx`) ----------
+  | select
+  | store
+deriving Repr, DecidableEq, Inhabited, Hashable
+
 inductive Op.DatatypeFuncs : Type where
   | constructor : Op.DatatypeFuncs
   | tester : Op.DatatypeFuncs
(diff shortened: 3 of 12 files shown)


StrataTest/DL/Imperative/Verify.lean
```diff
@@ -51,19 +51,20 @@ def verify (cmds : Commands) (verbose : Bool) :
...
         match ans with
         | .ok (result, estate) =>
-           results := results.push { obligation, result, estate }
+           let vcres := { obligation, result, estate }
+           results := results.push vcres
            if result ≠ .unsat then
             let prog := f!"\n\nEvaluated program:\n{format cmds}"
-            dbg_trace f!"\n\nObligation {obligation.label}: could not be proved!\
-                         \n\nResult: {result}\
+            dbg_trace f!"\n\nCould not be proved!\
```
atomb: I think it's useful to continue including the label here.

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -125,46 +114,122 @@ def runSolver (solver : String) (args : Array String) : IO IO.Process.Output :=
...
+    -- feedback (i.e., problem was `sat`) from the user.
+    match (processModel typedVarToSMTFn vars rawModel E) with
+    | .ok model => .ok (.sat model)
+    | .error _model_err => (.ok (.sat []))
   | "unsat"   =>  .ok .unsat
   | "unknown" =>  .ok .unknown
-  | _     =>  .error s!"stderr:{output.stderr}\nsolver stdout: {output.stdout}\n"
+  | _     =>
+    let stderr := output.stderr
+    let hasExecError := (stderr.splitOn "could not execute external process").length > 1
```
atomb: Is there not a `contains` function?

aqjune-aws: Missed it while blindly moving the code from Core, thanks.

Strata/Languages/Core/Verifier.lean
```diff
@@ -49,122 +49,15 @@ end Strata.SMT.Encoder
 namespace Core.SMT
 open Std (ToFormat Format format)
 open Lambda Strata.SMT
--- (TODO) Use DL.Imperative.SMTUtils.
```
shigoel, line 52: Thanks for doing all this work to remove this TODO! :)

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -125,46 +114,122 @@ def runSolver (solver : String) (args : Array String) : IO IO.Process.Output :=
...
+def dischargeObligation {P : PureExpr} [ToFormat P.Ident] [BEq P.Ident]
+  (encodeSMT : Strata.SMT.SolverM (List String × Strata.SMT.EncoderState))
   (typedVarToSMTFn : P.Ident → P.Ty → Except Format (String × Strata.SMT.TermType))
-  (vars : List P.TypedIdent) (smtsolver filename : String)
-  (terms : List Strata.SMT.Term) :
-  IO (Except Format (Result P.TypedIdent × Strata.SMT.EncoderState)) := do
+  (vars : List P.TypedIdent)
+  (md : Imperative.MetaData P)
+  (smtsolver filename : String)
+  (solver_options : Array String) (printFilename : Bool) :
```
shigoel, line 175: Ideally, I'd love to see `Options` or some such structure being passed in instead of `solver_options` for future proofing (e.g., we can imagine passing that in to `encodeSMT` to get configurable encodings, like we discussed elsewhere). We can use `getSolverFlags` from Core's Verifier then. Maybe not a blocker, but we know we need this.

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -125,46 +114,122 @@ def runSolver (solver : String) (args : Array String) : IO IO.Process.Output :=
...
   | .ok result => return .ok (result, estate)
 
 ---------------------------------------------------------------------
+end SMT
+
+
+/--
+SMT solver's `result` along with an SMT encoder state `estate` for a given
+verification condition `obligation`.
+Currently, this data structure is only used in the Arith example of StrataTest.
```
shigoel, line 198: Looks like this be moved into Arith then. Perhaps we can also move `VCResult` from Core's Verifier here?

Strata/DL/Imperative/SMTUtils.lean
aqjune-aws: At the moment the `Options` type is defined inside Core whereas `dischargeObligation` is in Imperative, so we will need a new data structure for this. But I fully agree that the options can be somewhere that is put inside `encodeSMT`. Indeed, the SolverM monad can be expanded to keep the list of arguments that will be passed to the commandline invocation.
Refactoring SolverM is one of my next todo items because it is really just a set of calls taking strings, but ideally we will want SolverM to take SMT.Terms (otherwise it is the caller's responsibility to pass reasonable strings representing SMT terms to the solver). Maybe I can have a look.

Strata/DL/Imperative/SMTUtils.lean
aqjune-aws: This is true. The `VCResult` depends on the `Outcome` type in Core/Verifier.lean, hence more code movements would be needed. I was wondering when writing this refactoring, should we have 'VCGen.lean' inside Imperative as well (which will have a quite big diff, as you can imagine..!). The diff in this patch was somewhat minimized to only target SMT stuffs for now.

Strata/DL/Imperative/SMTUtils.lean
shigoel: 100%

Strata/DL/Imperative/SMTUtils.lean
shigoel: Indeed! Thanks for your thoughts. Not a blocker, ofc.

aqjune-aws: This was not moved to Arith because when I was reading the tutorial I thought `VCResult` is supposed to be a data structure that must be reused across other applications. If Arith has `VCResult`, it would look slightly strange since it would give an impression of absence of universal result of verification condition. Probably before having a new VCGen.lean in Imperative, I thought it makes sense to keep it in here even if only used by the Arith example, yes...

PR #415, opened 2026-02-12
Add type unification to VCs from procedure pre/post conditions

joscoh:
*Issue #, if available:* #398

*Description of changes:* Performs type unification on generated VCs from (already typechecked) procedure pre and post conditions, so that polymorphic type variables are instantiated before SMT encoding.

### DIFF
--- a/Strata/DDM/AST.lean
+++ b/Strata/DDM/AST.lean
@@ -1070,6 +1070,7 @@ A spec for introducing a new binding into a type context.
 inductive BindingSpec (argDecls : ArgDecls) where
 | value (_ : ValueBindingSpec argDecls)
 | type (_ : TypeBindingSpec argDecls)
+| typeForward (_ : TypeBindingSpec argDecls)  -- Forward declaration (no AST node)
 | datatype (_ : DatatypeBindingSpec argDecls)
 | tvar (_ : TvarBindingSpec argDecls)
 deriving Repr
@@ -1079,6 +1080,7 @@ namespace BindingSpec
 def nameIndex {argDecls} : BindingSpec argDecls → DebruijnIndex argDecls.size
 | .value v => v.nameIndex
 | .type v => v.nameIndex
+| .typeForward v => v.nameIndex
 | .datatype v => v.nameIndex
 | .tvar v => v.nameIndex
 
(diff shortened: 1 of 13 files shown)


Strata/Languages/Core/StatementEval.lean
```diff
@@ -111,6 +113,48 @@ Get current values of global variables for old expression substitution.
 private def getCurrentGlobals (E : Env) : VarSubst :=
   E.exprEnv.state.oldest.map (fun (id, ty, e) => ((id.name, ty), e))
 
+/--
```
atomb, line 116: I'd argue that this and the following new definition belong somewhere in the Lambda directory (and probably not private there), but that's not blocking.

atomb (approved):

Strata/Languages/Core/StatementEval.lean
```diff
@@ -111,6 +113,48 @@ Get current values of global variables for old expression substitution.
...
+Compute type substitution by unifying actual argument types with input types
+and LHS types with output types.
+-/
+private def computeTypeSubst (input_tys output_tys: List LMonoTy)
+  (args : List Expression.Expr) (lhs : List Expression.Ident) (E : Env) :
+  Subst :=
+  let actual_tys := args.filterMap getExprType
+  let lhs_tys := lhs.filterMap (fun l =>
+    (E.exprEnv.state.findD l (none, .fvar () l none)).fst)
+  let input_constraints := actual_tys.zip input_tys
```
joehendrix, line 151: I asked Claude to review and it had this to say:

```
  1. (Correctness bug) filterMap + zip can misalign type constraints — StatementEval.lean:148-152

  This is the most important thing to check. In computeTypeSubst:

  let actual_tys := args.filterMap getExprType
  ...
  let input_constraints := actual_tys.zip input_tys

  filterMap drops any arg where getExprType returns none, shrinking the list. zip then pairs by position, so if arg i has no type, arg i+1's type gets paired with formal
  type i, producing an incorrect constraint. For example, if args = [a, b, c] and getExprType yields [some T1, none, some T3], the constraints become [(T1, input_0), (T3,
  input_1)] instead of the intended [(T1, input_0), (T3, input_2)]. The same issue applies to lhs_tys/output_constraints. A zipWith/pairwise approach that keeps index
  alignment (e.g., zip first, then filter pairs where the type is some) would be safer.
```

It's obvious to me why the zip is ok either.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -111,6 +113,48 @@ Get current values of global variables for old expression substitution.
...
+  let output_constraints := lhs_tys.zip output_tys
+  let constraints := input_constraints ++ output_constraints
+  match Constraints.unify constraints SubstInfo.empty with
+  | .ok substInfo => substInfo.subst
+  | .error _ => Subst.empty
```
joehendrix, line 156: More Claude:

```
 2. (Design) Silent failure on unification error — StatementEval.lean:154-156

  | .error _ => Subst.empty

  If unification fails, the code silently returns an empty substitution, meaning type variables remain uninstantiated in VCs sent to the SMT solver. Verify this is the
  desired behavior vs. propagating a diagnostic. At minimum, consider whether a warning should be emitted.
```

Strata/Languages/Core/StatementEval.lean
```diff
@@ -119,6 +163,10 @@ def Command.evalCall (E : Env) (old_var_subst : SubstMap)
     (md : Imperative.MetaData Expression) : Command × Env :=
```
joehendrix, line 163: Could you add some comments about the relationships between the variables in `old_var_subst` , `lhs` and `E`?  It's a little unclear to me which typing context the various expressions are in.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -111,6 +113,48 @@ Get current values of global variables for old expression substitution.
...
+  | .ite _ _ e1 e2 =>
+    let ty1 := getExprType e1
+    if ty1 == getExprType e2 then ty1 else none
+  | .bvar _ _ => none
+
+/--
+Compute type substitution by unifying actual argument types with input types
+and LHS types with output types.
+-/
+private def computeTypeSubst (input_tys output_tys: List LMonoTy)
```
joehendrix, line 145:
```suggestion
private def computeTypeSubst (input_tys output_tys : List LMonoTy)
```

Strata/Languages/Core/StatementEval.lean
joscoh: Basically, we are given a term `lhs := p(xs)` where `p` takes in arguments of type `input_tys` and returns arguments of types `output_tys`. `E` should contain the types for the free variables in `LHS` (guaranteed by the `StatementEval` execution). I believe that `old_var_subst` has to do with `var -> expression` maps, not type substitutions. I can add a comment explaining this.

Strata/Languages/Core/StatementEval.lean
joscoh: I kept it here because it explicitly relies on the fact that these are only valid for expressions that were already typechecked. I don't think anyone else should accidentally be calling these functions. But I can also move them if you prefer.

PR #416, opened 2026-02-12
Add provenance info to generated SMT files

atomb:
This consists of several additional commands before every `(check-sat)` command.
* `(set-info :file "<path>")` where `<path>` is the path to the original source file, extracted from metadata in the Strata artifact.
* `(set-info :start <n>)` where `<n>` is the starting character position in the original source file.
* `(set-info :stop <n>)` where `<n>` is the ending character position in the original source file.
* `(set-info :unsat-message "<msg>")` where `<msg>` is a message generated by Strata to be presented if the result of the `(check-sat)` command is `unsat`.

The changes to files in `StrataTest` are just to fix type errors caused by the refactoring in other files and probably don't need to be closely reviewed.

### DIFF
--- a/Strata/DL/Imperative/Stmt.lean
+++ b/Strata/DL/Imperative/Stmt.lean
@@ -185,6 +185,34 @@ end
 
 ---------------------------------------------------------------------
 
+/-! ### StripMetaData
+
+Functions to remove metadata from statements and blocks.
+Useful for cleaner formatting output in tests.
+-/
+
+mutual
+/-- Remove all metadata from a statement. -/
+def Stmt.stripMetaData (s : Stmt P C) : Stmt P C :=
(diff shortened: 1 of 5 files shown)


StrataTest/Languages/Python/run_py_analyze.sh
```diff
@@ -2,6 +2,8 @@
 
 failed=0
 
+(cd ../../.. && lake exe strata --help > /dev/null)
```
andrewmwells-amazon, line 5: What's the purpose of this?

atomb: This causes `strata` to be built once, at which point we don't need the overhead of `lake exe` later.

PR #417, opened 2026-02-13
Enable creating new objects

keyboardDrummer:
### Changes
- Fix a bug in SMTEncoder that would cause it to sometimes not emit all used sorts
- Fix a bug that prevented using impure expressions in opaque bodies
- Generate a single postcondition for the modifies clauses, instead of one per field in the whole program
- Simplify how we handle procedures without modifies clauses
- Enable creating new objects, and make sure these don't count for the modifies clause

### Testing
- Using object creation in `T1_MutableFields`. Checking that two created objects are not equal.
- Using object creation in `T2_ModifiesClauses`. Checking that newly allocated objects can be modified without being in the modifies clause.


ssomayyajula (commented):
No objection to the heap encoding itself nor how modifies/ensures clauses are transformed, leaving coding style etc. choices to @joscoh. Two questions:

1. Surprised that heap-parameterized procedures aren't their own data type that are `Coercion`-able to ordinary procedures, since e.g. one may forget to check `hasHeapOut` during processing and the type system won't save you. OK surprised isn't the right word, just curious if it would be a burden to make the distinction.
2. When a procedure has a `$heap_out` and no modifies clause, do you need that condition because the corresponding Core procedure will mark `$heap_out` as modified?

keyboardDrummer:
> leaving coding style etc. choices to @joscoh

Why? The PR needs two reviewers, one of which should be done by a code owner but the other can be done by you. If you approve the PR, could you approve it?

> Surprised that heap-parameterized procedures aren't their own data type that are Coercion-able to ordinary procedures, since e.g. one may forget to check hasHeapOut

How would a separate datatype prevent you from forgetting to transform the modifies clause?

> When a procedure has a $heap_out and no modifies clause, do you need that condition because the corresponding Core procedure will mark $heap_out as modified?

Then we don't need the condition. The condition is needed when there are modified clauses so we don't generate a clause that refers to a non-existent `heap_out` variable. Actually it would probably be better to emit an error when a procedure that does not modify the heap has a modifies clause, but that's not in scope of this PR.

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
```diff
@@ -55,6 +55,17 @@ procedure blockWithTwoAssignmentsInExpression() {
...
 }
+
+procedure nestedImpureStatementsAndOpaque()
+  ensures true
+{
+  var y: int := 0;
+  var x: int := y;
+  var z: int := y := y + 1;;
+    assert x == y;
+//  ^^^^^^^^^^^^^^ error: assertion does not hold
```
MikaelMayer:
```suggestion
  assert x == y;
//^^^^^^^^^^^^^^ error: assertion does not hold
```
Maybe?

StrataTest/Languages/Laurel/Examples/Objects/T1_MutableFields.lean
```diff
@@ -68,6 +63,33 @@ procedure implicitEquality(c: Container, d: Container) {
...
+  r := c#boolValue;
+}
+
+// Following test-cases can't be run because Core procedures are not transparent.
+// procedure modifiesFirst(c: Container, d: Container) returns (x: int) {
+//  c#intValue := 3;
+//  3
+// }
+
+// procedure caller() {
```
MikaelMayer, line 77: What's the plan to uncomment these procedures?

StrataTest/Languages/Laurel/Examples/Objects/T2_ModifiesClauses.lean
```diff
@@ -84,14 +86,21 @@ procedure multipleModifiesClauses(c: Container, d: Container, e: Container)
...
   modifies d
 
-procedure multipleModifiesClausesCaller(c: Container, d: Container, e: Container) {
-  assume c != d;
-  assume d != e;
-  assume c != e;
+procedure multipleModifiesClausesCaller() {
+  var c: Container := new Container;
+  var d: Container := new Container;
+  var e: Container := new Container;
```
MikaelMayer, line 92: Nice, would you be able to add assertions that `c != d && d != e && e != c` ?

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -288,6 +290,18 @@ where
...
       return ⟨ .PrimitiveOp op args', md ⟩
+    | .New _name =>
+        -- Allocate a new object: get the current counter, increment it, return the old value
+        -- 1. Save the current counter: $freshVar := Heap..counter(heapVar)
+        -- 2. Update the heap with incremented counter: heapVar := MkHeap(Heap..data(heapVar), Heap..counter(heapVar) + 1)
+        -- 3. Result is $freshVar (the old counter value, which is the new Composite reference)
+        let freshVar ← freshVarName
+        let getCounter := mkMd (.StaticCall "Heap..counter" [mkMd (.Identifier heapVar)])
+        let saveCounter := mkMd (.LocalVariable freshVar ⟨.TInt, #[]⟩ (some getCounter))
+        let newCounter := mkMd (.PrimitiveOp .Add [mkMd (.StaticCall "Heap..counter" [mkMd (.Identifier heapVar)]), mkMd (.LiteralInt 1)])
```
MikaelMayer: What about creating a prelude function that increases only the counter of the heap and calling it here?
I think it might be useful down the road if you ever need to trigger and assert "this heap is newer than this other one" so that you guarantee that counter increases

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -170,7 +170,12 @@ partial def SMT.Context.addType (E: Env) (id: String) (args: List LMonoTy) (ctx:
         ) ctx
       ) ctx
   | none =>
-    ctx.addSort { name := id, arity := args.length }
+    let ctx := ctx.addSort { name := id, arity := args.length }
+    args.foldl (fun (ctx : SMT.Context) arg =>
```
MikaelMayer: I'm not sure I understood everything but I remember it was mentioned in the PR description. Could you please point me where this change is tested?

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -23,9 +23,9 @@ This pass should run after heap parameterization, which has already:
 The frame condition states: for every object not mentioned in the modifies clause,
 all field values are preserved between the input and output heaps.
 
-For each field constant `fld`, generates:
-  forall $obj: Composite =>
-    notModified($obj) ==> readField($heap_in, $obj, fld) == readField($heap_out, $obj, fld)
+Generates:
+  forall $obj: Composite, $fld: Field =>
+    $obj < $heap_in.counter && notModified($obj) ==> readField($heap_in, $obj, $fld) == readField($heap_out, $obj, $fld)
```
MikaelMayer: Interesting ! That seems right to me. The counter is the next available reference, correct?

MikaelMayer (changes requested):
Minor questions and updates but otherwise looks good!

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
keyboardDrummer: Works!

StrataTest/Languages/Laurel/Examples/Objects/T1_MutableFields.lean
keyboardDrummer: As the above comment describes, the plan is to enable core procedures to be transparent during symbolic verification. Posted a message on Slack about that now.

StrataTest/Languages/Laurel/Examples/Objects/T2_ModifiesClauses.lean
keyboardDrummer: I could but similar assertions exist in the test `newsAreNotEqual`. I'd prefer to keep the test scopes small.

Strata/Languages/Laurel/HeapParameterization.lean
keyboardDrummer: Done

Strata/Languages/Laurel/ModifiesClauses.lean
keyboardDrummer: Yes, let me rename it to `nextReference`

Strata/Languages/Core/SMTEncoder.lean
keyboardDrummer: ~It's not separately tested. If I remove it the `T1_MutableFields.lean` Laurel test fails. I have tried to test it separately but had trouble doing so. Even if I take the printed Core created by `T1_MutableFields.lean` and verify that, I still can not reproduce the issue. The problem is that sometimes `Map A B` does not detect the type `A` and then that is not emitted in SMT, but I don't know exactly when.~

removed it. Maybe one of the refactorings, like introducing the increment function, hides the bug.

PR #385, opened 2026-02-04
Laurel Language Enhancements

fabiomadge:
## Summary

Expands the Laurel language with additional operators, data types, and verification constructs. Improves formatting across multiple dialects.

### Laurel Grammar & Translator
- New operators: `-`, `*`, `/`, `%`, `/t`, `%t` (truncating), `==>`, `!`, unary `-`
- Array support: `Array<T>` type, indexing, length
- Sequence operations: `Seq.From`, `Seq.Take`, `Seq.Drop`, `Seq.Contains`
- Constrained types with constraint injection into quantifiers
- Quantifiers: `forall`, `exists`
- Return statement works with postconditions
- While loops with multiple invariants
- Multiple `requires` clauses per procedure

### Infrastructure
- `NewlineSepBy` separator type for formatting preservation
- Java codegen: list separator preservation, `SourceRange` overloads
- Improved pretty printing for C_Simp, B3, Core

### CLI Commands
- `laurelParse`, `laurelAnalyze`, `laurelToCore`, `laurelPrint`

### Bug Fixes
- Fixed SMT encoding for multi-argument functions and `Map` type
- Added `substFvarLifting` for proper de Bruijn index handling

### Tests
- Reorganized Laurel tests T01-T17
- Added regression test for assignment lifting

PR #426, opened 2026-02-16
Infrastructure changes from PR 385

keyboardDrummer:
Contains a subset of the changes from https://github.com/strata-org/Strata/pull/385, rebased on current main.

### DDM Infrastructure
- Add `NewlineSepBy` separator and `SyntaxDef.passthrough` variant
- Replace `fromIonName?`/`toIonName` with `fromCategoryName?` for category-based lookup
- Add newline formatting case in `ArgF.mformatM`
- Update Java/Lean codegen for new constructs
- Comment parsing fix in `Parser.lean`

### Lambda/SMT Bug Fixes
- Add `liftBVars` with cutoff parameter for correct de Bruijn index shifting
- Add `substFvarLifting`/`substFvarsLifting` for substitution under binders (doc comment clarifies that `to`'s bvars must refer to binders outside `e`)
- Fix multi-argument function SMT encoding (was hardcoded to unary)
- Add Map type to SMT Array encoding

### Testing
- `LExprWFTests.lean`: tests for `substFvarLifting`
- `SMTEncoderTests.lean`: updated for multi-arg encoding
- `Functions.lean`: multi-argument function test + quantifier-in-body test

### Review feedback addressed
- The `≤` → `<` precedence change from the previous version has been removed (it was incorrect)
- Doc comment added to `substFvarLifting` per review feedback

### DIFF
--- a/Strata/DDM/Format.lean
+++ b/Strata/DDM/Format.lean
@@ -408,6 +408,13 @@ private partial def ArgF.mformatM {α} : ArgF α → FormatM PrecFormat
   | .spacePrefix =>
     .atom <$> entries.foldlM (init := .nil) fun p a =>
       return (p ++ " " ++ (← a.mformatM).format)
+  | .newline =>
+    if z : entries.size = 0 then
+      pure (.atom .nil)
+    else do
+      let f i q s := return s ++ "\n" ++ (← entries[i].mformatM).format
+      let a := (← entries[0].mformatM).format
+      .atom <$> entries.size.foldlM f (start := 1) a
 
 private partial def ppArgs (f : StrataFormat) (rargs : Array Arg) : FormatM PrecFormat :=
   if rargs.isEmpty then
--- a/Strata/DL/Lambda/LExprWF.lean
+++ b/Strata/DL/Lambda/LExprWF.lean
@@ -290,11 +290,24 @@ theorem varOpen_of_varClose {T} {GenericTy} [BEq T.Metadata] [LawfulBEq T.Metada
 /-! ### Substitution on `LExpr`s -/
 
 /--
-Substitute `(.fvar x _)` in `e` with `s`. Note that unlike `substK`, `varClose`,
-and `varOpen`, this function is agnostic of types.
+Increment bound variable indices in `e` by `n`. Only bvars at or above `cutoff`
+are shifted; bvars below `cutoff` (bound within `e`) are left alone. The cutoff
+increases when going under binders.
+-/
+def liftBVars (n : Nat) (e : LExpr ⟨T, GenericTy⟩) (cutoff : Nat := 0) : LExpr ⟨T, GenericTy⟩ :=
+  match e with
+  | .const _ _ => e | .op _ _ _ => e | .fvar _ _ _ => e
+  | .bvar m i => if i >= cutoff then .bvar m (i + n) else e
+  | .abs m ty e' => .abs m ty (liftBVars n e' (cutoff + 1))
+  | .quant m qk ty tr' e' => .quant m qk ty (liftBVars n tr' (cutoff + 1)) (liftBVars n e' (cutoff + 1))
+  | .app m fn e' => .app m (liftBVars n fn cutoff) (liftBVars n e' cutoff)
+  | .ite m c t e' => .ite m (liftBVars n c cutoff) (liftBVars n t cutoff) (liftBVars n e' cutoff)
+  | .eq m e1 e2 => .eq m (liftBVars n e1 cutoff) (liftBVars n e2 cutoff)
 
-Also see function `subst`, where `subst s e` substitutes the outermost _bound_
-variable in `e` with `s`.
+/--
+Substitute `(.fvar x _)` in `e` with `to`. Does NOT lift de Bruijn indices in `to`
+when going under binders - safe when `to` contains no bvars (e.g., substituting
+fvar→fvar). Use `substFvarLifting` when `to` contains bvars.
 -/
 def substFvar [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (fr : T.Identifier) (to : LExpr ⟨T, GenericTy⟩)
   : (LExpr ⟨T, GenericTy⟩) :=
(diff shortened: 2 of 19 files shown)


Strata/DDM/Format.lean
```diff
@@ -318,7 +318,7 @@ private def SyntaxDefAtom.formatArgs (opts : FormatOptions) (args : Array PrecFo
   match stx with
   | .ident lvl prec _ =>
     let ⟨r, innerPrec⟩ := args[lvl]!
-    if prec > 0 ∧ (innerPrec ≤ prec ∨ opts.alwaysParen) then
+    if prec > 0 ∧ (innerPrec < prec ∨ opts.alwaysParen) then
```
joehendrix: **Precedence change breaks associativity and same-precedence parenthesization**

The change from `innerPrec ≤ prec` to `innerPrec < prec` causes the formatter to drop parentheses that are needed to preserve expression structure.

**1. Non-associative operators at the same precedence**

B3 `==` and `!=` are both `@[prec(15)]` with no associativity annotation. For non-associative binary operators, both argument positions get required `prec = 15`, and both operators output precedence 15.

Formatting `(a == b) != c` — i.e. `not_equal(equal(a, b), c)`:

| Check | `prec` | `innerPrec` | Result | Output |
|-------|--------|-------------|--------|--------|
| Old `≤` | 15 | 15 | `15 ≤ 15` → parens | `(a == b) != c` ✓ |
| New `<` | 15 | 15 | `15 < 15` → no parens | `a == b != c` ✗ |

The output `a == b != c` is genuinely ambiguous — there's no associativity convention to disambiguate it.

**2. Associative operators**

B3 `==>` is `@[prec(5), rightassoc]`. The DDM precedence convention (`DialectM.lean:560-576`) assigns:
- Left arg (non-associative side): required `prec = 5`
- Right arg (associative side): required `prec - 1 = 4`
- Output precedence: `5`

Formatting `(true ==> false) ==> true` — i.e. `implies(implies(true, false), true)`:

| Check | `prec` | `innerPrec` | Result | Output |
|-------|--------|-------------|--------|--------|
| Old `≤` | 5 | 5 | `5 ≤ 5` → parens | `(true ==> false) ==> true` ✓ |
| New `<` | 5 | 5 | `5 < 5` → no parens | `true ==> false ==> true` ✗ |

Since `==>` is right-associative, `true ==> false ==> true` parses as `true ==> (false ==> true)` — a different logical proposition. The same issue affects left-associative operators symmetrically (e.g. B3 `<==`).

**Root cause:** The DDM precedence convention lowers the associative side by 1 and relies on `≤` triggering parens when `innerPrec == prec` on the non-associative side. Changing to `<` breaks that invariant.

Strata/DL/Lambda/LExprWF.lean
```diff
@@ -307,6 +320,28 @@ def substFvar [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (fr : T.Identifier)
...
+binders. Use this when `to` contains bound variables that should be preserved.
+-/
+def substFvarLifting [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (fr : T.Identifier) (to : LExpr ⟨T, GenericTy⟩)
+  : (LExpr ⟨T, GenericTy⟩) :=
+  go e 0
+where
+  go (e : LExpr ⟨T, GenericTy⟩) (depth : Nat) : LExpr ⟨T, GenericTy⟩ :=
+    match e with
+    | .const _ _ => e | .bvar _ _ => e | .op _ _ _ => e
+    | .fvar _ name _ => if name == fr then liftBVars depth to else e
```
joehendrix, line 339: **`substFvarLifting` assumes `to` has no bvars referring to binders inside `e`**

The call `liftBVars depth to` uses the default cutoff of 0, which shifts *all* bvars in `to`. This is incorrect when `to` has bvars that refer to binders shared with `e`.

Concrete example:

```
e  = quant(app(fvar "x", bvar 0))    -- bvar 0 bound by the quant
to = bvar 0                           -- intended to also refer to the quant
```

`substFvarLifting e "x" (bvar 0)` descends under the quant (`depth = 1`), then calls `liftBVars 1 (bvar 0)` with cutoff 0. Since `0 ≥ 0`, the bvar is shifted to `bvar 1` — but it should remain `bvar 0` since it refers to the quant binder right above.

Adding a `cutoff` parameter (defaulting to 0) and passing it through to `liftBVars` would handle this. The current call site in `SMTEncoder.lean:539` is safe because its `to` values (`bvar 0`, `bvar 1`, ...) always refer to the outer function definition, not to binders inside `body`.

This doesn't need to block the PR — at minimum, a doc comment noting the restriction (that `to`'s bvars must be external to `e`) would prevent misuse.

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -93,6 +93,7 @@ private def lMonoTyToSMTString (ty : LMonoTy) : String :=
   | .tcons "real" [] => "Real"
   | .tcons "string" [] => "String"
   | .tcons "regex" [] => "RegLan"
+  | .tcons "Map" [k, v] => s!"(Array {lMonoTyToSMTString k} {lMonoTyToSMTString v})"
```
aqjune-aws: I think the user must be able to choose which encoding option to use for `Map`. One option is `Array`, but another option is to use the axiomatized expression in SMT. This is being tracked in Asana & I am working on this too.

Strata/DL/Lambda/LExprWF.lean
```diff
@@ -307,6 +320,28 @@ def substFvar [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (fr : T.Identifier)
...
+    | .app m fn e' => .app m (go fn depth) (go e' depth)
+    | .ite m c t f => .ite m (go c depth) (go t depth) (go f depth)
+    | .eq m e1 e2 => .eq m (go e1 depth) (go e2 depth)
+
+def substFvarsLifting [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (sm : Map T.Identifier (LExpr ⟨T, GenericTy⟩))
+  : LExpr ⟨T, GenericTy⟩ :=
+  List.foldl (fun e (var, s) => substFvarLifting e var s) e sm
+
 def substFvars [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (sm : Map T.Identifier (LExpr ⟨T, GenericTy⟩))
   : LExpr ⟨T, GenericTy⟩ :=
```
aqjune-aws, line 351: Can we have extra `panic!` here checking that the range of `sm` does not include any `.bvar`?

Strata/DL/Lambda/LExprWF.lean
```diff
@@ -307,6 +320,28 @@ def substFvar [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (fr : T.Identifier)
...
+    | .const _ _ => e | .bvar _ _ => e | .op _ _ _ => e
+    | .fvar _ name _ => if name == fr then liftBVars depth to else e
+    | .abs m ty e' => .abs m ty (go e' (depth + 1))
+    | .quant m qk ty tr' e' => .quant m qk ty (go tr' (depth + 1)) (go e' (depth + 1))
+    | .app m fn e' => .app m (go fn depth) (go e' depth)
+    | .ite m c t f => .ite m (go c depth) (go t depth) (go f depth)
+    | .eq m e1 e2 => .eq m (go e1 depth) (go e2 depth)
+
+def substFvarsLifting [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (sm : Map T.Identifier (LExpr ⟨T, GenericTy⟩))
+  : LExpr ⟨T, GenericTy⟩ :=
```
aqjune-aws, line 347: This function will return an ill-formed expression because `WF` in LExprWF.lean states:
```
An `LExpr e` is well-formed if it has no dangling bound variables.
```
but if `sm`'s range has a bound variable, after lifting there will be dangling bound variables. The `LExpr`s are designed to have locally nameless representation: https://chargueraud.org/research/2009/ln/main.pdf

Let's make it clear that `substFvarsLifting` may break wellformedness.

keyboardDrummer:
Leaving the comments for @fabiomadge since I extracted this code from his PR.

Strata/Languages/Core/SMTEncoder.lean
aqjune-aws: A proposed patch with an extra option for this feature is at #441 .

fabiomadge:
Rebased on current main and addressed all review feedback:

- **Precedence bug (Format.lean):** The `≤` → `<` change has been removed. The new branch was built from jverify-strata-backend which never had this change.
- **`substFvarLifting` doc (LExprWF.lean):** Added a doc comment clarifying that `to`'s bvars must refer to binders *outside* `e`, not within it.
- **Map encoding (SMTEncoder.lean):** Already resolved by #441 (`useArrayTheory` option), which is now in main.
- **`panic!` on bvars in `substFvarsLifting`:** Not added — the only call site (SMTEncoder.lean:587) intentionally passes bvars (`bvar 0`, `bvar 1`, ...) as substitution targets. A panic would break it. The doc comment documents the contract instead.
- **DDM grammar spacing changes:** Dropped from this PR (stale against current main). Can be a separate follow-up.
- **Tests:** Added `LExprWFTests.lean`, updated `SMTEncoderTests.lean`, added multi-arg and quantifier-body tests to `Functions.lean`.

The branch has been force-pushed with a clean single commit based on current main. All 406 tests pass.

Strata/DL/Lambda/LExprWF.lean
aqjune-aws: Hi Fabio, do you want to add the explicit statement about WF to the comment of `substFvarLifting` above, to help readers expect when `substFvarLifting` can be used?

Strata/DL/Lambda/LExprWF.lean
aqjune-aws: It seems the sanity check is still missing - do you want to add it?

Strata/DL/Lambda/LExprWF.lean
fabiomadge: The doc comment on `substFvarLifting` already states: "Any bvars in `to` must refer to binders *outside* `e`, not to binders within `e`." I think this covers the WF concern — if callers violate this contract, the result will have dangling bvars. Do you think the existing wording is insufficient, or would you like a specific mention of the locally-nameless WF invariant?

Strata/DL/Lambda/LExprWF.lean
fabiomadge: We can't add a `panic!` here — the only call site (`SMTEncoder.lean:586`) intentionally passes bvars as substitution values (`bvar 0`, `bvar 1`, ...) to replace formals with de Bruijn indices in function bodies. A panic would break it. The contract is documented on `substFvarLifting` above: bvars in the replacement must refer to binders outside `e`.

aqjune-aws: It is right, it is temporarily breaking the WF criteria here. Updating it to keep the `WF`  property after each invocation of calls seems out of scope...

Strata/DL/Lambda/LExprWF.lean
aqjune-aws: I think I understood the motivation better in the sense that LExpr might not be locally-nameless in certain cases outside these calls.

Issue #419, opened 2026-02-13
Core VCG: goto inside if-then-else causes path multiplication across procedures


fabiomadge:
## Problem

When a procedure contains a `goto` inside an `if` branch, the VCG produces multiple `EnvWithNext` results (one per branch). These multiple results propagate through `ProgramEval`, causing subsequent procedures to be verified once per path from the previous procedure.

### Reproduction

```
program Core;

procedure first(a : int) returns (r : int)
spec { }
{
  if (a > 0) {
    r := 1;
    goto done;
  }
  r := 0;
  done: { }
};

procedure second(a : int) returns (r : int)
spec { }
{
  assert a > 0;
  r := a;
};
```

Expected: `assert_0` fails once.
Actual: `assert_0` fails twice (once per path from `first`).

### Root cause

In `StatementEval.lean`, `processIteBranches` has a special case (line ~530) that merges both branches into a single result when neither has a `nextLabel`. When one branch has a `goto` (`nextLabel := some l`), it falls through to the general case (line ~539) which returns `Ewns_t ++ Ewns_f` — separate paths.

These separate paths propagate up through:
1. `evalAuxGo` line ~497: `List.flatMap` continues remaining statements for each path
2. `Procedure.eval`: returns `List (Procedure × Env)` with multiple entries
3. `ProgramEval` line ~77: `pEs.flatMap` evaluates remaining declarations for each entry

Note: `Procedure.evalOne` (line ~84) explicitly expects a single result and errors on multiple, suggesting this multiplication is unintended.

### Proposed fix

In `processIteBranches`, when the general case produces multiple results, group them by `nextLabel` and merge environments within each group using `Env.merge`:

1. Partition results from each branch into goto paths (`nextLabel.isSome`) and normal paths
2. Merge normal paths from both branches using `Env.merge cond' E_t E_f` (same as the existing special case)
3. Merge goto paths targeting the same label using `Env.merge`
4. Return at most one result per distinct `nextLabel`

This ensures an `ite` with gotos produces at most N results (one per distinct target label + one for fall-through), not a product of branch paths.

### Impact

This blocks using `goto` for early return encoding in Laurel-to-Core translation. Currently Laurel uses `assume false` to model early returns, which is incorrect for execution and requires duplicating postcondition assertions at each return site. The `goto` approach is cleaner but unusable until this is fixed.

fabiomadge:
[...]

PR #429, opened 2026-02-16
Fix for issue 419

keyboardDrummer:
Almost purely LLM generated based on a bug report (https://github.com/strata-org/Strata/issues/419). I don't understand the original code well enough to understand this fix. This PR needs to be carefully reviewed.

### Changes
- Changed programEval so it calls `Procedure.evalOne` instead of `Procedure.eval`, which prevents the bug where one procedure causes a subsequent procedure to start with multiple paths.
- Fix so that a procedure ends with a single path even when jumping from a conditional
- Fix to the cover implementation to prevent the cover statement from causing a procedure to return with multiple paths.

### DIFF
--- a/Strata/Languages/Core/StatementEval.lean
+++ b/Strata/Languages/Core/StatementEval.lean
@@ -395,8 +395,8 @@ def processGoto : Statements → Option String → (Statements × Option String)
 
 /--
 Merge `EnvWithNext` results that originated from different branches of the same
-if-then-else and have reconverged to the same `nextLabel`. This prevents path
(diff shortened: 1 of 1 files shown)


StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -32,27 +32,27 @@ procedure Test() returns ()
 
 /--
 info:
-Obligation: unreachable_cover1
```
keyboardDrummer, line 35: It's just the order that is changed here

PR #430, opened 2026-02-16
Fix #420, bug in translation of function decl statement

### LINKED ISSUE #420: Use of function decl statement gives type error
Given the following program:
```
import Strata.Languages.Core.Core
import Strata.Languages.Core.DDMTransform.Translate

open Core
open Strata

def translate (t : Strata.Program) : Core.Program :=
  (TransM.run Inhabited.default (translateProgram t)).fst

def simpleFuncDeclPgm :=
#strata
program Core;

procedure test() returns ()
{
  var x : int := 1;
  function addX(y : int) : int
  { y + x }
  var z : int := addX(5);
};

#end

#eval (Std.format (Core.typeCheck Options.default (translate simpleFuncDeclPgm)))
```
The #eval panics with `Cannot infer the type of this operation: addX`.

joscoh:
*Issue #, if available:* #420

*Description of changes:*
There were two issues with the translation of function decl statements:
1. The local binding was getting added as a free var, not a bound var, which caused the bound var index lookups to be inconsistent.
2. Because bvars and fvars are looked up differently, the function decl was being stored with the wrong type.

Also includes functions to strip metadata from statements, procedures, and programs to make test output nicer.

### DIFF
--- a/StrataTest/Languages/Core/FuncDeclStmtTest.lean
+++ b/StrataTest/Languages/Core/FuncDeclStmtTest.lean
@@ -38,9 +38,9 @@ ok: procedure test :  () → ()
 {
   {
     init (x : int) := #1
-    #[<[fileRange]: :407-447>] funcDecl <function>
+    funcDecl <function>
     init (z : int) := ((~addX : (arrow int int)) #5)
   }
 }-/
 #guard_msgs in
-#eval (Std.format (Core.typeCheck Options.default (translate simpleFuncDeclPgm)))
+#eval (Std.format ((Core.typeCheck Options.default (translate simpleFuncDeclPgm).stripMetaData)))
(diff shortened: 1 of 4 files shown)


StrataTest/Languages/Core/FuncDeclStmtTest.lean
```diff
@@ -0,0 +1,46 @@
...
+---
+info:
+ok: procedure test :  () → ()
+  modifies: []
+  preconditions: 
+  postconditions: 
+{
+  {
+    init (x : int) := #1
+    #[<[fileRange]: :407-447>] funcDecl <function>
```
MikaelMayer: I would prefer for inline tests to not  display file ranges, as they prevent reordering, merging, etc. Is that something you can redact?

joscoh: I think the only way to do this is to strip the metadata from statements. I can add a function to do this.

PR #432, opened 2026-02-16
feat: Support optional RHS in Cmd.init for unconstrained variable declarations

MikaelMayer:
Makes `Cmd.init` RHS optional, enabling variable declarations without initial values (unconstrained/havoc semantics). Previously, `init` always required an expression, and uninitialized variables used dummy `init_...` free variables as a workaround.

With the optional RHS, `init x : int` with `none` directly represents an unconstrained variable — no dummy free variables needed.

## Changes

- `Cmd.init` signature: `(e : P.Expr)` → `(e : Option P.Expr)`
- Type checking, evaluation, and semantics handle both cases
- C_Simp `init_decl` translation uses `none` instead of generating dummy `init_...` free variables
- Removed unused `initVarValue` from C_Simp translate
- Fixed a bug in `ASTtoCST.lean` where the DDM formatter used a heuristic that accidentally hid real initialization values: it checked whether the init RHS was a free variable in scope, and if not, suppressed it — causing legitimate initializations (call elimination temp vars, loop measure vars) to be silently hidden in formatted output (see https://github.com/strata-org/Strata/pull/432/changes#diff-971209050e05a2cb431875f3d238398c9012c55d2b6264b0e44e91e941d1a140R76)
- Updated test expectations for corrected formatting and variable numbering

Existing tests pass, no new tests needed.

### DIFF
--- a/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
+++ b/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
@@ -4,7 +4,7 @@
   SPDX-License-Identifier: Apache-2.0 OR MIT
 -/
 
--- Minimal Laurel dialect for AssertFalse example
+-- Laurel dialect definition, loaded from LaurelGrammar.st
 import Strata.DDM.Integration.Lean
 
 namespace Strata.Laurel


Strata/Languages/Core/ProgramEval.lean
```diff
@@ -28,8 +28,8 @@ structure DeclsEnv where
 
 def initStmtToGlobalVarDecl (s : Statement) : Decl :=
   match s with
-  | .init x ty e md => (.var x ty e md)
-  | _ => panic s!"Expected a variable initialization; found {format s} instead."
+  | .init x ty (some e) md => (.var x ty e md)
+  | _ => panic s!"Expected a variable initialization with expression; found {format s} instead."
```
MikaelMayer: This is one place where I would have a doubt, but it's at least coherent.

MikaelMayer (commented):
I reviewed all files by hand and after asking the kiro-cli to review them.

Strata/DL/Imperative/Cmd.lean
```diff
@@ -34,9 +34,8 @@ Commands don't create local control flow, and are typically used as a parameter
 to `Imperative.Stmt` or other similar types.
 -/
 inductive Cmd (P : PureExpr) : Type where
-  /-- Define a variable called `name` with type `ty` and initial value `e`.
-    Note: we may make the initial value optional. -/
-  | init     (name : P.Ident) (ty : P.Ty) (e : P.Expr) (md : (MetaData P) := .empty)
+  /-- Define a variable called `name` with type `ty` and optional initial value `e`. -/
```
shigoel: A comment here what the semantics are when the initial value isn't provided would be welcome (similar to what you have in CmdEval.lean).

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -764,21 +764,13 @@ partial def stmtToCST {M} [Inhabited M] (s : Core.Statement)
...
-      | none => do
-        -- If this free variable isn't in scope, it's autogenerated in
-        -- Core (i.e., default/havoc init value).
-        let bind := Bind.bind_mk default nameAnn
-                    ⟨default, none⟩ tyCST
-        let dl := DeclList.declAtom default bind
-        pure (.varStatement default dl)
-    | _ => -- not an .fvar
-      let exprCST ← lexprToExpr expr 0
+    | none => do
```
shigoel, line 767: Nice! Glad that workaround went away.

Strata/Languages/Core/ProgramEval.lean
shigoel: I realized that we don't allow initialization of global variables in Core (at least in concrete syntax; and we shouldn't allow that in abstract syntax too). See `command_var` in Grammar.lean. It makes sense to me to change `Decl`'s `var` to remove the RHS too, which would mean that here we'd panic if `none` was not found in the RHS.

.gitignore
```diff
@@ -5,4 +5,3 @@
 vcs/*.smt2
 
 Strata.code-workspace
-
```
joscoh, line 8: Not very important but IMO we should try to avoid whitespace changes to unrelated files

MikaelMayer: Yes agree because otherwise it's always a back and forth.
Note: The agent had added a file to gitignore and I instructed it to remove the file and put it in git info exclude instead.

shigoel (approved):

joscoh (approved):

PR #434, opened 2026-02-17
Add documentation generation for Laurel

keyboardDrummer:
### Changes
- Add documentation for Laurel

### Testing
- Locally looked at the generated docs. After merging this should become available on https://strata-org.github.io/Strata/

### DIFF
--- a/Strata/Languages/Core/FactoryWF.lean
+++ b/Strata/Languages/Core/FactoryWF.lean
@@ -17,50 +17,63 @@ import Strata.DL.Lambda.IntBoolFactory
 namespace Core
 open Lambda
 
-set_option maxRecDepth 32768 in
-set_option maxHeartbeats 4000000 in
+theorem array_list_happend_toArray:
+  ∀ {α:Type} (a:Array α) (b:List α), a ++ b = (a.toList ++ b).toArray
+  := by
+  unfold HAppend.hAppend Array.instHAppendList Array.appendList
+  grind
+
+set_option maxRecDepth 1024 in
+set_option maxHeartbeats 400000 in
 /--
 Wellformedness of Factory
 -/
 theorem Factory_wf :
     FactoryWF Factory := by
+/- AL2: skip -/
   unfold Factory
+  simp only [array_list_happend_toArray, List.cons_append, List.nil_append]
   apply FactoryWF.mk
   · decide -- FactoryWF.name_nodup
-  · unfold HAppend.hAppend Array.instHAppendList
-    simp only []
-    unfold Array.appendList
-    simp only [List.foldl, Array.push, List.concat]
-    intros lf
-    rw [← Array.mem_toList_iff]
-    simp only []
-    intros Hmem
-    repeat (
-      rcases Hmem with _ | ⟨ a', Hmem ⟩
-      · rw [LFuncWF]
-        apply Strata.DL.Util.FuncWF.mk
-        · decide -- LFuncWF.arg_nodup
-        · decide -- LFuncWF.body_freevars
-        · -- LFuncWF.concreteEval_argmatch
-          simp (config := { ground := true })
-          try (
-            try unfold unOpCeval
-            try unfold binOpCeval
-            try unfold cevalIntDiv
-            try unfold cevalIntMod
-            try unfold cevalIntDivT
-            try unfold cevalIntModT
-            try unfold bvUnaryOp
-            try unfold bvBinaryOp
-            try unfold bvShiftOp
-            try unfold bvBinaryPred
-            intros lf md args res
-            repeat (rcases args with _ | ⟨ args0, args ⟩ <;> try grind))
-        · decide -- LFuncWF.body_or_concreteEval
-        · decide -- LFuncWF.typeArgs_nodup
-        · decide -- LFuncWF.inputs_typevars_in_typeArgs
-        · decide -- LFuncWF.output_typevars_in_typeArgs
+  · intros f Hmem
+    -- 176 is the number of functions in Factory (#eval Factory.size)
+    iterate 176 (any_goals (rcases Hmem with _ | ⟨ a', Hmem ⟩ <;> try contradiction))
+    all_goals (
+      rw [LFuncWF]
+      apply Strata.DL.Util.FuncWF.mk
+      · decide -- LFuncWF.arg_nodup
+      · decide -- LFuncWF.body_freevars
+      · -- LFuncWF.concreteEval_argmatch
+        intros lf md args res
+        -- Reduce '<func name>.concreteEval'
+        conv => lhs; simp (config := { ground := true })
+        -- Reduce 'List.length <func name>.inputs'
+        conv => rhs; rhs; rhs; whnf
+        try (solve | intro h; contradiction)
+        try (
+          try unfold unOpCeval
+          try unfold binOpCeval
+          try unfold cevalIntDiv
+          try unfold cevalIntMod
+          try unfold cevalIntDivT
+          try unfold cevalIntModT
+          try unfold bvUnaryOp
+          try unfold bvBinaryOp
+          try unfold bvShiftOp
+          try unfold bvBinaryPred
+          intro Hlf_def
+          rw [← Hlf_def]
+          -- Destruct the 'args' list until the goal is discharged.
+          repeat (rcases args with _ | ⟨ args0, args ⟩ <;> try (
+            conv => lhs; lhs; simp only []
+            intros Habsurd
+            contradiction))
+          -- When the [arg0,arg1,..].length = n exactly matches
+          intro _Hdummy; rfl)
+      · decide -- LFuncWF.body_or_concreteEval
+      · decide -- LFuncWF.typeArgs_nodup
+      · decide -- LFuncWF.inputs_typevars_in_typeArgs
+      · decide -- LFuncWF.output_typevars_in_typeArgs
     )
-    contradiction
-
+/- AL2: skip end -/
 end Core


docs/verso/index.html
```diff
@@ -32,6 +32,10 @@ <h2>DDM Documentation</h2>
             <h2>Strata Core Language Definition Documentation</h2>
             <p>Documentation for Strata Core language definition.</p>
         </a>
+        <a href="laurel/html-single/" class="doc-card">
+            <h2>Laurel Language Documentation</h2>
+            <p>Documentation for the Laurel intermediate verification language. Laurel attempts to provide features that are common between Java, Python, and JavaScript.</p>
```
tautschnig:
```suggestion
            <p>Documentation for the Laurel intermediate verification language. Laurel attempts to provide features that are common to Java, Python, and JavaScript.</p>
```
(but then English isn't my first language)

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+set_option pp.rawOnError true
+
+#doc (Manual) "The Laurel Language" =>
+%%%
+shortTitle := "Laurel"
+%%%
+
+# Introduction
+
+Laurel is an intermediate verification language designed to serve as a target for popular garbage collected languages that include imperative features, such as Java, Python, and JavaScript. Laurel tries to include any features that are common between those three languages.
```
tautschnig: 1. Is "garbage collected" really the defining criterion? If so, why?
2. 
```suggestion
Laurel is an intermediate verification language designed to serve as a target for popular garbage-collected languages that include imperative features, such as Java, Python, and JavaScript. Laurel tries to include any features that are common to those three languages.
```

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+- Model checking
+- Property based testing
+- Data-flow analysis
+
+Here are some Laurel language features that are shared between the source languages:
+- Statements such as loops and return statements
+- Mutation of variables, including in expressions
+- Reading and writing of fields of references
+- Object oriented concepts such as inheritance, type checking, up and down casting and dynamic dispatch
+- Error handling
```
tautschnig:
```suggestion
- Error handling via exceptions
```

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+- Higher order procedures and procedure types
```
tautschnig:
```suggestion
- Higher-order procedures and procedure types
```

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+- Parameteric polymorphism
```
tautschnig:
```suggestion
- Parametric polymorphism
```

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+
+On top of the above features, Laurel adds features that are useful specifically for verification:
```
tautschnig, line 55: Add loop invariants? Are there also "decreases" clauses for loops? Is there some support for non-deterministic choice, and for havocking?

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+- Assert and assume statements
+- Pre and postconditions for procedures
+- Modifies and reads clauses for procedures
+- Decreases clauses for procedures
+- Immutable fields and constructors that support assigning to them
+- Constrained types
+- Type invariants
+- Forall and exists expressions
+- Old and fresh expressions
+- Various constructs for writing proofs
```
tautschnig: I don't know what this means.

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+
+## Design Choices
+
+Laurel makes several design choices:
+
+- Procedures: instead of separate (functional) functions and (imperative) procedures, Laurel has a single
+  general concept called a *procedure*.
+- Determinism: procedures can be marked as deterministic or nondeterministic.
+  For deterministic procedures with a non-empty reads clause, the result can be
+  assumed unchanged if the read references are the same.
```
tautschnig: I don't understand this part. First, it doesn't explain at all what the effect of marking a procedure "nondeterministic" is. Second, it only even tries to cover a sub-case of the "deterministic" case. Third, I don't follow the explanation of that sub-case: "are the same" - same as what; "assumed unchanged" - unchanged compared to what?

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+- Opacity: procedures can have a body that is transparent or opaque. Only an
+  opaque body may declare a postcondition.
```
tautschnig: So ... procedures can also _not_ have a body? If so, what are the semantics? Is this transparency a verification hint or does it alter the program semantics?

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+- Unified StmtExpr: statements and expressions share a single type, reducing
+  duplication for constructs like conditionals and variable declarations.
```
tautschnig: What sort of "type" is this talking about? Is this a Strata/Lean type, or a Laurel type? Also, I find this decision quite surprising: my take would have been that expressions have a Laurel type (e.g., integer), while statements do not have a Laurel type.

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+types.
+
+Composite types have fields and procedures, and may extend other composite
+types. Fields declare whether they are mutable and specify their type.
+
+{docstring Strata.Laurel.CompositeType}
+
+{docstring Strata.Laurel.Field}
+
+Constrained types are defined by a base type and a constraint over that type.
```
tautschnig:
```suggestion
Constrained types are defined by a base type and a constraint over the values of the base type.
```

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+
+{docstring Strata.Laurel.ConstrainedType}
+
+{docstring Strata.Laurel.TypeDefinition}
+
+# Expressions and Statements
+
+Laurel uses a unified `StmtExpr` type that contains both expression-like and
+statement-like constructs. This avoids duplication of shared concepts such as
+conditionals and variable declarations.
```
tautschnig: Why do conditionals or variable declarations make any use of statement-like constructs in the first place?

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+{docstring Strata.Laurel.Parameter}
+
+{docstring Strata.Laurel.Determinism}
+
+{docstring Strata.Laurel.Body}
+
+# Programs
+
+A Laurel program consists of static procedures, static fields, type
+definitions, and constants.
```
tautschnig: What does "static" mean here, and why is it important?

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+Procedures that write the heap receive both an input and output heap parameter.
+Procedures that only read the heap receive an input heap parameter. Field reads
+and writes are rewritten to use `readField` and `updateField` functions.
+
+## Modifies Clauses
+
+The modifies clause transformation generates frame conditions for procedures
+with modifies clauses. For each field constant, it generates a universally
+quantified assertion that objects not mentioned in the modifies clause have
+their field values preserved between the input and output heaps.
```
tautschnig: What is a "field constant"?

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+
+## Lifting Expression Assignments
+
+The expression assignment lifting pass transforms assignments that appear in
+expression contexts into preceding statements. This is necessary because Strata
+Core does not support assignments within expressions.
```
tautschnig: Are there any other side-effects that need to be lifted, e.g., procedure calls within expressions?

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,187 @@
...
+
+## Translation to Core
+
+The final translation converts Laurel types, expressions, statements, and
+procedures into their Strata Core equivalents. Procedures with expression bodies are translated as Core functions, while other procedures become Core procedures.
```
tautschnig: What is an "expression bod[y]"?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -10,10 +10,11 @@ import Strata.Languages.Core.Procedure
 import Strata.Util.Tactics
 
 /-
-The Laurel language is supposed to serve as an intermediate verification language for at least Java, Python, JavaScript.
+Laurel is an intermediate verification language designed to serve as a target for popular garbage collected languages that include imperative features, such as Java, Python, and JavaScript. Laurel tries to include any features that are common between those three languages.
```
tautschnig: Many of the same comments that I gave above also apply to this file, not repeating them.

docs/verso/LaurelDoc.lean
keyboardDrummer: > Is "garbage collected" really the defining criterion?

The defining criterion is that we want to include features that are shared among the most popular programming languages. Regarding supporting verification of manual memory management, I think it's open to change but in the short term I don't see us adding support for it since it would be quite some effort.

docs/verso/LaurelDoc.lean
keyboardDrummer: What this'll be concretely is to be designed, but it would be ways to specify which code is visible when proving a particular VC. Something like hide and reveal statements, proof blocks, opaque blocks.

docs/verso/LaurelDoc.lean
keyboardDrummer: I've clarified the text.

docs/verso/LaurelDoc.lean
keyboardDrummer: Clarified

docs/verso/LaurelDoc.lean
keyboardDrummer: Clarified

docs/verso/LaurelDoc.lean
keyboardDrummer: What I'm saying is that for example a conditional only needs to be defined once, instead of once for the expression language and once for the statement language.

docs/verso/LaurelDoc.lean
keyboardDrummer: Clarified

docs/verso/LaurelDoc.lean
keyboardDrummer: We should also lift calls to non-deterministic procedures, but we don't do that yet.

For calls to deterministic procedures with statement bodies that occur in expressions, Laurel wants to support those but currently does not. How to support them is still a topic of discussion. One option is to translate the statement bodies of such procedures to expressions and convert them to Core functions. Another option is to change Core.

docs/verso/LaurelDoc.lean
keyboardDrummer: Bodies with constructs that fit what Core expressions support.

docs/verso/LaurelDoc.lean
keyboardDrummer: > Is there some support for non-deterministic choice, and for havocking?

Right now neither of those two are explicitly available. Calling a non-deterministic procedure without a body, like a file read operation, is the source of non-determinism.

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,160 @@
...
+- Constrained types
+- Type invariants
+- Forall and exists expressions
+- Old and fresh expressions
+- Unbounded integer and real types
+- To be designed constructs for supporting proof writing
+
+## Implementation Choices
+
+Here are some design choices that are more on the implementation than on the user facing level:
```
joscoh: I don't quite know what this sentence means.

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,160 @@
...
+
+- Procedures: instead of separate (functional) functions and (imperative) procedures, Laurel has a single general concept called a *procedure*.
```
joscoh: (functional) -> (pure)?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -10,190 +10,245 @@ import Strata.Languages.Core.Procedure
...
-  | TSet (elementType : WithMetadata HighType) /- Set type, e.g. Set Composite. Used in modifies clauses. -/
+  /-- 64-bit floating point type. Required for JavaScript (`number`), also used by Python (`float`) and Java (`double`). -/
+  | TFloat64
+  /-- String type for text data. -/
+  | TString
+  /-- Internal type representing the heap. Introduced by the heap parameterization pass; not accessible via grammar. -/
+  | THeap
+  /-- Internal type for a field constant with a known value type. Introduced by the heap parameterization pass; not accessible via grammar. -/
+  | TTypedField (valueType : WithMetadata HighType)
+  /-- Set type, e.g. `Set Composite`. Used in modifies clauses. -/
```
joscoh: Can a `Composite` be referred to at the user level or is it just an implementation detail?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -10,190 +10,245 @@ import Strata.Languages.Core.Procedure
...
+/--
+Specifies whether a procedure is deterministic or nondeterministic.
+
+For deterministic procedures with a non-empty reads clause, the result can be
+assumed unchanged if the read references are the same.
+-/
 inductive Determinism where
+  /-- A deterministic procedure. The optional reads clause lists the heap locations the procedure may read. -/
   | deterministic (reads : Option (WithMetadata StmtExpr))
+  /-- A nondeterministic procedure. -/
```
joscoh: So nondeterministic procedures cannot read from the heap?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -10,190 +10,245 @@ import Strata.Languages.Core.Procedure
...
+  /-- A boolean literal. -/
   | LiteralBool (value : Bool)
+  /-- A string literal. -/
   | LiteralString (value : String)
+  /-- A variable reference by name. -/
   | Identifier (name : Identifier)
-  /- For single target assignments, use a single-element list.
-     Multiple targets are only allowed when the value is a StaticCall to a procedure
-     with multiple outputs, and the number of targets must match the number of outputs. -/
+  /-- Assignment to one or more targets. Multiple targets are only allowed when the value is a `StaticCall` to a procedure with multiple outputs. -/
```
joscoh, line 212: All of these invariants expressed in comments makes me think that Laurel should have a proper type system with a typing relation in Lean. Is this on the roadmap?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -250,51 +305,80 @@ def HighType.isBool : HighType → Bool
...
 structure Field where
+  /-- The field name. -/
   name : Identifier
+  /-- Whether the field is mutable. Mutable fields require write permission. -/
   isMutable : Bool
+  /-- The field's type. -/
   type : HighTypeMd
 
+/--
+A composite (class-like) type with fields and instance procedures.
```
joscoh: As far as I can tell, `Composite` is used both to refer to heap references and to these class-like types. Is that intentional? I find it a bit confusing.

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -250,51 +305,80 @@ def HighType.isBool : HighType → Bool
...
+with the constraint `value is Some<T> || value is Unit`.
+-/
 structure ConstrainedType where
+  /-- The constrained type's name. -/
   name : Identifier
+  /-- The base type being refined. -/
   base : HighTypeMd
+  /-- The name bound to the value in the constraint expression. -/
   valueName : Identifier
+  /-- The predicate that values of this type must satisfy. -/
```
joscoh, line 355: Does this mean that Laurel allows arbitrary subset types?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -10,190 +10,245 @@ import Strata.Languages.Core.Procedure
...
-    /- Equality on composite types uses reference equality for impure types, and structural equality for pure ones -/
-  | Eq | Neq
-  | And | Or | Not | Implies
-  /- Works on Int/Float64 -/
-  | Neg | Add | Sub | Mul | Div | Mod | DivT | ModT
-  | Lt | Leq | Gt | Geq
+  /-- Equality test. Uses reference equality for impure composite types, structural equality for pure ones. -/
+  | Eq
+  /-- Inequality test. -/
+  | Neq
```
joscoh, line 34: Minor, but why do you need this when you have `Eq` and `Not` already?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -10,190 +10,245 @@ import Strata.Languages.Core.Procedure
...
-Example:
-ProveBy(
-  someCall(arg1, arg2),
-  ProveBy(
-    aLemmaToHelpProveThePreconditionOfTheCall(),
-    anotherLemmaToProveThePreconditionOfTheFirstLemma()
-  )
-)
--/
+  /-- Attach a proof hint to a value. The semantics are those of `value`, but `proof` helps discharge assertions in `value`. -/
```
joscoh, line 248: How is this different from adding an assertions before `value` or calling a lemma? Is it just a matter of convenience?

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: They can but defining a reads clause for them is useless because of what reads clauses are used for

joscoh:
Overall looks good, most of my questions are about the actual language definition rather than the documentation. Could be useful to have an example program in the documentation though.

docs/verso/LaurelDoc.lean
keyboardDrummer: Updated the text.

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: Fixed

docs/verso/LaurelDoc.lean
keyboardDrummer: Updated

docs/verso/LaurelDoc.lean
keyboardDrummer: Updated

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: > Can a Composite be referred to at the user level

Not right now, but maybe it would be good for it to be a supertype of all Composite types for use in modifies clauses. Removed the reference from the text.

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: They can. Made the comment more explicit.

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: > All of these invariants expressed in comments makes me think that Laurel should have a proper type system with a typing relation in Lean

Would be good yes.

> Is this on the roadmap?

No

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: Arbitrary non-empty subset types yes

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: We could leave it out, but this is slightly easier for the front-end. They can use `Neq <sourceLoc> <expr>` instead of `Not <sourceLoc> (Eq <sourceLoc> <expr>)`, where the duplication of the source location might be confusing.

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: If you have `Block [Assert <Expr>; <Value1>; <Value2>]` then `<Expr>` is available when verifying `<Value1>` and `<Value2>`, but if you have `Block <ProveBy <Value1> <Expr>; <Value2>`, then `<Expr>` is only available when verifying `<Value1>`

tautschnig (approved):
Thank you!

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,167 @@
...
+set_option pp.rawOnError true
+
+#doc (Manual) "The Laurel Language" =>
+%%%
+shortTitle := "Laurel"
+%%%
+
+# Introduction
+
+Laurel is an intermediate verification language designed to serve as a target for popular garbage-collected languages that include imperative features, such as Java, Python, and JavaScript. Laurel tries to include any features that are common to those three languages.
```
joehendrix: This will be easier for me to read in VSCode if lines aren't too long (say 80-100 characters).

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,167 @@
...
+
+Laurel enables doing various forms of verification:
+- Deductive verification
+- Model checking
+- Property based testing
+- Data-flow analysis
```
joehendrix: Maybe will-enable?  Does it have all these backends yet?

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,167 @@
...
+```
+procedure increment(counter: Counter)
+  // In Laurel, this ensures clause can be left out
+  ensures x.value == old(x.value) + 1
+{
+  x.value := x.value + 1;
+}
+```
+
+Another Laurel choice is that a procedure may only specify a postcondition when its body is opaque.
```
joehendrix: This is the first mention of opaque.  Does that mean a declaration only?

docs/verso/LaurelDoc.lean
keyboardDrummer: Can I ask, what's wrong with using word wrap in VSCode? That always seemed strictly better to me since than the line width matches the editor width. Do you use an automatic formatter to add the line breaks? The editing experience when having them is frustrating to me since manually doing it is a pain.

tautschnig: I'm with Joe on that: it's not just your local editor, also future edits to this text will yield unnecessarily large diffs. But maybe I'm just that old-school person.

joscoh (approved):

docs/verso/LaurelDoc.lean
joehendrix: @tautschnig makes a stronger argument than mine. When a paragraph is a single long line, any small edit (fixing a typo, rewording a clause) marks the entire paragraph as changed in the diff. With line breaks at ~80-100 characters, diffs stay localized to the lines actually affected, which makes code review faster and `git blame` more useful. This also reduces the chance of merge conflicts when two people edit different sentences in the same paragraph.

Similarly, line-oriented tools like `sed` and `awk` become much more useful when lines are reasonably sized. This extends to LLMs as well -- they typically apply edits by matching and replacing specific lines, so shorter lines mean more precise targeting and fewer spurious changes.

Word wrap in the editor solves the display problem but not the version control or tooling problems.

keyboardDrummer: Good arguments. Have you found any tools to help with inserting the breaks that works for comments in Lean files as well?

docs/verso/LaurelDoc.lean
keyboardDrummer: Updated

docs/verso/LaurelDoc.lean
keyboardDrummer: Removed the line

docs/verso/LaurelDoc.lean
```diff
@@ -0,0 +1,193 @@
...
+A peculiar choice of Laurel is that it does not require imperative code to be encapsulated
+using a functional specification. A reason for this is that sometimes the imperative code is
+as readable as the functional specification. For example:
+```
+procedure increment(counter: Counter)
+  // In Laurel, this ensures clause can be left out
+  ensures x.value == old(x.value) + 1
+{
+  x.value := x.value + 1;
+}
```
joehendrix, line 78: The parameter is `counter` but the body and ensures clause reference `x`.

```suggestion
procedure increment(counter: Counter)
  // In Laurel, this ensures clause can be left out
  ensures counter.value == old(counter.value) + 1
{
  counter.value := counter.value + 1;
}
```

docs/verso/LaurelDoc.lean
tautschnig: In `vim` I just use `V` -> mark the paragraph -> `gq`. I suppose other editors will have similar capabilities, but I don't actually know.

joehendrix: In VSCode, I tend to manually do it, or just ask kiro or claude code.   Either is bit heavyweight, but sometimes the right fix is to reword.

joscoh (approved):

PR #435, opened 2026-02-17
Python -> Laurel

andrewmwells-amazon:
Python -> Laurel translation.

This includes a new `TCore` type in Laurel to directly refer to Core types. For now the benchmarks don't work with CVC5.

### DIFF
--- a/Strata/Languages/Core/Core.lean
+++ b/Strata/Languages/Core/Core.lean
@@ -51,6 +51,21 @@ def typeCheck (options : Options) (program : Program)
     if options.verbose >= .normal then dbg_trace f!"[Strata.Core] Type checking succeeded.\n"
     return program
 
+def formatProofObligation (ob : Imperative.ProofObligation Expression) :
+    Std.Format :=
+  let assumptionPairs := ob.assumptions.flatMap (·.toList)
+  let assumptionFmt := assumptionPairs.map fun (label, expr) =>
+    f!"{label}: {Core.formatExprs [expr]}"
+  let assumptionLine := if assumptionPairs.isEmpty then f!""
+                        else f!"\nAssumptions:\n{Std.Format.joinSep assumptionFmt "\n"}"
+  f!"Label: {ob.label}\n\
+     Property: {ob.property}{assumptionLine}\n\
+     Obligation:\n{Core.formatExprs [ob.obligation]}\n"
+
+def formatProofObligations (obs : Array (Imperative.ProofObligation Expression)) :
+    Std.Format :=
+  Std.Format.joinSep (obs.toList.map formatProofObligation) "\n"
+
 def typeCheckAndPartialEval (options : Options) (program : Program)
     (moreFns : @Lambda.Factory CoreLParams := Lambda.Factory.default) :
     Except DiagnosticModel (List (Program × Env)) := do
@@ -69,7 +84,7 @@ def typeCheckAndPartialEval (options : Options) (program : Program)
   if options.verbose >= .normal then do
     dbg_trace f!"{Std.Format.line}VCs:"
     for (_p, E) in pEs do
-      dbg_trace f!"{ProofObligations.eraseTypes E.deferred}"
+      dbg_trace f!"{formatProofObligations E.deferred}"
   return pEs
 
 instance : ToString (Program) where
--- a/Strata/Languages/Core/DDMTransform/ASTtoCST.lean
+++ b/Strata/Languages/Core/DDMTransform/ASTtoCST.lean
@@ -75,8 +75,9 @@ def unknownTypeVar : String := "$__unknown_type"
 /-- Generate parameter names efficiently -/
 def mkParamName (i : Nat) : String := "a" ++ toString i
 
-/-- Generate quantifier variable names efficiently -/
-def mkQuantVarName (level : Nat) : String := "x" ++ toString level
+/-- Generate quantifier variable names with a `__` prefix to indicate that they
+    are generated names. In the future, we will store existing variable names in an extra field of quantifier expressions. -/
+def mkQuantVarName (level : Nat) : String := "__q" ++ toString level
 
 structure Scope where
   /-- Track bound variables in this scope -/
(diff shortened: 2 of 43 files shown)


Strata/Languages/Laurel/Laurel.lean
```diff
@@ -79,6 +79,10 @@ inductive HighType : Type where
   /- Java has implicit intersection types.
      Example: `<cond> ? RustanLeino : AndersHejlsberg` could be typed as `Scientist & Scandinavian`-/
   | Intersection (types : List (WithMetadata HighType))
+  /-
+    Type "passed through" from Core. Intended to allow translations to Laurel to refer directly to Core.
```
keyboardDrummer, line 83: I think it would be good to add that this is a temporary feature that we've added to migrate Python to Laurel

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -145,6 +146,7 @@ def translateExpr (constants : List Constant) (env : TypeEnv) (expr : StmtExprMd
       -- Field selects should have been eliminated by heap parameterization
       -- If we see one here, it's an error in the pipeline
       panic! s!"FieldSelect should have been eliminated by heap parameterization: {Std.ToFormat.format target}#{fieldName}"
+  | .Hole => .fvar () (Core.CoreIdent.locl s!"DUMMY_VAR_{env.length}") none -- TODO: don't do this
```
keyboardDrummer, line 149: What's the TODO for?

Strata/Languages/Python/CorePrelude.lean
```diff
@@ -331,6 +331,10 @@ spec {
   assume [assume_maybe_except_none]: (ExceptOrNone..isExceptOrNone_mk_none(maybe_except));
 };
 
+datatype PyAnyType () {
```
keyboardDrummer, line 334: How will we represent this in the long term? What's the behavior of this?

StrataTest/Languages/Python/expected_laurel/test_function_def_calls.expected
```diff
@@ -1,32 +1,13 @@
 
+==== Verification Results ====
```
keyboardDrummer, line 2: Why has this changed?

StrataTest/Languages/Python/expected/test_function_def_calls.expected
```diff
@@ -1,32 +1,13 @@
...
-
 assert_name_is_foo: ✅ pass (at byte 11075)
-
 assert_opt_name_none_or_str: ✅ pass (at byte 11125)
-
 assert_opt_name_none_or_bar: ✅ pass (at byte 11272)
-
 ensures_maybe_except_none: ✅ pass (at byte 10978)
-
-Assertion failed at byte 11075: test_helper_procedure_assert_name_is_foo_3: ❌ fail
```
keyboardDrummer: What happened here?

StrataTest/Languages/Python/expected_laurel/test_missing_models.expected
```diff
@@ -1,14 +1,19 @@
...
 ensures_str_strp_reverse: ✅ pass (at byte 8761)
-
 assert_name_is_foo: ✅ pass (at byte 11075)
-
 assert_opt_name_none_or_str: ✅ pass (at byte 11125)
-
 assert_opt_name_none_or_bar: ✅ pass (at byte 11272)
-
 ensures_maybe_except_none: ✅ pass (at byte 10978)
+(Origin_test_helper_procedure_Requires)req_name_is_foo: 🟡 unknown (at byte 10641)
```
keyboardDrummer: What happened here?

StrataTest/Languages/Python/run_py_analyze_laurel.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+failed=0
+
+# Initialize conda
+eval "$(conda shell.bash hook)"
+conda activate strata
```
aqjune-aws: `conda` is not available on my machine - can we run this test without `conda`?

I think it is good to merge this script with `run_py_analyze.sh` because the boilerplate lines are equivalent.

StrataTest/Languages/Python/expected_laurel/test_function_def_calls.expected
```diff
@@ -1,32 +1,13 @@
...
-test_helper_procedure_assert_opt_name_none_or_str_4: ✅ pass (at byte 11125)
-
-test_helper_procedure_assert_opt_name_none_or_bar_5: ✅ pass (at byte 11272)
-
-(Origin_test_helper_procedure_Requires)req_name_is_foo: ✅ pass (at byte 10641)
-
+(Origin_test_helper_procedure_Requires)req_name_is_foo: 🟡 unknown (at byte 10641)
 (Origin_test_helper_procedure_Requires)req_opt_name_none_or_str: ✅ pass (at byte 10690)
-
 (Origin_test_helper_procedure_Requires)req_opt_name_none_or_bar: ✅ pass (at byte 10836)
```
aqjune-aws: Do you have idea why the output messages disappeared in this test case? 
And the StrataTest/Languages/Python/expected/test_missing_models.expected case seems opposite. Are we having different procedure filtering policy compared to the old Python translator?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -0,0 +1,593 @@
...
+/-! ## Helper Functions -/
+
+/-- Create metadata from a SourceRange for attaching to Laurel statements. -/
+def sourceRangeToMetaData (filePath : String) (sr : SourceRange) : Imperative.MetaData Core.Expression :=
+  let uri : Uri := .file filePath
+  let fileRangeElt := ⟨ Imperative.MetaData.fileRange, .fileRange ⟨ uri, sr ⟩ ⟩
+  #[fileRangeElt]
+
+/-- Create default metadata for Laurel AST nodes -/
+def defaultMetadata : Imperative.MetaData Core.Expression :=
```
keyboardDrummer, line 102: I think having empty metadata would be a better default, so it crashes when metadata is incorrectly not specified, instead of showing diagnostics at an incorrect position.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -0,0 +1,593 @@
...
+
+    if numProvided < numExpected then
+      -- Fill remaining args with None of the appropriate type
+      for i in [numProvided:numExpected] do
+        let paramType := sig.inputs[i]!
+        translatedArgs := translatedArgs ++ [mkNoneForType paramType]
+
+    if sig.outputs.length > 0 then
+      let call := mkStmtExprMd (StmtExpr.StaticCall funcName translatedArgs)
+      return mkStmtExprMd (.Assign [mkStmtExprMd (.Identifier "maybe_except")] call)
```
keyboardDrummer: I'm confused. Do calls to prelude procedures always return just the `maybe_except` value? Do non-prelude procedures not throw exceptions?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -0,0 +1,593 @@
...
+
+  -- Expression statement (e.g., function call)
+  | .Expr _ value => do
+    let expr ← translateExpr ctx value
+    return (ctx, expr)
+
+  | .Import _ _ | .ImportFrom _ _ _ _ => return (ctx, mkStmtExprMd .Hole)
+
+  -- Try/except - wrap body with exception checks and handlers
+  | .Try _ body handlers _ _ => do
```
keyboardDrummer, line 405: Would be amazing if we can add exception handling that supports both Python and Java, to Laurel. I'm not sure how to make that work though.

I think ideally, we would add a `hasExceptions: HighType` flag to a procedure, where the specified type can be a constrained type so you can say what the base exception type is but also what specific exceptions it can throw. Wherever this thing is called you'd then translate the surrounding code to a monadic style until its encapsulated by a try/catch.

For higher order procedures, the procedure type of the parameter would include the `hasExceptions: HighType` property.

The above solution does require each procedure to specify what exceptions it can throw. If you don't know whether a procedure throws an exception you'll have to be defensive and specify something like `hasExceptions: <ExceptionBaseType>`, which it seems you would do with your current Python handling.

Since you're currently requiring the Python programs to be typed, wouldn't it be possible to also require that the exceptions are specified for the functions?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -0,0 +1,593 @@
...
+    let bodyStmts := prependExceptHandlingHelper bodyStmts
+    let bodyBlock := mkStmtExprMd (StmtExpr.Block bodyStmts none)
+
+    -- Create procedure with transparent body (no contracts for now)
+    let proc : Procedure := {
+      name := funcName
+      inputs := inputs
+      outputs := outputs
+      precondition := mkStmtExprMd (StmtExpr.LiteralBool true)
+      determinism := .deterministic none -- TODO: need to set reads
```
keyboardDrummer, line 502: Reads are optional. They help verification but not having them is still sound.

StrataTest/Languages/Python/expected_laurel/test_function_def_calls.expected
andrewmwells-amazon: The Python -> Laurel -> Core impl isn't doing procedure inlining.

StrataTest/Languages/Python/expected/test_function_def_calls.expected
andrewmwells-amazon: This is due to not inlining procedures in the Python -> Laurel -> Core implementation.

StrataTest/Languages/Python/expected_laurel/test_function_def_calls.expected
andrewmwells-amazon: The output format is slightly different:
- Has `==== Verification Results ====`
- Doesn't have newlines between results

Functional changes are:
- The Laurel impl uses z3
- The Laurel impl doesn't do procedure inlining

Strata/Languages/Python/CorePrelude.lean
andrewmwells-amazon: We'll port https://github.com/strata-org/Strata/pull/262 to this translation.

StrataTest/Languages/Python/expected_laurel/test_missing_models.expected
andrewmwells-amazon: See: https://github.com/strata-org/Strata/pull/435#discussion_r2823234879

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
andrewmwells-amazon: If I understand `.Hole` correctly, this generally isn't an acceptable translation.

Strata/Languages/Laurel/Laurel.lean
andrewmwells-amazon: I suspect this will live on for a while.

Strata/Languages/Python/PythonToLaurel.lean
andrewmwells-amazon: Certainly worth discussing, but won't be part of this PR. My main concern is that because languages implement exception handling differently, making decisions about exception handling in Laurel could encourage unsound translations to Laurel.

Strata/Languages/Python/PythonToLaurel.lean
andrewmwells-amazon: This was a bug. I changed it to something that I think will be correct if Laurel allows `a := b := call Foo();`.

Strata/Languages/Python/PythonToLaurel.lean
keyboardDrummer: > but won't be part of this PR

Of course ^^

aqjune-aws (approved):

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: Regardless of how long it'll live on, we should strive towards not using it, so let's mark it as such. Can do it in another PR though.

PR #437, opened 2026-02-17
Fix DDM precedence for single-argument wrapper ops

joehendrix:
## Summary

Reduce unnecessary parentheses in DDM pretty printing by overhauling syntax elaboration — convert `SyntaxDef` to an inductive type with `std`/`passthrough` variants, fix precedence for wrapper and string-literal ops, remove dead `unwrap` code, and rewrite the precedence documentation.

## Details

- Single string-literal syntax ops get `maxPrec + 1` (never parenthesized)
- Give `Nat` and `Decimal` literals `maxPrec + 1` so they are never spuriously
  parenthesized (e.g., `i + (1)` → `i + 1`)
- Remove unused `unwrap` field from `SyntaxDefAtom` and related dead code
  (`unwrapSpecs`, `unwrapTree`); unwrap is only used via metadata in Gen.lean
- Replace unsafe array indexing with bounds-checked access in Format.lean
- Fix missing whitespace in C_Simp `measure`/`invariant`/`pre`/`post` grammar

## Documentation

- Rewrite precedence section of DDM documentation; fix broken link, grammar,
  and section hierarchy
- Add lychee link checker to CI doc build

### DIFF
--- a/docs/verso/DDMDoc.lean
+++ b/docs/verso/DDMDoc.lean
@@ -31,13 +31,13 @@ distributed analysis across language and system boundaries.
 
 The Strata DDM is implemented in Lean.  Strata dialects and programs may be embedded
 in Lean or in standalone dialect files.  In a future release, we will support working with
-Strata from the command line as well as other languages such as .Net-based language,
+Strata from the command line as well as other languages such as .Net-based languages,
 JVM-based languages, Rust, Python and JavaScript/TypeScript.
 
 # Strata Dialect Definitions
 
 Dialects are the core mechanism in Strata used to declare and extend program
-intermediate representations.   A Strata dialect defines new syntactic categories,
+intermediate representations.  A Strata dialect defines new syntactic categories,
 operations, types, functions and metadata.  Strata dialects are composable and
 can be used to define extensions that can be used in multiple languages.  There
 is a builtin dialect, called `Init` that provides some basic declarations that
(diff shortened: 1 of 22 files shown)


docs/verso/DDMDoc.lean
```diff
@@ -336,38 +337,90 @@ describe the syntax of the operator or function.
...
 
 ```
 fn implies (p : Pred, q : Pred) : Pred => @[prec(20), rightassoc] p " ==> " q;
 ```
 
-Without the rightassoc operator, this is equivalent to the definition:
+Without the `rightassoc` annotation, this is equivalent to the definition:
 
 ```
 fn implies (p : Pred, q : Pred) : Pred => @[prec(20)] p:20 " ==> " q:19;
```
keyboardDrummer, line 397: Is there no way to infer the correct argument precedence values? I would think that the operator precedence already fully determines the parsing behavior, and from that we should be able to infer how a dual printer should behave.

docs/verso/DDMDoc.lean
```diff
@@ -336,38 +337,90 @@ describe the syntax of the operator or function.
 * For multiline declarations, the function `indent(`_n_`,` _tokens_`)` may
   be used.  The _tokens_ will appear indented when they appear on new lines.
 
-There are three metadata annotations in the Strata dialect that may affect
-precedence in syntax definitions.
+### Precedence
+
+Precedence controls when parentheses are inserted during pretty-printing.
```
keyboardDrummer, line 342: Cool! 

Do you assume that every language will use `(` `)` to group expressions? Seems fine but I'm asking out of curiosity. What about other ways to manipulate precedence, like inserting `$` in Lean, would there be ways to generate that during printing?

docs/verso/DDMDoc.lean
```diff
@@ -336,38 +337,90 @@ describe the syntax of the operator or function.
...
 ```
 
-As an example, a right associative implies operator can be defined with the syntax:
+*Default argument precedences.* When `arg:p` is not written explicitly,
+the argument precedence is derived from the operator precedence and
+associativity:
+
+* For a non-associative operator with precedence `p` and multiple arguments,
+  the first and last arguments get precedence `p` and middle arguments get
+  precedence `0`.
```
aqjune-aws, line 380: Will there be a case that introduces too many parentheses if the middle arguments also have precedence `p`? I was wondering because if we have a ternary operator `x ? y : z`, and if `y` is another ternary operator `p ? q : r`, it will be printed as `x ? p ? q : r : z`, but `x ? (p ? q : r) : z` seems slightly more natural.

docs/verso/DDMDoc.lean
joehendrix: I'm not sure what you mean, but I think this is out of scope for the PR, so I am going to resolve the conversation so I can merge.  If you'd like to discuss further, could you followup directly?

docs/verso/DDMDoc.lean
joehendrix: That's an interesting idea, but out of scope for this PR.

docs/verso/DDMDoc.lean
joehendrix: That's a good point, I'm not sure this is a great default, but I am also not sure what a correct default should be.

The auto-generated Python syntax uses the form `opname(arg0, ..., argn)` and in this case maybe we want a low required precedence for each argument.

I'm going to resolve this since I'd like to commit the fix, and this is just documenting the existing behavior; I haven't touched the code.

aqjune-aws (approved):

PR #438, opened 2026-02-17
Pretty printer for Core's Proof Obligations

shigoel:
*Description of changes:*
Print Core's proof obligations (which are Lambda expressions) using the DDM syntax (via translation to `#strata_gen` AST).

### DIFF
--- a/Strata/Languages/Core/DDMTransform/ASTtoCST.lean
+++ b/Strata/Languages/Core/DDMTransform/ASTtoCST.lean
@@ -76,8 +76,7 @@ def unknownTypeVar : String := "$__unknown_type"
 def mkParamName (i : Nat) : String := "a" ++ toString i
 
 /-- Generate quantifier variable names with a `__` prefix to indicate that they
-    are generated names. In the future, we will look for variables names in the
-    metadata of the quantifier expressions. -/
+    are generated names. In the future, we will store existing variable names in an extra field of quantifier expressions. -/
 def mkQuantVarName (level : Nat) : String := "__q" ++ toString level
 
 structure Scope where


Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -75,8 +75,10 @@ def unknownTypeVar : String := "$__unknown_type"
 /-- Generate parameter names efficiently -/
 def mkParamName (i : Nat) : String := "a" ++ toString i
 
-/-- Generate quantifier variable names efficiently -/
-def mkQuantVarName (level : Nat) : String := "x" ++ toString level
+/-- Generate quantifier variable names with a `__` prefix to indicate that they
+    are generated names. In the future, we will look for variables names in the
+    metadata of the quantifier expressions. -/
```
MikaelMayer: Since these names are very specific to quantifiers and abstractions, I would rather have them in an explicit field. It will make everything easier to develop since LExpr is anyway a parametric building block in its metadata.
```suggestion
    are generated names. In the future, we will store existing variable names in an extra field of quantifier expressions. -/
```
If you fear we might add too many fields, perhaps it's worth regrouping quantifier-specific metadata into its own field, for example we could regroup trigger and name.

StrataTest/Languages/Core/Examples/GeneratedLabels.lean
```diff
@@ -38,10 +38,10 @@ info: type Ref;
 type Field;
 type Struct := Map Field int;
 type Heap := Map Ref Struct;
-axiom [axiom_0]: forall x0 : Struct :: forall x1 : Field :: forall x2 : Field :: forall x3 : int :: !(x1 == x2) ==> x0[x1] == (x0[x2:=x3])[x1];
-axiom [axiom_1]: forall x0 : Struct :: forall x1 : Field :: forall x2 : int :: (x0[x1:=x2])[x1] == x2;
-axiom [axiom_2]: forall x0 : Heap :: forall x1 : Ref :: forall x2 : Ref :: forall x3 : Struct :: !(x1 == x2) ==> x0[x1] == (x0[x2:=x3])[x1];
-axiom [axiom_3]: forall x0 : Heap :: forall x1 : Ref :: forall x2 : Struct :: (x0[x1:=x2])[x1] == x2;
+axiom [axiom_0]: forall __q0 : Struct :: forall __q1 : Field :: forall __q2 : Field :: forall __q3 : int :: !(__q1 == __q2) ==> __q0[__q1] == (__q0[__q2:=__q3])[__q1];
```
MikaelMayer, line 41: Can't wait for all these generated names to go await.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
shigoel: I’m happy with your change, thanks. Not too worried about too many metadata fields yet esp since we don’t any apart from source locations :)

StrataTest/Languages/Core/Examples/GeneratedLabels.lean
shigoel: Me too me too.

PR #439, opened 2026-02-17
Fix funcDecl_statement scoping bugs

### LINKED ISSUE #436: Scoping bug with function declaration statements
Consider the following
```

namespace Strata

open Core

def translatePgm (p : Strata.Program) : Core.Program :=
  (TransM.run Inhabited.default (translateProgram p)).fst

-- Just for printing
partial def extractFuncDecls : Core.Statement → List (Imperative.PureFunc Core.Expression)
  | .funcDecl decl _ => [decl]
  | .block _ stmts _ => stmts.flatMap extractFuncDecls
  | .ite _ thenb elseb _ => thenb.flatMap extractFuncDecls ++ elseb.flatMap extractFuncDecls
  | .loop _ _ _ body _ => body.flatMap extractFuncDecls
  | _ => []

def bugRepro : Program :=
#strata
program Core;

procedure test(cond : bool, x : int, y : int) returns () {
  if (cond) {
    function f(a : int) : int { a + x }
    var r1 : int := f(10);
  } else {
    function f(a : int) : int { a + y }
    var r2 : int := f(20);
  }
};

#end

#eval do
  let prog := translatePgm bugRepro
  let _ ← prog.decls.foldlM (fun _ decl => do
    match decl with
    | Core.Decl.proc proc _ =>
      IO.println s!"Procedure: {proc.header.name.name}"
      let funcs := proc.body.flatMap extractFuncDecls
      let _ ← funcs.foldlM (fun idx (f : Imperative.PureFunc Core.Expression) => do
        IO.println s!"  Function #{idx}: {f.name.name}"
        IO.println s!"    Body: {f.body}"
        pure (idx + 1)
      ) 0
    | _ => pure ()
  ) ()
  pure ()
```
This shows that the body of the second `f` is translated as `a + r1` instead of `a + y` (the complicated printing mechanism is needed because `ToFormat` does not handle inline function statements). I believe this is a problem with the translation of nested statements in `Core/DDMTransform/Translate.lean`.

### LINKED ISSUE #445: Indices for function decl statements are wrong
Consider the following program:
```
import Strata.Languages.Core.Core
import Strata.Languages.Core.DDMTransform.Translate

open Core
open Strata

def translate' (t : Strata.Program) : Core.Program :=
  (TransM.run Inhabited.default (translateProgram t)).fst

def funcDeclRefAfter :=
#strata
program Core;

procedure test() returns ()
{
  var x : int := 1;
  function safeDiv(y : int) : int
    { y div x }
  assert 5 div x > 0;
  var z : int := safeDiv(5);
};

#end

#eval (Std.format (translate' funcDeclRefAfter).stripMetaData)
```
This outputs
```
procedure test () returns ()
{
  var x : int := 1;
  function safeDiv (y : int) : int { fvar!1 }
  assert [assert_0]: 5 div safeDiv > 0;
  var z : int := fvar!0(5);
  };
```
Which has 3 issues:
1. The body of `safeDiv` is being translated to `fvar!1` instead of `y div x`.
2. A reference to a variable `x` defined before `safeDiv` instead resolves to `safeDiv` in the assertion.
3. A reference to `safeDiv` itself instead resolves to `fvar!0` in `z`.

This appears to be an issue with the indices generated for function decl statements in `Core/DDMTransform/Translate.lean`.

MikaelMayer:
Fixes #436
Fixes #445

## Bugs fixed

**1. Arg order mismatch in `funcDecl_statement` translation**

The pattern match in `translateStmt` treated arg[4] as axioms (ignored) and arg[5] as the body, but the grammar defines arg[4] as the body and arg[5] as the inline flag. This caused function bodies to always translate as empty. Fixed the arg order to match the grammar: `#[namea, _typeArgsa, bindingsa, returna, bodya, _inlinea]`.

**2. Function name incorrectly in body scope**

The function name was included in the body's bound variables, but DDM's `@[declareFn]` annotation only adds the function name for *subsequent* statements, not inside the body itself. This shifted de Bruijn indices inside the body. Removed the function name from body bindings so only parameters are in scope.

**3. Pretty printer scope mismatch**

The pretty printer registered the function name as a free variable after processing a `funcDecl`, but DDM's `@[declareFn]` makes it a bound variable. This caused wrong variable names in the pretty-printed output after a function declaration (e.g., `x` displayed as `safeDiv`), and function calls displayed as `fvar!0(...)` instead of using the function name. Changed to register the function name as a bound variable (`pushBoundVar`), and updated `lopToExpr` to look up local function names in bound vars first.

**4. Variable leak across if-statement branches**

The old code shadowed `bindings` with each branch's result, so variables declared in the then-branch were visible in the else-branch. The fix passes the original bindings to each branch, only carrying forward the gen counters for sequential label numbering.

**5. Extra parentheses in if/else/block formatting**

Added `:0` precedence annotations to `if_statement`, `else1`, and `block_statement` grammar rules so the DDM pretty printer no longer wraps blocks and else-clauses in unnecessary parentheses.

## Other changes

- Added `@[scope(c)]` to the `block` grammar rule for proper parse-time scoping.

## Testing

- Added `NestedVarScopingTest` with exact repro programs from both issues, verifying correct translation and pretty-printing via `#guard_msgs`.
- Updated `FuncDeclStmtTest` expected output to match the corrected behavior.
- All existing tests pass.

### DIFF
--- a/Strata/DL/SMT/Op.lean
+++ b/Strata/DL/SMT/Op.lean
@@ -158,6 +158,12 @@ inductive Op.Strings : Type where
   | re_index : Nat → Op.Strings
 deriving Repr, DecidableEq, Inhabited, Hashable
 
+inductive Op.Arrays : Type where
+  ---------- SMTLib theory of arrays (`ArraysEx`) ----------
+  | select
+  | store
+deriving Repr, DecidableEq, Inhabited, Hashable
+
 inductive Op.DatatypeFuncs : Type where
   | constructor : Op.DatatypeFuncs
   | tester : Op.DatatypeFuncs
@@ -173,6 +179,8 @@ inductive Op : Type where
   | bv : Op.BV → Op
   -- SMTLib theory of unicode strings and regular expressions (`Strings`)
   | str : Op.Strings → Op
+  -- SMTLib theory of arrays (`ArraysEx`)
+  | arr : Op.Arrays → Op
   -- An operator to group triggers together
   | triggers
   -- Core ADT operators with a trusted mapping to SMT
@@ -231,13 +239,20 @@ elab "#genOpAbbrevs" : command => do
         let abbrevCmd ← `(command| abbrev $(mkIdent name) := Op.str $(mkIdent ctor))
         abbrevs := abbrevs.push (name, abbrevCmd)
 
+  if let some (.inductInfo arrInfo) := env.find? `Strata.SMT.Op.Arrays then
+    for ctor in arrInfo.ctors do
+      let ctorName := ctor.toString.splitToList (· == '.') |>.getLast!
+      let name := Lean.Name.mkStr2 "Op" ctorName
+      let abbrevCmd ← `(command| abbrev $(mkIdent name) := Op.arr $(mkIdent ctor))
+      abbrevs := abbrevs.push (name, abbrevCmd)
+
   for a in abbrevs do
     elabCommand a.snd
   logInfo s!"Generated abbrevs: {abbrevs.map (fun a => a.fst)}"
 
 
 /--
-info: Generated abbrevs: #[Op.not, Op.and, Op.or, Op.eq, Op.ite, Op.implies, Op.distinct, Op.uf, Op.neg, Op.sub, Op.add, Op.mul, Op.div, Op.mod, Op.abs, Op.le, Op.lt, Op.ge, Op.gt, Op.bvneg, Op.bvadd, Op.bvsub, Op.bvmul, Op.bvnot, Op.bvand, Op.bvor, Op.bvxor, Op.bvshl, Op.bvlshr, Op.bvashr, Op.bvslt, Op.bvsle, Op.bvult, Op.bvsge, Op.bvsgt, Op.bvule, Op.bvugt, Op.bvuge, Op.bvudiv, Op.bvurem, Op.bvsdiv, Op.bvsrem, Op.bvnego, Op.bvsaddo, Op.bvssubo, Op.bvsmulo, Op.bvconcat, Op.zero_extend, Op.str_length, Op.str_concat, Op.str_lt, Op.str_le, Op.str_at, Op.str_substr, Op.str_prefixof, Op.str_suffixof, Op.str_contains, Op.str_indexof, Op.str_replace, Op.str_replace_all, Op.str_to_re, Op.str_in_re, Op.re_none, Op.re_all, Op.re_allchar, Op.re_concat, Op.re_union, Op.re_inter, Op.re_star, Op.str_replace_re, Op.str_replace_re_all, Op.re_comp, Op.re_diff, Op.re_plus, Op.re_opt, Op.re_range, Op.re_loop, Op.re_index]
+info: Generated abbrevs: #[Op.not, Op.and, Op.or, Op.eq, Op.ite, Op.implies, Op.distinct, Op.uf, Op.neg, Op.sub, Op.add, Op.mul, Op.div, Op.mod, Op.abs, Op.le, Op.lt, Op.ge, Op.gt, Op.bvneg, Op.bvadd, Op.bvsub, Op.bvmul, Op.bvnot, Op.bvand, Op.bvor, Op.bvxor, Op.bvshl, Op.bvlshr, Op.bvashr, Op.bvslt, Op.bvsle, Op.bvult, Op.bvsge, Op.bvsgt, Op.bvule, Op.bvugt, Op.bvuge, Op.bvudiv, Op.bvurem, Op.bvsdiv, Op.bvsrem, Op.bvnego, Op.bvsaddo, Op.bvssubo, Op.bvsmulo, Op.bvconcat, Op.zero_extend, Op.str_length, Op.str_concat, Op.str_lt, Op.str_le, Op.str_at, Op.str_substr, Op.str_prefixof, Op.str_suffixof, Op.str_contains, Op.str_indexof, Op.str_replace, Op.str_replace_all, Op.str_to_re, Op.str_in_re, Op.re_none, Op.re_all, Op.re_allchar, Op.re_concat, Op.re_union, Op.re_inter, Op.re_star, Op.str_replace_re, Op.str_replace_re_all, Op.re_comp, Op.re_diff, Op.re_plus, Op.re_opt, Op.re_range, Op.re_loop, Op.re_index, Op.select, Op.store]
 -/
 #guard_msgs in
 #genOpAbbrevs
@@ -325,6 +340,8 @@ def Op.mkName : Op → String
   | .re_range      => "re.range"
   | .re_index _    => "re.^"
   | .re_loop _ _   => "re.loop"
+  | .select        => "select"
+  | .store         => "store"
 
 def Op.LT : Op → Op → Bool
   | .uf f₁, .uf f₂                    => f₁ < f₂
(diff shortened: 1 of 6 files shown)


joscoh (dismissed):

StrataTest/Languages/Core/NestedVarScopingTest.lean
```diff
@@ -0,0 +1,57 @@
...
+import Strata.Languages.Core.DDMTransform.Translate
+
+namespace Strata
+
+open Core
+
+def translatePgm (p : Strata.Program) : Core.Program :=
+  (TransM.run Inhabited.default (translateProgram p)).fst
+
+-- Test that variables from one branch don't leak into another branch
```
andrewmwells-amazon: Sorry, but I don't understand how the below tests this? Is there a program that was accepted before but shouldn't have been?

andrewmwells-amazon: Something like:
```
procedure test(cond : bool) returns () {
  if (cond) {
    var r1 : int := 10;
  } else {
    r1 := 20;
  }
};
```

Seems like what I'd expect to have passed before, but no longer should.

MikaelMayer: I'm embarrassed to admit that you are right.
Note to myself: Asking agents to create initially failing tests is not reliable.

joscoh: https://github.com/strata-org/Strata/issues/436 has a test that currently fails, though it isn't easy to include because printing function decl statements is not yet supported

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -1117,29 +1117,23 @@ partial def translateStmt (p : Program) (bindings : TransBindings) (arg : Arg) :
...
+    -- for subsequent statements only). So body bindings = outer + parameters.
     let funcType := Lambda.LMonoTy.mkArrow outputMono (inputs.values.reverse)
     let funcBinding : LExpr Core.CoreLParams.mono := .op () name (some funcType)
     let in_bindings := (inputs.map (fun (v, ty) => (LExpr.fvar () v ty))).toArray
-    -- Order: existing boundVars, then function, then parameters
-    let bodyBindings := { bindings with boundVars := bindings.boundVars ++ #[funcBinding] ++ in_bindings }
+    let bodyBindings := { bindings with boundVars := bindings.boundVars ++ in_bindings }
 
-    -- Translate body with function parameters in scope
+    -- Translate body (arg[4], not the inline flag)
```
joscoh: I'm not sure what this comment means.

StrataTest/Languages/Core/NestedVarScopingTest.lean
```diff
@@ -0,0 +1,94 @@
...
+info: [Strata.Core] Type checking succeeded.
+
+---
+info: ok: procedure test (cond : bool, x : int, y : int) returns ()
+{
+  if(cond){
+    |$_then|: ({
+      function f (a : int) : int { a + x }
+      var r1 : int := f(10);
+      })}(else({
```
joscoh: Not an issue for this PR, but what is going on with the formatting here?

Strata/Languages/Core/DDMTransform/Translate.lean
MikaelMayer: The order of arguments was wrong and this PR corrects it, but this is not a stateless comment so you're right let me remove it.

StrataTest/Languages/Core/NestedVarScopingTest.lean
MikaelMayer: I think it's because we added Scope to ensure the DDM scopes the then block correctly, and we displaying a scoped block here. Let me see if I can do something for labelled blocks.

andrewmwells-amazon (approved):

StrataTest/Languages/Core/NestedVarScopingTest.lean
MikaelMayer: So it was not related to adding scopes to the DDM, it was actually the type checker. I only pushed a fix that removes the extra parentheses.

joscoh (dismissed):

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -752,8 +758,9 @@ def funcDeclToStatement {M} [Inhabited M] (decl : Imperative.PureFunc Expression
     pure bodyExpr
   | some body => lexprToExpr body 0
   modify ToCSTContext.popScope
-  -- Register function name as a scoped free variable in the parent scope.
-  modify (·.addScopedFreeVars #[name.val])
+  -- Register function name as a scoped bound variable in the parent scope,
+  -- matching DDM's @[declareFn] which makes the name a bvar.
```
shigoel, line 754: Nice, thanks. The function `addScopedFreeVars` is then dead code. Okay to remove it in another PR.

shigoel (dismissed):

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
MikaelMayer: Excellent remark. Why wait. I'll do this as well.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -567,7 +567,13 @@ def lopToExpr {M} [Inhabited M]
     (name : String) (args : List (CoreDDM.Expr M))
     : ToCSTM M (CoreDDM.Expr M) := do
   let ctx ← get
-  -- User-defined functions.
+  -- User-defined functions: check bound vars first (local funcDecl via
+  -- @[declareFn]), then free vars (global declarations).
+  match ctx.findBoundVarIndex? name with
```
shigoel, line 564: Actually, this reminds me: does Core's typechecker disallow shadowing for function names? E.g., if `foo` is a global function, then no sub-function can be called `foo`.

MikaelMayer: No disallowing because sub-functions are not recursive so they can refer to previous functions of the same name.
We could discuss this further but I'm not in favor of disallowing shadowing in general, although I know that in some cases (especially for a top-level that could be splitted across multiple files) shadowing wouldn't be acceptable.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
shigoel: Thanks. I think we don't need scoped Free Vars in ASTtoCST anymore (i.e., free vars are all global, so we could define a `freeVar` field in `ToCSTContext` and remove it from `Scopes`), but I'm also okay to keep status quo until we fix the bvar/fvar issue.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
shigoel: We disallow shadowing in general.
E.g., see https://github.com/strata-org/Strata/blob/main/Strata/Languages/Core/ProgramType.lean#L68-L69

shigoel: Also: https://github.com/strata-org/Strata/blob/main/StrataTest/Languages/Core/Examples/ShadowedVars.lean

shigoel: (Anyway, this is a different issue; sorry for discussing on this PR!) :) Shouldn't block this one.

joscoh (approved):

shigoel (approved):

PR #441, opened 2026-02-17
Add useArrayTheory option, conditionally lower Core's Map operations to SMT Array

aqjune-aws:
This adds `useArrayTheory` option that conditionally lowers the `select` and `update` operations in Core (`map_get` and `map_set` rules in Core DDM) to SMT-LIB's Array theory operations.

Lots of helps from Kiro.

This pull request has overlapping edits with #414. 


Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -542,62 +551,67 @@ partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Conte
...
+          let ctx ← func.axioms.foldlM (fun acc_ctx (ax: LExpr CoreLParams.mono) => do
+            let current_axiom_ctx := acc_ctx.addSubst smt_ty_inst
+              let (axiom_term, new_ctx) ← toSMTTerm E [] ax current_axiom_ctx
+              .ok (new_ctx.addAxiom axiom_term)
+          ) ctx
+          let ctx := ctx.restoreSubst savedSubst
+          .ok (.app (Op.uf uf), smt_outty, ctx)
+        else
+          .ok (.app (Op.uf uf), smt_outty, ctx)
+
```
aqjune-aws, line 603: This change looks big, but in fact most of these lines are due to indentations. Turn on "Hide whitespace" on Github diff page and this diff will disappear.

StrataTest/Languages/Core/SMTEncoderTests.lean
```diff
@@ -101,4 +102,109 @@ info: "; f\n(declare-fun f0 (Int Int) Int)\n; x\n(declare-const f1 Int)\n(define
...
+    modifies m;
+    ensures result == v;
+}
+{
+    m := m[k := v];
+    result := m[k];
+};
+#end
+
+-- Test verification with axiomatized maps (default) - currently fails
```
atomb: Looks like this comment is no longer true?

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -284,41 +292,42 @@ partial def appToSMTTerm (E : Env) (bvs : BoundVars) (e : LExpr CoreLParams.mono
...
     .ok (((Term.app (.uf uf) [e1t] smt_outty)), ctx)
   | .app _ _ _ =>
     .error f!"Cannot encode expression {e}"
 
-  | _ => toSMTTerm E bvs e ctx
+  | _ => toSMTTerm E bvs e ctx useArrayTheory
 
-partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Context) :
+partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Context)
+  (useArrayTheory : Bool := false) :
```
joscoh, line 319: Are arrays the only construct that will have different SMT encodings? If not, I imagine we would later want some sort of configuration options here instead of having to pass parameters for every encoding choice. Not blocking for now though.

StrataTest/Languages/Core/SMTEncoderTests.lean
joscoh: It would be good to have an example that fails with the old encoding and now succeeds.

aqjune-aws: This was a wrong comment made by Claude. :/ The test was supposed to pass in both configurations, sorry.
I tried some tweaks to make the test fail on axiomatization, but could not make it.

aqjune-aws:
> ... a test for a Core program that uses Maps to see how well it works using the Array theory? Perhaps https://github.com/strata-org/Strata/blob/main/StrataTest/Languages/Core/Examples/AdvancedMaps.lean?

@shigoel Added one :)

Strata/Languages/Core/SMTEncoder.lean
aqjune-aws: Indeed if one wants to do approximation (e.g., certain bit-vector operations with UF) there can be lots of options to carry & having a data structure will make it succinct.

shigoel: +1 to a struct config option in the future.

joscoh (approved):

PR #442, opened 2026-02-17
Refactor PySpec type infrastructure and fix correctness bugs

joehendrix:
## Summary

PySpec types (`SpecType`) had no source location tracking, making it
impossible to report precise error locations for type-related diagnostics.
This PR adds a `loc : SourceRange` field to `SpecType` so that error
messages can point back to the source that produced each type.

Adding `loc` required replacing the derived `Ord` and `BEq` instances with
custom ones that compare types semantically (ignoring provenance metadata).
While implementing this, two pre-existing bugs were found and fixed:

- **`unionAux`** pushed the wrong element in the `.gt` case, producing
  incorrectly ordered union arrays. This could cause `binSearchContains`
  membership checks to miss valid members.
- **`ppSourceRange`** compared `spos.line == spos.line` (always true)
  instead of `spos.line == epos.line`, so multi-line ranges always
  rendered in single-line format.

The `SpecError` and `PySpecMClass` types are extracted into a separate
`PySpecM.lean` module and the `noneType` atom variant is unified with the
existing `ident`-based representation, removing the redundant `typeNoneType`
DDM op.

**Note:** Existing `.pyspec.st.ion` cache files will need regeneration.

### DIFF
--- a/Strata/DDM/Util/SourceRange.lean
+++ b/Strata/DDM/Util/SourceRange.lean
@@ -5,6 +5,8 @@
 -/
 module
 
+public import Lean.Data.Position
+
 public section
 namespace Strata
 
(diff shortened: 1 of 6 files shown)


Strata/DDM/Util/SourceRange.lean
```diff
@@ -34,5 +36,16 @@ def isNone (loc : SourceRange) : Bool := loc.start = 0 ∧ loc.stop = 0
 instance : Std.ToFormat SourceRange where
  format fr := f!"{fr.start}-{fr.stop}"
 
+/-- Format a SourceRange as a string using a FileMap for line:column conversion.
+    Renders location information in a format VSCode understands.
+    Returns "path:line:col-col" if on same line, otherwise "path:line:col". -/
+def format (loc : SourceRange) (path : System.FilePath) (fm : Lean.FileMap) : String :=
```
shigoel, line 42: Aside, perhaps out of scope for this PR. Maybe we should format `FileRange` in a similar way. https://github.com/strata-org/Strata/blob/main/Strata/Util/FileRange.lean#L51

joehendrix: I think I ported that, but didn't originally write it.  It has some more options, and I'm not entirely sure of the usage, so I'll consider that out of scope for now.

aqjune-aws (approved):

PR #443, opened 2026-02-18
Eliminate need for forward type declarations in recursive datatypes

joehendrix:
## Summary

Mutually recursive datatypes previously required users to manually write
`forward type` declarations before a `mutual` block. This PR eliminates
that boilerplate by having the elaborator automatically discover and
pre-register type names before elaborating the block's children.

## Key changes

- **Automatic pre-registration replaces forward declarations.** The
  `@[preRegisterTypes(scope)]` annotation on `command_mutual` triggers
  two-phase elaboration: Phase 1 partially elaborates each child to
  extract type names and params, then pre-registers them in the
  `GlobalContext`; Phase 2 elaborates the full block with mutual
  references already resolved.
- **`GlobalContext` simplified.** The `DeclState` enum
  (`.forward` / `.defined`) and its bookkeeping are removed. `define`
  now requires a non-membership proof, making illegal states
  unrepresentable. `defineChecked` takes a `preRegistered` flag to
  handle both fresh definitions and pre-registered confirmations.
- **Error propagation instead of silent failure.**
  `extractConstructorInfo` and helpers return `Except String` instead
  of silently returning empty arrays. Template expansion uses an atomic
  `buildAndDefine` that checks freshness and defines in one step.
- **Template expansion validated during elaboration.**
  `evalBindingSpec`'s `.datatype` branch calls `expandFunctionTemplates`
  to catch constructor and template function name collisions early via
  `logError`, rather than panicking in `addCommand`. Constructor names
  are no longer required to be globally unique — only the generated
  template function names must be distinct.
- **Elaboration refactors.** Extracted `elabSyntaxArg`, `resultContext`,
  `extractParamNames`, and `scopeSepFormat` from the monolithic
  `runSyntaxElaborator`. Lifted `extractConstructorInfo` out of the
  `GlobalContext` namespace (it only depends on `DialectMap`).
- **O(1) argument lookup via `argElabIndex`.** A precomputed
  `Array (Option Nat)` maps each `argLevel` to its position in
  `argElaborators`, replacing linear searches.
- **Translation layer adapted.** `translateMutualBlock` allocates
  placeholder entries on the fly when a datatype isn't already
  pre-registered, rather than requiring forward declarations.

## Test changes

- **`StrataTest/DDM/MutualDatatypes.lean`** (new): DDM-level test
  dialect with function templates (`perConstructor` tester,
  `perField` accessor). Tests mutual block visibility, three-way
  recursion, empty blocks, multi-constructor template expansion, and
  negative cases (undefined type, duplicate type, type clash, duplicate
  constructor name collisions, reused constructor name collisions).
- **`StrataTest/Languages/Core/Examples/MutualDatatypes.lean`** (new):
  Core-level integration tests verifying tester functions, accessor
  functions, equality, polymorphic havoc, and complex Stmt/StmtList
  patterns on mutually recursive types.
- **`StrataTest/Languages/Core/TestASTtoCST.lean`** (new):
  Round-trip tests for AST-to-CST translation.
- Removed all `forward type` declarations from existing test files.

### DIFF
--- a/Strata/DDM/Elab.lean
+++ b/Strata/DDM/Elab.lean
@@ -152,7 +152,6 @@ private def checkDialectName (ld : LoadedDialects) (actual : DialectName) (expec
       .ok ()
   | some expected =>
     if actual = expected then
-      assert! expected ∉ ld.dialects
       .ok ()
     else
       .error s!"Dialect header name of {actual} does not match expected name {expected}."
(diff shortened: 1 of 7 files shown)


Strata/DDM/Elab/Core.lean
```diff
@@ -982,9 +982,49 @@ def getSyntaxArgs (stx : Syntax) (ident : QualifiedIdent) (expected : Nat) : Ela
...
+Extract type parameter names from a partially-elaborated typeParams tree.
+Compares the tree's result context against the inherited binding count
+to find newly introduced type/tvar bindings.
+-/
+private def extractParamNames (inheritedCount : Nat) (tpTree : Tree) : List String :=
+  let tpCtx := tpTree.resultContext
+  let newBindings := tpCtx.bindings.toArray.extract inheritedCount tpCtx.bindings.size
+  let names := newBindings.filterMap fun (b : Binding) =>
+    match b.kind with
+    | .type _ [] _ => some b.ident
```
MikaelMayer: `extractParamNames` matches both `.type _ [] _` and `.tvar _ _`, but the analogous inline code in `runSyntaxElaborator` (around line 1101) only matches `.type _ [] _`:

```lean
|>.filterMap fun b =>
  match b.kind with
  | .type _ [] _ => some b.ident
  | _ => none
```

Is the `.tvar` case intentionally included here but not there? If both should behave the same, consider extracting the shared logic into `extractParamNames` and calling it from both sites.

MikaelMayer (commented):
Clean, well-motivated PR that removes a real source of user friction. The two-phase elaboration design in `elaborateWithPreRegistration` is sound, the `GlobalContext` simplification is a clear win, and the removal of `DeclState`/forward declarations is thorough — no dangling references remain.

One minor inconsistency between `extractParamNames` and the inline code in `runSyntaxElaborator` (`.tvar` handling) is worth a look. The rest is solid.

I'd approve after the one question below is addressed.

Strata/DDM/Elab/Core.lean
joehendrix: I've cleaned up this code quite a bit and added additional test cases.

Strata/DDM/BuiltinDialects/StrataDDL.lean
```diff
@@ -164,7 +164,7 @@ def StrataDDL : Dialect := BuiltinM.create! "StrataDDL" #[initDialect] do
   -- Metadata for marking an operation as a constructor list push (list followed by constructor)
   declareMetadata { name := "constructorListPush", args := #[.mk "list" .ident, .mk "constructor" .ident] }
   declareMetadata { name := "declareType", args := #[.mk "name" .ident, .mk "args" (.opt .ident)] }
-  declareMetadata { name := "declareTypeForward", args := #[.mk "name" .ident, .mk "args" (.opt .ident)] }
+  declareMetadata { name := "preRegisterTypes", args := #[.mk "scope" .ident] }
```
MikaelMayer, line 167: Is this annotation tested somewhere? I don't remember seeing an `@[preRegisterType]` in the tests.

MikaelMayer (approved):

Strata/DDM/BuiltinDialects/StrataDDL.lean
joehendrix: Yes, it's used sparingly, but the essential feature of the PR.

The dedicated DDM test file has it [here](https://github.com/strata-org/Strata/blob/b334bd5e911cc0c44625a0530a5d8fe264e5b19a/StrataTest/DDM/MutualDatatypes.lean#L57) and it is in Strata Core on this [line](https://github.com/strata-org/Strata/blob/b334bd5e911cc0c44625a0530a5d8fe264e5b19a/Strata/Languages/Core/DDMTransform/Grammar.lean#L372).

It'd probably be good to have some additional negative testing around its use.  However, this PR already improves the DDM testing of datatypes beyond what is in main, and I'd like to prioritize other work.

PR #444, opened 2026-02-18
Add function preconditions to Core

joscoh:
*Description of changes:* Adds support for function preconditions to Core. This is done via a transformation `precondElim` that is run before typechecking (to ensure that the generated assertions are typechecked).

Structure:
`Strata/DL/Lambda/Preconditions.lean` generates well-formedness conditions for Lambda expressions
`Strata/Transform/PrecondElim.lean` contains the transformation that 
    1. adds assertions at function call sites to verify that preconditions hold
    2. checks the well-formedness of procedure and function contracts by ensuring that any preconditions of functions used in the contract hold
    3. removes preconditions from functions, making everything total

Tests
`StrataTest/DL/Lambda/PreconditionsTests.lean` contains tests for generating well-formedness conditions for Lambda expressions (i.e. preconditions in called functions)
`StrataTest/Transform/PrecondElim.lean` contains tests of the `precondElim` transformation
`StrataTest/Languages/Core/Examples/FunctionPreconditions.lean` contains tests that involve the end-to-end verification of functions (including function decl statements) with preconditions

Other smaller changes include adding syntax for preconditions to the Core grammar, adding metadata for these preconditions, and adding a precondition that the denominator is nonzero to `Int.div` and `Int.mod` (updating the corresponding proof obligations in tests).

Note that ADT destructors do NOT by default have a precondition that they are called on the correct constructor (e.g. that `List..head` is called on something satisfying `List..isCons`). This will happen in a future PR, and we will include both checked and unchecked versions.

### DIFF
--- a/Strata/DDM/Elab/Core.lean
+++ b/Strata/DDM/Elab/Core.lean
@@ -443,6 +443,9 @@ partial def unifyTypes
         return args
       assert! ea.size = ia.size
       unifyTypeVectors b argLevel0 ea tctx exprSyntax ia args
+    | .tvar _ _ =>
+      -- tvar inferred types are passed through; type inference will catch mismatches
+      pure args
     | _ =>
       logErrorMF exprLoc mf!"Encountered {inferredHead} expression when {expectedType} expected."
       return args
@@ -454,6 +457,9 @@ partial def unifyTypes
         return args
       assert! ea.size = ia.size
       unifyTypeVectors b argLevel0 ea tctx exprSyntax ia args
+    | .tvar _ _ =>
+      -- tvar inferred types are passed through; type inference will catch mismatches
+      pure args
     | ih =>
       logErrorMF exprLoc mf!"Encountered {ih} expression when {expectedType} expected."
       return args
@@ -484,8 +490,11 @@ partial def unifyTypes
     pure args
   | .arrow _ ea er =>
     match inferredType with
-    | .ident .. | .bvar .. | .fvar .. | .tvar .. =>
+    | .ident .. | .bvar .. | .fvar .. =>
       logErrorMF exprLoc mf!"Expected {expectedType} when {inferredType} found"
+      pure args
+    | .tvar _ _ =>
+      -- tvar inferred types are passed through; type inference will catch mismatches
       pure args
     | .arrow _ ia ir =>
       let res ← unifyTypes b argLevel0 ea tctx exprSyntax ia args
--- a/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
+++ b/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
@@ -9,5 +9,4 @@ import Strata.DDM.Integration.Lean
 
 namespace Strata.Laurel
 
-
 #load_dialect "./LaurelGrammar.st"
--- a/Strata/Languages/Laurel/Laurel.lean
+++ b/Strata/Languages/Laurel/Laurel.lean
@@ -10,195 +10,250 @@ import Strata.Languages.Core.Procedure
 import Strata.Util.Tactics
 
 /-
-The Laurel language is supposed to serve as an intermediate verification language for at least Java, Python, JavaScript.
-
-It enables doing various forms of verification:
-- Deductive verification
-- Property based testing
-- Data-flow analysis
-
-Features currently not present:
-- Type inference. The source program needs to specify enough types so that no inference is needed.
-- Type checking. We assume types have already been checked before.
-- Namespaces. All definition and reference names consist of a single Identifier
-
-Design choices:
-- Pure contracts: contracts may only contain pure code. Pure code does not modify the heap, neither by modifying existing objects are creating new ones.
-- Procedures: instead of functions and methods we have a single more general concept called a 'procedure'.
-- Determinism: procedures can be marked as deterministic or not. For deterministic procedures with a non-empty reads clause, we can assume the result is unchanged if the read references are the same.
-- Opacity: procedures can have a body that's transparant or opaque. Only an opaque body may declare a postcondition.
-- StmtExpr: Statements and expressions are part of the same type. This reduces duplication since the same concepts are needed in both, such as conditions and variable declarations.
-- Loops: The only loop is a while, but this can be used to compile do-while and for loops to as well.
-- Jumps: Instead of break and continue statements, there is a labelled block that can be exited from using an exit statement inside of it.
-  This can be used to model break statements and continue statements for both while and for loops.
-
-- User defined types consist of two categories: composite types and constrained types.
-- Composite types have fields and procedures, and may extend other composite types.
-  - Fields state whether they are mutable, which impacts what permissions are needed to access them
-  - Fields state their type, which is needed to know the resulting type when reading a field.
-- Constrained types are defined by a base type and a constraint over that type.
-  - Algebraic datatypes do not exist directly but can be encoded using composite and constrained types.
-
-- For now there is no type polymorphism
-
-- Construction of composite types is WIP. It needs a design first.
-
+Documentation for Laurel can be found in docs/verso/LaurelDoc.lean
 -/
 namespace Strata
 namespace Laurel
 
 abbrev Identifier := String /- Potentially this could be an Int to save resources. -/
 
+/--
+Primitive operations available in Laurel expressions.
+
+Operations are grouped into boolean operations (`Eq`, `Neq`, `And`, `Or`,
+`Not`, `Implies`), arithmetic operations (`Neg`, `Add`, `Sub`, `Mul`, `Div`,
+`Mod`, `DivT`, `ModT`), and comparison operations (`Lt`, `Leq`, `Gt`, `Geq`).
+
+Equality on composite types uses reference equality for impure types and
+structural equality for pure ones.
+-/
 inductive Operation : Type where
-  /- Works on Bool -/
-    /- Equality on composite types uses reference equality for impure types, and structural equality for pure ones -/
-  | Eq | Neq
-  | And | Or | Not | Implies
-  /- Works on Int/Float64 -/
-  | Neg | Add | Sub | Mul | Div | Mod | DivT | ModT
-  | Lt | Leq | Gt | Geq
+  /-- Equality test. Uses reference equality for impure composite types, structural equality for pure ones. -/
+  | Eq
+  /-- Inequality test. -/
+  | Neq
+  /-- Logical conjunction. -/
+  | And
+  /-- Logical disjunction. -/
+  | Or
+  /-- Logical negation. -/
+  | Not
+  /-- Logical implication. -/
+  | Implies
+  /-- Arithmetic negation. Works on `Int` and `Float64`. -/
+  | Neg
+  /-- Addition. Works on `Int` and `Float64`. -/
+  | Add
+  /-- Subtraction. Works on `Int` and `Float64`. -/
+  | Sub
+  /-- Multiplication. Works on `Int` and `Float64`. -/
+  | Mul
+  /-- Euclidean division. Works on `Int` and `Float64`. -/
+  | Div
+  /-- Euclidean modulus. Works on `Int` and `Float64`. -/
+  | Mod
+  /-- Truncation division. -/
+  | DivT
+  /-- Truncation modulus. -/
+  | ModT
+  /-- Less than. Works on `Int` and `Float64`. -/
+  | Lt
+  /-- Less than or equal. Works on `Int` and `Float64`. -/
+  | Leq
+  /-- Greater than. Works on `Int` and `Float64`. -/
+  | Gt
+  /-- Greater than or equal. Works on `Int` and `Float64`. -/
+  | Geq
   deriving Repr
 
+/--
+A wrapper that pairs a value with source-level metadata such as source
+locations and annotations. All Laurel AST nodes are wrapped in
+`WithMetadata` so that error messages and verification conditions can
+refer back to the original source.
+-/
 structure WithMetadata (t : Type) : Type where
+  /-- The wrapped value. -/
   val : t
+  /-- Source-level metadata (locations, annotations). -/
   md : Imperative.MetaData Core.Expression
 
+/--
+The type system for Laurel programs.
+
+`HighType` covers primitive types (`TVoid`, `TBool`, `TInt`, `TFloat64`,
+`TString`), internal types used by the heap parameterization pass (`THeap`,
+`TTypedField`), collection types (`TSet`), user-defined types (`UserDefined`),
+generic applications (`Applied`), value types (`Pure`), and intersection types
+(`Intersection`).
+-/
 inductive HighType : Type where
+  /-- The void type, used for statements that produce no value. -/
   | TVoid
+  /-- Boolean type. -/
   | TBool
+  /-- Arbitrary-precision integer type. -/
   | TInt
-  | TFloat64 /- Required for JavaScript (number). Used by Python (float) and Java (double) as well -/
-  | TString /- String type for text data -/
-  | THeap /- Internal type for heap parameterization pass. Not accessible via grammar. -/
-  | TTypedField (valueType : WithMetadata HighType) /- Field constant with known value type. Not accessible via grammar. -/
-  | TSet (elementType : WithMetadata HighType) /- Set type, e.g. Set Composite. Used in modifies clauses. -/
+  /-- 64-bit floating point type. Required for JavaScript (`number`), also used by Python (`float`) and Java (`double`). -/
+  | TFloat64
+  /-- String type for text data. -/
+  | TString
+  /-- Internal type representing the heap. Introduced by the heap parameterization pass; not accessible via grammar. -/
+  | THeap
+  /-- Internal type for a field constant with a known value type. Introduced by the heap parameterization pass; not accessible via grammar. -/
+  | TTypedField (valueType : WithMetadata HighType)
+  /-- Set type, e.g. `Set int`. -/
+  | TSet (elementType : WithMetadata HighType)
+  /-- A reference to a user-defined composite or constrained type by name. -/
   | UserDefined (name : Identifier)
+  /-- A generic type application, e.g. `List<Int>`. -/
   | Applied (base : WithMetadata HighType) (typeArguments : List (WithMetadata HighType))
-  /- Pure represents a composite type that does not support reference equality -/
+  /-- A pure (value) variant of a composite type that uses structural equality instead of reference equality. -/
   | Pure (base : WithMetadata HighType)
-  /- Java has implicit intersection types.
-     Example: `<cond> ? RustanLeino : AndersHejlsberg` could be typed as `Scientist & Scandinavian`-/
+  /-- An intersection of types. Used for implicit intersection types, e.g. `Scientist & Scandinavian`. -/
   | Intersection (types : List (WithMetadata HighType))
-  /-
-    Type "passed through" from Core. Intended to allow translations to Laurel to refer directly to Core.
-  -/
+  /-- Temporary construct meant to aid the migration of Python->Core to Python->Laurel.
+  Type "passed through" from Core. Intended to allow translations to Laurel to refer directly to Core. -/
   | TCore (s: String)
 
 mutual
 
+/--
+A procedure in Laurel. Procedures are the main unit of specification and
+verification. Unlike separate functions and methods, Laurel uses a single
+general concept that covers both.
+-/
 structure Procedure : Type where
+  /-- The procedure's name. -/
   name : Identifier
+  /-- Input parameters with their types. -/
   inputs : List Parameter
+  /-- Output parameters with their types. Multiple outputs are supported. -/
   outputs : List Parameter
+  /-- The precondition that callers must satisfy. -/
   precondition : WithMetadata StmtExpr
+  /-- Whether the procedure is deterministic or nondeterministic. -/
   determinism : Determinism
+  /-- Optional termination measure for recursive procedures. -/
   decreases : Option (WithMetadata StmtExpr) -- optionally prove termination
+  /-- The procedure body: transparent, opaque, or abstract. -/
   body : Body
+  /-- Source-level metadata. -/
   md : Imperative.MetaData Core.Expression
 
+/--
+Specifies whether a procedure is deterministic or nondeterministic.
+
+For deterministic procedures with a non-empty reads clause, the result can be
+assumed unchanged if the read references are the same.
+-/
 inductive Determinism where
+  /-- A deterministic procedure. The optional reads clause lists the heap locations the procedure may read. -/
   | deterministic (reads : Option (WithMetadata StmtExpr))
+  /-- A nondeterministic procedure. They can read from the heap but there is no benefit from specifying a reads clause. -/
   | nondeterministic
 
+/--
+A typed parameter for a procedure.
+-/
 structure Parameter where
+  /-- The parameter name. -/
   name : Identifier
+  /-- The parameter type. -/
   type : WithMetadata HighType
 
-/- No support for something like function-by-method yet -/
+/--
+The body of a procedure. A body can be transparent (with a visible
+implementation), opaque (with a postcondition and optional implementation),
+or abstract (requiring overriding in extending types).
+-/
 inductive Body where
+  /-- A transparent body whose implementation is visible to callers. -/
   | Transparent (body : WithMetadata StmtExpr)
-/- Without an implementation, the postcondition is assumed -/
+  /-- An opaque body with a postcondition, optional implementation, and modifies clause. Without an implementation the postcondition is assumed. -/
   | Opaque
       (postcondition : List (WithMetadata StmtExpr))
       (implementation : Option (WithMetadata StmtExpr))
       (modifies : List (WithMetadata StmtExpr))
-/- An abstract body is useful for types that are extending.
-    A type containing any members with abstract bodies can not be instantiated. -/
+  /-- An abstract body that must be overridden in extending types. A type containing any members with abstract bodies cannot be instantiated. -/
   | Abstract (postcondition : WithMetadata StmtExpr)
 
-/-
-A StmtExpr contains both constructs that we typically find in statements and those in expressions.
-By using a single datatype we prevent duplication of constructs that can be used in both contexts,
-such a conditionals and variable declarations
--/
-/-
-It would be nice to replace `Type` on the next line with `(isPure: Bool) -> Type`,
-so that we can prevent certain constructors from being used for pure StmtExpr's,
-but when trying that Lean complained about parameters being used in a nested inductive,
-for example in `Option (StmtExpr isPure)`
+/--
+The unified statement-expression type for Laurel programs.
+
+`StmtExpr` contains both statement-like constructs (conditionals, loops,
+assignments, returns) and expression-like constructs (literals, identifiers,
+operations, calls). Using a single type avoids duplication of shared concepts
+such as conditionals and variable declarations.
 -/
 inductive StmtExpr : Type where
-/- Statement like -/
+  /-- Conditional with a then-branch and optional else-branch. -/
   | IfThenElse (cond : WithMetadata StmtExpr) (thenBranch : WithMetadata StmtExpr) (elseBranch : Option (WithMetadata StmtExpr))
+  /-- A sequence of statements with an optional label for `Exit`. -/
   | Block (statements : List (WithMetadata StmtExpr)) (label : Option Identifier)
-  /- The initializer must be set if this StmtExpr is pure -/
+  /-- A local variable declaration with a type and optional initializer. The initializer must be set if this `StmtExpr` is pure. -/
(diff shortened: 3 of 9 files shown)


Examples/expected/LoopSimple.csimp.expected
```diff
@@ -1,10 +1,15 @@
 Successfully parsed.
+ [assert_entry_invariant_calls_Int.Div_0]: ✅ pass
```
MikaelMayer: Why do we have new assertions here? Is it because you have no requires == requires true and that creates assertions? Ideally I would prefer to not mention assertions that were not in the code, especially if these assertions are generated to "true".

Strata/DL/Lambda/Preconditions.lean
```diff
@@ -0,0 +1,122 @@
...
+/--
+Collect all WF obligations from an expression by traversing it and finding
+all calls to functions with preconditions.
+
+For each call to a function with preconditions:
+1. Get the function's preconditions from the Factory
+2. Substitute actual arguments for formal parameters
+3. Create a WFObligation with the call-site metadata
+4. Wrap in enclosing quantifiers and implications
+-/
```
MikaelMayer, line 62: There is one more case: If we have .app (.abs ...) ...

We need to make sure we collect definitions when extracting the WF conditions.

StrataTest/DL/Lambda/PreconditionsTests.lean
```diff
@@ -0,0 +1,97 @@
...
+  | .ok f => f
+  | _ => (@IntBoolFactory TestParams _ _)
+
+
+-- forall x :: (x > 0) ==> (safeDiv(y, x) > 0)
+-- The WF obligation is: forall x :: (x > 0) ==> (x != 0)
+/-- info: [WFObligation(safeDiv, (∀ ((~Bool.Implies : (arrow bool (arrow bool bool))) (~Int.Gt %0 #0) (~!= %0 #0))), ())] -/
+#guard_msgs in
+#eval collectWFObligations factoryWithImplies
+  (.quant () .all LMonoTy.int (.true ())
```
MikaelMayer: Is there a way to make these tests use a human-readable syntax? It would make maintenance and understandability easier.

Strata/DL/Lambda/IntBoolFactory.lean
```diff
@@ -142,6 +120,57 @@ def cevalIntModT (m:T.Metadata) (args : List (LExpr T.mono)) : Option (LExpr T.m
...
+def intSubFunc : LFunc T :=
+  binaryOp "Int.Sub" .int
+  (some (binOpCeval Int Int (@intConst T.mono) LExpr.denoteInt Int.sub))
+
+def intMulFunc : LFunc T :=
+  binaryOp "Int.Mul" .int
+  (some (binOpCeval Int Int (@intConst T.mono) LExpr.denoteInt Int.mul))
+
+def intDivFunc [Inhabited T.mono.base.Metadata] : LFunc T :=
+  -- Precondition: y != 0
```
MikaelMayer: The precondition construction uses `default` for metadata, which means generated assertions won't have source location information. This will make debugging failures harder.

Consider requiring explicit metadata as a parameter to `intDivFunc` and `intModFunc`, or at minimum document that callers should provide proper metadata when these functions are used in contexts where error reporting matters.

Strata/DL/Util/Func.lean
```diff
@@ -24,6 +24,11 @@ open Std (ToFormat Format format)
 /-- Type identifiers for generic type arguments. Alias for String. -/
 abbrev TyIdentifier := String
 
+/-- A precondition with its associated metadata -/
```
MikaelMayer, line 27: `FuncPrecondition` duplicates the pattern of wrapping a value with metadata, which already exists as:
- `Laurel.WithMetadata` (val + md)
- `DDM.Ann` (val + ann)

Consider using one of these existing abstractions or creating a shared `WithMetadata` type in a common location that all three can use.

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,282 @@
...
+/-! ## Contract well-formedness procedures -/
+
+/--
+Generate a well-formedness checking procedure for a procedure's contract.
+
+For each precondition+postcondition (in order):
+  - Assert WF of partial function calls in the condition
+  - Assume the condition (for use by subsequent clauses)
+-/
+def mkContractWFProc (F : @Lambda.Factory CoreLParams) (proc : Procedure)
```
MikaelMayer, line 109: There's duplication between `mkContractWFProc` and `mkFuncWFStmts` in how they process preconditions (both do: collect asserts, add assume). Consider extracting a helper:

```lean
/-- Process a single precondition: assert WF obligations, then assume the precondition -/
def processPrecondition (F : @Lambda.Factory CoreLParams) (precond : FuncPrecondition Expression.Expr Expression.ExprMetadata)
    (labelPrefix : String) (label : String) : List Statement :=
  let asserts := collectAsserts F precond.expr labelPrefix
  let assume := Statement.assume label precond.expr precond.md
  asserts ++ [assume]
```

Then both functions can use this helper, reducing duplication and making the pattern more explicit.

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,282 @@
...
+        outputs := []
+      }
+      spec := { modifies := [], preconditions := [], postconditions := [] }
+      body := wfStmts
+    })
+
+/-! ## Statement transformation -/
+
+/-- Convert a PureFunc to an LFunc for adding to the factory -/
+def pureFuncToLFunc (decl : Imperative.PureFunc Expression) : Function := {
```
MikaelMayer: The `pureFuncToLFunc` conversion is duplicated in at least three other places:
- `Strata/Languages/Core/FunctionType.lean` (line 84)
- `Strata/Languages/Core/StatementWF.lean` (line 148)
- `Strata/Languages/Core/StatementEval.lean` (line 481)

Consider extracting this to a shared helper function in `Strata/Languages/Core/Statement.lean` or similar, since this conversion is needed whenever a `PureFunc` needs to be added to a factory or type-checked.

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,282 @@
...
+    -- Add function to factory before processing its preconditions/body
+    let func := pureFuncToLFunc decl
+    let F' := F.push func
+    let decl' := { decl with preconditions := [] }
+    match mkFuncWFStmts F' funcName decl.preconditions decl.body with
+    | none => ([.funcDecl decl' md], F')
+    | some wfStmts =>
+      -- Add init statements for function parameters so they're in scope
+      let paramInits := decl.inputs.toList.map fun (name, ty) =>
+        Statement.init name ty (.fvar () (CoreIdent.unres ("init_" ++ name.name)) none)
```
MikaelMayer: Note: https://github.com/strata-org/Strata/pull/432 will eliminate the need for these hack init statements with free variables in RHS.

MikaelMayer (commented):
This PR adds comprehensive support for function preconditions to Core, including well-formedness checking and automatic assertion generation. The implementation is well-structured with clear separation between precondition collection (Lambda layer), transformation (PrecondElim), and integration (Core). The test coverage is excellent, demonstrating various scenarios including nested calls, dependent preconditions, and ADT destructors.

The design choice to run precondElim before typechecking is sound, as it allows polymorphic preconditions to be properly unified. The transformation correctly handles both top-level function declarations and inline function declaration statements.

I have a few suggestions for improving code organization and reducing duplication.

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,282 @@
...
+2. For every function and procedure contract, it generates a well-formedness
+check asserting that all calls to functions preconditions within the contract
+hold, assuming earlier calls succeed.
+3. For function declarations, the well-formedness check also asserts the
+preconditions of any partial functions called within the body.
+4. The returned program consists only of total functions (no preconditions).
+
+See StrataTest/Transform/PrecondElim.lean for examples.
+
+Note that this transformation must be called BEFORE typechecking, since
```
MikaelMayer, line 27: Very interesting that we can do this before type checking.

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,282 @@
...
+Collect assertions for all expressions in a command.
+-/
+def collectCmdAsserts (F : @Lambda.Factory CoreLParams)
+  (cmd : Imperative.Cmd Expression) : List Statement :=
+  match cmd with
+  | .init _ _ e md => collectAsserts F e "init" md
+  | .set x e md => collectAsserts F e s!"set_{x.name}" md
+  | .assert l e md => collectAsserts F e s!"assert_{l}" md
+  | .assume l e md => collectAsserts F e s!"assume_{l}" md
+  | .cover l e md => collectAsserts F e s!"cover_{l}" md
```
MikaelMayer: Nice naming with a prefix.

StrataTest/Languages/Core/Examples/FunctionPreconditions.lean
```diff
@@ -0,0 +1,493 @@
...
+#eval verify funcCallsFuncPgm
+
+-- Function body calls another function but precondition does NOT hold (should fail)
+-- Expression: safeDiv(x, 0) - calling with literal 0
+def funcCallsFuncFailPgm :=
+#strata
+program Core;
+
+function badDiv(x : int) : int
+{ x div 0 }
```
MikaelMayer: Not for this PR but why aren't we using the standard division symbol?

MikaelMayer: Do we have a symbol for division that does not have preconditions?

Examples/expected/LoopSimple.csimp.expected
joscoh: It is because I changed `div` to have a precondition, so the invariant (which uses `div`) now has to check to ensure that the precondition is satisfied. That can be reverted if we prefer (I know you asked elsewhere about a total version of `div`). Happy to hear your thoughts.

Strata/DL/Lambda/Preconditions.lean
joscoh: Not sure what you mean exactly, could you give an example? Lambda abstractions cannot (yet) have preconditions.

Strata/DL/Util/Func.lean
joscoh: I don't think we want to use Laurel or the DDM mechanisms in Lambda though.

Strata/DL/Lambda/IntBoolFactory.lean
joscoh: I am unsure of how expression-level metadata is used (as opposed to statement-level metadata which includes file ranges). I can add metadata parameters here, but do you know where expression-level metadata is used to help me get a clearer picture?

Strata/Transform/PrecondElim.lean
joscoh: I think this ties in to a broader question about `Func` vs `LFunc` vs `PureFunc` that we've mentioned a few times. I can create a shared helper function but that doesn't resolve the underlying problem.

Strata/Transform/PrecondElim.lean
joscoh: Yes, once that is merged I will update!

StrataTest/Languages/Core/Examples/FunctionPreconditions.lean
joscoh: Currently, no. Again, that is something we should decide (if the default should be with a precondition with an additional non-precondition version or if we only want one or the other).

Strata/Transform/PrecondElim.lean
joscoh: Hmm, I thought I had fixed that. Thanks, will fix.

Strata/DL/Lambda/Preconditions.lean
MikaelMayer: What I mean is that if you have:

```
let x := a
2 div x
```
then when lifting the precondition that x != 0 you will want to actually extract
```
let x := a
x != 0
```

The let expression can be encoded as `.app (.abs ...) (let rhs)` in LExpr.

Strata/DL/Lambda/IntBoolFactory.lean
MikaelMayer: Ah I misread here. It looks like we are only defining a built-in precondition. We should probably use the metadata at the call site of this function anyway. Until now, expressions were total so their metadata was not used to report errors. But with this PR it seems that we should verify that the error reporting is done one the expression source.

Strata/DL/Util/Func.lean
MikaelMayer: I think that `Laurel.WithMetadata` should be moved closer to some utils so that you can reuse it.

Strata/Transform/PrecondElim.lean
MikaelMayer: What was the broader question?

The only reason why we have LFunc, Func and PureFunc is that the expression language is a type parameter of imperative when it's explicit in Lambda.  I think it's the price to pay to have type parameters.

StrataTest/Languages/Core/Examples/FunctionPreconditions.lean
MikaelMayer: Whatever the default is, I need to have at least one symbol that aligns with the SMT semantics of division without precondition.
I would suggest that `/` has preconditions when `div` has no preconditions. Thoughts?

Strata/Transform/PrecondElim.lean
joscoh: As in, the fact that one uses `LTy` and one uses `LMonoTy`. I know @shigoel had discussed trying to bridge the gap. But for now, I can create a common conversion function.

StrataTest/Languages/Core/Examples/FunctionPreconditions.lean
joscoh: That is fine. Do you want me to change the existing examples (e.g. in `Loops.lean`) to go back to using the non-precondition version?

Strata/DL/Util/Func.lean
joscoh: `Laurel.WithMetadata` involves `Imperative.Metadata`, which should not be used in Lambda either. It is possible to refactor these 3 to all use `Ann`, but that is a decently large refactor (at least for Laurel). Might be better for a future PR.

shigoel: Agreed. I've been wanting Imperative.Metadata to be in Strata/Util, or actually, even in Strata/DL if you view metadata as yet another structured language.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -742,6 +742,13 @@ def funcDeclToStatement {M} [Inhabited M] (decl : Imperative.PureFunc Expression
   let inline? : Ann (Option (Inline M)) M := ⟨default, none⟩
   -- Add formals to the context
   modify (·.addScopedBoundVars (reverse? := false) paramNames)
+  -- Convert preconditions
```
shigoel, line 759: Nit: looks like can be refactored into a helper function esp. since this is also used in `funcToCST`?

Strata/DL/Lambda/Preconditions.lean
```diff
@@ -0,0 +1,130 @@
...
+      | none => []
+    let subObligations := match e with
+      | .const _ _ | .op _ _ _ | .bvar _ _ | .fvar _ _ _ => []
+      -- Need to quantify over bound variable
+      -- e.g. λ x => 2 / x gives precondition ∀ x, x != 0
+      | .abs md ty body =>
+        (go F body implications).map fun ob =>
+          { ob with obligation := .quant md .all ty (.bvar md 0) ob.obligation }
+      | .quant md _ ty trigger body =>
+        let triggerObs := go F trigger implications
```
shigoel: Do we need to collect any obligations for trigger expressions?

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -357,7 +357,9 @@ partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Conte
     | "Int.Sub"      => .ok (.app Op.sub,        .int ,   ctx)
     | "Int.Mul"      => .ok (.app Op.mul,        .int ,   ctx)
     | "Int.Div"      => .ok (.app Op.div,        .int ,   ctx)
+    | "Int.SafeDiv"  => .ok (.app Op.div,        .int ,   ctx)
```
shigoel, line 362: A reminder comment that at the SMT level, why the safe versions of these functions map exactly to the unsafe ones would be welcome (i.e., Core already did precond elim and so have WF conditions in the env).

Strata/DL/Util/Func.lean
```diff
@@ -70,6 +75,7 @@ structure Func (IdentT : Type) (ExprT : Type) (TyT : Type) (MetadataT : Type) wh
   -- resulting expression of concreteEval if evaluation was successful.
   concreteEval : Option (MetadataT → List ExprT → Option ExprT) := .none
   axioms   : List ExprT := []  -- For axiomatic definitions
+  preconditions : List (FuncPrecondition ExprT MetadataT) := []
```
shigoel, line 78: Would be nice if we could add a label to each precondition (so that we can use that label to flag any related errors), but I suppose that could also go into the metadata. We haven't added labels to metadata yet though....

Strata/Languages/Core/Function.lean
```diff
@@ -26,6 +26,20 @@ instance : DecidableEq CoreLParams.IDMeta :=
 instance : ToFormat CoreLParams.IDMeta :=
   show ToFormat Visibility from inferInstance
 
+/-- Convert a `PureFunc Expression` (with polytypes) to a `Function` (with monotypes). -/
+def Function.ofPureFunc (decl : Imperative.PureFunc Expression) : Function := {
+  name := decl.name
+  typeArgs := decl.typeArgs
+  isConstr := decl.isConstr
+  inputs := decl.inputs.map (fun (id, ty) => (id, Lambda.LTy.toMonoTypeUnsafe ty))
+  output := Lambda.LTy.toMonoTypeUnsafe decl.output
```
shigoel: Makes sense to raise an exception if the `LTy` cannot be safely converted to `LMonoTy`. The caller of this function in FunctionType.lean does use the `Except` monad.

StrataTest/DL/Lambda/PreconditionsTests.lean
shigoel, line 1: Aside: Maybe it's time to have DDM-defined concrete syntax for Lambda...

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,280 @@
...
+def wfProcName (name : String) : String := s!"{name}{wfSuffix}"
+
+/-! ## Collecting assertions from expressions -/
+
+/--
+Given a Factory and an expression, collect all partial function call
+precondition obligations and return them as `assert` statements.
+The metadata from the original statement is attached to the generated assertions.
+-/
+def collectAsserts (F : @Lambda.Factory CoreLParams) (e : Expression.Expr)
```
shigoel: May be confused with this `Statement.collectAsserts`... Maybe rename to something that indicates that you're collecting the preconditions of partial functions?

Strata/DL/Lambda/Preconditions.lean
joscoh: Hmm I guess maybe not, since triggers have no semantic meaning.

StrataTest/Languages/Core/Examples/Loops.lean
shigoel, line 1: May I request a test where there's a partial function used in the loop's guard?

Strata/DL/Util/Func.lean
joscoh: That is basically the reason I added metadata, to give us the capability to do something like this in the future. Right now, the metadata here is not being using for anything. Do you have any suggestions on how to do this?

shigoel: As I commented elsewhere, I've been musing a move of Imperative/Metadata out to DL/Metadata, and perhaps adding a mechanism to enforce that certain fields are mandatory. Then we could add a field called `.label` that's mapped to a string value, and we'd be able to enforce that metadata in this place must have a label.  

I hope I answered the question you asked... :)

Strata/DL/Lambda/Preconditions.lean
shigoel: Yes, I'd love to not have extra obligations then!

Strata/Transform/PrecondElim.lean
shigoel: Let's deal with the LFunc vs. PureFunc vs. Func issue in another PR.

Strata/DL/Util/Func.lean
shigoel: To be clear, I did intend for metadata modifications to be out of scope for this PR. Thanks!

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,281 @@
...
+/--
+Transform an entire program:
+1. For each procedure, transform its body and if needed generate a WF procedure
+2. For each function, strip preconditions and if needed generate a WF procedure
+3. For each function call, assert that the preconditions hold
+
+Returns the transformed program.
+-/
+def precondElim (p : Program) (F : @Lambda.Factory CoreLParams)
+    : Except DiagnosticModel Program := do
```
aqjune-aws: Transformations are using `CoreTransformM (Bool \times Program)` where the first Bool returns true if there was any change, and the second Program returns the updated program. Can we use this monad?

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,281 @@
...
+    | [] => return ([], F)
+    | d :: rest =>
+      match d with
+      | .proc proc md =>
+        let (body', _) ← transformStmts F proc.body
+        let proc' := { proc with body := body' }
+        let procDecl := Decl.proc proc' md
+        let (rest', F') ← transformDecls F rest
+        match mkContractWFProc F proc with
+        | some wfDecl => return (wfDecl :: procDecl :: rest', F')
```
aqjune-aws: I am guessing that the generated procedures will be filtered out by the `FilterProcedures` pass which is mandatorily run during `Core.verify`. The pass removes all procedures that are unreachable.
Should FilterProcedures not filter some procedures that are marked as necessary..? These procedures will contain well-formedness assertions on the contracts of the procedures and functions. cc'ing @shigoel as well

Strata/Transform/PrecondElim.lean
```diff
@@ -0,0 +1,281 @@
...
+        | some wfDecl => return (wfDecl :: funcDecl :: rest', F'')
+        | none => return (funcDecl :: rest', F'')
+      | .type (.data block) _ =>
+        let bf ← Lambda.genBlockFactory (T := CoreLParams) block
+        let F' ← F.addFactory bf
+        let (rest', F'') ← transformDecls F' rest
+        return (d :: rest', F'')
+      | _ =>
+        let (rest', F') ← transformDecls F rest
+        return (d :: rest', F')
```
aqjune-aws: Can we split `precondElim` into two subtransformations: one for inserting asserts in the statement level, and one for inserting extra procedures at the top-level for checking contracts?

For the first sub-transformation, I think  `CoreTransform.runProcedures` can be used to implement this pass, after a small extension.
`CoreTransform.runProcedures` currently takes a function `f : Command -> CoreTransformM (Option (List Statement))`. It internally uses runStmtsRec to traverse the subblocks in the statements using `f`.
Can we extend CoreTransform.runProcedures to take a slightly more generic function `f` that receives `Statement` and returns `CoreTransformM (Option (List Statement))`? It can be passed to `runStmtsRec` which will still do the basic recursive traversal on the subblocks & update the `changed` Bool value.

For the second sub-transformation, the existing `CoreTransform.runProcedures` might not fit well, and probably will have to be written manually as the `FilterProcedures` pass did. However, the top-level `precondElim` pass will be a nice monadic sequential execution of the first sub-transformation, and the second sub-transformation.

joscoh: This does not handle all the cases. Function declaration statements create blocks containing the well-formedness conditions. So we could separate it into (insert asserts + insert well-formedness for function statements) and (insert well-formedness for top-level functions), though this distinction feels pretty artificial.

Also, see comment above for why we would need more substantial changes to `CoreTransform.runProcedures` (in addition to extending `Command` to `Statement`). That being said, I do like the idea of defining transformations compositionally using generic machinery.

aqjune-aws: I think the well-formedness block for `.funcDecl` is also Statement, so still seems to fit in the signature. But I can see that `transformDecls` is also returning an updated Factory - could be a state of CoreTransformM.

I was wondering whether this refactoring is possible, because the recursive traversal of Statement seems appearing universally across different transformations. It will help removal of duplication of `changed` boolean flag update handling.

Strata/Transform/PrecondElim.lean
joscoh: One complication: this transformation threads through a `Factory` to keep track of function declaration statements that are encountered, so we know the appropriate preconditions. So we would need to change the definition of `CoreTransformM`, which would unfortunately change things for all transformations. Let me know if you think this larger refactor is worth it.

aqjune-aws: I think `Factory` can be added as a field of `CoreTransformState` as an optional field: `currentFactory: Option Factory`.  This will minimize affecting other transformations.

Strata/Transform/PrecondElim.lean
joscoh: Yes, it fits within the signature, but the proof will be duplicated (which might have to be the case anyway). I'm not sure what you mean by "It will help removal of duplication of changed boolean flag update handling."

aqjune-aws: It is the `changed` result handling in `runStmtsRec`, e.g., `return (changed || changed', [.ite c thenb' elseb' md])` for the if-then-else case. This update will have to be in the precondElim anyway.

Strata/Transform/PrecondElim.lean
shigoel: Thanks, @aqjune-aws, for pointing this out. Indeed, I'd forgotten about the `FilterProcedures` transform. Would it be possible for this transform to not generate a new procedure, but add the body containing the transformed statements (`body'` above) in the procedure that's undergoing transformation (`proc` above)?

joscoh: This is not possible for function definitions. We have to check the well-formedness of the contract somewhere, and so we have to create a new procedure.

shigoel: Will there be a call to the generated procedure that checks the contracts in the procedure that contains the call to the partial function? If so, then the generated procedure won't be unreachable, and wouldn't be filtered out by the `FilterProcedures` transform. Or I guess I'm saying: can we have WF checks for the partial function right before its call?

shigoel: Or.. we can define a separate category of procedures that are marked `@[noFilter]`, as June suggested, and mark the WF procedures as such.

joscoh: I don't think we can ensure that the procedures are called. For example, suppose the entire file is a pure function e.g.
```
function foo (x: int)
requires x \ x > 0
{ ... }
```
Then there is nowhere we could possibly call the generated WF procedure. I think that June's suggestion is probably the way to go.

joscoh: Added a flag that ensures that the generated procedures are not removed by `FilterProcedures`

Strata/Languages/Core/Core.lean
```diff
@@ -69,16 +70,17 @@ def formatProofObligations (obs : Array (Imperative.ProofObligation Expression))
 def typeCheckAndPartialEval (options : Options) (program : Program)
     (moreFns : @Lambda.Factory CoreLParams := Lambda.Factory.default) :
     Except DiagnosticModel (List (Program × Env)) := do
+  let factory ← Core.Factory.addFactory moreFns
+  let program ← match Transform.run program (PrecondElim.precondElim · factory) with
```
shigoel: The filter procedures transform is applied in `Core.verify`. I'd like all the transforms to be applied in one place, just so it's clear what transforms are on by default. Happy to separate this out in another PR.

joscoh: This makes me realize that currently `PrecondElim` is run after `FilterProcedures` anyway, which was the reason I didn't encounter this issue.

shigoel: While you're at it: mind leaving some guidance in `Core.verify` doc string that we expect all program-wide transforms to be applied here?

joscoh: Some transformations should run after typechecking, right? If everything is in `Core.verify`, that implies that the only places for transformations are 1. Before typechecking and 2. After partial eval. Is this accurate?

shigoel: I should've been more precise: `Core.verify` should have program-wide transforms that occur before any kind of analyses, including type inference.

PR #446, opened 2026-02-18
Add PySpec-to-Laurel translation and method call dispatch

joehendrix:
## Summary

Enable end-to-end verification of Python programs that call AWS SDK methods by translating PySpec type stubs into Core prelude declarations and resolving method calls to the correct prelude procedures.

### PySpec-to-Laurel translation (`Specs/ToLaurel.lean`, new)

PySpec files describe Python library APIs as type signatures in Ion format. This PR adds a translator that converts those signatures into Laurel declarations (opaque procedures and composite types) which are then lowered to Core and merged into the verification prelude. This gives the verifier knowledge of library types and method signatures that were previously invisible.

The translator also builds an **overload dispatch table** — a map from factory function names to their string-literal-dispatched return types (e.g., `boto3.client("iam")` → `AWSIdentityManagementV20100508`). This is stored as a `HashMap String (HashMap String PythonIdent)` for O(1) lookup.

Types not yet supported in CorePrelude (e.g., `typing.Any`, `bytes`, `Optional[float]`) are mapped to `TString` via an `unsupportedType` placeholder, with TODO comments marking them for future CorePrelude additions.

### Method call dispatch (`PythonToLaurel.lean`)

The Python-to-Laurel translator previously rendered all method calls (e.g., `iam.get_role(...)`) as `pyExprToString`, producing names like `"iam_get_role"` that never matched any prelude procedure — so every SDK call became a `Hole` (opaque havoc). This PR adds dispatch logic that:

1. **Resolves factory calls via the dispatch table**: matches `boto3.client("iam")` against the overload table and produces a `New` expression for the appropriate service type.
2. **Resolves method calls on typed variables**: looks up the variable's type in context, constructs the correct prelude procedure name (e.g., `AWSIdentityManagementV20100508_get_role`), and prepends the object as the `self` argument.
3. **Falls back to a static call by flattened name** for unrecognized calls.

All three steps live inside `translateCall`, which takes the raw Python call expression directly. The `.Call` arm in `translateExpr` is a single delegation to `translateCall`.

**AnnAssign dispatch fallback**: Python type-stub names (e.g., `IAMClient`) don't match PySpec service names (e.g., `AWSIdentityManagementV20100508`). When an `AnnAssign` annotation resolves to `PyAnyType` and the initializer is a dispatched factory call, the translator falls back to the dispatch table to infer the correct `UserDefined` type. This ensures subsequent method calls on the variable resolve correctly.

Unrecognized type annotations in Python source are mapped to `TString` via `pyUnsupportedType`, allowing translation to proceed without modifying the CorePrelude.

### CLI integration (`StrataMain.lean`)

- `pyAnalyzeLaurel --pyspec <ion_file>` translates PySpec files and merges them into the Core prelude before verification.
- `pyAnalyzeLaurel --dispatch <ion_file>` extracts only the overload dispatch table (no Laurel translation) for lightweight factory-call resolution.
- `pySpecToLaurel` command for standalone PySpec-to-Laurel inspection.
- `buildPySpecPrelude` handles the multi-file merge pipeline with name collision detection.

### PySpec infrastructure (`Specs/Decls.lean`, `Specs.lean`, `Specs/DDM.lean`)

- `Inhabited` instance added to `PythonIdent` (needed for HashMap default values).
- `TypedDict` fields changed from `isTotal : Bool` to per-field `fieldRequired : Array Bool` to support `Required[T]`/`NotRequired[T]` annotations.
- Support for `Unpack[TypedDict]` in `**kwargs` expansion.

## Testing

`ToLaurelTest.lean` adds 15 `#guard_msgs` tests covering the PySpec-to-Laurel translation: primitive and complex type mapping, optional type patterns (`Union[None, T]`), literal and TypedDict types, class/type definitions, void returns, error cases (unknown types, empty types, unsupported unions), overload dispatch with both extern and local class return types, `extractOverloads` filtering, and a regression test for `externTypeDecl`.

### DIFF
--- a/Strata/Backends/CBMC/CoreToCBMC.lean
+++ b/Strata/Backends/CBMC/CoreToCBMC.lean
@@ -181,8 +181,6 @@ def blockToJson {P : Imperative.PureExpr} (I : Lambda.LExprParams) [IdentToStr (
     ]),
     ("sub", Json.arr (b.map (stmtToJson (I:=I) · loc)).toArray)
   ]
-  termination_by (Imperative.Block.sizeOf b)
-  decreasing_by term_by_mem [Stmt, Imperative.sizeOf_stmt_in_block]
 
 def stmtToJson {P : Imperative.PureExpr} (I : Lambda.LExprParams) [IdentToStr (Lambda.Identifier I.IDMeta)] [HasLExpr P I]
   (e : Imperative.Stmt P Command) (loc: SourceLoc) : Json :=
@@ -204,8 +202,6 @@ def stmtToJson {P : Imperative.PureExpr} (I : Lambda.LExprParams) [IdentToStr (L
       ])
     ]
   | _ => panic! "Unimplemented"
-  termination_by (Imperative.Stmt.sizeOf e)
-  decreasing_by all_goals term_by_mem
 end
 
 def listToExpr (l: ListMap CoreLabel Core.Procedure.Check) : Core.Expression.Expr :=
--- a/Strata/Backends/CBMC/StrataToCBMC.lean
+++ b/Strata/Backends/CBMC/StrataToCBMC.lean
@@ -302,8 +302,6 @@ def blockToJson (b: Imperative.Block Strata.C_Simp.Expression Strata.C_Simp.Comm
     ]),
     ("sub", Json.arr (b.map (stmtToJson · loc)).toArray)
   ]
-  termination_by b.sizeOf
-  decreasing_by term_by_mem [Stmt, Imperative.sizeOf_stmt_in_block]
 
  def stmtToJson (e : Strata.C_Simp.Statement) (loc: SourceLoc) : Json :=
   match e with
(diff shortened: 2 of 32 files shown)


StrataTest/Languages/Python/README.md
shigoel, line 1: Thanks for your changes on this file! Very helpful!

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -0,0 +1,394 @@
...
+              '{specTypeToString func.returnType}' is not a \
+              class type"
+          return
+  pushOverloadEntry func.name literalValue retType
+
+/-- Convert a single PySpec signature to Laurel declarations. -/
+def signatureToLaurel (sig : Signature) : ToLaurelM Unit :=
+  match sig with
+  | .externTypeDecl .. =>
+    pure ()
```
shigoel, line 348: Does an externTypeDecl not map to anything here?

shigoel (dismissed):

Strata/Languages/Python/Specs/ToLaurel.lean
joehendrix: Yes, I'll add a comment; they aren't really needed after all since PySpec fully qualifies imported classes.

joehendrix:
Hi Siva, you are right.  If every assignment is annotated or inferable from class definitions (for cases like `self.client = static_method('service_name')`) and there is consistency between the  class names in PySpec and user annotations, then the static dispatch code may not be needed.

Currently neither really seems to be the case, so I'd like to unblock this.  I think it's better to try to get things working now then postpone for future changes.

Issue #448, opened 2026-02-19
Some StrataMain subcommands require more verbose help text


tautschnig:
I'm trying to understand what `lake exe strata` will allow me to do. Yet some help descriptions require more verbosity:
```
$ lake exe strata --help
[...]
Usage: strata <command> [flags]...

  javaGen <dialect-file> <package> <output-dir>
      Generate Java classes from a DDM dialect file.
  check <file>
      Check a dialect or program file.
```
I don't know what is meant by "check" - what properties are being tested for?
```
  toIon <input> <output>
      Read a Strata text file and translate into Ion.
```
Strata has a collection of dialects, so I don't actually know what a "Strata text file" refers to. Would that be a Core or Laurel file? How is it determined which input dialect it is?
```
  print <file>
      Write a Strata text or Ion file to standard output.
```
Is that the same as the `cat` UNIX utility? I suspect there is more to it, but I don't know what it is.

Issue #447, opened 2026-02-18
Move non-DDM commands out of StrataMain


tautschnig:
See https://github.com/strata-org/Strata/pull/405#issuecomment-3923825014: `lake exe strata ...` will now do both DDM commands as well as uses of Strata as a program analysis tool. We should aim for a user experience that has the least amount of surprise and should, therefore, move the non-DDM commands to a separate CLI.

tautschnig:
I think groups is a great idea, but I'd still like to see a bit more separation by use case - perhaps that just means different group names? On a cursory look it would seem that both "Python" and "Java" groups are for similar use cases, just different input languages. The actual subcommands, however, are entirely different: `javaGen` produces files so as to use Strata from a Java program, while all of the `py*` subcommands seem related to analysis of Python code.

joehendrix:
Yes, I debated whether to put the command in a separate Java group or just put it in Core.  I decided Java was as good as name as any.

Do you have a concrete new group scheme?  I just want the extra flag functionality of #451 for another use case - I do not care at all what the category names are.

PR #451, opened 2026-02-19
Add optional flag support and improve CLI help output

joehendrix:
## Summary

Overhaul the strata CLI so that help output is organized by command group and flags are declared per-command rather than globally.

## Details

The previous `--help` output was a flat list of commands with minimal descriptions, and `--include` was handled as a global flag even though only some commands use it. This made the CLI harder to learn and extended awkwardly as new commands were added.

- Organize `--help` output into groups (Core, Python, Laurel, Java) with shared flags shown once per group.
- Add per-command help via `strata <command> --help`.
- Rewrite command descriptions to explain inputs, outputs, and purpose.
- Move `--include` from a global flag to a per-command flag on the commands that actually use it (`check`, `toIon`, `print`, `diff`, `javaGen`).
- Replace the positional verbose argument on `pyAnalyze` and `pyAnalyzeLaurel` with a `--verbose` flag.
- Fix `strata <cmd> --help` when mixed with other flags.
- Per-command errors now suggest `strata <cmd> --help` instead of the global help.
- Add error message and help output regression tests to `testStrataCommand.sh`.
- Update `run_py_analyze.sh` and its README to match the new CLI.

## Example output

<details>
<summary><code>strata --help</code></summary>

```
Usage: strata <command> [flags]...

Command-line utilities for working with Strata.

Core:
  check <file>
    Parse and validate a Strata file (text or Ion). Reports errors and exits.

  toIon <input> <output>
    Convert a Strata text file to Ion binary format.

  print <file>
    Pretty-print a Strata file (text or Ion) to stdout.

  diff <file1> <file2>
    Compare two program files for syntactic equality. Reports the first
    difference found.

  Common flags:
    --include <path>  Add a dialect search path.

Code Generation:
  javaGen <dialect-file> <package> <output-dir>
    Generate Java source files from a DDM dialect definition. Writes .java files
    under output-dir.

    Flags:
      --include <path>  Add a dialect search path.

Python:
  pyAnalyze <file>
    Verify a Python Ion program. Translates to Core, inlines procedures, and
    runs SMT verification.

    Flags:
      --verbose  Enable verbose output.

  pyAnalyzeLaurel <file>
    Verify a Python Ion program via the Laurel pipeline. Translates Python to
    Laurel to Core, then runs SMT verification.

    Flags:
      --verbose  Enable verbose output.

  pyTranslate <file>
    Translate a Python Ion program to Core and print the result to stdout.

  pySpecs <python_path> <strata_path>
    Translate a Python specification (.py) file into Strata DDM Ion format.
    Creates the output directory if needed. (Experimental)

Laurel:
  laurelAnalyze
    Verify Laurel Ion programs read from stdin and print diagnostics. Combines
    multiple input files.
```

</details>

## Related

- #447 — Move non-DDM commands out of StrataMain
- #448 — Some StrataMain subcommands require more verbose help text

### DIFF
--- a/.github/scripts/testStrataCommand.sh
+++ b/.github/scripts/testStrataCommand.sh
@@ -48,6 +48,8 @@ expect_output() {
 }
 
 # --- Error message and help output tests ---
+# Disable set -e so we can test commands that intentionally fail.
+set +e
 
 # Error cases
 expect_error "no args"            "Expected subcommand"              $strata
@@ -65,6 +67,8 @@ expect_error "hint for diff"      "strata diff --help"               $strata dif
 expect_output "global help"            "Command-line utilities"      $strata --help
 expect_output "per-command help"       "Usage: strata check"         $strata check --help
 expect_output "help with other flags"  "Usage: strata check"         $strata check --include foo --help
+
+set -e
 
 # --- Ion round-trip tests ---
 
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -546,13 +546,13 @@ def commandGroups : List CommandGroup := [
   { name := "Core"
     commands := [checkCommand, toIonCommand, printCommand, diffCommand]
     commonFlags := [includeFlag] },
+  { name := "Code Generation"
+    commands := [javaGenCommand] },
   { name := "Python"
     commands := [pyAnalyzeCommand, pyAnalyzeLaurelCommand, pyTranslateCommand,
                  pySpecsCommand] },
   { name := "Laurel"
     commands := [laurelAnalyzeCommand] },
-  { name := "Java"
-    commands := [javaGenCommand] },
 ]
 
 def commandList : List Command :=


.github/scripts/testStrataCommand.sh
```diff
@@ -9,11 +9,75 @@ echo "Storing temporary results in $temp_dir"
...
+  fi
+}
+
+# Expect a zero exit and a substring in stdout.
+expect_output() {
+  local description="$1" expected="$2"
+  shift 2
+  output=$("$@" 2>&1)
+  rc=$?
+  if [ $rc -ne 0 ]; then
```
tautschnig, line 37: Is this actually meaningful under `set -e`? I could be wrong, but I don't think line 37 will be reached if the command in line 35 had a non-zero exit code.

StrataMain.lean
```diff
@@ -453,56 +519,144 @@ def laurelAnalyzeCommand : Command where
...
+def commandGroups : List CommandGroup := [
+  { name := "Core"
+    commands := [checkCommand, toIonCommand, printCommand, diffCommand]
+    commonFlags := [includeFlag] },
+  { name := "Python"
+    commands := [pyAnalyzeCommand, pyAnalyzeLaurelCommand, pyTranslateCommand,
+                 pySpecsCommand] },
+  { name := "Laurel"
+    commands := [laurelAnalyzeCommand] },
+  { name := "Java"
```
tautschnig: Would it be ok to use something like "Strata Bindings" instead? I think all of the above groups use the _input_ language as the group descriptor, while "Java" is actually the output here (and Strata itself is the input).

.github/scripts/testStrataCommand.sh
joehendrix: Good catch — you're right, `set -e` would abort before reaching the error-checking logic. Fixed in 005a2c90 by disabling `set -e` (`set +e`) around the error message tests and re-enabling it before the Ion round-trip tests.

StrataMain.lean
joehendrix: Renamed to "Code Generation" in 005a2c90.

PR #452, opened 2026-02-19
[WIP] Remove sorry warnings via Kiro-generated proofs

tautschnig:
Summary of remaining uses of `sorry` after the changes in this PR:

| File | Active sorry's | Commented-out | Warning-suppressed |
|------|---------------|---------------|-------------------|
| ProcedureWF.lean | 0 | 9 (in /-...-/) | yes |
| StatementWF.lean | 4 | 0 | yes |
| ProgramWF.lean | 1 | 0 | yes |
| CmdEval.lean | 0 | 2 (in /-...-/) | n/a |
| CallElimCorrect.lean | 4 | 0 | yes |
| HashCommands.lean | 0 | 0 | n/a (programmatic use) |

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #453, opened 2026-02-19
Partially prove Procedure.typeCheckWF (5 of 9 fields)

tautschnig:
Replace the full 'by sorry' with a structured proof that establishes 5 of 9 WFProcedureProp fields from the type checker:

  - wfstmts: from Statement.typeCheckWF
  - ioDisjoint: from checkVariableScoping success
  - inputsNodup: from checkNoDuplicates success
  - outputsNodup: from checkNoDuplicates success
  - modNodup: from checkNoDuplicates success

The remaining 4 fields (wfloclnd, inputsLocl, outputsLocl, wfspec) are not currently enforced by the type checker and remain as sorry.

Kiro spent 21 minutes on this partial proof.


Strata/Languages/Core/ProcedureType.lean
```diff
@@ -48,6 +48,32 @@ private def checkVariableScoping (proc : Procedure) (sourceLoc : FileRange) :
               Formals: {proc.header.inputs.keys}\n\
               Returns: {proc.header.outputs.keys}"
 
+theorem checkNoDuplicates_ok {proc : Procedure} {sl : FileRange} {u : Unit} :
+    proc.checkNoDuplicates sl = Except.ok u →
+    proc.header.inputs.keys.Nodup ∧
+    proc.header.outputs.keys.Nodup ∧
+    proc.spec.modifies.Nodup := by
+  unfold checkNoDuplicates
+  simp only [bind, Except.bind, pure, Except.pure]
```
joscoh, line 57: Everything after this line in this proof can be replaced with `grind`

Strata/Languages/Core/ProcedureWF.lean
```diff
@@ -29,7 +29,40 @@ theorem snd_values_mem {ps : ListMap CoreLabel Procedure.Check} :
     case inr mem => right ; exact (ih mem)
   case nil => cases Hin
 
-theorem Procedure.typeCheckWF : Procedure.typeCheck C T p pp md = Except.ok (pp', T') → WFProcedureProp p pp := by sorry
+set_option maxHeartbeats 1600000 in
```
joscoh: I don't think these options are needed; there is nothing computation-heavy in this proof. And these options make things brittle so it best to avoid if not needed.

aqjune-aws (dismissed):
I approved this pull request because the added theorems looked useful and the partial proofs looked ok (one more approval would be necessary, though).

Strata/Languages/Core/ProcedureType.lean
tautschnig: Done.

Strata/Languages/Core/ProcedureWF.lean
tautschnig: Done.

tautschnig:
> Generally I'd recommend finishing proofs before submitting PRs. I don't think we waste time reviewing LLM failures to finish tasks.

What are you taking as an "LLM failure" here? The fact that 4 proofs were left unproven? If so: those actually require changes in the type checker to enable their proof. See https://github.com/strata-org/Strata/pull/452 for the full resolution, this was just my attempt to break this up into reviewable chunks. If something else is your definition of "LLM failure," then please explain.

PR #454, opened 2026-02-19
Prove EvalExpressionIsDefined by adding definedness propagation to WellFormedCoreEvalCong

tautschnig:
Add five new fields to WellFormedCoreEvalCong (absdef, appdef, eqdef, quantdef, itedef) asserting that if a compound expression is defined, its subexpressions are also defined. This enables proving all five remaining cases (abs, quant, app, ite, eq) of EvalExpressionIsDefined by structural induction.

Kiro spent 21 minutes on this proof.


Strata/Languages/Core/StatementSemantics.lean
```diff
@@ -72,7 +80,13 @@ structure WellFormedCoreEvalCong (δ : CoreEval): Prop where
       δ σ e₂ = δ σ' e₂' →
       δ σ e₃ = δ σ' e₃' →
       (δ σ (.ite m e₃ e₁ e₂) = δ σ' (.ite m e₃' e₁' e₂')))
+    /-- Definedness-propagation properties for compound expressions. -/
+    definedness : WellFormedCoreEvalDefinedness δ
 
+/-- Bundle both congruence and definedness properties for a core evaluator. -/
+structure WellFormedCoreEval (δ : CoreEval) : Prop where
```
joscoh: Is this structure used anywhere?

Strata/Languages/Core/StatementSemanticsProps.lean
```diff
@@ -2102,8 +2099,26 @@ theorem EvalExpressionIsDefined :
...
-  case abs => sorry
-  case quant => sorry
-  case app => sorry
-  case ite => sorry
-  case eq => sorry
+  case abs m ty e ih =>
+    exact ih (Hwfc.definedness.absdef σ m ty e Hsome) v Hin
+  case quant m k ty tr e trih eih =>
+    have ⟨htr, he⟩ := Hwfc.definedness.quantdef σ m k ty tr e Hsome
+    rcases Hin with Hin | Hin
```
joscoh: Optional, since it is a matter of style, but in all of these cases, after the `have ... := Hwfc.definedness...` line, the proof can be finished with `grind`. Generally, when the proof is just cases and applications of hypotheses, it is nicer to just let `grind` do the work.

Strata/Languages/Core/StatementSemantics.lean
tautschnig: Thanks, removed.

Strata/Languages/Core/StatementSemanticsProps.lean
tautschnig: Done.

PR #457, opened 2026-02-19
Support Any type, Operations, and Kwargs handling for Python->Laurel

thanhnguyen-aws:
*Description of changes:* 
This PR Any type, Operations, and Kwargs handling for Python->Laurel:
1. Define type Any and ListAny in PythonLaurelCorePrelude as inductive types.
2. Change the encoded type for all variables, functions' inputs and outputs to Any type.
3. Encode Python's operations in PythonLaurelCorePrelude so that they are consistent with Python's semantic.
4.  Support Kwargs handling for Python->Laurel

### DIFF
--- a/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
+++ b/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
@@ -212,15 +212,15 @@ def translateStmt (fieldNames : List Identifier) (funcNames : FunctionNames) (en
             -- Translate as: var name; call name := callee(args)
             let boogieArgs := args.map (translateExpr fieldNames env)
             let defaultExpr := defaultExprForType ty
-            let initStmt := Core.Statement.init ident boogieType (some defaultExpr)
-            let callStmt := Core.Statement.call [ident] callee boogieArgs
+            let initStmt := Core.Statement.init ident boogieType (some defaultExpr) md
+            let callStmt := Core.Statement.call [ident] callee boogieArgs md
             (env', [initStmt, callStmt])
       | some initExpr =>
           let boogieExpr := translateExpr fieldNames env initExpr
-          (env', [Core.Statement.init ident boogieType (some boogieExpr)])
+          (env', [Core.Statement.init ident boogieType (some boogieExpr) md])
       | none =>
           let defaultExpr := defaultExprForType ty
-          (env', [Core.Statement.init ident boogieType (some defaultExpr)])
+          (env', [Core.Statement.init ident boogieType (some defaultExpr) md])
   | .Assign targets value =>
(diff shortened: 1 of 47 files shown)


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -465,7 +465,7 @@ def translateDatatypeDefinition (dt : DatatypeDefinition) : Core.Decl :=
 /--
 Translate Laurel Program to Core Program
 -/
-def translate (program : Program) : Except (Array DiagnosticModel) (Core.Program × Array DiagnosticModel) := do
+def translate (program : Program) (preludeFunctionNames: List Identifier) : Except (Array DiagnosticModel) (Core.Program × Array DiagnosticModel) := do
```
keyboardDrummer, line 464: Hopefully we can get rid of `(preludeFunctionNames: List Identifier)`again when we change the Core prelude for Python into a Laurel prelude for Python.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,934 @@
+/-
```
keyboardDrummer, line 1: How does this file relate to Strata/Languages/Python/CorePrelude.lean ? Why do we need two versions?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -55,15 +55,33 @@ inductive UnmodeledFunctionBehavior where
...
+  name : String
+  --args include name, type, default value
+  args : List (String × String × Option (Python.expr SourceRange))
+  has_kwargs: Bool
+  ret : Option String
+deriving Repr, Inhabited
+
 structure TranslationContext where
+  /-Name of the current function that is being transalated-/
+  current_function : String
```
keyboardDrummer, line 69: The naming style seems to inconsistently use snake_case and camelCase

keyboardDrummer:
> This PR add support Kwargs handling for Python->Laurel

Is that all this PR changes? It looks like it changes much more.

Where are the test additions for this PR ?

Strata/Languages/Python/PythonLaurelCorePrelude.lean
thanhnguyen-aws: When I change the encoded type for all variables, functions' inputs and outputs to Any type, some procedures in Strata/Languages/Python/CorePrelude.lean needs to be modified. However, if I modify them, it would cause type checking errors with the PythonToCore translation. I think we should temporarily create a separate prelude for Python to Laurel to avoid those conflicts.

PR #458, opened 2026-02-20
Refactor Factory_wf to use less resource

aqjune-aws:
Confirmed that `lake build` works with `ulimit -S 10240`.
It could not still pass `ulimit -S 8192`, however.

The high-level idea of the optimization was to
- Specify which subexpression to simplify when using `simp`
- Avoid `simp` and `grind` when possible
- And interestingly, when the goal is `x \in [a, b, c, ...]`, split subgoals of `x = a`/`x = b`/.... in advance and apply the following proofs in parallel using `all_goals`.


joehendrix (approved):
This looks like a good result to fix an immediate issue.

However, I think the approach taken by the prior work is  brittle, and we should work on bundling the well-formed proofs at the individual LFunc level.  

This could like defined a struct WFLFunc that is a "LFunc + LFuncWF", and a WFLFactory that is an array of WFLFunc. 

Since the LFunc's are produced using combinators generally, we'd have versions of those combinators that produce WFLFuncs.  Proving would be local and incremental.

andrewmwells-amazon (approved):

PR #459, opened 2026-02-20
Support inheritance

keyboardDrummer:
### Changes
- Change naming scheme so the name `$heap` is always used as the heap to read/assign to in the body. For read-only procedures, `$heap` is the input parameter as well as the one to read. For read/write procedures, `$heap` is the out parameter and `$heap_in` the input one.
- Add a `Map.const` function to the Core factory
- Add datatypes to Laurel, to support a more incremental translation to Core
- During translation, introduce a `Field` datatype that has a constructor for each user defined field.
- During translation, introduce a `UserType` datatype that has a constructor for each user defined type.
- Support composites that have fields with the same name but different types
- Move the handling of `new` from HeapParameterization to a new pass `TypeHierarchy`, that also handles `<value> as <Type>` and `<value> is <Type>`.

### Testing
- Add a test `sameFieldNameDifferentType` to the mutable fields test file
- Added a test file related to inheritance


Strata/Languages/Core/FactoryWF.lean
```diff
@@ -17,50 +17,63 @@ import Strata.DL.Lambda.IntBoolFactory
...
-            try unfold bvShiftOp
-            try unfold bvBinaryPred
-            intros lf md args res
-            repeat (rcases args with _ | ⟨ args0, args ⟩ <;> try grind))
-        · decide -- LFuncWF.body_or_concreteEval
-        · decide -- LFuncWF.typeArgs_nodup
-        · decide -- LFuncWF.inputs_typevars_in_typeArgs
-        · decide -- LFuncWF.output_typevars_in_typeArgs
+  · intros f Hmem
+    -- 176 is the number of functions in Factory (#eval Factory.size)
```
andrewmwells-amazon: Can we just `def num_funcs_in_fac := Factory.size` and use that here?

andrewmwells-amazon (approved):

Strata/Languages/Core/FactoryWF.lean
keyboardDrummer: I'll try that in a follow-up

keyboardDrummer: Apparently `iterate` needs a literal number as argument 🤷

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -357,21 +360,48 @@ structure ConstrainedType where
...
+    to already translate to datatypes.
+     -/
+structure DatatypeDefinition where
+  name : Identifier
+  typeArgs : List Identifier
+  constructors : List DatatypeConstructor
+
 /--
-A user-defined type, either a composite type or a constrained type.
+A user-defined type, either a composite type, a constrained type, or an algebriac datatype.
```
joscoh: Nit: algebraic

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -354,30 +360,25 @@ def translateProcedure (constants : List Constant) (funcNames : FunctionNames) (
...
+  let initEnv : TypeEnv := proc.inputs.map (fun p => (p.name, p.type))
+  let body := match proc.body with
+    | .Transparent bodyExpr => some (translateExpr fieldNames initEnv bodyExpr)
+    | _ => none
+  .func {
+    name := Core.CoreIdent.unres proc.name
+    typeArgs := []
+    inputs := inputs
+    output := outputTy
+    body := body
```
joscoh, line 377: Not sure if you need this, but if you want this function to be inlined for the partial evaluator, you would have to give it the attribute `inline`.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -392,7 +393,35 @@ def isPureExpr(expr: StmtExprMd): Bool :=
...
+  | .Assign .. => false
+  | .FieldSelect .. => true
+  | .PureFieldUpdate .. => true
+  -- Instance related
+  | .This => panic s!"isPureExpr not implemented for This"
+  | .AsType .. => panic s!"isPureExpr not supported for AsType"
+  | .IsType .. => panic s!"isPureExpr not supported for IsType"
+  | .InstanceCall .. => panic s!"isPureExpr not implemented for InstanceCall"
+  -- Verification specific
+  | .Forall .. => panic s!"isPureExpr not implemented for Forall"
```
joscoh, line 412: I am confused by some of these. How would `.Forall` not be a pure expression, for instance?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -413,48 +442,68 @@ def canBeBoogieFunction (proc : Procedure) : Bool :=
...
-    | .Transparent bodyExpr => some (translateExpr constants initEnv bodyExpr)
-    | _ => none
-  .func {
-    name := Core.CoreIdent.unres proc.name
-    typeArgs := []
-    inputs := inputs
-    output := outputTy
-    body := body
-  }
+def translateDatatypeDefinition (dt : DatatypeDefinition) : Core.Decl :=
```
joscoh, line 448: Core datatypes have additional restrictions (e.g. strict positivity, uniformity, etc). Right now the translation could fail to pass the Core typechecker (not necessarily an issue for now, just pointing it out).

Strata/Languages/Laurel/TypeHierarchy.lean
```diff
@@ -0,0 +1,322 @@
...
+
+namespace Strata.Laurel
+
+open Strata
+
+/--
+Compute the flattened set of ancestors for a composite type, including itself.
+Traverses the `extending` list transitively.
+-/
+def computeAncestors (types : List TypeDefinition) (name : Identifier) : List Identifier :=
```
joscoh: Wondering if this can be combined with `canReachField` and `findFieldOwner` which basically do the same thing (traverse the inheritance tree). Maybe in a future PR.

Strata/Languages/Laurel/TypeHierarchy.lean
```diff
@@ -0,0 +1,322 @@
...
+            ct.extending.foldl (fun acc2 parent => acc2 ++ go fuel' parent) acc
+          else acc
+        | _ => acc) [])
+  (go types.length name).eraseDups
+
+private def mkMd (e : StmtExpr) : StmtExprMd := ⟨e, #[]⟩
+
+/--
+Generate Laurel constant definitions for the type hierarchy:
+- A `ancestorsFor<Type>` constant per composite type with the inner ancestor map
```
joscoh: Can you give an example of what this looks like or explain what these is used for?

Strata/Languages/Laurel/TypeHierarchy.lean
```diff
@@ -0,0 +1,322 @@
...
+  | .Return (some v) => collectDiamondFieldErrors uri types env v
+  | _ => []
+  termination_by sizeOf expr
+  decreasing_by all_goals (have := WithMetadata.sizeOf_val_lt expr; term_by_mem)
+
+/--
+Validate a Laurel program for diamond-inherited field accesses.
+Returns an array of DiagnosticModel errors.
+-/
+def validateDiamondFieldAccesses (uri : Uri) (program : Program) : Array DiagnosticModel :=
```
joscoh, line 185: Is this just the `Program` version of `collectDiamondFieldErrors`? If so, I would give it a name that makes that clearer (I was unclear what "collect" vs "validate" meant here).

StrataTest/Languages/Laurel/Examples/Objects/T1_MutableFields.lean
```diff
@@ -60,14 +60,27 @@ procedure implicitEquality(c: Container, d: Container) {
   if (c#intValue == d#intValue) {
     assert c == d;
   } else {
-    assert c != d;
+    // Somehow we can't prove this here
```
joscoh, line 63: Do you know why this was passing before and fails now?

StrataTest/Languages/Laurel/Examples/Objects/T5_inheritance.lean
```diff
@@ -0,0 +1,95 @@
...
+  assert b#bValue == 3;
+
+  assert b is Left;
+  assert b is Right;
+  assert b is Top;
+  assert b is Bottom;
+}
+
+// Currently does not pass. Implementation needs b type invariant mechanism that we have yet to add.
+//procedure typedParameter(b: Bottom) {
```
joscoh, line 85: What makes this procedure different from the previous one?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Okay. I think I'd prefer no inlining by default here.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: > How would .Forall not be a pure expression, for instance?

If it contains an assignment in its body, or calls a procedure that's not pure.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Thanks. Good callout.

StrataTest/Languages/Laurel/Examples/Objects/T5_inheritance.lean
keyboardDrummer: `b: Bottom` is never assigned, so knowledge about the type tag of `b` should come from the constraints on the type `Bottom`, but that knowledge is not added by the translation yet.

Strata/Languages/Laurel/TypeHierarchy.lean
keyboardDrummer: Good callout. What relates to this is that we don't have a pass that resolves references in Laurel, after which those resolved references could be used. Having that would greatly simplify `canReachField` and `findFieldOwner` and probably fix several latent bugs related to naming conflicts.

What I'm not sure about is how we should store resolution information. One option would be to give all the references/definitions in the AST a unique `id: int`, and outside of that have maps `definitions: Map Id AstNode` and `references: Map Id (Set AstNode)`. Feels a bit unfortunate that we'd have to generate these ids but it also seems there's no way create cyclic data structures directly.

Strata/Languages/Laurel/TypeHierarchy.lean
keyboardDrummer: Added more explanation

StrataTest/Languages/Laurel/Examples/Objects/T1_MutableFields.lean
keyboardDrummer: It flipped before and I have no idea why. Not sure how to investigate either.

joscoh: Hmm, would be interesting to look at the Core output to compare

PR #463, opened 2026-02-20
Fix #460, ensure type variables unify in DDM

### LINKED ISSUE #460: Type checking issue of Polymorphic datatype's destructor
Let's consider this code: 
```Lean
def smalltest :=
#strata
program Core;

datatype Option (a: Type) {Some (unwrap: a), None ()};

function somefunc (s: string) : int ;

function OptString_len (o: Option(string)) : string {
  somefunc(Option..unwrap(o))
}

#end

```

The type-checker complains at `Option..unwrap(o)` :  `Encountered a expression when string expected`.

joscoh:
*Issue #, if available:* #460

*Description of changes:* DDM `tvars` should not attempt to unify in the DDM directly, since all polymorphic type inference is done in Lambda for now. This PR fixes a bug where the DDM could not unify a `tvar` (which will later be instantiated as `string`) with the correct type. 

### DIFF
--- a/StrataTest/Languages/Core/PolyUnifTest.lean
+++ b/StrataTest/Languages/Core/PolyUnifTest.lean
@@ -60,7 +60,7 @@ function BadFunc (o: Option(int)) : int {
 #end
 
 /--
-info: error: (1284-1357) Impossible to unify (arrow string int) with (arrow int $__ty4).
+info: error: (1277-1350) Impossible to unify (arrow string int) with (arrow int $__ty4).
 First mismatch: string with int.
 -/
 #guard_msgs in


StrataTest/Languages/Core/PolyUnifTest.lean
```diff
@@ -0,0 +1,69 @@
...
+/-- info: ok: datatype Option (a : Type) {(
+  (Some(unwrap : a))),
+  (None())
+};
+function somefunc (s : string) : int;
+function OptString_len (o : (Option string)) : int {
+  somefunc(Option..unwrap(o))
+}-/
+#guard_msgs in
+#eval Core.typeCheck Options.quiet (TransM.run Inhabited.default (translateProgram polyUnifPgm)).fst
```
atomb: I've got a PR in creation that would lead this to fail to typecheck as-is, but it'll work without change if you adjust as follows:
```suggestion
#eval Core.typeCheck .quiet (TransM.run Inhabited.default (translateProgram polyUnifPgm)).fst
```

StrataTest/Languages/Core/PolyUnifTest.lean
```diff
@@ -0,0 +1,69 @@
...
+}
+
+#end
+
+/--
+info: error: (1284-1357) Impossible to unify (arrow string int) with (arrow int $__ty4).
+First mismatch: string with int.
+-/
+#guard_msgs in
+#eval Core.typeCheck Options.quiet (TransM.run Inhabited.default (translateProgram polyUnifBadPgm)).fst
```
atomb: Same change here:
```suggestion
#eval Core.typeCheck .quiet (TransM.run Inhabited.default (translateProgram polyUnifBadPgm)).fst
```

atomb (dismissed):

PR #464, opened 2026-02-20
Refactor to implement some I/O in SimpleAPI

atomb:
Perform some refactoring toward using the `SimpleAPI` interface from CLI implementations. This currently includes:
* Moving some definitions from `StrataMain.lean` to `Strata/Util/IO.lean`
* Refactoring several of those definitions to use `throw` for errors rather than exiting immediately
* Importing `Strata.Util.IO` in `SimpleAPI` and using some of the migrated definitions
* Calling `SimpleAPI` functions from `StrataMain`
* Renaming the `Options` type used by Core.verify to `VerifyOptions`

Although it covers a lot of files, the changes are largely trivial.


Strata/SimpleAPI.lean
```diff
@@ -27,7 +30,7 @@ possible. It is intended for use cases that are essentially equivalent to more
 fine-grained or more structured equivalents of what the `strata` CLI can
 currently do.
 
-**Note:** All definitions are opaque for the moment, so that we can discuss the
+**Note:** Some definitions are opaque for the moment, so that we can discuss the
 structure. Most can be implemented straightforwardly by calling existing code.
 Those that can't are noted explicitly.
```
joehendrix: Nit: the current wording is a bit ambiguous about what "opaque" means here. Suggested rewrite:

```suggestion
**Note:** This is a proposed API that is only partially implemented. Functions
that have been implemented are marked `def`. Proposed but unimplemented functions
are declared using `opaque` to define the intended API surface; these should not
be invoked.
```

Also worth considering: the `opaque` declarations silently return `Inhabited` defaults when evaluated at runtime. Marking them `noncomputable opaque` instead would prevent code generation entirely, so any attempt to compile code that calls an unimplemented function fails at build time rather than silently succeeding.

StrataMain.lean
```diff
@@ -200,15 +113,15 @@ def printCommand : Command where
...
+    let (ld, pd) ← Strata.readStrataFile searchPath v[0]
     match pd with
     | .dialect d =>
       let .isTrue mem := inferInstanceAs (Decidable (d.name ∈ ld.dialects))
         | IO.eprintln s!"Internal error reading file."
           return
       IO.print <| ld.dialects.format d.name mem
     | .program pgm =>
-      IO.print <| toString pgm
+      IO.print <| Strata.writeStrataText pgm
```
joehendrix: Bug: `Strata.writeStrataText` returns `ByteArray`, so `IO.print` writes out an array of bytes instead of the program text (e.g., `IO.print <| String.toByteArray "abc"` prints `[97, 98, 99]`). This breaks `strata print` for programs.

Suggest reverting to the old code:

```suggestion
      IO.print <| toString pgm
```

Strata/Util/IO.lean
```diff
@@ -28,5 +38,106 @@ def readBinInputSource (path : String) : IO ByteArray := do
...
+def readStrataText (fm : Strata.DialectFileMap) (path : System.FilePath) (bytes : ByteArray)
+    : IO (Strata.Elab.LoadedDialects × DialectOrProgram) := do
+  let leanEnv ← Lean.mkEmptyEnvironment 0
+  let contents ← match bytesToText path bytes with
+    | Except.ok c => pure c
+    | Except.error msg => throw (IO.userError (fileReadErrorMsg path msg))
+  let inputContext := Strata.Parser.stringInputContext path contents
+  let (header, errors, startPos) := Strata.Elab.elabHeader leanEnv inputContext
+  if errors.size > 0 then
+    IO.eprintln (← mkErrorReport path errors)
```
joehendrix: Header errors are printed to stderr but parsing continues. The old code in `StrataMain.lean` called `exitFailure` here. For a library function, hard-exiting is wrong, but silently continuing could produce confusing downstream failures. Consider throwing `IO.userError` for consistency with the other error paths in this function.

StrataMain.lean
```diff
@@ -186,12 +99,12 @@ def toIonCommand : Command where
...
   callback := fun v pflags => do
     let searchPath ← pflags.buildDialectFileMap
-    let (_, pd) ← readFile searchPath v[0]
+    let (_, pd) ← Strata.readStrataFile searchPath v[0]
     match pd with
     | .dialect d =>
       IO.FS.writeBinFile v[1] d.toIon
     | .program pgm =>
-      IO.FS.writeBinFile v[1] pgm.toIon
+      IO.FS.writeBinFile v[1] (Strata.writeStrataIon pgm)
```
joehendrix: I prefer the old code here.

joehendrix (changes requested):
I added a few comments, but the printCommand change is the only one I consider blocking.

StrataMain.lean
atomb: I was torn about this when I made the change, and I think I agree in the end. I've reverted it.

Strata/Util/IO.lean
atomb: Oh, good catch! Fixed.

Strata/SimpleAPI.lean
```diff
@@ -53,27 +59,19 @@ namespace Strata
...
-
 /--
 Parse a Strata dialect or program in textual format, possibly loading other
 dialects as needed along the way. The `DialectFileMap` indicates where to find
 the definitions of other dialects. The `FilePath` indicates the name of the file
 to be parsed. And the `ByteArray` includes the contents of that file. TODO:
 should it take just a file name and read it directly?
 -/
-opaque readStrataText :
+def readStrataText :
```
andrewmwells-amazon, line 69: Confused why you removed `opaque` here?

atomb: Because it now has a definition. It's just a direct call to another definition, so it's easy to miss. :)

StrataMain.lean
```diff
@@ -786,24 +699,27 @@ private def parseArgs (cmdName : String)
...
+          cmd.flags.foldl (init := {}) fun m f => m.insert f.name f
+        -- Split raw args into positional arguments and parsed flags.
+        let (args, pflags) ← parseArgs cmd.name flagMap #[] {} args
+        if p : args.size = cmd.args.length then
+          cmd.callback ⟨args, p⟩ pflags
+        else
+          exitCmdFailure cmd.name s!"{cmd.name} expects {cmd.args.length} argument(s)."
+    | [] => do
     exitFailure "Expected subcommand."
+  catch e =>
```
joehendrix, line 724: exitFailure above should be indented 2 more spaces.

PR #465, opened 2026-02-20
Correct-by-construction WF proofs for factory functions

joehendrix:
Introduces well-formed Lambda functions (`WFLFunc`) as an abstraction that bundles each factory function with its well-formedness proof at construction time, simplifying the `Factory_wf` proof in `FactoryWF.lean` which triggered stack overflows as the factory grew.

## Details

- Each factory function now carries its own well-formedness proof, so the top-level `Factory_wf` proof no longer needs to unfold every function individually. The proof shrinks from ~40 lines to 8 generic lines.
- A new `LambdaLeanType` typeclass lets eval combinators infer types automatically, reducing most function definitions from ~6 lines to a single line.
- Functions without concrete evaluation use `Uneval` combinators (`polyUneval`, `nullaryUneval`, `unaryFuncUneval`, `binaryFuncUneval`, `ternaryFuncUneval`) that handle common arities and delegate to `polyUneval` for the well-formedness proof.
- Guarded and unguarded binary operations are unified into a single `binaryOp` combinator via an optional guard parameter.
- `binaryOp` also accepts optional preconditions, used by `intSafeDivFunc`/`intSafeModFunc` to express the `y ≠ 0` guard as a proof obligation.
- The three parallel BV operation lists that had to be kept in sync are replaced by a single `BVOpSpec` array that serves as the source of truth.
- The section variable constraint changes from `Coe String T.Identifier` to `Inhabited T.IDMeta`, which is needed for the argument-uniqueness proofs in the new combinators.
- `Preconditions.lean` gains `[Inhabited T.IDMeta]` (needed since `boolImpliesFunc`/`boolNotFunc` are now `WFLFunc`s) and drops unused `[BEq T.IDMeta]`.

### DIFF
--- a/Strata/DL/Lambda/IntBoolFactory.lean
+++ b/Strata/DL/Lambda/IntBoolFactory.lean
@@ -5,6 +5,7 @@
 -/
 
 import Strata.DL.Lambda.LState
+import Strata.DL.Lambda.FactoryWF
 
 /-! ## A Minimal Factory with Support for Unbounded Integer and Boolean Operations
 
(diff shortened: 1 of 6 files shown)


Strata/DL/Lambda/IntBoolFactory.lean
```diff
@@ -19,180 +20,261 @@ open LExpr LTy
...
+    (hInTy2 : inTy2.freeVars = [] := by decide)
+    (hOutTy : outTy.freeVars = [] := by decide) : WFLFunc T :=
+  polyUneval n [] [("x", inTy1), ("y", inTy2)] outTy
+    (h_inputs := by
+      intro ty hty; simp [ListMap.values] at hty
+      rcases hty with rfl | rfl <;> simp [*])
+
+/-- Binary unevaluated function with same-type inputs and mixed output type. -/
+@[inline]
+def binaryMixedUneval (n : T.Identifier) (inTy outTy : LMonoTy)
```
joscoh: Wondering why we need `binaryMixedUneval` and `binaryOpUneval` instead of just using `binaryHetUneval`. Is it purely for convenience?

Strata/DL/Lambda/IntBoolFactory.lean
```diff
@@ -19,180 +20,261 @@ open LExpr LTy
...
+    (op : InValTy → OutValTy)
+    (hInTy : inTy.freeVars = [] := by decide)
+    (hOutTy : outTy.freeVars = [] := by decide) : WFLFunc T :=
+  let mkConst := LambdaLeanType.mkConst (ValTy := OutValTy) T
+  let cevalInTy := LambdaLeanType.cevalTy (ValTy := InValTy) T
+  ⟨{ name := n,
+     inputs := [("x", inTy)],
+     output := outTy,
+     concreteEval := some (fun md args => match args with
+       | [x] => match cevalInTy x with
```
joscoh, line 145: Now that a `LFunc` has a dependent type anyway, eventually, it would be nice to enforce the number of arguments to the concrete evaluator in the type. But that is for a future PR.

Strata/DL/Lambda/IntBoolFactory.lean
```diff
@@ -19,180 +20,261 @@ open LExpr LTy
...
+def intAddFunc : WFLFunc T :=
+  binaryOp "Int.Add" Int.add
+
+def intSubFunc : WFLFunc T :=
+  binaryOp "Int.Sub" Int.sub
+
+def intMulFunc : WFLFunc T :=
+  binaryOp "Int.Mul" Int.mul
+
+def intDivFunc : WFLFunc T :=
```
joscoh, line 225: This will conflict with #444, which adds a version of division with a precondition.

Strata/Languages/Core/Factory.lean
```diff
@@ -397,20 +295,23 @@ def bv64Extract_31_0_Func  := bvExtractFunc 64 31  0
...
-  @intNegFunc CoreLParams _,
-
-  @intLtFunc CoreLParams _,
-  @intLeFunc CoreLParams _,
-  @intGtFunc CoreLParams _,
-  @intGeFunc CoreLParams _,
+def WFFactory : Lambda.WFLFactory CoreLParams :=
+  -- (T := CoreLParams) annotations needed where Lean cannot infer T
+  -- from array context across combinator shape changes.
+  WFLFactory.ofArray (name_nodup := by native_decide) (#[
```
joscoh, line 310: I think we should avoid `native_decide` if possible, IIUC it adds the whole compiler to the proof checker TCB.

aqjune-aws:
Hi Joe, I like the elegance of the code structure - learning a lot.

One concern is that in the future if someone writes another proof that again takes a long time & memory usage, would there be a generic way to resolve it. Since Strata's executable codebase is pretty large now, proofs that reasons about the codebase will have some amount of opportunities to fail on limited memory use cases.

My two cents is that probably we can rely on a more simpler and somewhat dumb approach - adding a new target that collects expensive proofs. IIUC, the brazil build on AL2 is already not running tests because testing takes a while. Or, we can even use the lake exec build only, but this will omit too many proofs.

Another minor concern is that sometimes executable code around LFunc might be modified (like the precondition patch case), and after the refactoring this would require the code writer to be proficient in the proof language. Maybe this won't be Josh's case, but still possible for new beginners.

Strata/Languages/Core/Factory.lean
joehendrix: I typically agree, but it was necessary for efficiency here, this is a simple uniqueness check, and we are depending on the compiler in any case. 

My preference would be for ofArray to return `Except String (WFLFactory T)` and dynamically check function names are unique.  Then the code here could panic.

However, I decided to keep to the spirit of the original code for this PR.

Strata/DL/Lambda/IntBoolFactory.lean
joehendrix: Yes that's correct.  Those got added as Claude was working through migration and now that's done, I'm wondering whether they are worth the additional names.

I'm going to drop them in favor of fewer convenience functions.

Strata/Languages/Core/Factory.lean
joscoh: Makes sense here, but (this is more of a general comment) it does make me a bit concerned that trusting the compiler is necessary for some proofs to be efficient. I know Strata as a whole relies on the correctness of the compiler, but it wouldn't be great for the soundness story if, for example, proofs about the metatheory of Lambda need to rely on the compiler even without any execution.

Strata/DL/Lambda/IntBoolFactory.lean
joehendrix: Agreed

Strata/Languages/Core/Factory.lean
joehendrix: This larger discussion seems out of scope for a PR review.

Given that you say it makes sense here, I am going to resolve the comment.

PR #466, opened 2026-02-21
Add reachability checks for assert and cover statements

shigoel:
## Summary

Adds an optional **reachability check** to the verification pipeline for `assert` and `cover` statements. When enabled, the verifier emits an extra `(check-sat)` before each proof obligation to test whether the path-condition assumptions are satisfiable. If they are not (the code path is unreachable), the obligation is annotated with a `(❗path unreachable)` suffix — assertions on unreachable paths count as `✅ pass` (vacuously true), covers count as `❌ fail`.

### Key changes

- **Per-statement `@[reachCheck]` annotation** — New `@[reachCheck]` syntax for `assert` and `cover` in the Core grammar, parsed into a `ReachCheck` metadata element.
- **Global `--reach-check` CLI flag** — New `reachCheck` field in `Options` (default: `false`), wired to `--reach-check` in `StrataVerify`.
- **Two-check-sat SMT generation** — `encodeCore` accepts an `md` parameter and, when reachability checking is active, emits a `(check-sat)` after asserting assumptions (before the obligation). Each check-sat is preceded by `set-info` annotations with location metadata and a diagnostic message.
- **`addLocationInfo` moved to `SolverM`** — The function now lives in the `SolverM` monad and takes an explicit `message` parameter, enabling distinct `set-info` annotations for the reachability check (`"Path condition unreachable"`) and the proof check (`"Assertion cannot be proven"`).
- **Solver output parsing** — `solverResult` parses two verdict lines when `reachCheck` is true, storing the reachability result in `VCResult.smtReachResult`.
- **Reachability as diagnosis, not outcome** — Rather than a special `Outcome.unreachable` variant, the normal outcome (`pass`/`fail`) is preserved and `VCResult.isUnreachable` (derived from `smtReachResult == some .unsat`) drives a `(❗path unreachable)` suffix in `ToFormat` and richer diagnostic messages in `toDiagnosticModel`. This works correctly because when the path is unreachable, `path ∧ obligation` is already unsat, so `smt_result` is naturally `.unsat` and `smtResultToOutcome` yields the right outcome without any substitution.
- **PE interaction** — When `reachCheck` is active, the PE short-circuit in `preprocessObligation` is bypassed so the SMT solver can evaluate reachability.
- **Scope** — The global flag applies to all obligations including those generated by transforms (call elimination, procedure inlining).

## Test plan

- [x] Five test cases in `Cover.lean`:
  - Global `reachCheck` with contradictory assumptions (`✅ pass (❗path unreachable)` / `❌ fail (❗path unreachable)`)
  - Mixed reachable/unreachable paths (correct per-path results)
  - Per-statement `@[reachCheck]` annotation (only annotated statements get the check)
  - Diagnostic messages for unreachable outcomes
  - `reachCheck` overrides PE-optimized results

### DIFF
--- a/Strata/Languages/Core/Factory.lean
+++ b/Strata/Languages/Core/Factory.lean
@@ -220,6 +220,21 @@ def reNoneFunc : WFLFunc CoreLParams :=
 def polyOldFunc : WFLFunc CoreLParams :=
   polyUneval "old" ["a"] [("x", mty[%a])] mty[%a]
 
+/- A constant `Map` constructor with type `∀k, v. v → Map k v`.
+   `Map.const(d)` returns a map where every key maps to the value `d`. -/
+def mapConstFunc : WFLFunc CoreLParams :=
+  polyUneval "Map.const" ["k", "v"]
+    [("d", mty[%v])]
+    (mapTy mty[%k] mty[%v])
+    (axioms := [
+      ToCoreIdent esM[∀ (%v): -- %1 d
+          (∀ (%k): -- %0 kk
+            {(((~select : (Map %k %v) → %k → %v)
+                ((~Map.const : %v → (Map %k %v)) %1)) %0)}
+            (((~select : (Map %k %v) → %k → %v)
+                ((~Map.const : %v → (Map %k %v)) %1)) %0) == %1)]
+    ])
+
 /- A `Map` selection function with type `∀k, v. Map k v → k → v`. -/
 def mapSelectFunc : WFLFunc CoreLParams :=
   polyUneval "select" ["k", "v"]
@@ -359,6 +374,7 @@ def WFFactory : Lambda.WFLFactory CoreLParams :=
 
   polyOldFunc,
 
+  mapConstFunc,
   mapSelectFunc,
   mapUpdateFunc,
 
@@ -464,6 +480,7 @@ def reInterOp : Expression.Expr := reInterFunc.opExpr
 def reCompOp : Expression.Expr := reCompFunc.opExpr
 def reNoneOp : Expression.Expr := reNoneFunc.opExpr
 def polyOldOp : Expression.Expr := polyOldFunc.opExpr
+def mapConstOp : Expression.Expr := mapConstFunc.opExpr
 def mapSelectOp : Expression.Expr := mapSelectFunc.opExpr
 def mapUpdateOp : Expression.Expr := mapUpdateFunc.opExpr
 
--- a/Strata/Languages/Laurel/CorePrelude.lean
+++ b/Strata/Languages/Laurel/CorePrelude.lean
@@ -28,9 +28,15 @@ def corePreludeDDM :=
 #strata
 program Core;
 
-// Abstract types for the heap model
+// Field and TypeTag are declared as an opaque type for DDM resolution to pass; the Laurel translator
+// replaces them with datatypes that contain one constructor for each field and composite type
 type Field;
-type Composite := int;
+type TypeTag;
+
+// Composite is a datatype with a reference (int) and a runtime type tag
+datatype Composite () {
+  MkComposite(ref: int, typeTag: TypeTag)
+};
 
 // Tagged union for field values
 datatype Box () {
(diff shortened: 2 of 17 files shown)


Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -116,16 +116,37 @@ def runSolver (solver : String) (args : Array String) : IO IO.Process.Output :=
...
     (E : Strata.SMT.EncoderState) (smtsolver : String)
-    : Except Format (Result P.Ident) := do
+    (reachCheck : Bool := false)
+    : Except Format (Option (Result P.Ident) × Result P.Ident) := do
   let stdout := output.stdout
-  let pos := stdout.find (· == '\n')
-  let verdict := stdout.extract stdout.startPos pos |>.trimAscii
-  let rest := stdout.extract pos stdout.endPos
+  -- When reachCheck is true, the first line of stdout is the reachability
+  -- verdict; strip it and parse it separately.
```
atomb, line 134: Eventually I think it'd be good to write a proper parser for solver output, since we'll want that for things like complex models for satisfiable queries. It could be that we could even use a DDM dialect for that. But I'm just noting that for the future. The code below seems fine until we have that.

Strata/Languages/Core/Verifier.lean
```diff
@@ -116,6 +141,7 @@ inductive Outcome where
   | pass
   | fail
   | unknown
+  | unreachable
```
MikaelMayer: I think I would prefer to keep pass/fail/unknown and put unreachable not as an outcome but as a diagnosis along with the outcome. It's meta-information about the verification condition that is helpful.
Moreover we will have that `unreachable` implies `pass` for assertions and  implies `fail` for covers, so that it's easier for proper error reporting and filtering.

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -96,3 +96,209 @@ Result: ✅ pass
...
+  assert [unreach_assert]: (x > 5);
+  cover [unreach_cover]: (x > 5);
+};
+#end
+
+/--
+info:
+Obligation: unreach_assert
+Property: assert
+Result: ❗ unreachable
```
MikaelMayer: Here I would prefer to see `✅ pass (but because unreachable)`

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -96,3 +96,209 @@ Result: ✅ pass
...
+
+Obligation: unreach_cover
+Property: cover
+Result: ❗ unreachable
```
MikaelMayer: Here I would prefer to see `❌ fail (and unreachable)`

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -96,3 +96,209 @@ Result: ✅ pass
...
+def reachCheckGlobalPgm :=
+#strata
+program Core;
+procedure Test() returns ()
+{
+  var x : int;
+  assume (x >= 0);
+  assume (x < 0);
+
+  assert [unreach_assert]: (x > 5);
```
MikaelMayer, line 116: Shouldn't these assert and cover have the metadata annotation to trigger the unreachable check?

Strata/Languages/Core/Verifier.lean
atomb: Yeah, I also think this would be nicer.

StrataTest/Languages/Core/Examples/Cover.lean
shigoel: This PR provides a global `reachCheck` option, so the check is enabled.

Strata/Languages/Core/Verifier.lean
shigoel: I will change this PR to separate out the `unreachable` reporting. Thanks.

shigoel: PR updated.

PR #289, opened 2025-12-22
GOTO backend: Core-to-GOTO translation, CBMC pipeline tests, and CI

tautschnig:
*Description of changes:*

### Core-to-GOTO translation

Translate Strata Core programs to CProver GOTO binary format for CBMC
verification. Covers all Imperative statement types, Core commands,
procedure contracts, calls, axioms, datatypes, and source locations.

ToCProverGOTO.lean:
- Handle block, ite, loop, exit, funcDecl statements
- Emit loop invariants (#spec_loop_invariant) and measures (#spec_decreases)
- Detect unresolved exit statements (targeting nonexistent labels) and abort
- Extract helpers: emitGoto, emitCondGoto, emitLabel, patchGotoTargets

LambdaToCProverGOTO.lean (new, in Strata/Backends/CBMC/GOTO/):
- Map all arithmetic, comparison, boolean, bitvector, real, string,
  and regex operators to GOTO equivalents
- Signed BV operations (SDiv, SMod, SLt, SLe, SGt, SGe): cast operands
  to signedbv via typecast so CBMC interprets them correctly
- Euclidean integer division/modulo (Int.Div, Int.SafeDiv, Int.Mod,
  Int.SafeMod): encode as compound expressions built from truncating
  div/mod with correction terms
- BV hex encoding for all widths (was 32-bit only)
- Support BV extract, old(expr), quantifiers, ternary, Map.const

CoreToCProverGOTO.lean (new, in Strata/Backends/CBMC/GOTO/):
- End-to-end Core program to GOTO translation
- Call LHS type lookup from program context (not hardcoded)

InstToJson.lean:
- Extend JSON serialization for GOTO programs with function entries
- Deduplicate symbol collection and operator JSON generation

StrataMain.lean:
- Add coreAnalyzeToGoto, laurelAnalyzeToGoto, pyAnalyzeToGoto commands
- Translate procedure calls to FUNCTION_CALL instructions at any nesting
- Lift local funcDecl to top-level GOTO functions
- Emit contracts, axioms, distinct decls, global variables
- Propagate source locations from metadata to GOTO instructions

### Code reorganization

Production GOTO translation code moved from StrataTest/ to
Strata/Backends/CBMC/GOTO/ (LambdaToCProverGOTO.lean,
CoreToCProverGOTO.lean). Test files in StrataTest/ now import from
the production modules and contain only test code.

### Tests and CI

- Unit tests for expression, type, and statement translation
[...]

PR #467, opened 2026-02-23
Fix Statement.renameLhs to compare identifier names

tautschnig:
*Description of changes:*

renameLhs was comparing lhs.name against the full Identifier 'fr' instead of 'fr.name', causing renames to never match when the metadata differed.

### DIFF
--- a/StrataTest/Languages/Core/RenameLhsTests.lean
+++ b/StrataTest/Languages/Core/RenameLhsTests.lean
@@ -1,0 +1,49 @@
+/-
+  Copyright Strata Contributors
+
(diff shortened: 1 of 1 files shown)


tautschnig:
> I'm a bit confused about this PR. The visibility parameter in identifiers exists specifically to make sure we can know that auto-generated names are distinct from other names because they have different visibility. Because of this, visibility is fundamentally part of the name, not metadata. Can you elaborate on what problem the current approach was causing?

Let me reconsider - it seems #289 is now also passing all tests when reverting this change. I'll see whether it comes up again when Laurel evolves in `main`.

PR #471, opened 2026-02-23
Lift calls to imperative procedures that occur in expressions

keyboardDrummer:
### Changes
- Split Laurel procedures into functions and procedures. Only functions can be called from contracts, and functions do not support destructive assignment, exit jumps or loops in their body. Functions are always deterministic. The reason that Laurel only allows calling functions in contracts is because we believe this leads to easier to read contracts.

- Calls to imperative (non-functional) procedures that occur in expressions inside imperative procedures, are lifted to the statement level.

### Caveats
- Currently Laurel functions with postconditions are not supported, because Core functions don't support postconditions yet. Note that when they do support postconditions, such Core functions should behave opaquely, meaning their body won't be visible to callers during verification. This limitation is documented in [T8_PostconditionErrors](https://github.com/strata-org/Strata/pull/471/changes#diff-41333ed0e1ad3cd9ab069068526c97ab12923577680cd41f808be1202010813c)
- Currently Laurel procedures without a postcondition aren't transparent to the caller, because Core procedures don't support being transparent. This limitation is documented in [T5_ProcedureCalls](https://github.com/strata-org/Strata/pull/471/changes#diff-736ee141fc54df0bc99f42ccb7fc0bf8b4116237d49126614342491112192923)

### Testing
- Added new test-cases to `T2_ImpureExpressions.lean`
- Added a new test file `T2_ImpureExpressionsError.lean` that checks whether correct errors are thrown when using imperative code in contracts or functions
- Updated tests T5_ProcedureCalls, T6_Preconditions and T8_PostConditions with additional cases for functions.


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -396,15 +396,9 @@ def isPureExpr(expr: StmtExprMd): Bool :=
...
-/--
-Check if a procedure can be translated as a Core function.
-A procedure can be a function if:
-- It has a transparent body that is a pure expression
-- It has no precondition (or just `true`)
-- It has exactly one output parameter (the return type)
--/
-def canBeBoogieFunction (proc : Procedure) : Bool :=
+/-- Check if a pure-marked procedure can actually be represented as a Core function:
+    transparent body that is a pure expression, no precondition, exactly one output. -/
```
joscoh: Core functions now can have preconditions (PR https://github.com/strata-org/Strata/pull/444 was merged)

joscoh (commented):
Can you add a test (or point me to an existing one) for demonstrating that calling a procedure in a contract gives an error?

Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
```diff
@@ -322,6 +323,8 @@ def parseProcedure (arg : Arg) : TransM Procedure := do
 
   match op.name, op.args with
   | q`Laurel.procedure, #[nameArg, paramArg, returnTypeArg, returnParamsArg,
+      requiresArg, ensuresArg, modifiesArg, bodyArg]
+  | q`Laurel.function, #[nameArg, paramArg, returnTypeArg, returnParamsArg,
       requiresArg, ensuresArg, modifiesArg, bodyArg] =>
```
atomb, line 338: Does it make sense for functions to include `modifiesArg`?

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
```diff
@@ -66,6 +66,37 @@ procedure nestedImpureStatementsAndOpaque()
 //^^^^^^^^^^^^^^ error: assertion does not hold
   assert z == y;
 }
+
+// An imperative procedure call in expression position is lifted before the
+// surrounding expression is evaluated.
+procedure imperativeProc(x: int) returns (r: int)
+   // ensures clause requires because Core does not support transparent proceduces yet
```
atomb:
```suggestion
   // ensures clause required because the Core verification condition generator does not unfold procedure bodies
```

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -396,15 +396,9 @@ def isPureExpr(expr: StmtExprMd): Bool :=
...
+private def canBeCoreFunctionBody (proc : Procedure) : Bool :=
   match proc.body with
   | .Transparent bodyExpr =>
```
atomb, line 489: I'm not sure it needs to be transparent here. That's not blocking, though, because it's fine for this to start out overly conservative. That'd apply to @joscoh's comment, too, I think.

atomb (approved):

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
keyboardDrummer: By unfolding do you mean inlining? I don't think inlining resolves this issue. This should work for recursive procedures as well. The solution is to translate Core procedures to SMT functions.

In any case, if there is an ensures clause, then I don't want the body to be useable by callers in a verification context - bounded or unbounded. Core procedures should have a property `transparent: Bool` that determines whether the body is visible to callers or not.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Updated

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Updated

Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
keyboardDrummer: No it does not, but I prefer not to have separate types or grammar for functions and procedures. The separation between functions and procedures is something that I prefer to keep superficial. 

Assuming we want contracts to be as readable as possible, I think the argument for disallowing statements in contracts is the notion that expressions are easier to read than statements. However, while this holds true for me personally, I feel it is only a widely shared opinion rather than a fact, and we could find programmers that would find certain statement using programs more readable than their expression-only version.

Given the above, I'm open to merging functions and procedures again in the future. The reason I'm separating them right now is that Core does not support statements in contracts, so it's not easy to support, and I don't see enough reason to invest in supporting it right now.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -67,88 +67,169 @@ def isCoreFunction (funcNames : FunctionNames) (name : Identifier) : Bool :=
...
 -/
-def translateExpr (constants : List Constant) (env : TypeEnv) (expr : StmtExprMd)
-    (boundVars : List Identifier := []) : Core.Expression.Expr :=
+def translateExpr (env : TypeEnv) (expr : StmtExprMd)
+    (boundVars : List Identifier := []) (isPureContext : Bool := false)
+    : TranslateM Core.Expression.Expr := do
+  let s ← get
+  let constants := s.constants
+  let funcNames := s.funcNames
+  -- Dummy expression used as placeholder when an error is emitted in pure context
```
joscoh, line 116: I'm confused why we need all this ad-hoc error handling rather than just using the `Except` monad?

keyboardDrummer: So we can return multiple errors at once, which is nicer for the user (and our test)

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
atomb: All of that is specific to deductive verification, though. The purpose of the change I proposed was to make it clear that transparency or opacity isn't a fundamental property of the language. It's an instruction for the verifier.

keyboardDrummer: > All of that is specific to deductive verification, though. 

I think it should not be specific to unbounded verification. I think bounded verification could respect the opaqueness as well. Here's an example:

```
@Verify(Unbounded)
opaque procedure Foo(x: int): int
  ensures x > 0
{
  x
}

@Verify(Unbounded)
procedure UnboundedFooCaller(x: int): int
  ensures x > 0
{
  var x := Foo(3);
  assert x > 0; // pass
  assert x == 3; // fail
}

@Verify(Bounded(
  nonRecursiveInlineDepth := 5, 
  recursiveInlineDepth := 0, 
  ignoreOpaque := false)
procedure BoundedFooCaller() {
  var x := Foo(3);
  assert x > 0; // pass
  assert x == 3; // fail
}
```


> The purpose of the change I proposed was to make it clear that transparency or opacity isn't a fundamental property of the language.

1) It's unclear to me when something is a fundamental property of the language. Are only concepts that effect concrete execution part of that?
2) If I rephrased it to "symbolic verification in Core does not support transparent procedures", would that work for you?

Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
```diff
@@ -9,4 +9,5 @@ import Strata.DDM.Integration.Lean
 
 namespace Strata.Laurel
 
+
```
atomb: Nit: this seems like a spurious whitespace addition.

atomb (approved):

joscoh (dismissed):

Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -20,23 +20,31 @@ namespace Strata.Laurel
...
+      types.findSome? fun td =>
+        match td with
+        | .Composite ct =>
+          if ct.name == current then
+            match ct.fields.findSome? fun f => if f.name == fieldName then some f.type else none with
+            | some ty => some ty
+            | none => ct.extending.findSome? (go fuel')
+          else none
+        | _ => none
+  go types.length typeName
```
MikaelMayer, line 39: Clever hack.

Strata/Languages/Core/Factory.lean
```diff
@@ -221,18 +221,18 @@ def polyOldFunc : WFLFunc CoreLParams :=
   polyUneval "old" ["a"] [("x", mty[%a])] mty[%a]
 
 /- A constant `Map` constructor with type `∀k, v. v → Map k v`.
-   `Map.const(d)` returns a map where every key maps to the value `d`. -/
+   `const(d)` returns a map where every key maps to the value `d`. -/
 def mapConstFunc : WFLFunc CoreLParams :=
-  polyUneval "Map.const" ["k", "v"]
+  polyUneval "const" ["k", "v"]
```
MikaelMayer, line 226: I assume this change is motivated by the ability to use the `esM[]` syntax and not have to specify the entire qualified name, correct?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -68,95 +68,177 @@ def isCoreFunction (funcNames : FunctionNames) (name : Identifier) : Bool :=
...
+            translateExpr env e boundVars isPureContext
+      return .ite () bcond bthen belse
   | .StaticCall name args =>
-      let ident := Core.CoreIdent.unres name
-      let fnOp := .op () ident none
-      args.foldl (fun acc arg => .app () acc (translateExpr fieldNames env arg boundVars)) fnOp
-  | .Block [single] _ => translateExpr fieldNames env single boundVars
+      -- In a pure context, only Core functions (not procedures) are allowed
+      if isPureContext && !isCoreFunction funcNames name then
+        disallowed expr "calls to procedures are not supported in functions or contracts"
```
MikaelMayer, line 191: Could you make the error more actionable?
```suggestion
        disallowed expr "calls to procedures are not supported in functions or contracts - on calls to functions are"
```

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -68,95 +68,177 @@ def isCoreFunction (funcNames : FunctionNames) (name : Identifier) : Bool :=
...
-      LExpr.exist () (some coreTy) coreBody
+      let coreBody ← translateExpr env body (name :: boundVars) isPureContext
+      return LExpr.exist () (some coreTy) coreBody
+  | .Hole => return dummy
+  | .ReferenceEquals e1 e2 =>
+      let re1 ← translateExpr env e1 boundVars isPureContext
+      let re2 ← translateExpr env e2 boundVars isPureContext
+      return .eq () re1 re2
+  | .Assign _ _ =>
+      disallowed expr "destructive assignments are not supported in functions or contracts"
```
MikaelMayer, line 213:
```suggestion
      disallowed expr "destructive assignments are not supported in functions or contracts, you may shadow variable declarations"
```
if appropriate

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -68,95 +68,177 @@ def isCoreFunction (funcNames : FunctionNames) (name : Identifier) : Bool :=
...
+  | .While _ _ _ _ =>
+      disallowed expr "loops are not supported in functions or contracts"
```
MikaelMayer, line 215:
```suggestion
      disallowed expr "loops are not supported in functions or contracts, only recursive functions"
```

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -68,95 +68,177 @@ def isCoreFunction (funcNames : FunctionNames) (name : Identifier) : Bool :=
...
+  | .Exit _ => disallowed expr "exit is not supported in expression position"
```
MikaelMayer, line 216:
```suggestion
  | .Exit _ => disallowed expr "exit is not supported in expression position, but you can lift it before the statement in a conditional statement"
```
How hard would it be to support exit in expression positions?

Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -51,16 +59,19 @@ def computeExprType (env : TypeEnv) (types : List TypeDefinition) (expr : StmtEx
...
   -- Pure field update returns the same type as the target
-  | .PureFieldUpdate target _ _ => computeExprType env types target
-  -- Calls — we don't track return types here, so fall back to TVoid
-  | .StaticCall _ _ => panic "Not supported StaticCall"
+  | .PureFieldUpdate target _ _ => computeExprType env types target procedures
+  -- Calls — look up return type from first output of matching procedure
+  | .StaticCall name _ =>
+      match procedures.find? (·.name == name) with
+      | some proc => proc.outputs.head?.map (·.type) |>.getD ⟨ .TVoid, md ⟩
+      | none => ⟨ .TVoid, md ⟩
```
MikaelMayer, line 74: Shouldn't this be an error?

keyboardDrummer: Yes I think so. I will change this code though because I'm adding a resolution phase, so there will no longer be a lookup here.

atomb (approved):

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Regarding the error messages, these will not be user facing error messages. They still need to be translated on the front-end side to something that's specific to the given language.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: These are language developer facing diagnostics. We still need to specify the user facing ones.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: These are language developer facing diagnostics. We still need to specify the user facing ones.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: I think we can translate any imperative code to functional code, but I'm struggling with where in the pipeline that should happen. It almost seems simplest to remove all imperative code before even reaching Core, and to only use Core functions/expressions. Anyways, it's a design discussion that I've punted for now.

Strata/Languages/Core/Factory.lean
keyboardDrummer: This was to be consistent with the other definitions.

joscoh (approved):

PR #472, opened 2026-02-23
Replace goto with exit statement

MikaelMayer:
Replace `goto` with `exit` statement

The Imperative dialect's `goto` statement only supported forward jumps to labels, which was an awkward fit for structured control flow. This PR replaces it with an `exit` statement that exits the nearest enclosing block with a matching label (or the nearest block if no label is given).

The evaluator skips remaining statements when an exit is active and propagates upward through blocks until a matching block consumes it. The DDM syntax changes from `goto label;` to `exit label;`. The BoogieToStrata tool wraps block sequences in labeled wrappers so that Boogie's forward gotos translate correctly to exit semantics.

Unused helpers (`hasLabel`, `hasLabelInside`) were removed and stale comments updated.

Existing tests pass, test programs updated to exercise the new semantics.

### DIFF
--- a/Strata/Languages/Core/DDMTransform/ASTtoCST.lean
+++ b/Strata/Languages/Core/DDMTransform/ASTtoCST.lean
@@ -854,7 +854,7 @@ partial def stmtToCST {M} [Inhabited M] (s : Core.Statement)
       let labelAnn : Ann String M := ⟨default, l⟩
       pure (.exit_statement default labelAnn)
     | none =>
-      pure (.exit_statement default ⟨default, ""⟩)
+      pure (.exit_unlabeled_statement default)
   | .funcDecl decl _md => funcDeclToStatement decl
 
 partial def blockToCST [Inhabited M] (stmts : List Core.Statement)
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -228,6 +228,7 @@ op call_unit_statement (f : Ident, expr : CommaSepBy Expr) : Statement =>
 op block (c : Seq Statement) : Block => "{\n  " indent(2, c) "}";
 op block_statement (label : Ident, b : Block) : Statement => label ": " b:0 "\n";
 op exit_statement (label : Ident) : Statement => "exit " label ";\n";
+op exit_unlabeled_statement : Statement => "exit;\n";
 
 category SpecElt;
 category Free;
--- a/Strata/Languages/Core/DDMTransform/Translate.lean
+++ b/Strata/Languages/Core/DDMTransform/Translate.lean
@@ -1143,6 +1143,9 @@ partial def translateStmt (p : Program) (bindings : TransBindings) (arg : Arg) :
     let l ← translateIdent String la
     let md ← getOpMetaData op
     return ([.exit (some l) md], bindings)
+  | q`Core.exit_unlabeled_statement, #[] =>
+    let md ← getOpMetaData op
+    return ([.exit none md], bindings)
   | q`Core.funcDecl_statement, #[namea, _typeArgsa, bindingsa, returna, precondsa, bodya, _inlinea] =>
     let name ← translateIdent Core.CoreIdent namea
     let inputs ← translateMonoDeclList bindings bindingsa


Strata/DL/Imperative/Stmt.lean
```diff
@@ -117,30 +116,6 @@ end
 
 ---------------------------------------------------------------------
 
-/--
-`Stmt.hasLabel label s` checks whether statement `s` is a block labeled
-`label`. -/
-def Stmt.hasLabel : String → Stmt P C → Bool
```
MikaelMayer, line 124: We dropped these as they became dead code. They were used to skip statements until a label was found.  Same for the other functions below, they were used in type checking of goto statements.

Strata/Transform/DetToNondetCorrect.lean
```diff
@@ -161,20 +166,28 @@ theorem StmtToNondetCorrect
...
+        cases Heval with | block_sem h => exact ⟨_, h⟩
+      have Hδ : _ = δ := EvalBlock_noFuncDecl_preserves_δ extendEval bss δ _ σ σ' _ Hno Heval_inner
       specialize ih (Block.sizeOf bss) (by simp_all; omega)
       apply (ih _ _).2
-      omega
-      exact Hno
-      rw [← Hδ]; exact Heval
+      · omega
+      · exact Hno
+      · rw [← Hδ]; sorry
```
MikaelMayer: I think this will be irrelevant when we will merge the deterministic tree with the nondeterministic one, so I'm ok with this sorry.

Strata/DL/Imperative/StmtSemanticsSmallStep.lean
```diff
@@ -104,7 +104,7 @@ inductive StepStmt
       (.stmt (.loop g m inv body _) σ δ)
       (.terminal σ δ)
 
-  /- Goto: not implemented, because we plan to remove it. -/
+  /- Exit: not yet implemented in small-step semantics. -/
```
atomb: This definition of the semantics is the one I consider the current, standard semantics, even though the other definition is the one referred to by the call elimination proof. So this is the one I hoped you'd fill in as part of this PR, rather than the other one. Sorry for lack of clarity about that!

Strata/Languages/Core/Core.lean
```diff
@@ -26,8 +26,8 @@ open Strata
 
 3. Strata.Core does not (yet) support polymorphism.
 
-4. Strata.Core does not (yet) support arbitrary gotos. All gotos must
-   currently be to labels later in the program.
+4. Strata.Core supports `exit` statements that exit the nearest enclosing
+   block with a matching label (or the nearest block if no label is given).
```
atomb, line 30:
```suggestion
   block with a matching label (or the nearest block if no label is given).
   Strata does not support arbitrary `goto` statements.
```

Strata/Languages/Core/StatementType.lean
```diff
@@ -154,13 +154,9 @@ where
             -- Add source location to error messages.
             .error (errorWithSourceLoc e md)
 
-        | .goto label md => do try
+        | .exit _label md => do try
```
atomb: I think we should still check that this is a valid label to exit to.

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -17,97 +17,131 @@ namespace Imperative
...
+any block. An `exit l` matches a block labeled `l`. -/
+def ExitStatus.consumedBy : ExitStatus → String → ExitStatus
+  | .normal, _ => .normal
+  | .exiting .none, _ => .normal
+  | .exiting (.some l), blockLabel => if l == blockLabel then .normal else .exiting (.some l)
+
 mutual
 
+/-
+Theorem: exit semantics are well-defined and compositional.
```
joscoh: Is this an actual theorem? If so, can you point to it?

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -17,97 +17,131 @@ namespace Imperative
...
+3. When an exit is active, remaining statements in a block are skipped
+4. A block consumes the exit if the label matches (or exit has no label)
+5. Exit does not modify the store or evaluator
+6. Non-exit statements produce `normal` exit status
+-/
+
 /--
-An inductively-defined operational semantics that depends on
-environment lookup and evaluation functions for expressions.
+An inductively-defined operational semantics with exit support.
```
joscoh: I don't think the docstring should be changed here (the `exit` info can be added, but not in place of the existing description), particularly because this gets fed into the documentation.

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -17,97 +17,131 @@ namespace Imperative
...
-    -- For example, if we require definedness on a block, then we won't be able to evaluate
-    -- a block containing init x; havoc x, because it will require x to exist prior to the block
     isDefinedOver (HasVarsImp.modifiedVars) σ c →
     ----
-    EvalStmt P Cmd EvalCmd extendEval δ σ (Stmt.cmd c) σ' δ
+    EvalStmt P Cmd EvalCmd extendEval δ σ (Stmt.cmd c) σ' δ .normal
 
   | block_sem :
-    EvalBlock P Cmd EvalCmd extendEval δ σ b σ' δ' →
+    EvalBlock P Cmd EvalCmd extendEval δ σ b σ' δ' exit →
```
joscoh: This is minor, but it took me a bit realize this is an arbitrary exit status rather than `.exiting`. IMO renaming to `e` here would be clearer.

StrataTest/Languages/Core/Examples/Exit.lean
joscoh, line 1: Maybe the name of this file should be changed?

joscoh (dismissed):

StrataTest/Languages/Laurel/Examples/Fundamentals/T4_LoopJumps.lean
```diff
@@ -40,7 +40,7 @@ procedure whileWithBreakAndContinue(steps: int, continueSteps: int, exitSteps: i
 -- #eval! testInput "LoopJumps" program processLaurelFile
 
 /-
-Translation towards SMT:
+Translation towards SMT (TODO: update to use exit semantics):
```
atomb: It looks to me like it's already been updated to use exit semantics. Is this an obsolete comment from before you made that change?

Tools/BoogieToStrata/Source/StrataGenerator.cs
atomb, line 1: Could you add an unstructured Boogie program, with only forward gotos, as a test in the `BoogieToStrata` project? The new logic you've added here is complex enough that I'd like more evidence that it's correct.

atomb (changes requested):
This is a very nice improvement! I'd like to see a few more things, though:
* A test of the new functionality in `BoogieToStrata`
* Checking validity of exit labels in the type checker
* Small-step instead of big-step semantics. I think I'd even prefer to leave out the changes to the big-step semantics you made, though I'm sorry for that to have been wasted effort.

atomb:
I think the test failure is occurring because the expected output file for `BoogieToStrata` should contain the output of `StrataVerify`, not the translated Core program. Otherwise I think this is looking great!

Strata/DL/Imperative/StmtSemanticsSmallStep.lean
MikaelMayer: Done. Reverted the big-step changes and implemented exit in the small-step semantics instead. Added `Config.exiting` and `Config.block` to track exit propagation and block label consumption.

It matches what B3 was doing, although we might need to do more checks once we use the semantics to validate the interpreter. But I think it's good like that.

Tools/BoogieToStrata/Source/StrataGenerator.cs
MikaelMayer: Added `ForwardGotos.bpl` with four procedures covering simple forward skip, diamond pattern, chained gotos, and block skipping.

Strata/Languages/Core/StatementType.lean
MikaelMayer: Done. The type checker now threads enclosing block labels and validates that exit targets a label in scope. Unlabeled exit requires being inside at least one block.

Strata/DL/Imperative/StmtSemantics.lean
MikaelMayer: Reverted the big-step exit changes per Aaron's request, so the original docstring is back.

Strata/DL/Imperative/StmtSemantics.lean
MikaelMayer: Reverted the big-step exit changes per Aaron's request, so this is obsolete.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -984,9 +1005,85 @@ private void EmitBigBlock(BigBlock bigBlock) {
         }
     }
 
+    /// <summary>
+    /// Collect all forward goto target labels from a list of big blocks.
+    /// </summary>
+    private static HashSet<string> CollectGotoTargetsFromBigBlocks(IList<BigBlock> bigBlocks) {
```
joscoh: Minor, but this is essentially the same function as `CollectGotoTargets` and they could probably be merged

MikaelMayer: Done

MikaelMayer: https://github.com/strata-org/Strata/pull/472/commits/367f7b4baead891b62282e534469f1e24f61a4d9

atomb (approved):

pr472-review-comments.md
joscoh, line 1: You should not commit this file

PR #473, opened 2026-02-23
Use embedded Python dialect in pySpecs command

joehendrix:
Eliminates the need for pySpecs to locate the Python dialect file on disk by using the compile-time embedded dialect instead.

- pySpecsCommand and SpecsTest write `Strata.Python.Python.toIon` to a temp file via `withTempFile` instead of hardcoding `Tools/Python/dialects/Python.dialect.st.ion`
- Python `Dialect.from_ion()` now parses `syntax` and `metadata` fields in op declarations, since Lean's `Dialect.toIon` includes them


fabiomadge (approved):
Looks good — clean approach to eliminate the hardcoded dialect path.

PR #477, opened 2026-02-23
Refactor the Metadata field of Cmd/Stmt to have a mandatory value

aqjune-aws:
This patch removes the default `.empty` value of the metadata field from Cmd/Stmt/abbreviations of these and related definitions.

A few fixes on top of the refactoring are:
- The semantics of Stmt were assuming that metadata were .empty
- More metadata preservation when ranslating Core's CST to AST
- Procedure inlining and Call elim were also fixed to preserve the metadata
- C_Simp and PythonToCore were updated to preserve the metadata.

### DIFF
--- a/Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
+++ b/Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
@@ -126,7 +126,7 @@ instance : Inhabited Procedure where
     name := ""
     inputs := []
     outputs := []
-    precondition := mkStmtExprMdEmpty <| .LiteralBool true
+    preconditions := []
     determinism := .deterministic none
     decreases := none
     body := .Transparent ⟨.LiteralBool true, #[]⟩
@@ -353,12 +353,14 @@ def parseProcedure (arg : Arg) : TransM Procedure := do
         | .option _ none => pure []
         | _ => TransM.error s!"Expected returnParameters operation, got {repr returnParamsArg}"
       | _ => TransM.error s!"Expected optionalReturnType operation, got {repr returnTypeArg}"
-    -- Parse precondition (requires clause)
-    let precondition ← match requiresArg with
+    -- Parse preconditions (requires clause)
+    let preconditions ← match requiresArg with
       | .option _ (some (.op requiresOp)) => match requiresOp.name, requiresOp.args with
-        | q`Laurel.optionalRequires, #[exprArg] => translateStmtExpr exprArg
+        | q`Laurel.optionalRequires, #[exprArg] => do
+          let precond ← translateStmtExpr exprArg
+          pure [precond]
         | _, _ => TransM.error s!"Expected optionalRequires operation, got {repr requiresOp.name}"
-      | .option _ none => pure (mkStmtExprMdEmpty <| .LiteralBool true)
+      | .option _ none => pure []
       | _ => TransM.error s!"Expected optionalRequires operation, got {repr requiresArg}"
     -- Parse postconditions (ensures clauses - zero or more)
     let postconditions ← translateEnsuresClauses ensuresArg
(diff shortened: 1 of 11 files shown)


StrataTest/DL/Imperative/ArithEval.lean
```diff
@@ -185,11 +185,11 @@ error: none
 warnings: []
 deferred: #[Label: x_value_eq
  Property : assert
- Assumptions: 
+ Assumptions: ⏎
```
MikaelMayer, line 188: Why was a space added after the ":" ?

StrataTest/Languages/C_Simp/Examples/Coprime.lean
```diff
@@ -68,16 +68,16 @@ info: function coprime {
 {
   init (i : int)
   i := a
-  if (~Int.Lt b a) {
+  #[<[fileRange]: :320-351>] if (~Int.Lt b a) {
```
MikaelMayer: Displaying the file range in test expectations is brittle since we pack a lot of tests together. Could you please make it so that the file range is redacted ?

aqjune-aws: Makes sense, let me write a helper function that strips source ranges away & use it. :)

Strata/Languages/Core/ProcedureEval.lean
```diff
@@ -74,7 +74,10 @@ def eval (E : Env) (p : Procedure) : List (Procedure × Env) :=
                 | _ => (.assert label check.expr check.md))
       p.spec.postconditions
   let precond_assumes :=
-    List.map (fun (label, check) => (.assume label check.expr)) p.spec.preconditions
+    List.map (fun (label, check) =>
+      /- the assumptions from preconditions are set to have empty metadata  -/
+      (.assume label check.expr .empty))
```
shigoel: `Procedure.Check` has a metadata field too. Can we copy over that metadata here?

StrataTest/Languages/Core/Examples/ShadowedVars.lean
```diff
@@ -39,7 +39,7 @@ procedure Test() returns ()
 #end
 
 /--
-error: ❌ Type checking error.
+error: (0,(0-0)) ❌ Type checking error.
```
shigoel: Wondering if uninformative metadata can be elided when we're working off Strata programs in Lean files in the `#strata` environemnt. The offending construct isn't on line 0, so it can be confusing.

StrataTest/Languages/C_Simp/Examples/Coprime.lean
joscoh: Should be analogous to `Stmt.stripMetaData`, which was added for exactly the same purpose.

StrataTest/DL/Imperative/ArithEval.lean
aqjune-aws: I dug a rabbit hole initially, but It turns out that it was really that the previous test case was also having spaces at the end of lines. This update simply added `⏎` so editors like vscode don't accidently drop trailing spaces. :)

StrataTest/Languages/C_Simp/Examples/Coprime.lean
aqjune-aws: Fixed.

Strata/Languages/Core/ProcedureEval.lean
aqjune-aws: Missed it, sorry :)

StrataTest/Languages/Core/Examples/ShadowedVars.lean
aqjune-aws: This was happening when the InputContext input to verify was Inhabited.default, which is the default value. Special cased printing in this case.

keyboardDrummer (dismissed):
Sweet, thanks!

PR #478, opened 2026-02-23
Ion roundtrip tests + PreType.fvar removal

joehendrix:
## Summary

Ensures every Ion serialization path is covered by roundtrip tests and
removes dead code that the test investigation revealed.

## Details

- **Roundtrip test coverage for all Ion paths.** Two new test dialects
  (TestIon, TestIonDatatypes) exercise every `ArgF`, `ExprF`, `TypeExprF`,
  `PreType`, `SepFormat`, and metadata variant, so future serialization
  changes will break the build if they introduce regressions. These replace
  the old `DeclareFn`-based test program which only covered a handful of
  paths.
- **`PreType.fvar` and `PreType.ofType` removed.** While building tests
  we discovered `PreType.fvar` was only constructible through
  `PreType.ofType`, which converted `TypeExpr → PreType` solely to
  satisfy `runSyntaxElaborator`'s API. A small refactor to pass type
  information more precisely (`ElabArgKind`) eliminated the need for
  this conversion, making both definitions dead code.
- **Narrower elaboration APIs.** `runSyntaxElaborator` and `unifyTypes`
  now accept only the information they need rather than full `ArgDecl`
  vectors, reducing coupling between elaboration and the AST declaration
  types.
- **SepFormat serialization refactored.** `ArgF.toIon` now pattern-matches
  on `SepFormat` constructors directly instead of dispatching through
  `toIonName` strings, gaining exhaustiveness checking. `fromIonName?`
  moved from `AST.lean` to `Ion.lean` next to its serialization counterpart.
- **Fix missing `.bytes` case in `ArgF.beq`.** This pre-existing bug
  caused byte array arguments to always compare unequal, enabling
  strengthening `testProgramRoundTrip` to check full program equality
  instead of only command count.
- **`Program.create` simp lemmas.** New `Strata.DDM.AST.Lemmas` module
  provides `@[simp]` theorems for `Program.create` field projections,
  enabling safe (proof-carrying) array indexing in tests instead of `!`.

### DIFF
--- a/Strata/DDM/AST.lean
+++ b/Strata/DDM/AST.lean
@@ -435,6 +435,7 @@ private def ArgF.beq {α} [BEq α] (a1 a2 : ArgF α) : Bool :=
   | .num a1 n1, .num a2 n2 => a1 == a2 && n1 == n2
   | .decimal a1 v1, .decimal a2 v2 => a1 == a2 && v1 == v2
   | .strlit a1 i1, .strlit a2 i2 => a1 == a2 && i1 == i2
+  | .bytes a1 b1, .bytes a2 b2 => a1 == a2 && b1 == b2
   | .option a1 o1, .option a2 o2 => a1 == a2 &&
     match o1,o2 with
     | .none, .none => true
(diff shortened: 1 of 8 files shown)


Strata/DDM/AST.lean
```diff
@@ -2237,6 +2210,7 @@ This creates a program.  It is added in addition to `Program.mk` to simplify the
 def create (dialects : DialectMap) (dialect : DialectName) (commands : Array Operation) : Program :=
   { dialects, dialect, commands }
 
+
```
atomb: Tiny nit: did you want to add this newline?

Strata/DDM/AST.lean
```diff
@@ -208,6 +208,7 @@ def fromCategoryName? : QualifiedIdent → Option SepFormat
   | q`Init.NewlineSepBy => some .newline
   | _ => none
 
+
```
joehendrix:
```suggestion
```

Strata/DDM/AST.lean
joehendrix: Nope; will delete since I had a merge conflict too.

atomb (approved):

PR #428, opened 2026-02-16
Add more operators, while statement and quantifiers

keyboardDrummer:
Contains a subset of the changes from https://github.com/strata-org/Strata/pull/385

### Changes
- Fix to Strata/DDM/Parser.lean so it can handle parsing comments `//` in languages that have a division operator
- New operators: `-`, `*`, `/`, `%`, `/t`, `%t` (truncating), `==>`, `!`, unary `-`
- Quantifiers: `forall`, `exists`
- While loops with multiple invariants

### Testing
- Added tests for the new features

PR #481, opened 2026-02-25
Laurel API: precondition → preconditions, postcondition → postconditions

fabiomadge:
Part of the incremental split of #385. Can be merged independently of #426.

### Changes

Converts Laurel's `Procedure` from a single precondition to a list of preconditions, and `Body.Abstract` from a single postcondition to a list. This enables multiple `requires`/`ensures` clauses per procedure, which is needed by the upcoming constrained types and contract features.

**Laurel.lean:**
- `precondition : WithMetadata StmtExpr` → `preconditions : List (WithMetadata StmtExpr)`
- `Body.Abstract postcondition` → `Body.Abstract (postconditions : List ...)`
- `Body.Opaque postcondition` → `Body.Opaque postconditions` (name only, already a list)
- Reorder `Parameter` struct, add `Repr` instances

**Adapted callers:**
- `ConcreteToAbstractTreeTranslator` — parse `Option OptionalRequires` into list
- `LaurelToCoreTranslator` — iterate over preconditions list, generate indexed labels for multiple preconditions
- `HeapParameterization` — fold over preconditions/postconditions lists
- `LaurelFormat` — format preconditions/postconditions lists, relocate `formatDeterminism`
- `PythonToLaurel`, `Specs/ToLaurel` — adapt to new field names

**LiftExpressionAssignments.lean:**
- Process block-in-expression non-last elements left-to-right via `transformStmt` instead of right-to-left via `transformExprDiscarded`
- Remove now-unused `transformExprDiscarded`

### Testing

All existing Laurel, Python, and Core tests pass. No new tests needed — this is a structural refactor with no new features.

### DIFF
--- a/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
+++ b/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
@@ -5,6 +5,8 @@
 -/
 
 -- Laurel dialect definition, loaded from LaurelGrammar.st
+-- NOTE: Changes to LaurelGrammar.st are not automatically tracked by the build system.
+-- Update this file (e.g. this comment) to trigger a recompile after modifying LaurelGrammar.st.
 import Strata.DDM.Integration.Lean
 
 namespace Strata.Laurel


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -316,13 +316,16 @@ def translateProcedure (constants : List Constant) (funcNames : FunctionNames) (
...
                            constants.map (fun c => (c.name, c.type))
-  -- Translate precondition if it's not just LiteralBool true
+  -- Translate preconditions
   let preconditions : ListMap Core.CoreLabel Core.Procedure.Check :=
-    match proc.precondition with
-    | ⟨ .LiteralBool true, _ ⟩ => []
-    | precond =>
+    let (_, result) := proc.preconditions.foldl (fun (i, acc) precond =>
+      match precond with
+      | ⟨ .LiteralBool true, _ ⟩ => (i, acc)
```
keyboardDrummer: Do you think we still need this case, now that we no longer generate it?

Strata/Languages/Laurel/LiftExpressionAssignments.lean
```diff
@@ -261,9 +261,12 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
             match val with
             | .LocalVariable name ty _ => addToEnv name ty
             | _ => pure ()
-          -- Process all-but-last right to left using transformExprDiscarded
-          for nonLastStatement in stmts.dropLast.reverse.attach do
-            transformExprDiscarded nonLastStatement
+          -- Process all-but-last as statements and prepend them in order
```
keyboardDrummer, line 264: Why did you make this change? I'm worried this will adversely affect performance/idiomaticity of the translation, something that we're not testing. Did you consider that it might affect that, and if so, why do you think it does not? 

If you want to keep this change in, could you add a test that compares the generated Core against expected core for a Laurel program that contains heap updating statements in non-last block positions? Then we can easily inspect the effect on the translation. Also besides this PR such a test would be great to have.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
fabiomadge: You're right, removed the dead guard. No caller puts `LiteralBool true` in the preconditions list anymore — empty list is the new "no precondition".

Strata/Languages/Laurel/LiftExpressionAssignments.lean
fabiomadge: This change originated in `ede0ffe9` on `jverify-strata-backend`, where it was a bug fix: `transformExprDiscarded` only handles `Assign` and falls through to `transformExpr` for everything else, but `transformExpr` doesn't know how to process statement constructs like `While`, `Assert`, or `Assume` that can appear in non-last positions of block expressions. After merging #428 (which added while loops and quantifiers to Laurel), blocks containing these constructs would be mishandled. Switching to `transformStmt` fixes this since it properly handles all statement types.

Added `LiftExpressionAssignmentsTest.lean` which compares lifted Laurel against expected output. The `assertInBlockExpr` case demonstrates the bug: with the old `transformExprDiscarded`, it produces `var y: int := $x_0` (incorrect right-to-left snapshot semantics), while `transformStmt` correctly produces `var y: int := x` (sequential statement processing).

keyboardDrummer: Output in the new test looks good, thanks.

Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
```diff
@@ -12,3 +12,4 @@ import Strata.DDM.Integration.Lean
 namespace Strata.Laurel
 
 #load_dialect "./LaurelGrammar.st"
+
```
atomb: Spurious whitespace addition?

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -111,9 +111,9 @@ def analyzeProc (proc : Procedure) : AnalysisResult :=
           { readsHeapDirectly := r1.readsHeapDirectly || r2.readsHeapDirectly,
             writesHeapDirectly := r1.writesHeapDirectly || r2.writesHeapDirectly,
             callees := r1.callees ++ r2.callees }
-    | .Abstract postcond => (collectExprMd postcond).run {} |>.2
-  -- Also analyze precondition
-  let precondResult := (collectExprMd proc.precondition).run {} |>.2
+    | .Abstract postconds => postconds.foldl (fun (acc : AnalysisResult) p => let r := (collectExprMd p).run {} |>.2; { readsHeapDirectly := acc.readsHeapDirectly || r.readsHeapDirectly, writesHeapDirectly := acc.writesHeapDirectly || r.writesHeapDirectly, callees := acc.callees ++ r.callees }) {}
+  -- Also analyze preconditions
```
atomb, line 115: This seems like slightly different logic than before. What's the motivation for that?

Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
fabiomadge: Debugging artifact, removed.

Strata/Languages/Laurel/HeapParameterization.lean
fabiomadge: The logic is the same — adapted from a single value to a list since `precondition` became `preconditions` and `Body.Abstract` now takes a list. Cleaned up the one-liners to use `forM` which matches the original pattern: `(postconds.forM collectExprMd).run {} |>.2`.

atomb (approved):

Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
```diff
@@ -5,11 +5,8 @@
 -/
 
 -- Laurel dialect definition, loaded from LaurelGrammar.st
--- NOTE: Changes to LaurelGrammar.st are not automatically tracked by the build system.
```
andrewmwells-amazon, line 8: Do we want to remove this?

fabiomadge: That was a restore gone wrong — already removed. Merged main again so the diff should be clean now.

atomb (approved):

PR #484, opened 2026-02-26
Allow multiple invariants in Core's loop

shigoel:
## Summary

- **Change `Stmt.loop` invariant from `Option P.Expr` to `List P.Expr`**: Replaces the single optional invariant with a list, enabling multiple loop invariants natively. An empty list `[]` replaces `none` for invariant-free loops. Loop elimination now generates separate labeled `assert`/`assume` per invariant (`entry_invariant_0`, `assume_invariant_0`, etc.) rather than asserting a single conjunct. 

## Misc. changes:

- **Fix bug in `lquantToExpr`**: Fixed a bug in the ASTtoCST lambda-quantifier-to-expression translation, where the scope for bound variables was not properly pushed and popped.

- **Fix spacing for Core's `if` statements and blocks**


Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -661,27 +661,31 @@ partial def lquantToExpr {M} [Inhabited M]
     (qLevel : Nat)
     : ToCSTM M (CoreDDM.Expr M) := do
   let name : Ann String M := ⟨default, mkQuantVarName (qLevel - 1)⟩
+  modify ToCSTContext.pushScope
```
shigoel, line 664: To reviewers: this is the bug fix for rendering the concrete syntax for quantifiers; `pushScope` and `popScope` were inadvertently left out before.

StrataTest/Languages/Core/Examples/Quantifiers.lean
```diff
@@ -78,8 +78,8 @@ Model:
 [DEBUG] Evaluated program:
 procedure Test (x : int) returns (r : int)
 spec {
-  ensures [good]: forall __q0 : int :: exists __q1 : int :: __q1 + (__q1 + __q0) == __q0 + (__q1 + __q1);
-  ensures [bad]: forall __q0 : int :: __q0 < r;
+  ensures [good]: forall __q0 : int :: exists __q1 : int :: r + (__q1 + __q0) == __q0 + (__q1 + r);
```
shigoel, line 81: To reviewers: diff due to bugfix for `lquantToExpr`.

atomb (approved):
I was initially slightly unsure of this, worried that it would add complexity. But I'm really happy to see the number of places where it actually simplifies the logic, along with the benefits for providing narrowly-scoped feedback.

PR #485, opened 2026-02-26
Fix duplicate loop labels when procedure body has multiple loops

shigoel:
## Summary
- `ProcedureEval` was calling `p.body.map Stmt.removeLoops`, which invokes `StateT.run ... 0` independently for each top-level statement, resetting the loop counter to 0 each time. Any two loops in separate top-level blocks received the same `loop_num`, producing clashing labels (`assume_invariant_0`, `not_guard_0`, `invariant_0`, ...).
- Fix by replacing the `map` with a single `Block.removeLoopsM` pass that threads the counter across all statements in the body.
- Within a procedure, the fix ensures the counter threads through all loops in the entire body. Across procedures: the counter resets to 0 for each procedure, but since procedures are analyzed in isolation, their labels need not be globally unique.

- Adds `Examples/TwoLoops.core.st` as a minimal reproducer and hooks it into the existing test harness.

## Testing
- `lake exe StrataVerify Examples/TwoLoops.core.st` passes with unique labels `_0` and `_1`.

### DIFF
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -82,7 +82,7 @@ fn bv64Lit (n : Num) : bv64 => "bv{64}" "(" n ")";
 fn strLit (s : Str) : string => s;
 fn realLit (d : Decimal) : real => d;
 
-fn if (tp : Type, c : bool, t : tp, f : tp) : tp => "if " c:0 " then " t:50 " else " f:50;
+fn if (tp : Type, c : bool, t : tp, f : tp) : tp => "if " c:0 " then " t:0 " else " f:0;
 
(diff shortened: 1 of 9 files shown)


Strata/Languages/Core/ProcedureEval.lean
```diff
@@ -75,7 +75,7 @@ def eval (E : Env) (p : Procedure) : List (Procedure × Env) :=
       p.spec.postconditions
   let precond_assumes :=
     List.map (fun (label, check) => (.assume label check.expr)) p.spec.preconditions
-  let body' : List Statement := p.body.map Stmt.removeLoops
+  let body' : List Statement := (StateT.run (Block.removeLoopsM p.body) 0).fst
```
atomb, line 81: So simple but so critical. :)

atomb (approved):

PR #486, opened 2026-02-26
Support return statements in Laurel functions

keyboardDrummer:
### Changes
Functional:
- Adds a new EliminateReturnsInExpression pass that tries to eliminate return statements.
- Adds not supported message for assert, assume, and local variable declarations without an initializer, that occur in expressions.

Refactoring:
- Introduces a MetaData type abbreviation and refactors the disallowed helper signature
- Renames boogie* variables to core* in translateStmt

Performance:
- Partitions liftImperativeExpressions to skip functional procedures. This does not change the outcome of the pipeline but it might reduce work a bit.

### Testing
- Added cases to T3_ControlFlow
- Added T3_ControlFlowError to clarify that assert and assume are not yet supported in functions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,6 +110,9 @@ jobs:
           ./run_py_analyze.sh laurel
           python run_py_analyze_sarif.py
           python run_py_analyze_sarif.py --laurel
+      - name: Run regex differential tests
+        working-directory: StrataTest/Languages/Python/Regex
+        run: python diff_test.py
 
   lint_checks:
     name: Run lint checks
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -8,6 +8,8 @@ on:
 jobs:
   nightly-release:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@v4
         with:
--- a/DiffTestCore.lean
+++ b/DiffTestCore.lean
@@ -1,0 +1,141 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+-- Differential testing tool: reads (regex, string, mode) triples from stdin,
+-- checks each against Strata's SMT backend, and prints results to stdout.
+--
+-- Usage: DiffTestCore [--log-dir <path>]
+--
+-- Example: printf "a\ta\tfullmatch\n" | DiffTestCore
+--
+-- Input  (stdin):  one test per line, tab-separated: <regex>\t<string>\t<mode>
+-- Output (stdout): one result per line, tab-separated: <regex>\t<string>\t<mode>\t<result>
+--
+-- <result> is one of:
+--   match
+--   noMatch
+--   parseError:patternError
+--   parseError:unimplemented
+--   smtError:<detail>
+--
+-- When --log-dir <path> is given, each generated .core.st program is written to
+-- <path>/<n>_<mode>.core.st (1-indexed) with a header comment showing the
+-- original regex and string, for inspection and debugging.
+--
+-- Restriction: <regex> and <string> fields must not contain tab characters.
+
+import Strata.Languages.Core.Verifier
+import Strata.Languages.Python.Regex.ReToCore
+
+open Strata Strata.Python
+open Core (VerifyOptions)
+
+/-- Escape a string for embedding as a double-quoted Core string literal. -/
+def escapeForCore (s : String) : String :=
+  s.foldl (fun acc c => acc ++ match c with
+    | '\\' => "\\\\"
+    | '"'  => "\\\""
+    | '\n' => "\\n"
+    | '\r' => "\\r"
+    | '\t' => "\\t"
+    | _    => toString c) ""
+
+def parseMode (s : String) : Option MatchMode :=
+  match s with
+  | "match"     => some .match
+  | "fullmatch" => some .fullmatch
+  | "search"    => some .search
+  | _           => none
+
+def modeToStr : MatchMode → String
+  | .match     => "match"
+  | .fullmatch => "fullmatch"
+  | .search    => "search"
+
+inductive StrataResult where
+  | match
+  | noMatch
+  | parseError (kind : String)  -- "patternError" or "unimplemented"
+  | smtError   (msg  : String)
+
+def StrataResult.toStr : StrataResult → String
+  | .match           => "match"
+  | .noMatch         => "noMatch"
+  | .parseError kind => s!"parseError:{kind}"
+  | .smtError   msg  => s!"smtError:{msg}"
+
+/-- Build a Core program that asserts str.in.re(testStr, regexExpr). -/
+def mkProgText (testStr : String) (regexStr : String) : String :=
+  "program Core;\n" ++
+  "procedure main() returns () {\n" ++
+  s!"  assert [match_check]: (str.in.re(\"{escapeForCore testStr}\", {regexStr}));\n" ++
+  "};"
+
+/-- Check whether testStr matches pyRegex (in the given mode) via Strata's SMT backend.
+    If logPath is some path, the generated Core program is written there before checking. -/
+def checkMatch (pyRegex testStr : String) (mode : MatchMode)
+    (logPath : Option String := none) : IO StrataResult := do
+  let (regexExpr, parseErr) := pythonRegexToCore pyRegex mode
+  match parseErr with
+  | some (.patternError _ _ _)  => return .parseError "patternError"
+  | some (.unimplemented _ _ _) => return .parseError "unimplemented"
+  | none =>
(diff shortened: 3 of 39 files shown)


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -208,10 +209,23 @@ def translateExpr (env : TypeEnv) (expr : StmtExprMd)
...
+  | .Exit _ => disallowed expr.md "exit is not supported in expression position"
+
+  | .Block (⟨ .Assert _, md⟩ :: rest) label => do
+    _ ← disallowed md "asserts are not YET supported in functions or contracts"
+    translateExpr env ⟨ StmtExpr.Block rest label, md ⟩ boundVars isPureContext
+  | .Block (⟨ .Assume _, md⟩ :: rest) label =>
+    _ ← disallowed md "assumes are not YET supported in functions or contracts"
+    translateExpr env ⟨ StmtExpr.Block rest label, md ⟩ boundVars isPureContext
+  | .Block (⟨ .LocalVariable name ty (some initializer), md⟩ :: rest) label => do
+    let env' := (name, ty) :: env
```
joehendrix: Nit: `env'` is computed here but never used — lines 226–227 both pass `env`, not `env'`. Should one of them be `env'`?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -208,10 +209,23 @@ def translateExpr (env : TypeEnv) (expr : StmtExprMd)
...
+      disallowed expr.md "loops are not supported in functions or contracts"
+  | .Exit _ => disallowed expr.md "exit is not supported in expression position"
+
+  | .Block (⟨ .Assert _, md⟩ :: rest) label => do
+    _ ← disallowed md "asserts are not YET supported in functions or contracts"
+    translateExpr env ⟨ StmtExpr.Block rest label, md ⟩ boundVars isPureContext
+  | .Block (⟨ .Assume _, md⟩ :: rest) label =>
+    _ ← disallowed md "assumes are not YET supported in functions or contracts"
+    translateExpr env ⟨ StmtExpr.Block rest label, md ⟩ boundVars isPureContext
+  | .Block (⟨ .LocalVariable name ty (some initializer), md⟩ :: rest) label => do
```
joehendrix, line 222: This only matches `(some initializer)`. A `LocalVariable name ty none` (no initializer) in a function block falls through to the catch-all panic at line 236 with `"block expression not yet implemented"`, which is misleading. Consider adding a case that emits a diagnostic like "local variables in functions must have initializers".

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
```diff
@@ -0,0 +1,116 @@
...
+-/
+def stmtsToExpr (stmts : List StmtExprMd) (acc : StmtExprMd) (blockMd : MetaData)
+    : StmtExprMd :=
+  match stmts with
+  | [] => acc
+  | s :: rest =>
+    let acc' := stmtsToExpr rest acc blockMd
+    match s with
+    | ⟨.IfThenElse cond thenBr none, smd⟩ =>
+      ⟨.IfThenElse cond (lastStmtToExpr thenBr) (some acc'), smd⟩
```
fabiomadge, line 61: This only matches single-arm `if` (guard pattern). A full if-else as a non-last statement falls to `| _ =>`, wrapping it into a block with `Return` nodes still inside. The `| .Block _ _ => panic` catch-all in `translateExpr` (LaurelToCoreTranslator.lean line 236) then uses `panic` (not `panic!`), which returns the `Inhabited` default (`boolConst false` / `0`) instead of aborting. The panic message appears in debug trace but no diagnostic error is emitted, so compilation succeeds with wrong code.

Repro:
```
function deadCodeAfterIfElse(x: int) returns (r: int) {
  if (x > 0) { return 1; } else { return 2; }
  return 3;
}
```
Generates `function deadCodeAfterIfElse (x : int) : int { 0 }` — always returns 0.

One possible fix: handle `IfThenElse cond thenBr (some elseBr)` here by dropping the accumulator (it's dead code after an exhaustive if-else with returns in both branches) and producing `IfThenElse cond (lastStmtToExpr thenBr) (some (lastStmtToExpr elseBr))`. Though this assumes both branches return — a more robust approach might be post-transform validation that no `Return` nodes remain.

keyboardDrummer: Thanks for the find! Instead of supporting the repro you gave, I've added an unsupported error message.

And thanks for the note about panic vs panic!

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
```diff
@@ -0,0 +1,116 @@
...
+expression. Each `if-then` (no else) guard wraps as
+`if cond then lastStmtToExpr(body) else acc`; other statements produce
+`Block [stmt, acc]`.
+-/
+def stmtsToExpr (stmts : List StmtExprMd) (acc : StmtExprMd) (blockMd : MetaData)
+    : StmtExprMd :=
+  match stmts with
+  | [] => acc
+  | s :: rest =>
+    let acc' := stmtsToExpr rest acc blockMd
```
joehendrix, line 58: Minor nit.  Consider rewriting to either be tail recursive or a foldr so that you don't risk stack overflows.  The same goes for lastStmtToExpr, but I recognize that's pretty complex to rewrite.

Function bodies can be large.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -208,20 +209,39 @@ def translateExpr (env : TypeEnv) (expr : StmtExprMd)
...
+      let env' := (name, ty) :: env
+      let valueExpr ← translateExpr env initializer boundVars isPureContext
```
fabiomadge: These two `translateExpr` calls compute `valueExpr` and `bodyExpr` that are never used — `disallowed` on line 227 returns a dummy instead. Since `translateExpr` can emit diagnostics as a side effect, this could produce spurious errors (e.g., "calls to procedures not supported") alongside the "local variables not YET supported" error. Consider removing lines 225-226.

keyboardDrummer: Seems good to me. Why do you call those additional errors spurious?

fabiomadge: You're right, they're valid errors. I was thinking about prioritization (the "not YET supported" is the key issue), but fine to keep as-is.

fabiomadge (approved):

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
keyboardDrummer: Good callout, thanks. Any idea how we can avoid recursive functions on ASTs in general? I think I have littered the Laurel->Core translation with them.

keyboardDrummer: I tried to use foldr but this is also mutually recursive with another function. Using foldr complicated that termination proof

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
```diff
@@ -0,0 +1,116 @@
...
+def lastStmtToExpr (stmt : StmtExprMd) : StmtExprMd :=
+  match stmt with
+  | ⟨.Return (some val), _⟩ => val
+  | ⟨.Block stmts _, md⟩ =>
+    match h_last : stmts.getLast? with
+    | some last =>
+      have := List.mem_of_getLast? h_last
+      let lastExpr := lastStmtToExpr last
+      let dropped := stmts.dropLast
+      -- have := List.dropLast_subset stmts
```
shigoel, line 82:
```suggestion
```
Drop?

StrataTest/Languages/Laurel/Examples/Fundamentals/T3_ControlFlow.lean
```diff
@@ -10,10 +10,46 @@ import StrataTest.Languages.Laurel.TestExamples
...
+      return 2;
+    }
+  }
+
+  return 3;
+}
+
+procedure testFunctions() {
+  // assert letsInFunction() == 2;
+  // assert letsInFunction() == 3; error: assertion does not hold
```
shigoel, line 42: What is `letsInFunction`?

keyboardDrummer: It's in the Error file. These two lines should be removed. I'll do it in a follow-up PR.

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
keyboardDrummer: I'll remove it in a follow-up

PR #487, opened 2026-02-26
feat: Implement two-sided verification check with check modes

MikaelMayer:
## Summary

Replaces the single-sided `reachCheck` flag with a two-sided verification framework using orthogonal check mode and check amount flags. Each proof obligation now produces a `VCOutcome` with independent satisfiability and validity properties, enabling richer diagnostic feedback.

## Problem

We want to perform richer checks on assert statements beyond simple validity. Covers are existential checks where forking into two means the results are linked by an OR, so they are not suitable for detecting assertions that surely fail along a path. To find such failures, checks must be encoded as assertions, and we need extended diagnostics for them.

A previous PR opened the way by adding a reachability check, demonstrating that two checks per command are feasible. However, the reachability check missed an important case for bug-finding mode: from reachability + validity alone, we cannot derive the result of reachability + satisfiability. By testing both `P ∧ Q` (satisfiability) and `P ∧ ¬Q` (validity) where `P` is the path condition and `Q` is the property, we get two checks that together determine the validity and satisfiability of `Q` given `P`, and also derive reachability.

## Solution

Two orthogonal flags replace `reachCheck`:

- **Check Mode** (`--check-mode`): `deductive` (default) or `bugFinding`
- **Check Amount** (`--check-amount`): `minimal` (default) or `full`

A per-statement `@[fullCheck]` annotation can override the global check amount.

### Possible outcomes by mode

**Default mode** (`deductive`, `minimal`): validity check only for asserts, satisfiability check only for covers.

For assert statements (validity only, satisfiability masked to unknown):

| Emoji | Label | Meaning |
|-------|-------|---------|
| ✔️ | always true if reachable | Validity passed, property always true if reachable |
| ➖ | can be false and is reachable | Validity failed, solver found a reachable counterexample (with model) |
| ❓ | unknown | Solver could not determine validity |

For cover statements (satisfiability only, validity masked to unknown):

| Emoji | Label | Meaning |
|-------|-------|---------|
| ➕ | can be true and is reachable | Satisfiability passed, property can be true and is reachable from declaration entry |
| ✖️ | refuted if reachable | Satisfiability failed, property always false if reachable |
| ❓ | unknown | Solver could not determine satisfiability |

**Bug-finding mode** (`bugFinding`, `minimal`): satisfiability check only for all statement types. Same as the cover table above.

**Full mode** (`full`): both checks run, all 9 outcomes possible. The last two columns show the error reporting level in SARIF output for each mode (✅ = pass, 🔴 = error, 🟡 = warning, 🔵 = note).

| Emoji | Label | `P ∧ Q` | `P ∧ ¬Q` | Inferred reachability | Meaning | Deductive mode error level | Bug finding mode error level |
|-------|-------|-------------|--------------|----------------------|---------|----------------------------|------------------------------|
| ✅ | always true and is reachable | sat | unsat | ✅ | Property always true, reachable from declaration entry | ✅ | ✅ |
| ❌ | always false and is reachable | unsat | sat | ✅ | Property always false, reachable from declaration entry | 🔴 | 🔴 |
| 🔶 | indecisive and reachable | sat | sat | ✅ | Reachable from declaration entry, solver found models for both the property and its negation | 🔴 | 🔵 |
| ⛔/❌ | unreachable | unsat | unsat | ❌ | Dead code, path unreachable (⛔ warning for assert, ❌ error for cover) | 🟡/🔴 | 🟡/🔴 |
| ➕ | can be true and is reachable | sat | unknown | ✅ | Property can be true and is reachable from declaration entry, validity unknown | 🔴 | 🔵 |
| ✖️ | refuted if reachable | unsat | unknown | ❓ | Property always false if reachable, reachability unknown | 🔴 | 🔴 |
| ➖ | can be false and is reachable | unknown | sat | ✅ | Solver found a model for P ∧ ¬Q: path is reachable from declaration entry and Q can be false, but satisfiability of Q is unknown | 🔴 | 🔵 |
| ✔️ | always true if reachable | unknown | unsat | ❓ | Property always true if reachable, reachability unknown | ✅ | ✅ |
| ❓ | unknown | unknown | unknown | ❓ | Both checks inconclusive | 🔴 | 🔵 |

## Testing

All existing tests updated. New tests cover the full outcome matrix including per-statement `@[fullCheck]` annotations. BoogieToStrata integration tests, Python analysis tests, and SARIF output all updated.

### DIFF
--- a/Strata/DDM/AST.lean
+++ b/Strata/DDM/AST.lean
@@ -1787,7 +1787,7 @@ partial def resolveBindingIndices { argDecls : ArgDecls } (m : DialectMap) (src 
   | .datatype b =>
     -- For datatypes, resolveBindingIndices only returns the datatype type
     -- itself; the constructors and template-generated functions are handled
-    -- separately in addDatatypeBindings!.
+    -- separately in addDatatypeBindings.
     let params : Array String :=
         let addBinding (a : Array String) (_ : SourceRange) {argDecls : _} (b : BindingSpec argDecls) (args : Vector Arg argDecls.size) :=
             match args[b.nameIndex.toLevel] with
@@ -2177,7 +2177,7 @@ FreeVarIndex values are consistent with this order.
 this adds entries for: `Option` (type), `None` (constructor), `Some` (constructor),
 `Option..isNone` (tester), `Option..isSome` (tester).
 -/
-private def addDatatypeBindings!
+private def addDatatypeBindings
     (dialects : DialectMap)
     (gctx : GlobalContext)
     (src : SourceRange)
@@ -2186,7 +2186,7 @@ private def addDatatypeBindings!
     {argDecls : ArgDecls}
     (b : DatatypeBindingSpec argDecls)
     (args : Vector Arg argDecls.size)
-    : GlobalContext :=
+    : Except String GlobalContext := do
 
   let datatypeName :=
     match args[b.nameIndex.toLevel] with
@@ -2204,34 +2204,29 @@ private def addDatatypeBindings!
   -- When preRegistered, the type was already added by preRegisterTypeName;
   -- otherwise it must be fresh.
   let k := GlobalKind.type typeParams.toList none
-  let gctx :=
-    match gctx.defineChecked datatypeName k preRegistered with
-    | .ok gctx => gctx
-    | .error e => panic! s!"addDatatypeBindings!: {e}"
+  let gctx ← gctx.defineChecked datatypeName k preRegistered
   let datatypeIndex := gctx.findIndex? datatypeName |>.getD (gctx.vars.size - 1)
   let datatypeType := mkDatatypeTypeRef src datatypeIndex typeParams
 
   -- Step 2: Add constructor signatures and expand function templates
   let constrArg := args[b.constructorsIndex.toLevel]
-  let constructorInfo :=
-    match extractConstructorInfo dialects constrArg with
-    | .ok info => info
-    | .error e => panic! s!"Constructor extraction error: {e}"
+  let constructorInfo ← extractConstructorInfo dialects constrArg
   -- Errors from template expansion are reported during elaboration
   -- (evalBindingSpec); here we just take the updated context.
   let (gctx, _) := expandFunctionTemplates dialectName src
     datatypeName datatypeType constructorInfo
     b.functionTemplates gctx
-  gctx
+  return gctx
 
 /--
 Pre-register a type name in the `GlobalContext` before the main `addCommand`
 pass. Used by operations annotated with `@[preRegisterTypes]` (e.g., mutual
 blocks) so that forward references between sibling datatypes resolve correctly.
-Names must be fresh — panics if the name is already defined.
+Names must be fresh — returns an error if the name is already defined.
 -/
-private def preRegisterType (dialects : DialectMap) (gctx : GlobalContext) (l : SourceRange)
-    {argDecls} (b : BindingSpec argDecls) (args : Vector Arg argDecls.size) : GlobalContext :=
+private def preRegisterType (dialects : DialectMap) (acc : Except String GlobalContext) (l : SourceRange)
+    {argDecls} (b : BindingSpec argDecls) (args : Vector Arg argDecls.size) : Except String GlobalContext := do
+  let gctx ← acc
   match b with
   | .datatype _ | .type _ =>
     let name :=
@@ -2242,48 +2237,43 @@ private def preRegisterType (dialects : DialectMap) (gctx : GlobalContext) (l : 
     -- Names must be fresh: this is the pre-registration pass.
     | some kind =>
       if h : name ∈ gctx then
-        panic! s!"'{name}' already defined"
+        .error s!"'{name}' already defined"
       else
-        gctx.define name kind h
-    | none => gctx
-  | _ => gctx
+        pure (gctx.define name kind h)
+    | none => pure gctx
+  | _ => pure gctx
 
 private def addBinding (dialects : DialectMap) (dialectName : DialectName) (preRegistered : Bool)
-                       (gctx : GlobalContext) (l : SourceRange) {argDecls} (b : BindingSpec argDecls)
-                       (args : Vector Arg argDecls.size) :=
+                       (acc : Except String GlobalContext) (l : SourceRange) {argDecls} (b : BindingSpec argDecls)
+                       (args : Vector Arg argDecls.size) : Except String GlobalContext := do
+  let gctx ← acc
   match b with
   | .datatype datatypeSpec =>
-    addDatatypeBindings! dialects gctx l dialectName preRegistered datatypeSpec args
+    addDatatypeBindings dialects gctx l dialectName preRegistered datatypeSpec args
   | _ =>
     let name : Var :=
           match args[b.nameIndex.toLevel] with
           | .ident _ e => e
           | a => panic! s!"Expected ident at {b.nameIndex.toLevel} {repr a}"
     match resolveBindingIndices dialects l b args with
-    | some kind =>
-      match gctx.defineChecked name kind preRegistered with
-      | .ok gctx => gctx
-      | .error e => panic! s!"addCommand: {e}"
-    | none => gctx
+    | some kind => gctx.defineChecked name kind preRegistered
+    | none => pure gctx
 
-def addCommand (dialects : DialectMap) (gctx : GlobalContext) (op : Operation) : GlobalContext :=
+def addCommand (dialects : DialectMap) (gctx : GlobalContext) (op : Operation) : Except String GlobalContext := do
     let dialectName := op.name.dialect
     -- Pre-register types if op has @[preRegisterTypes] metadata
-    let (gctx, preRegistered) := Id.run do
-      let .op decl := dialects.decl! op.name
-        | return (panic! "Expected operator declaration", false)
-      let .isTrue h := decideProp (op.args.size = decl.argDecls.size)
-        | return (panic! "Expected arguments to match", false)
+    let .op decl := dialects.decl! op.name
+      | .error "Expected operator declaration"
+    let .isTrue h := decideProp (op.args.size = decl.argDecls.size)
+      | .error "Expected arguments to match"
+    let (gctx, preRegistered) ←
       match decl.metadata.preRegisterTypesLevel decl.argDecls.size with
       | some lvl =>
-        let gctx := foldOverArgAtLevel dialects
-          (preRegisterType dialects) gctx
-          decl.argDecls ⟨op.args, h⟩ lvl
-        (gctx, true)
-      | none =>
-        (gctx, false)
+        (foldOverArgAtLevel dialects (preRegisterType dialects) (.ok gctx)
+          decl.argDecls ⟨op.args, h⟩ lvl).map (·, true)
+      | none => .ok (gctx, false)
     -- Normal fold
-    op.foldBindingSpecs dialects (addBinding dialects dialectName preRegistered) gctx
+    op.foldBindingSpecs dialects (addBinding dialects dialectName preRegistered) (.ok gctx)
 
 end GlobalContext
 
@@ -2297,7 +2287,10 @@ structure Program where
   commands : Array Operation := #[]
   /-- Final global context for program. -/
   globalContext : GlobalContext :=
-    commands.foldl (init := {}) (·.addCommand dialects ·)
+    match commands.foldl (init := (Except.ok {} : Except String GlobalContext))
+        fun acc cmd => acc.bind (·.addCommand dialects cmd) with
+    | .ok gctx => gctx
+    | .error e => panic! s!"Program.globalContext: {e}" -- nopanic:ok
 
 namespace Program
 
@@ -2310,7 +2303,9 @@ instance : Inhabited Program where
 def addCommand (env : Program) (cmd : Operation) : Program :=
   { env with
     commands := env.commands.push cmd,
-    globalContext := env.globalContext.addCommand env.dialects cmd
+    globalContext := match env.globalContext.addCommand env.dialects cmd with
+      | .ok gctx => gctx
+      | .error e => panic! s!"Program.addCommand: {e}" -- nopanic:ok
   }
 
 /--
--- a/Strata/DDM/Elab.lean
+++ b/Strata/DDM/Elab.lean
@@ -75,12 +75,17 @@ private partial def runCommand (leanEnv : Lean.Environment) (commands : Array Op
   if newPos <= iniPos then
     logError { start := iniPos, stop := iniPos } "Syntax error: unrecognized syntax or unexpected token"
     return commands
-  let cmd := tree.info.asOp!.op
+  let opInfo := tree.info.asOp!
+  let cmd := opInfo.op
   let dialects := (← read).loader.dialects
-  modify fun s => { s with
-    globalContext := s.globalContext.addCommand dialects cmd
-  }
-  runCommand leanEnv (commands.push cmd) stopPos
+  let s ← get
+  match s.globalContext.addCommand dialects cmd with
+  | .ok newGctx =>
+    modify fun s => { s with globalContext := newGctx }
+    runCommand leanEnv (commands.push cmd) stopPos
+  | .error e =>
+    logError opInfo.loc e
+    runCommand leanEnv commands stopPos
 
 def elabProgramRest
     (loader : LoadedDialects)
--- a/StrataTest/Languages/Core/Tests/DuplicateDDMBindings.lean
+++ b/StrataTest/Languages/Core/Tests/DuplicateDDMBindings.lean
@@ -1,0 +1,26 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Core.Verifier
+
+---------------------------------------------------------------------
+namespace Strata
+
+-- Duplicate DDM bindings (polymorphic functions, datatypes) are caught during
+-- DDM elaboration and reported as errors (not panics).
+
+/--
+error: 'f1' already defined
+-/
+#guard_msgs in
+def dupPolyFunctions : Program :=
+#strata
+program Core;
+function f1<T1, T2>(x : T1) : Map T1 T2;
+function f1<T1, T2>(x : T1) : Map T1 T2;
+#end
+
+---------------------------------------------------------------------


StrataTest/Languages/Core/SMTEncoderTests.lean
```diff
@@ -187,24 +187,24 @@ spec {
...
 Obligation: UpdateAndRead_ensures_1
 Property: assert
 Result: ✅ pass
 -/
 #guard_msgs in
-#eval verify simpleMapProgram (options := {Options.quiet with useArrayTheory := false})
+#eval verify simpleMapProgram (options := {Options.quiet with useArrayTheory := false, checkMode := .full})
 
--- Test verification with Array theory
+-- Test verification with Array theory - use full mode for complete check
```
MikaelMayer: Resolved in latest commits

Strata/Languages/Core/Verifier.lean
```diff
@@ -106,26 +113,27 @@ def dischargeObligation
   (assumptionTerms : List Term)
   (obligationTerm : Term)
   (ctx : SMT.Context)
-  (reachCheck : Bool := false)
-  : IO (Except Format (Option SMT.Result × SMT.Result × EncoderState)) := do
+  (satisfiabilityCheck validityCheck : Bool)
+  : IO (Except Format (SMT.Result × SMT.Result × EncoderState)) := do
   -- CVC5 requires --incremental for multiple (check-sat) commands
   let baseFlags := getSolverFlags options
+  let needsIncremental := (satisfiabilityCheck && validityCheck) || satisfiabilityCheck || validityCheck
```
MikaelMayer: Resolved in latest commits

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -0,0 +1,225 @@
...
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Core.Verifier
+import Strata.Languages.Core.SarifOutput
+
+/-! ## Tests for VCOutcome
+
+Tests all nine outcome combinations from two-sided verification check,
```
MikaelMayer, line 12: Resolved in latest commits

Strata/Languages/Core/SarifOutput.lean
```diff
@@ -56,17 +90,23 @@ def extractLocation (files : Map Strata.Uri Lean.FileMap) (md : Imperative.MetaD
...
+  | .error msg =>
+    let level := .error
+    let messageText := s!"Verification error: {msg}"
+    let message : Strata.Sarif.Message := { text := messageText }
+    let locations := match extractLocation files vcr.obligation.metadata with
+      | some loc => #[locationToSarif loc]
+      | none => #[]
+    { ruleId, level, message, locations }
+  | .ok outcome =>
+    let level := outcomeToLevel .deductive outcome
```
MikaelMayer: Resolved in latest commits

Strata/Languages/Core/Options.lean
```diff
@@ -18,6 +18,12 @@ def VerboseMode.toNat (v : VerboseMode) : Nat :=
   | .normal => 2
   | .debug => 3
 
+/-- Verification mode for SARIF error level mapping -/
+inductive VerificationMode where
+  | deductive  -- Prove correctness (unknown is error)
+  | bugFinding -- Find bugs (unknown is warning)
```
MikaelMayer: Updated inline comment to clarify that provably false (if reached/reachable) is error, unreachable is warning

Strata/Languages/Core/SarifOutput.lean
```diff
@@ -19,27 +19,57 @@ open Strata.Sarif Strata.SMT
...
+  | .bugFinding =>
+    -- Bug finding: find counterexamples
+    if outcome.passAndReachable || outcome.passReachabilityUnknown then
+      .none
+    else if outcome.alwaysFalseAndReachable || outcome.alwaysFalseReachabilityUnknown then
+      .error
+    else if outcome.indecisiveAndReachable || outcome.canBeFalseAndReachable || outcome.unknown then
+      .warning
+    else
+      .note -- unreachable, satisfiableValidityUnknown
```
MikaelMayer: Updated bugFinding severity levels: unreachable→warning, indecisive/canBeFalse/unknown/satisfiable→note

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -0,0 +1,114 @@
...
+def mkOutcome (satisfiabilityProperty : Result) (validityProperty : Result) : VCOutcome :=
+  { satisfiabilityProperty, validityProperty }
+
+/-! ### Outcome: (sat, unsat) - always true and reachable -/
+
+/--
+info: Emoji: ✅, Label: pass, Predicate: true, Message: Always true and reachable, Deductive: none, BugFinding: none
+-/
+#guard_msgs in
+#eval do
```
MikaelMayer: Added testOutcome helper and 'SARIF output:' prefix

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -0,0 +1,99 @@
...
+namespace Core
+open Strata.SMT
+open Core.Sarif
+open Core.SMT (Result)
+
+def mkOutcome (satisfiabilityProperty : Result) (validityProperty : Result) : VCOutcome :=
+  { satisfiabilityProperty, validityProperty }
+
+def testOutcome (o : VCOutcome) (predicate : VCOutcome → Bool) : IO Unit := do
+  IO.println s!"Emoji: {o.emoji}, Label: {o.label}, Predicate: {predicate o}, Message: {outcomeToMessage o}, SARIF output: Deductive: {outcomeToLevel .deductive o}, BugFinding: {outcomeToLevel .bugFinding o}"
```
MikaelMayer: Added Sat/Val prefix, changed to 'SARIF', added 'level' suffix, removed verbose prefixes

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -0,0 +1,99 @@
...
+
+/-! ### Outcome: (sat, unsat) - always true and reachable -/
+
+/--
+info: Emoji: ✅, Label: pass, Predicate: true, Message: Always true and reachable, SARIF output: Deductive: none, BugFinding: none
```
MikaelMayer: Removed Emoji/Label/Message prefixes, added OutcomePredicate type to test all 9 predicates comprehensively

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -0,0 +1,155 @@
...
+/-! ### Outcome: (sat, sat) - true or false depending on inputs -/
+
+/--
+info:  isSatisfiable isReachable
+Sat:sat|Val:sat 🔶 indecisive, True or false depending on inputs, SARIF: Deductive level: error, BugFinding level: note
+-/
+#guard_msgs in
+#eval testOutcome (mkOutcome (.sat default) (.sat default)) .indecisiveAndReachable
+
+/-! ### Outcome: (unsat, unsat) - unreachable -/
```
MikaelMayer: Removed section comments

keyboardDrummer:
I have trouble parsing the description.

We want to perform richer checks on assert statements beyond simple validity.

> Covers are existential checks where forking into two means the results are linked by an OR, 

Forking into two? What do you mean by that?

> so they are not suitable for detecting assertions that surely fail along a path. 

What does surely fail along a path mean? Do you mean, there exists an input for which the condition does not hold?

Why can't we use assertions as they are? You check the assertion and you get:
- unsat => no bug
- sat => bug
- unknown => possible bug. Improve your proof until SMT returns sat or unsat to determine whether you have a bug or not.

If we're in a situation where the VC is such that the solver never returns sat, then if unknown is returned, solve again using a flipped assertion condition and interpret unsat as there being a bug.

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -186,9 +196,9 @@ def addLocationInfo {P : PureExpr} [BEq P.Ident]
 Writes the proof obligation to file, discharge the obligation using SMT solver,
 and parse the output of the SMT solver.
 
-When `reachCheck` is `true`, the generated SMT file will contain two
-`(check-sat)` commands (one for reachability, one for the proof obligation),
-and the return value includes the reachability decision.
+When two-sided checking is enabled, the generated SMT file will contain two
+`(check-sat-assuming)` commands (one for satisfiability, one for validity),
```
andrewmwells-amazon: I'd just say: "One for `P \and \not Q` and one for `P \and Q`

andrewmwells-amazon:
I think we should copy the explanation table somewhere more permanent than the PR description.

MikaelMayer:
> I have two questions about the table contents:
> 
> Emoji	Label	Sat `P ∧ Q`	Val `P ∧ ¬Q`	Inferred reachability	Meaning	Deductive mode error level	Bug finding mode error level
> ❌	refuted and reachable	unsat	sat	✅	Property always false, reachable from declaration entry	🔴	🔴
> Shouldn't "Property always false" be "Property _can_ be false" ?

On this line, we proved that P ∧ Q is unsat, and P ∧ ¬Q is sat. Hence, P is sat, meaning the path condition makes Q reachable.
That P ∧ Q is unsat means none of these assignments can make Q be true, so Q is proved to always be false. This is a bug for bug finding we will want to report.

> Emoji	Label	Sat `P ∧ Q`	Val `P ∧ ¬Q`	Inferred reachability	Meaning	Deductive mode error level	Bug finding mode error level
> 🔶	indecisive and reachable	sat	sat	✅	Reachable from declaration entry, solver found models for both the property and its negation	🔴	🔵
> How can that happen? Seems impossible to me

Here is the following.
```
procedure m(x: int) {
   assert x > 0
}
```
In this one, we have P == true, Q == (x > 0), ¬Q == !(x > 0).
Both P ∧ Q   on one side and P ∧ ¬Q  are satisfiable.
 Funny, on my side, I thought that (unsat, unsat) was the impossible case two weeks ago until I realized it encoded path condition reachability

> Emoji	Label	Sat `P ∧ Q`	Val `P ∧ ¬Q`	Inferred reachability	Meaning	Deductive mode error level	Bug finding mode error level
> ✖️	refuted if reachable	unsat	unknown	❓	Property always false if reachable, reachability unknown	🔴	🔴
> Given that we don't know whether the check is even reachable, should this really be red for bug finding mode? A check that can't be reached is not a bug.

That's an excellent question. What this case proves is that, if the line is reached, it will _surely_ cause a failure in the assertion. Since we assume in a well-formed codebase that every line is eventually reachable, and that assertions should never fail if reachable, I consider it as a bug, but I'm open to other narratives.

> Emoji	Label	Sat `P ∧ Q`	Val `P ∧ ¬Q`	Inferred reachability	Meaning	Deductive mode error level	Bug finding mode error level
> ➖	can be false if reachable	unknown	sat	✅	Solver found a counterexample reachable from declaration entry, reachability unknown	🔴	🔵
> Why is this not double red? I would think that any `sat` result for `P ∧ ¬Q` proves there is a bug. Also, why does the text say reachability unknown but the icon says it is known? I would think reachability is proven here. Instead of "can be false if reachable" shouldn't this say "refuted and reachable" ?

The problem is modularity. If we prove `sat` for `P ∧ ¬Q` , we have proved that the path condition is reachable from the procedure entry for a particular set of inputs and that with these inputs, Q will be false. This is a problem for deductive verification since it's the only place where we have to prove that Q is always true.
However, for bug finding mode, we cannot guarantee based on the verification condition that P will ever be called with these inputs, perhaps at every call site there are guards that will prevent these inputs.

You are right there was an issue it has to be "can be false and reachable". Excellent finding. Let me fix that.

> I'm rather confused about the meaning of both the `--check-mode` and `--check-amount` flags. IMO the user-facing output should answer these two questions:

Let's continue the conversation to see how I could best rephrase those

> What would your table look like if you asked not `P ^ Q` and `P ^ !Q`, but instead `P ^ !Q` (the first user-facing question) and `P` (the second user-facing question) ? It seems simpler to me to encode the user-facing questions as directly as you can to SMT questions.

If you check `P`  and  `P ^ !Q`, you will not get the satisfiability of  `P ^ Q`. At best, you can know that `P ^ Q` is unsat if `P` is unsat, but if `P` is sat, whatever the result of `P ^ !Q`, you cannot deduce the satisfiability of  `P ^ Q`.
There is redundancy of checking the satisfiability of the two conjuncts you mentioned because if `P ^ !Q` is sat, you automatically get that  `P`  is sat.

Strata/Languages/Core/Verifier.lean
```diff
@@ -44,17 +44,41 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
...
+    if satisfiabilityCheck then
+      Solver.comment "Satisfiability check (can property be true?)"
+      Imperative.SMT.addLocationInfo (P := Core.Expression) (md := md)
+        (message := ("sat-message", s!"\"Property can be satisfied\""))
+      Solver.assert obligationId
+    else if validityCheck then
+      Solver.comment "Validity check (can property be false?)"
+      Imperative.SMT.addLocationInfo (P := Core.Expression) (md := md)
+        (message := ("unsat-message", s!"\"Property is always true\""))
+      Solver.assert obligationId
```
shigoel: Shouldn't this be negated?

StrataTest/Languages/Core/Examples/RemoveIrrelevantAxioms.lean
```diff
@@ -108,128 +108,163 @@ def normalizeModelValues (s : String) : String :=
...
 Property: assert
-Result: 🟡 unknown
+Result: ❓ unknown
 
 Obligation: assert_4
 Property: assert
-Result: ❌ fail
-Model:
-($__x0, model_not_2)
+Result: ❓ unknown
```
shigoel: not sure why the fails became unknowns?

Strata/Languages/Core/Verifier.lean
MikaelMayer: obligationId is already negated at this stage but wait, it might indicate something else when we do bothCheck. Let me double check.

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -123,14 +131,14 @@ procedure Test() returns ()
 info:
 Obligation: unreach_assert
 Property: assert
-Result: ✅ pass (❗path unreachable)
+Result: ⛔ unreachable
 
 Obligation: unreach_cover
 Property: cover
-Result: ❌ fail (❗path unreachable)
+Result: ⛔ unreachable
```
MikaelMayer: This assert and this cover should produce pass and fail respectively, even if they are both reachable.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Ok I refactored the code so that it's not implicitly negated before. That was a bad architectural decision (although it did work) and now it looks better to me and it is correct.

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -290,21 +298,21 @@ procedure Test() returns ()
 info:
 Obligation: pe_assert_pass
 Property: assert
-Result: ✅ pass (❗path unreachable)
+Result: ✅ pass and reachable from declaration entry
```
MikaelMayer: I'm concerned about this. We say that the assertion is suddenly provably reachable when before this PR it was proved to be unreachable. Investigate the test and figure out what is wrong and what is right.

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -252,11 +260,11 @@ procedure Test() returns ()
 #end
 
 /--
-info: #["assertion holds vacuously (path unreachable)", "cover property is unreachable"]
+info: #[]
```
MikaelMayer: It seems that the test stops producing something that was meaningful. Why?

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -35,27 +35,35 @@ procedure Test() returns ()
...
 Obligation: unreachable_cover1
 Property: cover
-Result: ❌ fail
+Result: ✖️ refuted if reachable
+
+Obligation: unreachable_cover1
+Property: cover
+Result: ✖️ refuted if reachable
+
+Obligation: unreachable_cover2
```
MikaelMayer, line 40: Why are the unreachable_cover1 and unreachable_cover2 duplicated in the results?

Strata/Languages/Core/Options.lean
```diff
@@ -43,6 +49,12 @@ instance : DecidableRel (fun a b : VerboseMode => a ≤ b) :=
 /-- Default SMT solver to use -/
 def defaultSolver : String := "cvc5"
 
+/-- Check amount: how much information to gather -/
+inductive CheckAmount where
```
shigoel: smallest of nits/feel free to ignore: can we say `CheckLevel` instead? Amount to me signals that something numeric is involved...

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,112 +161,243 @@ open Std (ToFormat Format format)
...
-    | .pass => "✅ pass"
-    | .fail => "❌ fail"
-    | .unknown => "🟡 unknown"
-    | .implementationError e => s!"🚨 Implementation Error! {e}"
-
-/--
-A model expressed as Core `LExpr` values, suitable for display
-using Core's expression formatter and for future use as program metadata.
--/
-abbrev LExprModel := List (Expression.Ident × LExpr CoreLParams.mono)
```
shigoel, line 158: Did LExprModel go away in this PR, or am I missing something?

Strata/Languages/Core/Options.lean
```diff
@@ -22,6 +22,13 @@ def VerboseMode.toNat (v : VerboseMode) : Nat :=
   | .normal => 2
   | .debug => 3
 
+/-- Verification mode for SARIF error level mapping -/
+inductive VerificationMode where
+  | deductive  -- Prove correctness (unknown is error)
+  | bugFinding -- Find bugs assuming incomplete preconditions (only definite bugs are errors)
+  | bugFindingAssumingCompleteSpec -- Find bugs assuming complete preconditions (any counterexample is error)
+  deriving Inhabited, Repr, DecidableEq
```
MikaelMayer: We should not derive inhabited, same for checkLevel below. We should construct all instances explicitly to avoid not being to choose the default by mistake

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -682,11 +682,7 @@ def ProofObligation.toSMTTerms (E : Env)
     (λ ts => Term.app (.core .distinct) ts .bool)
   let (assumptions_terms, ctx) ← Core.toSMTTerms E assumptions ctx useArrayTheory
   let (obligation_pos_term, ctx) ← Core.toSMTTerm E [] d.obligation ctx useArrayTheory
-  let obligation_term :=
-    if d.property == .cover then
-      obligation_pos_term
-    else
-      Factory.not obligation_pos_term
+  let obligation_term := obligation_pos_term
```
MikaelMayer: The comment of this function might need to be updated since we now use check-sat-assuming.  Also, since we negate the term later, we should mention it here.

Strata/Languages/Core/SarifOutput.lean
```diff
@@ -19,24 +19,70 @@ open Strata.Sarif Strata.SMT
...
+      if !m1.isEmpty && !m2.isEmpty then s!" (true: {Std.format m1}, false: {Std.format m2})"
+      else if !m1.isEmpty then s!" (true: {Std.format m1})"
+      else if !m2.isEmpty then s!" (false: {Std.format m2})"
+      else ""
+    s!"True or false depending on inputs{models}"
+  | .unsat, .unsat => "Unreachable: path condition is contradictory"
+  | .sat _, .unknown => "Can be true, unknown if always true"
+  | .unsat, .unknown => "Always false if reached, reachability unknown"
+  | .unknown, .sat m =>
+    if m.isEmpty then "Can be false and reachable, unknown if always false"
```
MikaelMayer:
```suggestion
    if m.isEmpty then "Can be false and is reachable, unknown if always false"
```

Strata/Languages/Core/SarifOutput.lean
```diff
@@ -19,24 +19,70 @@ open Strata.Sarif Strata.SMT
...
+    else s!"Can be false and reachable, unknown if always false with counterexample: {Std.format m}"
```
MikaelMayer:
```suggestion
    else s!"Can be false and is reachable, unknown if always false with counterexample: {Std.format m}"
```

StrataMain.lean
```diff
@@ -295,27 +295,39 @@ def pyAnalyzeCommand : Command where
                   if path == pyPath then
                     let pos := (Lean.FileMap.ofString srcText).toPosition fr.range.start
                     -- For failures, show at beginning; for passes, show at end
-                    match vcResult.result with
-                    | .fail => (s!"Assertion failed at line {pos.line}, col {pos.column}: ", "")
-                    | _ => ("", s!" (at line {pos.line}, col {pos.column})")
+                    let isFail := match vcResult.outcome with
+                      | .error _ => true
+                      | .ok outcome => outcome.isRefuted || outcome.isRefutedIfReachable || outcome.isCanBeTrueOrFalse || outcome.isReachableAndCanBeFalse
```
MikaelMayer: We ought to have a (tested) predicate named .isFail, and reuse it below.

Strata/Languages/Core/SMTEncoder.lean
MikaelMayer: I don't know what comment you updated but I was talking about the comment of the function enclosing this let expression. def ProofObligation.toSMTTerms (E : Env)

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -650,26 +650,21 @@ def toSMTTerms (E : Env) (es : List (LExpr CoreLParams.mono)) (ctx : SMT.Context
...
+
+For two-sided verification, encodeCore uses check-sat-assuming:
+- Satisfiability: (check-sat-assuming (Q)) tests if P ∧ Q is satisfiable
+- Validity: (check-sat-assuming ((not Q))) tests if P ∧ ¬Q is satisfiable
+
+For single-check verification, encodeCore uses assert + check-sat:
+- For satisfiability: (assert Q) (check-sat) tests P ∧ Q
+- For validity: (assert (not Q)) (check-sat) tests P ∧ ¬Q
+
+See encodeCore in Verifier.lean for the complete encoding logic.
```
MikaelMayer: The comment of `def ProofObligation.toSMTTerms` and `def encodeCore` are too similar which creates duplication. Make sure there is no hard to maintain duplication between the two functions. Perhaps the second comment should be kept simple as of main.

Strata/Languages/Core/Verifier.lean
```diff
@@ -43,17 +58,42 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
...
+      (message := ("unsat-message", s!"\"Property is always true\""))
+    let negObligationStr := s!"(not {obligationStr})"
+    let _ ← Solver.checkSatAssuming [negObligationStr] ids
+  else
+    if satisfiabilityCheck then
+      -- P ∧ Q satisfiable?
+      Solver.comment "Satisfiability check (P ∧ Q)"
+      Imperative.SMT.addLocationInfo (P := Core.Expression) (md := md)
+        (message := ("sat-message", s!"\"Property can be satisfied\""))
+      Solver.assert obligationId
```
MikaelMayer, line 89: Architecturally, I'm bothered by the fact that this function performs both checks if bothChecks is enabled but performs no check if not (probably delegates the caller to do the check?!). Shouldn't we move the Solver.checkSat here after Solver.assert?

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,112 +176,329 @@ open Std (ToFormat Format format)
...
+  ❌     always false and is reachable                  unsat    sat      yes        error      error       error                Property always false, reachable from declaration entry
+  🔶     can be both true and false and is reachable    sat      sat      yes        error      note        error                Reachable, solver found models for both the property and its negation
+  ⛔     unreachable                                    unsat    unsat    no         warning    warning     warning              Dead code, path unreachable
+  ➕     can be true and is reachable                   sat      unknown  yes        error      note        note                 Property can be true and is reachable, validity unknown
+  ✖️     always false if reached                        unsat    unknown  unknown    error      error       error                Property always false if reached, reachability unknown
+  ➖     can be false and is reachable                  unknown  sat      yes        error      note        error                Q can be false and path is reachable, satisfiability of Q unknown
```
MikaelMayer: I would like messages to be harmonized. 
Property can be true and is reachable, validity unknown

Is not harmonized with

Q can be false and path is reachable, satisfiability of Q unknown

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,112 +176,329 @@ open Std (ToFormat Format format)
...
-    let first := f!"({id}, {formatCexValue e}) "
-    rest.foldl (fun acc (id', e') => acc ++ f!"({id'}, {formatCexValue e'}) ") first
-
-instance : ToFormat LExprModel where
-  format := LExprModel.format
+/-- Mask outcome properties based on requested checks.
+    This ensures that PE-optimized results only show the checks that were requested.
+    Special handling: When masking satisfiability for a refuted property, we preserve
+    the "always false" semantic by keeping validity as the primary signal. -/
+def maskOutcome (outcome : VCOutcome) (satisfiabilityCheck validityCheck : Bool) : VCOutcome :=
```
MikaelMayer, line 432: The existence of this function seems to assume that we do run both checks and then filter these checks afterwards. It seems illogical and aperformance penalty to do so. I understand there is a subtlety with partial evaluation that sometimes can simplify conditions, but could you please explain the rationale behind this decision to "mask outcomes" ? Should this function exist at all?

Strata/Languages/Core/Verifier.lean
```diff
@@ -493,27 +773,31 @@ def verify
...
+        if outcome.isUnreachable then some "assertion holds vacuously (path unreachable)"
+        else if outcome.isPass then none
+        else if outcome.isRefuted then some "assertion does not hold"
+        else if outcome.isCanBeTrueOrFalse then some "assertion does not hold"
+        else if outcome.isSatisfiable then none
+        else if outcome.isRefutedIfReachable then some "assertion does not hold if reached"
+        else if outcome.isReachableAndCanBeFalse then some "assertion does not hold"
+        else if outcome.isAlwaysTrueIfReachable then none
+        else some "assertion could not be proved"
+    message?.map fun message => { fileRange, message }
```
MikaelMayer: There are too many redundancies in the messages. Ideally, we never write one sentence more than once. Also, in the case of minimal vs minimalVerbose: minimal mode should reproduce the same error messages (no "if reached" in that case for example). Review this function carefully

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -214,15 +214,15 @@ procedure Test() returns ()
 info:
 Obligation: rc_assert
 Property: assert
-Result: ✅ pass (❗path unreachable)
+Result: ✅ pass
```
MikaelMayer: The diff with main indicates we have removed an unreachable path hint from the result. This needs to be reinstated. This line should have no diff with main or at least indicate that the assertion pass but the path is unreachable.

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -214,15 +214,15 @@ procedure Test() returns ()
...
 
 Obligation: no_rc_assert
 Property: assert
 Result: ✅ pass
 
 Obligation: rc_cover
 Property: cover
-Result: ❌ fail (❗path unreachable)
+Result: ❌ fail
```
MikaelMayer: Same here, we should indicate that the cover's path is unreachable like main.

StrataTest/Languages/Core/Examples/Regex.lean
```diff
@@ -181,8 +181,7 @@ str.in.re("a", bad_re_loop(1))
 
 
 Result: Obligation: assert_0
-Property: assert
-Result: 🚨 Implementation Error! SMT Encoding Error! Natural numbers expected as indices for re.loop.
+Implementation Error: SMT Encoding Error! Natural numbers expected as indices for re.loop.
```
MikaelMayer: Why are we missing "Property: assert" here?

StrataTest/Languages/Core/Tests/CounterExampleLiftTest.lean
```diff
@@ -129,7 +129,7 @@ Obligation: must_be_cons
 Property: assert
 Result: ❌ fail
 Model:
-($__xs1, Cons(0, Nil))
+($__xs1, (~Cons #0 ~Nil))
```
MikaelMayer: We should use the Core expression printer (aSTtoCST), not the direct expression displayer. Like in main.

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -269,23 +277,21 @@ def dischargeObligation {P : PureExpr} [ToFormat P.Ident] [BEq P.Ident]
   (md : Imperative.MetaData P)
   (smtsolver filename : String)
   (solver_options : Array String) (printFilename : Bool)
-  (reachCheck : Bool := false) :
-  IO (Except Format (Option (Result P.Ident) × Result P.Ident × Strata.SMT.EncoderState)) := do
+  (satisfiabilityCheck validityCheck : Bool) :
+  IO (Except Format (Result P.Ident × Result P.Ident × Strata.SMT.EncoderState)) := do
+  let _ := md  -- md is passed to encodeSMT which handles location info internally
```
MikaelMayer: Is md something we should just remove if not used?

Strata/Languages/Core/Verifier.lean
```diff
@@ -43,17 +58,44 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
...
+  if bothChecks then
+    -- Satisfiability check: P ∧ Q satisfiable?
+    Solver.comment "Satisfiability check (P ∧ Q)"
+    Imperative.SMT.addLocationInfo (P := Core.Expression) (md := md)
+      (message := ("sat-message", s!"\"Property can be satisfied\""))
+    let obligationStr ← Solver.termToSMTString obligationId
+    let _ ← Solver.checkSatAssuming [obligationStr] ids
+
+    -- Validity check: P ∧ ¬Q satisfiable?
+    Solver.comment "Validity check (P ∧ ¬Q)"
```
MikaelMayer: I think we should simplify the solver comments in which P and Q have no meaning. Just "Satisfiability" and "Validity"

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,67 +178,230 @@ open Std (ToFormat Format format)
...
+      match o.satisfiabilityProperty with
+      | .sat _ => "pass"
+      | .unsat => "fail"
+      | .unknown => "unknown"
+      | .err _ => "unknown"
+  -- MinimalVerbose and Full: detailed labels with unreachable indicator
+  else
+    -- Handle unreachable specially
+    if o.unreachable then
+      if property == .assert then "pass (❗path unreachable)"
```
MikaelMayer: I see already a if o.unreachable then with the same messages above. Is there duplication?

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,67 +178,230 @@ open Std (ToFormat Format format)
...
+      else "fail (❗path unreachable)"
+    -- For cover: satisfiability sat means the cover is satisfied (pass)
+    else if property == .cover && o.isSatisfiable then "satisfiable and reachable from declaration entry"
+    else if o.passAndReachable then "always true and is reachable from declaration entry"
+    else if o.alwaysFalseAndReachable then "always false and is reachable from declaration entry"
+    else if o.canBeTrueOrFalseAndIsReachable then "can be both true and false and is reachable from declaration entry"
+    else if o.satisfiableValidityUnknown then "can be true and is reachable from declaration entry"
```
MikaelMayer, line 334: Is this a redundant case with o.satisfiable above?

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,67 +178,230 @@ open Std (ToFormat Format format)
...
+      | .unknown => "unknown"
+      | .err _ => "unknown"
+  -- MinimalVerbose and Full: detailed labels with unreachable indicator
+  else
+    -- Handle unreachable specially
+    if o.unreachable then
+      if property == .assert then "pass (❗path unreachable)"
+      else "fail (❗path unreachable)"
+    -- For cover: satisfiability sat means the cover is satisfied (pass)
+    else if property == .cover && o.isSatisfiable then "satisfiable and reachable from declaration entry"
```
MikaelMayer: I don't think this message is exclusive to cover. It works for assertions as well. I would like to merge the two messages to use both "can be true and is reachable from declaration entry"

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,67 +178,230 @@ open Std (ToFormat Format format)
...
+    else if o.alwaysFalseReachabilityUnknown then "always false if reached"
+    else if o.canBeFalseAndIsReachable then "can be false and is reachable from declaration entry"
+    else if o.passReachabilityUnknown then "always true if reached"
+    else "unknown"
+
+def emoji (o : VCOutcome) (property : Imperative.PropertyType := .assert)
+    (checkLevel : CheckLevel := .minimal) (checkMode : VerificationMode := .deductive) : String :=
+  -- Unreachable is detected when both checks ran
+  if o.unreachable then
+    if property == .assert || property == .divisionByZero then "✅" else "❌"
```
MikaelMayer: I fear we might forget other cases. Can we have a predicate so that we can just run property.passWhenUnreachable

StrataTest/Languages/Core/Tests/ExprEvalTest.lean
```diff
@@ -4,6 +4,11 @@
   SPDX-License-Identifier: Apache-2.0 OR MIT
 -/
 
+-- TODO: This file needs to be updated for the new dischargeObligation API
+-- Temporarily disabled until the test is updated
```
MikaelMayer: Address this todo. No todo should be left in this PR

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -0,0 +1,139 @@
...
+/--
+info:  isPass isAlwaysTrue
+Sat:unknown|Val:unsat ✔️ always true if reached, Always true if reached, reachability unknown, SARIF: Deductive level: none, BugFinding level: none
+-/
+#guard_msgs in
+#eval testOutcome (mkOutcome (Imperative.SMT.Result.unknown (Ident := Core.Expression.Ident)) .unsat) .passReachabilityUnknown
+
+/--
+info:
+Sat:unknown|Val:unknown ❓ unknown, Unknown (solver timeout or incomplete), SARIF: Deductive level: error, BugFinding level: note
```
MikaelMayer, line 134: We need to also test the intermediate mode bugFindingAssumingCompleteSpec in case when we assume complete spec.

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,67 +178,226 @@ open Std (ToFormat Format format)
...
+  - satisfiabilityProperty: result of checking P ∧ Q  (is the property satisfiable given the path condition?)
+  - validityProperty:       result of checking P ∧ ¬Q (can the property be false given the path condition?)
+
+The 9 possible outcomes and their interpretations.
+For cover statements, any outcome where P ∧ Q is sat displays as ✅ (cover satisfied).
+Unreachable covers display as ❌ (error) instead of ⛔ (warning).
+
+  Emoji  Label                                          P ∧ Q    P ∧ ¬Q   Reachable  Deductive  BugFinding  BugFinding+Complete  Meaning
+  -----  ---------------------------------------------  -------  -------  ---------  ---------  ----------  -------------------  -------
+  ✅     always true and is reachable                   sat      unsat    yes        pass       pass        pass                 Property always true, reachable from declaration entry
```
MikaelMayer, line 190: This table ought to be a message before a #guard_msgs that makes sure its entries are in sync with the code.

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -143,4 +143,38 @@ Sat:unknown|Val:unknown ❓ unknown, Unknown (solver timeout or incomplete), SAR
 #guard outcomeToLevel .bugFindingAssumingCompleteSpec .assert (VCOutcome.mk (.sat []) .unsat) = Strata.Sarif.Level.none
 #guard outcomeToLevel .bugFindingAssumingCompleteSpec .assert (VCOutcome.mk .unknown (.sat [])) = Strata.Sarif.Level.error
 
+/-! ### Outcome table verification -/
+
+private def printOutcomeRow (sat val : Imperative.SMT.Result (Ident := Core.Expression.Ident)) : IO Unit := do
+  let o : VCOutcome := { satisfiabilityProperty := sat, validityProperty := val }
+  let e := o.emoji .assert .full .deductive
```
MikaelMayer: I would like here to differentiate the message and emoji if they differ for .cover
Also, make sure the table has (hard-coded) headers.

This line : "✅ pass (❗path unreachable) | Deductive: warning | BugFinding: warning | BugFinding+Complete: warning" should thus become something like
"✅pass (❗path unreachable) | Deductive: warning | BugFinding: warning | BugFinding+Complete: warning | Cover: ❌ fail (❗path unreachable)

All the other lines which have no difference between cover and assert should stay the same.

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -258,34 +266,31 @@ def addLocationInfo {P : PureExpr} [BEq P.Ident]
...
+When two-sided checking is enabled, the generated SMT file will contain two
+`(check-sat-assuming)` commands, one for `P ∧ Q` and one for `P ∧ ¬Q`,
+and the return value includes both decisions.
 -/
 def dischargeObligation {P : PureExpr} [ToFormat P.Ident] [BEq P.Ident]
   (encodeSMT : Strata.SMT.SolverM (List String × Strata.SMT.EncoderState))
   (typedVarToSMTFn : P.Ident → P.Ty → Except Format (String × Strata.SMT.TermType))
   (vars : List P.TypedIdent)
-  (md : Imperative.MetaData P)
+  (_md : Imperative.MetaData P)
```
joscoh: Why is this argument needed if it is never used?

Strata/Languages/Core/SarifOutput.lean
```diff
@@ -19,24 +19,70 @@ open Strata.Sarif Strata.SMT
...
-
-/-- Convert Core Outcome to a descriptive message -/
-def outcomeToMessage (outcome : Outcome) (cex : LExprModel) : String :=
-  match outcome with
-  | .pass => "Verification succeeded"
-  | .fail =>
-    if cex.isEmpty then
-      "Verification failed"
+/-- Convert VCOutcome to SARIF Level -/
+def outcomeToLevel (mode : VerificationMode) (property : Imperative.PropertyType) (outcome : VCOutcome) : Level :=
```
joscoh, line 23: Not blocking, but this function would be easier to read IMO if it matched on `mode, property, outcome` directly instead of all the nested boolean logic.

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,67 +178,226 @@ open Std (ToFormat Format format)
...
+def isAlwaysFalse (o : VCOutcome) : Bool :=
+  o.alwaysFalseAndReachable || o.alwaysFalseReachabilityUnknown
+
+def isAlwaysTrue (o : VCOutcome) : Bool :=
+  o.isPass
+
+def isReachable (o : VCOutcome) : Bool :=
+  o.passAndReachable || o.alwaysFalseAndReachable || o.canBeTrueOrFalseAndIsReachable
+
+-- Backward compatibility aliases (old names with "is" prefix)
```
joscoh, line 278: Why not just use these names directly instead of the prior ones+aliases?

Strata/Languages/Core/Verifier.lean
```diff
@@ -135,67 +178,226 @@ open Std (ToFormat Format format)
...
+    else if o.passAndReachable then "always true and is reachable from declaration entry"
+    else if o.alwaysFalseAndReachable then "always false and is reachable from declaration entry"
+    else if o.canBeTrueOrFalseAndIsReachable then "can be both true and false and is reachable from declaration entry"
+    else if o.satisfiableValidityUnknown then "can be true and is reachable from declaration entry"
+    else if o.alwaysFalseReachabilityUnknown then "always false if reached"
+    else if o.canBeFalseAndIsReachable then "can be false and is reachable from declaration entry"
+    else if o.passReachabilityUnknown then "always true if reached"
+    else "unknown"
+
+def emoji (o : VCOutcome) (property : Imperative.PropertyType := .assert)
```
joscoh, line 340: Why is this function separate from the previous one? It seems to duplicate the same logic, just with emojis instead of text.

Strata/Languages/Core/Verifier.lean
```diff
@@ -207,40 +409,88 @@ def LExprModel.format (cex : LExprModel) : Format :=
...
+    For example, in minimal deductive mode we only request validity, so if PE
+    also determined satisfiability, we mask it to `.unknown`. -/
+def maskOutcome (outcome : VCOutcome) (satisfiabilityCheck validityCheck : Bool) : VCOutcome :=
+  if satisfiabilityCheck && validityCheck then
+    -- Both checks requested: return outcome as-is
+    outcome
+  else if validityCheck && !satisfiabilityCheck then
+    -- Only validity requested: mask satisfiability
+    -- Special case: if property is refuted (.unsat, .sat), keep it as (.unknown, .sat)
+    -- which will display as "can be false and is reachable" instead of "always false and is reachable"
```
joscoh: Could you explain this comment? Isn't "always false and is reachable" the correct answer here?

StrataTest/DL/Imperative/Verify.lean
```diff
@@ -56,9 +56,9 @@ def verify (cmds : Commands) (verbose : Bool) :
                -- (FIXME)
                ((Arith.Eval.ProofObligation.freeVars obligation).map (fun v => (v, Arith.Ty.Num)))
                 Imperative.MetaData.empty "cvc5" filename.toString
-                #["--produce-models"] false false)
+                #["--produce-models"] false false true)
         match ans with
-        | .ok (_, result, estate) =>
+        | Except.ok (_satResult, result, estate) =>
```
joscoh: Why is this variable now needed?

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -305,6 +305,34 @@ Property: cover
...
+  assert [test_pass]: (x >= 0);
+  assert [test_fail]: (x < 0);
+};
+#end
+
+/--
+info:
+Obligation: test_pass
+Property: assert
+Result: ✔️ always true if reached
```
joscoh, line 331: Why isn't this reachable?

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -0,0 +1,185 @@
...
+    if value then IO.print s!" {name}"
+  let satStr := if let .sat _ := o.satisfiabilityProperty then "sat" else if let .unsat := o.satisfiabilityProperty then "unsat" else "unknown"
+  let valStr := if let .sat _ := o.validityProperty then "sat" else if let .unsat := o.validityProperty then "unsat" else "unknown"
+  IO.println s!"\nSat:{satStr}|Val:{valStr} {o.emoji .assert .full .deductive} {o.label .assert .full .deductive}, {outcomeToMessage o}, SARIF: Deductive level: {outcomeToLevel .deductive .assert o}, BugFinding level: {outcomeToLevel .bugFinding .assert o}"
+
+/-! ### Outcome: (sat, unsat) - always true and reachable -/
+
+/--
+info:  isPass isSatisfiable isAlwaysTrue isReachable
+Sat:sat|Val:unsat ✅ always true and is reachable from declaration entry, Always true and reachable, SARIF: Deductive level: none, BugFinding level: none
```
joscoh, line 78: Are these tests just checking if the messages are consistent?

Strata/DL/Imperative/SMTUtils.lean
MikaelMayer: Let me remove it.

Strata/Languages/Core/SarifOutput.lean
MikaelMayer: Let me try that.

StrataTest/Languages/Core/VCOutcomeTests.lean
MikaelMayer: Yes they are

Strata/Languages/Core/Verifier.lean
MikaelMayer: I don't guarantee that the emojis are in sync with the text. The emojis should be in sync with the error level, while the text only states the property that is proved.

Strata/Languages/Core/Verifier.lean
MikaelMayer: At least let me remove every definition that was not previously in `main` to reduce the number of aliases. I shouldn't also say "backward compatible" as it's not a stateless comment

StrataTest/DL/Imperative/Verify.lean
MikaelMayer: It's still not needed, it's just that its name is now explicit.

StrataTest/Languages/Core/Examples/Cover.lean
MikaelMayer: Because the check done is `minimalVerbose`, so only the validity check is performed, meaning we have no information about reachability.

Strata/Languages/Core/Verifier.lean
MikaelMayer: This comment seems outdated since we do not handle special cases here. thanks for reporting.

Strata/Languages/Core/Verifier.lean
joscoh: The "backwards compatible" is still there, but it's not blocking.

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -682,11 +665,7 @@ def ProofObligation.toSMTTerms (E : Env)
     (λ ts => Term.app (.core .distinct) ts .bool)
   let (assumptions_terms, ctx) ← Core.toSMTTerms E assumptions ctx useArrayTheory
   let (obligation_pos_term, ctx) ← Core.toSMTTerm E [] d.obligation ctx useArrayTheory
-  let obligation_term :=
-    if d.property == .cover then
-      obligation_pos_term
-    else
-      Factory.not obligation_pos_term
+  let obligation_term := obligation_pos_term
```
shigoel: nit: could rename obligation_pos_term to obligation_term, but don't want to block on this.

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -336,3 +336,17 @@ Result: ➖ can be false and is reachable from declaration entry
...
+-- Test: minimalVerbose in bugFinding mode (satisfiability check only)
+/--
+info:
+Obligation: test_pass
+Property: assert
+Result: ➕ can be true and is reachable from declaration entry
+
+Obligation: test_fail
+Property: assert
+Result: ✖️ always false if reached
```
shigoel, line 349: If `test_pass` is reachable from the declaration entry, why does `test_fail` say "if reached"?

PR #488, opened 2026-02-26
Remove Visibility from CoreIdent and old() expressions

### LINKED ISSUE #257: Soundness issue: old as operator.
Consider the following code:

```
procedure f()
  modifies x, y
  ensures x == y ==> old(x) != old(y)
{
   if x != y {
      x := y;
   } else {
     x := y + 1;
   }
}
```

Unfortunately, we encode `old()` as an application of "old" operator to the argument.
Hence, it should be provable in lambda that `x == y ==> old(x) == old(y)`, so the procedure above would theoretically able to prove false.

To fix this, we shouldn't use operator application to describe the old value of variables.

MikaelMayer:
Fixes #257

**Problem:** `old(g)` was encoded as an application of an `old` operator to its argument. This is unsound: since `old` is just a function, the SMT solver can prove `old(x) == old(y)` whenever `x == y`, even when `x` and `y` had different pre-state values. This allows the verifier to prove false.

**Solution:** Replace `old(g)` with a plain variable reference named `"old g"`. Since `"old x"` and `"old y"` are now independent variables, `x == y` no longer implies `old x == old y`. The verifier correctly initializes each `"old g"` variable to the pre-state value of `g` before evaluating postconditions.

**Changes:**
- `old(g)` syntax replaced by `old g` (identifier-only, no parentheses)
- `old g` translates to a plain variable reference `"old g"` (not a function application)
- `OldExpressions.lean` deleted — the old normalization/substitution logic is no longer needed
- `CoreIdent` simplified from `Lambda.Identifier Visibility` to `Lambda.Identifier Unit`
- Type checker adds `"old g"` to the postcondition context for all declared globals
- VCG substitutes `"old g"` with the pre-state value of `g` at procedure entry

All existing tests pass.

**Follow-up question for reviewers:** Now that `CoreIdent` uses `Unit` for `IDMeta`, and `C_Simp` also uses `Unit`, every concrete instantiation of `IDMeta` in the codebase is `Unit`. Would reviewers agree to remove `IDMeta` as a type parameter in a follow-up PR, hardcoding `Unit` throughout `DL/Lambda/`? This would simplify the signatures of `LConstr`, `LDatatype`, `TEnv`, `TContext`, `Identifier`, etc.

### DIFF
--- a/.github/workflows/cbmc.yml
+++ b/.github/workflows/cbmc.yml
@@ -52,8 +52,22 @@ jobs:
         run: |
           wget https://github.com/diffblue/cbmc/releases/download/cbmc-6.4.1/ubuntu-22.04-cbmc-6.4.1-Linux.deb
           sudo dpkg -i ubuntu-22.04-cbmc-6.4.1-Linux.deb
+      - name: Restore lake cache
+        uses: actions/cache/restore@v4
+        with:
+          path: .lake
+          key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
+          restore-keys: |
+            lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}
       - name: Build Strata
         uses: leanprover/lean-action@v1
+        with:
+          use-github-cache: false
+      - name: Save lake cache
+        uses: actions/cache/save@v4
+        with:
+          path: .lake
+          key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
       - uses: actions/setup-python@v5
         with:
           python-version: '3.14'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,8 +68,22 @@ jobs:
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '8.0.x'
+      - name: Restore lake cache
+        uses: actions/cache/restore@v4
+        with:
+          path: .lake
+          key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
+          restore-keys: |
+            lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}
       - name: Build and test Strata
         uses: leanprover/lean-action@v1
+        with:
+          use-github-cache: false
+      - name: Save lake cache
+        uses: actions/cache/save@v4
+        with:
+          path: .lake
+          key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
       - name: Build and run StrataVerify
         run: lake exe StrataVerify Examples/SimpleProc.core.st
       - name: Build BoogieToStrata
--- a/Strata/DDM/Elab.lean
+++ b/Strata/DDM/Elab.lean
@@ -10,7 +10,6 @@ public import Strata.DDM.Ion
 import Strata.DDM.Elab.DialectM
 import Strata.DDM.BuiltinDialects
 import Strata.DDM.Util.Ion.Serialize
-import Strata.Util.IO
 
 import all Strata.DDM.Util.ByteArray
 import all Strata.DDM.Util.Lean
--- a/Strata/DL/Imperative/SMTUtils.lean
+++ b/Strata/DL/Imperative/SMTUtils.lean
@@ -164,8 +164,8 @@ def solverResult {P : PureExpr} [ToFormat P.Ident]
     let hasExecError := stderr.contains "could not execute external process"
     let hasFileError := stderr.contains "No such file or directory"
     let suggestion :=
-      if (hasExecError || hasFileError) && smtsolver == defaultSolver then
-        s!" \nEnsure {defaultSolver} is on your PATH or use --solver to specify another SMT solver."
+      if (hasExecError || hasFileError) && smtsolver == Core.defaultSolver then
+        s!" \nEnsure {Core.defaultSolver} is on your PATH or use --solver to specify another SMT solver."
       else ""
     .error s!"stderr:{stderr}{suggestion}\nsolver stdout: {output.stdout}\n"
 
--- a/Strata/DL/SMT/Solver.lean
+++ b/Strata/DL/SMT/Solver.lean
@@ -59,7 +59,7 @@ def spawn (path : String) (args : Array String) : IO Solver := do
     }
     return ⟨IO.FS.Stream.ofHandle proc.stdin, IO.FS.Stream.ofHandle proc.stdout⟩
   catch e =>
-    let suggestion := if path == defaultSolver || path.endsWith defaultSolver then s!" Ensure {defaultSolver} is on your PATH or use --solver to specify another SMT solver." else ""
+    let suggestion := if path == Core.defaultSolver || path.endsWith Core.defaultSolver then s!" Ensure {Core.defaultSolver} is on your PATH or use --solver to specify another SMT solver." else ""
     throw (IO.userError s!"could not execute external process '{path}'.{suggestion} Original error: {e}")
 
 /--
--- a/Strata/Languages/C_Simp/Verify.lean
+++ b/Strata/Languages/C_Simp/Verify.lean
@@ -6,8 +6,11 @@
 
 import Strata.Languages.C_Simp.C_Simp
 import Strata.Languages.C_Simp.DDMTransform.Translate
+import Strata.Languages.Core.Options
 import Strata.Languages.Core.Verifier
 import Strata.DL.Imperative.Stmt
+
+open Core
 
 namespace Strata
 
@@ -133,13 +136,13 @@ def to_core(program : C_Simp.Program) : Core.Program :=
 def C_Simp.get_program (p : Strata.Program) : C_Simp.Program :=
   (Strata.C_Simp.TransM.run Inhabited.default (Strata.C_Simp.translateProgram (p.commands))).fst
 
-def C_Simp.typeCheck (p : Strata.Program) (options : Options := Options.default):
+def C_Simp.typeCheck (p : Strata.Program) (options : VerifyOptions := .default):
   Except DiagnosticModel Core.Program := do
   let program := C_Simp.get_program p
   Core.typeCheck options (to_core program)
 
 def C_Simp.verify (p : Strata.Program)
-    (options : Options := Options.default)
+    (options : VerifyOptions := .default)
     (tempDir : Option String := .none):
   IO Core.VCResults := do
   let program := C_Simp.get_program p
--- a/Strata/Languages/Core/Core.lean
+++ b/Strata/Languages/Core/Core.lean
@@ -34,7 +34,7 @@ open Strata
    requiring a tool like `BoogieToStrata` to desugar them.
 -/
 
-def typeCheck (options : Options) (program : Program)
+def typeCheck (options : VerifyOptions) (program : Program)
     (moreFns : @Lambda.Factory CoreLParams := Lambda.Factory.default) :
     Except DiagnosticModel Program := do
   let T := Lambda.TEnv.default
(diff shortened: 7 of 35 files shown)


Strata/Languages/Core/CmdEval.lean
```diff
@@ -47,8 +46,9 @@ def lookup (E : Env) (v : Expression.Ident) : Option Expression.TypedExpr :=
   | none => none
 
 def preprocess (E : Env) (c : Cmd Expression) (e : Expression.Expr) : Expression.Expr × Env :=
-  let substMap := oldVarSubst E.substMap E
-  let e' := OldExpressions.substsOldExpr substMap e
+  -- Apply "old g" substitutions from substMap (only entries for "old ..." variables)
+  let oldSubst := E.substMap.filter (fun (id, _) => id.name.startsWith "old ")
```
MikaelMayer: Where do we actually need this preprocessing? Please explain me as checking oldness with a prefix seems brittle.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -60,10 +59,10 @@ private def callConditions (proc : Procedure)
 Create substitution mapping from formal parameters to actual arguments.
 -/
 private def mkFormalArgSubst (proc : Procedure) (args : List Expression.Expr) (E : Env)
-    (old_var_subst : SubstMap) : VarSubst :=
-  let args' := args.map (fun a => E.exprEval (OldExpressions.substsOldExpr old_var_subst a))
+    (_old_var_subst : SubstMap) : VarSubst :=
```
MikaelMayer: If _old_var_subst is no longer used could we remove it?

Strata/Languages/Core/CmdEval.lean
MikaelMayer: Good point. Fixed:  now contains only  entries (set in  before calling ), so  can safely apply the full  without any prefix filtering. The prefix check is gone.

Strata/Languages/Core/StatementEval.lean
MikaelMayer: Done, removed.

StrataTest/Languages/Core/Examples/FunctionDeclDDMTest.lean
```diff
@@ -20,6 +20,7 @@ program Core;
 procedure testFuncDecl(c: int) returns () {
   function double(x : int) : int { x + x + c}
   var y : int := 5;
+  assume true;
```
MikaelMayer: Why is there an assume true that was added here?

MikaelMayer: The `assume true;` is part of the test program source (inside the `#strata` block), not something generated by our changes. It was already there before this PR — we only updated the expected output docstring to match what `IO.println` prints (which faithfully reflects the source). The test is checking that the pretty-printer round-trips the program correctly.

MikaelMayer: You're right, I was wrong. We introduced it — both in the source program and the expected output. Fixed: removed `assume true` from both.

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -252,9 +252,9 @@ partial def translateLMonoTy (bindings : TransBindings) (arg : Arg) :
                     -- Datatype Declaration (possibly mutual)
                     -- Look up the type name from the GlobalContext using the fvar index
                     let gctx := (← StateT.get).globalContext
-                    let ldatatype : LDatatype Core.Visibility := match gctx.nameOf? i, block with
+                    let ldatatype : LDatatype Unit := match gctx.nameOf? i, block with
```
atomb, line 255: Are we ever using this parameter for anything anymore?

Strata/Languages/Core/ProcedureEval.lean
```diff
@@ -53,6 +53,12 @@ def eval (E : Env) (p : Procedure) : List (Procedure × Env) :=
   -- `Statement.eval` will substitute `old <var>` where `<var>` is a local
   -- variable with the value of `<var>` at each given statement.
   let old_var_subst := E.exprEnv.state.oldest.map (fun (i, _, e) => (i, e))
+  -- Build "old g" → pre-state value substitutions for all declared globals.
+  -- These are passed as substMap so preprocess can substitute them in postcondition asserts.
+  let globalNames : List String := E.program.decls.filterMap fun d =>
+    match d with | .var name _ _ _ => some name.name | _ => none
+  let old_g_subst := old_var_subst.filterMap fun (id, e) =>
+    if globalNames.contains id.name then some (⟨"old " ++ id.name, ()⟩, e) else none
```
atomb: Given that the "old " ++ pattern occurs in several places, it would be nice to encapsulate it behind a function so that if we want to change how old names are encoded we can do it in a single place.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -196,10 +194,13 @@ def Command.evalCall (E : Env) (old_var_subst : SubstMap)
         (fun (l, c) => (l, { c with expr := c.expr.applySubst tySubst }))
     -- Create post-call substitution for postconditions.
     let postcond_subst_init := formal_arg_subst ++ return_lhs_subst
-    let postcond_subst_map := postcond_subst_init ++ current_globals
-    let postconditions := OldExpressions.substsOldInProcChecks postcond_subst_map postconditions_typed
-    let postcond_subst_full :=  postcond_subst_init ++ globals_post_subst
-    let postconditions := callConditions proc .Ensures postconditions postcond_subst_full
+    -- Build "old g" substitutions: map "old g" → pre-call value of g
+    let old_g_subst : VarSubst := current_globals.filterMap fun ((id, ty), e) =>
+      let oldId : CoreIdent := ⟨"old " ++ id.name, ()⟩
```
atomb: Is this logic to build a substitution duplicated in several parts of this PR?

Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
```diff
@@ -7,6 +7,7 @@
 -- Laurel dialect definition, loaded from LaurelGrammar.st
 -- NOTE: Changes to LaurelGrammar.st are not automatically tracked by the build system.
 -- Update this file (e.g. this comment) to trigger a recompile after modifying LaurelGrammar.st.
+-- Last updated: added function keyword support
```
atomb: This doesn't seem related to this PR?

StrataTest/Languages/Core/OldExpressionsTests.lean
atomb, line 1: You could probably just remove this file.

StrataTest/Languages/Core/ProcedureEvalTests.lean
```diff
@@ -18,7 +18,7 @@ open Procedure Statement Lambda Lambda.LTy.Syntax Lambda.LExpr.SyntaxMono Core.S
 info: Error:
 none
 Subst Map:
-(x, ($__x0 : int)) (y, ($__y1 : int))
+
```
atomb, line 21: What led to this change?

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -286,6 +286,29 @@ private void WriteLine(string text) {
...
+                break;
+            case NAryExpr { Fun: MapSelect } mapSelect:
+                WriteText("(");
+                EmitOldExpr(mapSelect.Args[0]);
+                WriteText(")[");
+                EmitSeparated(mapSelect.Args.Skip(1), (Expr e) => VisitExpr(e), "][");
+                WriteText("]");
+                break;
+            default:
+                // Fallback: wrap in old() — may not parse but better than silently wrong
```
atomb, line 318: I'd really like to handle more cases than the two above. I think Boogie may have code in it already to normalize `old` expressions that you could call.

Strata/Languages/Core/DDMTransform/Translate.lean
MikaelMayer: In the PR description.
Follow-up question for reviewers: Now that CoreIdent uses Unit for IDMeta, and C_Simp also uses Unit, every concrete instantiation of IDMeta in the codebase is Unit. Would reviewers agree to remove IDMeta as a type parameter in a follow-up PR, hardcoding Unit throughout DL/Lambda/? This would simplify the signatures of LConstr, LDatatype, TEnv, TContext, Identifier, etc.

I haven't found it used anywhere myself.

MikaelMayer: However, removing it might cover much more than what this PR is about. I would prefer to keep the two changes separate. Happy to do it in another PR

Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
MikaelMayer: It was a placeholder to force Laurel to recompile in the CI. A Hack. hope my other PR will solve this once and forall

Strata/Languages/Core/DDMTransform/Translate.lean
aqjune-aws: Yes, +1 for removing this type parameter. Can't imagine how many decidable instance parameters will be removed - will save environment by helping AIs use less token :)

Strata/Languages/Core/ProcedureEval.lean
MikaelMayer: Good point. Added `CoreIdent.mkOld` helper in `Identifiers.lean` and replaced all 6 occurrences throughout the codebase.

Strata/Languages/Core/StatementEval.lean
MikaelMayer: The two places build `old g` substitutions from different inputs: `ProcedureEval` filters the full pre-state map to globals only, while `StatementEval.evalCall` builds it from `current_globals` which is already global-only. They're not identical enough to share a helper without adding complexity.

Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
MikaelMayer: Reverted — the extra comment line was a workaround to bust the grammar cache after merging main's `function` keyword commit. The cache reload mechanism will be fixed properly in a separate PR.

StrataTest/Languages/Core/OldExpressionsTests.lean
MikaelMayer: Done, removed the file.

StrataTest/Languages/Core/ProcedureEvalTests.lean
MikaelMayer: The `Subst Map` is now empty because `substMap` only contains `"old g"` entries (set by `ProcedureEval` for globals). This test procedure has no globals, so the map is correctly empty. Previously it contained parameter substitutions which were a side effect of the old design.

Tools/BoogieToStrata/Source/StrataGenerator.cs
MikaelMayer: Added cases for `BvExtractExpr` (`old(bv[end:start])` → `(old bv)[end:start]`) and `BvConcatExpr` (`old(a ++ b)` → `(old a) ++ (old b)`).

Looking at the Boogie expression hierarchy (`AbsyExpr.cs`), the concrete `Expr` subclasses are: `LiteralExpr`, `IdentifierExpr`, `NAryExpr`, `OldExpr`, `BvExtractExpr`, `BvConcatExpr`, and `CodeExpr`. All arithmetic, boolean, comparison, map select/store, function calls, if-then-else, and field access are `NAryExpr`. `OldExpr` cannot appear inside itself (Boogie rejects nested `old`). `LiteralExpr` and `CodeExpr` cannot appear inside `old` in a meaningful way. So the only cases not yet handled were `BvExtractExpr` and `BvConcatExpr`, which are now covered. The fallback remains as a safety net.

Strata/Languages/Core/WF.lean
```diff
@@ -107,21 +105,12 @@ instance (p : Program) : ListP (WFStatementProp p) (WFStatementsProp p) where
 
 structure WFPrePostProp (p : Program) (d : Procedure) (pp : CoreLabel × Procedure.Check)
   : Prop where
```
aqjune-aws, line 107: What about removing WFPrePostProp and two variants below?

Strata/Languages/Core/WF.lean
```diff
@@ -107,21 +105,12 @@ instance (p : Program) : ListP (WFStatementProp p) (WFStatementsProp p) where
...
-  lvars : Forall (fun x =>
-            (CoreIdent.isLocl x) →
-            (x ∈ (ListMap.keys d.header.inputs) ++ (ListMap.keys d.header.outputs)))
-          (HasVarsPure.getVars (P:=Expression) pp.2.expr)
 
 structure WFPreProp (p : Program) (d : Procedure) (pp : CoreLabel × Procedure.Check)
-  : Prop extends WFPrePostProp p d pp
-  where
-  nold : ¬ OldExpressions.containsOldExpr pp.2.expr
+  : Prop extends WFPrePostProp p d pp where
```
aqjune-aws, line 110: Can `nold` still hold, or can preconditions have old global variables?

atomb (approved):

Strata/Languages/Core/WF.lean
MikaelMayer: Yes they were only there to check the well formedness of expressions in the presence of old constructs which is no longer necessary. We might want to add more well formedness conditions but I think we should think of that separately.

Strata/Languages/Core/WF.lean
MikaelMayer: Preconditions do not have the old variables in the context for type checking.

PR #489, opened 2026-02-26
Support Any type, Operations, and Kwargs handling for Python->Laurel

thanhnguyen-aws:
*Description of changes:* 
This PR Any type, Operations, and Kwargs handling for Python->Laurel:
1. Define type Any and ListAny in PythonLaurelCorePrelude as inductive types.
2. Change the encoded type for all variables, functions' inputs and outputs to Any type.
3. Encode Python's operations in PythonLaurelCorePrelude so that they are consistent with Python's semantic.
4.  Support Kwargs handling for Python->Laurel

### DIFF
--- a/Examples/StringTest.laurel.st
+++ b/Examples/StringTest.laurel.st
@@ -4,7 +4,7 @@ requires true
 {
   var message: string := "Hello, World!";
   return message;
-}
+};
 
 procedure testStringConcat()
 returns (result: string)
@@ -13,4 +13,4 @@ requires true
   var hello: string := "Hello";
   var world: string := "World";
   return hello;
-}
+};
--- a/Strata/DDM/Format.lean
+++ b/Strata/DDM/Format.lean
@@ -89,6 +89,8 @@ Options to control parenthesis
 structure FormatOptions where
   /-- Always add parenthesis when feasible. -/
   alwaysParen : Bool := false
+  /-- Use SMT-LIB 2.7 string escaping (`""` for quotes) instead of C-style (`\"`). -/
+  smtStringEscaping : Bool := false
 
 /--
 A format context provides callbacks and information needed to
@@ -383,7 +385,10 @@ private partial def ArgF.mformatM {α} : ArgF α → FormatM PrecFormat
 | .ident _ x => return .atom (formatIdent x)
 | .num _ x => pformat x
 | .decimal _ v => pformat v
-| .strlit _ s => return .atom (.text <| escapeStringLit s)
+| .strlit _ s => do
+    let ctx ← read
+    let esc := if ctx.opts.smtStringEscaping then escapeSMTStringLit s else escapeStringLit s
+    return .atom (.text esc)
 | .bytes _ v => return .atom <| .text <| ByteArray.escapeBytes v
 | .option _ ma =>
   match ma with
--- a/Strata/DDM/Util/Decimal.lean
+++ b/Strata/DDM/Util/Decimal.lean
@@ -41,9 +41,12 @@ def toString (d : Decimal) : String :=
   else if e < 0 ∧ e ≥ minPrettyExponent then
     let ms := if m < 0 then "-" else ""
     let ma := m.natAbs
-    let ne := 10^(-e).natAbs
+    let width := (-e).natAbs
+    let ne := 10^width
     let md := ma % ne
-    s!"{ms}{ma / ne}.{md}"
+    let fracStr := s!"{md}"
+    let padded := String.replicate (width - fracStr.length) '0' ++ fracStr
+    s!"{ms}{ma / ne}.{padded}"
   else
     s!"{m}e{e}"
 
--- a/Strata/DDM/Util/String.lean
+++ b/Strata/DDM/Util/String.lean
@@ -117,5 +117,23 @@ private def escapeStringLitAux (acc : String) (c : Char) : String :=
 def escapeStringLit (s : String) : String :=
   s.foldl escapeStringLitAux "\"" ++ "\""
 
+/--
+Escape a string literal for SMT-LIB 2.7 output.
+In SMT-LIB 2.7, double quotes are escaped by doubling them (`""`),
+and non-printable characters use `\u{XXXX}` escape sequences.
+Backslashes have no special meaning (they are literal).
+-/
+private def escapeSMTStringLitAux (acc : String) (c : Char) : String :=
+  if c == '"' then
+    acc ++ "\"\""
+  else if useXHex c then
+    let hex := String.ofList (Nat.toDigits 16 c.toNat)
+    s!"{acc}\\u\{{hex}}"
+  else
+    acc.push c
+
+def escapeSMTStringLit (s : String) : String :=
+  s.foldl escapeSMTStringLitAux "\"" ++ "\""
+
 end Strata
 end
--- a/Strata/DL/SMT/DDMTransform/Translate.lean
+++ b/Strata/DL/SMT/DDMTransform/Translate.lean
@@ -256,14 +256,14 @@ private def dummy_prg_for_toString :=
 def termToString (t:SMT.Term): Except String String := do
   let ddm_term <- translateFromTerm t
   let ddm_ast := SMTDDM.Term.toAst ddm_term
-  let ctx := dummy_prg_for_toString.formatContext {}
+  let ctx := dummy_prg_for_toString.formatContext { smtStringEscaping := true }
   let s := dummy_prg_for_toString.formatState
   return ddm_ast.render ctx s |>.fst
 
 def termTypeToString (t:SMT.TermType): Except String String := do
   let ddm_term <- translateFromTermType t
   let ddm_ast := SMTDDM.SMTSort.toAst ddm_term
-  let ctx := dummy_prg_for_toString.formatContext {}
+  let ctx := dummy_prg_for_toString.formatContext { smtStringEscaping := true }
   let s := dummy_prg_for_toString.formatState
   return ddm_ast.render ctx s |>.fst
 
--- a/Strata/Languages/Laurel/CoreDefinitionsForLaurel.lean
+++ b/Strata/Languages/Laurel/CoreDefinitionsForLaurel.lean
@@ -24,13 +24,13 @@ program Laurel;
 // The types for these Map functions are incorrect.
 // We'll fix them when Laurel supports polymorphism
 function select(map: int, key: int) : int
-  external
+  external;
 
 function update(map: int, key: int, value: int) : int
-  external
+  external;
 
 function const(value: int) : int
-  external
+  external;
 
 #end
 
--- a/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
+++ b/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
@@ -7,7 +7,7 @@
 -- Laurel dialect definition, loaded from LaurelGrammar.st
 -- NOTE: Changes to LaurelGrammar.st are not automatically tracked by the build system.
 -- Update this file (e.g. this comment) to trigger a recompile after modifying LaurelGrammar.st.
--- Last grammar change: added externalBody op for external procedure/function bodies.
+-- Last grammar change: require semicolon after procedure/function definitions.
 import Strata.DDM.Integration.Lean
 
 namespace Strata.Laurel
--- a/Strata/Languages/Laurel/Grammar/LaurelGrammar.st
+++ b/Strata/Languages/Laurel/Grammar/LaurelGrammar.st
@@ -150,7 +150,7 @@ op procedure (name : Ident, parameters: CommaSepBy Parameter,
   ensures: Seq EnsuresClause,
   modifies: Seq ModifiesClause,
   body : Option OptionalBody) : Procedure =>
-  "procedure " name "(" parameters ")" returnType returnParameters requires ensures modifies body;
+  "procedure " name "(" parameters ")" returnType returnParameters requires ensures modifies body ";";
 
 op function (name : Ident, parameters: CommaSepBy Parameter,
   returnType: Option OptionalReturnType,
@@ -159,7 +159,7 @@ op function (name : Ident, parameters: CommaSepBy Parameter,
   ensures: Seq EnsuresClause,
   modifies: Seq ModifiesClause,
   body : Option OptionalBody) : Procedure =>
-  "function " name "(" parameters ")" returnType returnParameters requires ensures modifies body;
+  "function " name "(" parameters ")" returnType returnParameters requires ensures modifies body ";";
 
 category TopLevel;
 op topLevelComposite(composite: Composite): TopLevel => composite;
--- a/Strata/Languages/Laurel/HeapParameterizationConstants.lean
+++ b/Strata/Languages/Laurel/HeapParameterizationConstants.lean
@@ -49,7 +49,7 @@ datatype Heap {
 // Read a field from the heap: readField(heap, obj, field) = Heap..data!(heap)[obj][field]
 function readField(heap: Heap, obj: Composite, field: Field): Box {
   select(select(Heap..data!(heap), obj), field)
-}
+};
 
 // Update a field in the heap
 function updateField(heap: Heap, obj: Composite, field: Field, val: Box): Heap {
@@ -57,18 +57,13 @@ function updateField(heap: Heap, obj: Composite, field: Field, val: Box): Heap {
     update(Heap..data!(heap), obj,
       update(select(Heap..data!(heap), obj), field, val)),
     Heap..nextReference!(heap))
-}
+};
 
 // Increment the heap allocation nextReference, returning a new heap
 function increment(heap: Heap): Heap {
   MkHeap(Heap..data!(heap), Heap..nextReference!(heap) + 1)
-}
+};
 
-// The #strata macro does not identify the end macro correctly,
-// because Laurel's grammar also parses # signs
-// Having this datatype here brings the parser in a state where it won't consume the #
-// A fix would be to require ';' after the body of functions/procedures
-datatype Workaround  {}
 #end
 
 /-- The Laurel Core prelude as a Laurel Program. -/
--- a/StrataTest/Backends/CBMC/contracts/test_contract.lr.st
+++ b/StrataTest/Backends/CBMC/contracts/test_contract.lr.st
@@ -3,9 +3,9 @@ procedure add(x: int, y: int) returns (r: int)
   ensures r >= x
 {
   r := x + y;
-}
+};
 
 procedure main() {
   var a: int := 42;
   assert a > 0;
-}
+};
--- a/StrataTest/DDM/Util/DecimalTests.lean
+++ b/StrataTest/DDM/Util/DecimalTests.lean
@@ -21,5 +21,11 @@ open Strata
 #guard s!"{Decimal.mk (-134) (-2)}" = "-1.34"
 #guard s!"{Decimal.mk (-142) 10}" = "-142e10"
 #guard s!"{Decimal.mk (-142) 10}" = "-142e10"
+-- leading zeros in fractional part
+#guard s!"{Decimal.mk 2 (-2)}"  = "0.02"
+#guard s!"{Decimal.mk 1 (-2)}"  = "0.01"
+#guard s!"{Decimal.mk 5 (-3)}"  = "0.005"
+#guard s!"{Decimal.mk 1 (-3)}"  = "0.001"
+#guard s!"{Decimal.mk (-2) (-2)}" = "-0.02"
 
 end Strata.Decimal.Tests
--- a/StrataTest/DL/SMT/DDMTransform/TranslateTests.lean
+++ b/StrataTest/DL/SMT/DDMTransform/TranslateTests.lean
@@ -18,10 +18,19 @@ namespace Strata.SMTDDM
 #guard_msgs in #eval (termToString
     (.app SMT.Op.add [(.prim (.int 10)), (.prim (.int (-20)))] .int))
 
-/-- info: Except.ok "(+ 0.1 0.2)" -/
+/-- info: Except.ok "(+ 0.1 0.02)" -/
 #guard_msgs in #eval (termToString
     (.app SMT.Op.add [(.prim (.real (Decimal.mk 1 (-1)))),
                       (.prim (.real (Decimal.mk 2 (-2))))] .int))
+
+/-- info: Except.ok "0.01" -/
+#guard_msgs in #eval (termToString (.prim (.real (Decimal.mk 1 (-2)))))
+
+/-- info: Except.ok "0.005" -/
+#guard_msgs in #eval (termToString (.prim (.real (Decimal.mk 5 (-3)))))
+
+/-- info: Except.ok "0.001" -/
+#guard_msgs in #eval (termToString (.prim (.real (Decimal.mk 1 (-3)))))
 
 /-- info: Except.ok "(_ bv1 32)" -/
 #guard_msgs in #eval (termToString
--- a/StrataTest/Languages/Core/Tests/SMTEncoderTests.lean
+++ b/StrataTest/Languages/Core/Tests/SMTEncoderTests.lean
@@ -195,6 +195,15 @@ info: "; m\n(declare-const m (Array Int Int))\n(define-fun t0 () (Array Int Int)
   (.quant () .all "x" (.some .int) (LExpr.noTrigger ())
    (.eq () (.bvar () 0) (.fvar () "x" (.some .int))))
 
+-- Test string literal containing double quotes is properly escaped for SMT-LIB 2.7
+-- In SMT-LIB 2.7, double quotes inside strings are escaped by doubling: "a""b" represents a"b
+/--
+info: "; x\n(declare-const x String)\n(define-fun t0 () String x)\n(define-fun t1 () Bool (= t0 \"{\"\"key\"\":\"\"val\"\"}\"))\n"
+-/
+#guard_msgs in
+#eval toSMTTermString
+  (.eq () (.fvar () "x" (.some .string)) (.strConst () "{\"key\":\"val\"}"))
+
 end ArrayTheory
 
 end Core
@@ -240,5 +249,33 @@ Result: ✅ pass
 -/
 #guard_msgs in
 #eval! verify simpleMapProgram (options := {Core.VerifyOptions.quiet with useArrayTheory := true})
+
+-- Test that string literals with embedded double quotes are correctly encoded for SMT
+def quotedStringProgram :=
+#strata
+program Core;
+
+var x: string;
+
+procedure Test() returns ()
+spec { ensures true; }
+{
+  assume x == "{\"key\":\"val\"}";
+  assert x == "{\"key\":\"val\"}";
+};
+#end
+
+/--
+info:
+Obligation: assert_0
+Property: assert
+Result: ✅ pass
+
+Obligation: Test_ensures_0
+Property: assert
+Result: ✅ pass
+-/
+#guard_msgs in
+#eval! verify quotedStringProgram (options := Core.VerifyOptions.quiet)
 
 end Strata
--- a/StrataTest/Languages/Laurel/DivisionByZeroCheckTest.lean
+++ b/StrataTest/Languages/Laurel/DivisionByZeroCheckTest.lean
@@ -24,31 +24,31 @@ procedure safeDivision() {
   var y: int := 2;
   var z: int := x / y;
   assert z == 5;
-}
+};
 
 // Error ranges are too wide because Core does not use expression locations
 procedure unsafeDivision(x: int) {
   var z: int := 10 / x;
 //^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
 // Error ranges are too wide because Core does not use expression locations
-}
+};
 
 function pureDiv(x: int, y: int): int
   requires y != 0
 {
   x / y
-}
+};
 
 procedure callPureDivSafe() {
   var z: int := pureDiv(10, 2);
   assert z == 5;
-}
+};
 
 procedure callPureDivUnsafe(x: int) {
   var z: int := pureDiv(10, x);
 //^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
 // Error ranges are too wide because Core does not use expression locations
-}
+};
 "
 
 #guard_msgs(drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T10_ConstrainedTypes.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T10_ConstrainedTypes.lean
@@ -23,7 +23,7 @@ composite None extends Option
 constrained SealedOption = x: Option where x is Some || x is None witness None
 
 procedure foo() returns (r: nat) {
-}
+};
 "
 
 -- Not working yet
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T12_Operators.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T12_Operators.lean
@@ -24,7 +24,7 @@ procedure testArithmetic() {
     assert z == 7;
     var r: int := 17 % 5;
     assert r == 2;
-}
+};
 
 procedure testLogical() {
     var t: bool := true;
@@ -37,20 +37,20 @@ procedure testLogical() {
     assert c == true;
     assert t ==> t;
     assert f ==> t;
-}
+};
 
 procedure testUnary() {
     var x: int := 5;
     var y: int := -x;
     assert y == 0 - 5;
-}
+};
 
 procedure testTruncatingDiv() {
     assert 7 /t 3 == 2;
     assert 7 %t 3 == 1;
     assert (0 - 7) /t 3 == 0 - 2;
     assert (0 - 7) %t 3 == 0 - 1;
-}
+};
 "
 
 #guard_msgs(drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T13_WhileLoops.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T13_WhileLoops.lean
@@ -21,7 +21,7 @@ procedure countDown() {
         i := i - 1;
     }
     assert i == 0;
-}
+};
 
 procedure countUp() {
     var n: int := 5;
@@ -33,7 +33,7 @@ procedure countUp() {
         i := i + 1;
     }
     assert i == n;
-}
+};
 "
 
 #guard_msgs(drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T14_Quantifiers.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T14_Quantifiers.lean
@@ -15,17 +15,17 @@ namespace Laurel
 def quantifiersProgram := r"
 procedure testForall() {
     assert forall(x: int) => x + 0 == x;
-}
+};
 
 procedure testExists() {
     assert exists(x: int) => x == 42;
-}
+};
 
 procedure testQuantifierInContract(n: int)
   requires n > 0
   ensures forall(i: int) => i >= 0 ==> i < n ==> i < n + 1
 {
-}
+};
 "
 
 #guard_msgs(drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T15_StringConcatLifting.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T15_StringConcatLifting.lean
@@ -20,7 +20,7 @@ requires true
   var y: string := x ++ (x := " World";);
   assert y == "Hello World";
   assert x == " World";
-}
+};
 
 procedure stringConcatOK()
 requires true
@@ -29,7 +29,7 @@ requires true
   var b: string := " World";
   var c: string := a ++ b;
   assert c == "Hello World";
-}
+};
 
 procedure stringConcatKO()
 requires true
@@ -39,7 +39,7 @@ requires true
   var c: string := a ++ b;
   assert c == "Goodbye";
 //^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
-}
+};
 "#
 
 #guard_msgs (error, drop all) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T1_AssertFalse.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T1_AssertFalse.lean
@@ -19,12 +19,12 @@ procedure foo() {
 //  ^^^^^^^^^^^^^ error: assertion does not hold
     assert false;
 //  ^^^^^^^^^^^^^ error: assertion does not hold
-}
+};
 
 procedure bar() {
     assume false;
     assert false;
-}
+};
 "
 
 #guard_msgs(drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
@@ -20,13 +20,13 @@ procedure nestedImpureStatements() {
   assert x == y;
 //^^^^^^^^^^^^^^ error: assertion does not hold
   assert z == y;
-}
+};
 
 procedure multipleAssignments() {
   var x: int := 1;
   var y: int := x + ((x := 2;) + x) + (x := 3;);
   assert y == 8;
-}
+};
 
 procedure conditionalAssignmentInExpression(x: int) {
   var y: int := 0;
@@ -38,14 +38,14 @@ procedure conditionalAssignmentInExpression(x: int) {
     assert z == 0;
     assert y == 0;
   }
-}
+};
 
 procedure anotherConditionAssignmentInExpression(c: bool) {
   var b: bool := c;
   var z: bool := (if (b) { b := false; } else (b := true;)) || b;
   assert z;
 //^^^^^^^^^ error: assertion does not hold
-}
+};
 
 procedure blockWithTwoAssignmentsInExpression() {
   var x: int := 0;
@@ -54,7 +54,7 @@ procedure blockWithTwoAssignmentsInExpression() {
   assert x == 1;
   assert y == 2;
   assert z == 2;
-}
+};
 
 procedure nestedImpureStatementsAndOpaque()
   ensures true
@@ -65,7 +65,7 @@ procedure nestedImpureStatementsAndOpaque()
   assert x == y;
 //^^^^^^^^^^^^^^ error: assertion does not hold
   assert z == y;
-}
+};
 
 // An imperative procedure call in expression position is lifted before the
 // surrounding expression is evaluated.
@@ -75,7 +75,7 @@ procedure imperativeProc(x: int) returns (r: int)
 {
   r := x + 1;
   r
-}
+};
 
 procedure imperativeCallInExpressionPosition() {
   var x: int := 0;
@@ -84,7 +84,7 @@ procedure imperativeCallInExpressionPosition() {
   var y: int := imperativeProc(x) + x;
   assert y == 1;
   assert x == 0;
-}
+};
 
 // An imperative call inside a conditional expression is also lifted.
 procedure imperativeCallInConditionalExpression(b: bool) {
@@ -96,7 +96,7 @@ procedure imperativeCallInConditionalExpression(b: bool) {
   } else {
     assert result == 0;
   }
-}
+};
 "
 
 #guard_msgs (error, drop all) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressionsError.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressionsError.lean
@@ -17,24 +17,24 @@ procedure impure(): int {
   var x: int := 0;
   x := x + 1;
   x
-}
+};
 
 function impureFunction1(x: int): int
 {
   x := x + 1;
 //^^^^^^^^^^^ error: destructive assignments are not supported in functions or contracts
-}
+};
 
 function impureFunction2(x: int): int
 {
   while(false) {}
 //^^^^^^^^^^^^^^^ error: loops are not supported in functions or contracts
-}
+};
 function impureFunction3(x: int): int
 {
   impure()
 //^^^^^^^^ error: calls to procedures are not supported in functions or contracts
-}
+};
 
 procedure impureContractIsNotLegal1(x: int)
   requires x == impure()
@@ -42,7 +42,7 @@ procedure impureContractIsNotLegal1(x: int)
 {
   assert impure() == 1;
 //       ^^^^^^^^ error: calls to procedures are not supported in functions or contracts
-}
+};
 
 procedure impureContractIsNotLegal2(x: int)
   requires (x := 2;) == 2
@@ -50,7 +50,7 @@ procedure impureContractIsNotLegal2(x: int)
 {
   assert (x := 2;) == 2;
 //        ^^^^^^^ error: destructive assignments are not supported in functions or contracts
-}
+};
 "
 
 #guard_msgs (error, drop all) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T3_ControlFlow.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T3_ControlFlow.lean
@@ -23,7 +23,7 @@ function returnAtEnd(x: int) returns (r: int) {
   } else {
     return 3;
   }
-}
+};
 
 function guardInFunction(x: int) returns (r: int) {
   if (x > 0) {
@@ -35,7 +35,7 @@ function guardInFunction(x: int) returns (r: int) {
   }
 
   return 3;
-}
+};
 
 procedure testFunctions() {
   // assert letsInFunction() == 2;
@@ -48,7 +48,7 @@ procedure testFunctions() {
   assert guardInFunction(1) == 1;
   assert guardInFunction(1) == 2;
 //^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
-}
+};
 
 procedure guards(a: int) returns (r: int)
 {
@@ -66,7 +66,7 @@ procedure guards(a: int) returns (r: int)
     assert e < 3;
 //  ^^^^^^^^^^^^^ error: assertion does not hold
   return e;
-}
+};
 
 procedure dag(a: int) returns (r: int)
 {
@@ -82,7 +82,7 @@ procedure dag(a: int) returns (r: int)
 //  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
 //  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
   return b;
-}
+};
 "
 
 #guard_msgs (error, drop all) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T3_ControlFlowError.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T3_ControlFlowError.lean
@@ -20,7 +20,7 @@ function assertAndAssumeInFunctions(a: int) returns (r: int)
   assume true;
 //^^^^^^^^^^^^ error: assumes are not YET supported in functions or contracts
   a
-}
+};
 
 // Lettish bindings in functions not yet supported
 // because Core expressions do not support let bindings
@@ -32,19 +32,19 @@ function letsInFunction() returns (r: int) {
   var z: int := y + 1;
 //^^^^^^^^^^^^^^^^^^^^ error: local variables in functions are not YET supported
   z
-}
+};
 
 function localVariableWithoutInitializer(): int {
   var x: int;
 //^^^^^^^^^^^ error: local variables in functions must have initializers
   3
-}
+};
 
 function deadCodeAfterIfElse(x: int) returns (r: int) {
   if (x > 0) { return 1; } else { return 2; }
 //^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: if-then-else only supported as the last statement in a block
   return 3;
-}
+};
 "
 
 #guard_msgs (error, drop all) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T4_LoopJumps.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T4_LoopJumps.lean
@@ -33,7 +33,7 @@ procedure whileWithBreakAndContinue(steps: int, continueSteps: int, exitSteps: i
     }
   } breakBlock;
   counter;
-}
+};
 "
 
 -- Not working yet
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T5_ProcedureCalls.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T5_ProcedureCalls.lean
@@ -19,14 +19,14 @@ procedure fooReassign(): int {
   assert x == 1;
   x := x + 1;
   x
-}
+};
 
 procedure fooSingleAssign(): int {
   var x: int := 0;
   var x2: int := x + 1;
   var x3: int := x2 + 1;
   x3
-}
+};
 
 procedure fooProof() {
   var x: int := fooReassign();
@@ -34,17 +34,17 @@ procedure fooProof() {
 // The following assertions fails while it should succeed,
 // because Core does not yet support transparent procedures
 //  assert x == y;
-}
+};
 
 function aFunction(x: int): int
 {
   x
-}
+};
 
 procedure aFunctionCaller() {
   var x: int := aFunction(3);
   assert x == 3;
-}
+};
 "
 
 #guard_msgs (drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
@@ -23,24 +23,24 @@ procedure hasRequires(x: int) returns (r: int)
     assert x > 3;
 //  ^^^^^^^^^^^^^ error: assertion does not hold
   x + 1
-}
+};
 
 procedure caller() {
   var x: int := hasRequires(1);
   var y: int := hasRequires(3);
-}
+};
 
 function aFunctionWithPrecondition(x: int): int
   requires x == 10
 {
   x
-}
+};
 
 procedure aFunctionWithPreconditionCaller() {
   var x: int := aFunctionWithPrecondition(0);
 //^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
 // Error ranges are too wide because Core does not use expression locations
-}
+};
 "
 
 #guard_msgs (drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T7_Decreases.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T7_Decreases.lean
@@ -18,30 +18,31 @@ A procedure with a decreases clause may be called in an erased context.
 -/
 
 def program := r"
-procedure noDecreases(x: int): boolean
+procedure noDecreases(x: int): boolean;
 procedure caller(x: int)
   requires noDecreases(x)
 //                    ^ error: noDecreases can not be called from a pure context, because it is not proven to terminate
+;
 
 procedure noCyclicCalls()
   decreases []
 {
   leaf();
-}
+};
 
-procedure leaf() decreases [1] { }
+procedure leaf() decreases [1] { };
 
 procedure mutualRecursionA(x: nat)
   decreases [x, 1]
 {
   mutualRecursionB(x);
-}
+};
 
 procedure mutualRecursionB(x: nat)
   decreases [x, 0]
 {
   if x != 0 { mutualRecursionA(x-1); }
-}
+};
 "
 
 -- Not working yet
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T8_Postconditions.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T8_Postconditions.lean
@@ -19,20 +19,20 @@ procedure opaqueBody(x: int) returns (r: int)
 {
   if (x > 0) { r := x; }
   else { r := 1; }
-}
+};
 
 procedure callerOfOpaqueProcedure() {
   var x: int := opaqueBody(3);
   assert x > 0;
   assert x == 3;
 //^^^^^^^^^^^^^^ error: assertion does not hold
-}
+};
 
 procedure invalidPostcondition(x: int)
     ensures false
 //          ^^^^^ error: assertion does not hold
 {
-}
+};
 "
 
 #guard_msgs (drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T8_PostconditionsErrors.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T8_PostconditionsErrors.lean
@@ -22,7 +22,7 @@ function opaqueFunction(x: int) returns (r: int)
 // The above limitation is because functions in Core do not support postconditions
 {
   x
-}
+};
 
 procedure callerOfOpaqueFunction() {
   var x: int := opaqueFunction(3);
@@ -30,7 +30,7 @@ procedure callerOfOpaqueFunction() {
 // The following assertion should fail but does not
 // Because Core does not support opaque functions
   assert x == 3;
-}
+};
 "
 
 #guard_msgs (drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T8b_EarlyReturnPostconditions.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T8b_EarlyReturnPostconditions.lean
@@ -20,7 +20,7 @@ procedure earlyReturnCorrect(x: int) returns (r: int)
     return -x;
   }
   return x;
-}
+};
 
 procedure earlyReturnBuggy(x: int) returns (r: int)
   ensures r >= 0
@@ -32,7 +32,7 @@ procedure earlyReturnBuggy(x: int) returns (r: int)
     return x;
   }
   return x;
-}
+};
 "
 
 #guard_msgs (drop info, error) in
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T9_Nondeterministic.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T9_Nondeterministic.lean
@@ -17,7 +17,7 @@ nondet procedure nonDeterministic(x: int): (r: int)
   ensures r > 0
 {
   assumed
-}
+};
 
 procedure caller() {
   var x = nonDeterministic(1)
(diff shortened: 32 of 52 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -294,25 +322,23 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
-      | .NotEq _ => .ok Operation.Neq
-      | .Lt _ => .ok Operation.Lt
-      | .LtE _ => .ok Operation.Leq
-      | .Gt _ => .ok Operation.Gt
-      | .GtE _ => .ok Operation.Geq
-      | .In _ => return mkStmtExprMd .Hole  -- Abstract: arbitrary bool (sound)
-      | .NotIn _ => return mkStmtExprMd .Hole  -- Abstract: arbitrary bool (sound)
+    let preludeOpnames ← match ops.val[0]! with
+      | .Eq _ => .ok "PEq"
+      | .NotEq _ => .ok "PNeq"
```
andrewmwells-amazon:
```suggestion
      | .NotEq _ => .ok "PNEq"
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,599 @@
...
+  else if (Any..isfrom_datetime(v1) && Any..isfrom_int(v2)) then
+    from_datetime(Any..as_datetime(v1) - Any..as_int(v2))
+  else if (Any..isfrom_datetime(v1) && Any..isfrom_datetime(v2)) then
+    from_int(Any..as_datetime(v1) - Any..as_datetime(v2))
+  else
+    exception(UndefinedError ("Operand Type is not defined"))
+}
+
+
+inline function PMul (v1: Any, v2: Any) : Any
```
andrewmwells-amazon, line 329: There are a bunch of bugs here from copying PAdd. E.g.,
`(Any..isfrom_string(v1) && Any..isfrom_string(v2))` does concat.

`from_float(int_to_real(Any..as_int(v1)) + Any..as_float(v2))`
uses `+` instead of `*`

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -401,53 +426,226 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
-        | some ⟨.UserDefined className, _⟩ =>
-          (s!"{className}_{methodAttr.val}", obj :: args)
-        | _ => (pyExprToString f, args)
-      | _ => (pyExprToString f, args)
-    | _ => (pyExprToString f, args)
-  -- Step 3: translate the resolved call
-  let mut translatedArgs ← finalArgs.mapM (translateExpr ctx)
-
-  -- Check if function has a model
-  if !hasModel ctx funcName then
```
andrewmwells-amazon, line 437: We want to keep this.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -264,27 +303,16 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
-        if highEq leftType { val := .TString, md := leftExpr.md } then
-          .ok Operation.StrConcat
-        else
-          .ok Operation.Add
-      | .Sub _ => .ok Operation.Sub
-      | .Mult _ => .ok Operation.Mul
-      | .FloorDiv _ => .ok Operation.Div  -- Python // maps to Laurel Div
-      | .Mod _ => .ok Operation.Mod
-      | .Div _ => .ok Operation.Div  -- Python / (true division)
-      | .BitAnd _ => .ok Operation.And  -- Bitwise & - abstract as logical And
```
andrewmwells-amazon, line 291: We probably want to keep these

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -294,25 +322,23 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
     let leftExpr ← translateExpr ctx left
     let rightExpr ← translateExpr ctx comparators.val[0]!
-    let laurelOp ← match ops.val[0]! with
-      | .Eq _ => .ok Operation.Eq
-      | .NotEq _ => .ok Operation.Neq
-      | .Lt _ => .ok Operation.Lt
-      | .LtE _ => .ok Operation.Leq
-      | .Gt _ => .ok Operation.Gt
-      | .GtE _ => .ok Operation.Geq
-      | .In _ => return mkStmtExprMd .Hole  -- Abstract: arbitrary bool (sound)
```
andrewmwells-amazon, line 313: We also want to keep these

Strata/Languages/Python/PythonToLaurel.lean
andrewmwells-amazon: Also, (a todo for me) CI should test this

Strata/Languages/Python/PythonLaurelCorePrelude.lean
thanhnguyen-aws: I fixed them

StrataTest/Languages/Python/expected_laurel/test_function_def_calls.expected
```diff
@@ -1,12 +1,10 @@
...
 ==== Verification Results ====
-datetime_now_ensures_0: ✅ pass (at byte 7134)
-datetime_utcnow_ensures_0: ✅ pass (at byte 7372)
-ensures_str_strp_reverse: ✅ pass (at byte 8766)
-assert_name_is_foo: ✅ pass (at byte 11081)
-assert_opt_name_none_or_str: ✅ pass (at byte 11131)
-assert_opt_name_none_or_bar: ✅ pass (at byte 11278)
-ensures_maybe_except_none: ✅ pass (at byte 10984)
-(Origin_test_helper_procedure_Requires)req_name_is_foo: 🟡 unknown (at byte 10646)
-(Origin_test_helper_procedure_Requires)req_opt_name_none_or_str: ✅ pass (at byte 10695)
```
shigoel, line 11: @thanhnguyen-aws Why don't I see this particular obligation in the new results?

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,611 @@
...
+inline function isInstance_of_Float (v: Any) : Any {
+  from_bool (Any..isfrom_float(v) || Any..isfrom_int(v) || Any..isfrom_bool(v))
+}
+
+inline function Any_to_bool (v: Any) : bool {
+  if (Any..isfrom_bool(v)) then Any..as_bool!(v) else
+  if (Any..isfrom_none(v)) then false else
+  if (Any..isfrom_string(v)) then !(Any..as_string!(v) == "") else
+  if (Any..isfrom_int(v)) then !(Any..as_int!(v) == 0) else
+  false
```
shigoel: I know `false` is a placeholder here, but it seems unsafe to use here because this function will silently return it on unsupported cases.

More generally, my point is: what's the plan to deal with unsupported cases? Are pure functions the right abstraction?

StrataTest/Languages/Python/expected_laurel/test_function_def_calls.expected
thanhnguyen-aws: I think this one "req_opt_name_none_or_bar" implies this one "req_opt_name_none_or_str", so I removed the later. But I am happy to add it back.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
shigoel, line 1: I'd like to request comments (similar to docString style) in the prelude code below, especially since this is hand-written and has to be maintained manually, and as such, should be held to the same high standard as Lean code in this repo.

shigoel: Preferably, please link to relevant Python documentation in these comments that justifies the design choices that were made.

shigoel: Also, please divide the file into sections, with descriptions of what constructs/operations.. etc. are being handled in each section.

StrataTest/Languages/Python/expected_laurel/test_function_def_calls.expected
shigoel: I'd love for the checks to be exactly the same so that we can compare apples to apples. Also true for other tests in this PR.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I added it back

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I added it back

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I added it back

Strata/Languages/Python/PythonLaurelCorePrelude.lean
thanhnguyen-aws: I added the documents into the prelude.

StrataTest/Languages/Python/expected_laurel/test_precondition_verification.expected
```diff
@@ -1,21 +1,25 @@
...
+(Origin_test_helper_procedure_Requires)req_opt_name_none_or_bar: ✅ pass (at byte 26855)
+(Origin_test_helper_procedure_Requires)req_name_is_foo: ✅ pass (at byte 26690)
+(Origin_test_helper_procedure_Requires)req_opt_name_none_or_str: ✅ pass (at byte 26752)
+(Origin_test_helper_procedure_Requires)req_opt_name_none_or_bar: ✅ pass (at byte 26855)
+(Origin_test_helper_procedure_Requires)req_name_is_foo: 🟡 unknown (at byte 26690)
+(Origin_test_helper_procedure_Requires)req_opt_name_none_or_str: ✅ pass (at byte 26752)
+(Origin_test_helper_procedure_Requires)req_opt_name_none_or_bar: ✅ pass (at byte 26855)
+(Origin_test_helper_procedure_Requires)req_name_is_foo: ✅ pass (at byte 26690)
+(Origin_test_helper_procedure_Requires)req_opt_name_none_or_str: ✅ pass (at byte 26752)
+(Origin_test_helper_procedure_Requires)req_opt_name_none_or_bar: 🟡 unknown (at byte 26855)
```
andrewmwells-amazon: Concerned about this regression

thanhnguyen-aws: I think the correct result should be "unknown" instead of "pass" because the input is "Bar", not "bar". It is also consistent with the corresponding result in expected_non_laurel.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,686 @@
...
+
+inline function Any_to_bool (v: Any) : bool
+  requires (Any..isfrom_bool(v) || Any..isfrom_none(v) || Any..isfrom_string(v) || Any..isfrom_int(v));
+{
+  if (Any..isfrom_bool(v)) then Any..as_bool!(v) else
+  if (Any..isfrom_none(v)) then false else
+  if (Any..isfrom_string(v)) then !(Any..as_string!(v) == "") else
+  if (Any..isfrom_int(v)) then !(Any..as_int!(v) == 0) else
+  false
+  //WILL BE ADDED
```
shigoel, line 161: Is this another instance of `false` being a placeholder value?

thanhnguyen-aws: This function is guarded by the requires clause, so in case `v` is not one of the 4 types, it will returns false, but the assertion of this function will fail.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+  else if (Any..isfrom_float(v1) && Any..isfrom_int(v2)) then
+    from_float(Any..as_float!(v1) + int_to_real(Any..as_int!(v2)) )
+  else if (Any..isfrom_float(v1) && Any..isfrom_float(v2)) then
+    from_float(Any..as_float!(v1) + Any..as_float!(v2))
+  else if (Any..isfrom_string(v1) && Any..isfrom_string(v1)) then
+    from_string(str.concat(Any..as_string!(v1),Any..as_string!(v2)))
+  else if (Any..isfrom_ListAny(v1) && Any..isfrom_ListAny(v2)) then
+    from_ListAny(List_extend(Any..as_ListAny!(v1),Any..as_ListAny!(v2)))
+  else if (Any..isfrom_datetime(v1) && Any..isfrom_int(v2)) then
+    from_datetime((Any..as_datetime!(v1) + Any..as_int!(v2)))
```
joehendrix: Bug: The second condition checks `v1` instead of `v2`. As written, string concatenation triggers whenever `v1` is a string regardless of `v2`'s type, and `Any..as_string!(v2)` on the next line will panic if `v2` is not a string.

```suggestion
  else if (Any..isfrom_string(v1) && Any..isfrom_string(v2)) then
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+inline function isTypeError (e: Error) : Any {
+  from_bool (Error..isTypeError(e))
+}
+
+inline function isAttributeError (e: Error) : Any {
+  from_bool (Error..isTypeError(e))
+}
+
+inline function isAssertionError (e: Error) : Any {
+  from_bool (Error..isAssertionError(e))
```
joehendrix, line 132: Bug: `isAttributeError` dispatches to `Error..isTypeError` instead of `Error..isAttributeError`. Any code checking for `AttributeError` will incorrectly match `TypeError` instead.

```suggestion
  from_bool (Error..isAttributeError(e))
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+    from_float(Any..as_float!(v1) + bool_to_real(Any..as_bool!(v2)))
+  else if (Any..isfrom_int(v1) && Any..isfrom_int(v2)) then
+    from_int(Any..as_int!(v1) + Any..as_int!(v2))
+  else if (Any..isfrom_int(v1) && Any..isfrom_float(v2)) then
+    from_float(int_to_real(Any..as_int!(v1)) + Any..as_float!(v2))
+  else if (Any..isfrom_float(v1) && Any..isfrom_int(v2)) then
+    from_float(Any..as_float!(v1) + int_to_real(Any..as_int!(v2)) )
+  else if (Any..isfrom_float(v1) && Any..isfrom_float(v2)) then
+    from_float(Any..as_float!(v1) + Any..as_float!(v2))
+  else if (Any..isfrom_string(v1) && Any..isfrom_string(v1)) then
```
joehendrix: Bug: The second condition checks `v1` instead of `v2`. As written, string concatenation triggers whenever `v1` is a string regardless of `v2`'s type, and `Any..as_string!(v2)` on the next line will panic if `v2` is not a string.

```suggestion
  else if (Any..isfrom_string(v1) && Any..isfrom_string(v2)) then
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+{
+  if Any..isexception(v1) then v1 else if Any..isexception(v2) then v2
+  else if (Any..isfrom_bool(v1) && Any..isfrom_bool(v2)) then
+    from_int(bool_to_int(Any..as_bool!(v1)) * bool_to_int(Any..as_bool!(v2)))
+  else if (Any..isfrom_bool(v1) && Any..isfrom_int(v2)) then
+    from_int(bool_to_int(Any..as_bool!(v1)) * Any..as_int!(v2))
+  else if (Any..isfrom_int(v1) && Any..isfrom_bool(v2)) then
+    from_int(Any..as_int!(v1) * bool_to_int(Any..as_bool!(v2)))
+  else if (Any..isfrom_bool(v1) && Any..isfrom_float(v2)) then
+    from_float(bool_to_real(Any..as_bool!(v1)) + Any..as_float!(v2))
```
joehendrix: Bug: This should be `*` (multiplication), not `+` (addition). In PMul, `True * 3.5` would incorrectly compute `1.0 + 3.5 = 4.5` instead of `1.0 * 3.5 = 3.5`. Note the symmetric case on line 344 correctly uses `*`.

```suggestion
    from_float(bool_to_real(Any..as_bool!(v1)) * Any..as_float!(v2))
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+//Functions that we provide to Python user
+//to write assertions/contracts about about types of errors
+// /////////////////////////////////////////////////////////////////////////////////////
+
+inline function isTypeError (e: Error) : Any {
+  from_bool (Error..isTypeError(e))
+}
+
+inline function isAttributeError (e: Error) : Any {
+  from_bool (Error..isTypeError(e))
```
shigoel: should be isAttributeError.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+}
+
+inline function isAssertionError (e: Error) : Any {
+  from_bool (Error..isAttributeError(e))
```
shigoel: isAssertionError

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+{
+  if Any..isexception(v1) then v1 else if Any..isexception(v2) then v2
+  else if (Any..isfrom_bool(v1) && Any..isfrom_bool(v2)) then
+    from_int(bool_to_int(Any..as_bool!(v1)) + bool_to_int(Any..as_bool!(v2)))
+  else if (Any..isfrom_bool(v1) && Any..isfrom_int(v2)) then
+    from_int(bool_to_int(Any..as_bool!(v1)) + Any..as_int!(v2))
+  else if (Any..isfrom_int(v1) && Any..isfrom_bool(v2)) then
+    from_int(Any..as_int!(v1) + bool_to_int(Any..as_bool!(v2)))
+  else if (Any..isfrom_int(v1) && Any..isfrom_float(v2)) then
+    from_float(bool_to_real(Any..as_bool!(v1)) + Any..as_float!(v2))
```
shigoel: I expected to see int_to_real here.

Strata/Languages/Python/PythonToLaurel.lean
shigoel, line 1: Let's not use `panic`s in general, and move towards throwing exceptions. The former silently suppress errors...

thanhnguyen-aws: I removed all panics

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.DDM.Elab
+import Strata.DDM.AST
+import Strata.Languages.Core.Verifier
+import StrataTest.Transform.ProcedureInlining
```
MikaelMayer: Production code should not import from `StrataTest`. The old `CorePrelude.lean` does not have this dependency. This creates a build-order issue and couples production code to test infrastructure. Can this import be removed or the needed definition moved to a non-test module?

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+
+namespace Strata
+namespace Python
+
+def pyThonLaurelPrelude :=
```
MikaelMayer: Inconsistent capitalization: `pyThonLaurelPrelude` — the `T` in `Thon` is capitalized mid-word. Should be `pythonLaurelPrelude`.

```suggestion
def pythonLaurelPrelude :=
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+    from_int(bool_to_int(Any..as_bool!(v1)) + Any..as_int!(v2))
+  else if (Any..isfrom_int(v1) && Any..isfrom_bool(v2)) then
+    from_int(Any..as_int!(v1) + bool_to_int(Any..as_bool!(v2)))
+  else if (Any..isfrom_int(v1) && Any..isfrom_float(v2)) then
+    from_float(int_to_real(Any..as_int!(v1)) + Any..as_float!(v2))
+  else if (Any..isfrom_float(v1) && Any..isfrom_bool(v2)) then
+    from_float(Any..as_float!(v1) + bool_to_real(Any..as_bool!(v2)))
+  else if (Any..isfrom_int(v1) && Any..isfrom_int(v2)) then
+    from_int(Any..as_int!(v1) + Any..as_int!(v2))
+  else if (Any..isfrom_int(v1) && Any..isfrom_float(v2)) then
```
MikaelMayer: This `int × float` branch is a duplicate of line 277 and is dead code (the earlier branch matches first). It should be removed.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+
+// /////////////////////////////////////////////////////////////////////////////////////
+// Modelling of Python comparision operations
+// /////////////////////////////////////////////////////////////////////////////////////
+
+function string_lt (s1: string, s2: string) : bool;
+function string_le (s1: string, s2: string) : bool;
+function string_gt (s1: string, s2: string) : bool;
+function string_ge (s1: string, s2: string) : bool;
+function List_lt (l1: ListAny, L2: ListAny): bool;
```
MikaelMayer: Inconsistent parameter naming: `L2` (capital L) here vs `l2` (lowercase) in `List_extend` on line 172. Should be `l2` for consistency.

```suggestion
function List_lt (l1: ListAny, l2: ListAny): bool;
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+  from_bool(Any_to_bool (v1) || Any_to_bool (v2))
+}
+
+
+// /////////////////////////////////////////////////////////////////////////////////////
+// Modelling of other Python operations, currrently unsupported
+// /////////////////////////////////////////////////////////////////////////////////////
+inline function PPow (v1: Any, v2: Any) : Any
+{
+  exception(UnimplementedError ("Pow operator is not suuported"))
```
MikaelMayer: Typo: "suuported" → "supported".

```suggestion
  exception(UnimplementedError ("Pow operator is not supported"))
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+  havoc d;
+  ret := from_datetime(d);
+};
+
+procedure timedelta(days: Any, hours: Any) returns (delta : Any, maybe_except: Error)
+spec{
+  requires [days_type]: Any..isfrom_none(days) || Any..isfrom_int(days);
+  requires [hours_type]: Any..isfrom_none(hours) || Any..isfrom_int(hours);
+  requires [days_pos]: Any..isfrom_int(days) ==> Any..as_int!(days)>=0;
+  requires [hours_type]: Any..isfrom_int(hours) ==> Any..as_int!(hours)>=0;
```
MikaelMayer: Duplicate label `hours_type` — this is the same label as line 608. Should be `hours_pos` to match the `days_pos` pattern.

```suggestion
  requires [hours_pos]: Any..isfrom_int(hours) ==> Any..as_int!(hours)>=0;
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+  assert [assert_name_is_foo]: req_name == from_string("foo");
+  assert [assert_opt_name_none_or_str]: (Any..isfrom_none(opt_name)) || (Any..isfrom_string(opt_name));
+  assert [assert_opt_name_none_or_bar]: (opt_name == from_none()) || (opt_name == from_string("bar"));
+  assume [assume_maybe_except_none]: (Error..isNoError(maybe_except));
+};
+
+procedure print(msg : Any) returns ();
+
+//This is only used to overwrite the Box datatype of Laurel prelude
+//WILL BE REMOVED
```
MikaelMayer, line 641: The `Box` datatype with "WILL BE REMOVED" comment, combined with the `.filter (λ d=> d.name.name != "Box")` calls in StrataMain (lines 451 and 1111), is technical debt. Could you create a tracking issue for removing this?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -56,15 +56,28 @@ inductive UnmodeledFunctionBehavior where
...
 structure TranslationContext where
-  /-- Map from variable names to their Laurel types -/
-  variableTypes : List (String × HighTypeMd) := []
+  variableTypes : List (String × String) := []
   /-- Map from function names to their signatures -/
-  functionSignatures : List (String × Procedure) := []
+  functionSignatures : List PythonFunctionDecl := []
   /-- Map from prelude procedure names to their full signatures -/
   preludeProcedures : List (String × CoreProcedureSignature) := []
+  /-- Map from prelude procedure names to their full signatures -/
```
MikaelMayer: Docstring is copy-pasted from `preludeProcedures`. Should describe what this field actually holds.

```suggestion
  /-- Names of prelude functions (non-procedure callables) -/
  preludeFunctions : List String := []
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -168,19 +184,55 @@ def translateType (ctx : TranslationContext) (typeStr : String) : Except Transla
...
-        if ctx.compositeTypes.any (fun ct => ct.name == typeStr) then
-          .ok (mkHighTypeMd (HighType.UserDefined typeStr))
-        else
-          -- Default to PyAnyType for unknown types
-          .ok (mkCoreType "PyAnyType")
+        -- Map it to a core PyAnyType
+        .ok (mkCoreType "Any")
+
+def AnyTy := mkCoreType "Any"
+def PackageTy := mkCoreType "Package"
```
MikaelMayer: Dead code: `PackageTy`, `intTy`, `boolTy`, `strTy` (lines 191-194), `HighTypeToString` (line 202), and `OptionNone` (line 220) are all defined but never used. They should be removed.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -275,25 +305,17 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+      | .Add _ => .ok "PAdd"
+      | .Sub _ => .ok "PSub"
+      | .Mult _ => .ok "PMul"
+      | .FloorDiv _ => .ok "PFloorDiv"  -- Python // maps to Laurel Div
+      | .Mod _ => .ok "PMod"
+      | .BitAnd _ => .ok "PBitAnd"
       -- Unsupported for now
       | _ => throw (.unsupportedConstruct s!"Binary operator not yet supported: {repr op}" (toString (repr e)))
-    return mkStmtExprMd (StmtExpr.PrimitiveOp laurelOp [leftExpr, rightExpr])
+    return mkStmtExprMd (StmtExpr.StaticCall preludeOpnames [leftExpr, rightExpr])
```
MikaelMayer, line 306: The `NotIn` comparison operator was previously handled (returning `.Hole` for soundness) but has been removed. It will now throw an `unsupportedConstruct` error. Similarly, `.Div` (true division), `.BitOr`, and `.BitXor` were previously handled but are now removed. These are regressions — if they're intentionally unsupported, they should still return `.Hole` or throw a clear error rather than falling through to a generic catch-all.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -410,53 +430,214 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+    let kws_and_exprs ← kwords.mapM (translateDictKWords ctx)
+    let ret := DictStrAny_mk kws_and_exprs
+    return ret
+
+partial def removePosargsFromKwargs (kwords : List (Python.keyword SourceRange)) (funcDecl: PythonFunctionDecl) : List (Python.keyword SourceRange) :=
+  kwords.filter (λ kw => match kw with
+    | .mk_keyword _ name _ =>
+      match name.val with
+        | some n => n.val ∉ funcDecl.args.unzip.fst
+        | none => True)
```
MikaelMayer: `True` (capital T) is `Prop`, not `Bool`. In a `List.filter` context this should be `true`.

```suggestion
        | none => true)
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -410,53 +430,214 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+    else [trans_kwords]
+  match f with
+  | .Name  _ _ _ =>  return mkStmtExprMd (StmtExpr.StaticCall funcName (trans_args ++ trans_kwords_exprs))
+  | .Attribute _ val attr _ =>
+      let target_trans ← translateExpr ctx val
+      if opt_firstarg.isSome then
+        return mkStmtExprMd (StmtExpr.InstanceCall target_trans attr.val (trans_args ++ trans_kwords_exprs))
+      else
+        return mkStmtExprMd (StmtExpr.StaticCall funcName (trans_args ++ trans_kwords_exprs))
+  | _ =>  throw (.unsupportedConstruct "Invalidcall construct" (toString (repr f)))
```
MikaelMayer: Typo: "Invalidcall" → "Invalid call".

```suggestion
  | _ =>  throw (.unsupportedConstruct "Invalid call construct" (toString (repr f)))
```

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -120,7 +120,7 @@ def SemanticModel.isFunction (model: SemanticModel) (id: Identifier): Bool :=
     -- expressions; any real errors will be caught during Core type checking.
     -- Make an exception for 'test_helper_procedure' since it's a procedure
     -- We will remove this hack when we enable the Python through Laurel pipeline to correctly resolve
-    id.text != "test_helper_procedure"
+    id.text ∉ ["test_helper_procedure", "datetime_now", "datetime_date"]
```
MikaelMayer: This hardcoded exclusion list is growing (`test_helper_procedure`, `datetime_now`, `datetime_date`). The comment says "We will remove this hack" — the `preludeFunctions` list added to `TranslateState` in `LaurelToCoreTranslator.lean` seems like it should be able to replace this hack. Is there a reason this list still needs to grow rather than being resolved by the prelude function list?

StrataTest/Languages/Python/run_py_analyze.sh
```diff
@@ -9,11 +9,11 @@ mode="${1:-core}"
 if [ "$mode" = "laurel" ]; then
     command="pyAnalyzeLaurel"
     expected_dir="expected_laurel"
-    skip_tests="test_datetime"
+    skip_tests=""
 else
     command="pyAnalyze"
     expected_dir="expected_non_laurel"
-    skip_tests=""
+    skip_tests="test_class_field_use"
```
MikaelMayer: The non-laurel path now skips `test_class_field_use`, which was previously not skipped. Is this an intentional regression or a temporary workaround? If temporary, it should be tracked.

Strata/Languages/Laurel/Resolution.lean
thanhnguyen-aws: Temporarily, we use those hacks to pass the CI. We will reorganize the Python and Laurel prelude with this PR: https://github.com/strata-org/Strata/pull/506. However, this PR must me merged first because it changes the PythonLaurelPrelude file.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
thanhnguyen-aws: I created this tracking issue https://github.com/strata-org/Strata/issues/533.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -275,25 +289,19 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
-      | .Mod _ => .ok Operation.Mod
-      | .Div _ => .ok Operation.Div  -- Python / (true division)
-      | .BitAnd _ => .ok Operation.And  -- Bitwise & - abstract as logical And
-      | .BitOr _ => .ok Operation.Or    -- Bitwise | - abstract as logical Or
-      | .BitXor _ => .ok Operation.Neq  -- Bitwise ^ - abstract as inequality
+      | .Add _ => .ok "PAdd"
+      | .Sub _ => .ok "PSub"
+      | .Mult _ => .ok "PMul"
+      | .FloorDiv _ => .ok "PFloorDiv"  -- Python // maps to Laurel Div
+      | .Mod _ => .ok "PMod"
```
MikaelMayer, line 299: Regression: The `.Div _` case (Python's `/` true division operator) was previously handled (`| .Div _ => .ok Operation.Div`) but is now missing entirely. It will fall through to the catch-all error. Please add a `| .Div _ => .ok "PDiv"` case (and define `PDiv` in the prelude), or at minimum map it to `.Hole` with a comment explaining why.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -275,25 +289,19 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+      | .BitAnd _ => .ok "PBitAnd"
```
MikaelMayer: `PBitAnd` is referenced here but not defined in `PythonLaurelCorePrelude.lean`. This will cause a runtime error when a Python program uses `&`. Either define `PBitAnd` in the prelude or map this to `.Hole` like `BitOr`/`BitXor`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -275,25 +289,19 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+      | .BitOr _ => return mkStmtExprMd .Hole
```
MikaelMayer, line 301: Regression: `.BitOr` was previously mapped to `Operation.Or` (a sound abstraction). Now it returns `.Hole`, which loses information. Same for `.BitXor` on line 301 (was `Operation.Neq`). If the prelude doesn't model these yet, at least add a TODO comment explaining the regression.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -168,19 +184,39 @@ def translateType (ctx : TranslationContext) (typeStr : String) : Except Transla
...
+        -- Map it to a core PyAnyType
+        .ok (mkCoreType "Any")
+
+def AnyTy := mkCoreType "Any"
+def strToAny (s: String) := mkStmtExprMd (.StaticCall "from_string" [mkStmtExprMd (StmtExpr.LiteralString s)])
+def intToAny (i: Int) := mkStmtExprMd (.StaticCall "from_int" [mkStmtExprMd (StmtExpr.LiteralInt i)])
+def boolToAny (b: Bool) := mkStmtExprMd (.StaticCall "from_bool" [mkStmtExprMd (StmtExpr.LiteralBool b)])
+def AnyNone := mkStmtExprMd (.StaticCall "from_none" [])
+def Any_to_bool (b: StmtExprMd) := mkStmtExprMd (.StaticCall "Any_to_bool" [b])
+def FreeVar (name: String) := mkStmtExprMd (StmtExpr.Identifier name)
```
MikaelMayer: Dead code: `FreeVar` is defined but never used. Same for `mkNoneForType` (line 199) and `OptionNone` (line 204). Please remove them.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1037,20 +1293,27 @@ def pythonToLaurel (prelude: Core.Program)
...
       types := compositeTypes.map TypeDefinition.Composite
       constants := []
     }
 
-    return program
+    return (program, ctx)
 
   | _ => throw (.internalError "Expected Module")
 
+def pythonToLaurelWithFuncSigature (prelude: Core.Program) (funsig : Python.Command SourceRange) (pyModule : Python.Command SourceRange)
```
MikaelMayer: Two issues:
1. Typo in name: `pythonToLaurelWithFuncSigature` → `pythonToLaurelWithFuncSignature`
2. This function is defined but never called anywhere in the codebase. If it's intended for future use, add a docstring explaining its purpose. Otherwise, remove it.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -834,6 +1057,22 @@ def extractPreludeProcedures (prelude : Core.Program) : List (String × CoreProc
...
     | none => none
 
+def preludeSignatureToPythonFunctionDecl (prelude : Core.Program) : List PythonFunctionDecl :=
+  prelude.decls.filterMap fun decl =>
+    match Core.Program.Procedure.find? prelude decl.name with
+    | some proc =>
+      let inputTypes := proc.header.inputs.values.map getTypeName
+      let inputnames := proc.header.inputs.keys.map (λ n => n.name)
+      let outputtypes := proc.header.outputs.values.map getTypeName
+      let noneexpr : Python.expr SourceRange := .Constant default (.ConNone default) default
```
MikaelMayer, line 1067: Every prelude procedure argument gets `noneexpr` (Python `None`) as its default value. This means the kwargs-filling logic in `combinePositionalAndKeywordArgs` would use `None` for any missing argument to a prelude procedure, which is semantically incorrect for arguments that don't have a default. Consider using `none` (no default) instead of `some noneexpr` for prelude args:
```
args:= (inputnames.zip inputTypes).map (λ(n,t) => (n,t, none))
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,682 @@
...
+// Modelling of other Python operations, currrently unsupported
+// /////////////////////////////////////////////////////////////////////////////////////
+inline function PPow (v1: Any, v2: Any) : Any
+{
+  exception(UnimplementedError ("Pow operator is not supported"))
+}
+
+inline function PMod (v1: Any, v2: Any) : Any
+{
+  exception(UnimplementedError ("Mod operator is not suuported"))
```
MikaelMayer: Typo: "suuported" → "supported"
```suggestion
  exception(UnimplementedError ("Mod operator is not supported"))
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,682 @@
...
+// /////////////////////////////////////////////////////////////////////////////////////
+// Modelling of Python comparision operations
+// /////////////////////////////////////////////////////////////////////////////////////
+
+function string_lt (s1: string, s2: string) : bool;
+function string_le (s1: string, s2: string) : bool;
+function string_gt (s1: string, s2: string) : bool;
+function string_ge (s1: string, s2: string) : bool;
+function List_lt (l1: ListAny, l2: ListAny): bool;
+function List_le (l1: ListAny, L2: ListAny): bool;
```
MikaelMayer: Inconsistent parameter naming: `L2` (uppercase) in `List_le`, `List_gt`, `List_ge` vs `l2` (lowercase) in `List_lt`. Should all be `l2`.
```suggestion
function List_le (l1: ListAny, l2: ListAny): bool;
function List_gt (l1: ListAny, l2: ListAny): bool;
function List_ge (l1: ListAny, l2: ListAny): bool;
```

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,682 @@
...
+def getDatatypeFunctions (decls: List Core.Decl) : List String :=
+  match decls with
+  | [] => []
+  | decl::t => match h: decl.kind with
+        |.type =>
+          let typedec := decl.getTypeDecl (by simp_all)
+          match typedec with
+          | .data dtypes =>
+            let constructors := dtypes.flatMap (λ t => t.constrs.map (λ c => c.name.name))
+            let destructors := dtypes.flatMap (λ t => (t.constrs.flatMap (λ c => c.args.map (fun (n,y) => t.name ++ ".." ++ n.name))))
```
MikaelMayer: Unused variable `y` in the lambda. Use `_` instead:
```suggestion
            let destructors := dtypes.flatMap (λ t => (t.constrs.flatMap (λ c => c.args.map (fun (n, _) => t.name ++ ".." ++ n.name))))
```

Strata/Languages/Laurel/Resolution.lean
MikaelMayer: This hardcoded exclusion list is growing (`test_helper_procedure`, `datetime_now`, `datetime_date`). The comment says "We will remove this hack" — is there a tracking issue for this? Each new prelude procedure that's not a function will need to be added here manually, which is fragile.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,682 @@
...
+            let testers := dtypes.flatMap (λ t => t.constrs.map (λ c => c.testerName))
+            constructors ++ destructors ++ testers ++ getDatatypeFunctions t
+          | _ => getDatatypeFunctions t
+        | _ => getDatatypeFunctions t
+
+
+def get_preludeFunctions (prelude: Core.Program) : List String := (getFunctions prelude.decls) ++ (getDatatypeFunctions prelude.decls)
```
MikaelMayer: Naming: `get_preludeFunctions` and `CorePrelude_functions` (line 678) use snake_case, which is inconsistent with Lean conventions. Consider `getPreludeFunctions` and `corePreludeFunctions`.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -185,7 +186,7 @@ def translateExpr (expr : StmtExprMd)
       return .ite () bcond bthen belse
   | .StaticCall callee args =>
       -- In a pure context, only Core functions (not procedures) are allowed
-      if isPureContext && !model.isFunction callee then
+      if isPureContext && !((model.isFunction callee) || (callee.text ∈ s.preludeFunctions)) then
```
MikaelMayer:
```suggestion
      if isPureContext && !model.isFunction callee && callee.text ∉ s.preludeFunctions then
```
That helps removing parentheses (this is \notin)

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -281,7 +291,7 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
       match initializer with
       | some (⟨ .StaticCall callee args, callMd⟩) =>
           -- Check if this is a function or a procedure call
-          if model.isFunction callee then
+          if (model.isFunction callee) || (callee.text ∈ s.preludeFunctions) then
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -291,7 +301,14 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
             let defaultExpr := defaultExprForType model ty
             let initStmt := Core.Statement.init ident coreType (some defaultExpr) md
             let callStmt := Core.Statement.call [ident] callee.text coreArgs md
-            return [initStmt, callStmt]
+            return ([initStmt, callStmt])
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -291,7 +301,14 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
...
+      | some (⟨ .InstanceCall .., _⟩) =>
+          -- Instance method call as initializer: var name := target.method(args)
+          -- Havoc the result since instance methods may be on unmodeled types
+          let defaultExpr := defaultExprForType model ty
+          let initStmt := Core.Statement.init ident coreType (some defaultExpr) md
+          let havocStmt := Core.Statement.havoc ident md
+          return ([initStmt, havocStmt])
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -305,14 +322,17 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
           -- Check if RHS is a procedure call (not a function)
           match value.val with
           | .StaticCall callee args =>
-              if model.isFunction callee then
+              if (model.isFunction callee) || (callee.text ∈ s.preludeFunctions) then
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -305,14 +322,17 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
...
                 -- Functions are translated as expressions
                 let coreExpr ← translateExpr value
                 return [Core.Statement.set ident coreExpr md]
               else
                 -- Procedure calls need to be translated as call statements
                 let coreArgs ← args.mapM (fun a => translateExpr a)
-                return [Core.Statement.call [ident] callee.text coreArgs md]
+                return ([Core.Statement.call [ident] callee.text coreArgs md])
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -305,14 +322,17 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
...
+          | .InstanceCall .. =>
+              -- Instance method call: havoc the target variable
+              return ([Core.Statement.havoc ident md])
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -326,7 +346,14 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
                 match t.val with
                 | .Identifier name => some (⟨name.text, ()⟩)
                 | _ => none
-              return [Core.Statement.call lhsIdents callee.text coreArgs value.md]
+              return ([Core.Statement.call lhsIdents callee.text coreArgs value.md])
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -338,12 +365,15 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
       return [Imperative.Stmt.ite bcond bthen belse .empty]
   | .StaticCall callee args =>
       -- Check if this is a function or procedure
-      if model.isFunction callee then
+      if (model.isFunction callee) || (callee.text ∈ s.preludeFunctions) then
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -338,12 +365,15 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
...
         -- Functions as statements have no effect (shouldn't happen in well-formed programs)
         return []
       else
         let coreArgs ← args.mapM (fun a => translateExpr a)
-        return [Core.Statement.call [] callee.text coreArgs md]
+        return ([Core.Statement.call [] callee.text coreArgs md])
```
MikaelMayer: Parentheses not needed.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,684 @@
...
+inline function PNEq (v: Any, v': Any) : Any {
+  from_bool(normalize_any(v) != normalize_any (v'))
+}
+
+// /////////////////////////////////////////////////////////////////////////////////////
+// Modelling of Python Boolean operations And and Or
+// /////////////////////////////////////////////////////////////////////////////////////
+
+inline function PAnd (v1: Any, v2: Any) : Any
+  requires Any..isfrom_bool(v1) && Any..isfrom_bool(v2);
```
MikaelMayer: `and` in python evaluates the truthfullness of the first argument, and if false returns it, otherwise returns the second argument. There is no coercion to bool needed.
Same for `or` but it returns the first argument iff it is truthy.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,684 @@
...
+  }
+};
+
+procedure datetime_now() returns (ret: Any)
+spec {
+  ensures [ret_type]: Any..isfrom_datetime(ret);
+}
+{
+  var d: int;
+  havoc d;
```
MikaelMayer:
```suggestion
```
Havoc not needed when a variable does not have a rhs

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,684 @@
...
+
+procedure datetime_date(d: Any) returns (ret: Any, error: Error)
+spec {
+  requires [d_type]: Any..isfrom_datetime(d);
+  ensures [ret_type]: Any..isfrom_datetime(ret) && Any..as_datetime!(ret) <= Any..as_datetime!(d);
+}
+{
+  var timedt: int;
+  if (Any..isfrom_datetime(d)) {
+    havoc timedt;
```
MikaelMayer: Not needed
```suggestion
```

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: This one is only applied for the procedures that is written in Core directly (those testing procedures in the CorePrelude), while combinePositionalAndKeywordArgs is applied for procedures that is written in Python. For those testing procedures (such as  "test_helper_procedure", "datetime_date"), there is no way to express the default value but they are assumed to be None if the input argument is  missing. In the future, we may consider rewriting those testing procedures in Python instead of Core.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I added a TODO for those BitVector operations. I will add them back when we add BitVector to Any type. For this PR, all user-defined variables are translated as Any type variables, so we have to replace all operations with those of Any type.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
thanhnguyen-aws: This coercion is needed because of 2 reasons: 1) All variables and expressions are now of `Any` type, so if their actual type is `bool` it needs to be converted back to `bool` (by a destructor of Core datatype). 2/  If the expressions' types are not `bool`, Python can automatically converted them to `bool` by some conventions (such as an empty List is converted into `false`), so we need to model that behavior.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -292,6 +302,13 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
             let initStmt := Core.Statement.init ident coreType (some defaultExpr) md
             let callStmt := Core.Statement.call [ident] callee.text coreArgs md
             return [initStmt, callStmt]
+      | some (⟨ .InstanceCall .., _⟩) =>
+          -- Instance method call as initializer: var name := target.method(args)
+          -- Havoc the result since instance methods may be on unmodeled types
+          let defaultExpr := defaultExprForType model ty
+          let initStmt := Core.Statement.init ident coreType (some defaultExpr) md
+          let havocStmt := Core.Statement.havoc ident md
+          return [initStmt, havocStmt]
```
MikaelMayer:
```suggestion
          let initStmt := Core.Statement.init ident coreType none md
          return [initStmt]
```

I just realized you don't need an havoc here either since no RHS means havoc.

StrataTest/Languages/Python/tests/cbmc_expected.txt
```diff
@@ -4,11 +4,11 @@
 # FAIL = VERIFICATION FAILED expected (known limitations)
 # SKIP = pipeline error expected (test not yet supported)
 
-test_arithmetic.py.ion          PASS
-test_comparisons.py.ion         PASS
-test_control_flow.py.ion        PASS
+test_arithmetic.py.ion          FAIL
```
MikaelMayer, line 7: Why are these tests failing suddenly?

StrataTest/Languages/Python/run_py_analyze_sarif.py
```diff
@@ -21,7 +21,7 @@
 
 BOTH_SKIP = {"test_foo_client_folder", "test_invalid_client_type", "test_unsupported_config"}
 SKIP_TESTS = BOTH_SKIP | {"test_class_field_use"}
-SKIP_TESTS_LAUREL = BOTH_SKIP | {"test_datetime"}
+SKIP_TESTS_LAUREL = BOTH_SKIP
```
MikaelMayer, line 24: Nice, one more test going through !

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1037,20 +1283,30 @@ def pythonToLaurel (prelude: Core.Program)
...
 
-    return program
+    return (program, ctx)
 
   | _ => throw (.internalError "Expected Module")
 
+/- This function is used to add the external functions signatures to the TranslationContext
+TODO: Rewrite testing procedures that is currently in PythonLaurelCorePrelude in Python
+ -/
+def pythonToLaurelWithFuncSignature (prelude: Core.Program) (funsig : Python.Command SourceRange) (pyModule : Python.Command SourceRange)
```
MikaelMayer: Is this function used somewhere? If not consider remove this or adding a test so that we can capture what it does.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -466,109 +625,103 @@ Translate Python statements to Laurel StmtExpr nodes.
...
+def withException (ctx : TranslationContext) (funcname: String) : Bool :=
+  if funcname ∈ ctx.preludeFunctions then false else
+  match ctx.preludeProcedures.lookup funcname with
+  | some sig => sig.outputs.length > 0 && sig.outputs.getLast! == "Error"
+  | _ => false
+
+def maybeExceptVar := mkStmtExprMd (.Identifier "maybe_except")
+def nullcall_var := mkStmtExprMd (.Identifier "nullcall_ret")
+
+partial def translateAssign  (ctx : TranslationContext)
```
MikaelMayer, line 647: I'm not able to check mentally if any of this code is correct. Can you please confirm this code is well tested?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -56,15 +56,28 @@ inductive UnmodeledFunctionBehavior where
...
+  ret : Option String
+deriving Repr, Inhabited
+
 structure TranslationContext where
-  /-- Map from variable names to their Laurel types -/
-  variableTypes : List (String × HighTypeMd) := []
+  variableTypes : List (String × String) := []
   /-- Map from function names to their signatures -/
-  functionSignatures : List (String × Procedure) := []
+  functionSignatures : List PythonFunctionDecl := []
```
MikaelMayer, line 70: The comment of this line indicates it's a map but it's no longer a map?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -410,53 +409,214 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+
+partial def reMapFunctionName (ctx: TranslationContext) (fname: String) : String :=
+  if fname ∈ ctx.userClasses then
+    fname ++ "___init__"
+  else
+    match fname with
+    | "str" => "to_string_any"
+    | "int" => "to_int_any"
+    | "len" => "Any_len_to_Any"
+    | _ => fname
```
MikaelMayer, line 437: I'm not sure about this passthrough case, would it be worth it to sanitize it? Or return an exception?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -410,53 +409,214 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+  let (root, _):= getListAttributes expr
+  match root with
+  | .Name _ n _ => n.val ∉ ctx.variableTypes.unzip.fst
+  | _ => false
+
+partial def inferExprType (ctx : TranslationContext) (e: Python.expr SourceRange) : Except TranslationError String := do
+  match e with
+  -- Integer literals
+  | .Constant _ (.ConPos _ _) _
+  | .Constant _ (.ConNeg _ _) _ => return "int"
```
MikaelMayer: It would be worth it, every time there is a hard code `"int"`, "`string`", `"bool"` or `"Any"` that you instead refer to a programmatic name such as `PyLauType.Int` for example. It will make code maintenance much easier.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -410,53 +409,214 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+partial def getListAttributes (expr: Python.expr SourceRange): (Python.expr SourceRange) × List String :=
+  match expr with
+  | .Attribute _ v attr _ =>
+      let ret := (getListAttributes v)
+      (ret.fst , ret.snd ++ [attr.val])
+  | _ => (expr, [])
+
+partial def reMapFunctionName (ctx: TranslationContext) (fname: String) : String :=
+  if fname ∈ ctx.userClasses then
+    fname ++ "___init__"
```
MikaelMayer: Could you please avoid the name collision by using a delimiter that is not supported by python, if that exists?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -410,53 +409,214 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+  match func with
+    | .Name _ n _ => return (reMapFunctionName ctx n.val, none , false)
+    | .Attribute _ v attr _ =>
+        let callerTy ←  inferExprType ctx v
+        let callname := attr.val
+        if isPackage ctx v then
+          return (pyExprToString func, none, false)
+        else
+        if callerTy == "Any" then
+          return ("AnyTyInstance" ++ "_" ++ callname, some v, true)
```
MikaelMayer: If there is no possible name collision with user-defined names, mark that comment as resolved. Otherwise, it is possible to use a character splitter that is not valid python? For example, `@` or `#`.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,688 @@
...
+          | _ => getDatatypeFunctions t
+        | _ => getDatatypeFunctions t
+
+
+def getPreludeFunctions (prelude: Core.Program) : List String := (getFunctions prelude.decls) ++ (getDatatypeFunctions prelude.decls)
+
+def corePreludeFunctions := getPreludeFunctions Core.PythonLaurelPrelude
+
+def getProcedures (decls: List Core.Decl) : List String :=
+  match decls with
```
MikaelMayer: Could you use a single-line combinator? I don't know the equivalent of flatMap in Lean but it looks like it would be more maintainable.
It could help for getDatatypes and getFunctions above

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: This function is mainly for renaming the a couple of Python functions which may conflicts with names of some built-in functions or types in Core. Otherwise, the function names stay the same.

StrataTest/Languages/Python/tests/cbmc_expected.txt
thanhnguyen-aws: I think @tautschnig will handle it in a different PR.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: Yes, I just move the code from `translateStmt` to a separate function.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -75,6 +75,7 @@ structure TranslateState where
   nextId : Nat := 1
   /-- Constants known to the program (field constants, etc.) -/
   model : SemanticModel
+  preludeFunctions : List String
```
keyboardDrummer: Now that `model : SemanticModel` is available, we shouldn't need `preludeFunctions` any more. You should add functions to Laurel with an "external" body when you want to use functions from Core. Look at `CoreDefinitionsForLaurel.lean` for examples.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -233,7 +234,16 @@ def translateExpr (expr : StmtExprMd)
   | .ContractOf _ _ => panic "contractOf expression not implemented"
   | .Abstract => panic "abstract expression not implemented"
   | .All => panic "all expression not implemented"
-  | .InstanceCall _ _ _ => panic "InstanceCall not implemented"
+  | .InstanceCall target callee args =>
+      if isPureContext then
+        disallowed expr "calls to instance methods are not supported in functions or contracts"
```
keyboardDrummer: Why not?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -233,7 +234,16 @@ def translateExpr (expr : StmtExprMd)
   | .ContractOf _ _ => panic "contractOf expression not implemented"
   | .Abstract => panic "abstract expression not implemented"
   | .All => panic "all expression not implemented"
-  | .InstanceCall _ _ _ => panic "InstanceCall not implemented"
+  | .InstanceCall target callee args =>
```
keyboardDrummer: Since this is a fairly large PR, considering extracting support for instance calls to a separate PR. This PR seems to add support for them but is missing tests and probably some more consideration. We can also delay that until later so I'll leave it up to you, but it might help with debugging issues to have separate tests for instance calls.

Also, why are you using instance calls? It seems like they're not supporting dynamic dispatch at the moment, but then I wonder what their use is.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -233,7 +234,16 @@ def translateExpr (expr : StmtExprMd)
...
+      if isPureContext then
+        disallowed expr "calls to instance methods are not supported in functions or contracts"
+      else
+        let fnOp : Core.Expression.Expr := .op () ⟨callee.text, ()⟩ none
```
keyboardDrummer: Don't we need to check whether `callee` is functional? If not, I think we need to emit an error similar to what we're doing for static call.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -327,6 +346,13 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
                 | .Identifier name => some (⟨name.text, ()⟩)
                 | _ => none
               return [Core.Statement.call lhsIdents callee.text coreArgs value.md]
+          | .InstanceCall .. =>
+              -- Instance method call: havoc all target variables
```
keyboardDrummer, line 357: Why the havoc instead of generating a Core call? I'm reluctant to merge this because we'll have to change it when adding support for instance calls and this PR might complicate that.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -454,7 +483,7 @@ private def isPureExpr(expr: StmtExprMd): Bool :=
   | .This => panic s!"isPureExpr not implemented for This"
   | .AsType .. => panic s!"isPureExpr not supported for AsType"
   | .IsType .. => panic s!"isPureExpr not supported for IsType"
-  | .InstanceCall .. => panic s!"isPureExpr not implemented for InstanceCall"
+  | .InstanceCall .. => false
```
keyboardDrummer: You can call an instance function right?

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -0,0 +1,685 @@
...
+{
+  assert [assert_name_is_foo]: req_name == from_string("foo");
+  assert [assert_opt_name_none_or_str]: (Any..isfrom_none(opt_name)) || (Any..isfrom_string(opt_name));
+  assert [assert_opt_name_none_or_bar]: (opt_name == from_none()) || (opt_name == from_string("bar"));
+  assume [assume_maybe_except_none]: (Error..isNoError(maybe_except));
+};
+
+procedure print(msg : Any) returns ();
+
+//This is only used to overwrite the Box datatype of Laurel prelude
```
keyboardDrummer, line 640: Could you point me to the code doing the overriding? I couldn't find it. I see code removing Box datatypes, but that's it.

thanhnguyen-aws: It is for overriding this code in Laurel/CorePrelude.lean:
```
// Tagged union for field values
datatype Box () {
  BoxInt(intVal: int),
  BoxBool(boolVal: bool),
  BoxFloat64(float64Val: real),
  BoxComposite(compositeVal: Composite)
};
```

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
andrewmwells-amazon: Sounds good. Saving InstanceCall for a future PR.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
keyboardDrummer: That I knew, but could you point me to the code doing the overrriding?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
andrewmwells-amazon: Sorry, I'm having trouble imagining how this would complicate InstanceCall support. Can you explain?

keyboardDrummer: If the current Python->Laurel translation depends on the current Laurel->Core translation of InstanceCall, which we'll need to change, then it'll be harder to change since we might need to update Python->Laurel as well. I doesn't look like changing InstanceCall to what it should be will break Python things, but there's no guarantee so if we can avoid it that'd be better.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
andrewmwells-amazon: Reverted.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
andrewmwells-amazon: Let's save this for a follow-up PR.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Well I won't block on this if you want to merge this ASAP :) but it doesn't have my preference. I'll look into adding support for InstanceCall tomorrow.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
thanhnguyen-aws: It is in StrataMain.lean, line 453 and 1123

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
andrewmwells-amazon: I still don't see how a more precise implementation of `InstanceCall` for `Laurel->Core` could require changes in `Python->Laurel` if `Python->Laurel->Core` previously worked with the call abstracted as a `havoc` (ignoring modifications of `self`). I suspect we could move the abstraction into `Python->Laurel` if you feel strongly, but then I think we'd have to add significantly more complexity there and I don't think the current version will be hard to replace.

shigoel (approved):

Strata/Languages/Python/PythonLaurelCorePrelude.lean
keyboardDrummer: But on those lines it's just filtering out `Box` definitions. Where is the final Box definition that you're using?

Issue #491, opened 2026-02-26
Precedence issue with if-then-else in Strata Core


joscoh:
The following fails to verify:
```
def ifElsePlusPgm : Program :=
#strata
program Core;

function myFunc (b : bool, x : int) : int {
  if b then 10 else 1 + x
}

procedure Test() returns ()
spec {
  ensures true;
}
{
  assert [trueCase]: myFunc(true, 5) == 10;
};
#end

#eval verify ifElsePlusPgm (options := .quiet)
```
The reason is that it is parsed as `(if b then 10 else 1) + x` instead of the expected `if b then 10 else (1 + x)`.

PR #492, opened 2026-02-26
Fix if-then-else precedence issue

joscoh:
*Issue #, if available:* #491

*Description of changes:* Fix if-then-else precedence issue that led to unexpected parsing of expressions involving arithmetic and if-then-else


atomb (approved):

MikaelMayer (approved):
I wouldn't have added the new test case since it's already tested but it's good to have.

PR #493, opened 2026-02-27
Add datatype destructors with precondition checks

joscoh:
*Issue #, if available:*

*Description of changes:* Each datatype now generates 2 varieties of constructors: "safe" versions which check the precondition that instance of the datatype belongs to the correct constructor, and "unsafe" versions which do not (preserve existing behavior). The default destructors (e.g. `List..head`) are now safe, and the unsafe versions have an additional `!` at the end. This PR updates existing datatype tests to use both varieties. Laurel and Python use unsafe versions in its translation to Core to keep existing behavior.

### DIFF
--- a/StrataTest/Languages/Python/expected_laurel/test_class_decl.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_class_decl.expected
@@ -7,4 +7,3 @@ assert_name_is_foo: ✅ pass (at byte 11081)
 assert_opt_name_none_or_str: ✅ pass (at byte 11131)
 assert_opt_name_none_or_bar: ✅ pass (at byte 11278)
 ensures_maybe_except_none: ✅ pass (at byte 10984)
-
--- a/StrataTest/Languages/Python/expected_laurel/test_class_field_use.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_class_field_use.expected
@@ -8,4 +8,3 @@ assert_opt_name_none_or_str: ✅ pass (at byte 11131)
 assert_opt_name_none_or_bar: ✅ pass (at byte 11278)
 ensures_maybe_except_none: ✅ pass (at byte 10984)
 assert(285): 🟡 unknown (at line 14, col 4)
-


Strata/DL/Lambda/TypeFactory.lean
```diff
@@ -484,22 +484,40 @@ def destructorConcreteEval {T: LExprParams} [BEq T.Identifier] (d: LDatatype T.I
...
-Generate destructor functions for a constructor, which extract the
-constructor components, e.g.
-`List..head (Cons h t) = h`
-`List..tail (Cons h t) = t`
-These functions are partial, `List..head Nil` is undefined.
+Generate destructor functions with a precondition that the corresponding tester holds, e.g.
+`List..head(x)` requires `List..isCons(x)`
 -/
-def destructorFuncs {T} [BEq T.Identifier] [Inhabited T.IDMeta]  (d: LDatatype T.IDMeta) (c: LConstr T.IDMeta) : List (LFunc T) :=
+def destructorFuncs {T} [BEq T.Identifier] [Inhabited T.IDMeta] [Inhabited T.Metadata] (d: LDatatype T.IDMeta) (c: LConstr T.IDMeta) : List (LFunc T) :=
```
MikaelMayer: `destructorFuncs` and `unsafeDestructorFuncs` share everything except the function name and the presence of `preconditions`. Consider extracting a helper to avoid the duplication:

```lean
private def mkDestructorFunc {T} [BEq T.Identifier] [Inhabited T.IDMeta] [Inhabited T.Metadata]
    (d : LDatatype T.IDMeta) (c : LConstr T.IDMeta) (i : Nat)
    (name : Identifier T.IDMeta) (ty : T.mono) (safe : Bool) : LFunc T :=
  let arg := genArgName
  let argExpr : LExpr T.mono := .fvar default arg .none
  let testerExpr : LExpr T.mono := .app default (.op default c.testerName .none) argExpr
  { name := if safe then destructorFuncName d name else unsafeDestructorFuncName d name,
    typeArgs := d.typeArgs,
    inputs := [(arg, dataDefault d)],
    output := ty,
    concreteEval := some (fun _ => destructorConcreteEval d c i),
    attr := #[.evalIfConstr 0],
    preconditions := if safe then [⟨testerExpr, default⟩] else [] }

def destructorFuncs ... :=
  c.args.mapIdx (fun i (name, ty) => mkDestructorFunc d c i name ty true)

def unsafeDestructorFuncs ... :=
  c.args.mapIdx (fun i (name, ty) => mkDestructorFunc d c i name ty false)
```

MikaelMayer (commented):
The PR is well-motivated and the implementation is correct. The main actionable issues are duplication in `destructorFuncs`/`unsafeDestructorFuncs` and in `filterDatatypeDecls`. The `!`-stripping in SMTEncoder is a minor fragility. The removal of `safeHead`/`safeTail` wrappers from `PrecondElim` tests nicely demonstrates the user-facing benefit.

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -1541,7 +1543,12 @@ def filterDatatypeDecls (ldatatype : LDatatype Core.Visibility) (funcDecls : Lis
     | .func f => fieldAccessorNames.contains f.name.name
     | _ => false
 
-  (constructorDecls, testerDecls, fieldAccessorDecls)
+  let unsafeFieldAccessorDecls := funcDecls.filter fun decl =>
```
MikaelMayer: The four `funcDecls.filter` blocks in `filterDatatypeDecls` are identical except for the name list. A local helper would reduce the repetition:

```lean
let filterByNames (names : List String) := funcDecls.filter fun decl =>
  match decl with | .func f => names.contains f.name.name | _ => false
```

Then each block becomes a one-liner.

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -341,10 +342,11 @@ partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Conte
     let adtApp := fun (args : List Term) (retty : TermType) =>
         /-
         Note: testers use constructor, translated in `Op.mkName` to is-foo
-        Selectors use full function name (Datatype..fieldName) for uniqueness
+        Selectors use full function name (Datatype..fieldName) for uniqueness.
+        Unsafe selectors (e.g. List..head!) use the safe name (List..head).
         -/
         let name := match kind with
-          | .selector => fn.name
+          | .selector => if fn.name.endsWith "!" then (fn.name.dropEnd 1).toString else fn.name
```
MikaelMayer: The `!`-stripping relies on the naming convention being stable. If the convention changes, this silently produces wrong SMT names. Consider using `unsafeDestructorFuncName` / `destructorFuncName` directly, or at least extracting a named helper `stripUnsafeSuffix` so the coupling is explicit.

PR #482, opened 2026-02-25
Update SMT.Encoder to avoid manual query formatting in String, prove termination of `encodeTerm`

aqjune-aws:
Currently, Strata's SMT.Encoder (1) rewrites the asserting SMT term into the A-normal form, and (2) emits the statements as concrete SMT queries with multiple `define-fun`s in String.

However, to prove that the A-normal form translation in SMT.Encoder is correct, we need to split out step (2): SMT Term -> String conversion.
Since SMT.Solver was already partially doing this task as well, this patch refactors SMT.Encoder to purely work on SMT.Term/SMT.TermTypes, and let SMT.Solver do the whole SMT Term -> String formatting.
Another feature that was delegated from SMT.Encoder to SMT.Solver, as an effect, is cacheing the translated results of SMT.Term to String. However, SMT.Encoder still keeps the mappings from Terms to abbreviated variables.

During this refactoring, SMT.Solver was updated to use the printer of SMTDDM.

After the refactoring, termination of `encodeTerm` is proven. :) One `partial` is added to a translation function from SMT.Term to SMTDDM.Term however, due to the added complexity of correctly translating triggers.

---

Special thanks to Kiro. :)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #495, opened 2026-02-27
Parse and lift the SMT counterexamples to LExpr

aqjune-aws:
This patch

- Parses the output of SMT solver containing counterexample using SMTDDM,
- Lifts SMTDDM.Term to SMT.Term, then to LExpr, and
- Let Core.verify properly print the counterexample using the new printer of Core.

To implement this,
- A new `parseCategoryFromDialect` function is added to DDM. It parses a string using a specific syntax category.
- A bug in SpaceSepBy making only odd'th symbols to be parsed was fixed. It seems SpaceSepBy was only used by SMTDDM.
- The old CexParser.lean is erased.

A new StrataTest/Languages/Core/CounterExampleLiftTest.lean test shows a handful of interesting results.

---

 #482 and this pull request have overlapping edits. Merging #482 does not introduce merge conflict, but tests need updates. Would be great if #482 is reviewed first! :)

---

Thanks to Kiro :)

### DIFF
--- a/Strata/DL/SMT/DDMTransform/Translate.lean
+++ b/Strata/DL/SMT/DDMTransform/Translate.lean
@@ -26,7 +26,6 @@ private def mkSimpleSymbol (s:String):SimpleSymbol SourceRange :=
     | "percent" => .simple_symbol_percent SourceRange.none
     | "questionmark" => .simple_symbol_questionmark SourceRange.none
     | "period" => .simple_symbol_period SourceRange.none
-    | "dollar" => .simple_symbol_dollar SourceRange.none
     | "tilde" => .simple_symbol_tilde SourceRange.none
     | "amp" => .simple_symbol_amp SourceRange.none
     | "caret" => .simple_symbol_caret SourceRange.none
--- a/Strata/DL/SMT/DDMTransform/Parse.lean
+++ b/Strata/DL/SMT/DDMTransform/Parse.lean
@@ -89,7 +89,6 @@ def specialCharsInSimpleSymbol := [
     ("questionmark", "?"),
     -- ("bang", "!"),
     ("period", "."),
-    ("dollar", "$"),
     -- ("underbar", "_"),
     ("tilde", "~"),
     ("amp", "&"),


Strata/Languages/Core/Verifier.lean
```diff
@@ -152,6 +151,12 @@ instance : ToFormat Outcome where
     | .unknown => "🟡 unknown"
     | .implementationError e => s!"🚨 Implementation Error! {e}"
 
+/--
+A counterexample expressed as Core `LExpr` values, suitable for display
+using Core's expression formatter and for future use as program metadata.
+-/
```
atomb, line 157: :heart:

Strata/DDM/Elab.lean
```diff
@@ -412,5 +412,57 @@ def parseStrataProgramFromDialect (dialects : LoadedDialects) (dialect : Dialect
         return s!"{msg}  {e.pos.line}:{e.pos.column}: {← e.data.toString}\n"
       throw (IO.userError errMsg)
 
+/--
+Parse a single syntax category from a loaded dialect set.
```
atomb, line 416: Oh, nice! I was thinking we'd need something like this eventually.

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -114,9 +103,83 @@ def runSolver (solver : String) (args : Array String) : IO IO.Process.Output :=
...
+  | .ok (.specific_success_response _ (.ssr_check_sat _ (.csr_unknown _))) => return some .unknown
+  | _ => return none
+
+/--
+Parse a `(get-value ...)` model response using the SMTResponse DDM dialect.
+Uses `parseCategoryFromDialect` targeting `SMTResponse.GetValueResponse`
+directly, which avoids the ambiguity that arises when parsing at the
+`Command` level.
+
+Returns a list of (key-string, value-Term) pairs on success.
```
atomb, line 135: ❤️

StrataTest/Languages/Core/CounterExampleLiftTest.lean
```diff
@@ -0,0 +1,192 @@
...
+};
+#end
+
+/--
+info:
+Obligation: must_be_cons
+Property: assert
+Result: ❌ fail
+Model:
+($__xs1, Cons(0, Nil))
```
atomb, line 132: ❤️

atomb (approved):

Strata/DL/SMT/CexParser.lean
shigoel, line 1: So happy to see this eliminated!

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -16,16 +20,22 @@ namespace SMT
...
+to an `SMT.Term`.
 -/
-abbrev CounterEx (Ident : Type) := Map Ident String
+abbrev CounterEx (Ident : Type) := Map Ident Strata.SMT.Term
+
+/-- Render an `SMT.Term` to a string via the SMTDDM translation. -/
+private def termToString (t : Strata.SMT.Term) : String :=
+  match Strata.SMTDDM.termToString t with
+  | .ok s => s
+  | .error _ => repr t |>.pretty
```
shigoel, line 31: Errors are not propagated here?

Strata/DL/SMT/DDMTransform/Translate.lean
```diff
@@ -246,4 +246,81 @@ def termTypeToString (t:SMT.TermType): Except String String := do
...
+partial def translateFromDDMTerm (t : Strata.SMTResponseDDM.Term Strata.SourceRange)
+    : Strata.SMT.Term :=
+  match t with
+  | .spec_constant_term _ sc =>
+    match sc with
+    | .sc_numeral _ n     => .prim (.int n)
+    | .sc_numeral_neg _ n => .prim (.int (-(n : Int)))
+    | .sc_decimal _ d     => .prim (.real d)
+    | .sc_str _ s         => .prim (.string s)
+    | _  => panic! s!"translateFromDDMTerm: don't know how to convert {repr t}"
```
shigoel: Can we have an exception here instead of a panic?

Strata/DL/SMT/DDMTransform/Translate.lean
```diff
@@ -246,4 +246,81 @@ def termTypeToString (t:SMT.TermType): Except String String := do
...
+    | .qi_isort _ iden sort =>
+      match iden with
+      | .iden_simple _ sym | .iden_indexed _ sym _
+      => some (symbolToString sym, some sort)
+  /-- Extract the raw string from a Symbol. -/
+  symbolToString (sym : Strata.SMTResponseDDM.Symbol Strata.SourceRange) : String :=
+    formatArg (.op (Strata.SMTResponseDDM.Symbol.toAst sym))
+  /-- Build a `Term.app` with a UF op for a named function/constructor. -/
+  mkUFApp (name : String) (args : List Strata.SMT.Term) : Strata.SMT.Term :=
+    let uf : Strata.SMT.UF := { id := name, args := [], out := .bool }
```
shigoel: Not sure where the `.bool` is coming from?

Strata/Languages/Core/Verifier.lean
```diff
@@ -163,6 +168,10 @@ structure VCResult where
   result : Outcome := .unknown
   estate : EncoderState := EncoderState.init
   verbose : VerboseMode := .normal
+  /-- Counterexample with values converted from `SMT.Term` to Core `LExpr`.
+      The contents must be consistent with smtObligationResult, if
+      smtObligationResult was .sat. -/
+  counterExample : LExprCounterEx := []
```
shigoel: Do we even need SMT.Result anymore? I'd vote for changing the types of both `smtObligationResult` and `smtReachResult` by `LExprCounterEx`.

shigoel: Another reason to have just one representation is so that the SARIF output rendering can use just one kind of model then.

Strata/Languages/Core/Verifier.lean
```diff
@@ -152,6 +151,12 @@ instance : ToFormat Outcome where
     | .unknown => "🟡 unknown"
     | .implementationError e => s!"🚨 Implementation Error! {e}"
 
+/--
+A counterexample expressed as Core `LExpr` values, suitable for display
+using Core's expression formatter and for future use as program metadata.
+-/
+abbrev LExprCounterEx := List (Expression.Ident × LExpr CoreLParams.mono)
```
shigoel: Could we use another term, like `LExprModel` here? Not all models obtained from SMT solvers will be counterexamples; some will be witnesses too.

StrataTest/Languages/Core/PolymorphicProcedureTest.lean
```diff
@@ -56,7 +56,7 @@ Result: Obligation: (Origin_Extract_Requires)Extract_requires_0
 Property: assert
 Result: ❌ fail
 Model:
-($__xs3, (as Nil (List Int))
+($__xs3, Nil)
```
shigoel, line 59: Umm, I thought we'd see `Nil()` here?

shigoel: Also true for examples in the newly added CexLift test file.

aqjune-aws: On my Strata `Nil` seemed to be a valid syntax too. :) In fact I was also in this rabbit hole and tried lifting the counterexample printer to print `Nil()`, but after confirming that e.g., `Cons(0, Nil)` could be accepted I chose to do so.

Strata/DL/Imperative/SMTUtils.lean
aqjune-aws: Hmm, yes. I will have a look.

StrataTest/Languages/Core/PolymorphicProcedureTest.lean
shigoel: I'm okay with either right now, but thought I'd bring it up because of the comment above `smtTermToLExpr` in SMTEncoder.lean which says something more specific here.

Strata/DL/Imperative/SMTUtils.lean
aqjune-aws: This is only used in the `ToFormat (CounterEx Ident)` instance, and `ToFormat.format` has to return `Format` :(

Strata/Languages/Core/Verifier.lean
aqjune-aws: It seems `smtObligationResult` and `smtReachResult` can bring good information from being `.unsat` or `.unknown` (and `.none` for `smtReachResult`) - they still seem to have meaningful values.

StrataTest/Languages/Core/PolymorphicProcedureTest.lean
aqjune-aws: That's right... It was legacy of the old trials. I fixed the comments accordingly, thanks.

Strata/DL/SMT/DDMTransform/Translate.lean
aqjune-aws: Fixed. :)

Strata/DL/SMT/DDMTransform/Translate.lean
aqjune-aws: Actually this was a really great catch. The result of translateFromDDMTerm could not have the correct type information because the term of SMTDDM does not have type information. Therefore, all type placeholders were filled with a temporary `.bool` and the number of arguments must have been changed. Indeed, the users of this function in this patch are all ignoring the types annotated in SMT.Term after parsing response (search `_retTy`).
I think this should be mentioned very explicitly - I changed the function name to include 'ToUntyped', and also left an explicit warning message.

Strata/Languages/Core/Verifier.lean
aqjune-aws: Updated.

Strata/DL/Imperative/SMTUtils.lean
shigoel: Ah alright. Thx.

Strata/DL/SMT/DDMTransform/Translate.lean
shigoel: Sorry, I don't think I understand completely, but some general comments first: I'm a little uncomfortable with `.bool` being a placeholder -- can we have a fresh SMT sort here, for instance?

Also: could you remind me why SMTDDM term doesn't have type info?

aqjune-aws: I added a commit that uses a "_placeholder" type. :)

Yeah, SMTDDM term does not have a type info because the parsing tree of e.g., `(f a b)` does not have that. SMTDDM is indeed a concrete syntax tree without type information. The full type information in LExpr is reconstructed when lifting SMT.Term to LExpr.

It is possible to annotate SMT.Term with the right types after getting the right signatures of UF as well as natively supported operators like `+`, `-`, ... of SMT, but this seemed redundant because filling in the types of LExpr could be done regardless of the existence of types in SMT.Term.

aqjune-aws:
Fixed the latest failure after #496 by updating DDM's Parser to also correspondingly recognize '$'. :)

Strata/DL/SMT/DDMTransform/Translate.lean
shigoel: Thanks for the placeholder type addition!

Indeed -- I agree with you that reconstructing types isn't really necessary because we can get them from Lambda.

atomb (approved):

PR #498, opened 2026-02-27
ci: Include .st files in lake cache key to prevent stale Laurel .olean

MikaelMayer:
The `lean-action` default cache key does not include `.st` grammar files. When `.st` files change, Laurel `.olean` files can become stale, causing incorrect CI results where old compiled grammar is used.

This PR adds a custom lake cache step to both `ci.yml` and `cbmc.yml` that includes `hashFiles('**/*.st')` in the cache key, and disables `lean-action`'s built-in cache (`use-github-cache: false`) to avoid conflicts.

Tested by verifying CI passes with the custom cache in place.

### DIFF
--- a/.github/workflows/cbmc.yml
+++ b/.github/workflows/cbmc.yml
@@ -52,8 +52,22 @@ jobs:
         run: |
           wget https://github.com/diffblue/cbmc/releases/download/cbmc-6.4.1/ubuntu-22.04-cbmc-6.4.1-Linux.deb
           sudo dpkg -i ubuntu-22.04-cbmc-6.4.1-Linux.deb
+      - name: Restore lake cache
+        uses: actions/cache/restore@v4
+        with:
+          path: .lake
+          key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/cbmc.yml
```diff
@@ -52,8 +52,22 @@ jobs:
         run: |
           wget https://github.com/diffblue/cbmc/releases/download/cbmc-6.4.1/ubuntu-22.04-cbmc-6.4.1-Linux.deb
           sudo dpkg -i ubuntu-22.04-cbmc-6.4.1-Linux.deb
+      - name: Restore lake cache
+        uses: actions/cache/restore@v4
+        with:
+          path: .lake
+          key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
+          restore-keys: |
+            lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}
```
atomb, line 61: It's only one of the several `.st` files that triggers this issue, but I guess this is still reasonable.

MikaelMayer: Ok, I'll revisit this if we need more of these files. Thanks.

PR #499, opened 2026-02-28
Add recursive functions to Strata Core

joscoh:
*Issue #, if available:*

*Description of changes:* This PR adds support for recursive functions in Strata Core, including their definition in concrete syntax, partial evaluation, and an axiom-based SMT encoding.

Main changes:
1. This PR adds an additional attribute to the DDM `@[scopeSelf]` that acts like `@[scope]` but also adds the function's name to its body's scope. This is structured and handled very similarly to `@[declareDatatype]`, which involves a similar recursive structure. Main changes are in `AST.lean`, `Format.lean`, `Core.lean`. This adds a `@[cases]` metadata attribute to indicate the parameter on which case splitting should occur.
2. Adds fields to `Func` indicating whether a function is recursive and the argument to case-split on. This information is used in `Env.lean` when the function is added to the environment. Here we also rule out recursive functions that are not yet supported: polymorphic functions, functions with more sophisticated termination measures, and functions without an explicit `cases` clause.
3. `RecursiveAxioms.lean` contains the generation of axioms as `LExpr` for a recursive function. Generating the axiom is fairly simple, since we just create term e.g. `forall h t, length (List.cons h t) = length (List.cons h t)` and then partially evaluate the right-hand side. Since recursive functions are marked as `inlineIfConstr` on their case-splitting parameter, this produces the simplified form.

Other changes
1. Adds support in `SMT/DDMTransform/Translate` for trigger patterns complex enough to handle constructors applied to arguments (e.g. `List.cons h t`).
2. Changes the evaluation of if-then-else to always reduce within the `then` and `else` branches and changes the small-step semantics to match.

Tests
1. `RecursiveAxiomTests.lean` tests the generation of recursive axioms as Lambda terms.
2. This PR adds a test to `SMTEncoderDatatypeTests.lean` that demonstrates the SMT encoding of the recursive function axioms.
3. `RecursiveFunctionDDMTest.lean` tests the DDM parsing for recursive function.
4. `RecursiveFunctionErrorTest.lean` demonstrates not-yet-supported recursive functions.
5. `RecursiveFunctionTests.lean` contain a variety of programs with recursive functions, some of which contain VCs which can be solved purely by partial evaluation and others which require SMT-based reasoning.
6. `BinaryTreeSize.lean` gives a more complex example that demonstrates inductive reasoning over types and functions.

What is not supported (yet): polymorphic and mutually recursive functions, termination checking.

### DIFF
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -276,10 +276,6 @@ op command_procedure (name : Ident,
 op command_typedecl (name : Ident, args : Option Bindings) : Command =>
   "type " name args ";\n";
 
-@[declareTypeForward(name, some args)]
-op command_forward_typedecl (name : Ident, args : Option Bindings) : Command =>
-  "forward type " name args ";\n";
-
 @[aliasType(name, some args, rhs)]
 op command_typesynonym (name : Ident,
                         args : Option Bindings,
@@ -383,8 +379,8 @@ op command_datatype (name : Ident,
       "datatype " name typeParams " {" constructors "\n}" ";\n";
 
 // Mutual block for defining mutually recursive types
-// Types should be forward-declared before the mutual block
-@[scope(commands)]
+// Type names are pre-registered via @[preRegisterTypes] before elaboration
+@[scope(commands), preRegisterTypes(commands)]
 op command_mutual (commands : SpacePrefixSepBy Command) : Command =>
   "mutual\n  " indent(2, commands) "end;\n";
 
--- a/Strata/DDM/AST.lean
+++ b/Strata/DDM/AST.lean
@@ -13,6 +13,7 @@ public import Strata.DDM.Util.SourceRange
 
 import Std.Data.HashMap
 import all Strata.DDM.Util.Array
+import Strata.Util.DecideProp
 import all Strata.DDM.Util.ByteArray
 
 set_option autoImplicit false
@@ -622,6 +623,23 @@ def resultLevel (varCount : Nat) (metadata : Metadata) : Option (Fin varCount) :
     else
       panic! s!"Scope index {idx} out of bounds (varCount = {varCount})"
 
+/-- Returns the argument index from @[preRegisterTypes] metadata, if present. -/
+def preRegisterTypesIndex (metadata : Metadata) : Option Nat :=
+  match metadata[q`StrataDDL.preRegisterTypes]? with
+  | none => none
+  | some #[.catbvar idx] => some idx
+  | some _ => panic! s!"Unexpected argument count to preRegisterTypes"
+
+/-- Returns the level for @[preRegisterTypes] metadata, if present. -/
+def preRegisterTypesLevel (varCount : Nat) (metadata : Metadata) : Option (Fin varCount) :=
+  match metadata.preRegisterTypesIndex with
+  | none => none
+  | some idx =>
+    if _ : idx < varCount then
+      some ⟨varCount - (idx + 1), by omega⟩
+    else
+      panic! s!"preRegisterTypes index {idx} out of bounds (varCount = {varCount})"
+
 end Metadata
 
 abbrev Var := String
@@ -932,122 +950,6 @@ def buildFunctionType (template : FunctionTemplate)
   .ok <| paramTypes.foldr (init := returnType) fun argType tp => .arrow default argType tp
 
 /--
-Result of expanding a single template.
-Contains the generated function signatures and any errors encountered.
--/
-structure TemplateExpansionResult where
-  /-- Generated function signatures as (name, type) pairs -/
-  functions : Array (String × TypeExpr)
-  /-- Errors encountered during expansion -/
-  errors : Array String
-  deriving Repr
-
-/--
-Expand a single function template based on its scope.
-
-Function templates specify patterns for generating auxiliary functions
-from datatype declarations. This function expands one template according to
-its iteration scope:
-
-- **perConstructor**: Generates one function per constructor (e.g., testers
-like `..isNone`)
-- **perField**: Generates one function per unique field across all constructors
-(e.g., accessors)
-
-**Parameters:**
-- `datatypeName`: Name of the datatype (used in name pattern expansion)
-- `datatypeType`: TypeExpr for the datatype (used in function signatures)
-- `constructorInfo`: Array of constructor information
-- `template`: The function template to expand
-- `dialectName`: Dialect name (for resolving builtin types)
-- `existingNames`: Set of already-used names (for duplicate detection)
-
-**Example:** For a `perConstructor` template defined as:
-```
-perConstructor([.datatype, .literal "..is", .constructor], [.datatype],
-.builtin "bool")
-```
-This specifies:
-- Name pattern: `[.datatype, .literal "..is", .constructor]` → generates names
-like `Option..isNone`
-- Parameter types: `[.datatype]` → takes one parameter of the datatype type
-- Return type: `.builtin "bool"` → returns a boolean
-
-Applied to `Option<T>` with constructors `None` and `Some`, this generates:
-- `Option..isNone : Option<T> -> bool`
-- `Option..isSome : Option<T> -> bool`
--/
-def expandSingleTemplate
-    (datatypeName : String)
-    (datatypeType : TypeExpr)
-    (constructorInfo : Array ConstructorInfo)
-    (template : FunctionTemplate)
-    (dialectName : String)
-    (existingNames : Std.HashSet String) : TemplateExpansionResult :=
-  -- First validate the pattern
-  match validateNamePattern template.namePattern template.scope with
-  | some err => { functions := #[], errors := #[err] }
-  | none =>
-    match template.scope with
-    | .perConstructor =>
-      -- Generate one function per constructor
-      let (funcs, errs, _) := constructorInfo.foldl (init := (#[], #[], existingNames)) fun (funcs, errs, names) constr =>
-        let funcName := expandNamePattern template.namePattern datatypeName (some constr.name)
-        if names.contains funcName then
-          (funcs, errs.push s!"Duplicate function name: {funcName}", names)
-        else
-          match buildFunctionType template datatypeType none dialectName with
-          | .ok funcType =>
-            (funcs.push (funcName, funcType), errs, names.insert funcName)
-          | .error e =>
-            (funcs, errs.push e, names)
-      { functions := funcs, errors := errs }
-
-    | .perField =>
-      -- Generate one function per unique field across all constructors
-      -- Error if the same field name appears with different types
-      let allFields := constructorInfo.foldl (init := #[]) fun acc c => acc ++ c.fields
-      let (funcs, errs, _) := allFields.foldl (init := (#[], #[], existingNames)) fun (funcs, errs, names) (fieldName, fieldTp) =>
-        let funcName := expandNamePattern template.namePattern datatypeName none (some fieldName)
-        if names.contains funcName then
-          (funcs, errs.push s!"Duplicate field name '{fieldName}' across constructors in datatype '{datatypeName}'", names)
-        else
-          match buildFunctionType template datatypeType (some fieldTp) dialectName with
-          | .ok funcType =>
-            (funcs.push (funcName, funcType), errs, names.insert funcName)
-          | .error e =>
-            (funcs, errs.push e, names)
-      { functions := funcs, errors := errs }
-
-/--
-This function generates function signatures for an array of function templates
-in order. Templates are specified in `@[declareDatatype]` annotations
-to automatically generate auxiliary functions like testers and field accessors.
-Within each template, functions are generated in constructor/field declaration order.
-
-**Parameters:**
-- `datatypeName`: Name of the datatype
-- `datatypeType`: TypeExpr for the datatype
-- `constructorInfo`: Array of constructor information
-- `templates`: Array of function templates to expand
-- `dialectName`: Dialect name (for resolving builtin types)
-- `existingNames`: Optional set of pre-existing names to avoid
--/
-def expandFunctionTemplates
-    (datatypeName : String)
-    (datatypeType : TypeExpr)
-    (constructorInfo : Array ConstructorInfo)
-    (templates : Array FunctionTemplate)
-    (dialectName : String)
-    (existingNames : Std.HashSet String := {}) : TemplateExpansionResult :=
-  templates.foldl (init := { functions := #[], errors := #[] }) fun acc template =>
-    -- Track names from previous templates to detect cross-template duplicates
-    let currentNames := acc.functions.foldl (init := existingNames) fun s (name, _) => s.insert name
-    let result := expandSingleTemplate datatypeName datatypeType constructorInfo template dialectName currentNames
-    { functions := acc.functions ++ result.functions
-      errors := acc.errors ++ result.errors }
-
-/--
 Specification for datatype declarations.
 Includes indices for extracting datatype information and optional function templates.
 -/
@@ -1077,7 +979,6 @@ A spec for introducing a new binding into a type context.
 inductive BindingSpec (argDecls : ArgDecls) where
 | value (_ : ValueBindingSpec argDecls)
 | type (_ : TypeBindingSpec argDecls)
-| typeForward (_ : TypeBindingSpec argDecls)  -- Forward declaration (no AST node)
 | datatype (_ : DatatypeBindingSpec argDecls)
 | tvar (_ : TvarBindingSpec argDecls)
 deriving Repr
@@ -1087,7 +988,6 @@ namespace BindingSpec
 def nameIndex {argDecls} : BindingSpec argDecls → DebruijnIndex argDecls.size
 | .value v => v.nameIndex
 | .type v => v.nameIndex
-| .typeForward v => v.nameIndex
 | .datatype v => v.nameIndex
 | .tvar v => v.nameIndex
 
@@ -1128,12 +1028,19 @@ private def mkValueBindingSpec
     newBindingErr "Arguments only allowed when result is a type."
   return { nameIndex, argsIndex, typeIndex, allowCat }
 
-/-- Parse function templates from metadata arguments. -/
-private def parseFunctionTemplates (args : Array MetadataArg) : Array FunctionTemplate :=
-  args.filterMap fun arg =>
+/-- Parse and validate function templates from metadata arguments. -/
+private def parseFunctionTemplates (args : Array MetadataArg)
+    : NewBindingM (Array FunctionTemplate) := do
+  let mut result := #[]
+  for arg in args do
     match arg with
-    | .functionTemplate t => some t
-    | _ => none
+    | .functionTemplate t =>
+      if let some err := validateNamePattern t.namePattern t.scope then
+        newBindingErr s!"Function template error: {err}"
+      else
+        result := result.push t
+    | _ => pure ()
+  return result
 
 def parseNewBindings (md : Metadata) (argDecls : ArgDecls) : Array (BindingSpec argDecls) × Array String :=
   let ins (attr : MetadataAttr) : NewBindingM (Option (BindingSpec argDecls)) := do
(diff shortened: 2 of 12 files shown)


Strata/DL/Lambda/RecursiveAxioms.lean
```diff
@@ -0,0 +1,104 @@
...
+    Assumes the function is well-formed (use `checkRecursiveFunc` first).
+    The PE must have the function in its factory with `inlineIfConstr`. -/
+def mkRecursiveAxioms [Inhabited T.Metadata] [DecidableEq T.Metadata] [DecidableEq T.IDMeta]
+    (func : LFunc T) (recIdx : Nat) (dt : LDatatype T.IDMeta)
+    (pe : LExpr T.mono → LExpr T.mono) (m : T.Metadata)
+    : Except Format (List (LExpr T.mono)) :=
+  let formals := func.inputs.keys
+  let inputTys := func.inputs.values
+  dt.constrs.mapM fun c => do
+    let numFields := c.args.length
```
MikaelMayer, line 60: The index arithmetic is subtle. Consider naming intermediate results to make the calculation clearer:

```lean
let numFields := c.args.length
let numOtherParams := formals.length - 1
let totalBvs := numFields + numOtherParams
let dtTy : LMonoTy := .tcons dt.name []
let constrTy := c.args.foldr (fun (_, argTy) acc => .arrow argTy acc) dtTy
let constrApp := c.args.foldlIdx (fun acc i _ =>
  .app m acc (.bvar m (numFields - 1 - i))
) (.op m c.name (.some constrTy) : LExpr T.mono)
let otherParamIdx (idx : Nat) : Nat := if idx < recIdx then idx else idx - 1
let formalExpr (idx : Nat) : LExpr T.mono :=
  if idx == recIdx then constrApp
  else 
    let paramPos := otherParamIdx idx
    .bvar m (totalBvs - 1 - paramPos)
```

Key changes: introduce `numOtherParams` (all params except recursive one), rename `otherIdx` to `otherParamIdx`, extract `paramPos` intermediate variable in `formalExpr`.

Strata/DL/Lambda/RecursiveAxioms.lean
```diff
@@ -0,0 +1,104 @@
...
+      else .bvar m (totalBvs - 1 - otherIdx idx)
+    -- LHS: f(bvars..., C(fields...), bvars...) — not PE'd (serves as trigger)
+    let lhs := formals.foldlIdx (fun acc idx _ =>
+      .app m acc (formalExpr idx)
+    ) (LFunc.opExpr func)
+    -- RHS: PE inlines the function since the recursive arg is a constructor
+    let rhs := pe lhs
+    if lhs == rhs then
+      .error f!"Recursive function {func.name}: PE did not reduce axiom for \
+               constructor {c.name}. Ensure the function is in the factory with inlineIfConstr."
```
MikaelMayer: The error message when PE doesn't reduce is helpful, but it could be more actionable. Consider adding what the user should check:

```lean
.error f!"Recursive function {func.name}: PE did not reduce axiom for \
         constructor {c.name}. Ensure the function is in the factory with inlineIfConstr \
         and that the function body can be partially evaluated when the recursive parameter \
         is a constructor."
```

Strata/Languages/Core/Env.lean
```diff
@@ -201,7 +201,25 @@ def Env.addFactory (E : Env) (f : (@Lambda.Factory CoreLParams)) : Except Diagno
   let exprEnv ← E.exprEnv.addFactory f
   .ok { E with exprEnv := exprEnv }
 
+/-- Add a function to the environment. For recursive functions, resolves the
+    `decreases` parameter name to an `inlineIfConstr` attribute index, and
+    rejects cases not yet supported for SMT verification (polymorphic recursive
+    functions, missing decreases clauses, non-parameter decreases expressions). -/
 def Env.addFactoryFunc (E : Env) (func : (Lambda.LFunc CoreLParams)) : Except DiagnosticModel Env := do
+  if func.isRecursive && !func.typeArgs.isEmpty then
```
MikaelMayer, line 209: The error message for polymorphic recursive functions could explain why this limitation exists:

```suggestion
    .error (.fromFormat f!"Polymorphic recursive functions are not yet supported for SMT verification: '{func.name}'. SMT solvers require monomorphic axioms.")
```

StrataTest/Languages/Core/RecursiveFunctionTests.lean
```diff
@@ -0,0 +1,455 @@
+/-
```
MikaelMayer, line 1: Excellent test coverage! The tests progress nicely from simple ground assertions to symbolic reasoning to complex examples. Well done.

Strata/DL/Lambda/RecursiveAxioms.lean
joscoh: I will add a comment explaining the arithmetic but I changed the code from introducing all of these extra variables because I thought it was overcomplicated and harder to read.

Strata/DL/Lambda/RecursiveAxioms.lean
```diff
@@ -0,0 +1,117 @@
...
+## Axiom Generation for Recursive Functions
+
+Given a recursive function with a `decreases` parameter over an algebraic datatype,
+generates per-constructor axioms. Each axiom is a quantified equation:
+
+  ∀ (other_params..., fields...). f(..., C(fields...), ...) = PE(f(..., C(fields...), ...))
+
+The LHS is left unevaluated (it serves as the trigger pattern). The RHS is obtained by
+passing the LHS to the partial evaluator, which inlines the function (since the recursive
+argument is a constructor application) and reduces.
```
MikaelMayer, line 21: Wait it looks like this axiom generation has nothing to do with recursion. It's about emitting axioms for functions applied to constructor. What is the relationship of this problem with the decreases clauses and the recursivity? It seems pretty orthogonal to me.

joscoh: The (structural) recursion occurs over constructors. So if we define recursive function 
```
recursive function length (l: IntList) : int 
{
  if List..isNil l then 0 else 1 + length (List..tail l)
}
```
This function satisfies the properties:
```
length Nil = 0
forall h t, length (Cons h t) = 1 + length t
```
These properties define both the evaluation behavior and the axiomatization for the SMT solver. The evaluation behavior actually follows automatically from the already existing evaluation behavior of pieces like `List..isNil`, `List..tail`, etc which simplify when applied to constructors. But in this file we construct the axioms to send to the SMT solver.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -93,6 +89,18 @@ as `ite x (λ.e1) (λ.e2)`.  -/
     Step F rf econd econd' →
     Step F rf (.ite m econd ethen eelse) (.ite m' econd' ethen eelse)
 
+/-- Evaluation of `ite` "then" branch (when condition is not yet resolved). -/
+| ite_reduce_then_branch:
+  ∀ (econd ethen ethen' eelse:LExpr Tbase.mono),
+    Step F rf ethen ethen' →
+    Step F rf (.ite m econd ethen eelse) (.ite m' econd ethen' eelse)
```
MikaelMayer, line 96: Why do you add these semantic steps?

joscoh: Because the evaluation of `.ite` changed in the partial evaluator (among other things, this lets us use the partial evaluator to generate the axioms), and this keeps the semantics consistent.

Strata/DL/Util/Func.lean
```diff
@@ -66,6 +66,8 @@ structure Func (IdentT : Type) (ExprT : Type) (TyT : Type) (MetadataT : Type) wh
   name     : IdentT
   typeArgs : List TyIdentifier := []
   isConstr : Bool := false --whether function is datatype constructor
+  isRecursive : Bool := false
+  decreases : Option ExprT := none
```
MikaelMayer: Decreases clauses in Dafny are a List, I would prefer that. The list represents a lexicographic tuple, which is extremely useful when you want to refactor a function. I know that your PR only needs to decrease over a datatype, but in the future we will need more than just a datatype.

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -309,6 +309,20 @@ op command_fndef (name : Ident,
...
+@[declareFn(name, b, r)]
+op command_recfndef (name : Ident,
+                     typeArgs : Option TypeArgs,
+                     @[scope(typeArgs)] b : Bindings,
+                     @[scope(typeArgs)] r : Type,
+                     @[scope(b)] preconds : Seq SpecElt,
+                     @[scope(b)] dec : Option DecreasesClause,
+                     @[scopeSelf(name, b, r)] c : r,
+                     inline? : Option Inline) : Command =>
+  "recursive " inline? "function " name typeArgs b " : " r indent(2, preconds) "\n  " dec "\n{\n  " indent(2, c) "\n}\n";
```
MikaelMayer: I'll leave the final choice to you since the concrete syntax is not very important at this stage, but what about using "rec" instead of "recursive" to make benchmarks less verbose?

StrataTest/Languages/Core/Examples/BinaryTreeSize.lean
```diff
@@ -0,0 +1,175 @@
...
+{ IntTree..val(t) }
+
+recursive function listLen (xs : IntList) : int
+  decreases xs
+{
+  if IntList..isNil(xs) then 0 else 1 + listLen(tail(xs))
+}
+
+recursive function append (xs : IntList, ys : IntList) : IntList
+  decreases xs
```
MikaelMayer: I would also like to see negative tests on what decreases does not support yet:
- Decreases with multiple arguments
- Decreases with a type that is not a datatype
- Decreases with a type that does not decrease in the body and trying to prove false after that to verify we don't emit an unsound axiom.

joscoh: Did you see `RecursiveFunctionErrorTest.lean`? We cannot check the third because we don't have termination checking yet, this is for a future PR. Right now, we can emit unsound axioms.

Strata/Languages/Core/DDMTransform/Grammar.lean
joscoh: Yes, I prefer that as well.

Strata/DL/Util/Func.lean
joscoh: Yes, but that is explicitly for a future PR where more flexible decreases clauses will be supported. It is not yet determined exactly what kinds of recursion we will support and how, so I would prefer not to explicitly commit to supporting lexicographic tuples quite yet.

Strata/DL/Lambda/Semantics.lean
MikaelMayer: I would have hoped that you could have _derived_ these rules by invoking that every boolean is either true or false, and that every expression evaluates to a value that is of its type?

joscoh: This is not true because there can be uninterpreted symbols that cannot be immediately evaluated. We need to relax the `ite` evaluation from call-by-value to nondeterministic. Let me try to find the example that necessitated this change.

StrataTest/Languages/Core/Examples/BinaryTreeSize.lean
MikaelMayer: I saw that file after writing my comment indeed, but still, I think there is no negative test of decreases with multiple arguments, or things that are not datatypes.

Also, since we can introduce unsound axioms since we don't do termination checks, I fear now that the "decreases" clause will introduce an unexpected expectation that termination checks are already there.

What about we keep recursive keyword, but remove "decreases" clauses in favor of a metadata attribute that just says to partially the function's axioms for a given argument? e.g.

```
recursive function listLen (@[cases] xs : IntList) : int
{
  if IntList..isNil(xs) then 0 else 1 + listLen(IntList..tl(xs))
}
```
or even better, we could perhaps just do this by default?

MikaelMayer: Bonuses are:

- No need to check if decreases clauses is a single argument which is an parameter name
- We could also use that metadata for non-recursive functions. For example, a function that tests if a value is a subset of the constructors of a datatype.

Strata/DL/Lambda/Semantics.lean
joscoh: The example is:
```
function lookup (x: int, xs: IntList) : bool
{
  if List..isNil xs then false
  else if List..head xs == x then true
  else lookup(x, List..tail xs)
}
```
Without the more permissive evaluation, the `cons` axiom cannot simplify in the `else` branch, because the `if List..head xs == x` does not simplify to `true` or `false`.

StrataTest/Languages/Core/Examples/BinaryTreeSize.lean
joscoh: Changed this

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -245,6 +245,8 @@ op spec_mk (elts : Seq SpecElt) : Spec => "spec " indent(2, "{\n" elts "} ");
 category Binding;
 @[declare(name, tp)]
 op mkBinding (name : Ident, tp : TypeP) : Binding => @[prec(40)] name " : " tp;
+@[declare(name, tp)]
+op casesBinding (name : Ident, tp : TypeP) : Binding => @[prec(40)] "@[cases] " name " : " tp;
```
MikaelMayer, line 254: I thought there was a general mechanism for attributes?

joscoh: What do you mean by this exactly? I know we have attributes in `Func`, but those need to be handled specially in the grammar (e.g. with the `inline?` syntax)

Strata/DL/Lambda/LExprEval.lean
```diff
@@ -208,14 +208,8 @@ def evalIte (n' : Nat) (σ : LState TBase) (m: TBase.Metadata) (c t f : LExpr TB
...
   | _ =>
-    -- It's important to at least substitute `.fvar`s in both branches of the
-    -- `ite` here so that we can replace the variables by the values in the
-    -- state; these variables can come from an imperative dialect.
-    -- (TODO): Is it worth it to evaluate these branches instead?
-    -- let t' := eval n' σ t
-    -- let f' := eval n' σ f
-    let t' := substFvarsFromState σ t
-    let f' := substFvarsFromState σ f
+    let t' := eval n' σ t
```
shigoel, line 217: Could we add a comment at the top of this function that says that this evaluator is now eager instead of lazy? Perhaps a little bit of historical info. will help too -- i.e., this evaluator was originally lazy when we didn't have recursive functions, and we need it to be eager for that.

The reason I'm bringing this up is because I can imagine us wanting a mode where we switch this evaluator to lazy (or perhaps design a new lazy one), and such a comment will help retrieve context.

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -314,6 +316,16 @@ op command_fndef (name : Ident,
...
   inline? "function " name typeArgs b " : " r indent(2, preconds) " {\n  " indent(2, c) "\n}\n";
 
+@[declareFn(name, b, r)]
+op command_recfndef (name : Ident,
+                     typeArgs : Option TypeArgs,
+                     @[scope(typeArgs)] b : Bindings,
+                     @[scope(typeArgs)] r : Type,
+                     @[scope(b)] preconds : Seq SpecElt,
+                     @[scopeSelf(name, b, r)] c : r,
+                     inline? : Option Inline) : Command =>
```
shigoel: What does it mean for a recursive function to be "inline"? In the case of non-rec functions, we replace their calls by the bodies (after formals -> actuals, etc. ofc). But we don't have that behavior here.

I was expecting we wouldn't make the `inline` keyword a part of the syntax for recursive functions.

shigoel: Actually, I'm wondering what the axiom generation would look like for recursive functions marked `inline`....

Strata/Languages/Core/ProgramType.lean
```diff
@@ -123,7 +123,10 @@ def typeCheck (C: Core.Expression.TyContext) (Env : Core.Expression.TyEnv) (prog
 
       | .func func _ => try
         let Env := Env.pushEmptySubstScope
-        let (func', Env) ← Function.typeCheck C Env func |>.mapError (fun e => DiagnosticModel.withRange fileRange e)
+        -- For recursive functions, add to context before type-checking so the
+        -- body can reference itself.
```
shigoel: nit: so that the call to the function in the body can reference itself, right?

The only relevant thing about this temporary addition of the pre-type-checked recursive function to the context is the name and the signature. The body is irrelevant, right, because that's precisely what's being typechecked. Wonder if it makes sense to remove `body` from `func` added to `C'`... Perhaps just an expanded comment will suffice.

Strata/Languages/Core/StatementType.lean
```diff
@@ -168,6 +168,9 @@ where
             .error (errorWithSourceLoc e md)
 
         | .funcDecl decl md => do try
+          -- Recursive functions are only allowed as top-level declarations
```
shigoel, line 171: Just curious: any plans to provide them locally as well?

Strata/DL/Lambda/RecursiveAxioms.lean
```diff
@@ -0,0 +1,117 @@
...
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.DL.Lambda.Factory
+import Strata.DL.Lambda.TypeFactory
+import Strata.DL.Util.List
+
+/-!
+## Axiom Generation for Recursive Functions
```
shigoel, line 12: Is there a comment somewhere that describes the entire pipeline for rec. functions, from the use of PE to generate axioms so that at the SMT-level, we have an UF + axioms (with triggers)? I know this info is scattered across various comments, but it'd be nice to have an overview, along with some idea of what's the next TODO.

Strata/Languages/Core/DDMTransform/Grammar.lean
joscoh: Good point, this should be removed.

Strata/Languages/Core/StatementType.lean
joscoh: Yes, eventually. I don't think there is SMT support for function statements right now IIRC, but other than that I don't think there are any fundamental blockers.

Strata/DL/Lambda/RecursiveAxioms.lean
shigoel: It'd also be good to mention that Core's recursive functions support recursing on ADTs (IIUC), not, say `int`.

joscoh: I added a file with an overview of the pipeline and what is/is not yet supported (`docs/RecursiveFunctions.md`). This will certainly change as things are implemented, and I will keep this doc up to date. Eventually, we should expand the Strata Core language reference to handle declarations and include info about recursive functions, function preconditions, etc.

docs/RecursiveFunctions.md
```diff
@@ -0,0 +1,95 @@
...
+(ADTs). At the SMT level, a recursive function is encoded as an **uninterpreted
+function (UF)** together with **per-constructor axioms** equipped with
+**triggers**.
+
+## Syntax
+
+Recursive functions are declared with the `rec` keyword. Exactly one parameter
+must be annotated with `@[cases]` to indicate the ADT argument being recursed on:
+
+```boogie
```
shigoel: Umm, `boogie`?

joscoh: Ugh, sorry. Don't know why that got in there

docs/RecursiveFunctions.md
shigoel, line 1: Thank you!

StrataTest/Languages/Core/RecursiveFunctionErrorTest.lean
shigoel, line 1: Minor: not sure why some Core-related files are in Core/Examples and some in Core/.

joscoh: I tried to put more "unit-test" type files (especially those testing errors) in just Core/ and examples of actual programs in Core/Examples, which seems to mostly match the existing patterns. But this distinction is blurry, and maybe we should just have 1 directory anyway.

shigoel: Yes, I think we need two directories: `Tests` and `Examples`. Worth doing in a separate PR. I can take this on.

PR #501, opened 2026-03-01
Add a resolution phase to Laurel

keyboardDrummer:
### Changes
Instead of redefining how to do name resolution in each phase in the translation from Laurel to Core, define resolution once in a single phase. Currently resolution is done again after each phase. This introduces partially duplicated work but we expect to reduce that in the future using caching.

The Python through Laurel pipeline currently can not resolve each reference because it adds some prelude definitions, that are referenced by the generated code, only at the Core level instead of at the Laurel level. To still enable the Python through Laurel pipeline to work, I had to turn off reporting of resolution errors. I'll resolve that in a follow-up PR.

For Laurel tests, this PR includes changes to ensure that the program resolves without errors after each phase:
- Fix issue in DDM parsing caused by Lean.Name.toString inserting guillemets "« »"
- Datatype support has been improved. References to the type of a datatype now translate to Core correctly. There's now grammar support for datatypes
- The prelude for Laurel is now added during HeapParameterization phase, since that's the phase that depends on these definitions, instead of after the translation to Core. 
- The `typeTag: TypeTag` field of the composite datatype is only added to that datatype during the typeHierarchy phase, since that's the phase where `TypeTag` is defined.
- `const`,`select` and `update` are now added as part of `CoreDefinitionsForLaurel.lean`, as Laurel functions with an "External" body, meaning they're ignored during the translation to Core. This way, references to them can be resolved during resolution.
- During `HeapParameterization`, fields for composite type definitions are deleted, so their names don't conflict with the constructors of the generated `Field` datatype
- Small refactoring to simplify the definition of Forall and Exists, by reusing the Parameter type.

### Testing
Refactoring. No changes to existing tests. I will follow-up with first removing the hacks, and then adding tests containing name conflicts that'll make use of this new resolution phase.


Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
```diff
@@ -55,7 +55,12 @@ def checkOp (op : Strata.Operation) (name : QualifiedIdent) (argc : Nat) :
 def translateIdent (arg : Arg) : TransM Identifier := do
   let .ident _ id := arg
     | TransM.error s!"translateIdent expects ident"
-  return id
+  -- Strip Lean's «» guillemet notation that the DDM parser adds to identifiers
+  -- containing special characters (e.g. dots like Heap..data)
+  let id := if id.startsWith "«" && id.endsWith "»" then
```
joscoh: Why do you need to do this? This seems very brittle.

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -15,7 +15,30 @@ Documentation for Laurel can be found in docs/verso/LaurelDoc.lean
...
+    Carries a mandatory unique ID assigned by the resolution pass. -/
+structure Identifier where
+  /-- The declared name. -/
+  text : String
+  /-- Unique ID assigned by the resolution pass. -/
+  uniqueId : Option Nat := none
+  deriving Repr, Inhabited, DecidableEq
+
+-- Temporary hack because the Python through Laurel pipeline doesn't resolve
+instance : BEq Identifier where
```
joehendrix, line 29: This `BEq` compares only `text`, but the derived `DecidableEq` (line 26) compares both `text` and `uniqueId`. That means `a == b` and `decide (a = b)` can disagree, which can cause subtle bugs in code that dispatches on one vs the other.

Consider dropping `DecidableEq` from the `deriving` clause until this "hack" can be removed.

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -15,7 +15,30 @@ Documentation for Laurel can be found in docs/verso/LaurelDoc.lean
...
+  beq a b := a.text == b.text
+
+instance : Inhabited Identifier where
+ default := { text := "defaultIdentifier" }
```
joehendrix, line 33: There is already an Inhabited instance in the deriving version.  Pick one.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -589,31 +616,38 @@ def translate (program : Program) : Except (Array DiagnosticModel) (Core.Program
...
       typeArgs := []
       inputs := []
       output := coreTy
       body := body
     }
 
-  -- Collect ALL errors from both functions and procedures before deciding whether to fail
-  let allErrors := pureErrors ++ procDiags ++ constantsState.diagnostics
+  -- Collect ALL errors from both functions, procedures, and resolution before deciding whether to fail
+  let allErrors := -- resolutionDiags.toList ++
```
joehendrix: `_resolutionDiags` is not included here — the addition is commented out.  Is that intentional?

keyboardDrummer: Added a clarifying comment

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: > Consider dropping DecidableEq from the deriving clause until this "hack" can be removed.

Done

Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
joscoh: Why do you need to do this? It seems brittle; is it possible for a user to write an identifier that uses `«»`?

Strata/Languages/Laurel/Grammar/LaurelGrammar.st
```diff
@@ -106,6 +108,20 @@ op optionalExtends(parents: CommaSepBy Ident): OptionalExtends => "extends " par
 category Composite;
 op composite (name: Ident, extending: Option OptionalExtends, fields: Seq Field): Composite => "composite " name extending "{" fields "}";
 
+// Datatype definitions
+category DatatypeConstructorArg;
```
joscoh, line 112: This will not allow recursive type definitions, right?

Strata/Languages/Laurel/CoreDefinitionsForLaurel.lean
```diff
@@ -0,0 +1,46 @@
...
+function const(value: int) : int
+  external
+
+#end
+
+/--
+The core map operation definitions as a `Laurel.Program`, parsed at compile time.
+-/
+def coreDefinitionsForLaurel : Program :=
+  let uri := Strata.Uri.file "Strata/Languages/Laurel/CoreDefinitionsForLaurel.lean"
```
joscoh, line 41: Why does this involve its own filepath in the implementation, that seems strange to me.

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -0,0 +1,728 @@
...
+/-- Insert a definition into the refToDef map using the ID already on the identifier. -/
+private def register (map : Std.HashMap Nat AstNode) (iden : Identifier) (node : AstNode)
+    : Std.HashMap Nat AstNode :=
+  match iden.uniqueId with
+  | some uuid => map.insert uuid node
+  | none => map  -- shouldn't happen after Phase 1
+
+mutual
+
+private partial def collectStmtExpr (map : Std.HashMap Nat AstNode) (expr : StmtExprMd)
```
joscoh: Any reason these are `partial`?

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -455,26 +415,28 @@ def heapTransformProcedure (proc : Procedure) : TransformM Procedure := do
...
+  -- Remove fields from composite types since they are now stored in the heap
+  let types' := program.types.map fun td =>
+    match td with
+    | .Composite ct => .Composite { ct with fields := [] }
+    | other => other
   { program with
-    staticProcedures := procs',
-    types := program.types ++ [fieldDatatype] }
+    staticProcedures := heapConstants.staticProcedures ++ procs',
+    types := [fieldDatatype] ++ heapConstants.types ++ types' }
```
joscoh: Nit: `fieldDatatype :: heapConstants.types`

Strata/Languages/Laurel/HeapParameterizationConstants.lean
```diff
@@ -0,0 +1,77 @@
...
+      update(select(Heap..data!(heap), obj), field, val)),
+    Heap..nextReference!(heap))
+}
+
+// Increment the heap allocation nextReference, returning a new heap
+function increment(heap: Heap): Heap {
+  MkHeap(Heap..data!(heap), Heap..nextReference!(heap) + 1)
+}
+
+datatype Workaround  {}
```
joscoh, line 71: What is this for? It should be commented at least

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -15,7 +15,30 @@ Documentation for Laurel can be found in docs/verso/LaurelDoc.lean
 namespace Strata
 namespace Laurel
 
-abbrev Identifier := String /- Potentially this could be an Int to save resources. -/
+
+/-- A name-introduction site (variable declaration, procedure, field, type, etc.).
+    Carries a mandatory unique ID assigned by the resolution pass. -/
+structure Identifier where
```
joscoh, line 21: This could use `Lambda.Identifier` with `IDMeta` as `Option Nat`, which would avoid needing to add all of these instances.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -253,122 +253,117 @@ def defaultExprForType (ty : HighTypeMd) : Core.Expression.Expr :=
...
-        return (e', acc ++ ss)) (env, [])
-      return (env', stmtsList)
-  | .LocalVariable name ty initializer =>
-      let env' := (name, ty) :: env
-      let boogieMonoType := translateType ty
+      let coreExpr ← translateExpr cond [] (isPureContext := true)
+      return [Core.Statement.assume ("assume" ++ getNameFromMd md) coreExpr md]
+  | .Block stmts _ => stmts.flatMapM (fun s => translateStmt outputParams s)
+  | .LocalVariable id ty initializer =>
+      let boogieMonoType := translateType model ty
```
joscoh: Why Boogie?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -559,24 +553,56 @@ def tryTranslatePureToFunction (proc : Procedure) (initState : TranslateState)
...
-    | .Datatype dt => if dt.name == "Field" then acc ++ dt.constructors.map (·.name) else acc
-    | _ => acc) []
+  let program := { program with
+    staticProcedures := coreDefinitionsForLaurel.staticProcedures ++ program.staticProcedures
+  }
+
+  let result := resolve program
+  let (program, model) := (result.program, result.model)
+  let mut _resolutionDiags := result.errors
+  -- dbg_trace "=== Before validateDiamondFieldAccesses ==="
```
joscoh: Is there a reason these are commented out? I would prefer either `dbg_trace` that is present or else just removing these lines entirely.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -589,31 +615,41 @@ def translate (program : Program) : Except (Array DiagnosticModel) (Core.Program
...
       body := body
     }
 
-  -- Collect ALL errors from both functions and procedures before deciding whether to fail
-  let allErrors := pureErrors ++ procDiags ++ constantsState.diagnostics
+  -- Collect ALL errors from both functions, procedures, and resolution before deciding whether to fail
+  let allErrors :=
+    -- Not including resolution diagnostics yet because the Python through Laurel pipeline
+    -- does not resolve yet.
+    -- resolutionDiags.toList ++
```
joscoh, line 617: Dead code?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -589,31 +615,41 @@ def translate (program : Program) : Except (Array DiagnosticModel) (Core.Program
...
+  let preludeDecls : List Core.Decl := []
 
   -- Translate Laurel datatype definitions to Core datatype declarations
   let laurelDatatypeDecls := program.types.filterMap fun td => match td with
-    | .Datatype dt => some (translateDatatypeDefinition dt)
+    | .Datatype dt => some (translateDatatypeDefinition model dt)
     | _ => none
-  pure ({ decls := laurelDatatypeDecls ++ preludeDecls ++ constantDecls ++ pureFuncDecls.toList ++ procDecls }, modifiesDiags)
+  let program := {
+    decls := laurelDatatypeDecls ++ preludeDecls ++ constantDecls ++ pureFuncDecls.toList ++ procDecls }
```
joscoh: Does this mean that Laurel datatypes cannot reference types defined in the prelude?

Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
keyboardDrummer: > Why do you need to do this?

I've changed the DDM parser so it's no longer needed. `Lean.Name.toString` adds the guillemets so when you parse this output using a Lean parser, it doesn't conflict with Lean keywords, but 1) this escaping should only happen when printing source code, not already when the identifiers are in the AST 2) the DDM isn't meant for just Lean code so it shouldn't be applying Lean specific escaping.

Strata/Languages/Laurel/Grammar/LaurelGrammar.st
keyboardDrummer: Grammar wise I think it would be allowed, but it might fail later in the pipeline, depending on whether Core datatypes support this or not.

Strata/Languages/Laurel/CoreDefinitionsForLaurel.lean
keyboardDrummer: To create source location information for what's parsed from this file, but I will change the code so the source locations for AST nodes from this file are abscent.

Strata/Languages/Laurel/Resolution.lean
keyboardDrummer: No, removed it. Sorry!

Strata/Languages/Laurel/HeapParameterizationConstants.lean
keyboardDrummer: Yes of course. Sorry! Thought I had already commented this.

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: HMM, I prefer this more explicit definition over reducing the code size a bit.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: I'll add it back in the next PR, where we resolve the resolution errors in the Python through Laurel pipeline.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Cleaned up the code and now the question doesn't apply any more since there's no longer a prelude. There are some definitions that are used to define the heap, and these shouldn't be referenced. When we turn on errors for the resolution phase (follow-up PR), hopefully Laurel will prevent you from referencing these :)

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Removed them. The right solution is to add logging to the compilation options

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Fixed

Strata/Languages/Laurel/Grammar/LaurelGrammar.st
joscoh: Core datatypes support this. The Laurel grammar does not right now I believe because the type definition is not in scope in the constructor types.  We use the `@[scopeDatatype]` annotation for this in the Core grammar to handle this. I am not saying that Laurel should or not support this, just wanted to let you know that I do not believe this is supported right now.

Strata/Languages/Laurel/Resolution.lean
joscoh: There are still `partial` functions (`resolveHighType` and `resolveStmtExpr`).

Strata/Languages/Laurel/Grammar/LaurelGrammar.st
keyboardDrummer: I see, thanks.

Strata/Languages/Laurel/HeapParameterizationConstants.lean
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// increment(heap) = MkHeap(Heap..data(heap), Heap..nextReference(heap) + 1)
+function increment(heap: Heap): Heap {
+  MkHeap(Heap..data(heap), Heap..nextReference(heap) + 1)
+}
+
+
+
+datatype Workaround  {}
```
joehendrix, line 71: ~~The `Workaround` datatype can be deleted — I removed it locally and `lake build strata:exe` completed successfully (368/368 jobs, no errors).strata:exe` completed successfully (368/368 jobs, no errors).~~

Scratch that I see the test failure.  I'll investigate further for a fix, but don't consider this blocking obviously.

keyboardDrummer: Thank you!

PR #504, opened 2026-03-03
Fix early return soundness in Laurel: replace `assume false` with `exit`

fabiomadge:
### Problem

Early returns in Laurel were encoded as `result := value; assume false`. The `assume false` made the return path vacuously true, so Core's VCG never checked postconditions on early return paths. A procedure like:

```
procedure buggy(x: int) returns (r: int)
  ensures r >= 0
{
  if (x < 0) { return x; }  // violates postcondition — was not caught
  return x;
}
```

passed verification silently.

### Fix

Replace `assume false` with `exit $body` targeting a labeled body block. The `exit` skips remaining body statements, but postconditions — appended after the block by `ProcedureEval` — are still checked on all paths.

### Changes

- **LaurelToCoreTranslator**: `assume "return" false` → `.exit (some "$body") md`, wrap procedure body in labeled `$body` block
- **TestDiagnostics**: support consecutive annotation comments targeting the same code line (needed for duplicate diagnostics from #419)
- **T8b_EarlyReturnPostconditions**: test proving the buggy early return is now caught
- **T3_ControlFlow**: updated annotations for #419 duplicate diagnostics

### Context

See [PR #385 discussion](https://github.com/strata-org/Strata/pull/385#discussion_r2798679880) for the original analysis of this soundness issue. Duplicate diagnostics from `exit` inside `if` branches are a known Core VCG issue (#419).


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -355,11 +355,9 @@ def translateStmt (env : TypeEnv) (outputParams : List Parameter) (stmt : StmtEx
           let ident := ⟨outParam.name, ()⟩
           let coreExpr ← translateExpr env value
           let assignStmt := Core.Statement.set ident coreExpr md
-          let noFallThrough := Core.Statement.assume "return" (.const () (.boolConst false)) .empty
-          return (env, [assignStmt, noFallThrough])
+          return (env, [assignStmt, .exit none md])
```
keyboardDrummer: Don't you need to specify the target for the exit? What if someone adds another labelled block in between?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -422,11 +420,13 @@ def translateProcedure (proc : Procedure) : TranslateM Core.Procedure := do
...
     | _ => pure []
   let modifies : List Core.Expression.Ident := []
-  let body : List Core.Statement ←
+  let bodyStmts : List Core.Statement ←
     match proc.body with
     | .Transparent bodyExpr => (·.2) <$> translateStmt initEnv proc.outputs bodyExpr
     | .Opaque _postconds (some impl) _ => (·.2) <$> translateStmt initEnv proc.outputs impl
     | _ => pure [Core.Statement.assume "no_body" (.const () (.boolConst false)) .empty]
+  -- Wrap body in a labeled block so early returns (exit) work correctly
+  let body : List Core.Statement := [.block "$body" bodyStmts .empty]
```
keyboardDrummer, line 429: Why do you need the label here if you're never using it?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -653,7 +653,10 @@ def verifyToDiagnostics (files: Map Strata.Uri Lean.FileMap) (program : Program)
   let results <- verifyToVcResults program options
   match results with
   | .error errors => return errors.map (fun dm => dm.toDiagnostic files)
-  | .ok results => return results.filterMap (fun dm => dm.toDiagnostic files)
+  -- Deduplicate: early returns use `exit` inside a block, which causes the VCG
+  -- to explore multiple paths that check the same postcondition, producing
+  -- duplicate diagnostics at the same source location.
+  | .ok results => return (results.filterMap (fun dm => dm.toDiagnostic files)).toList.eraseDups.toArray
```
keyboardDrummer: I'd rather have the duplicate assertion for now and push fixing https://github.com/strata-org/Strata/issues/419. Also in case you keep this, it would be good to mention https://github.com/strata-org/Strata/issues/419 in the comments here.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
fabiomadge: Changed to `exit (some "$body")` to explicitly target the body block.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
fabiomadge: It's now used — `exit (some "$body")` targets it.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
fabiomadge: Removed the dedup. Updated the test framework to support consecutive annotation comments targeting the same code line, so duplicate diagnostics from #419 can be annotated directly. The duplicate annotations reference #419 in a comment.

keyboardDrummer (approved):

keyboardDrummer:
Looks great! Thanks

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
joehendrix: Could you amend your comment in #419 about how this removes duplication?

fabiomadge: I'm not sure I follow. The duplication is still present, but I removed the dedup code. The test works around it by allowing multiple test annotations on a single line.

PR #505, opened 2026-03-03
Laurel: Add constrained type support

fabiomadge:
## Constrained Types for Laurel

Adds constrained types to Laurel via a Laurel-to-Laurel elimination pass that inserts verification checks at type boundaries.

### Syntax

```
constrained nat = x: int where x >= 0 witness 0
constrained posnat = x: nat where x != 0 witness 1
```

### How it works

The pass (`ConstrainedTypeElim.lean`) generates a named constraint function per constrained type (e.g. `function nat$constraint(x: int): bool { x >= 0 }`) and eliminates constrained types by:

1. **`requires constraintFunc(param)`** for constrained-typed inputs — Core handles caller asserts and body assumes via call elimination
2. **`ensures constraintFunc(result)`** for constrained-typed outputs — Core handles body checks and caller assumes
3. **`assert constraintFunc(var)`** after local variable init and reassignment of constrained-typed variables
4. **Witness as default** for uninitialized constrained-typed variables
5. **Witness validation** via synthetic procedures that assert the witness satisfies the constraint function
6. **Quantifier constraint injection** — `forall(n: nat) => body` becomes `forall(n: int) => nat$constraint(n) ==> body`; `exists` uses `&&`
7. **Type resolution** — all constrained type references resolved to base types

The Core translator sees only base types and regular requires/ensures/assert with constraint function calls — no translator changes needed beyond pipeline wiring.

### Functions

Functions (`isFunctional` procedures) with constrained return types emit a "constrained return types on functions are not yet supported" diagnostic. The function is translated as-is with types resolved, but the return constraint is not checked.

### Changes

- `ConstrainedTypeElim.lean` — the elimination pass
- `ConstrainedTypeElimTest.lean` — before/after comparison test and scope regression test
- `LaurelGrammar.st` / `LaurelGrammar.lean` — constrained type syntax
- `ConcreteToAbstractTreeTranslator.lean` — parser for `constrained` keyword
- `LaurelToCoreTranslator.lean` — pipeline wiring (import + pass + resolve + diagnostics)
- `T10_ConstrainedTypes.lean` — 24 test procedures covering inputs, outputs, assignments, arguments, nested types, functions, witnesses, quantifiers, capture avoidance

### Known limitations

- `resolveBaseType` is `partial` — cyclic constrained type definitions loop forever

### DIFF
--- a/StrataTest/Languages/Laurel/ConstrainedTypeElimTest.lean
+++ b/StrataTest/Languages/Laurel/ConstrainedTypeElimTest.lean
@@ -103,7 +103,7 @@ def uninitProgram : String := r"
 constrained posint = x: int where x > 0 witness 1
 procedure f() {
   var x: posint;
-  assert x > 0;
+  assert x == 1;
 };
 "
 
@@ -115,7 +115,7 @@ deterministic
 procedure f() returns ⏎
 ()
 deterministic
-{ var x: int; assume posint$constraint(x); assert x > 0 }
+{ var x: int; assume posint$constraint(x); assert x == 1 }
 procedure $witness_posint() returns ⏎
 ()
 deterministic
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T10_ConstrainedTypes.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T10_ConstrainedTypes.lean
@@ -103,17 +103,24 @@ procedure constrainedInExpr() {
 constrained bad = x: int where x > 0 witness -1
 //                                           ^^ error: assertion does not hold
 
-// Uninitialized constrained variable — witness used as default
+// Uninitialized constrained variable — havoc + assume constraint
 procedure uninitNat() {
   var y: nat;
   assert y >= 0;
 };
 
-// Uninitialized nested constrained variable — outermost witness used
+// Uninitialized nested constrained variable — havoc + assume constraint
 procedure uninitPosnat() {
   var y: posnat;
   assert y != 0;
   assert y >= 0;
+};
+
+// Uninitialized constrained variable — witness value is not provable
+procedure uninitNotWitness() {
+  var y: posnat;
+  assert y == 1;
+//^^^^^^^^^^^^^^ error: assertion does not hold
 };
 
 // Function with valid constrained return — constraint not checked (not yet supported)


StrataTest/Languages/Laurel/Examples/Fundamentals/T09_ConstrainedTypes.lean
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// Uninitialized nested constrained variable — outermost witness used
+procedure uninitPosnat() {
+  var y: posnat;
+  assert y > 0;
+  assert y >= 0;
+}
+
+// Function with valid constrained return — cleared isFunctional, ensures passes
```
keyboardDrummer: Why clear isFunctional instead of reporting this case as not supported?

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -0,0 +1,271 @@
...
+-/
+
+import Strata.Languages.Laurel.Laurel
+import Strata.Languages.Laurel.Resolution
+
+/-!
+# Constrained Type Elimination
+
+A Laurel-to-Laurel pass that eliminates constrained types by:
+1. Adding `requires` for constrained-typed inputs (Core handles caller asserts and body assumes)
```
keyboardDrummer: I think using requires clauses for constrained types has this downside:
1. Unnecessary assertions for calls where the type of the provided value was already constrained

I think using ensures clauses for constrained types has these downsides:
1. Transparent procedures and functions (so procedures and functions that did not already have an ensures clause) should not have ensures clauses. Adding an ensures clause leads to duplication of information since everything about the returned value is already visible from the transparent body.
2. I think this does not extend to polymorphic procedures

My suggestion would be:
- Replace these requires clauses by: 
  - Detect casts from base to constrained types and add assertions there. 
  - Add assumptions at the start of procedures that have constrained parameters
- Replace these ensures clauses by: 
  - for calls to opaque procedures/functions, assume the constraint of any returned values.

StrataTest/Languages/Laurel/Examples/Fundamentals/T09_ConstrainedTypes.lean
fabiomadge: Without clearing `isFunctional`, the translator drops the postcondition with a warning but still translates the function — the constraint is not checked at all (we verified contradictory asserts both pass). Clearing it converts to a procedure with `ensures`, which Core verifies correctly.

Alternative: reject constrained return types on functions entirely. Would you prefer that?

Strata/Languages/Laurel/ConstrainedTypeElim.lean
fabiomadge: Agreed the redundancy is real. The current design trades precision for simplicity — the pass only looks at the callee's signature, never at call sites or caller context, keeping it a straightforward Laurel-to-Laurel transformation.

Your cast-based alternative would produce smaller Core programs and reduce SMT work, but needs cast detection, opaque/transparent call distinction, and call-site walking. OK to keep the simpler approach for this PR and refine in a follow-up, or is the redundancy a blocker?

keyboardDrummer: > Your cast-based alternative would produce smaller Core programs and reduce SMT work, but needs cast detection, opaque/transparent call distinction, and call-site walking

Is that difficult? For every .StaticCall in the program, compare the type of each argument with the parameter to see if there's a cast, and if so insert the assert. It seems not too different in complexity to me. Why did you opt for the current design?

> OK to keep the simpler approach for this PR and refine in a follow-up, or is the redundancy a blocker?

The requires point 1 and ensures point 1 are performance related and not blocking I think, although given that the work seems easy I would prefer to do it now already. Point 2 from ensures clauses means that I think we'll have to change this implementation when we introduce polymorphism.

We can merge this with the design as is, but it means we will have to change the primary part of the implementation soon. I would prefer a solution that's more durable but I'll leave that up to you.

StrataTest/Languages/Laurel/Examples/Fundamentals/T09_ConstrainedTypes.lean
keyboardDrummer: > Without clearing isFunctional, the translator drops the postcondition with a warning but still translates the function — the constraint is not checked at all (we verified contradictory asserts both pass). Clearing it converts to a procedure with ensures, which Core verifies correctly.

I understood that, but why implicitly convert a function to a procedure, which causes confusing side-effects to the user, instead of telling them we do not YET support constrained types on function return types ?

fabiomadge: Done — latest push emits a "constrained return types on functions are not yet supported" diagnostic instead of converting. The function is still translated as-is with types resolved, but the return constraint is not checked.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
fabiomadge: I opted for requires/ensures because it keeps the pass simple — no call-site analysis or model lookups needed. You're right that the cast-based approach is feasible since the `SemanticModel` provides the needed info, and that it would extend to polymorphism where requires/ensures doesn't.

Since you're OK merging as-is, I'd prefer to land this and revisit the design when polymorphism lands. Happy to change course if you feel strongly.

StrataTest/Languages/Laurel/Examples/Fundamentals/T09_ConstrainedTypes.lean
keyboardDrummer: The reason for my question is because it feels to me like this choice in the PR wasn't reviewed before publishing, which I think is something the publisher is responsible for. With LLMs the effort in changing code moves more heavily toward reading what the LLM wrote instead of writing ourselves, and I think it's important that the submitter still does the reading part.

fabiomadge: I made that choice. I felt it was better do to something than nothing, but didn't see a point in arguing.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
keyboardDrummer: What I don't understand is that despite me having provided critical feedback on the requires/ensures design that was in #385 and sharing the assert/assume design beforehand, you still opted for the requires/ensures design. 

The complexity "I opted for requires/ensures because it keeps the pass simple — no call-site analysis or model lookups needed." does not seem significant since this is just traversing the program (which we do all the time) and calling functions that already exist. It feels like a bad use of time to submit and review this temporary approach. I can approve this but I would still like to understand what happened.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -0,0 +1,275 @@
...
+    let stmtss ← stmts.mapM (elimStmt ptMap)
+    pure [⟨.Block stmtss.flatten sep, md⟩]
+
+  | .IfThenElse cond thenBr (some elseBr) =>
+    let thenSs ← elimStmt ptMap thenBr
+    let elseSs ← elimStmt ptMap elseBr
+    pure [⟨.IfThenElse cond (wrap thenSs md) (some (wrap elseSs md)), md⟩]
+  | .IfThenElse cond thenBr none =>
+    let thenSs ← elimStmt ptMap thenBr
+    pure [⟨.IfThenElse cond (wrap thenSs md) none, md⟩]
```
shigoel, line 176: Can the if's branches have any local effects? If so, then some scope management is needed here. Probably the same comment for `While`.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -0,0 +1,275 @@
...
+
+/-- Inject constraints into a quantifier body for a constrained type -/
+private def injectQuantifierConstraint (ptMap : ConstrainedTypeMap) (ty : HighType)
+    (varName : Identifier) (body : StmtExprMd) (isForall : Bool) : StmtExprMd :=
+  let constraints := getAllConstraints ptMap ty
+  match constraints with
+  | [] => body
+  | _ =>
+    let preds := constraints.map fun (vn, pred) => substId vn varName pred
+    let conj := preds.tail.foldl (init := preds.head!) fun acc p =>
```
shigoel: I'd recommend using a non-panicing version of `head!` here. Looks like this is derivable because constraints are not empty.

fabiomadge: I replaced it with a pattern match.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
fabiomadge: Laurel rejects variable redeclaration, so a constrained variable declared in a branch can't shadow an outer variable.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
shigoel: Thanks!

Strata/Languages/Laurel/ConstrainedTypeElim.lean
shigoel: More than shadowing, I was wondering about the lifetime of a local variable declared in a branch. Does that leak into the outer scope?

shigoel: For instance, I was expecting to see a `withScope` for IfThenElses and Whiles in https://github.com/strata-org/Strata/blob/main/Strata/Languages/Laurel/Resolution.lean#L274, but I don't.

shigoel:
Here is a counterexample demonstrating a scope management bug in `elimStmt` (added to `StrataTest/Languages/Laurel/ConstrainedTypeElimTest.lean`):

```lean
/-
Counterexample demonstrating the scope management bug in `elimStmt`.

The `PredVarMap` (which tracks which variables have constrained types) is
threaded through sibling blocks via `mapM` in the `Block` case, but is never
saved/restored when entering/exiting an `IfThenElse` branch or a `Block`.

The resolution pass uses `withScope` inside `Block`, so two sibling blocks can
legitimately declare variables with the same *text* name but distinct uniqueIds.
The elimination pass's `PredVarMap` is keyed by `name.text`, so after the first
block adds `x: pos`, that entry persists (stale) when the second sibling block
declares a plain `x: int` and later assigns to it.

Bug: the assignment `x := -10` in the second block (where x: int) incorrectly
receives a spurious `assert x > 0` because the stale PredVarMap entry for the
first `x` (x: pos) is still present.

Expected: no assert after `x := -10` (x is plain int in that block).
Actual:   a spurious `assert x > 0` is inserted.
-/
def counterexampleProgram : String := r"
constrained pos = v: int where v > 0 witness 1
procedure test(b: bool) {
  if (b) {
    var x: pos := 1;
  }
  {
    var x: int := -5;
    x := -10;
  }
};
"

-- The CORRECT output should have no assert after `x := -10`:
--   { if b then { var x: int := 1; assert x > 0 }; { var x: int := -5; x := -10 } }
-- The ACTUAL (buggy) output inserts a spurious `assert x > 0` after `x := -10`:
--   { if b then { var x: int := 1; assert x > 0 }; { var x: int := -5; x := -10; assert x > 0 } }
/--
info: procedure test(b: bool) returns
()
deterministic
{ if b then { var x: int := 1; assert x > 0 }; { var x: int := -5; x := -10; assert x > 0 } }
procedure $witness_pos() returns
()
deterministic
{ var $witness: int := 1; assert $witness > 0 }
-/
#guard_msgs (whitespace := lax) in
#eval! do
  let program ← parseLaurelAndElim counterexampleProgram
  for proc in program.staticProcedures do
    IO.println (toString (Std.Format.pretty (Std.ToFormat.format proc)))
```

The key point: the no-redeclaration property only prevents two variables with the same name from coexisting in a *single live scope*. The resolution pass's `withScope` inside `Block` means sibling blocks can each declare their own `x` with distinct `uniqueId`s — this is not redeclaration. The `PredVarMap` keyed by `name.text` conflates them.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
shigoel: Claude came up with the following counterexample, btw: see https://github.com/strata-org/Strata/pull/505#issuecomment-4035342659

fabiomadge: Thanks for calling this out! I updated the definition and included your test.

shigoel: Thanks!

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -0,0 +1,284 @@
...
+7. Resolving all constrained type references to their base types
+-/
+
+namespace Strata.Laurel
+
+open Strata
+
+abbrev ConstrainedTypeMap := Std.HashMap String ConstrainedType
+/-- Map from variable name to its constrained HighType (e.g. UserDefined "nat") -/
+abbrev PredVarMap := Std.HashMap String HighType
```
keyboardDrummer, line 32: We should use the `model: SemanticModel` to lookup types and definitions like HighType and ConstrainedType, instead of implementing a form of name binding in this pass.

~This change should also fix the bug that @shigoel mentioned.~ prevent the possibility for bugs like that

Strata/Languages/Laurel/ConstrainedTypeElim.lean
MikaelMayer: The approach of this PR has the upside that it does not introduce assumptions when translating to Core, and that it is simple to conceptualize.

We still want to address the performance penalty of having to prove trivial assertions.
For that, I would like to evaluate how the traversal you suggest @keyboardDrummer  could actually be used as a transformation from Core to Core directly. Broadly speaking, this transformation consists of assuming we know all call sites of a given procedure, eliminating the procedure calls, and then eliminating assertions that we can prove immediately true by traversal + partial evaluation of assertions that make use of abstract path condition. It would not be only useful for the subtypes of this PR, but also for subtypes encoded differently, and possibly other preconditions that are assumed earlier in the path condition.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -0,0 +1,284 @@
...
+
+/-!
+# Constrained Type Elimination
+
+A Laurel-to-Laurel pass that eliminates constrained types by:
+1. Adding `requires` for constrained-typed inputs (Core handles caller asserts and body assumes)
+2. Adding `ensures` for constrained-typed outputs (Core handles body checks and caller assumes)
+   - Skipped for `isFunctional` procedures since the Laurel translator does not yet support
+     function postconditions. Constrained return types on functions are not checked.
+3. Inserting `assert` for local variable init and reassignment of constrained-typed variables
```
MikaelMayer: If we have a constraint, instead of substituting the constraint with the provided variable - which is tedious, I would prefer to see a proper function call. That is, the constraint gives rise to a function returning a boolean. Not only it will help for triggers, but also for proofs later on.
To keep the current behavior with this approach, you can mark this function as inline so that it's automatically inlined when building verification conditions. But I would prefer to avoid doing substitution if we can do it automatically already in Core. It will also make the generated Core code clearer.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -0,0 +1,284 @@
...
+partial def getAllConstraints (ptMap : ConstrainedTypeMap) (ty : HighType)
+    : List (Identifier × StmtExprMd) :=
+  match ty with
+  | .UserDefined name => match ptMap.get? name.text with
+    | some ct => (ct.valueName, ct.constraint) :: getAllConstraints ptMap ct.base.val
+    | none => []
+  | _ => []
+
+/-- Substitute `Identifier old` with `Identifier new` in a constraint expression -/
+partial def substId (old new : Identifier) : StmtExprMd → StmtExprMd
```
MikaelMayer: The approach I suggest above will get rid of this substitution function and make it much clearer in the generated Core code what we are actually checking, without incurring performance penalty if the function is marked as inline.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -0,0 +1,284 @@
...
+    outputs := []
+    body := .Transparent ⟨.Block ([witnessInit] ++ asserts) none, md⟩
+    preconditions := []
+    isFunctional := false
+    determinism := .deterministic none
+    decreases := none
+    md := md }
+
+/-- Eliminate constrained types from a Laurel program.
+    The `witness` field is used as the default initializer for uninitialized
```
MikaelMayer: Since the base type is always inhabited after this elimination phase, we don't need to initialize variables with constrained type, we can just assume the constraint, can't we? That way the code we emit does not depend on the witness. We would use the witness only to prove the type is inhabited, so that we never create procedures which would be vacuously true because the type.

Also, looking forward based on what we had in Dafny, it's sometimes very tedious to build a witness and the proof that the witness is a witness does not always goes through automatically. So since we have this one chance of reinventing witnesses from scratch, I would prefer to see a **witness statement** instead of a witness expression.

For example:
```
constrainted nat := x: int | x >= 0 witness {
  x := 1;
  assert x >= 0; // Optional, or anything that can help the proof go through.
}
````

It makes it easier to determine what is the proof obligation (x >= 0), makes it possible to compute the witness using any statements, not just expressions, and also inline elements that can help the witness.
Plus, the block will just become the body of the witness check procedure:

```
procedure witnessNat() returns (x: int)
  ensures x >= 0 {
  x := 1;
  assert x >= 0;
}
````

An evaluator could always evaluate witness procedures to get a witness if necessary, but for verification witnesses should not be needed.

I recommend this approach in the Laurel AST rather than storing an expression, especially if we get rid of injecting the witness at every initialization site.

File: StrataTest/Languages/Laurel/Examples/Fundamentals/T10_ConstrainedTypes.lean [hunk 9]
```diff
@@ -8,23 +8,161 @@ import StrataTest.Util.TestDiagnostics
...
 
 open StrataTest.Util
-open Strata
 
+namespace Strata
 namespace Laurel
 
 def program := r"
 constrained nat = x: int where x >= 0 witness 0
+constrained posnat = x: nat where x > 0 witness 1
```
MikaelMayer:
```suggestion
constrained posnat = x: nat where x != 0 witness 1
```
Just to make sure we test that the condition of nat is included whenever we check something is posnat.

StrataTest/Languages/Laurel/ConstrainedTypeElimTest.lean
```diff
@@ -0,0 +1,91 @@
...
+
+open Strata
+open Strata.Elab (parseStrataProgramFromDialect)
+
+namespace Strata.Laurel
+
+def testProgram : String := r"
+constrained nat = x: int where x >= 0 witness 0
+procedure test(n: nat) returns (r: nat) {
+  var y: nat := n;
```
MikaelMayer, line 28:
```suggestion
  assert r >= 0;
  var y: nat := n;
```
Does it make sense to verify that uninitialized output variables satisfy their constraint?

Strata/Languages/Laurel/ConstrainedTypeElim.lean
keyboardDrummer: > I would prefer to see a witness statement instead of a witness expression.

Isn't everything a StmtExpr in Laurel? What's the benefit of assigning to `x` instead of returning that value?

Strata/Languages/Laurel/ConstrainedTypeElim.lean
fabiomadge: Done — latest push generates named constraint functions (e.g. `nat$constraint`) and uses them in requires/ensures/asserts. This eliminates `substId` and its capture avoidance complexity.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
fabiomadge: The translator initializes uninitialized variables to 0, not havoc. An `assume` on a variable with value 0 that doesn't satisfy the constraint (e.g. `x > 0`) makes the path infeasible. Witness injection sets the variable to a satisfying value instead.

Witness statements — happy to explore in a follow-up.

StrataTest/Languages/Laurel/ConstrainedTypeElimTest.lean
fabiomadge: Done — added `assert r >= 0` before the body in the elim test. The before/after output confirms the ensures is generated alongside the user assert.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
fabiomadge: The `SemanticModel` is keyed by `uniqueId` (`Nat`), but we need name-based lookup for two reasons: our generated identifiers start with `uniqueId = none` (only assigned by subsequent resolution), and we need to follow the type chain transitively (`posnat → nat → int`) by name. The `ConstrainedTypeMap` gives us direct name-based access for both.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
MikaelMayer: > The translator initializes uninitialized variables to 0, not havoc. An assume on a variable with value 0 that doesn't satisfy the constraint (e.g. x > 0) makes the path infeasible

The translator should not initialize uninitialized variables. You don't need that. Especially if the witness is a complex expression, you are going to regret it later.

Now that Core's init statements support having no right-hand side, you should use that instead, so the constrained type elimination phase should simply change the underlying type and emit the constraint assumption. That way you can keep uninitialized variables uninitialized as long as you prove that they have a witness.
You don't need to worry about executability because the executable cases we care about usually don't involve branching over or building constructs with uninitialized variables.

So here is what I recommend:
Line 152 of Strata/Languages/Laurel/ConstrainedTypeElim.lean : 

Just don't add a right-hand-side to that, and it will be perfectly fine, and it will avoid overwhelming the SMT solver with a witness (I've seen witnesses that were pretty complicated) - and also it will avoid proving the following:
```
constrained nat = x: int where x >= 0 witness 0
procedure f() {
  var x: nat;
  assert x == 0; // Proved in your branch, but I don't think this is expected, what if the source language produced 1?
}
```
which I recommend should be added in your benchmarks, so that we know the behavior.

fabiomadge: Good point — the witness injection makes `assert x == 0` provable, which is wrong. The default initialization comes from `defaultExprForType` in the translator (#435), which converts `var x: int;` to `init x int 0` instead of havoc. Once the translator is fixed to emit `init` without RHS, we can switch from witness injection + assert to assume. I'll add a TODO and address in a follow-up.

MikaelMayer (approved):

Strata/Languages/Laurel/ConstrainedTypeElim.lean
keyboardDrummer: > our generated identifiers start with uniqueId = none (only assigned by subsequent resolution)

yes, so you would run resolution before running this pass

> and we need to follow the type chain transitively

yes, why can't do that using the model?

keyboardDrummer (approved):

shigoel (approved):

Strata/Languages/Laurel/ConstrainedTypeElim.lean
shigoel: I assume you're tracking these TODOs internally, but if not, feel free to use issues.

StrataTest/Languages/Laurel/ConstrainedTypeElimTest.lean
```diff
@@ -97,14 +97,13 @@ deterministic
...
--- TODO: Once the translator emits havoc for uninitialized variables (#550),
--- switch to assume instead of witness injection + assert.
+-- Uninitialized constrained variable: havoc + assume constraint.
+-- The variable has no known value, only the type constraint is assumed.
 def uninitProgram : String := r"
 constrained posint = x: int where x > 0 witness 1
 procedure f() {
   var x: posint;
-  assert x == 1;
+  assert x > 0;
```
MikaelMayer: I would have kept this test the same to verify we can't prove the variable is equal to the witness.

PR #506, opened 2026-03-03
Define Python Prelude in Laurel instead of Core

keyboardDrummer:
Putting this in draft for now, because my impression is that since the current PySpecs are translated to Core instead of Python or Laurel, it is not possible to reliably turn on resolution error reporting for Laurel when using Python.

---- 

This PR turns on resolution error reporting in Laurel, and works towards removing the need for the `TCore` type in Laurel. It conflicts with changes from another in-progress PR, https://github.com/strata-org/Strata/pull/489. I think that one will need more time to be ready for merging, so I think it makes sense to review/merge this one first. I will happily resolve conflicts in 489 after this is merged.

### Changes
1. In the Python through Laurel pipeline, replace the use of `CorePrelude.lean` with `PythonRuntimeLaurelPart.lean` and `PythonRuntimeCorePart.lean`. The CorePart is needed because Python requires axioms in Core and Laurel does not support axioms. I think we should not add axioms to Laurel. A long term solution would be that the Python pipeline stops relying on axioms.
2. Turn on reporting of resolution errors
3. Add `--update` option to run_py_analyze.sh to update the test expect files (the bytecode offsets in them change often)
4. Update Laurel grammar to support multiple preconditions

#### Caveats
- Because DDM parsing also does a form of resolution, `PythonRuntimeCorePart.lean` needs a lot of type definitions that are already provided in `PythonPreludeInLaurel.lean`. Ideally we'd find a way so that's no longer necessary, using something like `#strata_parsing_only`. @joehendrix is that something we could add? I looked into adding that but it wasn't so simple.

### Tested
1. Updated Python expect files for Laurel because the reported locations are different
2. No tests for resolution errors added yet. I will do so in a future PR and I expect this will trigger the need for some fixes as well.
3. Manually used the --update option for `run_py_analyze.sh`
4. Update T6_Preconditions with tests that check multiple preconditions. Tests suffer from #531 
5. Bonus: Added T16_Datatypes.lean test for Laurel datatypes, something that was missing in previous PRs

### DIFF
--- a/StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
@@ -6,13 +6,13 @@ PFloorDiv_body_calls_Int.SafeDiv_2: ✅ pass
 PFloorDiv_body_calls_Int.SafeDiv_3: ✅ pass
 PAnd_body_calls_Any_to_bool_0: ✅ pass
 POr_body_calls_Any_to_bool_0: ✅ pass
-ret_type: ✅ pass (at byte 25063)
-ret_type: ✅ pass (at byte 25478)
-ret_pos: ✅ pass (at byte 25964)
-assert_name_is_foo: ✅ pass (at byte 26968)
-assert_opt_name_none_or_str: ✅ pass (at byte 27031)
-assert_opt_name_none_or_bar: ✅ pass (at byte 27135)
-ensures_maybe_except_none: ✅ pass (at byte 26891)
+postcondition: ✅ pass (at byte 26210)
+postcondition: ✅ pass (at byte 26625)
+postcondition: ✅ pass (at byte 27125)
+assert(28134): ✅ pass (at byte 28134)
+assert(28211): ✅ pass (at byte 28211)
+assert(28320): ✅ pass (at byte 28320)
+postcondition: ✅ pass (at byte 28071)
 assert_assert(102)_calls_Any_to_bool_0: ✅ pass (at line 7, col 4)
 assert(102): ✅ pass (at line 7, col 4)
 assert_assert(226)_calls_Any_to_bool_0: ✅ pass (at line 12, col 4)
--- a/StrataTest/Languages/Python/expected_laurel/test_class_decl.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_class_decl.expected
@@ -6,10 +6,10 @@ PFloorDiv_body_calls_Int.SafeDiv_2: ✅ pass
 PFloorDiv_body_calls_Int.SafeDiv_3: ✅ pass
 PAnd_body_calls_Any_to_bool_0: ✅ pass
 POr_body_calls_Any_to_bool_0: ✅ pass
-ret_type: ✅ pass (at byte 25063)
-ret_type: ✅ pass (at byte 25478)
-ret_pos: ✅ pass (at byte 25964)
-assert_name_is_foo: ✅ pass (at byte 26968)
-assert_opt_name_none_or_str: ✅ pass (at byte 27031)
-assert_opt_name_none_or_bar: ✅ pass (at byte 27135)
-ensures_maybe_except_none: ✅ pass (at byte 26891)
+postcondition: ✅ pass (at byte 26210)
+postcondition: ✅ pass (at byte 26625)
+postcondition: ✅ pass (at byte 27125)
+assert(28134): ✅ pass (at byte 28134)
+assert(28211): ✅ pass (at byte 28211)
+assert(28320): ✅ pass (at byte 28320)
+postcondition: ✅ pass (at byte 28071)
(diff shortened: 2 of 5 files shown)


thanhnguyen-aws:
We will use the "PythonLaurelCorePrelude" in this PR https://github.com/strata-org/Strata/pull/489 for Python-> Laurel translation instead of the "CorePrelude".

joehendrix:
### Request: Handle unresolved identifiers gracefully in `SemanticModel.isFunction`

`Resolution.lean:116-122` — The PR removes the hack that returned `true` for unresolved identifiers (`uniqueId == none`), which is the right move now that prelude declarations resolve properly. However, the remaining `panic!` fallback is now reachable for unresolved callees:

```lean
def SemanticModel.isFunction (model: SemanticModel) (id: Identifier): Bool :=
  match model.get id with
    | .staticProcedure proc => proc.isFunctional
    | .parameter _ => true
    | .datatypeConstructor _ _ => true
    | .constant _ => true
    | node => panic! s!"id: {repr id}, is not a procedure, node {repr node}"
```

When resolution fails for an identifier, it gets `uniqueId := none`. `model.get` returns `default` for those, and if the default `AstNode` doesn't match any of the listed patterns, the `panic!` fires. This can happen because `translate` in `LaurelToCoreTranslator.lean` continues through the full translation pipeline (which calls `isFunction` at lines 189, 312, 336, 369) before checking collected resolution errors at line 585.

Could you either:
- Replace the `panic!` with a graceful default (e.g., `| _ => false`) and optionally log a diagnostic, or
- Rename to something like `isFunction!` / `isFunction_unsafe` to make the partial nature explicit

The first option is preferred since it allows the pipeline to finish collecting all errors before reporting, rather than crashing on the first unresolved callee.

joehendrix:
### Suggestion: Add test case for unsafe destructor resolution

The PR registers unsafe destructors (`DT..field!`) during resolution (`Resolution.lean:488-489`):

```lean
_ ← defineName p.name (.parameter p) (some $ dt.name.text ++ ".." ++ p.name.text ++ "!")
```

And the Laurel prelude uses them, e.g. in `PythonRuntimeLaurelPart.lean:207`:

```laurel
days_i := IntOrNone..int_val!(days);
```

However, no Laurel-level test directly exercises this syntax. The new `T16_Datatypes.lean` tests safe destructors (`IntList..head(xs)`) but not the unsafe variant. Consider adding a case:

```laurel
procedure testUnsafeDestructor() {
  var xs: IntList := Cons(42, Nil());
  assert IntList..isCons(xs);
  var h: int := IntList..head!(xs);
  assert h == 42;
}
```

This would verify that the `!` suffix is correctly parsed, resolved, and translated to Core.

Strata/Languages/Python/PythonRuntimeCorePart.lean
```diff
@@ -0,0 +1,175 @@
...
+that precede the `type CoreOnlyDelimiter;` sentinel (and the sentinel itself).
+Everything after the delimiter is a genuine Core-only declaration.
+-/
+def coreOnlyFromRuntimeCorePart : List Core.Decl :=
+  let decls := pythonRuntimeCorePart.decls
+  -- Drop everything up to and including the CoreOnlyDelimiter sentinel
+  match decls.dropWhile (fun d => d.name.name != "CoreOnlyDelimiter") with
+  | _ :: rest => rest   -- drop the delimiter itself
+  | [] => panic! "CoreOnlyDelimiter sentinel not found in pythonRuntimeCorePart"
+
```
joehendrix, line 73: If the `CoreOnlyDelimiter` sentinel is accidentally removed or renamed, this panics at runtime. A compile-time guard right here would catch that immediately:

```suggestion
  | [] => panic! "CoreOnlyDelimiter sentinel not found in pythonRuntimeCorePart"

-- Verify at compile time that the sentinel exists and Core-only declarations follow it.
#guard coreOnlyFromRuntimeCorePart.length > 0
```

Strata/Languages/Laurel/TypeHierarchy.lean
```diff
@@ -260,8 +260,8 @@ def rewriteTypeHierarchyExpr (exprMd : StmtExprMd) : THM StmtExprMd :=
   | .InstanceCall t callee args => do
       let args' ← args.attach.mapM fun ⟨a, _⟩ => rewriteTypeHierarchyExpr a
       return ⟨.InstanceCall (← rewriteTypeHierarchyExpr t) callee args', md⟩
-  | .Forall p b => do return ⟨.Forall p (← rewriteTypeHierarchyExpr b), md⟩
-  | .Exists p b => do return ⟨.Exists p (← rewriteTypeHierarchyExpr b), md⟩
+  | .Forall p trigger b => do return ⟨.Forall p (← trigger.attach.mapM (fun pv => have := pv.property; rewriteTypeHierarchyExpr pv.val)) (← rewriteTypeHierarchyExpr b), md⟩
+  | .Exists p trigger b => do return ⟨.Exists p (← trigger.attach.mapM (fun pv => have := pv.property; rewriteTypeHierarchyExpr pv.val)) (← rewriteTypeHierarchyExpr b), md⟩
```
joehendrix, line 264: These lines are 172 chars each (hard limit is 100). Extracting the trigger transformation into a `let` binding also gives Lean the membership proof via destructuring, which means no explicit `decreasing_by` clause is needed for the trigger.

```suggestion
  | .Forall p trigger b => do
      let trigger' ← trigger.attach.mapM fun ⟨t, _⟩ => rewriteTypeHierarchyExpr t
      return ⟨.Forall p trigger' (← rewriteTypeHierarchyExpr b), md⟩
  | .Exists p trigger b => do
      let trigger' ← trigger.attach.mapM fun ⟨t, _⟩ => rewriteTypeHierarchyExpr t
      return ⟨.Exists p trigger' (← rewriteTypeHierarchyExpr b), md⟩
```

PR #507, opened 2026-03-03
feat: Add statement-level type declarations

MikaelMayer:
Add statement-level type declarations to Strata Core

This PR adds support for declaring uninterpreted types as statements within procedures, similar to how function declarations work at the statement level.

## Motivation

To support translation to CoreSMT (a subset of Core that maps 1-1 with SMT), we need the ability to declare uninterpreted types that can be mapped to SMT's `declare-sort`. Statement-level type declarations enable this while maintaining proper scoping.

## Changes

- Test cases for statement-level type declarations covering:
  - Basic uninterpreted types
  - Polymorphic types with multiple parameters
  - Constructor and accessor functions with axioms
  - Type scoping within procedures
  - Integration with function declarations
  - Type mismatch error detection

## Implementation Plan

1. ✅ Create test cases (this commit)
2. Add grammar support for type declaration statements
3. Extend Statement AST with type declaration constructor
4. Implement translation from DDM to Core
5. Add type checking support
6. Update statement evaluation and semantics
7. Ensure all tests pass

### DIFF
--- a/Strata/DL/Imperative/Stmt.lean
+++ b/Strata/DL/Imperative/Stmt.lean
@@ -6,7 +6,7 @@
 module
 
 public import Strata.DL.Imperative.Cmd
-public import Strata.DL.TypeConstructor
+public import Strata.DL.Lambda.TypeConstructor
 
 namespace Imperative
 
--- a/Strata/DL/Lambda/TypeConstructor.lean
+++ b/Strata/DL/Lambda/TypeConstructor.lean
@@ -14,9 +14,6 @@ public section
 /-! # Type Constructor
 
 A type constructor declaration that can be shared across dialects.
-This file lives under `DL/` rather than `DL/Lambda/` because type
-constructors are dialect-independent; the Lambda import is only needed
-for `LTy`/`LMonoTy` in `toType`.
 -/
 
 inductive Boundedness where
--- a/Strata/Languages/Core/Statement.lean
+++ b/Strata/Languages/Core/Statement.lean
@@ -13,7 +13,7 @@ import Strata.Languages.Core.Factory
 import Strata.DL.Imperative.Stmt
 import Strata.DL.Imperative.HasVars
 import Strata.DL.Lambda.LExpr
-import Strata.DL.TypeConstructor
+import Strata.DL.Lambda.TypeConstructor
 import Strata.Util.Tactics
 
 namespace Core
--- a/Strata/Languages/Core/TypeDecl.lean
+++ b/Strata/Languages/Core/TypeDecl.lean
@@ -7,7 +7,7 @@
 
 
 import Strata.Languages.Core.Statement
-import Strata.DL.TypeConstructor
+import Strata.DL.Lambda.TypeConstructor
 
 ---------------------------------------------------------------------
 


Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -865,6 +865,18 @@ partial def stmtToCST {M} [Inhabited M] (s : Core.Statement)
...
       pure (.exit_unlabeled_statement default)
   | .funcDecl decl _md => funcDeclToStatement decl
+  | .typeDecl tc _md =>
+    let nameAnn : Ann String M := ⟨default, tc.name⟩
+    let args : Ann (Option (Bindings M)) M :=
+      if tc.numargs == 0 then
+        ⟨default, none⟩
+      else
+        let bindings := List.range tc.numargs |>.map fun i =>
+          let argName : Ann String M := ⟨default, s!"_ty{i}"⟩
```
MikaelMayer: could we store the type arguments names as part of this PR? That way we wouldn't need to generate these weird identifiers.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -230,13 +230,13 @@ def typeConToCST {M} [Inhabited M] (tcons : TypeConstructor)
...
   modify (·.addGlobalFreeVars #[name.val])
   let args : Ann (Option (Bindings M)) M :=
-    if tcons.numargs = 0 then
+    if tcons.params.isEmpty then
       ⟨default, none⟩
     else
-      let bindings := List.range tcons.numargs |>.map fun i =>
-        let paramName : Ann String M := ⟨default, mkParamName i⟩
+      let bindings := tcons.params.map fun paramName =>
+        let paramNameAnn : Ann String M := ⟨default, paramName⟩
```
MikaelMayer: is mkParamName still used? One use was removed by this commit.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -852,6 +852,18 @@ partial def stmtToCST {M} [Inhabited M] (s : Core.Statement)
     | none =>
       pure (.exit_unlabeled_statement default)
   | .funcDecl decl _md => funcDeclToStatement decl
+  | .typeDecl tc _md =>
+    let nameAnn : Ann String M := ⟨default, tc.name⟩
+    let args : Ann (Option (Bindings M)) M :=
+      if tc.numargs == 0 then
+        ⟨default, none⟩
+      else
+        let bindings := tc.params.map fun paramName =>
```
MikaelMayer: Is duplicated? It seems it appears twice in the same file. See above.

Strata/DL/Lambda/TypeConstructor.lean
```diff
@@ -0,0 +1,44 @@
...
+A type constructor declaration that can be shared across dialects.
+-/
+
+inductive Boundedness where
+  | Finite
+  | Infinite -- Default
+  deriving Repr
+
+structure TypeConstructor where
+  -- (TODO) Add SMT support for Boogie's Finite types.
```
joscoh, line 25: Should not mention Boogie

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -332,6 +332,11 @@ op funcDecl_statement (name : Ident,
                        inline? : Option Inline) : Statement =>
   inline? "function " name typeArgs b " : " r indent(2, preconds) " { " body " }\n";
 
+// Type declaration statement
+@[declareScopedType(name, some args)]
```
joscoh, line 336: What does the `@[declareScopedType]` do here?

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
```diff
@@ -0,0 +1,133 @@
...
+---
+info:
+Obligation: trans
+Property: assert
+Result: ✅ pass
+-/
+#guard_msgs in
+#eval verify typeDeclStmt1
+
+/-- Type scoping - type declared in one procedure not visible in another -/
```
joscoh: Why does this test scoping? Shouldn't it have something in `P2` try to reference `T` and hit an error?

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
```diff
@@ -0,0 +1,133 @@
...
+
+---
+info:
+Obligation: reflexive
+Property: assert
+Result: ✅ pass
+-/
+#guard_msgs in
+#eval verify typeDeclStmt3
+
```
joscoh, line 204: Could you add tests for a type declaration with arguments, including one that succeeds and at least one that gives a type error e.g.
```
procedure P () returns () {
  type T a b;
  var x : T int int;
  var y : T bool bool;
  assert (x != y);
}
```

.

Strata/DL/Imperative/StmtSemantics.lean
joscoh, line 1: I think that the small step semantics should also be updated.

Strata/Languages/Core/DDMTransform/Grammar.lean
MikaelMayer: I defined it so that the DDM can recognize this definition as a type position in the same scope.

Strata/DL/Lambda/TypeConstructor.lean
MikaelMayer: I know we renamed it to Core, but I'm not sure about whether the finite types that are mentioned are already in Core somehow or if they still need to be imported from Boogie. Anyway, this comment was already there before, it has just been refactored, let me know if you have a better suggestion.

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
MikaelMayer: Good catch. P2 should be identical to P1 and not hit an error. I remember precisely that I had done this originally, not sure why this test was changed to something meaningless.

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
MikaelMayer: Shoot I swear these tests were there. I should have checked the "consolidation" step of the agent.

Strata/DL/Imperative/Stmt.lean
```diff
@@ -44,6 +45,8 @@ inductive Stmt (P : PureExpr) (Cmd : Type) : Type where
   | exit     (label : Option String) (md : MetaData P)
   /-- A function declaration within a statement block. -/
   | funcDecl (decl : PureFunc P) (md : MetaData P)
+  /-- A type declaration within a statement block. -/
+  | typeDecl (tc : TypeConstructor) (md : MetaData P)
```
shigoel, line 49: Just checking: do you have any plans to go beyond uninterpreted types here? E.g., type synonyms, ADTs?

shigoel: More generally, I'm wondering what your thoughts are re. attaining feature parity between current top-level declarations and statement-level ones, and whether that's even important.

Strata/DL/Imperative/Stmt.lean
```diff
@@ -136,6 +143,7 @@ def Stmt.noFuncDecl (s : Stmt P C) : Bool :=
   | .loop _ _ _ bss _ => Block.noFuncDecl bss
   | .exit _ _ => true
   | .funcDecl _ _ => false
+  | .typeDecl _ _ => true  -- Type declarations are allowed
```
shigoel:
```suggestion
  | .typeDecl _ _ => true
```
Extraneous comment

Strata/DL/Imperative/StmtSemantics.lean
shigoel: +1

Strata/DL/Lambda/TypeConstructor.lean
shigoel: We could do away with this entire `bound` field altogether, but we could also do it in another PR.

joscoh: Oh I see, sorry. I thought this was a TODO from when we wanted Core to encompass all of Boogie's features, including these finite types. Is that still on the roadmap? In any case, I guess it is fine to keep, it just might be a stale comment.

Strata/DL/Lambda/TypeConstructor.lean
shigoel, line 1: This file imports Lambda but is put directly under DL.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -72,9 +72,6 @@ section ToCST
 /-- Constants for consistent naming -/
 def unknownTypeVar : String := "$__unknown_type"
 
-/-- Generate parameter names efficiently -/
```
shigoel, line 75: Nice!

Strata/Languages/Core/StatementType.lean
```diff
@@ -179,6 +179,14 @@ where
           catch e =>
             .error (errorWithSourceLoc e md)
 
+        | .typeDecl tc md => do try
+          -- Add the type to the context
```
shigoel: Want to add a little comment about shadowing? I realized we haven't elsewhere, but we really should've. :)

Strata/DL/Imperative/Stmt.lean
MikaelMayer: Anything that can be done at the top-level, should be permissible at the statement level at some point. So I'd love to keep the two in sync as possible.

Strata/DL/Lambda/TypeConstructor.lean
```diff
@@ -0,0 +1,44 @@
...
+  bound    : Boundedness := .Infinite
+  name     : String
+  params   : List String
```
shigoel, line 28: What does `type Foo (a : Type) (a : Type)` mean? Does that mean that we can instantiate `a` with the same type only? Should we disallow such duplicate var names?

Strata/DL/Imperative/Stmt.lean
MikaelMayer: I think that's important because statement provide implicit entry point, ordering of declarations, whereas our top-level is a set of declarations that needs to be resolved, graph dependencies to be made, etc.

Whereas we can state with a statement what it means for a procedure to be an entry point (just have non-deterministic calls to these procedures), for top-level we don't have this capability yet and we will have to extend the AST for that (typically using private or public modifiers).

Strata/DL/Lambda/TypeConstructor.lean
MikaelMayer: Until we have a RHS for these types, for now, variables are only pretty names and discarded in our pipeline. There is no ambiguity because of their position and we never refer to them by name (or by any other way).

I would start disallowing duplicate var names as soon as they are used for semantics, but for now, they are just hints and don't correlate type arguments if they are the same.

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
shigoel, line 1: Can we have a test case (in the Core/Tests/ directory) that shows that a top-level type `T` cannot be shadowed by a local one?

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
```diff
@@ -0,0 +1,169 @@
...
+---
+info:
+Obligation: trans
+Property: assert
+Result: ✅ pass
+-/
+#guard_msgs in
+#eval verify typeDeclStmt1
+
+/-- Type scoping - same type name in different procedures -/
```
joscoh, line 53: I would still like a test that a type declared in one procedure is not visible in another.

Strata/Languages/Core/DDMTransform/Grammar.lean
joscoh: I guess I am confused why the args need to be part of this, because it seems like they are ignored/never referenced?

Strata/Languages/Core/StatementType.lean
MikaelMayer: Added in https://github.com/strata-org/Strata/pull/507/commits/05970a49c3548a657a48502cae7709b6c600d34a

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
MikaelMayer: See the new test `shadowTopLevelType`

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
MikaelMayer: See the new test `typeScopeError`

Strata/DL/Lambda/TypeConstructor.lean
MikaelMayer: Moved to lambda

StrataTest/Languages/Core/Examples/TypeDeclStmt.lean
MikaelMayer: Added it in the same test file as I think it's better to have them all here.

Strata/Languages/Core/DDMTransform/Grammar.lean
MikaelMayer: Oh that's fair, since type declarations don't have a RHS (yet). Would you be ok if I still leave it like that?

Strata/DL/Lambda/TypeConstructor.lean
shigoel: Indeed, and I remember Boogie does the same. However, it's worth at least documenting so avoid this confusion.

Strata/Languages/Core/DDMTransform/Grammar.lean
joscoh: Yes that's fine, I just wanted to make sure I wasn't missing anything

shigoel (approved):
Thank you!

PR #510, opened 2026-03-03
Add division-by-zero verification for Laurel via Core safe operators

tautschnig:
Laurel division and modulo operations (/, %, /t, %t) now fail verification when the divisor may be zero.

This is achieved by translating Laurel's division/modulo to Core's safe operator variants (Int.SafeDiv, Int.SafeMod,
Int.SafeDivT, Int.SafeModT), which carry a built-in y ≠ 0 precondition. The existing PrecondElim transform then
automatically generates the necessary verification conditions at each use site — no Laurel-level instrumentation is
required.

*Description of changes:*

New Core operators — Int.SafeDivT and Int.SafeModT added to the Lambda IntBoolFactory and Core Factory, mirroring the
existing Int.SafeDiv/Int.SafeMod. SMT encoding and AST-to-CST printing updated accordingly.

Laurel-to-Core translator — Div/Mod/DivT/ModT now emit the safe operator variants instead of the unchecked ones.

End-to-end tests — New DivisionByZeroCheckTest.lean verifies:
- Safe division with a known non-zero divisor passes ✅
- Division by an unconstrained parameter fails ❌
- A pure function with an explicit requires y != 0 precondition passes, but callers that don't satisfy it fail ❌

### DIFF
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -131,8 +131,8 @@ fn div_expr (tp : Type, a : tp, b : tp) : tp => @[prec(30), leftassoc] a " div "
 fn mod_expr (tp : Type, a : tp, b : tp) : tp => @[prec(30), leftassoc] a " mod " b;
 fn safediv_expr (tp : Type, a : tp, b : tp) : tp => @[prec(30), leftassoc] a " / " b;
 fn safemod_expr (tp : Type, a : tp, b : tp) : tp => @[prec(30), leftassoc] a " % " b;
-fn divt_expr (tp : Type, a : tp, b : tp) : tp => @[prec(30), leftassoc] a " divt " b;
-fn modt_expr (tp : Type, a : tp, b : tp) : tp => @[prec(30), leftassoc] a " modt " b;
+fn divt_expr (tp : Type, a : tp, b : tp) : tp => "Int.DivT(" a ", " b ")";
+fn modt_expr (tp : Type, a : tp, b : tp) : tp => "Int.ModT(" a ", " b ")";
 fn safedivt_expr (tp : Type, a : tp, b : tp) : tp => @[prec(30), leftassoc] a " /t " b;
 fn safemodt_expr (tp : Type, a : tp, b : tp) : tp => @[prec(30), leftassoc] a " %t " b;
 


StrataTest/Languages/Laurel/DivisionByZeroCheckTest.lean
```diff
@@ -0,0 +1,54 @@
...
+}
+
+procedure callPureDivUnsafe(x: int) {
+  var z: int := pureDiv(10, x);
+//              ^^^^^^^^^^^^^^ error: assertion does not hold
+}
+"
+
+#guard_msgs(drop info, error) in
+#eval testInputWithOffset "DivByZeroE2E" e2eProgram 22 processLaurelFile
```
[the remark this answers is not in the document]
tautschnig: The offset of 22 is correct: def e2eProgram is on line 21 (0-indexed), and the raw string content starts on line 22. The caret positions in the test were calibrated to match the actual diagnostics.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -528,6 +528,8 @@ def handleBinaryOps {M} [Inhabited M] (name : String)
   | "Int.SafeDiv" => pure (.safediv_expr default ty arg1 arg2)
   | "Int.Mod" => pure (.mod_expr default ty arg1 arg2)
   | "Int.SafeMod" => pure (.safemod_expr default ty arg1 arg2)
+  | "Int.DivT" | "Int.SafeDivT" => pure (.safediv_expr default ty arg1 arg2)
+  | "Int.ModT" | "Int.SafeModT" => pure (.safemod_expr default ty arg1 arg2)
```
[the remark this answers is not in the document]
tautschnig: Addressed in https://github.com/strata-org/Strata/pull/510/commits/7cdbbd5be944eeeed963a2f401bc2682b3c1c94c.

keyboardDrummer (approved):
Sweet! Thanks

joscoh (approved):

StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
```diff
@@ -11,4 +11,5 @@ assert(102): ✅ pass (at line 7, col 4)
 assert(226): ✅ pass (at line 12, col 4)
 assert(345): ✅ pass (at line 16, col 4)
 assert(458): ✅ pass (at line 20, col 4)
+init_calls_Int.SafeDiv_0: ✅ pass
```
andrewmwells-amazon: Can we get line+col info here?

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -528,6 +528,10 @@ def handleBinaryOps {M} [Inhabited M] (name : String)
   | "Int.SafeDiv" => pure (.safediv_expr default ty arg1 arg2)
   | "Int.Mod" => pure (.mod_expr default ty arg1 arg2)
   | "Int.SafeMod" => pure (.safemod_expr default ty arg1 arg2)
+  | "Int.DivT" => pure (.safedivt_expr default ty arg1 arg2)
+  | "Int.SafeDivT" => pure (.safedivt_expr default ty arg1 arg2)
+  | "Int.ModT" => pure (.safemodt_expr default ty arg1 arg2)
```
joehendrix: This doesn't look right `"Int.ModT"` is mapping to `.safemodt_expr`.  Same with Int.DivT above.

Either Int.ModT should be dropped or `modt_expr` should be declared.

StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
tautschnig: Implemented in https://github.com/strata-org/Strata/pull/510/commits/f37c246b5712dd8cd89fe187e25cf9c7bb738e1e. Commit https://github.com/strata-org/Strata/pull/510/commits/3942b5091ec7d811d9c117e39203a38886d788a4 goes one step further and fixes the missing source locations for all other statement types.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
tautschnig: Implemented the additional nodes in https://github.com/strata-org/Strata/pull/510/commits/256952b8db36e7c74e95c27178d0cb5bfa0a8cc8.

PR #511, opened 2026-03-03
feat(laurel): Add direct operational semantics for Laurel IR

olivier-aws:
## Summary

This PR implements Option A from the design document `design-formal-semantics-for-laurel-ir.md`: a standalone big-step operational semantics for all ~35 Laurel StmtExpr constructors, independent of Core semantics.

### New Files

1. **Strata/Languages/Laurel/LaurelSemantics.lean** (~800 LOC)
   - Core semantic definitions including `LaurelValue`, `LaurelStore`, `LaurelHeap`, `Outcome` types
   - Mutually inductive `EvalLaurelStmt`/`EvalLaurelBlock` relations
   - Coverage: literals, variables, primitive operations, control flow (if/while/block/exit/return), assignments, verification constructs (assert/assume), static calls, OO features (new/field select/instance call/reference equals/is type/as type), and specification constructs

2. **Strata/Languages/Laurel/LaurelSemanticsProps.lean** (~500 LOC)
   - Store monotonicity theorems
   - Store operation determinism
   - `catchExit` and `evalPrimOp` properties
   - Block value semantics
   - Full evaluation determinism theorem (stated, admitted pending mutual induction proof)

3. **StrataTest/Languages/Laurel/LaurelSemanticsTest.lean** (~300 LOC)
   - 30+ concrete evaluation tests covering all major constructs
   - Tests for literals, identifiers, primitive ops, blocks, if-then-else, exit/return propagation, local variables, assert/assume, nested control flow

4. **docs/designs/design-formal-semantics-for-laurel-ir.md**
   - Design document describing three options (A: direct, B: shallow, C: hybrid)
   - This implementation is Option A, serving as a reference semantics for future translation correctness proofs

## Key Design Decisions

- **Outcome type** explicitly models non-local control flow (exit/return)
- **EvalArgs** uses the expression evaluator δ (non-mutual) for call argument evaluation, avoiding Lean 4 mutual inductive limitations
- **Heap model** is separate from store for OO features
- **Specification constructs** (forall, exists, old, fresh, assigned, contract_of) are delegated to the expression evaluator δ
- **Static/instance calls** restore the caller store after body evaluation

## Bug Fixes (second commit)

- Fixed `EvalLaurelBlock` non-determinism by adding `rest ≠ []` side condition to `cons_normal`
- Added void-returning procedure rules (`static_call_return_void`, `instance_call_return`, `instance_call_return_void`)
- Fixed `bindParams` to enforce lexical scoping (start from empty store)
- Added documentation for purity requirements and known limitations

## Known Limitations

- Determinism theorem is admitted (requires careful mutual induction)
- While loop semantics only captures terminating executions (inductive)
- Call argument evaluation uses δ rather than full `EvalLaurelStmt`
- Translation correctness proof is out of scope (separate task)

## Testing

All tests pass:
```bash
lake build
```

## Related

Design document: `docs/designs/design-formal-semantics-for-laurel-ir.md`

### DIFF
--- a/Strata/Languages/Laurel/LaurelSemantics.lean
+++ b/Strata/Languages/Laurel/LaurelSemantics.lean
@@ -1,0 +1,399 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Laurel.Laurel
+
+/-!
+# Direct Operational Semantics for Laurel IR
+
+This file defines a standalone big-step operational semantics for Laurel's
+`StmtExpr` AST, independent of Core semantics (Option A from the design
+document `docs/designs/design-formal-semantics-for-laurel-ir.md`).
+
+## Design
+
+- **LaurelValue**: runtime values (int, bool, string, void, ref)
+- **LaurelStore**: variable store (`Identifier → Option LaurelValue`)
+- **LaurelHeap**: object heap (`Nat → Option (Identifier × (Identifier → Option LaurelValue))`)
+- **Outcome**: non-local control flow (normal, exit, return)
+- **EvalLaurelStmt / EvalLaurelBlock**: mutually inductive big-step relations
+
+The judgment form is: `δ, π, heap, σ, stmt ⊢ heap', σ', outcome`
+-/
+
+namespace Strata.Laurel
+
+/-! ## Values -/
+
+inductive LaurelValue where
+  | vInt    : Int → LaurelValue
+  | vBool   : Bool → LaurelValue
+  | vString : String → LaurelValue
+  | vVoid   : LaurelValue
+  | vRef    : Nat → LaurelValue
+  deriving Repr, BEq, Inhabited, DecidableEq
+
+/-! ## Store and Heap -/
+
+abbrev LaurelStore := Identifier → Option LaurelValue
+abbrev LaurelHeap := Nat → Option (Identifier × (Identifier → Option LaurelValue))
+abbrev LaurelEval := LaurelStore → StmtExpr → Option LaurelValue
+abbrev ProcEnv := Identifier → Option Procedure
+
+/-! ## Outcomes -/
+
+inductive Outcome where
+  | normal : LaurelValue → Outcome
+  | exit   : Identifier → Outcome
+  | ret    : Option LaurelValue → Outcome
+  deriving Repr, BEq, Inhabited, DecidableEq
+
+/-! ## Store Operations -/
+
+inductive UpdateStore : LaurelStore → Identifier → LaurelValue → LaurelStore → Prop where
+  | update :
+    σ x = .some v' →
+    σ' x = .some v →
+    (∀ y, x ≠ y → σ' y = σ y) →
+    UpdateStore σ x v σ'
+
+inductive InitStore : LaurelStore → Identifier → LaurelValue → LaurelStore → Prop where
+  | init :
+    σ x = none →
+    σ' x = .some v →
+    (∀ y, x ≠ y → σ' y = σ y) →
+    InitStore σ x v σ'
+
+/-! ## Heap Operations -/
+
+inductive AllocHeap : LaurelHeap → Identifier → Nat → LaurelHeap → Prop where
+  | alloc :
+    h addr = none →
+    h' addr = .some (typeName, fun _ => none) →
+    (∀ a, addr ≠ a → h' a = h a) →
+    AllocHeap h typeName addr h'
+
+def heapFieldRead (h : LaurelHeap) (addr : Nat) (field : Identifier) : Option LaurelValue :=
+  match h addr with
+  | some (_, fields) => fields field
+  | none => none
+
+inductive HeapFieldWrite : LaurelHeap → Nat → Identifier → LaurelValue → LaurelHeap → Prop where
+  | write :
+    h addr = .some (tag, fields) →
+    h' addr = .some (tag, fun f => if f == field then some v else fields f) →
+    (∀ a, addr ≠ a → h' a = h a) →
+    HeapFieldWrite h addr field v h'
+
+/-! ## Helpers -/
+
+def catchExit : Option Identifier → Outcome → Outcome
+  | some l, .exit l' => if l == l' then .normal .vVoid else .exit l'
+  | _, o => o
+
+def evalPrimOp (op : Operation) (args : List LaurelValue) : Option LaurelValue :=
+  match op, args with
+  | .And,     [.vBool a, .vBool b] => some (.vBool (a && b))
+  | .Or,      [.vBool a, .vBool b] => some (.vBool (a || b))
+  | .Not,     [.vBool a]           => some (.vBool (!a))
+  | .Implies, [.vBool a, .vBool b] => some (.vBool (!a || b))
+  | .Add, [.vInt a, .vInt b] => some (.vInt (a + b))
+  | .Sub, [.vInt a, .vInt b] => some (.vInt (a - b))
+  | .Mul, [.vInt a, .vInt b] => some (.vInt (a * b))
+  | .Neg, [.vInt a]          => some (.vInt (-a))
+  | .Div, [.vInt a, .vInt b] => if b != 0 then some (.vInt (a / b)) else none
+  | .Mod, [.vInt a, .vInt b] => if b != 0 then some (.vInt (a % b)) else none
+  | .Eq,  [.vInt a, .vInt b] => some (.vBool (a == b))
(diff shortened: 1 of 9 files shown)


Strata/Languages/Laurel/LaurelSemanticsProps.lean
```diff
@@ -0,0 +1,131 @@
...
+EvalLaurelBlock simultaneously. The proof is admitted here; the store
+operation determinism lemmas above are the key building blocks.
+-/
+
+mutual
+theorem EvalLaurelStmt_deterministic :
+    EvalLaurelStmt δ π h σ s h₁ σ₁ o₁ →
+    EvalLaurelStmt δ π h σ s h₂ σ₂ o₂ →
+    h₁ = h₂ ∧ σ₁ = σ₂ ∧ o₁ = o₂ := by
+  sorry
```
tautschnig: Can we actually do the proof?

Strata/Languages/Laurel/LaurelSemanticsProps.lean
```diff
@@ -0,0 +1,131 @@
...
+
+theorem EvalLaurelBlock_deterministic :
+    EvalLaurelBlock δ π h σ ss h₁ σ₁ o₁ →
+    EvalLaurelBlock δ π h σ ss h₂ σ₂ o₂ →
+    h₁ = h₂ ∧ σ₁ = σ₂ ∧ o₁ = o₂ := by
+  sorry
```
tautschnig: Can we actually do the proof?

Strata/Languages/Laurel/LaurelSemanticsProps.lean
olivier-aws: Yes, updated the PR (thanks Kiro for the proof)

Strata/Languages/Laurel/LaurelSemanticsProps.lean
olivier-aws: Yes, updated the PR (thanks Kiro for the proof)

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,493 @@
...
+  | vBool   : Bool → LaurelValue
+  | vString : String → LaurelValue
+  | vVoid   : LaurelValue
+  | vRef    : Nat → LaurelValue
+  deriving Repr, BEq, Inhabited, DecidableEq
+
+/-! ## Store and Heap -/
+
+abbrev LaurelStore := Identifier → Option LaurelValue
+abbrev LaurelHeap := Nat → Option (Identifier × (Identifier → Option LaurelValue))
```
aqjune-aws, line 78: What is the first `Identifier` in the range of this type?

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,493 @@
...
+
+  -- Verification Constructs
+  -- Note: assert_true and assume_true require the condition to be pure
+  -- (no side effects on heap or store). Conditions with side effects have
+  -- no derivation. This is intentional — assert/assume conditions should
+  -- be specification-level expressions, not effectful computations.
+
+  | assert_true :
+    EvalLaurelStmt δ π h σ c h σ (.normal (.vBool true)) →
+    EvalLaurelStmt δ π h σ ⟨.Assert c, md⟩ h σ (.normal .vVoid)
```
aqjune-aws, line 310: We are thinking of adding a top-level soundness specification of assert to Strata which looks like this:
```
forall (s0 s:ProgramState),
  init_state(s0) /\ s0 ->* s /\ s.programCounter.instruction = "assert(P)"
  ==> P(s) holds
```
which decouples the operational semantics of `assert(P)` with what actually `assert(P)` being valid means.
For this reason, https://github.com/strata-org/Strata/pull/509 is amending `assert` as a skip instruction. Unless a program state explicitly has "isError?" flag and `assert(P)` mark as true when `P` doesn't hold, I think operational semantics of `assert` shouldn't matter, and simpler semantics of `assert` is always better.
Could Laurel also use the same style with Core, for easier proof of transformation in the future?

Strata/Languages/Laurel/LiftImperativeExpressionsCorrectness.lean
```diff
@@ -0,0 +1,672 @@
...
+`_v'` — these are independent. This means `TransformOK` does not require
+intermediate normal values to agree, only that the final (heap, store,
+outcome) match. This is sufficient for `TransformOK_eval` (since
+`EvalLaurelBlock_append` only needs *some* normal value), but it weakens
+`TransformOK` as a specification of semantic preservation. A stronger
+version could require `_v = _v'` in `cons_normal`. -/
+inductive TransformOK :
+    LaurelEval → ProcEnv → LaurelHeap → LaurelStore →
+    List StmtExprMd → List StmtExprMd →
+    LaurelHeap → LaurelStore → Outcome → Prop where
```
aqjune-aws, line 614: It seems TransformOK doesn't directly use the executable `transformExpr` function in LiftImperativeExpressions - or am I missing it?

aqjune-aws (commented):
Do you also have a plan to add small-step semantics for Laurel? This is helpful for describing the top-level soundness statement of `assert` (written as a comment below), where `assert` can be embedded in the middle of an inner block, e.g.,
```
int x = 1;
...
if (...) {
  assert(x == 1);
}
```

To describe the validity of `assert` in the `if` block, we will need steps from the beginning of a program to the middle of `if`, but big-step style would not represent the program state at this position. In the small-step style, this is possible because what small-step does is really mimicking the interpreter (and hitting the 'stop' on a debugger :) )

Strata/Languages/Laurel/LiftImperativeExpressionsCorrectness.lean
```diff
@@ -0,0 +1,672 @@
+/-
```
keyboardDrummer, line 1: It's not entirely clear to me what the meaning of this correctness proof is. The left translation pass is partial: it panics for certain Laurel programs. How is that reflected in this proof?

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,493 @@
...
+/-! ## Store and Heap -/
+
+abbrev LaurelStore := Identifier → Option LaurelValue
+abbrev LaurelHeap := Nat → Option (Identifier × (Identifier → Option LaurelValue))
+abbrev LaurelEval := LaurelStore → StmtExpr → Option LaurelValue
+abbrev ProcEnv := Identifier → Option Procedure
+
+/-! ## Outcomes -/
+
+inductive Outcome where
```
keyboardDrummer, line 84: What about a program with resolution errors? Shouldn't reporting a resolution error be a possible outcome?

Strata/Languages/Laurel/LaurelSemantics.lean
keyboardDrummer, line 78: Please use `structure` instead of `×` when it's not entirely obvious

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,493 @@
...
+
+mutual
+inductive EvalLaurelStmt :
+    LaurelEval → ProcEnv → LaurelHeap → LaurelStore →
+    StmtExprMd → LaurelHeap → LaurelStore → Outcome → Prop where
+
+  -- Literals
+
+  | literal_int :
+    EvalLaurelStmt δ π h σ ⟨.LiteralInt i, md⟩ h σ (.normal (.vInt i))
```
keyboardDrummer, line 204: Please don't use single letter identifiers unless their meaning is obvious from the line that they're in (which is rarely the case)

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,493 @@
...
+
+/-- Non-mutual argument evaluation using the expression evaluator δ. -/
+inductive EvalArgs : LaurelEval → LaurelStore → List StmtExprMd → List LaurelValue → Prop where
+  | nil  : EvalArgs δ σ [] []
+  | cons : δ σ e.val = some v → EvalArgs δ σ es vs → EvalArgs δ σ (e :: es) (v :: vs)
+
+/-! ## Main Semantic Relations -/
+
+mutual
+inductive EvalLaurelStmt :
```
keyboardDrummer, line 197: Could we compare proofs written using these semantics versus those written using the more interpreter style semantics? I'm really excited about extracting the resolution semantics from the latter.

olivier-aws:
New PR with better design in #631

PR #512, opened 2026-03-03
Improve pySpecs robustness: graceful failures, regex support, and reusable API

joehendrix:
## Summary

Improves the robustness of the `pySpecs` command so it reports failures
gracefully — including regex usage — instead of aborting, and extracts
the core logic into a reusable API with CLI flags for skipping
definitions, controlling verbosity, and selecting log events.

## Details

- **Graceful failure handling.** When a Python import cannot be resolved,
  the translator now registers the imported names as opaque extern types
  and continues, rather than aborting. This means downstream references
  to those names still produce valid (if incomplete) output. Overload
  stubs with unexpected bodies produce a warning instead of failing.
- **Regex support.** `from re import compile` is recognized via the
  built-in prelude, so Python files that use regex no longer trigger
  unknown-identifier errors. A test confirms this import resolves
  without warnings.
- **Reusable API.** The translation logic that was inline in the CLI
  callback is extracted into a self-contained entry point in SimpleAPI.
  It validates inputs, parses skip-name strings, translates, writes
  output, and reports warnings — all returning structured errors instead
  of calling `exitFailure`. The CLI callback now delegates to this.
- **CLI flags.** `--skip <name>` omits named top-level definitions from
  output (overloaded variants are always kept). `--quiet` suppresses
  logging and warnings. `--log <event>` enables specific event types.
- **Warning pipeline.** Warnings are now collected during translation
  and returned to the caller as data, rather than being printed to
  stderr internally. A three-level verbosity setting (none / summary /
  detail) controls how they are reported.
- **Composable assertion expressions.** The expression tree for
  preconditions and postconditions is refactored from fused nodes
  (e.g., "length >= N" as one node) into orthogonal building blocks
  (length, integer literal, greater-or-equal, etc.). This makes
  serialization a direct structural recursion with no special cases.
- **Readable DDM output.** Assertions print as `ensure(pred, msg)`.
  Comparisons use infix `>=`/`<=`. Class member blocks are indented
  on separate lines. These changes make the generated specs easier to
  read and diff.
- **Event logging cleanup.** Logging no longer uses a global mutable
  reference; the set of enabled events is threaded through the
  translation context. Import messages now report after completion
  (past tense) and include warning counts.

### DIFF
--- a/Strata/Languages/Python/Specs.lean
+++ b/Strata/Languages/Python/Specs.lean
@@ -1322,17 +1322,6 @@ end
 /-- Maps file paths to their FileMap for error location reporting. -/
 public abbrev FileMaps := Std.HashMap System.FilePath Lean.FileMap
 
-namespace FileMaps
-
-def ppSourceRange (fmm : FileMaps)  (path : System.FilePath) (loc : SourceRange) : String :=
-  match fmm[path]? with
-  | none =>
-    panic! "Invalid path {file}"
-  | some fm =>
-    loc.format path fm
-
-end FileMaps
-
 /-- Translates Python AST statements to PySpec signatures with dependency resolution. -/
 def translateModule
     (dialectFile searchPath strataDir pythonFile : System.FilePath)
@@ -1406,13 +1395,18 @@ public def translateFile
         (pythonFile := pythonFile)
         (.ofString contents)
         body
+  let ppErr (e : SpecError) : EIO String String :=
+        match fmm[e.file]? with
+        | none =>
+          throw s!"No location information for {e.file}"
+        | some fm =>
+          pure s!"{e.loc.format e.file fm}: {e.message}"
   if errors.size > 0 then
     let msg := "Translation errors:\n"
-    let msg := errors.foldl (init := msg) fun msg e =>
-      s!"{msg}{fmm.ppSourceRange pythonFile e.loc}: {e.message}\n"
+    let msg ←
+      errors.foldlM (init := msg) fun msg e =>
+        return s!"{msg}{← ppErr e}\n"
     throw msg
-  let warningMsgs := warnings.map fun w =>
-    s!"{fmm.ppSourceRange w.file w.loc}: {w.message}"
-  pure (sigs, warningMsgs)
+  pure (sigs, ← warnings.mapM ppErr)
 
 end Strata.Python.Specs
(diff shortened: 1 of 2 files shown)


Strata/SimpleAPI.lean
```diff
@@ -184,3 +185,87 @@ Do deductive verification of a Core program, including any external solver
...
+  match ← pythonFile.metadata |>.toBaseIO with
+  | .ok md =>
+    if md.type != .file then
+      throw s!"Expected {pythonFile} to be a regular file"
+  | .error e => throw s!"Cannot access {pythonFile}: {e}"
+
+  -- Parse skip names into PythonIdents
+  let some fileStem := pythonFile.fileStem
+    | throw s!"No file stem for {pythonFile}"
+  let stem := fileStem
```
shigoel: Nit: looks like `fileStem` can be used everywhere and `stem` isn't needed

Strata/Languages/Python/Specs.lean
```diff
@@ -1275,29 +1322,37 @@ end
 /-- Maps file paths to their FileMap for error location reporting. -/
 public abbrev FileMaps := Std.HashMap System.FilePath Lean.FileMap
 
-def FileMaps.ppSourceRange (fmm : Strata.Python.Specs.FileMaps) (path : System.FilePath) (loc : SourceRange) : String :=
+namespace FileMaps
+
+def ppSourceRange (fmm : FileMaps)  (path : System.FilePath) (loc : SourceRange) : String :=
   match fmm[path]? with
   | none =>
     panic! "Invalid path {file}"
```
shigoel: Out of scope for this PR, but noting a panic here that'd be a good candidate to convert to an exception.

joehendrix: I just migrated it to reduce future friction.

Strata/SimpleAPI.lean
joehendrix: Done

Strata/Languages/Python/Specs.lean
```diff
@@ -25,17 +25,11 @@ public abbrev EventType := String
 def importEvent : EventType := "import"
 
 /--
-Set of event types to log to stderr. Test code can modify this to
-enable/disable logging.
--/
-initialize stdoutEventsRef : IO.Ref (Std.HashSet EventType) ← IO.mkRef {}
```
aqjune-aws, line 31: Wasn't aware of this feature; declaring a global variable in Lean was possible!

PR #515, opened 2026-03-04
Add automatic overload resolution and argument validation for PySpec dispatch

joehendrix:
Resolves overloaded service dispatch (e.g., `servicelib.connect("storage")`) by walking the Python AST to identify which PySpec modules are needed, replacing manual `--pyspec` flags. Restructures the PySpec-to-Core pipeline into composable steps and adds argument validation with user-friendly error messages.

## Changes

- **Overload resolution** (`Specs/IdentifyOverloads.lean`) — New AST walker collects the exact set of PySpec modules used by matching `Call` nodes against an `OverloadTable`. A new `pyResolveOverloads` CLI command exposes this for tooling. `pyAnalyzeLaurel` now auto-discovers needed PySpec files, eliminating manual `--pyspec` flags.

- **Single Laurel.translate pass** (`PySpecPipeline.lean`, `StrataMain.lean`) — PySpec and user Laurel declarations are merged into one program before translation to Core, eliminating `TypeTag` collisions from independent translation.

- **PreludeInfo abstraction** (`PythonToLaurel.lean`) — Decouples `pythonToLaurel'` from `Core.Program` so prelude info can come from either Core or Laurel programs. External stubs are generated separately, allowing `pythonToLaurel'` to produce a clean user program.

- **Module-prefix PySpec identifiers** (`Specs/ToLaurel.lean`) — Type/procedure names are prefixed by PySpec filename (e.g., `Tag` → `S3_Tag`) to prevent collisions. Name collisions across PySpec files are now hard errors.

- **Argument validation** (`PythonToLaurel.lean`) — `combinePositionalAndKeywordArgs` rejects unknown kwargs and reports missing required args by original Python method name (e.g., `'put_item' called with missing required arguments: [Key, Data]`). Unknown methods and invalid service names also produce specific errors.

- **User class method recognition** (`PythonToLaurel.lean`) — `userFunctions` now includes class methods in `ClassName_methodName` format, so calls to user-defined class methods are recognized by `hasModel`.

- **Pipeline extraction** (`PySpecPipeline.lean`) — Implementation moved from `SimpleAPI.lean` into a dedicated module with composable steps (`buildPreludeInfo`, `combinePySpecLaurel`, `translateCombinedLaurel`, `replaceStubsWithPrelude`), keeping SimpleAPI as a thin stable surface.

- **Self/kwargs handling** (`PySpecPipeline.lean`, `Specs/ToLaurel.lean`) — PySpec method signatures strip `self` (with validation that it exists) and expand `**kwargs: Unpack[TypedDict]` into individual parameters. Non-TypedDict kwargs report errors via `expandKwargsArgsM`.

- **SpecDefault in DDM** (`Specs/DDM.lean`, `Specs/Decls.lean`) — `Arg.hasDefault : Bool` replaced with `Arg.default : Option SpecDefault`

- **Field access Any coercions** (`PythonToLaurel.lean`) — After #568 resolved field types to concrete types, heap parameterization produces typed Box destructors (`Box..intVal!` returns `int`) but Python operators expect `Any`. Fixed by wrapping `FieldSelect` expressions in `from_int`/`from_bool`/etc. at the PythonToLaurel layer, keeping HeapParameterization language-agnostic. Replaced unused `classAttributeType` with nested `classFieldHighType` HashMap. `lookupFieldHighType` throws explicit errors for `self.field` on known classes; `tryLookupFieldHighType` falls back gracefully for `obj.field` on unknown types.

- **Infrastructure** (`ReadPython.lean`, `Specs.lean`) — Extracted `runPyToStrata`/`readParseStrataFile`, added `PythonToStrataOptions` with perf logging. `List` → `HashMap`/`HashSet` for `preludeProcedures`, `preludeTypes`, `compositeTypeNames`.

## Test coverage

Two new test files exercise the dispatch pipeline against a mock service fixture (`Storage`, `Messaging`, `servicelib` dispatch file) with 17 Python test scripts in `StrataTest/Languages/Python/Specs/dispatch_test/`:

**`AnalyzeLaurelTest.lean`** — 12 data-driven end-to-end tests running through the full `pyAnalyzeLaurel` pipeline with concurrent execution via `IO.asTask`:
- 4 positive: single service (`test_single_service.py`), multi-service (`test_multi_service.py`), type annotation fallback (`test_annotation_fallback.py`), required+optional args (`test_required_with_optional.py`)
- 8 negative with exact error message string matching:
  - `test_invalid_service.py` — `'connect' called with unknown service name "invalid"; known services: [storage, messaging]`
  - `test_invalid_method.py` — `Unknown method 'nonexistent_method'`
  - `test_invalid_args.py` — `'put_item' called with unknown keyword arguments: [Wrong]`
  - `test_missing_required.py` — `'put_item' called with missing required arguments: [Key, Data]`
  - `test_extra_kwarg.py` — `'get_item' called with unknown keyword arguments: [Bogus]`
  - `test_no_args.py` — `'put_item' called with missing required arguments: [Bucket, Key, Data]`
  - `test_optional_missing_required.py` — `'list_items' called with missing required arguments: [Bucket]`
  - `test_positional_missing.py` — `'delete_item' called with missing required arguments: [Key]`

**`IdentifyOverloadsTest.lean`** — 7 unit tests calling `resolveOverloads` directly, asserting it detects modules.

- `test_single_service.py` → exactly `{Storage}`
- `test_multi_service.py` → exactly `{Storage, Messaging}`
- `test_class_dispatch.py` → dispatch inside class methods → `{Storage}`
- `test_dispatch_in_conditional.py` → both if/else branches → `{Storage, Messaging}`
- `test_dispatch_in_try.py` → dispatch in try block → `{Storage}`
- `test_no_dispatch.py` → no dispatch calls → `{}`
- `test_dispatch_in_loop.py` → variable argument (non-literal) → `{}` (correctly not resolved)

**Existing test updates:**
- `SpecsTest.lean` — Updated DDM format (`hasDefault: Bool` → `default: Option SpecDefault`)
- `ToLaurelTest.lean` — Test fixtures updated with `self` parameter for method validation
- `test_class_field_use.expected` — Updated with passing verification results now that field access type mismatch is fixed

### DIFF
--- a/Strata/Languages/Python/PySpecPipeline.lean
+++ b/Strata/Languages/Python/PySpecPipeline.lean
@@ -249,21 +249,11 @@ public def combinePySpecLaurel (info : Python.PreludeInfo)
     constants := pySpec.constants ++ user.constants
   }
 
-/-- `coreFromLaurel` contains External stub declarations for every
-    prelude name (e.g. `print`, `from_string`, `Any`).  These stubs
-    were added so the Laurel resolver could classify names as
-    functions vs procedures, but they carry no type information or
-    bodies.  This function drops those stubs and prepends the full
-    `pythonRuntimeCorePart` (with datatype definitions, procedure
-    bodies, etc.) that the Core type-checker and verifier require. -/
-private def replaceStubsWithPrelude (coreFromLaurel : Core.Program) : Core.Program :=
-  let preludeDecls := Python.pythonRuntimeCorePart.decls
-  let preludeNames : Std.HashSet String :=
-    preludeDecls.foldl (init := {}) fun s d =>
-      d.names.foldl (init := s) (·.insert ·.name)
-  let newDecls := coreFromLaurel.decls.filter fun d =>
-    !(d.names.any (·.name ∈ preludeNames))
-  { decls := preludeDecls ++ newDecls }
+/-- Prepend the full Python runtime Core prelude (datatype definitions,
+    procedure bodies, etc.) to the Core program produced by Laurel
+    translation. -/
+private def prependPrelude (coreFromLaurel : Core.Program) : Core.Program :=
+  { decls := Python.pythonRuntimeCorePart.decls ++ coreFromLaurel.decls }
 
 /-- Translate a combined Laurel program to Core and prepend the full
     runtime prelude.  Resolution errors are suppressed because PySpec
@@ -276,7 +266,7 @@ public def translateCombinedLaurel (combined : Laurel.Program)
     : Except (Array DiagnosticModel) (Core.Program × Array DiagnosticModel) :=
   match Laurel.translate { emitResolutionErrors := false } combined with
   | .error e => .error e
-  | .ok (core, diags) => .ok (replaceStubsWithPrelude core, diags)
+  | .ok (core, diags) => .ok (prependPrelude core, diags)
 
 /-- Errors from the pyAnalyzeLaurel pipeline, distinguishing user code
     errors (detected bugs in Python source) from internal tool errors. -/


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -261,16 +268,26 @@ def resolveDispatch (ctx : TranslationContext)
...
     match args[0] with
     | .Constant _ (.ConString _ s) _ =>
-      return (fnOverloads.get? s.val).map (·.name)
+      match fnOverloads.get? s.val with
+      | some ident =>
+        let className := if ident.pythonModule.isEmpty then ident.name
+          else ident.pythonModule ++ "_" ++ ident.name
+        return some className
+      | none =>
+        let knownServices := fnOverloads.keys.take 2
```
shigoel: `fnOverloads.keys` iterates a `Std.HashMap`, whose key order is not guaranteed. With the two-entry test fixture (`storage`, `messaging`), this error message could produce `[messaging, storage]` instead of `[storage, messaging]`, making the exact-string assertion in `AnalyzeLaurelTest.lean` flaky.

Suggest sorting before taking:
```lean
let knownServices := fnOverloads.keys.toList.mergeSort.take 2
```

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -255,10 +266,47 @@ def argToParameter (arg : Arg) : ToLaurelM Parameter := do
...
+      pure <| fields.mapIdx fun i name =>
+        { name := name
+          type := fieldTypes.getD i default
+          default := if fieldRequired.getD i true then none else some .none }
+    | _ => do
+      reportError default s!"**{name} has non-TypedDict type; kwargs not expanded"
+      pure #[]
+
+/-- Pure version of `expandKwargsArgsM` for use outside `ToLaurelM`. -/
+public def expandKwargsArgs (kwargs : Option (String × SpecType)) : Array Arg :=
```
shigoel: The pure `expandKwargsArgs` silently returns `#[]` when kwargs is present but not a TypedDict, while `expandKwargsArgsM` reports an error via `reportError`. Since `expandKwargsArgs` is used to build `functionSignatures` (which drives argument validation), a non-TypedDict `**kwargs` method will have an empty arg list there, causing argument validation to incorrectly accept or reject calls to that method.

Consider having `funcDeclToFunctionDecl` in `PySpecPipeline.lean` return `Except String` and call `expandKwargsArgsM` (or an equivalent pure-but-fallible version) so the error surfaces at `buildPySpecLaurel` time rather than silently producing wrong signatures.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1333,51 +1377,153 @@ def getPreludeProcedures (prelude: Core.Program) : List String :=
...
+  functions := getPreludeFunctions prelude
+  procedureNames := getPreludeProcedures prelude
+
+/-- Convert a Laurel `HighType` to the same string name that `getTypeName` would
+    produce from the corresponding Core `LMonoTy` after translation. -/
+def getHighTypeName : Laurel.HighType → String
+  | .TInt => "int"
+  | .TBool => "bool"
+  | .TString => "string"
+  | .TVoid => "bool"
```
shigoel: Void -> Bool?

shigoel: Agentic comment:
In practice this only affects procedures that declare a TVoid output in PreludeInfo.ofLaurelProgram.procedures, which populates TranslationContext.preludeProcedures used by withException. The withException check looks for "Error" as the last output type, so mapping TVoid → "bool" vs "void" might not cause visible failures today — but it's a latent bug if code ever inspects these type names.

Strata/Languages/Python/PythonToLaurel.lean
joehendrix: Good catch.  I just pushed this fix to sort keys.

Strata/Languages/Python/Specs/ToLaurel.lean
joehendrix: Good catch.  I deduplicated and changed callers.

Strata/Languages/Python/PythonToLaurel.lean
joehendrix: Fixed to "void"

StrataTest/Languages/Python/expected_laurel/test_class_field_use.expected
```diff
@@ -1,30 +1 @@
...
-POr_body_calls_Any_to_bool_0: ✅ pass
-ret_type: ✅ pass (in prelude file)
-ret_type: ✅ pass (in prelude file)
-ret_pos: ✅ pass (in prelude file)
-assert_name_is_foo: ✅ pass (in prelude file)
-assert_opt_name_none_or_str: ✅ pass (in prelude file)
-assert_opt_name_none_or_bar: ✅ pass (in prelude file)
-ensures_maybe_except_none: ✅ pass (in prelude file)
-assert_assert(285)_calls_Any_to_bool_0: ✅ pass (at line 14, col 4)
-assert(285): ❓ unknown (at line 14, col 4)
```
aqjune-aws, line 30: It seems the type checking failure is only printed to stderr, making this file empty.
What about slightly tweaking run_py_analyze.sh so that the error message to stderr is compared when stdout is empty? Claude made a patch for me. :) https://gist.github.com/aqjune-aws/a42ae93e0f2f8a2f07d1e05d0555c120

joehendrix: I decided to just go ahead and fix the underlying field type checking issue in f8f735d71a7c92989b231cfeb6ecd6d559da23f7.

joehendrix:
Since this wasn't going to be merged today, I fixed the field dereference issue and added additional logic to distinguish PythonToLaurel errors/limitations from errors in the program under test.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -223,6 +233,32 @@ def intToAny (i: Int) := mkStmtExprMd (.StaticCall "from_int" [mkStmtExprMd (Stm
 def boolToAny (b: Bool) := mkStmtExprMd (.StaticCall "from_bool" [mkStmtExprMd (StmtExpr.LiteralBool b)])
 def AnyNone := mkStmtExprMd (.StaticCall "from_none" [])
 def Any_to_bool (b: StmtExprMd) := mkStmtExprMd (.StaticCall "Any_to_bool" [b])
+
+/-- Wrap a field access expression in the appropriate Any constructor based on HighType.
```
keyboardDrummer, line 246: @joehendrix I think we're both missing some context on how @andrewmw94 and @thanhnguyen-aws intended to use Any.

I would think that a simple Python translation uses Any for all variable types. All fields would be typed `Any` and you wouldn't need to wrap/unwrap anywhere except in the implementations of the operators.

If we're using types more specific than Any for fields, I think that only improves performance if we don't immediately wrap these values in Any but instead replace some of the calls to Any operators to specific operator calls.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -0,0 +1,328 @@
...
+    functions vs procedures, but they carry no type information or
+    bodies.  This function drops those stubs and prepends the full
+    `pythonRuntimeCorePart` (with datatype definitions, procedure
+    bodies, etc.) that the Core type-checker and verifier require. -/
+private def replaceStubsWithPrelude (coreFromLaurel : Core.Program) : Core.Program :=
+  let preludeDecls := Python.pythonRuntimeCorePart.decls
+  let preludeNames : Std.HashSet String :=
+    preludeDecls.foldl (init := {}) fun s d =>
+      d.names.foldl (init := s) (·.insert ·.name)
+  let newDecls := coreFromLaurel.decls.filter fun d =>
```
keyboardDrummer: Is this filtering necessary? Laurel will already filter out any functions/procedures with an `External` body, which the stubs have, when translating to Core.

joehendrix: Thanks.  I seem to recall getting conflicts at one point, but I just tested and simplifying this works.

Strata/Languages/Python/PythonToLaurel.lean
joehendrix: I could imagine that a Core to Core rewriting transform could locally optimize away projection/removal while having all fields declared as Any would requite a more complex analysis.

However, that's just speculation.  This seemed like the more light touch than changing how class fields are represented.

keyboardDrummer: > This seemed like the more light touch than changing how class fields are represented.

My impression is that since https://github.com/strata-org/Strata/pull/489 all fields were represented as Any in Core, so this PR does change how they are represented.

That PR overwrites the Laurel Box datatype so that only `Any` can be represented in a Box:
```
//This is only used to overwrite the Box datatype of Laurel preludeExpand commentComment on line R640Resolved
//WILL BE REMOVEDExpand commentComment on line R641Resolved
datatype Box () {
  BoxInt(intVal: Any)
};
```

While the heap type is `type Heap = Map Ref Map Field Box`

PR #516, opened 2026-03-04
Refactor DialectFileMap to use IO.Ref for LoadedDialects

joehendrix:
## Summary

This PR replaces the pure `DialectFileMap.preloaded` field with a mutable
`IO.Ref LoadedDialects`, eliminating the need to thread `LoadedDialects`
through parameters and return types across every dialect-loading call
site.

The main benefit is that `strata print` and other DDM CLI subcommands should be much easier to work with since they don't require dialects that are already precompiled into Strata.

## Details

- **Mutable ref in `DialectFileMap`.** Added `loaded : IO.Ref
  LoadedDialects` with `new`, `getLoaded`, `modifyLoaded` helpers.
- **Removed `LoadedDialects` threading.** Removed `LoadedDialects` from
  parameters and return tuples throughout the dialect-loading mutual
  block, public API (`loadDialect`, `elabDialect`), `Util/IO`, and
  `SimpleAPI`. Functions now read/write the ref directly.
- **Removed `DialectState.loaded` field.** Added `getLoadedDialects` in
  `DialectM` that reads from the ref, eliminating the sync pattern and
  stale-snapshot risk.
- **Simplified `LoadDialectCallback`.** Reduced from two-argument
  callback returning a pair to a single-argument callback.
- **Refactored text dialect loading.** Extracted header parsing into
  `readDialectTextfileHeader`; made `loadDialectFromPath` private with
  a public `loadDialectFromFile` wrapper.
- **Preload production dialects.** Core, Laurel, Python, PythonSpecs and the SMT family are
  now registered in `buildDialectFileMap` so strata DDM commands do not need them added as arguments.

### DIFF
--- a/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
+++ b/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
@@ -355,11 +355,9 @@ def translateStmt (env : TypeEnv) (outputParams : List Parameter) (stmt : StmtEx
           let ident := ⟨outParam.name, ()⟩
           let coreExpr ← translateExpr env value
           let assignStmt := Core.Statement.set ident coreExpr md
-          let noFallThrough := Core.Statement.assume "return" (.const () (.boolConst false)) .empty
-          return (env, [assignStmt, noFallThrough])
+          return (env, [assignStmt, .exit (some "$body") md])
       | none, _ =>
-          let noFallThrough := Core.Statement.assume "return" (.const () (.boolConst false)) .empty
-          return (env, [noFallThrough])
+          return (env, [.exit (some "$body") md])
       | some _, none =>
           panic! "Return statement with value but procedure has no output parameters"
   | .While cond invariants decreasesExpr body =>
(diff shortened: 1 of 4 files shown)


StrataMain.lean
```diff
@@ -66,7 +66,16 @@ def insertRepeated (pf : ParsedFlags) (name : String) (value : String) : ParsedF
   { pf with repeated := pf.repeated.insert name (arr.push value) }
 
 def buildDialectFileMap (pflags : ParsedFlags) : IO Strata.DialectFileMap := do
-  let mut sp : Strata.DialectFileMap := {}
+  let preloaded := Strata.Elab.LoadedDialects.builtin
```
shigoel, line 69: Looks like the general guidance is to update this list whenever we add a new known/"production" dialect.

Question: if we add dialect X to this list -- and it has dependencies (i.e., imports other dialects) -- do the imported dialects get added automatically too? Asking because I'm mulling splitting up Core's DDM into (at least) two dialects, one of which will be (Lambda) expressions.

Strata/DDM/Elab.lean
```diff
@@ -152,7 +152,6 @@ private def checkDialectName (ld : LoadedDialects) (actual : DialectName) (expec
       .ok ()
   | some expected =>
     if actual = expected then
-      assert! expected ∉ ld.dialects
```
shigoel, line 155: umm, is this not true anymore?

shigoel (approved):
LGTM, left a couple of questions for my own understanding.

Strata/DDM/Elab.lean
joehendrix: It should be true at all the call sites.  The assertion tripped up Claude in a review, and I decided I didn't want this line of code triggering an assertion failure.  It's a private function and the callers should enforce this.

StrataMain.lean
joehendrix: The dialects have to be added with its dependencies added first.  It will trigger a runtime error to add a  dialect without doing so (which should get caught in testing when Strata print is run).

If this becomes a pain point, we could extend the interface to take a dialect's `DialectMap`.  This is a downward closed set that includes dependencies.

The other option I could implement would be to query the Lean Environment in this file in meta-programming and have it automatically add all dialects it knows about.

I decided that at least for now, an explicit list is reasonable and makes it clear what dialects are in scope.

PR #517, opened 2026-03-05
Fix anchor bugs in pythonRegexToCore; add differential test suite

shigoel:
## Summary

- **Differential testing framework** (`DiffTestCore.lean`, `diff_test.py`): pipes `(regex, string, mode)` triples through Strata's SMT backend and compares results against Python's `re` module; classifies outcomes as agree / bug / known_gap / investigate
- **Fixes in `ReToCore.lean`**:
  - Remove `partial`
  - Fix complement translation (intersect with `re.allchar()` to match `[^...]` semantics)
  - Fix anchor-interaction bugs in all four `concat` cases (`true,false` / `false,true` / `false,false`) — anchors were incorrectly forwarded across potentially-empty siblings
  - Fix the same anchor bug in `star | false =>` and `loop(0,m) | false =>` branches
  - Refactor `star` and `loop(0,m)` to unify the `alwaysConsume` and anchor-split paths under a single condition
- **Test corpus** (`diff_test.py`): Cases covering literals, character classes, quantifiers, alternation, all anchor/mode combinations, and unimplemented features; every entry annotated with its Python ground-truth answer; corpus organized into labeled sections
- **CI**: add `Run regex differential tests` step to `build_and_test_lean` job

### DIFF
--- a/Strata/Languages/Core/SMTEncoder.lean
+++ b/Strata/Languages/Core/SMTEncoder.lean
@@ -155,6 +155,11 @@ we leave as `partial` for now.
 -/
 partial def SMT.Context.addType (E: Env) (id: String) (args: List LMonoTy) (ctx: SMT.Context) :
   SMT.Context :=
+  -- Always recurse into concrete args to register any type references
+  let ctx := args.foldl (fun ctx arg =>
+    match arg with
+    | .tcons id1 args1 => SMT.Context.addType E id1 args1 ctx
+    | _ => ctx) ctx
   match E.datatypes.getType id with
   | some d =>
     if ctx.hasDatatype id then ctx else
--- a/StrataTest/Languages/Core/Tests/PolymorphicDatatypeTest.lean
+++ b/StrataTest/Languages/Core/Tests/PolymorphicDatatypeTest.lean
@@ -416,4 +416,119 @@ Result: ✅ pass
 #guard_msgs in
 #eval verify optionHavocPgm (options := .quiet)
 
+---------------------------------------------------------------------
+-- Test 10: Polymorphic datatype instantiated with user-defined datatype
+-- (Issue #525: Inner not declared as a type in SMT)
+---------------------------------------------------------------------
+
+def polyWithUserDatatypePgm : Program :=
+#strata
+program Core;
+
+datatype Option (a : Type) { None(), Some(val: a) };
+
+datatype Inner () {
+  MkInner(flag: bool)
+};
+
+datatype Outer () {
+  MkOuter(
+    name: Option string,
+    inner: Option Inner
+  )
+};
+
+procedure TestPolyUserDatatype() returns ()
+spec { ensures true; }
+{
+  var x : Outer;
+  havoc x;
+  assume Option..isSome(Outer..inner(x));
+  assert [innerIsSome]: Option..isSome(Outer..inner(x));
+};
+#end
+
+/-- info: true -/
+#guard_msgs in
+#eval TransM.run Inhabited.default (translateProgram polyWithUserDatatypePgm) |>.snd |>.isEmpty
+
+/--
+info:
+Obligation: assume_assume_0_calls_Outer..inner_0
+Property: assert
+Result: ✅ pass
+
+Obligation: assert_innerIsSome_calls_Outer..inner_0
+Property: assert
+Result: ✅ pass
+
+Obligation: innerIsSome
+Property: assert
+Result: ✅ pass
+
+Obligation: TestPolyUserDatatype_ensures_0
+Property: assert
+Result: ✅ pass
+-/
+#guard_msgs in
+#eval verify polyWithUserDatatypePgm (options := .quiet)
+
+---------------------------------------------------------------------
+-- Test 11: Non-datatype parameterized by a datatype
+-- (Abstract type Seq's argument Inner must still be registered in SMT)
+---------------------------------------------------------------------
+
+def nonDatatypeWithDatatypeArgPgm : Program :=
+#strata
+program Core;
+
+type Seq (a: Type);
+
+datatype Option (a : Type) { None(), Some(val: a) };
+
+datatype Inner () {
+  MkInner(x: int)
+};
+
+datatype Middle () {
+  MkMiddle(items: (Seq Inner))
+};
+
+datatype Outer () {
+  MkOuter(
+    mid: Option Middle,
+    flag: Option bool
+  )
+};
+
+var v : Outer;
+
+procedure Test() returns ()
+spec { ensures true; }
+{
+  assert [test]: Option..isSome(Outer..flag(v));
+};
+#end
+
+/-- info: true -/
+#guard_msgs in
+#eval TransM.run Inhabited.default (translateProgram nonDatatypeWithDatatypeArgPgm) |>.snd |>.isEmpty
+
+/--
+info:
+Obligation: assert_test_calls_Outer..flag_0
+Property: assert
+Result: ✅ pass
+
+Obligation: test
+Property: assert
+Result: ❌ fail
+
+Obligation: Test_ensures_0
+Property: assert
+Result: ✅ pass
+-/
+#guard_msgs in
+#eval verify nonDatatypeWithDatatypeArgPgm (options := .quiet)
+
 end Strata.PolymorphicDatatypeTest


Strata/Languages/Python/Regex/ReToCore.lean
```diff
@@ -75,98 +121,192 @@ Unmatchable regex pattern.
...
+        let r2b := toCore r2 false atEnd
+        mkApp () (.op () reConcatFunc.name none) [r1b, r2b]
     | false, true =>
-      let r1b := toCore r1 atStart false
-      let r2b := toCore r2 true atEnd
-      mkApp () (.op () reConcatFunc.name none) [r1b, r2b]
+      -- `r2` always consumes; `r1` may be empty. `r2` might be first.
+      -- Case 1: (`r1=""`, `r2` starts at original position, `atStart` propagates) and
+      -- Case 2: (`r1` non-empty, `r2` starts after `r1`, `atStart` must not propagate).
+      if atStart && r2.containsAnchorStart && r1.hasNonAnchorContent then
```
MikaelMayer: The `false, true` case (lines 247-255) and `false, false` case (lines 264-272) have nearly identical anchor-splitting logic. Both check `atStart && r2.containsAnchorStart && r1.hasNonAnchorContent` and produce the same union structure:

```lean
let r1b_eps := mkApp () (.op () reInterFunc.name none) [r1b, Core.emptyRegex]
let r2b_start := toCore r2 atStart atEnd
let r2b_mid := toCore r2 false atEnd
mkApp () (.op () reUnionFunc.name none)
  [mkApp () (.op () reConcatFunc.name none) [r1b_eps, r2b_start],
   mkApp () (.op () reConcatFunc.name none) [r1b, r2b_mid]]
```

Consider extracting this into a helper function that takes the pre-computed `r1b` and returns the union expression. This would reduce duplication and make the anchor-splitting pattern more explicit.

StrataTest/Languages/Python/Regex/diff_test.py
```diff
@@ -0,0 +1,728 @@
...
+    ("(a)(b$)",  "ab",  "search"),     # match — b$ forces end
+    ("(a)(b$)",  "abx", "search"),     # noMatch — $ blocked by x
+    ("(^a)(b$)", "ab",  "search"),     # match — both anchors: exactly "ab"
+    ("(^a)(b$)", "xab", "search"),     # noMatch — ^ blocks
+    ("(^a)(b$)", "abx", "search"),     # noMatch — $ blocks
+
+    # Match mode: $ inside group cuts off trailing content
+    ("(a$|b)c",  "bc",  "match"),      # match
+    ("(a$|b)c",  "bcd", "match"),      # match — trailing d allowed
+
```
joehendrix, line 541: **Missing test coverage for `$` in the `(false, false)` concat branch**

The corpus has thorough tests for `^` in the `(false, false)` case (`a?^b`, `a?^b?`) but no symmetric tests for `$`. The `(false, false)` branch in `ReToCore.lean:260-275` only case-splits on `atStart && r2.containsAnchorStart` — it does not handle the symmetric `atEnd && r1.containsAnchorEnd && r2.hasNonAnchorContent`.

`a?$b?` exposes this: `$` should only fire when `b?=""`, but without the split, `$` fires even when `b?` matches non-empty.

| Regex | String | Mode | Python | Strata | Verdict |
|-------|--------|------|--------|--------|---------|
| `a?$b?` | `""` | fullmatch | match | match | agree |
| `a?$b?` | `"a"` | fullmatch | match | match | agree |
| `a?$b?` | **`"b"`** | fullmatch | **noMatch** | **match** | **false positive** |
| `a?$b?` | **`"ab"`** | fullmatch | **noMatch** | **match** | **false positive** |
| `a?$b?` | `""` | match | match | match | agree |
| `a?$b?` | `"a"` | match | match | match | agree |
| `a?$b?` | **`"b"`** | match | **noMatch** | **match** | **false positive** |

Suggested test additions:

```suggestion
    ("(a$|b)c",  "bcd", "match"),      # match — trailing d allowed

    # ── Bug-exposing: $ in (false,false) concat — missing atEnd case-split ──────
    #
    # a?$b?: both sides may be empty. $ in r1 should only fire when r2="".
    # Without the atEnd split, $ fires even when b? matches non-empty.

    ("a?$b?", "",   "fullmatch"),  # match — a?="", $ fires, b?=""
    ("a?$b?", "a",  "fullmatch"),  # match — a?="a", $ fires, b?=""
    ("a?$b?", "b",  "fullmatch"),  # noMatch — $ at pos 0 but string non-empty; no valid split
    ("a?$b?", "ab", "fullmatch"),  # noMatch — $ can't fire: b? non-empty after $
    ("a?$b?", "",   "match"),      # match
    ("a?$b?", "a",  "match"),      # match
    ("a?$b?", "b",  "match"),      # noMatch

```

Strata/Languages/Python/Regex/ReToCore.lean
```diff
@@ -75,98 +121,192 @@ Unmatchable regex pattern.
...
+        let r1b_eps   := mkApp () (.op () reInterFunc.name none) [r1b, Core.emptyRegex]
+        let r2b_start := toCore r2 atStart atEnd
+        let r2b_mid   := toCore r2 false atEnd
+        mkApp () (.op () reUnionFunc.name none)
+          [mkApp () (.op () reConcatFunc.name none) [r1b_eps, r2b_start],
+           mkApp () (.op () reConcatFunc.name none) [r1b, r2b_mid]]
+      else
+        let r1b := toCore r1 atStart false
+        let r2b := toCore r2 true atEnd
+        mkApp () (.op () reConcatFunc.name none) [r1b, r2b]
```
joehendrix: **`(false, true)` else branch passes `atStart=true` unconditionally to `r2`**

When the `if` guard on line 247 fails because `atStart=false` (outer context says something was already consumed), the else branch unconditionally passes `true` as the `atStart` argument to `r2`:

```lean
let r2b := toCore r2 true atEnd
```

This causes `^` in `r2` to fire even though the position is not at the start of the string.

**Example:** In `search` mode, `pythonRegexToCore` generates `core_ft` and `core_ff` with `atStart=false`. For `concat(^, ^a)` where `r1 = ^` (no non-anchor content), the guard fails. The else branch passes `atStart=true` to `r2 = ^a`, so `^` in `r2` fires mid-string.

**Concrete test case — `a?(^b)` in search mode:**

| regex | string | mode | Python | Strata | verdict |
|-------|--------|------|--------|--------|---------|
| `a?(^b)` | `"xb"` | search | noMatch | **match** | BUG |
| `a?(^b)` | `"ab"` | search | noMatch | **match** | BUG |
| `a?(^b)` | `"xab"` | search | noMatch | **match** | BUG |

The fix should pass the inherited `atStart` flag instead of literal `true`:

```suggestion
      else
        let r1b := toCore r1 atStart false
        let r2b := toCore r2 atStart atEnd
        mkApp () (.op () reConcatFunc.name none) [r1b, r2b]
```

StrataTest/Languages/Python/Regex/diff_test.py
```diff
@@ -0,0 +1,728 @@
...
+    ("[a-z]+", "hello", "fullmatch"),  # match
+    ("[a-z]+", "Hello", "fullmatch"),  # noMatch — H is uppercase
+
+    # Negated classes [^...]: complement intersected with allchar
+    ("[^a-z]",  "A",     "fullmatch"),  # match
+    ("[^a-z]",  "a",     "fullmatch"),  # noMatch
+    (r"[^b]",   "a",     "fullmatch"),  # match
+    (r"[^b]",   "b",     "fullmatch"),  # noMatch
+    (r"[^A-Z]+","hello",  "fullmatch"),  # match
+    (r"[^A-Z]+","Hello",  "fullmatch"),  # noMatch — H is uppercase
```
joehendrix, line 79: **Missing negated-class tests for multi-char and empty strings**

The complement fix (intersecting `re.comp` with `re.allchar()`) is one of the core changes in this PR, but all `[^x]` tests use single-character test strings. The old code — bare `re.comp(X)` without the `re.allchar` intersection — would incorrectly match multi-char strings and empty strings, which is exactly what the fix addresses.

Adding these cases would serve as regression tests for the complement fix:

```suggestion
    # Negated classes [^...]: complement intersected with allchar
    ("[^a-z]",  "A",     "fullmatch"),  # match
    ("[^a-z]",  "a",     "fullmatch"),  # noMatch
    (r"[^b]",   "a",     "fullmatch"),  # match
    (r"[^b]",   "b",     "fullmatch"),  # noMatch
    (r"[^A-Z]+","hello",  "fullmatch"),  # match
    (r"[^A-Z]+","Hello",  "fullmatch"),  # noMatch — H is uppercase
    # Regression: complement must be single-char, not multi-char or empty
    (r"[^a]",   "",      "fullmatch"),  # noMatch — must be exactly 1 char
    (r"[^a]",   "bc",    "fullmatch"),  # noMatch — 2 chars, not 1
    (r"[^a]",   "bb",    "fullmatch"),  # noMatch — 2 chars
    (r"[^b]+",  "",      "fullmatch"),  # noMatch — below min (1)
```

These all return `noMatch` in Python and would have returned `match` under the old `re.comp`-only implementation.

StrataTest/Languages/Python/Regex/diff_test.py
```diff
@@ -0,0 +1,728 @@
...
+    st_smt_error     = st_result.startswith("smtError:")
+
+    # Agreement
+    if py_match   and st_match:        return "agree"
+    if py_nomatch and st_nomatch:      return "agree"
+    if py_error   and st_parse_error:  return "agree"
+
+    # Bugs
+    if py_match   and st_nomatch:                    return "bug"
+    if py_nomatch and st_match:                      return "bug"
```
joehendrix, line 884: **Subprocess failure is silently tolerated**

If the subprocess fails entirely (e.g., `lake` not found, build failure), all results become `"smtError:missing_output"` via the `.get()` default in `main()` (line 690) → classified as `"investigate"`, not a hard error. A total infrastructure failure could be masked as individual test anomalies.

Consider checking for an empty result set and raising early:

```python
    strata_results = run_strata(CORPUS, args.lake_exe, log_dir=args.log_dir)
    if not strata_results:
        print("ERROR: No results from DiffTestCore — check that 'lake exe DiffTestCore' works.",
              file=sys.stderr)
        return 2
```

This would surface build/path issues immediately rather than producing a wall of `"investigate"` entries.

Strata/Languages/Python/Regex/ReToCore.lean
```diff
@@ -75,98 +121,192 @@ Unmatchable regex pattern.
...
-  | .union r1 r2 =>
-      let r1b := toCore r1 atStart atEnd
-      let r2b := toCore r2 atStart atEnd
-      mkApp () (.op () reUnionFunc.name none) [r1b, r2b]
+      -- Both sides may be empty.
+      -- Case 1: (`r1=""`, `^` fires correctly) and
+      -- Case 2: (`r1` non-empty, `^` must not fire, so `atStart=false` for `r2`).
+      if atStart && r2.containsAnchorStart && r1.hasNonAnchorContent then
+        let r1b       := toCore r1 atStart atEnd
+        let r1b_eps   := mkApp () (.op () reInterFunc.name none) [r1b, Core.emptyRegex]
```
joehendrix: `r1b_eps` (and a few other variables here) is snake case while other variables are consistently lowerCamelCase.

DiffTestCore.lean
```diff
@@ -0,0 +1,142 @@
...
+--
+-- When --log-dir <path> is given, each generated .core.st program is written to
+-- <path>/<n>_<mode>.core.st (1-indexed) with a header comment showing the
+-- original regex and string, for inspection and debugging.
+--
+-- Restriction: <regex> and <string> fields must not contain tab characters.
+
+import Strata.Languages.Core.Verifier
+import Strata.Languages.Python.Regex.ReToCore
+import Strata.Languages.Core.DDMTransform.ASTtoCST
```
joehendrix: **Redundant import**

`ASTtoCST` is already transitively imported via `Strata.Languages.Core.Verifier` (line 30).

```suggestion
```

Strata/Languages/Python/Regex/ReToCore.lean
```diff
@@ -63,6 +63,52 @@ def RegexAST.alwaysConsume (r : RegexAST) : Bool :=
...
+  | .anychar       => true
+  | .complement _  => true
+  | .concat r1 r2  => r1.hasNonAnchorContent || r2.hasNonAnchorContent
+  | .union  r1 r2  => r1.hasNonAnchorContent || r2.hasNonAnchorContent
+  | .star   r      => r.hasNonAnchorContent
+  | .plus   r      => r.hasNonAnchorContent
+  | .optional r    => r.hasNonAnchorContent
+  | .loop   r _ _  => r.hasNonAnchorContent
+  | .group  r      => r.hasNonAnchorContent
+  | _              => false
```
joehendrix: **Three AST-walking helpers share identical recursive structure**

`containsAnchorEnd`, `containsAnchorStart`, and `hasNonAnchorContent` all exhaustively match every `RegexAST` constructor with the same recursive pattern — they differ only in which leaf constructors return `true`. A generic fold would eliminate ~30 lines of duplication:

```lean
/-- Generic predicate: returns true if any node in the AST satisfies `p`. -/
def RegexAST.anyNode (p : RegexAST → Bool) (r : RegexAST) : Bool :=
  p r || match r with
  | .concat r1 r2 => r1.anyNode p || r2.anyNode p
  | .union  r1 r2 => r1.anyNode p || r2.anyNode p
  | .star   r     => r.anyNode p
  | .plus   r     => r.anyNode p
  | .optional r   => r.anyNode p
  | .loop   r _ _ => r.anyNode p
  | .group  r     => r.anyNode p
  | _             => false
```

Then the three helpers become one-liners:

```lean
def RegexAST.containsAnchorEnd   := anyNode (· matches .anchor_end)
def RegexAST.containsAnchorStart := anyNode (· matches .anchor_start)
def RegexAST.hasNonAnchorContent  := anyNode fun
  | .char _ | .range _ _ | .anychar | .complement _ => true
  | _ => false
```

Not blocking — the current code is correct and clear. Just a readability/maintenance improvement.

StrataTest/Languages/Python/Regex/diff_test.py
```diff
@@ -0,0 +1,728 @@
...
+    (".*^a",  "ba",   "search"),     # noMatch
+    (".*^a",  "xa",   "search"),     # noMatch
+
+    # a?(^b): a? non-consuming with content, (^b) always-consuming with ^
+    ("a?(^b)", "b",   "fullmatch"),  # match — a? = "", ^ fires, b matches
+    ("a?(^b)", "ab",  "fullmatch"),  # noMatch — a? = "a" → ^ past start; a? = "" → b ≠ a
+    ("a?(^b)", "b",   "match"),      # match
+    ("a?(^b)", "ba",  "match"),      # match — a? = "", ^ fires, b matches, trailing a ok
+    ("a?(^b)", "ab",  "match"),      # noMatch
+
```
joehendrix, line 375: **Missing `a?(^b)` tests in `search` mode — exposes issue discussed [here](https://github.com/strata-org/Strata/pull/517#discussion_r2894469162).

The existing `a?(^b)` tests only cover `fullmatch` and `match` modes, where `atStart=true`. In `search` mode, `pythonRegexToCore` generates `core_ft` and `core_ff` with `atStart=false`, which triggers the `(false, true)` else branch bug (line 258 passes `atStart=true` unconditionally).

| regex | string | mode | Python | Strata | verdict |
|-------|--------|------|--------|--------|---------|
| `a?(^b)` | `"b"` | search | match | match | agree |
| `a?(^b)` | `"xb"` | search | noMatch | **match** | **BUG** |
| `a?(^b)` | `"ab"` | search | noMatch | **match** | **BUG** |
| `a?(^b)` | `"xab"` | search | noMatch | **match** | **BUG** |
| `a?(^b)` | `"ba"` | search | match | match | agree |
| `a?(^b)` | `"bx"` | search | match | match | agree |

```suggestion
    ("a?(^b)", "ab",  "match"),      # noMatch

    # a?(^b) in search: core_ft/core_ff have atStart=false; ^ must not fire mid-string
    ("a?(^b)", "b",   "search"),     # match — pos 0: a?="", ^ fires, b matches
    ("a?(^b)", "xb",  "search"),     # noMatch — BUG: Strata says match (^ fires mid-string)
    ("a?(^b)", "ab",  "search"),     # noMatch — BUG: Strata says match
    ("a?(^b)", "xab", "search"),     # noMatch — BUG: Strata says match
    ("a?(^b)", "ba",  "search"),     # match — pos 0: a?="", ^ fires, b matches
    ("a?(^b)", "bx",  "search"),     # match — pos 0: a?="", ^ fires, b matches
```

Strata/Languages/Python/Regex/ReToCore.lean
shigoel: I don't write code as consistently as an agent does, and it shows :) Thanks.

Strata/Languages/Python/Regex/ReToCore.lean
shigoel: Thanks! I'd kept them separate as I was figuring out what to do with the complement case which doesn't contain any positional constraints, but this consolidation makes the semantics easier to follow.

StrataTest/Languages/Python/Regex/diff_test.py
shigoel: Thanks, suggestion applied which'd raise this error if the lengths of the results and corpus don't match.

Strata/Languages/Python/Regex/ReToCore.lean
shigoel: Will refactor, thanks!

shigoel:
> This is great! I really like the encoding.
> 
> One nitpick: desugaring `r+` to `rr*` in `RegexAST.toCore` causes an exponential blowup when you have nested `+` constructors. Do we really need to do that? The core language includes a `+` operator, why not just encode it directly?
> 
> One simple test case to add that shows this off would be `#eval Std.format $ pythonRegexToCore "((((((a+)+)+)+)+)+)+" .fullmatch`.
> 
> This is a nitpick because it is probably unlikely to come up in practice very frequently.

The reason we do this is anchors, yet again. But you're right: I can add a fast path that uses `re.+` when the regex doesn't have an anchor.

PR #518, opened 2026-03-05
Add SMT denotations and translation support

kondylidou:
*Issue #, if available:*

*Description of changes:*
This PR introduces the SMT denotation layer and related translation utilities as a standalone upstream step.

It adds:
- `Strata/DL/SMT/Denote.lean`: semantic interpretation of SMT terms/queries into Lean-level meaning.
- `Strata/DL/SMT/Translate.lean`: translation/state machinery used to bridge Strata/Core-side structures with SMT representations.


Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1276 @@
...
+mutual
+
+/--
+Interpret an SMT `TermType` as a Lean `Type`, when supported.
+
+Returns `none` when we lack an interpretation (e.g. for reals).
+-/
+def denoteSort (sctx : SortContext) (ty : TermType) : Option (SortDenoteResult sctx) := do
+  match ty with
+  | .prim .bool => return fun _ => Prop
```
joscoh: I would split the `prim` cases into a separate function to avoid duplication.

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1276 @@
...
+-/
+structure TermContext where
+  vs  : TermVarContext := []
+  ufs : UFContext := []
+
+/--
+Semantic environment providing Lean values for the elements stored in a
+`Context`.
+-/
+structure TermEnvironment (sΓ : SortDenoteInput sctx) where
```
joscoh, line 342: It's not clear to me why you need both of these (e.g. `UFContext` and `UFEnvironment`). The `Environment` gives strictly more information than the `Context`.

shigoel:
@kondylidou We do need copyright notices on all files. Thanks!

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+  h : (denoteSort ctx.sctx ty).isSome
+  res : (tdi : TermDenoteInput ctx) → (denoteSort ctx.sctx ty).get h ⟨tdi.sΓ, tdi.hsΓ⟩
+
+-- Note: `noncomputable` due to a compiler error
+/--
+Check that the denotations `as` match the types of the variables `vs` when the
+two lists have the same length.
+-/
+@[simp]
+noncomputable def argTypesAlignAux (as : List (TermDenoteResult ctx)) (vs : List TermVar) (hl : as.length = vs.length) : Bool :=
```
joscoh: Why does this need the `as.length = vs.length` hypothesis? It is well-defined even without it, and it will always return `false` if the lengths are different (so `argTypesAlign` would have the exact same behavior).

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+  ∀ i, (hi : i < as.length) → as[i].ty = (vs[i]'(argTypesAlign_length_eq h ▸ hi)).ty := by
+  unfold argTypesAlign at h
+  by_cases hl : as.length = vs.length
+  case pos => exact argTypesAlignAux_true (dif_pos hl ▸ h)
+  case neg =>
+    rewrite [dif_neg hl] at h; contradiction
+
+-- Note: `noncomputable` because of a compiler error
+/--
+Apply an interpreted uninterpreted function to a list of argument denotations.
```
joscoh: What is an "interpreted uninterpreted function"? I think from the definition it means that you apply a function symbol to some inputs, but I could be wrong.

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+-/
+@[simp]
+noncomputable def applyUF (tdi : TermDenoteInput ctx) (uf : (denoteFunSort ctx.sctx args out).get h ⟨tdi.sΓ, tdi.hsΓ⟩)
```
joscoh: Very minor, but is there a reason this doesn't just take in `argTypesAlign` and then construct the needed proof with `argTypesAlign_arg_types`?

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+  match args, as with
+  | [], []            => uf
+  | arg :: _, a :: as =>
+    let uf := arrow_of_denoteFunSortCons_isSome h ▸ uf
+    have ha : denoteSort ctx.sctx arg.ty = denoteSort ctx.sctx a.ty := has 0 (Nat.zero_lt_succ _) ▸ rfl
+    let uf := uf (Option.get_congr ha ▸ a.res tdi)
+    applyUF tdi uf as (Nat.succ.inj hl) (fun i hi => (has (i + 1) (Nat.succ_lt_succ hi)))
+
+@[simp]
+def bindForallVar (ctx : Context) (hTy : (denoteSort ctx.sctx ty).isSome) :
```
joscoh: There should be a comment explaining what this function is doing.

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+          | i + 1 =>
+            have hivs' := Nat.lt_of_succ_lt_succ hivs
+            have hivΓ := Nat.succ_lt_succ (tdi.htΓ.hv.h ▸ hivs')
+            (List.getElem_cons_succ _ ctx.tctx.vs i hivs).symm ▸ (List.getElem_cons_succ _ tdi.tΓ.vs i hivΓ).symm ▸ tdi.htΓ.hv.ha i hivs'
+        { h := h', ha := ha' }
+      let tdi' : TermDenoteInput ⟨ctx.sctx, tctx'⟩ :=
+        { sΓ := tdi.sΓ, hsΓ := tdi.hsΓ, tΓ := { ufs := tdi.tΓ.ufs, vs := vΓ' }, htΓ := { hv := hv', huf := tdi.htΓ.huf } }
+      ft' tdi'
+
+def buildExists (ctx : Context) (vs : List TermVar)
```
joscoh: As far as I can tell, this is the exact same as `buildForall`, but in uses `bindExistsVar` instead of `bindForallVar` (both of which have the same type). They should be combined into a single function.

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+/-
+Noncomputable because of `ite` case. Two conditions are needed to make this function computable:
+- Disallow quantifiers in the condition of `ite`.
+- Return a proof of decidability for each other `Prop`.
+-/
+/--
+Attempt to interpret a single SMT term under `ctx`, returning its Lean type
+and semantics when successful.
+-/
+noncomputable def denoteTerm (ctx : Context) (t : Term) : Option (TermDenoteResult ctx) := do
```
joscoh, line 563: The `var` and `app` cases are difficult to read because they contain a large mix of proof terms and only a small part is the actual "interesting" result. Giving a short comment explaining the intended behavior would help.

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+    let as ← denoteTerms ctx as
+    leftAssoc ctx (.prim .int) rfl (fun _ => @HSub.hSub Int Int Int _) as
+  | .app .add as _ =>
+    let as ← denoteTerms ctx as
+    leftAssoc ctx (.prim .int) rfl (fun _ => @HAdd.hAdd Int Int Int _) as
+  | .app .mul as _ =>
+    let as ← denoteTerms ctx as
+    leftAssoc ctx (.prim .int) rfl (fun _ => @HMul.hMul Int Int Int _) as
+  | .app .div as _ =>
+    -- Semantic mismatch with SMT-Lib: `x / 0` is defined as `0` in Lean, but as UF in SMT-Lib.
```
joscoh, line 674: Should the semantics return `none` if the denominator is `0`?

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+/--
+Interpret a ground integer term in the empty context.
+-/
+@[simp]
+noncomputable def denoteIntTermAux (t : Term) : Option Int := do
+  let some ⟨.prim .int, _, fi⟩ := denoteTerm {} t | none
+  return fi ⟨[], { h := rfl, ha := fun _ hi => nomatch hi }, ⟨[], []⟩, ⟨{ h := rfl, ha := fun _ hi => nomatch hi }, { h := rfl, ha := fun _ hi => nomatch hi }⟩⟩
+
+@[simp]
+noncomputable def bindUS (uss iss) {us'} (ft' : SortDenoteInput ⟨(us' :: uss), iss⟩ → Prop) :
```
joscoh, line 984: What is this function doing?

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+    else
+      none
+  else
+    none
+
+/--
+Interpret a boolean term using the bindings defined in an SMT context.
+-/
+@[simp]
+noncomputable def denoteBoolTermFromContext
```
joscoh, line 1149: I would appreciate more detail about these functions (`bindIF`, this one, etc). Why do we need a separate boolean denotation function beyond `denoteTerm`, which already handles booleans?

Strata/DL/SMT/Denote.lean
```diff
@@ -0,0 +1,1282 @@
...
+  if !ctx.typeFactory.isEmpty || !ctx.seenDatatypes.isEmpty || !ctx.datatypeFuns.isEmpty then none
+  let stmt := assums.foldr (.app .implies [·, ·] (.prim .bool)) conc
+  let t := ctx.axms.foldr (.app .implies [·, ·] (.prim .bool)) stmt
+  let uss := ctx.sorts.toList.reverse
+  let iss := (mkISContext ctx.tySubst).reverse
+  let ufs := ctx.ufs.toList.reverse
+  let ifs := ctx.ifs.toList.reverse
+  (denoteBoolTermFromContext uss iss ufs ifs t).map PLift.down
+
+#reduce denoteIntTermAux (.app .add [.prim (.int 1), .prim (.int 2)] (.prim .int))
```
joscoh: These should use `#guard_msgs` so the output is not visible. Also, to be consistent with Strata conventions, these should be in a separate test file in `StrataTest/DL/SMT/DenoteTest.lean` (or similar).

Strata/DL/SMT/Translate.lean
```diff
@@ -0,0 +1,645 @@
...
+    leftAssocOp mkIntSub as
+  | .app .add as _ =>
+    leftAssocOp mkIntAdd as
+  | .app .mul as _ =>
+    leftAssocOp mkIntMul as
+  | .app .div as _ =>
+    leftAssocOp mkIntDiv as
+  | .app .mod as _ =>
+    leftAssocOp mkIntMod as
+  -- | .app .abs [a] _ =>
```
joscoh: Is there a reason this is commented out?

Strata/DL/SMT/Translate.lean
```diff
@@ -0,0 +1,645 @@
...
+    modify fun s => { s with level := s.level + 1 }
+    return (as.push a)
+  let as ← as.foldlM f #[]
+  let (_, a) ← translateTerm a
+  modify fun s => { s with level := level + 1 }
+  return as.foldr mkArrow a
+
+def withTypeDecls (uss : Array Core.SMT.Sort) (k : TranslateM Expr) : TranslateM Expr := do
+  let state ← get
+  let mut decls ← uss.mapM translateTypeDecl
```
joscoh: Why is this mutable? It doesn't seem to ever be updated

Strata/DL/SMT/Translate.lean
```diff
@@ -0,0 +1,645 @@
...
+  translateTypeDef (is : String × TermType) : TranslateM (Name × Expr × Expr) := do
+    let n := symbolToName is.fst
+    let t := .sort 1
+    let v ← translateSort is.snd
+    modify fun s => { level := s.level + 1, bvars := s.bvars.insert (.us { name := is.fst, arity := 0 }) (t, s.level) }
+    return (n, t, v)
+
+def withFunDecls (ufs : Array UF) (k : TranslateM Expr) : TranslateM Expr := do
+  let state ← get
+  let mut decls ← ufs.mapM translateFunDecl
```
joscoh: Same as above

Strata/DL/SMT/Translate.lean
```diff
@@ -0,0 +1,645 @@
...
+    let t := ps.foldr (fun (n, t) b => .forallE n t b .default) s
+    set { level := state.level + 1, bvars := state.bvars.insert (.uf uf) (t, state.level) : Translate.State }
+    return (n, t)
+  translateParam (v : TermVar) : TranslateM (Name × Expr) := do
+    let n := symbolToName v.id
+    let t ← translateSort v.ty
+    modify fun s => { level := s.level + 1, bvars := s.bvars.insert (.bv v) (t, s.level) }
+    return (n, t)
+
+def withFunDefs (ifs : Array Core.SMT.IF) (k : TranslateM Expr) : TranslateM Expr := do
```
joscoh, line 589: It would be helpful for this group of functions (`withFunDecls`, `withTypeDefs`, etc) to have comments explaining their purpose. Am I correct that they are collecting uninterpreted type and function symbols to quantify over?

Strata/DL/SMT/Denote.lean
kondylidou: `Environment` is richer, but we keep `Context` separate, so denotation is indexed by declarations, and `WF` explicitly proves the environment matches those declarations (especially during binder extension/lookups). I’ve clarified this in comments and added `toContext` projections.

Strata/DL/SMT/Denote.lean
kondylidou: cleaned it up

Strata/DL/SMT/Denote.lean
kondylidou: You're right. I changed it to `Apply the semantic interpretation of a UF symbol to denoted arguments.
`

Strata/DL/SMT/Denote.lean
kondylidou: At this stage, the denominator is usually symbolic, so we can’t reliably check = 0 yet. Returning none only for / 0 isn’t sound here.

Strata/DL/SMT/Denote.lean
kondylidou: Discharges one uninterpreted sort declaration by quantifying over all its possible semantic realizations.

Strata/DL/SMT/Denote.lean
joscoh: I know you cannot check syntactically if the denominator is zero, but is it possible to evaluate the denotation of the denominator? Maybe not, I am just wondering.

Strata/DL/SMT/Denote.lean
```diff
@@ -377,21 +393,23 @@ Check that the argument denotations align with the declared term-variable types.
 -/
 @[simp]
 noncomputable def argTypesAlign (as : List (TermDenoteResult ctx)) (vs : List TermVar) : Bool :=
```
joscoh: This definition is not needed. It is exactly the same as `argTypesAlignAux` (because if the lengths are different, `argTypesAlignAux` gives `false` anyway).

Strata/DL/SMT/Denote.lean
```diff
@@ -377,21 +393,23 @@ Check that the argument denotations align with the declared term-variable types.
...
-  if h : as.length = vs.length then
-    argTypesAlignAux as vs h
+  if _ : as.length = vs.length then
+    argTypesAlignAux as vs
   else
     false
 
-theorem argTypesAlignAux_true (h : argTypesAlignAux as vs hl) :
+theorem argTypesAlignAux_true (h : argTypesAlignAux as vs) (hl : as.length = vs.length) :
```
joscoh: It is fine to keep this, but I will note that this length hypothesis is not necessary here. You can rule out the false case directly from the assumption that `argTypesAlignAux as vs` holds (or indirectly via `argTypesAlign_length_eq` and your current proof).

Strata/DL/SMT/Denote.lean
```diff
@@ -421,8 +441,26 @@ noncomputable def applyUF (tdi : TermDenoteInput ctx) (uf : (denoteFunSort ctx.s
...
+@[simp]
+noncomputable def applyUF (tdi : TermDenoteInput ctx) (uf : (denoteFunSort ctx.sctx args out).get h ⟨tdi.sΓ, tdi.hsΓ⟩)
+  (as : List (TermDenoteResult ctx)) (hAlign : argTypesAlign as args) :
+    (denoteSort ctx.sctx out).get (denoteSortOut_isSome_of_denoteFunSort_isSome h) ⟨tdi.sΓ, tdi.hsΓ⟩ :=
+  applyUFAux tdi uf as (argTypesAlign_length_eq hAlign).symm (argTypesAlign_arg_types hAlign)
+
+/--
+Semantics helper for universal quantification.
 
+Given a body interpretation over the context extended with one bound variable
```
joscoh, line 462: Thanks, this is helpful. Can you give a similar comment for `bindExistsVar`?

Strata/DL/SMT/Denote.lean
```diff
@@ -496,7 +519,7 @@ def bindExistsVar (ctx : Context) (hTy : (denoteSort ctx.sctx ty).isSome) :
         { sΓ := tdi.sΓ, hsΓ := tdi.hsΓ, tΓ := { ufs := tdi.tΓ.ufs, vs := vΓ' }, htΓ := { hv := hv', huf := tdi.htΓ.huf } }
       ft' tdi'
 
-def buildExists (ctx : Context) (vs : List TermVar)
+def buildQuant (isForall : Bool) (ctx : Context) (vs : List TermVar)
```
joscoh: Nit: instead of a bool, this could take in a function 
```
f: let tctx := { ufs := ctx.tctx.ufs, vs := ctx.tctx.vs }
    let v' := { id := n, ty := ty }
    let vs' := v' :: ctx.tctx.vs
    let tctx' := { ufs := ctx.tctx.ufs, vs := vs' }
    (TermDenoteInput ⟨ctx.sctx, tctx'⟩ → Prop)
```
And the the last part would be:
```
let ft' := buildQuant isForall ctx' vs hTys' (hvs ▸ bodyFt)
f ctx (denoteFunSortCons_isSome hTys).left ft' tdi
```
But this type is very long and ugly, so it is ultimately a matter of preference.

Strata/DL/SMT/Denote.lean
```diff
@@ -519,7 +519,20 @@ def bindExistsVar (ctx : Context) (hTy : (denoteSort ctx.sctx ty).isSome) :
         { sΓ := tdi.sΓ, hsΓ := tdi.hsΓ, tΓ := { ufs := tdi.tΓ.ufs, vs := vΓ' }, htΓ := { hv := hv', huf := tdi.htΓ.huf } }
       ft' tdi'
 
-def buildQuant (isForall : Bool) (ctx : Context) (vs : List TermVar)
+/--
+Shape of a one-variable quantifier binder combinator (`∀` or `∃`).
+-/
+abbrev QuantVarBinder :=
```
joscoh: Can you update `bindExistsVar` and `bindForallVar` to return this type?

Strata/DL/SMT/Denote.lean
```diff
@@ -376,8 +376,9 @@ structure TermDenoteResult (ctx : Context) where
...
 -- Note: `noncomputable` due to a compiler error
 /--
-Check that the denotations `as` match the types of the variables `vs` when the
-two lists have the same length.
+Check that denoted argument types match declared variable types.
+
+If lengths differ, this returns `false`.
 -/
 @[simp]
 noncomputable def argTypesAlignAux (as : List (TermDenoteResult ctx)) (vs : List TermVar) : Bool :=
```
joscoh: Minor and non-blocking, but this could just be `argTypesAlign`, since it not an auxiliary definition anymore.

joscoh (approved):

aqjune-aws:
Hi, I promised to leave some questions about SMT semantics in the CSLib meeting by yesterday but I was slightly late, sorry. :) I have a naive question - I would like to prove translation from Lambda to SMT-LIB someday. Can the style of semantics of shallow embedding in this PR be used to a possibly new semantics that support deep-embedding? Of course we cannot immediately reuse `Prop` which is the return type of the functions, but can we somehow use the same structure in the future.

aqjune-aws (approved):

PR #520, opened 2026-03-05
Fix VCG path multiplication with goto in if-then-else

olivier-aws:
## Summary

This PR fixes a critical bug where procedures containing `goto` (exit) statements inside `if-then-else` branches caused verification condition (VC) path multiplication. When multiple paths converged to the same exit label, each path was treated independently through `ProgramEval`, duplicating proof obligations for subsequent procedures.

*Issue #419

> 🤖 **Note:** This PR was created with assistance from an AI agent.

## Changes

**1. Block-level path merging** (`fix(StatementEval): Merge block handler results`)
* Added `mergeByExitLabel` in the block handler to merge environments that converge to the same exit label
* Prevents path multiplication by ensuring at most one result per distinct exit label

**2. Fixed path merging logic** (`fix(StatementEval): Address review feedback`)
* Fixed `Env.merge` argument order in `mergeByExitLabel` (true-branch = E1, false-branch = E2)
* Fixed condition extraction to inspect path condition structure instead of blindly using newest entry
* Fixed repeated merge corruption by replacing `foldl`-based merge with exact 2-path matching
* Moved merge logic into `processIteBranches` for same-label exits

**3. Documentation** (`docs(StatementEval): Clarify safety of label-based path merging`)
* Added comprehensive documentation explaining why block-level merging is safe
* Created `LABEL_MERGING_SAFETY.md` with detailed explanation of the two-level merging strategy

**4. Test updates** (`test: Update T8b_EarlyReturnPostconditions`)
* Updated test expectations to reflect that only one error is now produced instead of the previous duplicate

**5. Distinct block label validation** (`feat(StatementType): Add check for distinct block labels`)
* Added type checker validation to enforce that all labels within a single block are distinct
* The path merging fix relies on this property; the check rejects Core programs that violate it

**6. BoogieToStrata duplicate label elimination** (`fix(BoogieToStrata): Eliminate duplicate block labels`)
* Fixed duplicate labels from exit-target wrappers in `EmitWithExitWrappers`
* Renamed synthetic `_exit` label to `__strata_return` to avoid name collisions with user-defined labels
* Updated expect files for ForwardGotos, B, and Bubble

**7. Procedure inlining unique labels** (`fix(inlining): Generate unique labels for inlined procedure blocks`)
* Used `genIdent` to produce a unique label for each inlined block instead of the static `procName ++ "$inlined"` string
* Prevents duplicate label errors when the same procedure is inlined multiple times within a single block
* Updated Python test expected outputs to reflect corrected behavior (removed duplicate assertion results that were artifacts of the bug)

## Files Changed
* `Strata/Languages/Core/StatementEval.lean` — path merging logic
* `Strata/Languages/Core/StatementType.lean` — distinct label validation
* `Strata/Transform/ProcedureInlining.lean` — unique inlined block labels
* `Tools/BoogieToStrata/` — duplicate label fixes
* `StrataTest/Languages/Core/Examples/ExitItePathMerge.lean` — new tests
* `StrataTest/Languages/Python/expected_non_laurel/` — updated expected outputs

## Status

🚧 **Work in Progress** - Testing and review in progress

### DIFF
--- a/Examples/StringTest.laurel.st
+++ b/Examples/StringTest.laurel.st
@@ -4,7 +4,7 @@ requires true
 {
   var message: string := "Hello, World!";
   return message;
-}
+};
 
 procedure testStringConcat()
 returns (result: string)
@@ -13,4 +13,4 @@ requires true
   var hello: string := "Hello";
   var world: string := "World";
   return hello;
-}
+};
--- a/Strata/DDM/Format.lean
+++ b/Strata/DDM/Format.lean
@@ -89,6 +89,8 @@ Options to control parenthesis
 structure FormatOptions where
   /-- Always add parenthesis when feasible. -/
   alwaysParen : Bool := false
+  /-- Use SMT-LIB 2.7 string escaping (`""` for quotes) instead of C-style (`\"`). -/
+  smtStringEscaping : Bool := false
 
 /--
 A format context provides callbacks and information needed to
@@ -383,7 +385,10 @@ private partial def ArgF.mformatM {α} : ArgF α → FormatM PrecFormat
 | .ident _ x => return .atom (formatIdent x)
 | .num _ x => pformat x
 | .decimal _ v => pformat v
-| .strlit _ s => return .atom (.text <| escapeStringLit s)
+| .strlit _ s => do
+    let ctx ← read
+    let esc := if ctx.opts.smtStringEscaping then escapeSMTStringLit s else escapeStringLit s
+    return .atom (.text esc)
 | .bytes _ v => return .atom <| .text <| ByteArray.escapeBytes v
 | .option _ ma =>
   match ma with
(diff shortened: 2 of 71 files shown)


keyboardDrummer:
Have you reviewed the change yourself? I think it would be good to let someone who understands this code well review this change since it's a complicated piece of code and I think if we're not careful it'll only get more complex.

Also, while this PR fixes the issue for the test-case that we have, we can make the following fix in ProgramEval to ensure at a code level that we don't run into this issue again: https://github.com/strata-org/Strata/blob/9bdd2cc87a3737315f8ad155166c0c8a6340203d/Strata/Languages/Core/ProgramEval.lean

However, making that fix uncovers more issues in the partial evaluator so you'd have to fix those as well then.

olivier-aws:
> Have you reviewed the change yourself? I think it would be good to let someone who understands this code well review this change since it's a complicated piece of code and I think if we're not careful it'll only get more complex.
> 

I asked @shigoel and @atomb to review it

PR #301, opened 2026-01-03
Modes for controlling irrelevant axiom removal

shigoel:
*Description of changes:*

Make the detection of relevant axioms for a given function more precise. See test in `irrelevantAxiomsTestPgm2` in `Strata/StrataTest/Languages/Boogie/Examples/RemoveIrrelevantAxioms.lean` for details.

There are three modes to choose from for detecting and removing irrelevant axioms:

- `Off`: do not remove any axioms.

- `Aggressive`: Only the functions in the consequent `Q` of a goal `P ==> Q` will be taken into account for relevant axiom analysis. This means that axioms relevant to some function in `P` may be removed. This is sound, but may yield unknown results from the solver.

- `Precise`: Functions in both `P` and `Q` of a goal `P ==> Q` will be considered for relevant axiom analysis. A caveat is that only the antecedents that appear after a marker called `__end_of_prelude_marker` will be considered as the real assumptions here.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #524, opened 2026-03-06
Irrelevant axiom removal implemented as a program transform

shigoel:
## Summary

Revival of #301 (which was written in the Boogie days!), reimplemented for Core.

- Adds `IrrelevantAxiomsMode` (`Off` / `Aggressive` / `Precise`) to
  `Options.lean`, replacing the old `Bool` flag. Soundness documentation is
  included: axiom removal is sound for deductive verification but must not be
  applied to cover obligations(weakening assumptions can make unreachable paths
  appear satisfiable).

- New module `Strata/Transform/IrrelevantAxioms.lean` provides a two-step API:
  `Cache.build` precomputes the function call graph and immediate
  function-to-axiom map once per program; `getIrrelevantAxioms` uses the cached
  data for efficient per-goal relevance queries.

- Relevance is computed via a transitive fixed-point (`computeRelevantAxioms`):
  starting from the functions mentioned in a goal, it finds all axioms that
  mention those functions, expands to the functions mentioned in those axioms
  (plus their call-graph neighbours), and repeats until no new axioms are
  discovered.

- `Aggressive` mode uses only the consequent functions of the goal (`P ==> Q` →
  functions in `Q`); `Precise` mode includes antecedent functions too.
  `Aggressive` is better for finding counterexamples (fewer axioms → solver can
  negate the goal more easily); `Precise` is safer for proving goals.

- Proves termination for all previously-`partial` functions in `CallGraph.lean`:
  BFS closures use an edge-count fuel bound; the mutual `extractCalls*` block
  uses `sizeOf`; `computeRelevantAxioms` uses a declaration-count fuel bound.
  
### DIFF
--- a/Strata/Languages/Core/CallGraph.lean
+++ b/Strata/Languages/Core/CallGraph.lean
@@ -175,6 +175,8 @@ def extractCallsFromProcedure (proc : Procedure) : List String :=
 abbrev ProcedureCG := CallGraph
 abbrev FunctionCG := CallGraph
 
+/-- Build a call graph whose nodes are procedures and whose edges represent
+direct procedure calls. -/
 def Program.toProcedureCG (prog : Program) : ProcedureCG :=
   let procedures := prog.decls.filterMap (fun decl =>
     match decl with
@@ -183,6 +185,9 @@ def Program.toProcedureCG (prog : Program) : ProcedureCG :=
     | _ => none)
   buildCallGraph procedures
 
+/-- Build a call graph whose nodes are user-defined functions and whose edges
+represent calls between them. Builtin functions (e.g. `Bool.And`, arithmetic
+operators) are not listed as nodes; calls to builtins are silently dropped. -/
 def Program.toFunctionCG (prog : Program) : FunctionCG :=
   let functions := prog.decls.filterMap (fun decl =>
     match decl with
@@ -199,6 +204,11 @@ Map from functions to their _immediately_ relevant axioms. An axiom `a` is
 _immediately_ relevant for a function `f` if `f` occurs in the body of `a`,
 including in any trigger expressions. Callees and callers of `f` are not
 associated with `a` in this map.
+
+Builtin functions (e.g. `Bool.And`, `Bool.Implies`) are excluded from the map
+keys. Because builtins appear in nearly every axiom body, including them would
+make almost every axiom "immediately relevant" to any goal that touches a
+builtin, collapsing the relevance filter entirely.
 -/
 def Program.functionImmediateAxiomMap (prog : Program) : FuncAxMap :=
   let axioms := prog.decls.filterMap (fun decl =>
@@ -208,7 +218,9 @@ def Program.functionImmediateAxiomMap (prog : Program) : FuncAxMap :=
 
   let functionAxiomPairs := axioms.flatMap (fun ax =>
     let ops := Lambda.LExpr.getOps ax.e
-    ops.map (fun op => (CoreIdent.toPretty op, ax)))
+    ops.filterMap (fun op =>
+      let fname := CoreIdent.toPretty op
+      if builtinFunctions.contains fname then none else some (fname, ax)))
 
   functionAxiomPairs.foldl
     (fun acc (funcName, ax) =>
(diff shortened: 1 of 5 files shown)


keyboardDrummer:
Are triggers for quantifiers not taken into account here? Boogie uses them in its pruning algorithm. Explanation from Claude:

Consider:
```
axiom forall x, y :: {P(x, y), Q(y)} {R(x)} P(x, y) ==> R(x)
```
This axiom has two trigger groups: `{P, Q}` and `{R}`. The axiom becomes relevant if either all symbols in a trigger group are reachable, or if R alone is reachable. So trigger group `{P, Q}` becomes a "merge node" — both P and Q must be visited before the axiom is reachable through that path. Trigger group {R} is a simple single-symbol incoming edge. 

For triggered forall in positive position (and exists in negative position), the axiom body's incoming edges are discarded — only the trigger-derived incoming edges are kept. This is sound because the solver can only instantiate those quantifiers via their triggers.

-----

Have you also considered removal of irrelevant assume statements? Conceptually the two seem so similar that it seems like we should be able to come up with a method that can prune both of them. Also, the more assumptions you prune the more axioms become irrelevant. When working on Dafny I ran into cases where some assertions only needed a tiny segment of a program to be proven, but the assumptions in the same procedure prevented all of that from being pruned. Pruning the assumptions as well could decrease the final program size by 10-100x for some goals.

Strata/Languages/Core/CallGraph.lean
```diff
@@ -61,32 +61,43 @@ def CallGraph.decrementEdge (cg : CallGraph) (caller : String) (callee : String)
   }
 
 /-- Compute transitive closure of callees; the result does not contain `name`. -/
-partial def CallGraph.getCalleesClosure (cg : CallGraph) (name : String) : List String :=
-  let rec go (visited : List String) (toVisit : List String) : List String :=
+def CallGraph.getCalleesClosure (cg : CallGraph) (name : String) : List String :=
+  -- Fuel bound: initially 1 item is enqueued; processing each new node `n`
+  -- enqueues |callees(n)| more. Total enqueued ≤ 1 + Σ |callees(n)| ≤ 1 + totalEdges.
+  let fuel := cg.callees.toList.foldl (fun acc (_, v) => acc + v.size) 0 + 1
```
aqjune-aws, line 67: It seems you already found a good measure for proving termination of function "go". Would Kiro be able to actually prove termination? It would be fantastic if it could. :)

StrataTest/Languages/Python/expected_non_laurel/test_datetime.expected
```diff
@@ -19,4 +19,4 @@ py_assertion: ✅ pass (at line 25, col 0)
 
 py_assertion: ✅ pass (at line 27, col 0)
 
-Assertion failed at line 30, col 0: py_assertion: ❌ fail
+py_assertion: ✅ pass (at line 30, col 0)
```
aqjune-aws, line 22: Have any idea why this assertion was changed from failure to pass? Was failure a timeout?

Strata/Transform/CoreTransform.lean
```diff
@@ -81,7 +81,8 @@ def isGlobalVar (p : Program) (ident : Expression.Ident) : Bool :=
 /-- Cached results of program analyses that are helpful for program
     transformation. -/
 structure CachedAnalyses where
-  callGraph: Option CallGraph := .none
+  callGraph : Option CallGraph := .none
+  axiomMap  : Option FuncAxMap := .none
```
aqjune-aws, line 85: It seems this field wasn't used.

Strata/Transform/IrrelevantAxioms.lean
```diff
@@ -0,0 +1,58 @@
...
+-/
+def getIrrelevantAxioms (prog : Program) (cache : Cache) (functions : List String)
+    : List String :=
+  let allAxioms := prog.decls.filterMap (fun decl =>
+    match decl with | .ax a _ => some a.name | _ => none)
+  let relevantAxioms := functions.flatMap (fun f =>
+    let initialFns :=
+      (f :: cache.funcCG.getCalleesClosure f ++ cache.funcCG.getCallersClosure f).dedup
+    computeRelevantAxioms prog cache.funcCG cache.axiomMap initialFns []) |>.dedup
+  allAxioms.filter (fun a => a ∉ relevantAxioms)
```
aqjune-aws, line 55: IIUC IrrelevantAxiom is more like program analysis! It traverses a program and finds irrelevant axioms. It doesn't seem to directly change the program, however (making it a proper target of being CoreTransformM).
Should we have, say, a new Strata/Analysis/ directory and make `IrrelevantAxioms` the first descendant of the new category? :)

StrataTest/Languages/Python/expected_non_laurel/test_datetime.expected
shigoel: Quick answer for now: the old irrelevant axiom removal was throwing away a relevant axiom, so we couldn't prove this.

PR #526, opened 2026-03-06
Unify recursive and mutually recursive types

joscoh:
*Issue #, if available:*

*Description of changes:*

Replaces the separate `command_datatype / command_mutual` operations with a single `command_datatypes` that handles both single and mutually recursive datatypes uniformly. All datatype declarations now go through `preRegisterTypes`, so type names are always visible to all constructors regardless of whether
recursion is self- or mutual.

This also changes the syntax for mutually recursive datatypes: instead of `mutual ... end`, datatypes in a mutual block have a newline between them, and the block ends with a semicolon. This means that single-datatype mutual blocks and single datatypes are exactly equivalent. This also makes the translation in `Translate.lean` much simpler.

In the DDM, this removes `scopeDatatype`, which is no longer needed, and adds a `scopeTVar`, which converts `.type` bindings to `.tvar` bindings for constructor elaboration.

Note that this is NOT backwards-compatible, since it changes the syntax for mutually recursive types.

### Changes

### DIFF
--- a/.kiro/settings/mcp.json
+++ b/.kiro/settings/mcp.json
@@ -24,7 +24,8 @@
         "lean_state_search",
         "lean_hammer_premise",
         "lean_file_outline",
-        "lean_local_search"
+        "lean_local_search",
+        "lean_verify"
       ]
     }
   }
--- a/Strata/DL/Lambda/FactoryWF.lean
+++ b/Strata/DL/Lambda/FactoryWF.lean
@@ -10,7 +10,6 @@ import all Strata.DL.Lambda.Factory
 public import Strata.DL.Lambda.LExprEval
 public import Strata.DL.Lambda.LExprWF
 public import Strata.DL.Lambda.LTy
-public import Strata.DL.Lambda.LExprTypeSpec
 public import Strata.DL.Lambda.Semantics
 public import Strata.DDM.Util.Array
 public import Strata.DL.Util.Func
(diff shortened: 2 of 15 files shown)


Strata/DDM/AST.lean
```diff
@@ -832,20 +833,15 @@ def argScopeLevel (argDecls : ArgDecls) (level : Fin argDecls.size) : Option (Fi
...
-      if h2 : typeParamsIdx < level.val then
-        some (⟨level.val - (nameIdx + 1), by omega⟩, ⟨level.val - (typeParamsIdx + 1), by omega⟩)
-      else
-        panic! s!"scopeDatatype typeParams index {typeParamsIdx} out of bounds ({level.val})"
+  | some idx =>
+    if h : idx < level.val then
+      some ⟨level.val - (idx + 1), by omega⟩
     else
-      panic! s!"scopeDatatype name index {nameIdx} out of bounds ({level.val})"
+      panic! s!"scopeTVar index {idx} out of bounds ({level.val})"
```
MikaelMayer, line 847: Would it be a place where we should use a throw / Except than panic/option?

joscoh: Yes, but this is a larger refactor that I thought was better for a future PR. For now, I am preserving the existing types/behavior.

StrataTest/Languages/Python/expected_laurel/test_control_flow.expected
```diff
@@ -6,13 +6,13 @@ PFloorDiv_body_calls_Int.SafeDiv_2: ✅ pass
...
 PAnd_body_calls_Any_to_bool_0: ✅ pass
 POr_body_calls_Any_to_bool_0: ✅ pass
-ret_type: ✅ pass (at byte 25063)
-ret_type: ✅ pass (at byte 25478)
-ret_pos: ✅ pass (at byte 25964)
-assert_name_is_foo: ✅ pass (at byte 26968)
-assert_opt_name_none_or_str: ✅ pass (at byte 27031)
-assert_opt_name_none_or_bar: ✅ pass (at byte 27135)
-ensures_maybe_except_none: ✅ pass (at byte 26891)
+ret_type: ✅ pass (at byte 25049)
```
MikaelMayer: Not blocking for this PR but it is the n-th PR in which this byte offset is being changed. On the next PR I see this, I will request that we redact this byte offset.

PR #530, opened 2026-03-08
refactor(CBMC): replace panic! with Except/throw in GOTO conversion paths

tautschnig:
Replace all uses of panic! in the CBMC/GOTO conversion pipeline with proper error handling via Except String or throw, making failure paths explicit and recoverable.

Source files changed:
- Common.lean: i32ToHex, bvToHex, mkConstant, mkConstantTrue, returnStmt, opToStr, opToOutTypeJson, mkBinaryOp now return Except String
- CoreToCBMC.lean: exprToJson, cmdToJson, blockToJson, stmtToJson, and symbol creation functions return Except String
- StrataToCBMC.lean (both CBMC and C_Simp): same pattern
- GOTO/InstToJson.lean: exprToJson, codeToJson, instructionToJson, programToJson, programsToJson return Except String
- GOTO/CoreToCProverGOTO.lean: Context.toJson returns Except String, getGotoJson uses throw in IO
- StrataMain.lean: updated callers to propagate Except

Test files updated:
- LambdaToCProverGOTO.lean, E2E_CoreToGOTO.lean: panic! replaced with IO.throwServerError
- ToCProverGOTO.lean: updated exprToJson call site

### DIFF
--- a/Strata/Backends/CBMC/Common.lean
+++ b/Strata/Backends/CBMC/Common.lean
@@ -236,22 +236,22 @@ def mkSymbol (identifier : String) (symbolType : Json) : Json :=
 
 /-! # Constants -/
 
-def i32ToHex (s : String) : Except String String :=
+def i32ToHex (s : String) : Except String String := do
   match s.toInt? with
   | some n =>
     let unsigned := if n < 0 then UInt32.size + n else n
-    .ok ("".intercalate ((Nat.toDigits 16 unsigned.natAbs).map (λ c => c.toUpper.toString)))
-  | none => .error "Failed to convert String to int"
+    return "".intercalate ((Nat.toDigits 16 unsigned.natAbs).map (λ c => c.toUpper.toString))
+  | none => throw "Failed to convert String to int"
 
 /-- Convert a decimal integer string to hex for a bitvector of the given bit width.
     Negative values are two's-complement encoded. -/
-def bvToHex (s : String) (width : Nat) : Except String String :=
+def bvToHex (s : String) (width : Nat) : Except String String := do
   match s.toInt? with
   | some n =>
     let modulus := 2 ^ width
     let unsigned := if n < 0 then modulus + n else n
-    .ok ("".intercalate ((Nat.toDigits 16 unsigned.natAbs).map (λ c => c.toUpper.toString)))
-  | none => .error s!"Failed to convert '{s}' to int"
+    return "".intercalate ((Nat.toDigits 16 unsigned.natAbs).map (λ c => c.toUpper.toString))
+  | none => throw s!"Failed to convert '{s}' to int"
 
 def mkConstant (value : String) (base : String) (sourceLocation : Json) (config : CBMCConfig := .empty) : Except String Json := do
   let hex ← i32ToHex value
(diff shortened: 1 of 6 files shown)


Strata/Languages/C_Simp/StrataToCBMC.lean
```diff
@@ -77,16 +77,16 @@ def lexprToCBMC (expr : Strata.C_Simp.Expression.Expr) (functionName : String) :
           ("type", mkIntType defaultConfig)
         ])
       ])
-      (mkConstant value "10" (mkSourceLocation "from_andrew.c" functionName "2" defaultConfig) defaultConfig)
-  | .const "true" _ =>
-    Json.mkObj [
+      (← mkConstant value "10" (mkSourceLocation "from_andrew.c" functionName "2" defaultConfig) defaultConfig)
```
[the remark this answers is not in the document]
tautschnig: All 18 comments flag the same non-issue. Lean 4 supports `←` in arbitrary expression positions within `do` blocks via automatic lifting (see [Lean 4 documentation on do-notation](https://lean-lang.org/lean4/doc/do.html)). The code compiles and all tests pass — these are false positives.

Strata/Backends/CBMC/Common.lean
```diff
@@ -236,43 +236,44 @@ def mkSymbol (identifier : String) (symbolType : Json) : Json :=
...
 /-! # Constants -/
 
-def i32ToHex (s : String) : String :=
+def i32ToHex (s : String) : Except String String :=
   match s.toInt? with
   | some n =>
     let unsigned := if n < 0 then UInt32.size + n else n
-    ("".intercalate ((Nat.toDigits 16 unsigned.natAbs).map (λ c => c.toUpper.toString)))
-  | none => panic! "Failed to convert String to int"
+    .ok ("".intercalate ((Nat.toDigits 16 unsigned.natAbs).map (λ c => c.toUpper.toString)))
```
MikaelMayer:
```suggestion
    return "".intercalate ((Nat.toDigits 16 unsigned.natAbs).map (λ c => c.toUpper.toString))
```
It saves one set of parentheses and avoid depending on the monad constructor. If you could do this everywhere I would prefer it because later agents will more easily pick up the pattern.

Note that I think you don't even need the do notation, it's automatically wrappped.

Strata/Backends/CBMC/Common.lean
```diff
@@ -236,43 +236,44 @@ def mkSymbol (identifier : String) (symbolType : Json) : Json :=
...
+  | none => .error "Failed to convert String to int"
```
MikaelMayer:
```suggestion
  | none => throw "Failed to convert String to int"
```

MikaelMayer (commented):
I'm so happy to see panics disappear. This is much better.
For the codd style, I would prefer to avoid one level of parentheses if possible, see  my remake below.

Strata/Backends/CBMC/Common.lean
tautschnig: Thanks, applied consistently across the changes.

Strata/Backends/CBMC/Common.lean
tautschnig: Thanks, applied consistently across the changes.

PR #532, opened 2026-03-09
Prove soundness of `LExpr.resolve`

aqjune-aws:
This pull request states and proves soundness of type inference & check algorithm of Lambda, `Expr.resolve`.

### Top-level specification

The statement is named `resolve_HasType` in LExprTypeSpec.lean, and it is defined as follows.

```
theorem resolve_HasType :
    ∀ (e : LExpr T.mono) (e_typed : LExprT T.mono) (C : LContext T)
      (Env : TEnv T.IDMeta) Env',
      e.resolve C Env = .ok ⟨e_typed, Env'⟩ →
      TEnvWF Env →
      FactoryWF C.functions →
      WellScoped e Env.context →
      Subst.allKeysFresh Env.stateSubstInfo.subst Env.context →
      LExpr.checkContextTypesClosed Env →
      HasType C (TContext.subst Env.context Env'.stateSubstInfo.subst) e (.forAll [] e_typed.toLMonoTy) ∧
      TEnvWF Env' ∧
      LExpr.checkContextTypesClosed Env' ∧
      Subst.allKeysFresh Env'.stateSubstInfo.subst Env'.context := by ...
```

The explanation is as follows.

Assumptions:
1. `e.resolve C Env = .ok ⟨e_typed, Env'⟩`: Resolving types `e` using TEnv `Env` and LContext `C` yielded `e_typed` and `Env'`
2. The input TEnv is well-formed, and the input Factory is also well-formed
3. `e` is well-scoped, meaning that `e`'s free variables are all registered to the TContext `Env.context` (sometimes Gamma)
4. All substitutions in `Env.stateSubstInfo.subst` are fresh with respect to the TContext `Env.context`, meaning that there is no 'variable -> type' mapping registered to Env.context whose types are not substituted by `Env.stateSubstInfo.subst` yet
5. `checkContextTypesClosed Env` holds, meaning that there is no unbound free type variables in the ranges of 'variable -> type' mappings

Conclusions:
1. `HasType C (TContext.subst Env.context Env'.stateSubstInfo.subst) e (.forAll [] e_typed.toLMonoTy)`: the resulting `e` is well-typed by the `HasType` rules and the substituted type context (Env' is the output TEnv)
2. The output `Env'` is well-formed
3. `checkContextTypesClosed Env'` holds
4. All substitutions in `Env'.stateSubstInfo.subst` are fresh with respect to the TContext `Env'.context`

The conclusion 2 can be fed to the next Expr.resolve call's assumption 2, conclusion 3 to assumption 5, conclusion 4 to assumption 4. This guarantees that a sequential call of `LExpr.resolve` can be validly reasoned using resolve_HasType
.

Among the 5 assumptions, the 5th assumption could be a target of discussion because we might want to allow, say ("x", .ftvar ("A")) (where "A" is a free type variable) in our TEnv when calling Expr.resolve. I found that without this 5th assumption, the theorem became ridiculously hard to prove. This assumption 5 holded empirically, from the following facts:
1. Inductively: (1) If the type environment starts from empty (or only with Factory registered), it will start from empty (or all type variables legitimately bounded). (2) After each LExpr.resolve, the conclusion will guarantee that `checkContextTypesClosed Env'` still holds. Therefore, `checkContextTypesClosed Env` can always hold, unless someone manually modifies TEnv and breaks this condition.
2. To check whether this happens in practice, at LExprT.lean I wrote `resolveChecked` which is `LExpr.resolve` with the extra check, and `lake test` successfully passed!
 
### HasType and other updates

Another updates are made to HasType, to allow
* The new `talias` rule: two types are equal modulo aliasing
* The updates to rules that handle when LExpr has type annotations: the rules needed to deal with (1) aliases, because LExpr.resolveAlias can deal with that, and (2) free variables in the annotations. For (1), the same `AliasEquiv` used in the `talias` rule was reused, For (2), it seems in this case the free variables really does not give any restriction to possible types. For example, `((x:A) == true) == ((x:A) == 0)` types even of `x` is annotated with `.some (.ftvar "A")`! For this reason, it was just handled in a relaxed way. Another possibility would be to fix LExpr.resolve so that it respects the type annotation consistently.
* As shared before, a few `panic!` in fresh variable generation were fixed to the `Except` monad.
* The fresh type variable's prefix was changed from TEnv-specific value to a constant `$__ty`, and similarly fresh variable's prefix was to constant `$__var`. In practice, they were never changed to something else.
* `LMonoTy.tconsAlias` was slightly refactored to `LMonoTy.tconsAliasSimple`. Theorem `tconsAlias_eq_simple` shows that `LMonoTy.tconsAlias` is equivalent to `LMonoTy.tconsAliasSimple`. The update can be seen from StrataTest/DL/Lambda/LExprTypeEnvTests.lean , whose interest was to check FooAlias -> Foo, but not the parameter ty vars.

Also, there are a few new smaller wellformedness:
* typeArgs_no_gen_prefix: Type arguments must not start with the reserved generated-variable prefix `$__ty` used by the type-checker.

### Quest: No aliasing remains after `Expr.resolve`

This can be independently added to the conclusion of `LExpr.resolve`, saying something like `e_typed`'s type does not include any alias. :) I wanted to avoid making the pull request too large and was not strictly related to type-safety, but if we want this can be a very reasonable next goal.

### DIFF
--- a/Strata/Languages/Laurel/Resolution.lean
+++ b/Strata/Languages/Laurel/Resolution.lean
@@ -115,20 +115,15 @@ def SemanticModel.get (model: SemanticModel) (iden: Identifier): AstNode :=
   | none => default -- panic! s!"model.get called on identifier {iden.text} without number"
 
 def SemanticModel.isFunction (model: SemanticModel) (id: Identifier): Bool :=
-  if id.uniqueId == none then
-    -- The Python pipeline generates constructor/discriminator calls that may not
-    -- be resolved at the Laurel level. Treating them as functions keeps them as
-    -- expressions; any real errors will be caught during Core type checking.
-    -- Make an exception for 'test_helper_procedure' since it's a procedure
-    -- We will remove this hack when we enable the Python through Laurel pipeline to correctly resolve
-    id.text ∉ Strata.Python.corePreludeProcedures
-  else
-    match model.get id with
-    | .staticProcedure proc => proc.isFunctional
-    | .parameter _ => true
-    | .datatypeConstructor _ _ => true
-    | .constant _ => true
-    | node => panic! s!"id: {repr id}, is not a procedure, node {repr node}"
+  match model.get id with
+      | .staticProcedure proc => proc.isFunctional
+      | .parameter _ => true
+      | .datatypeConstructor _ _ => true
+      | .constant _ => true
+      | .unresolved => true -- functions calls are more permissive, so true avoids possibly incorrect errors
+      | node =>
+          dbg_trace s!"Sound but incomplete BUG! id: {repr id}, is not a procedure, but a node {repr node}"
+          false
 
 /-- The output of the resolution pass. -/
 structure ResolutionResult where
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -1289,6 +1289,11 @@ def getDatatypeFunctions (decls: List Core.Decl) : List String :=
 
 
 def getPreludeFunctions (prelude: Core.Program) : List String := (getFunctions prelude.decls) ++ (getDatatypeFunctions prelude.decls)
+def getPreludeProcedures (prelude: Core.Program) : List String :=
+  prelude.decls.filterMap (λ decl =>
+    match decl.kind with
+        |.proc => some decl.name.name
+        | _ => none)
 
 /-- Translate Python module to Laurel Program -/
 def pythonToLaurel (prelude: Core.Program)
@@ -1374,13 +1379,19 @@ def pythonToLaurel (prelude: Core.Program)
       isFunctional := false
       }
 
+    /-
+Compute partial Laurel functions and procedures from the Core functions and procedures
+These are needed by the Laurel pipeline to determine how to translate calls.
+In the future, we will replace this Core=>Laurel translation by defining the Python prelude
+in Laurel.
+    -/
     let preludeFunctions : List Procedure := (getPreludeFunctions prelude).map (λ funcname =>
     {
-      name := {text:= funcname} ,
+      name := { text:= funcname},
       inputs := [],
       outputs := [],
       preconditions := [],
-      determinism := .deterministic none, --TODO: need to set reads
+      determinism := .deterministic none,
       decreases := none,
       body := .External
       md := default
@@ -1388,8 +1399,22 @@ def pythonToLaurel (prelude: Core.Program)
       }
     )
 
+    let preludeProcedures : List Procedure := (getPreludeProcedures prelude).map (λ funcname =>
+    {
+      name := { text:= funcname},
+      inputs := [],
+      outputs := [],
+      preconditions := [],
+      determinism := .deterministic none,
+      decreases := none,
+      body := .External
+      md := default
+      isFunctional := false
+      }
+    )
+
     let program : Laurel.Program := {
-      staticProcedures := preludeFunctions ++ procedures ++ [mainProc]
+      staticProcedures := preludeFunctions ++ preludeProcedures ++ procedures ++ [mainProc]
       staticFields := []
       types := compositeTypes.map TypeDefinition.Composite
       constants := []


Strata/DL/Lambda/LExprT.lean
```diff
@@ -344,11 +344,33 @@ def resolveAux (C: LContext T) (Env : TEnv T.IDMeta) (e : LExpr T.mono) :
...
+  -- This is required by resolveAux for the abs/quant cases, where
+  -- addInNewestContext/eraseFromContext round-trip correctly only
+  -- when there is at least one scope.
+  let Env := if Env.context.types.isEmpty then
+      Env.updateContext { Env.context with types := [[]] }
+    else Env
   let (et, Env) ← resolveAux C Env e
   .ok (LExpr.applySubstT et Env.stateSubstInfo.subst, Env)
 
+/-- Wrapper around `resolve` that checks the closed-context-types precondition
```
joscoh, line 364: So is the soundness theorem false when `checkContextTypesClosed` does not hold, or just harder to prove? If it is false, is it possible for the user to create an ill-typed expression?

Strata/DL/Lambda/LExprTypeEnv.lean
```diff
@@ -12,6 +12,8 @@ public import Strata.DL.Lambda.Factory
 public import Strata.DL.Lambda.TypeFactory
 public import Strata.DL.Util.Maps
 import all Strata.DL.Util.Maps
+public import Strata.DL.Util.String
+import all Strata.DL.Util.String
```
joscoh: Do you need both of these?

Strata/DL/Lambda/LExprTypeEnv.lean
```diff
@@ -52,6 +54,68 @@ structure TypeAlias where
...
+  fvs_closed : ∀ tv, tv ∈ LMonoTy.freeVars a.type → tv ∈ a.typeArgs
+  typeArgs_nodup : a.typeArgs.Nodup
+
+mutual
+/-- Replace free type variables in `mty` according to a positional mapping
+    from `vars` to `vals`. Variables not in `vars` are unchanged. -/
+def LMonoTy.openVars (vars : List TyIdentifier) (vals : LMonoTys) (mty : LMonoTy) : LMonoTy :=
+  match mty with
+  | .ftvar x =>
+    match List.zip vars vals |>.find? (fun p => p.1 == x) with
```
joscoh, line 67: It's not too important, but is there a reason that this does not just take in a `(List (TyIdentifier * LMonoTy))`? Then you also don't need to worry about whether the lengths align in `List.zip`.

Strata/DL/Lambda/LExprTypeEnv.lean
```diff
@@ -898,5 +1267,541 @@ def TEnv.addTypeAlias (alias : TypeAlias) (C: LContext T) (Env : TEnv T.IDMeta)
 
 ---------------------------------------------------------------------
 
+/-! ### Context preservation and freshness lemmas -/
+
+/--
+`genTyVar` preserves the context.
+
+Now that `genTyVar` returns `Except Format` (instead of using `panic`),
```
joscoh: I don't think that the comments should refer to the old `panic` version.

Strata/DL/Lambda/LExprTypeEnv.lean
```diff
@@ -898,5 +1267,541 @@ def TEnv.addTypeAlias (alias : TypeAlias) (C: LContext T) (Env : TEnv T.IDMeta)
...
+theorem TGenEnv.genTyVar_context {IDMeta : Type} [ToFormat IDMeta]
+    (Env : TGenEnv IDMeta) (tv : TyIdentifier) (Env' : TGenEnv IDMeta)
+    (h : TGenEnv.genTyVar Env = .ok (tv, Env')) :
+    Env'.context = Env.context := by
+  simp [TGenEnv.genTyVar] at h
+  split at h
+  · simp at h
+  · simp at h; obtain ⟨_, h2⟩ := h; rw [← h2]
+
+/-- `genTyVars` preserves the context (by induction, using `genTyVar_context`). -/
```
joscoh: I don't think that the `by induction, using genTyVar_context` adds anything, but that is a matter of preference.

Strata/DL/Util/Map.lean
```diff
@@ -213,5 +232,102 @@ theorem Map.findNone_eq_notmem_mapfst {m: Map α β} [DecidableEq α]: ¬ a ∈
...
+    · right; simp [h]
+  | cons p m' ih =>
+    obtain ⟨a, b⟩ := p
+    unfold Map at *; simp only [List.cons_append, Map.find?]
+    by_cases h : a = y
+    · right; simp [h]
+    · simp only [h, ↓reduceIte]; exact ih
+
+/-- When `x` is not in the map, `Map.insert` appends `(x, v)` at the end. -/
+theorem Map.insert_fresh_eq_append [DecidableEq α]
```
joscoh, line 287: These lemmas concern me a bit. Ideally the List-based representation of the Map is just an implementation detail and we should not need to rely on the internal structure so much.

Strata/DL/Util/Maps.lean
```diff
@@ -341,5 +359,222 @@ theorem Maps.mem_values_of_mem_keys_remove [DecidableEq α] [BEq (Map α β)]
       · simp [@Map.mem_values_of_mem_keys_remove _ _ _ m k v (by assumption)]
       · simp_all
 
+/-- `Map.find?` returns `none` when the key is not in `Map.keys`. -/
```
joscoh: Should this be in Map.lean to be consistent? I'm fine either way as long as things are consistent.

Strata/DL/Util/Maps.lean
```diff
@@ -341,5 +359,222 @@ theorem Maps.mem_values_of_mem_keys_remove [DecidableEq α] [BEq (Map α β)]
...
+    (ms : Maps α β) (x : α) (v : β) :
+    Maps.find? (Maps.insert ms x v) x = some v := by
+  simp only [Maps.insert]; split
+  · match ms with
+    | [] => simp [Maps.pop, Maps.push, Maps.newest, Maps.find?, Map.find?_insert_self]
+    | _ :: _ => simp [Maps.pop, Maps.push, Maps.newest, Maps.find?, Map.find?_insert_self]
+  · exact Maps.find?_update_self ms x v (by simp_all)
+
+/-- `Maps.find?` is unchanged for a different key after `Maps.insert`. -/
+theorem Maps.find?_insert_ne [DecidableEq α]
```
joscoh, line 408: Unless I am missing something, this seems strictly stronger than `Maps.find?_insert_ne_of_none`.

Strata/DL/Util/Maps.lean
```diff
@@ -341,5 +359,222 @@ theorem Maps.mem_values_of_mem_keys_remove [DecidableEq α] [BEq (Map α β)]
...
+    (ms : Maps α β) (x : α) (h : ∀ m, m ∈ ms → Map.find? m x = none) :
+    Maps.erase ms x = ms := by
+  induction ms with
+  | nil => simp [Maps.erase]
+  | cons m rest ih =>
+    simp only [Maps.erase]; congr 1
+    · exact Map.erase_of_find?_none m x (h m List.mem_cons_self)
+    · exact ih (fun r hr => h r (List.mem_cons_of_mem m hr))
+
+/-- Erasing a key that was just added to the newest scope restores the original types,
```
joscoh: Minor, but should refer to keys/values, not types.

joscoh:
That's fine if you prefer the lemmas in there for module purposes. I would have just called it `LExprTypeEnvProofs.lean` or something like that (unless there is a more specific name). In the long run, if we eventually have other proofs about the type system (e.g. progress and preservation, etc), we will likely want to split proofs between multiple files anyway. But that's not blocking for this PR.

Strata/DL/Lambda/LTyUnify.lean
```diff
@@ -540,6 +573,136 @@ def LTy.subst (S : Subst) (ty : LTy) : LTy :=
...
+/--
+Substitution on `LMonoTy.bool` is the identity (ground type).
+-/
+theorem LMonoTy.subst_bool (S : Subst) : LMonoTy.subst S LMonoTy.bool = LMonoTy.bool := by
+  simp [LMonoTy.bool, LMonoTy.subst]
+  intro h
+  simp [LMonoTys.subst, h, LMonoTys.subst.substAux]
+
+/-- Substitution distributes over a 2-element `tcons`, giving component-wise results. -/
+theorem LMonoTy.subst_tcons_pair (S : Subst) (name : String) (a b : LMonoTy) :
```
joscoh, line 611: Just curious, is there a reason this is only proved for pairs? I imagine the general lemma (`subst S (.tcons name tys) = List.map (subst S) tys`) is no harder to prove.

Strata/DL/Lambda/LTyUnify.lean
```diff
@@ -540,6 +573,136 @@ def LTy.subst (S : Subst) (ty : LTy) : LTy :=
...
+  intro h_x_key
+  exact h_wf x h_x_key h_x_in_fvs
+
+/--
+If no key of a substitution `S` appears free in `ty`, then applying `S` to
+`ty` leaves it unchanged. This is the key lemma for proving idempotence.
+-/
+theorem LMonoTy.subst_no_key_free (S : Subst) (ty : LMonoTy)
+    (h : S.keys.all (fun k => k ∉ ty.freeVars)) :
+    LMonoTy.subst S ty = ty := by
```
joscoh, line 714: The following proof also works:
```
apply subst_no_relevant_keys
grind
```

Strata/DL/Lambda/LTyUnify.lean
```diff
@@ -1183,6 +1346,827 @@ def Constraints.unify (constraints : Constraints) (S : SubstInfo) :
...
+      rw [h_inner]
+  | bitvec n =>
+    simp [LMonoTy.subst]
+  | tcons name args ih =>
+    simp only [LMonoTy.subst, hSingle, hS, Bool.false_eq_true, ↓reduceIte]
+    congr 1
+    exact LMonoTys.subst_absorbs_single S a t args hS hSingle ih
+
+/-!
+#### Absorption: relating substitutions produced by successive resolveAux calls
```
joscoh: This comment should be placed above the prior few lemmas, which already reference absorption.

Strata/DL/Lambda/LTyUnify.lean
```diff
@@ -1183,6 +1346,827 @@ def Constraints.unify (constraints : Constraints) (S : SubstInfo) :
...
+This is a key lemma for proving that sequential `tinst` applications
+(each substituting one bound variable) produce the same result as a
+single parallel substitution with all bindings.
+-/
+
+theorem LMonoTy.subst_cons_single
+    (v : TyIdentifier) (t : LMonoTy) (ys : SubstOne) (mty : LMonoTy)
+    (h_t : LMonoTy.subst [ys] t = t) :
+    LMonoTy.subst [ys] (LMonoTy.subst [[(v, t)]] mty) =
+    LMonoTy.subst [((v, t) :: ys)] mty := by
```
joscoh, line 1558: Not blocking, but I wonder if some of these lemmas can be proved with a generic `LMonoTy.subst` composition result and an "extensionality" result like: `(forall x, x \in ty.freeVars -> Maps.find? S1 x = Maps.find? S2 x) -> LMonoTy.subst S1 ty = LMonoTy.subst S2 ty`? In other words, just show pointwise that these substitutions are equal. Maybe not, but it would make a lot of these tedious proofs much simpler if so.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -19,7 +19,10 @@ import all Strata.DL.Lambda.Identifiers
 import all Strata.DL.Util.Func
 import all Strata.DL.Util.ListMap
 import all Strata.DL.Util.List
-
+public import Strata.DL.Lambda.LExprT
+import all Strata.DL.Lambda.LExprT
+public import Strata.DL.Lambda.FactoryWF
+import all Strata.DL.Lambda.FactoryWF
```
joscoh: Are these redundant?

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+
+/--
+Helper: `toLMonoTy` commutes with `applySubstT` in the expected way.
+For most constructors, `(applySubstT et S).toLMonoTy = LMonoTy.subst S et.toLMonoTy`.
+For quantifiers, `toLMonoTy` always returns `LMonoTy.bool`.
+-/
+theorem applySubstT_toLMonoTy {T : LExprParamsT}
+    (et : LExprT T) (S : Subst) :
+    (LExpr.applySubstT et S).toLMonoTy = LMonoTy.subst S et.toLMonoTy := by
+  cases et with
```
joscoh: Entirely a matter of preference, but this can be:
```
  cases et <;>  simp [LExpr.applySubstT, LExpr.replaceMetadata, LExpr.toLMonoTy]
  case quant m k _ ty tr e =>
    unfold LMonoTy.subst
    split <;> simp [LMonoTys.subst, LMonoTys.subst.substAux]; rfl
```

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+
+1. **`resolveAux_HasType`**: The core theorem, proved by induction on `e`.
+   States that if `resolveAux C Env e = .ok (et, Env')`, then:
+   - `Env'.context = Env.context` (context is preserved), and
+   - for any substitution `S` that absorbs `Env'.stateSubstInfo.subst`,
+     `HasType C (TContext.subst Env.context S) e (.forAll [] (LMonoTy.subst S et.toLMonoTy))`.
+
+   The conclusion is universally quantified over `S` so that the caller
+   (including recursive cases) can instantiate it with the final substitution.
+   This eliminates the need for `HasType_subst_upgrade` in recursive cases
```
joscoh: I'm not sure what this sentence means.

Edit: I figured it out by reading comments much later in the file (I believe that `upgrade` is basically a weakening lemma, and by quantifying over the substitution, you don't need to explicitly weaken, is that correct?), but it doesn't have much context here.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+   `resolveAux_properties.absorbs`, `unify_absorbs`, and `Subst.absorbs_trans`.
+
+2. **`resolve_HasType`**: The top-level theorem. Since `resolve` is just
+   `resolveAux` followed by `applySubstT`, we decompose the hypothesis,
+   apply `resolveAux_HasType` (instantiating `S` with the final substitution),
+   and use `applySubstT_toLMonoTy`.
+
+#### Key supporting lemmas (quite a few of these are in LTyUnify.lean):
+
+- **`Subst.absorbs`**: `S_outer` absorbs `S_inner` when every binding in
```
joscoh, line 329: Minor, but this is a definition, not a lemma, right?

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+  simp only [Constraints.unify, Bind.bind, Except.bind] at h_unify
+  split at h_unify
+  · simp at h_unify
+  · rename_i relS h_core
+    simp at h_unify; subst h_unify
+    exact (Constraints.unifyCore_sound cs S relS h_core).keys_incl
+
+/-- `Map.values` of a `zipWith Prod.mk` is the second list truncated to the
+    length of the first. -/
+private theorem Map.values_zipWith_eq_take (as : List TyIdentifier) (bs : List LMonoTy) :
```
joscoh: Probably this should be polymorphic and in `Map.lean`

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+        simp at h
+        obtain ⟨_, _, h_env⟩ := h; rw [← h_env]
+        simp only [TEnv.addInNewestContext, TEnv.updateContext]
+        -- genTyVar preserves stateSubstInfo
+        have h_subst := TEnv.genTyVar_subst _ xtyid Env1 h_genTy
+        rw [h_subst, h_gen_subst]
+        exact Subst.absorbs_refl _ Env.stateSubstInfo.isWF
+
+/-- Removing a key `k` from a map doesn't affect lookups of other keys `a ≠ k`. -/
+private theorem Map.find?_remove_ne {α β : Type} [DecidableEq α]
```
joscoh: Also should be in `Map.lean`

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    by_cases h_xk : x = k
+    · -- x = k: Map.remove skips this entry; result is `rest`
+      simp only [h_xk, ↓reduceIte]
+      simp only [Map.find?, show k ≠ a from Ne.symm h_ne, ↓reduceIte]
+    · -- x ≠ k: entry preserved
+      simp only [h_xk, ↓reduceIte, Map.find?]
+      grind
+
+/-- Removing a key `k` from maps doesn't affect lookups of other keys `a ≠ k`. -/
+private theorem Maps.find?_remove_ne
```
joscoh: Does `ms` have to be a `Subst`? If not, should be in `Maps.lean`

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+      | cons _ _ => simp [Map.isEmpty] at h
+    · apply ih
+      -- Need: hasEmptyScopes rest
+      simp [Subst.hasEmptyScopes]
+      exact h.2
+
+/-- `subst (remove S k) mty = subst S mty` when `k ∉ freeVars mty`.
+    Since `LMonoTy.subst` is single-pass, removing a key that doesn't
+    appear in the type doesn't change the result. -/
+private theorem LMonoTy.subst_remove_not_fv (S : Subst) (k : TyIdentifier) (mty : LMonoTy)
```
joscoh, line 1517: Also would be much easier to prove with an "extensionality" lemma for `LMonoTy.subst` (e.g. if `(forall x, x \in LMonoTy.freeVars t \to Maps.find S1 x = Maps.find S2 x)`, then `LMonoTy.subst S1 t = MonoTy.subst S2 t`

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+              (LMonoTy_instantiateWithCheck_tyGen_mono fty_val C Env1 fty_inst Env2 h_iwc2)
+
+omit [ToString T.IDMeta] [ToFormat (LFunc T)] [ToFormat T.Metadata] [DecidableEq T.IDMeta] in
+private theorem typeBoundVar_tyGen_mono
+    (C : LContext T) (Env : TEnv T.IDMeta) (bty : Option LMonoTy)
+    (xv : T.Identifier) (xty : LMonoTy) (Env' : TEnv T.IDMeta)
+    (h : typeBoundVar C Env bty = .ok (xv, xty, Env')) :
+    Env'.genEnv.genState.tyGen ≥ Env.genEnv.genState.tyGen := by
+  simp only [typeBoundVar, liftGenEnv, Bind.bind, Except.bind] at h
+  -- Split on the result of HasGen.genVar (now returns Except)
```
joscoh: Should not say `now`

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+        -- In either case, v is either a known type var (gen-fresh) or was substituted away
+        -- For the overapprox case: v ∈ freeVars body and v ∈ (x::xs) means v is a bound var.
+        -- Context's knownTypeVars include bound vars, so v is gen-fresh by ContextFreshForGen.
+        -- Use h_ty_fresh which covers LTy.freeVars = freeVars(body) \ bound vars
+        -- But freeVars_of_subst_subset includes ALL body fvs, not just unbound ones.
+        -- For bound vars: they are in context's knownTypeVars, hence gen-fresh by h_cfg.
+        -- Since h_cfg : ContextFreshForGen Env.context Env.genState, and
+        -- n ≥ genEnv'.genState.tyGen ≥ Env.genState.tyGen:
+        -- v ∈ freeVars body where ty = forAll (x::xs) body
+        -- Split: v ∈ (x::xs) (bound var) or v ∉ (x::xs) (free var)
```
joscoh: I don't have a strong opinion either way, but do we need all of these comments inline? It seems inconsistent which lemmas have lots of comments and which do not.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    suffices ∀ Env1, typeBoundVar C Env bty = .ok (xv, xty, Env1) →
+        Env1.context.types = Env.context.types.addInNewest [(xv, .forAll [] xty)] ∧
+        Env1.context.aliases = Env.context.aliases by
+      intro h_ctx h_nonempty h_fresh_xv; exact this Env1 h
+    -- Decompose typeBoundVar to extract that Env1.context = addInNewest on Env.context
+    -- typeBoundVar does: liftGenEnv genVar >> (instantiateWithCheck|genTyVar) >> addInNewestContext
+    -- Each intermediate step preserves context, then addInNewestContext modifies types
+    intro Env1 h_tbv
+    -- typeBoundVar C Env bty =
+    --   do let (xv, Env_g) ← liftGenEnv genVar Env
```
joscoh: Is this dead code?

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    --      let (xty, Env_mid) ← (instantiateWithCheck | genTyVar)
+    --      return (xv, xty, Env_mid.addInNewestContext [(xv, .forAll [] xty)])
+    -- We need: Env1.context.types = Env.context.types.addInNewest [...]
```
joscoh: Comments with `[...]` should probably be removed.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    | cons m rest =>
+      exact Maps.erase_addInNewest_fresh xv _ (fun s hs => h_fresh_xv s (h_types_ne ▸ hs))
+  have h1 : (Env2.eraseFromContext xv).context.types = Env.context.types := by
+    rw [h_erase_types, h_cancel]
+  have h2 : (Env2.eraseFromContext xv).context.aliases = Env.context.aliases := by
+    rw [h_erase_aliases, h_al]
+  exact TContext.mk.injEq .. ▸ ⟨h1, h2⟩
+
+/-- `removeAll` produces `[]` when every element of the source is in the filter list. -/
+private theorem List.removeAll_eq_nil_of_forall_mem [BEq α] [LawfulBEq α]
```
joscoh: Maybe `List.lean`?

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    rw [ih m]; simp
+
+/-- `LMonoTys.freeVars (xs ++ ys) = freeVars xs ++ freeVars ys`. -/
+private theorem LMonoTys.freeVars_append (xs ys : LMonoTys) :
+    LMonoTys.freeVars (xs ++ ys) = LMonoTys.freeVars xs ++ LMonoTys.freeVars ys := by
+  induction xs with
+  | nil => simp [LMonoTys.freeVars]
+  | cons x xrest ih => simp [LMonoTys.freeVars, ih, List.append_assoc]
+
+-- Combined proof by well-founded induction on sizeOf, proving both single-type
```
joscoh: This comment does not apply here, it is a definition, not a proof.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    rcases hv with hv_hd | hv_tl
+    · left; exact LMonoTy_resolveAliases_freeVars_subset mty Env mty' Env1 h_hd h_aw v hv_hd
+    · right; exact LMonoTys_resolveAliases_freeVars_subset mrest Env1 mrest' Env2 h_tl
+        (h_ctx_eq ▸ h_aw) v hv_tl
+end
+
+
+omit [ToString T.IDMeta] [ToFormat T.IDMeta] [HasGen T.IDMeta] [ToFormat (LFunc T)] [ToFormat T.Metadata] in
+/-- Combined result: context preservation, SubstFreshForGen preservation, and output type freshness.
+    These are proved together by strong induction to avoid circular dependencies. -/
```
joscoh: I'm not sure what the strong induction comment is referring to here

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    intro v hv; simp [LMonoTys.freeVars, List.mem_append] at hv ⊢
+    rcases hv with hv_hd | hv_tl
+    · left; exact LMonoTy_resolveAliases_freeVars_subset mty Env mty' Env1 h_hd h_aw v hv_hd
+    · right; exact LMonoTys_resolveAliases_freeVars_subset mrest Env1 mrest' Env2 h_tl
+        (h_ctx_eq ▸ h_aw) v hv_tl
+end
+
+
+omit [ToString T.IDMeta] [ToFormat T.IDMeta] [HasGen T.IDMeta] [ToFormat (LFunc T)] [ToFormat T.Metadata] in
+/-- Combined result: context preservation, SubstFreshForGen preservation, and output type freshness.
```
joscoh: This comment seems to be referring to something else, maybe `resolveAux_properties_aux`?

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+  · intro a t h_find h_fv
+    have h_in_fvs := Subst.freeVars_of_find_subset Env.stateSubstInfo.subst h_find h_fv
+    exact h_fresh fresh_name (Or.inr h_in_fvs) Env2.genEnv.genState.tyGen h_mono h_name
+
+
+omit [ToString T.IDMeta] [ToFormat (LFunc T)] [ToFormat T.Metadata] in
+/-- Combined inductive proof of all `ResolveAuxProperties` for `resolveAux`:
+    generator monotonicity, context preservation, SubstFreshForGen preservation,
+    output type freshness, and absorption.
+    Uses strong induction on `e.sizeOf` to handle `varOpen` in abs/quant cases. -/
```
joscoh: Nit: strong induction is needed for the `.app`, `.eq`, `.ite`cases, not `abs`/`quant`, which have size 1 smaller than the original term.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+  let ⟨h1, h2, h3, h4⟩ := resolveAux_properties_aux e.sizeOf e rfl et C Env Env' h h_ne h_aw h_fwf h_sf h_cf h_bvf
+  { genState_mono := h1, context := h2, preserves := h3, absorbs := h4 }
+
+omit [ToString T.IDMeta] [ToFormat T.IDMeta] [HasGen T.IDMeta] [ToFormat (LFunc T)] [ToFormat T.Metadata] in
+/--
+Upgrade lemma: if `e` has type `subst S_inner mty` and `S_outer` absorbs
+`S_inner`, then `e` has type `subst S_outer mty` (provided `S_outer`'s keys
+are fresh in the context).
+
+This is the key mechanism for composing IHs in the new formulation: each
```
joscoh: What is the "new formulation"?

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+  rw [LMonoTy.subst_absorbs S_outer S_inner mty h_absorbs] at h1
+  exact h1
+
+
+private theorem removeAll_not_mem {x : TyIdentifier} {xs : List TyIdentifier}
+    (h : x ∉ xs) : xs.removeAll [x] = xs := by
+  induction xs <;> grind
+
+/-- Keys of a zipped map are a subset of the first list. -/
+private theorem Map.keys_zip_subset {α β : Type} [DecidableEq α]
```
joscoh: `Map.lean`?

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+      simp [LMonoTys.openVars, LMonoTys.substLogic, hS_ne]
+      constructor
+      · exact subst_openVars_comm S vars vals hd (by
+          intro tv h; exact h_wf tv (by simp [LMonoTys.freeVars]; left; exact h)) h_len
+      · exact subst_openVarsList_comm S vars vals tl (by
+          intro tv h; exact h_wf tv (by simp [LMonoTys.freeVars]; right; exact h)) h_len
+end
+
+/-- `Map.find?` on a zip agrees with `List.find?` using BEq on the first component. -/
+private theorem map_find_eq_list_find' (vars : List TyIdentifier) (vals : LMonoTys) (x : TyIdentifier) :
```
joscoh: `Map.lean`? This doesn't seem specific to `TyIdentifier` or `LMonoTys`.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    | nil => simp [LMonoTys.substLogic]
+    | cons fv fvs' ih =>
+      unfold LMonoTys.substLogic; split
+      · rename_i hS; simp [LMonoTy.subst_emptyS hS]
+      · simp only [List.map]; exact congrArg _ ih
+  rw [h_substLogic_map]
+  -- Step 4: openVars → subst (reverse direction)
+  rw [← subst_single_scope_eq_openVars ids _ mty h_closed h_vals_len']
+
+private theorem Map.keys_erase_subset [DecidableEq α] (m : Map α β) (x : α) :
```
joscoh: `Map.lean` (and the next few)?

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+                | tail _ h_rest => exact List.mem_append_right _ (ih_l h_rest)
+            exact this args ha hv)
+        -- Reduce: subst distributes over tcons when non-empty
+        -- Use h_per_arg to show each arg matches, then combine for the whole tcons.
+        -- Avoid simp on the subst structure; instead use the per-arg proof directly.
+        -- Show: subst S (subst [zip1] (tcons name args)) = subst [zip2] (tcons name args)
+        -- This is exactly what the goal is after the `induction mty` + `tcons` case.
+        -- The approach: rewrite both sides via LMonoTy.subst_tcons or similar,
+        -- then reduce to per-arg.
+        -- Actually, let's just note that subst_compose_ftvar_closed' on each arg gives us
```
joscoh: I don't think we should have the LLM thought process in here

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+    simp only [List.map, LMonoTys.openVars, LMonoTy.openVars,
+               List.zip, List.zipWith, List.find?, BEq.beq]
+    simp only [Ne.symm h_w_ne]
+    congr 1
+    exact ih (fun h => h_v_notin (.tail _ h))
+
+private theorem openVarsList_map_ftvar_id
+    (vars : List TyIdentifier) (vals : LMonoTys)
+    (h_len : vars.length = vals.length)
+    (h_nodup : vars.Nodup := by assumption) :
```
joscoh: Why does a theorem argument have a default value? I've never seen this before, maybe I am missing something but it seems strange.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+      have := LMonoTys.instantiateEnv_length _ _ _ _ _ h_inst; simp at this; omega
+    -- Decompose: unify
+    generalize h_u : Constraints.unify _ _ = u at h_tcons
+    match u with
+    | .error e => simp at h_tcons
+    | .ok substInfo =>
+      simp at h_tcons
+      obtain ⟨h_mty, h_env⟩ := h_tcons
+      rw [← h_mty, ← h_env]
+      -- Now goal uses getD instead of dependent indexing:
```
joscoh: I think this "Now" refers to an older version, so I think it should be removed.

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+end
+
+/-- `subst S (ftvar v) = t` when `S` is non-empty and `find? S v = some t`. -/
+private theorem LMonoTy.subst_ftvar_eq (S : Subst) (v : TyIdentifier) (t : LMonoTy)
+    (h_ne : Subst.hasEmptyScopes S = false) (h_find : Maps.find? S v = some t) :
+    LMonoTy.subst S (.ftvar v) = t := by
+  simp only [LMonoTy.subst, h_ne, h_find, Bool.false_eq_true, ↓reduceIte]
+
+/-- Helper: `Map.find?` on `l.map (fun v => (v, f v))` returns `some (f v)` for `v ∈ l`. -/
+private theorem Map.find?_of_map_self {α : Type} [DecidableEq α] {β : Type}
```
joscoh: `Map.lean`

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+      intro l h_all
+      induction l with
+      | nil => simp [LMonoTys.substLogic, LMonoTys.freeVars]
+      | cons hd tl ih_tl =>
+        simp only [LMonoTys.substLogic, hSF]
+        intro h_abs; rcases List.mem_append.mp h_abs with h_hd | h_tl
+        · exact h_all hd (List.mem_cons_self ..) h_hd
+        · exact ih_tl (fun m hm => h_all m (List.mem_cons_of_mem _ hm)) h_tl
+
+private theorem Map.values_erase_subset [DecidableEq α] (m : Map α β) (x : α) :
```
joscoh: `Map.lean` (and the following)

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+
+private theorem SubstWF_go (S : Subst) (xs : List TyIdentifier) (h_wf : SubstWF S) :
+    SubstWF (LTy.subst.go xs S) := by
+  induction xs generalizing S with
+  | nil => simp [LTy.subst.go]; exact h_wf
+  | cons x rest ih =>
+    simp [LTy.subst.go]
+    exact ih (Maps.erase S x) (SubstWF_erase S x h_wf)
+
+private theorem Map.keys_erase_mem_of_ne [DecidableEq α] (m : Map α β) {a x : α}
```
joscoh: `Map.lean`

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -234,6 +255,7148 @@ theorem HasType.regularity [DecidableEq T.IDMeta] (h : HasType (T := T) C Γ e t
...
+  | cons scope rest ih =>
+    simp only [Maps.erase, Maps.keys] at h_key ⊢
+    rcases List.mem_append.mp h_key with h | h
+    · exact List.mem_append_left _ (Map.keys_erase_mem_of_ne scope h h_ne)
+    · exact List.mem_append_right _ (ih h)
+
+private theorem keys_go_mem (S : Subst) (xs : List TyIdentifier) (a : TyIdentifier)
+    (h_key : a ∈ Maps.keys S) (h_not_xs : a ∉ xs) :
+    a ∈ Maps.keys (LTy.subst.go xs S) := by
+  induction xs generalizin
```
joscoh: What is this docstring for? It doesn't have a definition immediately following.

joscoh: `Map.lean`

joscoh: I don't quite understand what this means. Is this a placeholder for the full goal conclusion?

joscoh: Don't think LLM options should be included in output. The comment is fine, just probably shouldn't have `Option A:`

Strata/DL/Lambda/LExprTypeSpec.lean
aqjune-aws: Removed because the comment over `resulveAux_HasType` already had a similar explanation

Strata/DL/Lambda/LExprTypeSpec.lean
aqjune-aws: This is a legacy of Kiro who enjoyed writing proof plans. Let me omit lengthy comments...

Strata/DL/Lambda/LExprTypeSpec.lean
aqjune-aws: Added LMonoTy.subst_ext.

Strata/DL/Lambda/LExprTypeSpec.lean
aqjune-aws: It is a remnant of a previous decomposition the goal. HasType_subst_upgrade was not used in the last proof anymore; I removed it.

Strata/DL/Lambda/LExprTypeSpec.lean
aqjune-aws: I think this must have been attached to a theorem about the output of `resolveAux`, specifically wellformedness of the typed expression. It existed as a separate theorem. Now removed.

Strata/DL/Lambda/LTyUnify.lean
aqjune-aws: It seems this lemma was used for `rw`. :)

Strata/DL/Lambda/LExprT.lean
aqjune-aws: Not immediately, but a chain of `LExpr.resolve` can cause an ill-typed expression.
Let's assume that `TContext.types` has a typing information ("x", `\alpha`) (where `\alpha` is a type variable). For some reason, type unification may unify `\alpha` with `int`, adding `\alpha |-> int` to the type substitution (TEnv.stateSubstInfo).

This will break `Subst.allKeysFresh Env'.stateSubstInfo.subst Env'.context` in the postcondition of `resolve_HasType`. And I have a proof that if`Subst.allKeysFresh ...` doesn't hold, the result of `LExpr.resolve` can break `\neg HasType`. :) https://gist.github.com/aqjune-aws/47713c51b5d43e121eb1651ba44c11fc

aqjune-aws: Since I could not find a counter example that immediately breaks `HasType`, there might be a better definition of well-formedness that relaxes `checkContextTypesClosed`. I could not find one (yet) :/

Strata/DL/Lambda/FactoryWF.lean
```diff
@@ -32,13 +31,18 @@ public section
...
+/-- Well-formedness properties for LFunc — extends generic `FuncWF` with
+    Lambda-specific extractors and the generated-prefix guard on `typeArgs`. -/
+structure LFuncWF {T : LExprParams} (f : LFunc T) extends
+    FuncWF
+      (fun id => id.name) -- getName
+      (fun e => (LExpr.freeVars e).map (·.1.name)) -- getVarNames
+      (fun e => e.freeVars) -- getTyFreeVars
+      f where
+  /-- Type arguments must not start with the reserved generated-variable
+      prefix `$__ty` used by the type-checker. -/
```
shigoel, line 43: Nice!

Strata/DL/Lambda/LExprT.lean
```diff
@@ -230,7 +230,7 @@ def resolveAux (C: LContext T) (Env : TEnv T.IDMeta) (e : LExpr T.mono) :
     /- Infer the type of an operation `.op o oty`, where an operation is defined in
       the factory. -/
     let (ty, Env) ← do
-      match C.functions.find? (fun fn => fn.name.name == o.name) with
+      match C.functions.find? (fun fn => fn.name == o) with
```
shigoel, line 233: Wondering why the `.name` was removed. Won't this cause this equality check to be sensitive to the identifier's metadata field too?

aqjune-aws: This was to make the comparison of identifiers consistent for proofs - some places were comparing `name.name`, some places were using `name`. Before the metadata type parameter is fully removed, this seemed to be a safer choice (at the expense of bringing BEq) :)

shigoel: Is it on the list to remove the metadata? I recall @andrewmwells-amazon ran into issues with the Python benchmarks because of a missing `.name` here.

aqjune-aws: Yes - after https://github.com/strata-org/Strata/pull/488 metadata (which was `Core.Visibility`) is simply Unit. https://github.com/strata-org/Strata/pull/488#discussion_r2866263346 has a short thread that talks about removing the type eventually.

shigoel: Thank you! I am realizing that a lot happened in #488 (which I didn't review) and so I have some missing context. Resolving...

aqjune-aws: Yes, many things are happening in a short period...

Strata/DL/Lambda/LExprTypeEnv.lean
```diff
@@ -52,6 +52,68 @@ structure TypeAlias where
...
+def TypeAlias.expandsTo (aliases : List TypeAlias) (name : String) (args : LMonoTys) (mty : LMonoTy) : Prop :=
+  ∃ alias ∈ aliases,
+    alias.name = name ∧
+    alias.typeArgs.length = args.length ∧
+    mty = alias.expand args
+
+mutual
+/-- Two monotypes are alias-equivalent if one can be obtained from the other
+    by expanding (or collapsing) type aliases, including inside subtrees. -/
+inductive AliasEquiv (aliases : List TypeAlias) : LMonoTy → LMonoTy → Prop where
```
shigoel, line 94: Very cool!

Strata/DL/Lambda/LExprTypeEnv.lean
```diff
@@ -121,34 +188,160 @@ def TContext.allFreshVars (xs : List TyIdentifier) (Γ : (TContext T.IDMeta)) :
   | [] => True
   | x :: rest => (TContext.isFresh x Γ) ∧ (TContext.allFreshVars rest Γ)
 
+def TContext.types.subst.go (S : Subst) (tmap : Map (Identifier IDMeta) LTy) :
```
shigoel, line 191: Just curious: did this kind of refactoring help in proofs? If so, maybe a guidance to folks contributing to Strata is to not use `where` with implicit formals anymore

aqjune-aws: I need more experiment with this (and since AI explored this as well... I have limited understanding), but indeed my impression was that especially with the new module system, explicitly extracting out `go` function like this was more helpful in unfolding this `go` definition. When I observe this repetitively in the next proofs, I will write some suggestions to people.

Strata/DL/Lambda/LTyUnify.lean
```diff
@@ -540,6 +573,156 @@ def LTy.subst (S : Subst) (ty : LTy) : LTy :=
   match xs with
   | [] => S | x :: rest => go rest (S.erase x)
 
+/--
+Open `ty` by instantiating the bound type variable `x` with `xty`.
+-/
+def LTy.open (x : TyIdentifier) (xty : LMonoTy) (ty : LTy) : LTy :=
```
shigoel, line 581: Nit: LTy.close is in Strata/DL/Lambda/LTy.lean. These two functions look like they belong together.

shigoel: Also: openFull below.

Strata/DL/Lambda/LExprT.lean
```diff
@@ -344,11 +344,33 @@ def resolveAux (C: LContext T) (Env : TEnv T.IDMeta) (e : LExpr T.mono) :
...
+  let Env := if Env.context.types.isEmpty then
+      Env.updateContext { Env.context with types := [[]] }
+    else Env
   let (et, Env) ← resolveAux C Env e
   .ok (LExpr.applySubstT et Env.stateSubstInfo.subst, Env)
 
+/-- Wrapper around `resolve` that checks the closed-context-types precondition
+    required by the soundness theorem `resolve_HasType`. All context types must
+    be closed (no free type variables). Use this when the soundness guarantee is
+    needed; call plain `resolve` when the context may have non-ground annotations
```
shigoel: Can you give other examples of the kinds of context where it's suitable to call `resolve`? Top-level global vars?

Strata/DL/Lambda/LTyUnify.lean
aqjune-aws: It seems this might need some amount of work because it uses `LMonoTy.subst` and we have theorems about it - but I can move forward and relocate these as well.

shigoel: I don't want to block on this though if it's a lot of work; perhaps a note for later (unless you're really okay to do this in this PR)

Strata/DL/Lambda/LTyUnify.lean
```diff
@@ -540,6 +573,156 @@ def LTy.subst (S : Subst) (ty : LTy) : LTy :=
...
+  LMonoTy.subst [(List.zip (LTy.boundVars ty) tys)] (LTy.toMonoTypeUnsafe ty)
+
+---------------------------------------------------------------------
+
+/-! ### Substitution Properties -/
+
+/--
+Substitution on `LMonoTy.bool` is the identity (ground type).
+-/
+theorem LMonoTy.subst_bool (S : Subst) : LMonoTy.subst S LMonoTy.bool = LMonoTy.bool := by
```
shigoel, line 605: ooh interesting. Looks like this could be generalized to any `.tcons <name> []`, right?

Strata/DL/Lambda/LExprT.lean
aqjune-aws: Oh dang, I think this comment is wrong because is fine to call `resolve` in practically any case in Core (unless we implement type inference in Imperative in the future). Generic procedure parameters in Strata Core is fine. I will remove this comment.

Strata/DL/Lambda/LTyUnify.lean
aqjune-aws: Yes, I think so. :) This kind of theorems seem to be used for rewrite rules (which want to preserve RHS exactly as 'LMonoTy.bool').

Strata/DL/Lambda/LTyUnify.lean
shigoel, line 1: So very cool to see theorems in this file that were just invariants in my head before! <3

Strata/DL/Lambda/LTyUnify.lean
```diff
@@ -1183,6 +1368,819 @@ def Constraints.unify (constraints : Constraints) (S : SubstInfo) :
...
+    Map.find? (SubstOne.applyLogic new old) a = some (LMonoTy.subst [new] t) := by
+  induction old with
+  | nil => simp [Map.find?] at h
+  | cons hd rest ih =>
+    simp only [SubstOne.applyLogic, Map.find?]
+    simp only [Map.find?] at h
+    grind
+
+-- Helper: applyLogic preserves none bindings.
+private theorem Map.find?_applyLogic_none_h' {new old : SubstOne} {a : TyIdentifier}
```
shigoel, line 1623: If we ever change our implementation of Maps, we'd have to make sure we preserve the interface. Just bringing this up because Maps can be inefficient... No action item for this PR.

shigoel (approved):

Strata/DL/Lambda/LTyUnify.lean
aqjune-aws: Indeed. I am pretty sure that AI will be good at proof repair. AIs are really good at this.

PR #534, opened 2026-03-09
Convert DDM duplicate polymorphic function panic to a proper error

shigoel:
`addDatatypeBindings!` (and callers up through `addBinding` and
`GlobalContext.addCommand`) called `panic!` when `defineChecked` returned
`.error` for a duplicate name.  This manifested when declaring the same
polymorphic function twice: the DDM already raised a proper error for
duplicate monomorphic function names, but the polymorphic case went through
`addDatatypeBindings!` and panicked instead.

### Changes

**`Strata/DDM/AST.lean`**
- `addDatatypeBindings` (renamed from `addDatatypeBindings!` — the `!`
  convention implies panic-on-error): return type changed to
  `Except String GlobalContext`; uses `do`-notation with
  `← gctx.defineChecked` to short-circuit on error.
- `preRegisterType`: accumulator changed from `GlobalContext` to
  `Except String GlobalContext`, matching the established pattern in
  `extractFieldsFromBindings`; returns `.error` instead of `panic!` for
  duplicate names.
- `addBinding`: same accumulator change; threads `Except` through to
  callers.
- `GlobalContext.addCommand`: return type changed to
  `Except String GlobalContext`; `Id.run do` pre-registration block
  replaced with direct `Except`-monad `do` notation; `foldBindingSpecs`
  initial value changed from `gctx` to `.ok gctx`.
- `Program.globalContext` default and `Program.addCommand` updated to
  handle the new return type.

**`Strata/DDM/Elab.lean`**
- `runCommand`: matches on `addCommand`'s result; on `.error e` calls
  `logError opInfo.loc e` (with the command's source location) and skips
  adding the command to `commands`, so the program's `globalContext`
  default recomputation also stays clean.

**`StrataTest/Languages/Core/Tests/DuplicateFnsDDMError.lean`**
- New `#guard_msgs` test that pins the error message for a program with a
  duplicate top-level function declaration. Before this PR's changes, this
  program would panic.

By submitting this pull request, I confirm that you can use, modify, copy,
and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/Strata/Languages/Laurel/ConstrainedTypeElim.lean
+++ b/Strata/Languages/Laurel/ConstrainedTypeElim.lean
@@ -1,0 +1,257 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Laurel.Laurel
+import Strata.Languages.Laurel.Resolution
+
+/-!
+# Constrained Type Elimination
+
+A Laurel-to-Laurel pass that eliminates constrained types by:
+1. Generating a constraint function per constrained type (e.g. `nat$constraint(x: int): bool`)
+2. Adding `requires constraintFunc(param)` for constrained-typed inputs
+3. Adding `ensures constraintFunc(result)` for constrained-typed outputs
+   - Skipped for `isFunctional` procedures since the Laurel translator does not yet support
+     function postconditions. Constrained return types on functions are not checked.
+4. Inserting `assert constraintFunc(var)` for local variable init and reassignment
+5. Assuming the constraint for uninitialized constrained-typed variables (havoc + assume)
+6. Adding a synthetic witness-validation procedure per constrained type
+7. Injecting constraint function calls into quantifier bodies (`forall` → `implies`, `exists` → `and`)
+8. Resolving all constrained type references to their base types
+-/
+
+namespace Strata.Laurel
+
+open Strata
+
+abbrev ConstrainedTypeMap := Std.HashMap String ConstrainedType
+/-- Map from variable name to its constrained HighType (e.g. UserDefined "nat") -/
+abbrev PredVarMap := Std.HashMap String HighType
+
+def buildConstrainedTypeMap (types : List TypeDefinition) : ConstrainedTypeMap :=
+  types.foldl (init := {}) fun m td =>
+    match td with | .Constrained ct => m.insert ct.name.text ct | _ => m
+
+partial def resolveBaseType (ptMap : ConstrainedTypeMap) (ty : HighType) : HighType :=
+  match ty with
+  | .UserDefined name => match ptMap.get? name.text with
+    | some ct => resolveBaseType ptMap ct.base.val | none => ty
+  | .Applied ctor args =>
+    .Applied ctor (args.map fun a => ⟨resolveBaseType ptMap a.val, a.md⟩)
+  | _ => ty
+
+def resolveType (ptMap : ConstrainedTypeMap) (ty : HighTypeMd) : HighTypeMd :=
+  ⟨resolveBaseType ptMap ty.val, ty.md⟩
+
+def isConstrainedType (ptMap : ConstrainedTypeMap) (ty : HighType) : Bool :=
+  match ty with | .UserDefined name => ptMap.contains name.text | _ => false
+
+/-- Build a call to the constraint function for a constrained type, or `none` if not constrained -/
+def constraintCallFor (ptMap : ConstrainedTypeMap) (ty : HighType)
+    (varName : Identifier) (md : Imperative.MetaData Core.Expression) : Option StmtExprMd :=
+  match ty with
+  | .UserDefined name => if ptMap.contains name.text then
+      some ⟨.StaticCall (mkId s!"{name.text}$constraint") [⟨.Identifier varName, md⟩], md⟩
+    else none
+  | _ => none
+
+/-- Generate a constraint function for a constrained type.
+    For nested types, the function calls the parent's constraint function. -/
+def mkConstraintFunc (ptMap : ConstrainedTypeMap) (ct : ConstrainedType) : Procedure :=
+  let md : Imperative.MetaData Core.Expression := #[]
+  let baseType := resolveType ptMap ct.base
+  let bodyExpr := match ct.base.val with
+    | .UserDefined parent =>
+      if ptMap.contains parent.text then
+        let parentCall : StmtExprMd :=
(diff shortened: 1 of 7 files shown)


Strata/DDM/AST.lean
```diff
@@ -2297,7 +2290,10 @@ structure Program where
   commands : Array Operation := #[]
   /-- Final global context for program. -/
   globalContext : GlobalContext :=
-    commands.foldl (init := {}) (·.addCommand dialects ·)
+    match commands.foldl (init := (Except.ok {} : Except String GlobalContext))
+        fun acc cmd => acc.bind (·.addCommand dialects cmd) with
+    | .ok gctx => gctx
+    | .error _ => {}
```
joscoh: Should this be silently swallowing the error?

Strata/DDM/AST.lean
```diff
@@ -2242,48 +2240,43 @@ private def preRegisterType (dialects : DialectMap) (gctx : GlobalContext) (l :
...
+def addCommand (dialects : DialectMap) (gctx : GlobalContext) (op : Operation) : Except String GlobalContext := do
     let dialectName := op.name.dialect
     -- Pre-register types if op has @[preRegisterTypes] metadata
-    let (gctx, preRegistered) := Id.run do
-      let .op decl := dialects.decl! op.name
-        | return (panic! "Expected operator declaration", false)
-      let .isTrue h := decideProp (op.args.size = decl.argDecls.size)
-        | return (panic! "Expected arguments to match", false)
+    let .op decl := dialects.decl! op.name
+      | return (panic! "Expected operator declaration")
```
joscoh: Why is a `panic` being returned here?

shigoel: Oversight, sorry. Will fix.

Strata/DDM/AST.lean
shigoel: No idea how to deal with this, given we're giving a default value to `globalContext`. Recommendations?

joscoh: Oh yeah, I missed the context. This is probably fine for a default then.

Strata/DDM/AST.lean
joscoh: This comment keeps getting resolved, but as far as I can see the `panic` is still there.

joehendrix: In the `Except String GlobalContext` do-block, `return (panic! "...")` evaluates `panic!` at type `GlobalContext` — in release builds this returns `default` (i.e. `{}`), then `return` wraps it as `.ok {}`, silently succeeding with an empty context instead of signaling failure.

```suggestion
      | .error "Expected operator declaration"
```

Strata/DDM/AST.lean
```diff
@@ -2242,48 +2240,43 @@ private def preRegisterType (dialects : DialectMap) (gctx : GlobalContext) (l :
...
+    let .isTrue h := decideProp (op.args.size = decl.argDecls.size)
+      | return (panic! "Expected arguments to match")
```
joehendrix: Same issue as above — `panic!` evaluates at type `GlobalContext` and `return` wraps it as `.ok {}`.

```suggestion
      | .error "Expected arguments to match"
```

Strata/DDM/AST.lean
joehendrix: Silently falling back to an empty `{}` context hides errors. If commands are pre-validated by the elaborator, this should never happen — a `panic!` would be more appropriate to surface bugs:

```suggestion
    | .error e => panic! s!"Program.globalContext: {e}"
```

Strata/DDM/AST.lean
```diff
@@ -2310,7 +2306,9 @@ instance : Inhabited Program where
 def addCommand (env : Program) (cmd : Operation) : Program :=
   { env with
     commands := env.commands.push cmd,
-    globalContext := env.globalContext.addCommand env.dialects cmd
+    globalContext := match env.globalContext.addCommand env.dialects cmd with
+      | .ok gctx => gctx
+      | .error _ => env.globalContext
```
joehendrix: Same issue — on error, the command is still appended to `commands` but the context is stale, creating an inconsistent `Program`. Should either `panic!` (if this is unreachable) or change the return type to `Except String Program`:

```suggestion
      | .error e => panic! s!"Program.addCommand: {e}"
```

joehendrix:
**Minor: Inconsistent panic in `addDatatypeBindings` (line 2216)**

Since `addDatatypeBindings` now returns `Except String GlobalContext`, the `panic!` at line 2216 for `extractConstructorInfo` errors could be propagated instead:

```lean
-- Current (line 2213-2216):
let constructorInfo :=
  match extractConstructorInfo dialects constrArg with
  | .ok info => info
  | .error e => panic! s!"Constructor extraction error: {e}"

-- Suggested:
let constructorInfo ← extractConstructorInfo dialects constrArg
```

This is outside the diff so I couldn't attach it as an inline suggestion.

shigoel:
> Some of the comments are minor, but I'd strongly prefer we don't replace the panic! with empty defaults.

Agreed, thanks. I haven't had the chance to clean this up, but will get to it later.

Strata/DDM/AST.lean
shigoel: Should be fixed now. Sorry about that.

Strata/DDM/AST.lean
```diff
@@ -2297,7 +2287,10 @@ structure Program where
   commands : Array Operation := #[]
   /-- Final global context for program. -/
   globalContext : GlobalContext :=
-    commands.foldl (init := {}) (·.addCommand dialects ·)
+    match commands.foldl (init := (Except.ok {} : Except String GlobalContext))
+        fun acc cmd => acc.bind (·.addCommand dialects cmd) with
+    | .ok gctx => gctx
+    | .error e => panic! s!"Program.globalContext: {e}"
```
shigoel:
```suggestion
    | .error e => panic! s!"Program.globalContext: {e}" -- nopanic:ok
```

Strata/DDM/AST.lean
```diff
@@ -2310,7 +2303,9 @@ instance : Inhabited Program where
 def addCommand (env : Program) (cmd : Operation) : Program :=
   { env with
     commands := env.commands.push cmd,
-    globalContext := env.globalContext.addCommand env.dialects cmd
+    globalContext := match env.globalContext.addCommand env.dialects cmd with
+      | .ok gctx => gctx
+      | .error e => panic! s!"Program.addCommand: {e}"
```
shigoel:
```suggestion
      | .error e => panic! s!"Program.addCommand: {e}" -- nopanic:ok
```

PR #542, opened 2026-03-10
Fix CallElim: substitute `old g` for globals not in `modifies`

shigoel:
## Problem

`callElimCmd` in `Strata/Transform/CallElim.lean` built the `old g` substitution
map by filtering only over `proc.spec.modifies`. For any global `g` that appeared
as `old g` in a postcondition but was **not** listed in `modifies`, no substitution
entry was produced. When the partial evaluator later encountered `old g` in the
generated `assume`, it resolved it against the oldest in-scope binding — the value
of `g` at the entry of the *calling* procedure — rather than at the call site.

## Fix

`Strata/Transform/CallElim.lean` — extend the substitution map with entries for
globals that are **not** in `modifies` but do appear as `old g` in postconditions.
For these globals the callee cannot change them, so `old g` at the call site equals
the current value of `g`; the substitution is simply `old g → createFvar g`, with
no fresh snapshot variable needed.

This mirrors what the reference evaluator already does in `ProcedureEval.lean`,
where `old_g_subst` is built for all program globals unconditionally.

## Test

`StrataTest/Languages/Core/Tests/CallElimWithOldGlobals.lean` — new `#guard_msgs`
tests covering both the reference-evaluator path and the CallElim path
(`proceduresToVerify`):

- `f` has a body (`z := g`) that provably satisfies `ensures (z == old g)` where
  `g` is not in `modifies`. `f_ensures_0` passes cleanly, keeping the scenario free
  of bodyless-procedure proof-obligation concerns.
- `h_correct` (`ensures (r == true)`) must pass on both paths.
- `h_incorrect` (`ensures (r == false)`) must fail on both paths.

---

By submitting this pull request, I confirm that you can use, modify, copy,
and redistribute this contribution, under the terms of your choice.

Thanks, @aaronjeline!

### DIFF
--- a/Strata/DDM/AST.lean
+++ b/Strata/DDM/AST.lean
@@ -1787,7 +1787,7 @@ partial def resolveBindingIndices { argDecls : ArgDecls } (m : DialectMap) (src 
   | .datatype b =>
     -- For datatypes, resolveBindingIndices only returns the datatype type
     -- itself; the constructors and template-generated functions are handled
-    -- separately in addDatatypeBindings!.
+    -- separately in addDatatypeBindings.
     let params : Array String :=
         let addBinding (a : Array String) (_ : SourceRange) {argDecls : _} (b : BindingSpec argDecls) (args : Vector Arg argDecls.size) :=
             match args[b.nameIndex.toLevel] with
@@ -2177,7 +2177,7 @@ FreeVarIndex values are consistent with this order.
 this adds entries for: `Option` (type), `None` (constructor), `Some` (constructor),
 `Option..isNone` (tester), `Option..isSome` (tester).
 -/
-private def addDatatypeBindings!
+private def addDatatypeBindings
     (dialects : DialectMap)
     (gctx : GlobalContext)
     (src : SourceRange)
@@ -2186,7 +2186,7 @@ private def addDatatypeBindings!
     {argDecls : ArgDecls}
     (b : DatatypeBindingSpec argDecls)
     (args : Vector Arg argDecls.size)
-    : GlobalContext :=
+    : Except String GlobalContext := do
 
   let datatypeName :=
     match args[b.nameIndex.toLevel] with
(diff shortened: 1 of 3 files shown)


StrataTest/Languages/Core/Tests/CallElimWithOldGlobals.lean
```diff
@@ -0,0 +1,82 @@
...
+Result: ❌ fail
+-/
+#guard_msgs in
+#eval verify oldModifiesPgm (options := .quiet)
+
+/--
+info:
+Obligation: h_correct_ensures_1
+Property: assert
+Result: ✅ pass
```
shigoel, line 75: Before this PR, this was reported as fail.

StrataTest/Languages/Core/Tests/CallElimWithOldGlobals.lean
```diff
@@ -0,0 +1,82 @@
...
+
+Obligation: h_incorrect_ensures_1
+Property: assert
+Result: ❌ fail
```
shigoel, line 79: This was a pass: hence the unsoundness!

StrataTest/Languages/Core/Tests/CallElimWithOldGlobals.lean
```diff
@@ -0,0 +1,82 @@
...
+def oldModifiesPgm :=
+#strata
+program Core;
+
+var g : bool;
+var i : bool;
+
+procedure f(x : bool) returns (z : bool)
+spec {
+  ensures (z == old g);
```
MikaelMayer, line 24: I think the translation from the CST to the AST should instead throw an error. There is no old value of a global if it's not modified. That indicate a code issue.
Also how does it work for globals that are just read without `old` ? How are they resolved in the AST? 
If global variables in modifies clauses become local variables (which is currently the case), then maybe we should have a way to  declare that a procedure's output depend on a global variable it only reads and not closes on.

shigoel: I'd like to limit the scope of this PR to fix the CallElim bug, esp. since using `old` is allowed without non-CallElim right now. 

I agree with you that eventually we should flag such uses of `old`, like we discussed earlier. However, I don't know if the CST->AST translator is the right place; perhaps Core's type checker is though.

MikaelMayer: FYI There is already no more "old" construct in the Core AST.

Strata/Transform/CallElim.lean
```diff
@@ -60,8 +60,20 @@ def callElimCmd (cmd: Command)
...
         let oldInit := createInitVars oldTripsRaw md
 
-        -- Substitute "old g" with the fresh pre-call variable in postconditions
-        let oldSubst := createOldVarsSubst oldTrips
+        -- Substitute "old g" in postconditions:
+        -- - For globals IN modifies: use the fresh snapshot variable.
+        -- - For globals NOT in modifies: old g == g at the call site (the callee
+        --   cannot modify them), so substitute old g directly with the current fvar.
+        let unmodifiedOldSubst : Map Expression.Ident Expression.Expr :=
+          p.decls.filterMap fun d => match d with
```
MikaelMayer, line 68: If there are 1000 globals, we are going to do 1000 traversals of all the post conditions and for each postcondition we are going to compute a set of all its free variables and for each free variable perform a concatenation.

Here is what I suggest to avoid this huge performance penalty.
- Create a function Lambda.LExpr.freeVarsAny that avoids computing the set, just runs the given predicate.
- Pre-compute a hashmap containing a mapping of old global names to global names (only once)
- Use this hashmap to ensure a linear traversal and not concatenating.

Note that if you implement my other suggestion, you would have no performance penalty to fix. The other suggestion would be to not accept references to global variables in the contract if they are not declared in a reads clause. There no hashmap, only verification that all free identifiers of contracts are actually referenced in the procedure's signature.

StrataTest/Languages/Core/Tests/CallElimWithOldGlobals.lean
shigoel: I'm confused: https://github.com/strata-org/Strata/blob/main/Strata/Languages/Core/Factory.lean#L203

shigoel: Removed this function now based on an offline conversation; should be unused.

PR #543, opened 2026-03-11
Boole frontend

kondylidou:
*Issue #, if available:*

*Description of changes:*

## Summary

This PR presents the Boole frontend end-to-end: grammar support, Boole-to-Core lowering, meta-verification integration, regression tests, and feature-request examples.

Boole is the IR for [CSLib](https://www.cslib.io/) ([Pillar 2](https://arxiv.org/abs/2602.04846)). Boole is built on Strata Core and serves as a staging ground for Strata Core feature requests from the CSLib community. Not all of these may make it into Core, but we will strive to have Boole and Core share as many components as possible. This will allow both CSLib and Strata to make progress without blocking on each other (due to differences in philosophies, priorities, etc.) and to keep up with each other.

## Main changes

- Added and refined the Boole frontend pipeline across parsing, lowering, and verification
- Integrated Boole programs with `#strata`, Core VC generation, and SMT-based checking
- Added Boole regression tests and moved Boole examples into `StrataTest/Languages/Boole`
- Added `StrataTest/Languages/Boole/FeatureRequests/` with Boole-facing seeds for unsupported or partially supported features
- Added `docs/BooleFeatureRequests.md` to summarize the current Boole feature backlog and point to the example files

## Structure

The PR touches the Boole frontend in four layers:
- grammar and surface syntax
- Boole-to-Core translation
- meta-verifier integration
- Boole tests and feature-request examples

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,8 +187,8 @@ jobs:
         auto-config: false
         build: true
         build-args: "strata:exe"
-    - name: Run PySpec test that may be skipped without Python
-      run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest
+    - name: Run PySpec and dispatch tests
+      run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest StrataTest.Languages.Python.AnalyzeLaurelTest StrataTest.Languages.Python.Specs.IdentifyOverloadsTest
     - name: Run test script
       run: FAIL_FAST=1 ./scripts/run_cpython_tests.sh ${{ matrix.python_version }}
       working-directory: Tools/Python
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -4,7 +4,9 @@
   SPDX-License-Identifier: Apache-2.0 OR MIT
 -/
 
+import Strata.Backends.CBMC.CollectSymbols
 import Strata.Backends.CBMC.GOTO.CoreToCProverGOTO
+import Strata.Transform.ProcedureInlining
 
 /-! ## Core-to-GOTO translation pipeline
 
@@ -426,5 +428,98 @@ def emitProcWithLifted (Env : Core.Expression.TyEnv) (procName : String)
       symtabObj := symtabObj.insert k v
   | _ => pure ()
   return (Lean.Json.obj symtabObj, Lean.Json.mkObj [("functions", Lean.Json.arr gotoFns)])
+
+/-! ## High-level pipeline steps
+
+Composable building blocks for translating Core programs to GOTO.
+-/
+
+/-- Inline procedure calls repeatedly until a fixpoint is reached.
+    By default inlines all procedures except `main`. -/
+public def inlineCoreFixpoint (program : Core.Program)
+    (doInline : String → Core.Transform.CachedAnalyses → Bool := fun name _ => name ≠ "main")
+    (maxIterations : Nat := 10)
+    : Except String Core.Program := do
+  let mut pgm := program
+  for _ in List.range maxIterations do
+    match Core.Transform.runProgram (targetProcList := .none)
+          (Core.ProcedureInlining.inlineCallCmd (doInline := doInline))
+          pgm .emp with
+    | ⟨.ok (changed, pgm'), _⟩ =>
+      pgm := pgm'
+      if !changed then break
+    | ⟨.error e, _⟩ => throw e
+  return pgm
+
+/-- Type-check a Core program using the standard context and factory.
+    Returns the type-checked program and the resulting type environment. -/
+public def typeCheckCore (program : Core.Program)
+    : Except String (Core.Program × Core.Expression.TyEnv) := do
+  let Ctx := { Lambda.LContext.default with
+    functions := Core.Factory, knownTypes := Core.KnownTypes }
+  let Env := Lambda.TEnv.default
+  match Core.Program.typeCheck Ctx Env program with
+  | .ok (tcPgm, Env') => return (tcPgm, Env')
+  | .error e => throw s!"{e.format none}"
+
+/-- Translate a type-checked Core program to CProver GOTO JSON files.
+    Finds an entry-point procedure from `entryPoints` (defaulting to
+    `["main", "__main__"]`), translates it to GOTO, and writes
+    `<baseName>.symtab.json` and `<baseName>.goto.json`. -/
+public def coreToGotoFiles (tcPgm : Core.Program) (Env : Core.Expression.TyEnv)
+    (baseName : String)
+    (sourceText : Option String := none)
+    (entryPoints : List String := ["main", "__main__"])
+    : EIO String Unit := do
+  let findProc (name : String) := tcPgm.decls.find? fun d =>
+      match d with
+      | .proc p _ => Core.CoreIdent.toPretty p.header.name == name
+      | _ => false
+  let mainDecl ← match entryPoints.findSome? findProc with
+    | some d => pure d
+    | none => throw s!"No entry-point procedure found (tried {entryPoints})"
+  let some p := mainDecl.getProc?
+    | throw "entry point is not a procedure"
+  -- Always use "main" as the GOTO function name (CBMC expects --function main)
+  let procName := "main"
+  let axioms := tcPgm.decls.filterMap fun d => d.getAxiom?
+  let distincts := tcPgm.decls.filterMap fun d => match d with
+    | .distinct name es _ => some (name, es) | _ => none
+  match procedureToGotoCtx Env p sourceText (axioms := axioms) (distincts := distincts)
+        (varTypes := tcPgm.getVarTy?) with
+  | .error e => throw s!"{e}"
+  | .ok (ctx, liftedFuncs) =>
+    let extraSyms ← match collectExtraSymbols tcPgm with
+      | .ok s => pure (Lean.toJson s)
+      | .error e => throw s!"{e}"
+    let (symtab, goto) ←
+      match ← emitProcWithLifted Env procName ctx liftedFuncs extraSyms |>.toBaseIO with
+      | .ok r => pure r
+      | .error e => throw s!"{e}"
+    let symTabFile := s!"{baseName}.symtab.json"
+    let gotoFile := s!"{baseName}.goto.json"
+    match ← IO.FS.writeFile symTabFile symtab.pretty |>.toBaseIO with
+    | .ok () => pure ()
+    | .error e => throw s!"Error writing {symTabFile}: {e}"
+    match ← IO.FS.writeFile gotoFile goto.pretty |>.toBaseIO with
+    | .ok () => pure ()
+    | .error e => throw s!"Error writing {gotoFile}: {e}"
+    let _ ← IO.println s!"Written {symTabFile} and {gotoFile}" |>.toBaseIO
+
+/--
+Inline procedures, type-check, and emit CProver GOTO JSON files.
+-/
+public def inlineCoreToGotoFiles (program : Core.Program)
+    (baseName : String)
+    (sourceText : Option String := none)
+    (entryPoints : List String := ["main", "__main__"])
+    : EIO String Unit := do
+  let inlined ← match inlineCoreFixpoint program with
+    | .ok r => pure r
+    | .error msg => throw msg
+  let (tcPgm, Env) ← match typeCheckCore inlined with
+    | .ok r => pure r
+    | .error msg => throw msg
+  coreToGotoFiles tcPgm Env baseName sourceText entryPoints
 
 end Strata
--- a/Strata/Languages/Python/PySpecPipeline.lean
+++ b/Strata/Languages/Python/PySpecPipeline.lean
@@ -1,0 +1,318 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+import Strata.Languages.Python.PythonToCore
+public import Strata.Languages.Python.PythonToLaurel
+import Strata.Languages.Python.Specs
+import Strata.Languages.Python.Specs.DDM
+import Strata.Languages.Python.Specs.IdentifyOverloads
+import Strata.Languages.Python.Specs.ToLaurel
+
+/-! ## PySpec Pipeline
+
+Implementation of the Python-to-Core pipeline via PySpec and Laurel.
+Reads PySpec Ion files, resolves overloads, builds Laurel declarations,
+and translates through to Core for verification.
+-/
+
+namespace Strata
+
+open Python.Specs.ToLaurel (OverloadTable)
+
+/-! ### Types -/
+
+/-- Result of reading PySpec files: combined Laurel declarations and overload table. -/
+public structure PySpecLaurelResult where
+  laurelProgram : Laurel.Program
+  overloads : OverloadTable
+  functionSignatures : List Python.PythonFunctionDecl := []
+
+/-! ### Private Helpers -/
+
+/-- Convert a SpecDefault to a Python None expression. -/
+private def specDefaultToExpr : Python.Specs.SpecDefault → Python.expr SourceRange
+  | .none => .Constant default (.ConNone default) default
+
+/-- Convert a pyspec Arg to a PythonFunctionDecl arg tuple. -/
+private def specArgToFuncDeclArg (arg : Python.Specs.Arg)
+    : String × String × Option (Python.expr SourceRange) :=
+  (arg.name, "Any", arg.default.map specDefaultToExpr)
+
+/-- Build a PythonFunctionDecl from a PySpec FunctionDecl or class method,
+    expanding `**kwargs` TypedDict fields into individual parameters. -/
+private def funcDeclToFunctionDecl (name : String) (args : Python.Specs.ArgDecls)
+    : Except String Python.PythonFunctionDecl := do
+  let kwargsArgs ← Python.Specs.ToLaurel.expandKwargsArgs args.kwargs
+  let allArgs := args.args ++ args.kwonly ++ kwargsArgs
+  pure { name, args := allArgs.toList.map specArgToFuncDeclArg,
+         hasKwargs := false, ret := none }
+
+/-- Extract PythonFunctionDecl entries from pyspec signatures.
+    Handles both top-level functions and class methods.
+    Strips `self` from class methods and expands `**kwargs` TypedDict fields. -/
+private def extractFunctionSignatures (sigs : Array Python.Specs.Signature)
+    (modulePrefix : String) : Except String (List Python.PythonFunctionDecl) := do
+  let prefixName (n : String) := if modulePrefix.isEmpty then n else modulePrefix ++ "_" ++ n
+  let mut result : List Python.PythonFunctionDecl := []
+  for sig in sigs do
+    match sig with
+    | .functionDecl func =>
+      if !func.isOverload then
+        result := result ++ [← funcDeclToFunctionDecl (prefixName func.name) func.args]
+    | .classDef cls =>
+      let clsName := prefixName cls.name
+      for method in cls.methods do
+        if method.args.args.size == 0 then
+          throw s!"Method '{cls.name}.{method.name}' has no arguments (expected 'self' as first parameter)"
+        let posArgs := method.args.args.extract 1 method.args.args.size  -- strip self
+        let decl ← funcDeclToFunctionDecl (clsName ++ "_" ++ method.name) { method.args with args := posArgs }
+        result := result ++ [decl]
+    | _ => pure ()
+  return result
+
+/-! ### Building PySpec Laurel Declarations -/
+
+private def mergeOverloads (old new : OverloadTable) : OverloadTable :=
+  new.fold (init := old) fun o name n =>
+    o.alter name fun s => some <| s.getD {} |>.union n
+
+/-- Read a Python Ion program and return the single module command.
+    Throws if the file is not Ion or does not contain exactly one module. -/
+public def readPythonIonModule (strataPath : String)
+    : EIO String (Python.Command SourceRange) := do
+  let bytes ←
+    match ← IO.FS.readBinFile strataPath |>.toBaseIO with
+    | .ok b => pure b
+    | .error msg => throw s!"Error reading {strataPath}: {msg}"
+  if !Ion.isIonFile bytes then
+    throw s!"{strataPath} is not an Ion file"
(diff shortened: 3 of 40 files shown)


Strata/DDM/Integration/Lean/HashCommands.lean
```diff
@@ -127,17 +127,57 @@ def strataDialectImpl: CommandElab := fun (stx : Syntax) => do
...
+        (acc ++ "forall ", true)
+      else if !inBinder && ch == '∃' then
+        (acc ++ "exists ", true)
+      else if inBinder && ch == '.' then
+        (acc ++ " :: ", false)
+      else
+        (acc.push ch, inBinder))
+  out
+
+private def isBooleProgram (src : String) : Bool :=
```
joscoh: The DDM should not be hard-coded to mention Boole. Rather than `normalizeQuantifierUnicode`, this translation of syntax should be in Boole's DDM-defined grammar IMO.

Strata/Languages/Boole/CoreBridge.lean
```diff
@@ -0,0 +1,247 @@
...
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Core.Program
+import Strata.Languages.Core.Statement
+import Strata.DL.Lambda.LExpr
+
+namespace Strata.Boole.CoreBridge
```
joscoh: This file should have a comment with a small explanation of its purpose. I'm a bit confused about this file anyway, since it seems to consist essentially of redefinitions of Core constructors.

Strata/Languages/Boole/Grammar.lean
```diff
@@ -0,0 +1,141 @@
...
+op prog (commands : SpacePrefixSepBy Command) : Program =>
+  commands;
+
+#end
+
+---------------------------------------------------------------------
+
+end Strata
+
+def test : Strata.Program :=
```
joscoh: Should be in separate test file. Also, what is this testing exactly?

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,669 @@
...
+  | .sub_expr m ty a b => toCoreTypedBin m ty "Sub" (← toCoreExpr a) (← toCoreExpr b)
+  | .mul_expr m ty a b => toCoreTypedBin m ty "Mul" (← toCoreExpr a) (← toCoreExpr b)
+  | .div_expr m ty a b => toCoreTypedBin m ty "Div" (← toCoreExpr a) (← toCoreExpr b)
+  | .mod_expr m ty a b => toCoreTypedBin m ty "Mod" (← toCoreExpr a) (← toCoreExpr b)
+  | .old _ _ a =>
+      match CoreBridge.mkExprOld (← toCoreExpr a) with
+      | .ok e => return e
+      | .error msg => throw (.fromMessage msg)
+  | .forall _ ds body
+  | .forall_unicode _ ds body
```
joscoh: I am confused why you need special support in the DDM for the unicode quantifiers when you seem to be translating them here anyway.

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,669 @@
...
+  match idxs with
+  | [] => rhs
+  | [i] => CoreBridge.mkExprMapUpdate base i rhs
+  | i :: rest =>
+    let innerMap := CoreBridge.mkExprMapSelect base i
+    let updatedInner := nestMapSet innerMap rest rhs
+    CoreBridge.mkExprMapUpdate base i updatedInner
+
+def toCoreInvariants (is : BooleDDM.Invariants SourceRange) : TranslateM (List Core.Expression.Expr) := do
+  go is
```
joscoh: Very minor style point, but not sure why this isn't just directly matching on `is`, since `toCoreInvariants` and `toCoreInvariants.go` are exactly the same after unfolding.

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,669 @@
...
+
+def lowerVarStatement (ds : BooleDDM.DeclList SourceRange) : TranslateM (List Core.Statement) := do
+  let mut out : List Core.Statement := []
+  let mut newBVars : List Core.Expression.Expr := []
+  for d in declListToList ds do
+    let (id, ty) ← toCoreBind d
+    let n := (← get).globalVarCounter
+    modify fun st => { st with globalVarCounter := n + 1 }
+    let initName := mkIdent s!"init_{id.name}_{n}"
+    newBVars := newBVars ++ [(.fvar () id none : Core.Expression.Expr)]
```
joscoh: Appending to the end in each iteration is `O(n^2)`, while using `::` in each iteration then reversing the list at the end is `O(n)`. For large Boole programs this could make a difference.

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,669 @@
...
+  modify fun st => { st with bvars := st.bvars ++ newBVars.toArray }
+  return out
+
+mutual
+
+def toCoreBlock (b : BooleDDM.Block SourceRange) : TranslateM (List Core.Statement) := do
+  match b with
+  | .block _ ⟨_, ss⟩ =>
+    let mut out : List Core.Statement := []
+    for s in ss do
```
joscoh: This loop could just be a single `mapM`, which would also be more efficient (for the same reason as before).

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,669 @@
...
+  let some _ := ctorList.head?
+    | throwAt m s!"Datatype {dtypeName} must have at least one constructor"
+  let constrs ← ctorList.mapM (toCoreDatatypeConstr dtypeName)
+  match h : constrs with
+  | [] => throwAt m s!"Datatype {dtypeName} must have at least one constructor"
+  | _ :: _ =>
+    return { name := dtypeName
+             typeArgs := typeParams
+             constrs := constrs
+             constrs_ne := by simp [h] }
```
joscoh: Optional, but instead of checking whether the list is nil twice, you could store the proof that `ctorList.head?` is `some` (or not `nil`) and use the fact that `mapM` preserves length. `grind` may be able to handle this automatically, but I am not sure.

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,669 @@
...
+    let (id, ty) ← toCoreBind b
+    let i := (← get).globalVarCounter
+    modify fun s => { s with globalVarCounter := i + 1 }
+    return [CoreBridge.mkVarDecl id ty (some (.fvar () (mkIdent s!"init_{id.name}_{i}") none))]
+  | .command_axiom m ⟨_, l?⟩ e =>
+    return [CoreBridge.mkAxiomDecl (← defaultLabel m "axiom" l?) (← toCoreExpr e)]
+  | .command_distinct m ⟨_, l?⟩ ⟨_, es⟩ =>
+    return [CoreBridge.mkDistinctDecl (mkIdent (← defaultLabel m "distinct" l?)) (← es.toList.mapM toCoreExpr)]
+  | .command_block _ b =>
+    return [CoreBridge.mkProcDecl {
```
joscoh: Why does translating a block create a new procedure?

Strata/MetaVerifier.lean
```diff
@@ -0,0 +1,510 @@
...
+def genVCs (program : Program) (options : VerifyOptions := .default) : Option coreVCs := do
+  match Core.typeCheckAndPartialEval options program with
+  | .error _ => none
+  | .ok pEs =>
+    let VCss ← List.mapM (fun pE => genVCsSingleENV pE) pEs
+    return VCss.flatten.reverse
+
+end Core
+
+namespace C_Simp
```
joscoh, line 67: Why is C_Simp relevant here?

Strata/MetaVerifier.lean
```diff
@@ -0,0 +1,510 @@
...
+  else if program.dialect == "Boole" then
+    match Boole.getProgram program with
+    | .ok booleProgram =>
+      Boole.genVCs booleProgram program.globalContext { (default : Core.VerifyOptions) with verbose := .quiet : Core.VerifyOptions }
+    | .error _ => none
+  else
+    none
+
+/--
+Remove solver-only context caches that are irrelevant for denotation and
```
joscoh: Note: the removed entries (e.g. `seenDatatypes`) are only irrelevant for the denotation because the denotation does not yet handle datatypes, I believe.

Strata/MetaVerifier.lean
```diff
@@ -0,0 +1,510 @@
...
+  exact hp
+
+noncomputable def denoteProofObligations (vcs : Core.coreVCs) : Option Prop := do
+  match vcs with
+  | [] => return True
+  | (E, ob) :: vcs =>
+    let p ← denoteProofObligation E ob
+    go vcs p
+where
+  go vcs p : Option Prop := do
```
joscoh: Optional possible alternative that does not require a separate recursive function:
```
vcs.foldlM (fun p (E, ob) => do 
  let q <- denoteProofOblication E ob 
  return (p /\ q)
) (return p)
```

Strata/MetaVerifier.lean
```diff
@@ -0,0 +1,510 @@
...
+noncomputable def coreVCsCorrect (program : Program) : Prop :=
+  match genCoreVCs program with
+  | some vcs => (denoteProofObligations vcs).getD False
+  | none     => False
+
+/--
+Interpret a list of SMT verification conditions as the conjunction of their
+denotations.
+-/
+noncomputable def denoteQueries (vcs : SMT.SMTVCs) : Option Prop := do
```
joscoh, line 126: This seems almost identical to `denoteProofObligations`. The two can be factored out into a separate function.

Strata/MetaVerifier.lean
```diff
@@ -0,0 +1,510 @@
...
+-/
+def genSMTVCs (program : Program) : Option SMT.SMTVCs := do
+  let coreVCs ← genCoreVCs program
+  toSMTVCs coreVCs
+
+/--
+State semantic correctness of the SMT verification conditions generated for a
+program.
+-/
+def smtVCsCorrect (program : Program) : Prop :=
```
joscoh, line 159: What is the difference between this function and `coreVCsCorrect`?

StrataTest/Languages/Boole/array_assignment.lean
```diff
@@ -0,0 +1,45 @@
...
+  while (i < m)
+  {
+    j := 0;
+    while (j < n)
+    {
+      // Array assignment is parsed as regular assignment over recursive LHS
+      // indexing (`arr[i][j]... := v`).
+      // Lowering recursively nests map `select`/`update`, so any index depth works.
+      B[i][j] := A[j][i];
+      // Previously would have been:
```
joscoh: Not sure the comment about the previous version is needed here, but up to you.

StrataTest/Languages/Boole/deterministic.lean
```diff
@@ -0,0 +1,51 @@
...
+private def deterministic :=
+#strata
+program Boole;
+
+function f(a:int) : int;
+
+procedure Foo(x:int) returns (r:int)
+spec
+{
+  free ensures r == f(x);
```
joscoh, line 20: What is the purpose of the body in this example if the postcondition is `free`?

StrataTest/Languages/Boole/demo.lean
joscoh, line 1: This should also be in `StrataTest/Languages/Boole` (along with `FindMax.lean`)

StrataTest/Languages/Boole/find_max_verus.lean
joscoh, line 1: Why are `find_max.lean` and `FindMax.lean` both needed? It seems like this one proves a stronger spec.

docs/BooleFeatureRequests.md
```diff
@@ -0,0 +1,85 @@
...
+18. Datatype constructor/selector verification robustness
+    Improve solver/type-checker handling for datatype VCs that are already emitted faithfully.
+19. Complex recursive type shapes
+    Support more nested recursive datatype shapes during type-checking.
+20. Non-Boole SST artifacts
+    Decide whether `RevealString` / `Air`-style statements need first-class treatment or an explicit erase/lower policy.
+
+## Boole seed examples
+
+These are now separate Lean examples under [`StrataTest/Languages/Boole/FeatureRequests`](/home/lydia/Desktop/PhD/projects/Strata/StrataTest/Languages/Boole/FeatureRequests).
```
joscoh: Should not link to your personal computer :)

StrataTest/Languages/Boole/FeatureRequests/decreases_metadata.lean
```diff
@@ -0,0 +1,48 @@
...
+spec {
+  requires 0 <= n;
+  ensures i == n;
+}
+{
+  i := 0;
+  while (i < n)
+    invariant 0 <= i
+    invariant i <= n
+    // TODO(feature:decreases): preserve `decreases n - i` as semantic metadata.
```
joscoh: FYI `decreases` clauses for loops are now in Core https://github.com/strata-org/Strata/pull/581

Strata/Languages/Boole/Verify.lean
kondylidou: Unicode support in the DDM is needed to parse the concrete syntax.
The separate handling in Verify exists only because the generated Boole AST currently keeps ASCII and Unicode spellings as different constructors.

Strata/Languages/Boole/Verify.lean
kondylidou: `command_block` is not meant to be semantically a procedure. The issue is that `toCoreDecls` must return `Core.Decl`, and there is no standalone top-level block declaration in Core, so we wrap a command-level block as an anonymous procedure. I added a comment to make that explicit.

Strata/MetaVerifier.lean
kondylidou: C_Simp is only relevant here because this file defines the frontend-specific wrapper `C_Simp.genVCs`, which is used below. If we inline that wrapper into `Strata.genCoreVCs`, then we can remove the namespace.`

Strata/MetaVerifier.lean
kondylidou: `coreVCsCorrect` is about the semantics of the Core proof obligations directly. `smtVCsCorrect` is about the semantics after translating those obligations into SMT queries. So `coreVCsCorrect` is before SMT encoding, and `smtVCsCorrect` is after SMT encoding. I was unsure about keeping both but we have an example with both so keeping it for now.

StrataTest/Languages/Boole/deterministic.lean
kondylidou: The example tests that free ensures are assumed at call sites, even for a recursive procedure that has been implemented, not just for a declaration-only stub.

docs/BooleFeatureRequests.md
kondylidou: whoops :p

Strata/DDM/Integration/Lean/HashCommands.lean
kondylidou: This needs to happen before the Boole DDM, because of the dotted syntax. I'm trying to find a nicer way to do this

Strata/MetaVerifier.lean
joscoh: I am still confused because `coreVCsCorrect` ultimately references `Core.ProofObligation.toSMTObligation`, so both seem to be after the Core proof obligations have been translated to SMT. It is fine to keep both, but I still don't understand the difference.

Strata/DDM/Elab.lean
```diff
@@ -36,6 +36,47 @@ inductive Header where
...
+          (acc.push ch, .outside)
+        else
+          (acc.push ch, .inBinder))).fst
+
+private def normalizeInputContext (inputContext : InputContext) : InputContext :=
+  let inputString := normalizeUnicodeQuantifierSeparators inputContext.inputString
+  Strata.Parser.stringInputContext (System.FilePath.mk inputContext.fileName) inputString
+
+private def normalizeStopPos (original normalized : InputContext) (stopPos : String.Pos.Raw) : String.Pos.Raw :=
+  if stopPos == original.endPos then normalized.endPos else stopPos
```
shigoel: `normalizeStopPos` only patches the edge case where `stopPos == original.endPos`, but `normalizeUnicodeQuantifierSeparators` replaces `.` (1 byte) with `::` (2 bytes) inside quantifier binders, so the normalized string can be strictly longer than the original. Any `stopPos` pointing into the middle of the string that follows a normalization site will be off-by-N in the normalized context. This could silently produce wrong error locations or misparsing that is hard to diagnose. At minimum, add a comment explaining the assumption that only the end-position case matters here; ideally, build a proper position map from original → normalized byte offsets.

Strata/MetaVerifier.lean
```diff
@@ -0,0 +1,513 @@
...
+    go vcs p
+where
+  go vcs p : Option Prop := do
+  match vcs with
+  | [] => return p
+  | (_, ctx, ts, t) :: vcs =>
+    let q ← denoteQuery ctx.toCore ts t
+    go vcs (p ∧ q)
+
+def toSMTVCs vcs := do
```
shigoel: Could we please add type annotation to `vcs` for readability?

kondylidou: done:)

Strata/MetaVerifier.lean
```diff
@@ -0,0 +1,513 @@
...
+  match vcs with
+  | [] => return []
+  | (E, ob) :: vcs =>
+    let (ctx, ts, t) ← Core.ProofObligation.toSMTObligation E ob
```
shigoel: Minor, but looks like `Core.ProofObligation.toSMTObligation` returns a list of `SMTVC`, which is `String × SanitizedContext × List Term × Term`. So this destructuring is a tad brittle.

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,733 @@
...
+    return [.var id ty (some (.fvar () (mkIdent s!"init_{id.name}_{i}") none))]
+  | .command_axiom m ⟨_, l?⟩ e =>
+    return [.ax { name := ← defaultLabel m "axiom" l?, e := ← toCoreExpr e }]
+  | .command_distinct m ⟨_, l?⟩ ⟨_, es⟩ =>
+    return [.distinct (mkIdent (← defaultLabel m "distinct" l?)) (← es.toList.mapM toCoreExpr)]
+  | .command_block _ b =>
+    -- Core decls do not have a standalone "top-level block" form, so a Boole
+    -- command-level block is wrapped as an anonymous procedure declaration.
+    return [.proc {
+      header := { name := mkIdent "", typeArgs := [], inputs := [], outputs := [] }
```
shigoel: Indeed, Core doesn't have a top-level block, but an empty string as a name worries me. I'd rather have some other sentinel name here (`"__Boole_top"`, `"__Boole_init__"`, or some such).

Strata/DDM/Elab.lean
kondylidou: I see your concern. I'm working on a mapping

Strata/Languages/Boole/Verify.lean
shigoel: Also, in case you missed it, Core verifier has the capability to select procedures to verify, which incurs a pass that includes only reachable procedures (e.g., see https://github.com/strata-org/Strata/blob/main/StrataTest/Languages/Core/Examples/SelectiveVerification.lean). You could imagine having Boole's top there.

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,733 @@
...
+  | .arrow _ a b => return .arrow (← toCoreMonoType a) (← toCoreMonoType b)
+  | .bool _ => return .bool
+  | .int _ => return .int
+  | .bv1 _ => return .bitvec 1
+  | .bv8 _ => return .bitvec 8
+  | .bv16 _ => return .bitvec 16
+  | .bv32 _ => return .bitvec 32
+  | .bv64 _ => return .bitvec 64
+  | .Map _ v k => return .tcons "Map" [← toCoreMonoType k, ← toCoreMonoType v]
+  | _ => throwAt default "Unsupported type"
```
shigoel: Let's have a more informative error message here. What's the unsupported type? What's the source location (i.e., don't use `default`)?

Strata/Languages/Boole/Verify.lean
```diff
@@ -0,0 +1,733 @@
...
+  | .neg_expr m ty a => toCoreTypedUn m ty "Neg" (← toCoreExpr a)
+  | .add_expr m ty a b => toCoreTypedBin m ty "Add" (← toCoreExpr a) (← toCoreExpr b)
+  | .sub_expr m ty a b => toCoreTypedBin m ty "Sub" (← toCoreExpr a) (← toCoreExpr b)
+  | .mul_expr m ty a b => toCoreTypedBin m ty "Mul" (← toCoreExpr a) (← toCoreExpr b)
+  | .div_expr m ty a b => toCoreTypedBin m ty "Div" (← toCoreExpr a) (← toCoreExpr b)
+  | .mod_expr m ty a b => toCoreTypedBin m ty "Mod" (← toCoreExpr a) (← toCoreExpr b)
+  | .old _ _ a =>
+      match (← toCoreExpr a) with
+      | .fvar m ident ty => return .fvar m (Core.CoreIdent.mkOld ident.name) ty
+      | other => throw (.fromMessage s!"old: expected an identifier, got {other}")
```
shigoel, line 309: Out of scope for this PR but prompted by reading it, so I'll share :)  

We don't do this in Strata yet (AFAIK), but it'd be nice if for currently unsupported expressions like`old(x[i])`, we gave an suggestion for alternatives (e.g., use `freshvar := x[i]` and then `old(freshvar)`).

kondylidou: Indeed a nice suggestion!

Strata/DDM/Elab.lean
kondylidou: fixed

Strata/Languages/Boole/Verify.lean
kondylidou: fixed

Strata/Languages/Boole/Verify.lean
kondylidou: done:)

StrataTest/Languages/Boole/top_level_block_selection.lean
```diff
@@ -0,0 +1,37 @@
...
+
+/--
+info:
+Obligation: top_assert
+Property: assert
+Result: ✅ pass
+-/
+#guard_msgs in
+#eval Strata.Boole.verify "cvc5" topLevelBlockSelection
+        (proceduresToVerify := (some [Strata.Boole.topLevelBlockProcedureName]))
```
shigoel, line 36: Sweet!

PR #546, opened 2026-03-11
Add support for Dict and List in Python->Laurel

thanhnguyen-aws:
This PR add support for Dict and List  in Python->Laurel, which includes:
1. Encoding literal Dict and List. Limitation: Dict keys must be string.
2. Encoding In and NotIn operators for Dict and List 
3. Subscription and subscription assign. Limitation: not support range slice for List yet.


Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -163,43 +165,204 @@ inline function Any_to_bool (v: Any) : bool
...
+rec function List_get (@[cases] l : ListAny, i : int) : Any
+  requires i >= 0 && i < List_len(l);
+{
+  if ListAny..isListAny_nil(l) then from_none()
+  else if  i == 0 then ListAny..head!(l)
+  else List_get(ListAny..tail!(l), i - 1)
+}
+
+rec function List_take (@[cases] l : ListAny, i: int) : ListAny
+  requires i >= 0 && i < List_len(l);
```
shigoel: **Bug: `List_take` precondition is too restrictive — should use `<=`**

`requires i >= 0 && i < List_len(l)` prevents two valid calls:
- `List_take([], 0)` — taking nothing from an empty list is valid but `0 < 0` fails.
- `List_take(l, List_len(l))` — taking all elements is a valid Python slice idiom (`l[:len(l)]`).

Should be `requires i >= 0 && i <= List_len(l)`.

The same issue applies to `List_drop` (line 208), and the supporting axioms `List_take_len` and `List_drop_len` should be extended to cover `i == List_len(l)` as well.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -212,6 +204,15 @@ def AnyNone := mkStmtExprMd (.StaticCall "from_none" [])
...
 
+def getSubscriptList (expr:  Python.expr SourceRange) : List ( Python.expr SourceRange) :=
+  match expr with
+  | .Subscript _ val slice _ => (getSubscriptList val) ++ [slice]
+  | _ => [expr]
+
+def pyOptExprToString (e : Python.opt_expr SourceRange) : Except TranslationError String := do
+  match e with
+  | .some_expr _ (.Constant _ (.ConString _ s) _) => return s.val
+  | _ => throw (.internalError "Expected some constant string: {e}")
```
shigoel: **Bug: string interpolation `{e}` will not work without the `s!` prefix**

`"Expected some constant string: {e}"` is a plain Lean string literal — `{e}` will appear literally rather than being interpolated. Should be:
```lean
s!"Expected some constant string: {repr e}"
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -212,6 +212,15 @@ def AnyNone := mkStmtExprMd (.StaticCall "from_none" [])
...
 
+def getSubscriptList (expr:  Python.expr SourceRange) : List ( Python.expr SourceRange) :=
+  match expr with
+  | .Subscript _ val slice _ => (getSubscriptList val) ++ [slice]
+  | _ => [expr]
+
+def pyOptExprToString (e : Python.opt_expr SourceRange) : Except TranslationError String := do
+  match e with
+  | .some_expr _ (.Constant _ (.ConString _ s) _) => return s.val
+  | _ => throw (.internalError "Expected some constant string: {repr e}")
```
shigoel:
```suggestion
  | _ => throw (.internalError s!"Expected some constant string: {repr e}")
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -257,8 +266,26 @@ def hasModel (ctx : TranslationContext) (funcName : String) : Bool :=
...
 
+partial def translateList (ctx : TranslationContext) (elmts: List (Python.expr SourceRange))
+    : Except TranslationError StmtExprMd := do
+  let trans_elmts ←  elmts.mapM (translateExpr ctx)
+  return  mkStmtExprMd (.StaticCall "from_ListAny" [ListAny_mk trans_elmts])
+
+partial def translateDictStrAny (ctx : TranslationContext)
+    (keys: List (Python.opt_expr SourceRange)) (values: List (Python.expr SourceRange))
+      : Except TranslationError StmtExprMd := do
+  assert! keys.length == values.length
```
shigoel: Why `assert!` instead of `throw`?

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -163,43 +165,204 @@ inline function Any_to_bool (v: Any) : bool
...
+  if ListAny..isListAny_nil(l) then ListAny_nil()
+  else if  i == 0 then l
+  else List_drop(ListAny..tail!(l), i - 1)
+}
+
+axiom [List_drop_len]: forall l : ListAny, i: int :: {List_len(List_drop(l,i))}
+  (i >= 0 && i <= List_len(l)) ==> List_len(List_drop(l,i)) == List_len(l) - i;
+
+inline function List_slice (l : ListAny, start : int, stop: int) : ListAny
+  requires start >= 0 && start < List_len(l) && stop >= 0 && stop < List_len(l) && start <= stop;
```
shigoel: Should be `stop <= List_len(l)`.

shigoel: Actually, consider this Python List slicing operation:
```
l = [1,2,3]
l[0:4] # gives back the whole list
```

PR #549, opened 2026-03-11
Convert SimpleAPI and transitive imports to Lean 4 modules

joehendrix:
## Summary

This PR converts 68 files in the transitive import closure of
`Strata/SimpleAPI.lean` into Lean 4 modules, controlling visibility of
internal definitions and enabling downstream consumers to depend on a
stable API surface. Along the way, several proof files required rework
due to changed reduction behavior, and ~500 lines of GOTO pipeline code
were extracted from `StrataMain` into proper library modules.

### Module conversion (68 files)

Every `.lean` file reachable from `SimpleAPI.lean` now begins with
`module`. For each file:

- A `module` declaration was added after the copyright header.
- Definitions needed by downstream files were wrapped in `public
  section` blocks or individually marked `public`.
- Imports used in public signatures were changed from `import` to
  `public import`.
- Type aliases and abbreviations used in `simp` or pattern matching
  downstream were annotated with `@[expose]` so they remain transparent
  across module boundaries.
- A handful of files use `import all` to access private helper lemmas
  needed by proofs (e.g., `simp` lemmas that extend the `List` namespace
  in utility modules).

The 66 originally-planned files plus 2 new files added on `main` during
development (`EliminateReturnsInExpression`, `PythonLaurelCorePrelude`)
were converted for a total of 68.

### Proof rework in well-formedness files

Making `Forall` transparent with `@[expose]` changed how `simp`
normalizes proof goals throughout the well-formedness proofs. This
required updating proofs in `StatementWF.lean`, `ProgramWF.lean`, and
`ListUtils.lean`. The `ProgramWF` rework was the most substantial,
rewriting the main inductive proof to use `grind` and `simp` with
explicit unfolding. No functional changes—only proof strategies differ.

### GOTO pipeline extraction from StrataMain

Approximately 500 lines of GOTO translation code were extracted from
`StrataMain.lean` into two new library modules:

- **`Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean`**: Core-to-GOTO
  translation logic for procedures and functions.
- **`Strata/Backends/CBMC/CollectSymbols.lean`**: Symbol table
  collection for datatypes, type constructors, and global variables.

This extraction was necessary because the E2E GOTO test previously
imported `StrataMain` directly, which caused interpreter failures with
compiler-generated specializations across module boundaries. The test now
imports the pipeline modules directly. `StrataMain` imports the pipeline
module and no longer contains duplicated code.

### Test file adjustments

- `E2E_CoreToGOTO.lean` was updated to import the new pipeline modules
  instead of `StrataMain`.
- All `#eval!` tests were changed to `#eval` to avoid runtime issues
  with evaluating code containing `sorry`.

### Known issues / follow-ups

- **ListUtils public definitions**: `Forall`, `List.Disjoint`, and
  related definitions in `ListUtils.lean` must remain public because
  they appear in public structure field types in `WF.lean`. A follow-up
  should either move these into a namespaced module or refactor `WF.lean`
  so its public API doesn't depend on them directly.


shigoel (dismissed):

tautschnig (dismissed):

joscoh (dismissed):

shigoel (dismissed):

joscoh (approved):

Strata/Languages/Core/Factory.lean
```diff
@@ -3,19 +3,29 @@
...
-import Strata.Languages.Core.Expressions
-import Strata.DL.Lambda.Factory
-import Strata.DL.Lambda.FactoryWF
-import Strata.DL.Lambda.IntBoolFactory
+module
+
+public meta import Lean.Elab.Command
+
+public import Strata.Languages.Core.Identifiers
+public meta import Strata.Languages.Core.Identifiers
```
MikaelMayer, line 11: Nit: This file imports `Strata.Languages.Core.Identifiers` twice — once as `public import` (line 10) and once as `public meta import` (line 11). Is the `meta` variant needed in addition to the regular one, or is this a leftover from iterating on the conversion?

Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
```diff
@@ -0,0 +1,430 @@
...
+        let argExprs ← renamedArgs.mapM toExpr
+        let lhsExpr := match renamedLhs with
+          | id :: _ =>
+            let name := Core.CoreIdent.toPretty id
+            let ty := match trans.T.context.types.find? id with
+              | some lty =>
+                match lty.toMonoTypeUnsafe.toGotoType with
+                | .ok gty => gty
+                | .error _ =>
+                  dbg_trace s!"warning: type conversion failed for {name}, defaulting to Integer"
```
MikaelMayer, line 156: The old code silently fell back to `.Integer` on type conversion failure. The new code adds `dbg_trace` warnings, which is an improvement for debuggability. Just noting this is a deliberate behavioral change (not just a code move) — the `dbg_trace` output will appear in the Lean server log during evaluation.

PR #553, opened 2026-03-12
Change Laurel block grammar

keyboardDrummer:
### Changes
1. Add SemicolonSepBy to DDM
1. Use a separate DDM command for each top level Laurel declaration, to reduce stack usage during parsing
1. Change the Laurel block grammar so `;` occurs only between consecutive statements, not as part of statements. This makes the grammar more regular and reduces hard to understand parse errors
1. Reduce priority of the "else" part of the if-then-else parser, to enable parsing for example `else foo(3)` as `else (foo(3))` instead of `(else foo)(3)`
1. Remove unused `CorePrelude.lean`

### Testing
1. Made small modifications to existing tests to test this new construct.
1. Removed `set_option maxRecDepth 10000` from some files since this is no longer needed
1. Updated existing Laurel tests to use the new grammar
1. Added test-case to T3_ControlFlow
1. No test needed


Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
```diff
@@ -7,7 +7,7 @@
 -- Laurel dialect definition, loaded from LaurelGrammar.st
 -- NOTE: Changes to LaurelGrammar.st are not automatically tracked by the build system.
 -- Update this file (e.g. this comment) to trigger a recompile after modifying LaurelGrammar.st.
--- Last grammar change: require semicolon after procedure/function definitions.
+-- Last grammar change: 1314
```
shigoel: Where does `1314` come from? Is this a checksum or something of some form?

shigoel: Worth adding a comment.

keyboardDrummer: Its my local clock. This entire line doesn't need to be checked in but it is useful for locally resetting the cache.

PR #554, opened 2026-03-12
Various Laurel improvements that enable defining the Python runtime/prelude in Laurel

keyboardDrummer:
### Changes
1. Support for mutually referencing datatypes in Laurel
2. Support for referencing unsafe datatype destructors
3. Support for a single trigger with a single component for quantifiers
4. Grammar for multiple requires clauses
5. Improved resolution support related to instance procedures. Better handling of calls to them and of resolving their bodies.
6. Support for decimal literals and the `real` type
7. Fix a bug in Core, so it correctly translates division of reals
8. When translating Laurel to Core, unused expressions are not silently dropped but instead assigned to a fresh variable
9. Enable optionally turning on resolution errors for Laurel, and turn it on unless Python is the front-end
10. Remove unused support for opaque types. This is not a feature Laurel should have.

### Testing
1. Added a test for datatypes with a test-case for this
2. Added a test for datatypes with a test-case for this
3. Added tests for triggers in quantifiers
4. Extended T6_Preconditions test
5. No test yet. Will add dedicated tests for instance calls when we fully support that feature
6. Added test T14_Decimals
7. Updated SMTEncoderTests with a new case
8. Debug feature. No test added
9. Absence of resolution errors is tested in all Laurel tests. We will add a dedicated test for presence of them in a future PR
10. There were no tests for opaque types

### DIFF
--- a/Strata/Languages/Core/StatementEval.lean
+++ b/Strata/Languages/Core/StatementEval.lean
@@ -476,7 +476,8 @@ def evalAuxGo (steps : Nat) (old_var_subst : SubstMap) (Ewn : EnvWithNext) (ss :
                   let deadPathConds := Ewn.env.pathConditions.push [(deadLabel, LExpr.false ())]
                   let deferred := createUnreachableCoverObligations deadPathConds ss_f_covers
                   let deferred := deferred ++ createUnreachableAssertObligations deadPathConds ss_f_asserts
-                  [{ Ewn with env.deferred := Ewn.env.deferred ++ deferred }]
+                  -- No need to retain older deferred obligations in a dead branch.
+                  [{ Ewn with env.deferred := deferred }]
                 else
                   []
               let Ewns_t :=
--- a/Strata/Transform/LoopElim.lean
+++ b/Strata/Transform/LoopElim.lean
@@ -14,12 +14,55 @@ public section
 
 /-! ## Loop elimination
 
-This transformation recursively removes loops from a statement, resulting in a
-new, acyclic statement. If a loop invariant is present, it adds checks that the
-invariant is established on entry and re-established at the end of each
-iteration. If a loop invariant is not present, this transformation isn't very
-useful.
+This transformation converts a loop into an acyclic passive statement suitable
+for symbolic verification. A loop invariant `I` is required; without one the
+transformation produces no useful verification conditions.
 
+### Role of the invariant
+
+Unlike the classical Hoare triple `{P} while G { S } {Q}` — which keeps the
+loop pre-condition `P`, inductive invariant `I`, and post-condition `Q`
+separate — this encoding folds all three into `I`. The user must choose `I`
+strong enough so that:
+
+- the pre-loop state satisfies `I` (checked at entry), and
+- `I ∧ ¬G` implies some desired post-condition `Q` (checked after the loop).
+
+### Passive encoding recipe
+
+Let `M` be the set of variables modified by the loop body.
+
+```
+assert(I);              -- VC1: I holds at loop entry (unconditional)
+assume(I);              -- make I available on the 0-iteration path
+                        --   (assert-then-assume; intentional exception to the
+                        --    usual assert/assume separation, needed so that
+                        --    assert(Q) can use I when G is false at entry)
+if (G) {
+  havoc(M);             -- non-deterministically pick a mid-loop state
+  assume(G);            -- guard holds at this state (live iteration)
+  assume(I);            -- invariant holds at this state
+  S;                    -- execute one iteration of the body
+  assert(I);            -- VC2: I is maintained by S
+
+  havoc(M);             -- non-deterministically pick an exit state
+  assume(¬G);           -- guard is false at exit (loop has terminated)
+  assume(I);            -- invariant holds at exit (by induction from VC1+VC2)
+}
+assert(Q);              -- checked with I ∧ ¬G available on both paths:
+                        --   G=false path: M is pre-loop state, I from assume above
+                        --   G=true  path: M is arbitrary exit state, I ∧ ¬G from then-branch
+```
+
+Note: the `if(G)` then-branch does double duty — VC2 check and exit-state model —
+so the mid-loop path conditions (`G(M_iter)`, `I(M_iter)`) are present alongside the
+exit-state facts when `Q` is checked. These linger as irrelevant assumptions,
+indeed.
+
+Note: `assume(I)` after VC1 is not strictly necessary. If VC1 passes, then it
+means `I` is derivable for a backend solver and we could skip adding it to the
+path conditions. However, we choose to keep this assumption for a higher-quality
+encoding.
 -/
 
 mutual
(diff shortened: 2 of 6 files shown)


Strata/DL/SMT/Op.lean
```diff
@@ -83,7 +83,8 @@ inductive Op.Num : Type where
   | sub
   | add
   | mul
-  | div
+  | div   -- integer division (`div` in SMT-LIB)
+  | rdiv  -- real division (`/` in SMT-LIB)
```
shigoel, line 87: TIL.

Strata/Languages/Laurel/DatatypeGrouping.lean
```diff
@@ -0,0 +1,117 @@
...
+  | ⟨.Intersection ts, _⟩ => ts.flatMap collectTypeRefs
+  | _ => []
+
+/-- Get all datatype names that a `DatatypeDefinition` references in its constructor args. -/
+def datatypeRefs (dt : DatatypeDefinition) : List String :=
+  dt.constructors.flatMap fun c => c.args.flatMap fun p => collectTypeRefs p.type
+
+/-! ### Tarjan's SCC -/
+
+private structure TarjanState where
```
shigoel: Did you consider using Strata/DDM/Util/Graph/Tarjan.lean?

Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -24,13 +24,14 @@ No inference is performed — all types are determined by annotations on paramet
...
-  match expr with
+  match _: expr with
   | WithMetadata.mk val md =>
-  match val with
+  match _: val with
   -- Literals
   | .LiteralInt _ => ⟨ .TInt, md ⟩
   | .LiteralBool _ => ⟨ .TBool, md ⟩
   | .LiteralString _ => ⟨ .TString, md ⟩
+  | .LiteralDecimal _ => ⟨ .TFloat64, md ⟩
```
shigoel: Should this be real instead of float?

Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -70,22 +79,22 @@ def computeExprType (model : SemanticModel) (expr : StmtExprMd) : HighTypeMd :=
...
   | .Fresh _ => ⟨ .TBool, md ⟩
   -- Proof related
   | .ProveBy v _ => computeExprType model v
-  | .ContractOf _ _ => panic "Not supported"
+  | .ContractOf _ _ => panic "ContractOf Not supported"
   -- Special
-  | .Abstract => panic "Not supported"
-  | .All => panic "Not supported"
-  | .Hole => panic "Not supported"
+  | .Abstract => panic "Abstract Not supported"
```
shigoel, line 99: I was surprised that the panic linter that @tautschnig wrote didn't catch these new ones, but then realized that the linter checks for `panic!`. In any case, It'd be good to very soon use throw or similar here, as discussed offline. OK to do in a separate PR.

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -96,10 +99,11 @@ def AstNode.getType (node: AstNode): Option HighTypeMd := match node with
  | .field _ f => f.type
  | .datatypeConstructor type _ => some ⟨ .UserDefined type, default ⟩
  | .constant c => c.type
+ | .quantifierVar _ type => type
  | .unresolved =>
     -- The Python through Laurel pipeline does not resolve yet
     some ⟨ .UserDefined "dummyName", default ⟩
- | _ => panic! s!"getType called on {repr node}"
+ | _ => dbg_trace s!"SOUND BUG: getType called on {repr node}"; none
```
shigoel, line 106: "Sound" bug? Also, curious why you have `dbg_trace` here now. Is it really safe to go on here? Is this case unreachable?

keyboardDrummer: This case is a bit strange because the meaning of returning `none` is unclear. If we interpret it as `.error` then returning none here is sound (does not lead to unfound claims) but is still a bug. Code will change in a follow-up.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -149,30 +155,35 @@ def translateExpr (expr : StmtExprMd)
...
     | .Eq => return .eq () re1 re2
     | .Neq => return .app () boolNotOp (.eq () re1 re2)
     | .And => return binOp boolAndOp
     | .Or => return binOp boolOrOp
     | .Implies => return binOp boolImpliesOp
-    | .Add => return binOp intAddOp
-    | .Sub => return binOp intSubOp
-    | .Mul => return binOp intMulOp
-    | .Div => return binOp intSafeDivOp
+    | .Add => return binOp (if isReal then realAddOp else intAddOp)
```
andrewmwells-amazon, line 176: [Nit]: Personally, I'd prefer if the LangToLaurel pass could choose which it wanted, but I'm ok changing that later if it is indeed useful.

keyboardDrummer: Can't the LangToLaurel pass choose which type it uses and get the right operator based on that?

Strata/Languages/Laurel/LaurelEval.lean
```diff
@@ -176,6 +176,7 @@ partial def eval (expr : StmtExpr) : Eval TypedValue :=
   | StmtExpr.LiteralBool b => pure <| TypedValue.mk (Value.VBool b) HighType.TBool
   | StmtExpr.LiteralInt i => pure <| TypedValue.mk (Value.VInt i) HighType.TInt
   | StmtExpr.LiteralString s => pure <| TypedValue.mk (Value.VString s) HighType.TString
+  | StmtExpr.LiteralDecimal d => pure <| TypedValue.mk (Value.VFloat64 (Float.ofScientific d.mantissa.natAbs (d.mantissa < 0) d.exponent.natAbs)) HighType.TFloat64
```
fabiomadge, line 179: `LaurelEval.lean` isn't imported or tested anywhere — is it still in use? If it is, this should be `TReal` not `TFloat64` to match the rest of the PR. If it isn't, it might be worth removing the file to avoid maintaining dead code.

StrataMain.lean
```diff
@@ -607,7 +601,7 @@ def pyTranslateLaurelCommand : Command where
     | .error e =>
       exitFailure s!"Python to Laurel translation failed: {e}"
     | .ok (laurelProgram, _) =>
-      match Strata.Laurel.translate laurelProgram with
+      match Strata.Laurel.translate { } laurelProgram with
```
fabiomadge, line 604: This is a Python pipeline (`pythonToLaurel`), but it uses the default `emitResolutionErrors := true`. The other Python pipelines (`buildPySpecPrelude` at line 371 and `pyAnalyzeLaurelCommand` at line 453) use `{ emitResolutionErrors := false }`. Same issue at line 629 in `pyAnalyzeLaurelToGotoCommand`. Should these also be `{ emitResolutionErrors := false }`?

Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
```diff
@@ -557,14 +586,11 @@ def parseTopLevel (arg : Arg) : TransM (Option Procedure × Option TypeDefinitio
   | q`Laurel.datatypeCommand, #[datatypeArg] =>
     let typeDef ← parseDatatype datatypeArg
     return (none, some typeDef)
-  | q`Laurel.opaqueTypeCommand, #[opaqueTypeArg] =>
-    let typeDef ← parseOpaqueType opaqueTypeArg
-    return (none, some typeDef)
   | q`Laurel.constrainedTypeCommand, #[ctArg] =>
```
fabiomadge, line 589: `parseOpaqueType` (line 551) is now dead code — its only caller was this removed case. Should be deleted.

PR #555, opened 2026-03-12
Remove Python specific hack from Laurel

keyboardDrummer:
### Changes
Remove Python specific hack in SemanticModel.isFunction

### Testing
Covered by existing Python tests


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1346,8 +1351,22 @@ def pythonToLaurel (prelude: Core.Program)
...
+      inputs := [],
+      outputs := [],
+      preconditions := [],
+      determinism := .deterministic none, --TODO: need to set reads
+      decreases := none,
+      body := .External
+      md := default
+      isFunctional := false
+      }
+    )
```
joehendrix, line 1414: I think I am missing some context, but I would have expected procedures to have inputs and output parameters.  It's also think the TODO should be expanded a bit to explain what remains to be done for reads (or the expected consequences of setting this to "none").

keyboardDrummer: Added a comment.

PR #563, opened 2026-03-12
Fix bugs in precondition elimination

joscoh:
*Issue #, if available:*

*Description of changes:*
Ensure preconditions in guard of `.ite` are encoded
Ensure preconditions arising in loop guards appear at end of loop as well 
Changes order so assertions are emitted in dependency order 

Includes tests for new cases

### DIFF
--- a/.github/workflows/cbmc.yml
+++ b/.github/workflows/cbmc.yml
@@ -45,22 +45,41 @@ jobs:
               unzip -q "${ARCHIVE_NAME}.zip"
               chmod +x "${ARCHIVE_NAME}/bin/z3"
               echo "$GITHUB_WORKSPACE/${ARCHIVE_NAME}/bin/" >> $GITHUB_PATH
+      - name: Prepare ccache
+        uses: actions/cache@v4
+        with:
+          save-always: true
+          path: .ccache
+          key: cbmc-${{ runner.os }}-${{ runner.arch }}-cbmc-${{ github.sha }}
+          restore-keys: |
+            cbmc-${{ runner.os }}-${{ runner.arch }}-cbmc
       - name: Build CBMC from source (with string support patch)
         shell: bash
         run: |
           sudo apt-get -qq update
-          sudo apt-get -qq install -y cmake ninja-build flex bison
+          sudo apt-get -qq install -y cmake ninja-build flex bison ccache
           git clone --depth 1 --branch cbmc-6.8.0 https://github.com/diffblue/cbmc.git cbmc-src
           cd cbmc-src
           git apply "$GITHUB_WORKSPACE/StrataTest/Languages/Python/cbmc-string-support.patch"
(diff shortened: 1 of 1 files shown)


Strata/DL/Lambda/Preconditions.lean
```diff
@@ -126,7 +126,7 @@ where
         let fObs := go F f ((md, .app md (@boolNotFunc T).opExpr c) :: implications)
         cObs ++ tObs ++ fObs
       | .eq _ e1 e2 => go F e1 implications ++ go F e2 implications
-    callObligations ++ subObligations
+    subObligations ++ callObligations
```
shigoel, line 131: oof! Maybe worth a comment why this order matters?

joscoh: I can leave a comment in the code. It's not a bug exactly, but if we have e.g. `x / (y / z)`, previously this outputted (in order) `assert (y/z != 0); assert (z != 0)`, when the order should really be the other way around. This was not directly related to the bugs found by Aaron but I thought it was worth fixing.

Strata/Transform/PrecondElim.lean
```diff
@@ -224,20 +224,23 @@ def transformStmt (s : Statement)
...
+    let condAsserts := collectPrecondAsserts F c "ite_cond" md
     let savedF ← getFactory
     let (changed, thenb') ← transformStmts thenb
     setFactory savedF
     let (changed', elseb') ← transformStmts elseb
     setFactory savedF
-    return (changed || changed', [.ite c thenb' elseb' md])
+    return (changed || changed' || !condAsserts.isEmpty,
+      condAsserts ++ [.ite c thenb' elseb' md])
   | .loop guard measure invariant body md => do
```
shigoel, line 235: Wondering why we don't need to gather function preconds from the loop measure (right now, doesn't matter since we don't do termination checking, but it will soon).

joscoh: Good call, we should be doing that as well. Will add

joscoh: Actually, that will depend on how the termination checking is handled. Will the measure be checked at the end of each loop iteration as well?

shigoel: Yes, it'd be done alongside the inductive step (i.e., the loop invariant is maintained).

I'm also okay to leave this out of scope for this PR, if you're okay with throwing an exception if the measure is non-none.

joscoh: It's not hard to add, but since we can't yet write a decreases clause for loops in the concrete syntax, I just threw an error on non-None measures for the moment. It can be changed once termination checking is added.

shigoel: Yes, indeed.

PR #566, opened 2026-03-12
feat: conservatively abstract unsupported Python constructs

MikaelMayer:
## What

Instead of failing on unsupported Python constructs, translate them to sound over-approximations so the verifier can still run.

## How

| Construct | Abstraction |
|-----------|-------------|
| Unsupported expressions | `Hole` (fresh uninterpreted value) |
| Unsupported binary operators | `Hole` |
| Compound LHS assignment (`a[i]=v`, `(a,b)=v`) | Havoc all in-scope variables (aliasing may affect any of them) |
| Tuple unpacking in `for` (`for a,b in ...`) | Havoc all target names |
| `with` statement | Body translated; `as`-targets declared as `Hole` |
| Unsupported statements | Havoc all in-scope variables + `assert *` |

Unsupported statements and compound LHS assignments are recorded in `TranslationContext.abstractedStatements` and emitted as warnings by `pyAnalyzeLaurel`.

## Tests

Added `test_with_statement.py`.


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -674,119 +479,120 @@ Translate Python statements to Laurel StmtExpr nodes.
...
+    | .Attribute _ _ _ _ =>
+      -- Field assignment: obj.field = expr or self.field = expr
+      let valueExpr ← translateExpr ctx value
+      let targetExpr ← translateExpr ctx targets.val[0]!  -- This will handle self.field via translateExpr
+      let assignStmt := mkStmtExprMdWithLoc (StmtExpr.Assign [targetExpr] valueExpr) md
+      return (ctx, assignStmt)
+    | _ =>
+      -- Compound LHS (e.g. a[i] = v, (a, b) = v) — havoc all Name targets
+      -- This is a sound over-approximation: we lose the assignment but keep
+      -- the program structure.
```
olivier-aws: I'm not sure this is sound in Python, due to aliasing. This Python script prints 999:
```python
a = [1, 2, 3]
b = a
b[0] = 999
print(a[0])
```
Here you would only havoc `b`, not `a`

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -944,25 +744,94 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+    let targetExpr ← translateExpr ctx target
+    let valueExpr ← translateExpr ctx value
+    -- Translate the operator; fall back to Hole if unsupported
+    let rhs ← match op with
+      | .Add _    => .ok (mkStmtExprMd (StmtExpr.PrimitiveOp Operation.Add [targetExpr, valueExpr]))
+      | .Sub _    => .ok (mkStmtExprMd (StmtExpr.PrimitiveOp Operation.Sub [targetExpr, valueExpr]))
+      | .Mult _   => .ok (mkStmtExprMd (StmtExpr.PrimitiveOp Operation.Mul [targetExpr, valueExpr]))
+      | _         => .ok (mkStmtExprMd .Hole)
+    let assignStmt := mkStmtExprMdWithLoc (StmtExpr.Assign [targetExpr] rhs) md
+    return (ctx, assignStmt)
```
olivier-aws: Can this be in a separate PR? This is a correct and precise handling of compound assignment, so conflicts with the name of the PR (abstract unsupported Python constructs)

StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
```diff
@@ -0,0 +1,13 @@
+
+==== Verification Results ====
+datetime_now_ensures_0: ✅ pass (at byte 7134)
+datetime_utcnow_ensures_0: ✅ pass (at byte 7372)
+ensures_str_strp_reverse: ✅ pass (at byte 8766)
+assert_name_is_foo: ✅ pass (at byte 11081)
+assert_opt_name_none_or_str: ✅ pass (at byte 11131)
+assert_opt_name_none_or_bar: ✅ pass (at byte 11278)
+ensures_maybe_except_none: ✅ pass (at byte 10984)
```
olivier-aws: side remark: would be great to find a way to now show these VC that come from the prelude, it would make our expected files smaller and more representative of what the test actually is for

olivier-aws (commented):
As a general comment, I'm not a big fan of hiding away what we don't support. There are many risks of unsoundness with this. At the very least, we need a message in the log that the translation encountered an unsupported expression/statement and over-approximated it.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Indeed good catch. The answer there is that we should havoc not just all name targets but all names in scope.

I hadn't reviewed that part yet but I'm happy you did!

StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
MikaelMayer: Oh indeed.

StrataTest/Languages/Python/tests/test_break_continue.py
```diff
@@ -0,0 +1,24 @@
+def test_while_break() -> None:
+    count: int = 0
+    i: int = 0
+    while i < 10:
+        if i == 3:
+            break
+        i = i + 1
+        count = count + 1
+    assert count >= 0, "count non-negative after break"
```
MikaelMayer: We need to add an assertion here that i == 3 or i >= 10 to catch the ability to prove or not useful results

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: On a second thought, if we stored dictionaries via an indirection to the heap, then no target would be havoc-ed, we would here only havoc the heap to indicate something changed, so we would still be able to prove a[0] == b[0]
What is the status of the heap in Laurel? Is it fully supported?

ssomayyajula: AFAIK Laurel does allow arbitrary StmtExprs to occur on the LHS of an assignment, it's just that Laurel->Core only supports assignmets to identifiers https://github.com/strata-org/Strata/blob/main/Strata/Languages/Laurel/LaurelToCoreTranslator.lean#L362 (that is, Laurel supports heap-parameterized procedures + the creation of objects of composite type on the heap, but not assigning to their fields after the fact)

keyboardDrummer: > (that is, Laurel supports heap-parameterized procedures + the creation of objects of composite type on the heap, but not assigning to their fields after the fact)

What? You can assign to fields whenever you like: https://github.com/strata-org/Strata/blob/main/StrataTest/Languages/Laurel/Examples/Objects/T1_MutableFields.lean#L40

I would prefer to change the field selector syntax from `#` to `.` but I ran into some technical issues. Will resolve them at some point.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -591,7 +610,7 @@ partial def translateVarKwargs (ctx : TranslationContext) (kwords : List (Python
     match name.val with
     | some _ => throw (.internalError s!"Keyword arg should be a Dict")
     | none =>
-        let expr ← translateExpr ctx expr
+        let (expr, _) ← (translateExpr ctx expr).run'
```
joehendrix: Is this dropping "warnings"?

MikaelMayer: Yes and it should not. Thanks for spotting.

MikaelMayer: Fixed: `translateDictKWords`, `translateVarKwargs`, `translateKwargs`, and `translateCall` now use `ExprM` so warnings from nested expression translations are propagated rather than dropped. `translateStmt` also uses `translateExprW` for condition/value expressions to merge warnings into `ctx.abstractedStatements`.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: AugAssign has been moved to PR #573 which is dedicated to augmented assignment support.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Agreed. The unsupported statement catch-all now records the abstraction in `ctx.abstractedStatements` and emits a warning via `pyAnalyzeLaurel`. Unsupported expressions also record warnings via the `ExprM` monad.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -762,7 +788,7 @@ partial def translateCall (ctx : TranslationContext)
   match f with
   | .Name  _ _ _ =>  return mkStmtExprMd (StmtExpr.StaticCall funcName (trans_args ++ trans_kwords_exprs))
   | .Attribute _ val attr _ =>
-      let target_trans ← translateExpr ctx val
+      let (target_trans, _) ← (translateExpr ctx val).run'
```
MikaelMayer: `translateCall` is already inside the `mutual` block and returns `ExprM`, so calling `.run'` here exits the monad and silently drops any warnings from translating `val`. You can just use `translateExpr` directly:

```suggestion
      let target_trans ← translateExpr ctx val
```

I verified this compiles.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -793,7 +830,7 @@ partial def translateAssign  (ctx : TranslationContext)
                              (rhs: Python.expr SourceRange)
                              (md: Imperative.MetaData Core.Expression)
                     : Except TranslationError (TranslationContext × List StmtExprMd) := do
-  let rhs_trans ←  translateExpr ctx rhs
+  let rhs_trans ←  translateExprVal ctx rhs
```
MikaelMayer: `translateExprVal` discards warnings. This means if the RHS of an assignment contains an unsupported expression that gets abstracted to `Hole`, the warning is silently lost. The same applies to lines 900, 901, and 917.

Since `translateAssign` already returns a `TranslationContext`, it could use `translateExprW` and propagate warnings through the context, consistent with how `translateStmt` handles it.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -121,6 +123,20 @@ def TranslationError.toString : TranslationError → String
...
 
+/-- Warnings about abstracted constructs (location × description) -/
+abbrev TranslationWarnings := List (SourceRange × String)
+
+/-- Monad for expression translation: can throw errors and accumulate warnings -/
+abbrev ExprM := StateT TranslationWarnings (Except TranslationError)
+
+/-- Record an abstraction warning -/
+def ExprM.warn (loc : SourceRange) (kind : String) : ExprM Unit :=
+  modify (· ++ [(loc, kind)])
```
MikaelMayer: Minor: `modify (· ++ [(loc, kind)])` is O(n) per call since it appends to the end of a list. Not a problem at current scale, but if this ever becomes hot, consider building in reverse and reversing at the end (consistent with the project convention).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1160,7 +1204,16 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     | some lbl => return (ctx, [mkStmtExprMdWithLoc (StmtExpr.Exit lbl) md])
     | none => return (ctx, [mkStmtExprMdWithLoc (StmtExpr.Assert (mkStmtExprMd .Hole)) md])
 
-  | _ => throw (.unsupportedConstruct "Statement type not yet supported" (toString (repr s)))
+  -- Unsupported statement — record the abstraction, havoc all in-scope
+  -- variables (sound: the statement could modify anything), and emit
+  -- `assert *` so the verifier flags the location.
+  | _ =>
+    let stmtName := ((toString (repr s)).takeWhile (· != ' ')).toString
```
MikaelMayer: Calling `repr` on the entire statement AST just to extract the constructor name is expensive. Consider pattern-matching on the statement constructors you expect to reach here (e.g., `Global`, `Nonlocal`, `Import`, `Delete`, etc.) and returning a string literal for each, with a catch-all like `"unknown"`. This would be both cheaper and produce cleaner warning messages.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Fixed: using `translateExpr ctx val` directly since `translateCall` is already in `ExprM` context.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Fixed: `translateAssign` now uses `translateExprW` for the RHS, subscript target/slices, and attribute LHS, propagating warnings into `ctx.abstractedStatements`.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Fixed: `ExprM.warn` now prepends (O(1)) and `ExprM.run'` reverses at the boundary, consistent with the project convention.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Fixed: pattern-matching on the expected constructors (`Global`, `Nonlocal`, `Import`, `ImportFrom`, `Delete`, `With`) with a catch-all `"unknown"`.

PR #568, opened 2026-03-12
Define Python Prelude in Laurel instead of Core

keyboardDrummer:
### Changes
1. Support datatypes in field types.
2. Remove the override `datatype Box` hack that was used in Python. This is possible because of the previous change.
3. In the Python through Laurel pipeline, introduce `PythonRuntimeLaurelPart` and move a small part of `PythonLaurelCorePrelude` to it. In the final generated Core, declarations from `PythonRuntimeLaurelPart.lean` come before `PythonLaurelCorePrelude`. I will rename `PythonLaurelCorePrelude` to `PythonRuntimeCorePart` in a separate PR, to avoid conflicts.
4. Add `--update` option to run_py_analyze.sh to update the test expect files (the bytecode offsets in them change often)

#### Caveats
- Because DDM parsing also does a form of resolution, `PythonRuntimeCorePart.lean` needs a lot of type definitions that are already provided in `PythonPreludeInLaurel.lean`. Ideally we'd find a way so that's no longer necessary, using something like `#strata_parsing_only`. @joehendrix is that something we could add? I looked into adding that but it wasn't so simple.

- Axioms in `PythonLaurelCorePrelude` can't be moved to `PythonRuntimeLaurelPart` because Laurel does not and I think should not support axioms. A long term solution would be that the Python pipeline stops relying on axioms.

### Future work
- Turn on reporting of resolution errors in the Python through Laurel pipeline. This requires moving most definitions from PythonLaurelCorePrelude to PythonRuntimeLaurelPart.

### Tested
1. Updated T1_MutableFields test with a new test-case.
2. Refactoring, no tests needed
3. Updated Python expect files for Laurel because the reported locations are different
4. Manually used the --update option for `run_py_analyze.sh`

### DIFF
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -680,6 +680,17 @@ def pyAnalyzeToGotoCommand : Command where
         IO.FS.writeFile gotoFile goto.pretty
         IO.println s!"Written {symTabFile} and {gotoFile}"
 
+def pyTranslateLaurelCommand : Command where
+  name := "pyTranslateLaurel"
+  args := [ "file" ]
+  help := "Translate a Strata Python Ion file through Laurel to Strata Core. Write results to stdout."
+  callback := fun v _ => do
+    let pgm ← readPythonStrata v[0]
+    let cmds := Strata.toPyCommands pgm.commands
+    assert! cmds.size == 1
+    let mut coreProgram ← translatePythonToCore false v[0] Array.empty Array.empty
+    IO.print coreProgram
+
 def pyAnalyzeLaurelToGotoCommand : Command where
   name := "pyAnalyzeLaurelToGoto"
   args := [ "file" ]


Strata/Languages/Laurel/CoreDefinitionsForLaurel.lean
```diff
@@ -25,6 +25,8 @@ def coreDefinitionsForLaurelDDM :=
 #strata
 program Laurel;
 
+datatype Float64IsNotSupportedYet {}
```
keyboardDrummer, line 28: Moved from HeapParameterizationConstants, because it fits here better. Unrelated to the rest of the PR

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -5,25 +5,41 @@
 -/
 module
 
+-- TODO: rename file to PythonRuntimeCorePart
```
keyboardDrummer, line 8: Will do the rename in a follow-up

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1042,7 +1042,7 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
   | _ => throw (.internalError "Expected FunctionDef")
 
 /-- Translate Python function to Laurel Procedure -/
```
keyboardDrummer, line 1044: Contains a few changes to pass more source locations to Laurel

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -155,28 +155,67 @@ structure TransformState where
...
   | .TReal => "BoxReal"
-  | .UserDefined _ => "BoxComposite"
-  | _ => "BoxInt"  -- fallback
+  | .UserDefined name =>
+      if isDatatype model name then s!"Box..{name.text}"
+      else "BoxComposite"
+  | _ => dbg_trace "BUG, boxConstructorName bad type"; "boxConstructorNameError"
+
+/-- Build the DatatypeConstructor for a Box variant from a HighType, for datatype generation -/
+private def boxConstructorDef (model : SemanticModel) (ty : HighType) : Option DatatypeConstructor :=
```
shigoel, line 198: No .TReal support here?

StrataMain.lean
```diff
@@ -470,7 +522,14 @@ def pyAnalyzeLaurelCommand : Command where
           if !collisions.isEmpty then
             let names := ", ".intercalate (collisions.map (·.name))
             exitFailure s!"Core name collision between program and prelude: {names}"
-          let coreProgram := {decls := pyPrelude.decls ++ programDecls }
+          let (preludeDecls, userDecls) := programDecls.span (fun d => !(toString d.name).contains "END_MARKER")
```
shigoel: The `contains` check is fragile — any user declaration whose name happens to contain the substring `"END_MARKER"` (e.g. `MY_END_MARKER`) would silently break the split. Prefer an exact match (`== "END_MARKER"`) to be consistent with how the `CoreOnlyDelimiter` sentinel is handled in `PythonLaurelCorePrelude.lean`.

StrataMain.lean
```diff
@@ -370,22 +378,61 @@ def buildPySpecPrelude (pyspecPaths : Array String) : IO PySpecPrelude := do
...
-      for d in pyspecDecls do
+      for d in coreSpec.decls do
         for n in Core.Decl.names d do
           if existingNames.contains n.name then
             exitFailure s!"Core name collision in PySpec {ionPath}: {n.name}"
           existingNames := existingNames.insert n.name
-      preludeDecls := preludeDecls ++ pyspecDecls.toArray
-  let pyPrelude : Core.Program := { decls := preludeDecls.toList }
-  return { corePrelude := pyPrelude, overloads := allOverloads }
+      preludeInCoreDecls := preludeInCoreDecls ++ coreSpec.decls
```
shigoel: Potential duplication: `translate` is called on the whole accumulated `preludeInLaurel` (which includes all previously-processed PySpec programs), so `coreSpec.decls` from iteration N contains all declarations from iterations 1..N-1 as well. Those get appended to `preludeInCoreDecls` again, duplicating earlier entries. The collision check by name prevents outright conflicts, but silent duplicates will still reach the final `Core.Program`.

StrataMain.lean
```diff
@@ -333,22 +335,28 @@ def pyAnalyzeCommand : Command where
...
-  -- We no longer need to strip it from translate output.
-  let laurelPreludeSize := 0
-  let mut preludeDecls : Array Core.Decl :=
-    Strata.Python.Core.PythonLaurelPrelude.decls.toArray
+  let mut preludeInLaurel : Strata.Laurel.Program := Strata.Python.pythonRuntimeLaurelPart
+  let mut preludeInCoreDecls : List Core.Decl := Strata.Python.coreOnlyFromRuntimeCorePart
   let mut existingNames : Std.HashSet String :=
-    preludeDecls.foldl (init := {}) fun s d =>
-      (Core.Decl.names d).foldl (init := s) fun s n => s.insert n.name
+    preludeInLaurel.staticProcedures.foldl (init := {}) fun s d =>
```
shigoel: `existingNames` is seeded only from `preludeInLaurel.staticProcedures`. Types, constants, and static fields defined in the Laurel prelude are not tracked, so name collisions between PySpec files for those declaration kinds will go undetected.

StrataMain.lean
```diff
@@ -446,21 +493,26 @@ def pyAnalyzeLaurelCommand : Command where
     match laurelPgm with
       | .error e =>
         exitFailure s!"Python to Laurel translation failed: {e}"
-      | .ok (laurelProgram, ctx)  =>
+      | .ok (laurelProgram, ctx) =>
+        -- Combine the Laurel prelude declarations with the translated program
+        let pythonRuntimeLaurelPart := Strata.Python.pythonRuntimeLaurelPart
```
shigoel: This local binding `pythonRuntimeLaurelPart` shadows the module-level name but is never read — the prelude is already available via `pySpecResult.preludeInLaurel`. Dead code.

StrataMain.lean
```diff
@@ -599,13 +658,13 @@ def pyTranslateLaurelCommand : Command where
     let pgm ← readPythonStrata v[0]
     let cmds := Strata.toPyCommands pgm.commands
     assert! cmds.size == 1
-    let prelude := Strata.Python.Core.PythonLaurelPrelude
+    let prelude := Strata.Python.pythonRuntimeCorePart -- TODO incorrect
```
shigoel: Incorrect in what way? Same applies to comment below.

StrataMain.lean
keyboardDrummer: > any user declaration whose name happens to contain the substring "END_MARKER" (e.g. MY_END_MARKER) would silently break the split

I don't think so, since ours always comes first so it will break the span first. Replaced the contains though

Strata/Languages/Laurel/HeapParameterization.lean
keyboardDrummer: Yikes. Added a test for real fields as well

StrataMain.lean
keyboardDrummer: It's missing the complex logic that's in pyAnalyzeLaurel. I think this command was already broken. It doesn't seem to be tested. I will see if we can delete it in this PR.

Update: I've deleted one of these and the other I've fixed.

PR #569, opened 2026-03-13
Add assertion descriptions in Laurel

keyboardDrummer:
### Changes
Allow adding descriptions to assertions, in this way:
```
assert y == 0 summary "divisor is zero"
```

which then change the reported error messages. In the above case, if it fails with unsat, it would report `divisor is zero does not hold`

### Testing
Added a test file

### DIFF
--- a/.github/workflows/cbmc.yml
+++ b/.github/workflows/cbmc.yml
@@ -45,22 +45,41 @@ jobs:
               unzip -q "${ARCHIVE_NAME}.zip"
               chmod +x "${ARCHIVE_NAME}/bin/z3"
               echo "$GITHUB_WORKSPACE/${ARCHIVE_NAME}/bin/" >> $GITHUB_PATH
+      - name: Prepare ccache
+        uses: actions/cache@v4
+        with:
+          save-always: true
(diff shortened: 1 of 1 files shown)


Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
```diff
@@ -8,7 +8,7 @@ module
 -- Laurel dialect definition, loaded from LaurelGrammar.st
 -- NOTE: Changes to LaurelGrammar.st are not automatically tracked by the build system.
 -- Update this file (e.g. this comment) to trigger a recompile after modifying LaurelGrammar.st.
--- Last grammar change: added externalBody op for external procedure/function bodies.
+-- Last grammar change: added externalBody op for external procedure/function bodies.2
```
MikaelMayer: Do you still need these hacks? I remember fixing something that was hopefully taking care of that a few weeks ago.

keyboardDrummer: I use them for local development

joscoh (approved):

Issue #565, opened 2026-03-12
Track Python→Laurel translation coverage


MikaelMayer:
This issue tracks which Python constructs are fully supported vs conservatively abstracted in the Python→Laurel translator.

## Conservatively abstracted (sound but imprecise)

| Construct | Abstraction | Semantics | Status |
|-----------|-------------|-----------|--------|
| Unsupported binary operators (e.g. `**`) | `Hole` | Fresh uninterpreted value | ⚠️ Needs full support |
| Unsupported expressions | `Hole` | Fresh uninterpreted value | ⚠️ Needs full support |
| Compound LHS assignment (`a[i] = v`, `(a,b) = v`) | Havoc named targets | Targets assigned fresh uninterpreted values | ⚠️ Needs full support |
| Tuple unpacking in `for` (`for a, b in ...`) | Havoc all target names | Targets assigned fresh uninterpreted values | ⚠️ Needs full support |
| `with` statement | Body translated, context manager ignored | Sound: body still verified | ⚠️ Needs full support |
| Augmented assignment (`+=`, `-=`, etc.) | `x = x op rhs` | Precise for supported ops; `Hole` for unsupported | ✅ Fully supported |
| `break` / `continue` | `Exit` labelled block | Correct control flow | ✅ Fully supported |
| Unsupported statements | `assert *` | Assert a fresh nondeterministic bool — verifier may flag but not a guaranteed failure | ⚠️ Needs full support |

## Fully supported

- Simple variable assignment
- Field assignment (`self.field = ...`)
- Annotated assignment
- `if`/`elif`/`else`
- `while` loop (including `break`/`continue`)
- `for` loop (simple and tuple targets, including `break`/`continue`)
- `return`
- `assert`
- `try`/`except`
- `raise`
- Function calls, method calls
- Arithmetic, comparison, boolean operators
- `import` (ignored)

PR #572, opened 2026-03-13
feat: support break and continue in Python→Laurel translation

MikaelMayer:
## What

Translates Python `break` and `continue` to Laurel `Exit` statements targeting labelled blocks. Also restructures try/except translation to use proper labelled blocks, enabling uniform `Exit` handling across the pipeline.

### break/continue
- Adds `loopBreakLabel` and `loopContinueLabel` to `TranslationContext`
- `while` and `for` loops wrap their body in a labelled block (for `continue`) and the whole loop in a labelled block (for `break`)
- `break` emits `Exit breakLabel`; `continue` emits `Exit continueLabel`

### try/except restructuring
Previously, `Exit "exception_handlers"` was a no-op because it tried to jump *into* a sibling block, which Core doesn't support. The translation is now restructured as:
```
try: {
  catchers: {
    body_stmt; if (isError) { exit catchers; }
    ...
    exit try;   -- normal completion, skip handlers
  }
  handler_code
}
```
This makes `Exit` semantics uniform: all exits jump *out of* their enclosing labelled block.

### LaurelToCoreTranslator
- `Exit` now translates to `Core.Stmt.exit` for all labels
- Labelled `Block` now translates to `Core.Stmt.block` for all labelled blocks

## Tests

Added `test_break_continue.py` covering `for` loops with `break` and `continue`.
`pyAnalyzeLaurel` SARIF is validated; `pyAnalyze` SARIF is skipped since `PythonToCore` does not yet support these constructs.

Tracks: #565

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -93,6 +93,8 @@ structure TranslationContext where
   compositeTypeNames : Std.HashSet String := {}
   /-- Track current class during method translation -/
   currentClassName : Option String := none
+  loopBreakLabel : Option String := none
+  loopContinueLabel : Option String := none
 deriving Inhabited
 
 /-! ## Error Handling -/
@@ -956,15 +958,19 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
         ([varDecl], varRef, { ctx with variableTypes := ctx.variableTypes ++ [(freshVar, "bool")] })
       | _ => ([], condExpr, ctx)
 
-    let (loopCtx, bodyStmts) ← translateStmtList condCtx body.val.toList
-    let bodyBlock := mkStmtExprMd (StmtExpr.Block bodyStmts none)
-    let whileStmt := mkStmtExprMdWithLoc (StmtExpr.While (Any_to_bool finalCondExpr) [] none bodyBlock) md
+    let breakLabel := s!"loop_break_{test.toAst.ann.start.byteIdx}"
+    let continueLabel := s!"loop_continue_{test.toAst.ann.start.byteIdx}"
+    let loopCtx := { condCtx with loopBreakLabel := some breakLabel, loopContinueLabel := some continueLabel }
+    let (_, bodyStmts) ← translateStmtList loopCtx body.val.toList
+    let bodyBlock := mkStmtExprMd (StmtExpr.Block bodyStmts (some continueLabel))
+    let whileStmt := mkStmtExprMd (StmtExpr.While (Any_to_bool finalCondExpr) [] none bodyBlock)
+    let whileWrapped := mkStmtExprMdWithLoc (StmtExpr.Block [whileStmt] (some breakLabel)) md
 
     -- Wrap in block if we hoisted condition
     let result := if condStmts.isEmpty then
-      whileStmt
+      whileWrapped
     else
-      mkStmtExprMdWithLoc (StmtExpr.Block (condStmts ++ [whileStmt]) none) md
+      mkStmtExprMdWithLoc (StmtExpr.Block (condStmts ++ [whileWrapped]) none) md
 
     return (loopCtx, [result])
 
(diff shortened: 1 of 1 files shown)


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -330,7 +330,18 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
   | .Assume cond =>
       let coreExpr ← translateExpr cond [] (isPureContext := true)
       return [Core.Statement.assume ("assume" ++ getNameFromMd md) coreExpr md]
-  | .Block stmts _ => stmts.flatMapM (fun s => translateStmt outputParams s)
+  | .Block stmts label =>
+      let innerStmts ← stmts.flatMapM (fun s => translateStmt outputParams s)
+      -- Wrap in a Core labelled block for loop labels (break/continue)
+      -- and try/except labels (exception_handlers, try_end).
+      let isLabelledBlock := label.any fun l =>
```
keyboardDrummer: I prefer to keep the LaurelToCore pass as dumb as possible. The `l.startsWith "loop_break_"` checks don't seem maintainable. How about we always translate to a `Imperative.Stmt.block` when the Laurel Block has a label?

joehendrix (commented):
The PR scope has evolved from break/continue support to primarily: (1) `exit` and labelled-block grammar support in Laurel, (2) restructured try/except translation with unique labels and proper variable hoisting, and (3) composite type re-assignment fixes. The labelled-block infrastructure is solid and sets the stage for future break/continue work.

**One blocker:** `test_break_continue.py` and its expected output are still in the tree, but break/continue translation was removed during the branch's evolution. Running `run_py_analyze.sh laurel` fails on this test:
```
Exception: Python to Laurel translation failed: Unsupported construct: Statement type not yet supported
ERROR: Analysis output for test_break_continue does not match expected result
```
Either remove these files (and the `"test_break_continue"` entry in `SKIP_TESTS`) or re-implement break/continue.

**Everything else looks good** — all other Laurel tests pass, the try/except scoping tests verify nested blocks and variable visibility, and `StrataTest.Languages.Python.AnalyzeLaurelTest` builds cleanly.

**Merge conflicts:** Two trivial conflicts with `origin/main`:
1. `LaurelGrammar.lean` — comment-only, keep the PR's line.
2. `run_py_analyze_sarif.py` — combine skip-list additions from both sides (main added `test_class_field_init`).

StrataTest/Languages/Python/tests/test_break_continue.py
```diff
@@ -0,0 +1,20 @@
+def test_while_break() -> None:
```
joehendrix, line 1: This test file exercises `break` and `continue` statements, but support for these was removed from `PythonToLaurel.lean` during the branch's evolution. The catch-all in `translateStmt` now throws `unsupportedConstruct` for `.Break` and `.Continue`.

Suggest removing this file, `expected_laurel/test_break_continue.expected`, and `"test_break_continue"` from `SKIP_TESTS` in `run_py_analyze_sarif.py`.

StrataTest/Languages/Python/expected_laurel/test_try_except.expected
```diff
@@ -0,0 +1,32 @@
...
+ret_type: ✅ pass (in prelude file)
+ret_pos: ✅ pass (in prelude file)
+assert_name_is_foo: ✅ pass (in prelude file)
+assert_opt_name_none_or_str: ✅ pass (in prelude file)
+assert_opt_name_none_or_bar: ✅ pass (in prelude file)
+ensures_maybe_except_none: ✅ pass (in prelude file)
+assert_assert(129)_calls_Any_to_bool_0: ✅ pass (at line 7, col 4)
+assert(129): ✅ pass (at line 7, col 4)
+assert_assert(358)_calls_Any_to_bool_0: ✅ pass (at line 17, col 4)
+assert(358): ❓ unknown (at line 17, col 4)
```
joehendrix, line 32:
```
assert(358): ❓ unknown (at line 17, col 4)
```
This corresponds to `assert result == 2` in `test_try_with_exception`. The verifier can't prove the handler executes because `raise` is translated to `.Hole` (havoc), so there's no guarantee the exception path is taken. This is an expected limitation of the current abstraction — worth a brief comment in the test file noting it's intentional.

StrataTest/Languages/Python/run_py_analyze_sarif.py
```diff
@@ -20,7 +20,7 @@
 )
 
 BOTH_SKIP = {"test_foo_client_folder", "test_invalid_client_type", "test_unsupported_config"}
-SKIP_TESTS = BOTH_SKIP | {"test_class_field_use", "test_list", "test_subscription", "test_with_statement", "test_class_field_init"}
+SKIP_TESTS = BOTH_SKIP | {"test_class_field_use", "test_list", "test_subscription", "test_with_statement", "test_class_field_init", "test_break_continue", "test_try_except", "test_try_except_scoping"}
```
joehendrix, line 23: Note: this line will conflict with `origin/main`, which added `"test_class_field_init"` to `SKIP_TESTS`. Resolution is straightforward — combine both additions. Also remove `"test_break_continue"` if the test file is removed per the other comment.

StrataTest/Languages/Python/expected_laurel/test_try_except.expected
MikaelMayer: We'll get to that eventually

PR #573, opened 2026-03-13
feat: support augmented assignment (+=, -=, *=) in Python→Laurel translation

MikaelMayer:
## What

Translates Python augmented assignment (`x += y`, `x -= y`, `x *= y`) to `x = x op rhs`.

## How

- Handles `AugAssign` for `+`, `-`, `*` operators precisely
- Unsupported operators (e.g. `**`, `//`) fall back to `Hole` (uninterpreted value)

## Tests

Added `test_augmented_assign.py` verifying `+=`, `-=`, `*=`.


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1160,6 +1160,18 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+  -- Augmented assignment: x += expr  →  x = x op expr
+  | .AugAssign _ target op value => do
+    let targetExpr ← translateExpr ctx target
+    let valueExpr ← translateExpr ctx value
+    let rhs := match op with
+      | .Add _  => mkStmtExprMd (StmtExpr.StaticCall "PAdd" [targetExpr, valueExpr])
+      | .Sub _  => mkStmtExprMd (StmtExpr.StaticCall "PSub" [targetExpr, valueExpr])
+      | .Mult _ => mkStmtExprMd (StmtExpr.StaticCall "PMul" [targetExpr, valueExpr])
+      | _       => mkStmtExprMd .Hole
+    let assignStmt := mkStmtExprMdWithLoc (StmtExpr.Assign [targetExpr] rhs) md
```
shigoel, line 1174: The wildcard here silently degrades `x //= y` and `x %= y` to `Hole`, but `BinOp` already handles these: `FloorDiv` → `"PFloorDiv"` and `Mod` → `"PMod"` (lines 402–403). These two cases should be added to `AugAssign` for consistency.

shigoel, line 1174: Nit: `BinOp` throws `(.unsupportedConstruct ...)` for operators it doesn't recognize (line 408), but here the wildcard silently emits `Hole` for everything. This inconsistency could mask translation bugs. Consider restricting the `Hole` fallback to the same set `BinOp` degrades gracefully (`Div`, `BitAnd`, `BitOr`, `BitXor`), and throwing for truly unrecognized operators.

StrataTest/Languages/Python/run_py_analyze_sarif.py
```diff
@@ -20,7 +20,7 @@
 )
 
 BOTH_SKIP = {"test_foo_client_folder", "test_invalid_client_type", "test_unsupported_config"}
-SKIP_TESTS = BOTH_SKIP | {"test_class_field_use", "test_list", "test_subscription", "test_with_statement", "test_class_field_init", "test_break_continue", "test_try_except", "test_try_except_scoping"}
+SKIP_TESTS = BOTH_SKIP | {"test_class_field_use", "test_list", "test_subscription", "test_with_statement", "test_class_field_init", "test_break_continue", "test_try_except", "test_try_except_scoping", "test_augmented_assign"}
 SKIP_TESTS_LAUREL = BOTH_SKIP
```
shigoel, line 25: Worth a brief inline comment explaining why this test is skipped in the sarif path (e.g. `# sarif pipeline doesn't yet handle AugAssign`). The Laurel path correctly runs it (it's not in `SKIP_TESTS_LAUREL`), so this will likely confuse readers who notice the asymmetry.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Fixed: added `FloorDiv` → `PFloorDiv` and `Mod` → `PMod` to the AugAssign operator mapping, consistent with BinOp.

MikaelMayer: Acknowledged. The wildcard Hole fallback is intentional for now — PR #566 (conservative abstraction) will make BinOp consistent by also returning Hole instead of throwing for unrecognized operators. Once #566 merges, both will behave the same way.

StrataTest/Languages/Python/run_py_analyze_sarif.py
MikaelMayer: Added a comment explaining the skip. The Laurel path runs the test correctly; the sarif path uses PythonToCore which doesn't yet support AugAssign.

PR #574, opened 2026-03-13
Add CFG-to-CProverGOTO translation

tautschnig:
The existing Stmt-to-GOTO translation (ToCProverGOTO.lean) interleaves control-flow lowering — emitting conditional jumps, managing forward references, patching targets — with CBMC-specific concerns like source locations and instruction encoding. This makes the code hard to extend to additional backends and difficult to reason about independently.

*Description of changes:*

This commit introduces a CFG-based path that separates those responsibilities:

  Core Stmt → CFG (via stmtsToCFG) → CProverGOTO instructions

The first step (stmtsToCFG from StructuredToUnstructured) handles control-flow lowering once, producing a backend-agnostic CFG. The second step (detCFGToGotoTransform, added here) is a straightforward mapping from CFG blocks to GOTO instructions. This separation makes it easier to target additional backends — each only needs to consume the CFG — and to reason about control-flow lowering independently.

Three gaps remain before this can replace the direct path:

1. Source locations on control flow: StructuredToUnstructured discards MetaData from ite/loop/block/exit statements, so transfer commands in the CFG carry no source location information.

2. Loop contracts: the direct path emits #spec_loop_invariant and #spec_decreases as named sub-expressions on the backward-edge GOTO (for CBMC DFCC). In the CFG, invariants become plain asserts and measures are discarded.

3. Core.CmdExt.call: this translation handles Imperative.Cmd only. Core procedure calls need a command translator analogous to coreStmtsToGoto in CoreToGOTOPipeline.lean.

No changes to StructuredToUnstructured are needed: the existing stmtsToCFG already works with Core.Command because its CmdT parameter is generic, and Core.Command has the required HasPassiveCmds and HasBool instances.

Tests cover sequential commands, if-then-else, loops (including back-edge verification), empty programs, assert/assume, and havoc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,34 @@ jobs:
         auto-config: false
         build: true
         build-args: "strata:exe"
+    - name: Install cvc5
+      shell: bash
+      run: |
+            ARCH=$(uname -m)
+            if [ "$ARCH" = "x86_64" ]; then
+              ARCH_NAME="x86_64"
+            elif [ "$ARCH" = "aarch64" ] || [ "$ARCH" = "arm64" ]; then
+              ARCH_NAME="arm64"
+            else
+              echo "Unsupported architecture: $ARCH"
+              exit 1
+            fi
+            wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-Linux-${ARCH_NAME}-static.zip
+            unzip cvc5-Linux-${ARCH_NAME}-static.zip
+            chmod +x cvc5-Linux-${ARCH_NAME}-static/bin/cvc5
+            sudo cp cvc5-Linux-${ARCH_NAME}-static/bin/cvc5 /usr/local/bin/
+    - name: Install z3
+      shell: bash
+      run: |
+            pip install z3-solver
+            Z3_PATH=$(python -c "import z3; import os; print(os.path.join(os.path.dirname(z3.__file__), 'lib', 'z3'))")
+            sudo cp "$Z3_PATH" /usr/local/bin/z3 || true
+            # Fallback: install from GitHub release
+            if ! command -v z3 &> /dev/null; then
+              wget https://github.com/Z3Prover/z3/releases/download/z3-4.13.4/z3-4.13.4-x64-glibc-2.35.zip
+              unzip z3-4.13.4-x64-glibc-2.35.zip
+              sudo cp z3-4.13.4-x64-glibc-2.35/bin/z3 /usr/local/bin/
+            fi
     - name: Run PySpec and dispatch tests
       run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest StrataTest.Languages.Python.AnalyzeLaurelTest StrataTest.Languages.Python.Specs.IdentifyOverloadsTest
     - name: Run test script
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -64,7 +64,8 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
   -- Encode the obligation term Q (not negated)
   let (obligationId, estate) ← (encodeTerm False obligationTerm) |>.run estate
 
-  let ids := estate.ufs.values
+  let ids := estate.ufs.toList.filterMap fun (uf, id) =>
+    if uf.args.isEmpty then some id else none
 
   -- Choose encoding strategy: use check-sat-assuming only when doing both checks
   let bothChecks := satisfiabilityCheck && validityCheck
--- a/Strata/Languages/Laurel/Laurel.lean
+++ b/Strata/Languages/Laurel/Laurel.lean
@@ -375,6 +375,10 @@ def Body.isExternal : Body → Bool
   | .External => true
   | _ => false
 
+def Body.isTransparent : Body → Bool
+  | .Transparent _ => true
+  | _ => false
+
 def HighTypeMd.isBool (t : HighTypeMd) : Bool := t.val.isBool
 
 /--
--- a/Strata/Languages/Python/PySpecPipeline.lean
+++ b/Strata/Languages/Python/PySpecPipeline.lean
@@ -34,6 +34,8 @@ public structure PySpecLaurelResult where
   laurelProgram : Laurel.Program
   overloads : OverloadTable
   functionSignatures : List Python.PythonFunctionDecl := []
+  /-- Maps unprefixed class names to prefixed names for type resolution. -/
+  typeAliases : Std.HashMap String String := {}
 
 /-! ### Private Helpers -/
 
@@ -96,6 +98,7 @@ public def buildPySpecLaurel (pyspecPaths : Array String)
   let mut combinedTypes : Array (Laurel.TypeDefinition × String) := #[]
   let mut allOverloads := overloads
   let mut funcSigs : List Python.PythonFunctionDecl := []
+  let mut allTypeAliases : Std.HashMap String String := {}
   for ionPath in pyspecPaths do
     let ionFile : System.FilePath := ionPath
     let some mod := ionFile.fileStem
@@ -107,7 +110,7 @@ public def buildPySpecLaurel (pyspecPaths : Array String)
       match ← Python.Specs.readDDM ionFile |>.toBaseIO with
       | .ok t => pure t
       | .error msg => throw s!"Could not read {ionFile}: {msg}"
-    let { program, errors, overloads } :=
+    let { program, errors, overloads, typeAliases } :=
       Python.Specs.ToLaurel.signaturesToLaurel ionPath sigs modulePrefix
     if errors.size > 0 then
       let _ ← IO.eprintln
@@ -115,6 +118,7 @@ public def buildPySpecLaurel (pyspecPaths : Array String)
       for err in errors do
         let _ ← IO.eprintln s!"  {err.file}: {err.message}" |>.toBaseIO
     allOverloads := mergeOverloads allOverloads overloads
+    allTypeAliases := typeAliases.fold (init := allTypeAliases) fun m k v => m.insert k v
     match extractFunctionSignatures sigs modulePrefix with
     | .ok fs => funcSigs := funcSigs ++ fs
     | .error msg => throw msg
@@ -149,7 +153,7 @@ public def buildPySpecLaurel (pyspecPaths : Array String)
     constants := []
   }
   return { laurelProgram := combinedLaurel, overloads := allOverloads
-           functionSignatures := funcSigs }
+           functionSignatures := funcSigs, typeAliases := allTypeAliases }
 
 /-- Read dispatch Ion files and merge their overload tables. -/
 public def readDispatchOverloads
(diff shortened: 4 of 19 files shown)


Strata/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,124 @@
...
+       - `condGoto c lt lf` → GOTO [!c] lf; GOTO lt (conditional + fallthrough)
+       - `finish` → no instruction (handled by END_FUNCTION in the caller)
+
+    After all blocks are emitted, a second pass patches GOTO targets using the
+    label-to-location map built during the first pass.
+-/
+def detCFGToGotoTransform {P} [G : ToGoto P] [BEq P.Ident]
+    (T : P.TyEnv) (functionName : String)
+    (cfg : CFG String (DetBlock String (Cmd P) P))
+    (loc : Nat := 0)
```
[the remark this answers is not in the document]
tautschnig: Good catch. Fixed in de6537e6: `detCFGToGotoTransform` now validates that `cfg.entry` matches the first block's label and returns an error if not. Added a test that verifies the error is raised for a mismatched entry.

Strata/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,124 @@
...
+      trans := trans'
+      pendingPatches := (trueIdx, lt) :: pendingPatches
+    | .finish _md =>
+      -- No instruction needed; the caller appends END_FUNCTION
+      pure ()
+  -- Second pass: patch all GOTO targets
+  let patches := pendingPatches.filterMap fun (idx, label) =>
+    match labelMap.find? (fun (l, _) => l == label) with
+    | some (_, loc) => some (idx, loc)
+    | none => none
```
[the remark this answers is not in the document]
tautschnig: Agreed. Fixed in de6537e6: the patch loop now returns an `Except.error` with the unresolved label name instead of silently dropping it via `filterMap`.

Strata/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,124 @@
...
+    --       emits `FUNCTION_CALL` instructions for `.call`, delegating `.cmd`
+    --       to `Cmd.toGotoInstructions`.
+    for cmd in block.cmds do
+      trans ← Cmd.toGotoInstructions trans.T functionName cmd trans
+    -- Translate the transfer command
+    match block.transfer with
+    | .condGoto cond lt lf _md =>
+      let cond_expr ← G.toGotoExpr cond
+      -- Emit: GOTO [!cond] lf
+      let (trans', falseIdx) := emitCondGoto (Expr.not cond_expr) srcLoc trans
```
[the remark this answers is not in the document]
tautschnig: You're right — `DetTransferCmd.condGoto` and `.finish` already have an `md : MetaData P` field. The real gap is that `StructuredToUnstructured.stmtsToBlocks` fills it with `MetaData.empty` (the metadata from `ite`/`loop`/`block`/`exit` is discarded as `_md`). Fixed the comment in de6537e6 to reflect this.

StrataTest/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,238 @@
...
+  let ans ← Imperative.detCFGToGotoTransform Lambda.TEnv.default "test" cfg
+  return format ans.instructions
+
+-- Verify GOTOs have resolved targets
+#eval do
+  let cfg := Imperative.stmtsToCFG iteCmds
+  let ans ← Imperative.detCFGToGotoTransform Lambda.TEnv.default "test" cfg
+  let gotos := ans.instructions.toList.filter (fun (i : CProverGOTO.Instruction) =>
+    i.type == CProverGOTO.InstructionType.GOTO)
+  assert! gotos.any (fun (i : CProverGOTO.Instruction) => i.target.isSome)
```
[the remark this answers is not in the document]
tautschnig: Agreed. Fixed in de6537e6: changed `any ... target.isSome` to `all ... target.isSome`, and added a separate all-GOTOs-resolved check for sequential programs as well.

Strata/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,141 @@
...
+      trans := trans'
+      pendingPatches := (trueIdx, lt) :: pendingPatches
+    | .finish _md =>
+      -- No instruction needed; the caller appends END_FUNCTION
+      pure ()
+  -- Second pass: patch all GOTO targets, failing on unresolved labels
+  let mut patchedTrans := trans
+  for (idx, label) in pendingPatches do
+    match labelMap.find? (fun (l, _) => l == label) with
+    | some (_, loc) =>
```
shigoel: **Issue 1: `loc` variable shadowing**

The pattern binding `loc` here shadows the outer `loc : Nat` parameter (the starting location offset passed by the caller). If someone later adds code in this branch expecting the parameter, they'll silently get the wrong value.

Suggest renaming the inner binding:
```lean
| some (_, targetLoc) =>
  patchedTrans := patchGotoTargets patchedTrans [(idx, targetLoc)]
```

Strata/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,141 @@
...
+      patchedTrans := patchGotoTargets patchedTrans [(idx, loc)]
+    | none =>
+      throw f!"[detCFGToGotoTransform] Unresolved label '{label}' referenced \
+               by GOTO at instruction index {idx}."
+  return patchedTrans
```
shigoel: **Issue 2: `patchGotoTargets` called once per patch instead of once total**

`patchGotoTargets` already accepts a `List (Nat × Nat)`, so calling it in a loop is unnecessary. More importantly, `labelMap.find?` inside the loop is O(n) per lookup, making the second pass O(n²) overall.

Suggested fix — resolve all labels first, then patch in one call:
```lean
-- Second pass: resolve all pending labels, then patch in one call
let mut resolvedPatches : List (Nat × Nat) := []
for (idx, label) in pendingPatches do
  match labelMap.find? (fun (l, _) => l == label) with
  | some (_, targetLoc) => resolvedPatches := (idx, targetLoc) :: resolvedPatches
  | none =>
    throw f!"[detCFGToGotoTransform] Unresolved label '{label}' referenced              by GOTO at instruction index {idx}."
return patchGotoTargets trans resolvedPatches
```

This also eliminates `patchedTrans` as a separate mutable variable. As a further optional improvement, replacing `labelMap : List (String × Nat)` with `Std.HashMap String Nat` would reduce the lookup to O(1).

StrataTest/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,263 @@
...
+#eval do
+  let cfg := Imperative.stmtsToCFG seqCmds
+  let ans ← Imperative.detCFGToGotoTransform Lambda.TEnv.default "test" cfg
+  let gotos := ans.instructions.toList.filter (fun (i : CProverGOTO.Instruction) =>
+    i.type == CProverGOTO.InstructionType.GOTO)
+  assert! gotos.all (fun (i : CProverGOTO.Instruction) => i.target.isSome)
+
+-------------------------------------------------------------------------------
+
+end
```
shigoel, line 310: **Issue 3: Missing test for the unresolved-label error path**

There's a test for the "entry not first" error, but no test for the other `throw` in `detCFGToGotoTransform`: a CFG whose `condGoto` references a label that doesn't appear in `cfg.blocks`. This path is easy to exercise directly:

```lean
#eval do
  let cfg : CFG String (DetBlock String (Cmd LExprTP) LExprTP) :=
    { entry := "entry",
      blocks := [("entry", { cmds := [],
        transfer := .condGoto HasBool.tt "missing_label" "also_missing" })] }
  match Imperative.detCFGToGotoTransform Lambda.TEnv.default "test" cfg with
  | .error e => assert! (s!"{e}".splitOn "Unresolved label").length > 1
  | .ok _    => assert! false
```

Strata/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,141 @@
...
+  | (firstLabel, _) :: _ =>
+    if firstLabel != cfg.entry then
+      throw f!"[detCFGToGotoTransform] Entry label '{cfg.entry}' does not match \
+               first block label '{firstLabel}'. The entry block must be listed first."
+  | [] => pure ()
+  -- First pass: emit instructions and build label→locationNum map
+  let mut trans : GotoTransform P.TyEnv :=
+    { instructions := #[], nextLoc := loc, T := T, sourceText := sourceText }
+  -- Pending GOTO patches: (instruction array index, target label)
+  let mut pendingPatches : List (Nat × String) := []
```
shigoel: **Nit: `pendingPatches` uses `List` with prepend while `instructions` uses `Array` with `push`**

The reversed accumulation order doesn't affect correctness (patching is order-independent), but using `Array` with `push` would be more consistent with the rest of the code:

```lean
let mut pendingPatches : Array (Nat × String) := #[]
-- ...
pendingPatches := pendingPatches.push (falseIdx, lf)
-- ...
pendingPatches := pendingPatches.push (trueIdx, lt)
```

Strata/DL/Imperative/CFGToCProverGOTO.lean
tautschnig: Fixed in 3d22700d: renamed the inner binding from `loc` to `targetLoc`.

Strata/DL/Imperative/CFGToCProverGOTO.lean
tautschnig: Fixed in 3d22700d: the second pass now collects all resolved patches into a list first, then calls `patchGotoTargets` once. Also replaced `labelMap : List (String × Nat)` with `Std.HashMap String Nat` for O(1) lookups, and eliminated the `patchedTrans` mutable variable.

StrataTest/DL/Imperative/CFGToCProverGOTO.lean
tautschnig: Fixed in 3d22700d: added a test that constructs a CFG with a `condGoto` referencing labels not present in `cfg.blocks`, and asserts that the error message contains "Unresolved label".

Strata/DL/Imperative/CFGToCProverGOTO.lean
tautschnig: Fixed in 3d22700d: switched to `Array` with `push`, consistent with how `instructions` is accumulated.

shigoel (approved):

StrataTest/DL/Imperative/CFGToCProverGOTO.lean
```diff
@@ -0,0 +1,298 @@
...
+ GOTO skip [((not(true : bool)) : bool)],
+ GOTO skip,
+ LOCATION skip,
+ GOTO skip [((not(true : bool)) : bool)],
+ GOTO skip,
+ LOCATION skip,
+ ASSIGN (assign (i : unsignedbv[32]) (((i : unsignedbv[32]) + (1 : unsignedbv[32])) : unsignedbv[32])),
+ GOTO skip [((not(true : bool)) : bool)],
+ GOTO skip,
+ LOCATION skip]
```
atomb: Are all of the `skip`s in this intentional? It's not clear to me how the `GOTO` instructions work here.

atomb (commented):
There's one bit that seems odd to me, but I don't know if it's as intended or not.

StrataTest/DL/Imperative/CFGToCProverGOTO.lean
tautschnig: Thank you, good catch! The `skip` was the default `Code.skip` field being printed — not a target. The actual GOTO targets are location numbers stored in the `target` field, which `toString` wasn't displaying. Fixed in c54d18df: GOTO instructions now show their target location number instead of the code field, so the control flow is visible in the test output:

```
Before: GOTO skip [((not(true : bool)) : bool)]
After:  GOTO 5 [((not(true : bool)) : bool)]
```

For example, in the loop test you can now see the back-edge: `GOTO 5` at location 10 jumps back to the loop entry at location 5.

atomb: Is the same true for the `LOCATION` instructions? Is that how labels are created?

tautschnig: Thanks, fixed for `LOCATION` and then also `ASSUME` and `ASSERT`, which had the same problem.

PR #576, opened 2026-03-13
Eliminate Python dependency from CBMC pipeline

tautschnig:
The CBMC backend relied on a Python post-processing step (process_json.py) to merge a 4000-line defaults.json of built-in symbols into the Lean-generated symbol table before passing it to symtab2gb. This made the pipeline harder to maintain, required Python as a runtime dependency, and kept critical symbol definitions outside of Lean where they could not benefit from type checking.

*Description of changes:*

Generate all 38 CBMC built-in symbols directly in Lean (DefaultSymbols.lean), parameterized by ArchConfig and module name, and emit the symbol table in the {"symbolTable": ...} wrapper that symtab2gb expects. This turns the pipeline into a simple two-step flow (strata → symtab2gb) with no Python intermediary.

Remove the now-unused Python scripts (process_json.py, defaults.py), defaults.json, and legacy shell wrappers (run_strata_cbmc.sh, compare_cbmc_json.sh). Update all pipeline scripts and documentation to reflect the simplified flow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,8 @@ name: Build
 on:
   pull_request:
   merge_group:
+  push:
+    branches: [main]
 
 env:
   SOLUTION: Tools/BoogieToStrata/BoogieToStrata.sln
@@ -75,6 +77,8 @@ jobs:
           key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
           restore-keys: |
             lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}
+            lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}
+            lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}
       - name: Build and test Strata
         uses: leanprover/lean-action@v1
         with:
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -938,15 +938,15 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     let whileWrapped := mkStmtExprMdWithLoc (StmtExpr.Block [whileStmt] (some breakLabel)) md
     return (loopCtx, [whileWrapped])
 
-  -- Return statement
+  -- Return statement: assign to the LaurelResult output parameter, then exit $body.
   | .Return _ value => do
-    let retVal ← match value.val with
+    let stmts ← match value.val with
       | some expr => do
         let e ← translateExpr ctx expr
-        .ok (some e)
-      | none => .ok none
-    let retStmt := mkStmtExprMd (StmtExpr.Return retVal)
-    return (ctx, [retStmt])
+        let assign := mkStmtExprMd (StmtExpr.Assign [mkStmtExprMd (StmtExpr.Identifier "LaurelResult")] e)
+        .ok [assign, mkStmtExprMd (StmtExpr.Exit "$body")]
+      | none => .ok [mkStmtExprMd (StmtExpr.Exit "$body")]
+    return (ctx, stmts)
 
   -- Assert statement
   | .Assert _ test _msg => do
@@ -1227,7 +1227,8 @@ def translateFunction (ctx : TranslationContext) (sourceRange: SourceRange) (fun
     -- Translate return type
 
 
-    -- Determine outputs based on return type
+    -- Declare an output parameter when the function has a return type annotation.
+    -- Return statements explicitly assign to LaurelResult and exit $body.
     let outputs : List Parameter :=
       match funcDecl.ret with
       | none => []
--- a/StrataTest/Languages/Python/AnalyzeLaurelTest.lean
+++ b/StrataTest/Languages/Python/AnalyzeLaurelTest.lean
@@ -94,8 +94,12 @@ private meta def runAnalyze (dispatchIon : System.FilePath)
     | .ok r => pure r
     | .error err => return .error (toString err)
   match Strata.translateCombinedLaurel laurel with
-  | (some core, []) => .ok $ pure core
-  | (_, errors) => .error s!"Laurel to Core translation failed: {errors}"
+  | (some core, []) =>
+    -- Also run Core type checking to catch semantic errors (e.g. Heap vs Any)
+    match Core.typeCheck Core.VerifyOptions.quiet core with
+    | .error diag => return .error s!"Core type checking failed: {diag}"
+    | .ok _ => return .ok core
+  | (_, errors) => return .error s!"Laurel to Core translation failed: {errors}"
 
 /-- Expected outcome for a test case. -/
 private inductive Expected where
@@ -109,6 +113,10 @@ private meta def testCases : List (String × Expected) := [
   .mk "test_multi_service.py" .success,
   .mk "test_annotation_fallback.py" .success,
   .mk "test_required_with_optional.py" .success,
+  .mk "test_heap_return.py" .success,
+  .mk "test_list_str.py" .success,
+  .mk "test_nested_try.py" .success,
+  .mk "test_try_scope.py" .success,
   -- Negative tests
   .mk "test_invalid_service.py" $
     .fail "User code error: 'connect' called with unknown string \"invalid\"; known services: #[messaging, storage]",
--- a/StrataTest/Languages/Python/Specs/dispatch_test/test_heap_return.py
+++ b/StrataTest/Languages/Python/Specs/dispatch_test/test_heap_return.py
@@ -1,0 +1,6 @@
+import servicelib
+
+def create_and_check() -> bool:
+    client: Storage = servicelib.connect("storage")
+    client.put_item(Bucket="b", Key="k", Data="v")
+    return True
--- a/StrataTest/Languages/Python/Specs/dispatch_test/test_list_str.py
+++ b/StrataTest/Languages/Python/Specs/dispatch_test/test_list_str.py
@@ -1,0 +1,5 @@
+import servicelib
+
+def use_list_str() -> bool:
+    names: List[str] = ["alice", "bob"]
+    return True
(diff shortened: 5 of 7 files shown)


Strata/Backends/CBMC/GOTO/DefaultSymbols.lean
```diff
@@ -0,0 +1,542 @@
...
+  let mkStrSym (name : String) (val : String) (line : String) : String × CBMCSymbol :=
+    (name, {
+      baseName := name, name := name, prettyName := name,
+      mode := "C", module := moduleName,
+      isLvalue := true, isStaticLifetime := true,
+      prettyType := "char *",
+      prettyValue := s!"\"{val}\"",
+      type := pointerTo (signedBvType 8 "char") cfg.pointerWidth,
+      value := stringConstValue val file line cfg.pointerWidth
+    })
```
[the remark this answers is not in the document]
tautschnig: Good catch. Fixed in e979d091: `prettyType` is now `"const char *"` and the symbol type uses `constSignedBvType` which includes the `#constant` qualifier on the pointee. The `address_of` type in the value intentionally omits `#constant`, matching the original `defaults.json`.

Strata/Backends/CBMC/GOTO/DefaultSymbols.lean
```diff
@@ -0,0 +1,542 @@
...
+    -- __CPROVER_constant_infinity_uint : const unsigned int
+    ("__CPROVER_constant_infinity_uint", {
+      baseName := "__CPROVER_constant_infinity_uint",
+      name := "__CPROVER_constant_infinity_uint",
+      prettyName := "__CPROVER_constant_infinity_uint",
+      mode := "C", module := moduleName,
+      isLvalue := true, isStaticLifetime := true,
+      prettyType := "const unsigned int",
+      type := unsignedBvType cfg.intWidth "unsigned_int",
+      value := nilValue
```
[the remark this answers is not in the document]
tautschnig: Fixed in e979d091: the type now uses `constUnsignedBvType` which includes `"#constant": {"id": "1"}` in the namedSub, matching the original `defaults.json`.

Strata/Backends/CBMC/GOTO/DefaultSymbols.lean
```diff
@@ -0,0 +1,542 @@
...
+    -- __CPROVER_initialize(void)
+    ("__CPROVER_initialize", {
+      baseName := "__CPROVER_initialize", name := "__CPROVER_initialize",
+      prettyName := "__CPROVER_initialize",
+      mode := "C", module := "",
+      isLvalue := true,
+      prettyType := "void (void)",
+      type := voidCodeType #[],
+      value := nilValue
+    }),
```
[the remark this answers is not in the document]
tautschnig: Fixed in e979d091: `builtinFunctionSymbols` now takes a `moduleName` parameter. `__CPROVER_initialize` and the assignable/freeable family (`__CPROVER_assignable`, `__CPROVER_object_whole`, `__CPROVER_object_upto`, `__CPROVER_object_from`, `__CPROVER_freeable`, `__CPROVER_is_freeable`, `__CPROVER_was_freed`) use it. `__CPROVER_assert` and `__CPROVER_assume` intentionally keep `module := ""`, matching the original `defaults.json`.

Strata/Backends/CBMC/GOTO/DefaultSymbols.lean
```diff
@@ -0,0 +1,542 @@
...
+            ("namedSub", Lean.Json.mkObj [
+              ("#source_location", Lean.Json.mkObj [("id", ""), ("namedSub", mkSourceLoc file line)]),
+              ("type", arrayType),
+              ("value", Lean.Json.mkObj [("id", s)])
+            ])
+          ],
+          Lean.Json.mkObj [
+            ("id", "constant"),
+            ("namedSub", Lean.Json.mkObj [
+              ("type", signedBvType 32 "signed_int"),
```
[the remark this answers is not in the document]
tautschnig: Fixed in e979d091: the zero-index constant now uses `signedBvType ptrWidth "signed_long_int"` (parameterized by pointer width rather than hardcoded), matching the original `defaults.json` for 64-bit targets.

docs/CoreToGOTO_Gaps.md
```diff
@@ -160,6 +158,29 @@ not handled:
 
 ### Known Limitations
 
+#### Hardcoded architecture configuration
+
+The CBMC default symbols (`DefaultSymbols.lean`) use `ArchConfig.default`,
+which is hardcoded to arm64/macOS/LP64. This means all generated symbol tables
+assume 64-bit pointers, 64-bit `long`, little-endian byte order, etc.
+
+To support cross-platform verification (e.g., 32-bit targets, Windows LLP64,
```
shigoel, line 167: Very cool that this is a config option exposed to Strata now, should it be needed!

Strata/Backends/CBMC/GOTO/DefaultSymbols.lean
```diff
@@ -0,0 +1,563 @@
...
+private def stringConstValue (s : String) (file : String) (line : String)
+    (ptrWidth : Nat) : Lean.Json :=
+  let len := s.length + 1
+  let arrayType := Lean.Json.mkObj [
+    ("id", "array"),
+    ("namedSub", Lean.Json.mkObj [
+      ("size", Lean.Json.mkObj [
+        ("id", "constant"),
+        ("namedSub", Lean.Json.mkObj [
+          ("type", signedBvType 64 "signed_long_int"),
```
shigoel: The array-size type is hardcoded to width 64 here, but the zero-index constant below (already fixed per the earlier Copilot thread) correctly uses `ptrWidth`. On a 32-bit target these two would be inconsistent. Suggest:

```lean
("type", signedBvType ptrWidth "signed_long_int"),
```

Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
```diff
@@ -398,7 +398,7 @@ symtab/goto JSON.
 -/
 def emitProcWithLifted (Env : Core.Expression.TyEnv) (procName : String)
     (ctx : CoreToGOTO.CProverGOTO.Context) (liftedFuncs : List Core.Function)
-    (extraSyms : Lean.Json)
+    (extraSyms : Lean.Json) (moduleName : String := "")
```
shigoel: The `moduleName := ""` default means any future call site that forgets to pass the argument will silently produce CBMC symbols (e.g. `__CPROVER_initialize`) with an empty module field rather than a compile error. Since `moduleName` is now semantically meaningful, consider removing the default and requiring callers to be explicit — or at minimum add a comment that `""` is only valid for built-in-only symbol tables, not user programs.

StrataToCBMC.lean
```diff
@@ -33,13 +38,17 @@ def main (args : List String) : IO Unit := do
         if file.endsWith ".csimp.st" then
           let csimp_prog := C_Simp.get_program pgm
           match CSimp.testSymbols csimp_prog.funcs.head! with
-          | .ok s => IO.println s
+          | .ok s =>
+            let wrapped ← wrapOutput s (moduleName := "")
```
shigoel: Both the `.csimp.st` path (line 42) and the `.core.st` path (line 50) pass `moduleName := ""`, so `__CPROVER_initialize` and the assignable/freeable family will have an empty module field in the output. The other entry points (`writeToGotoJson`, `emitProcWithLifted`, `laurelAnalyzeToGotoCommand`) all derive the module name from the file base name. For consistency, and in case `StrataToCBMC` is used outside the single legacy CI test, consider doing the same here:

```lean
let baseName := System.FilePath.mk file |>.stem
let wrapped ← wrapOutput s (moduleName := baseName)
```

Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
shigoel: BTW, I'm very wary of default arguments. E.g., see https://github.com/strata-org/Strata/pull/579.

Strata/Backends/CBMC/GOTO/DefaultSymbols.lean
tautschnig: Good catch - fixed in fc9bb19c. The array-size type now uses `ptrWidth` instead of the hardcoded `64`, consistent with the zero-index constant below.

Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
tautschnig: Agreed, and point taken re #579. Removed the default in b846434e - callers must now pass moduleName explicitly.

StrataToCBMC.lean
tautschnig: Fixed in 99eb4972. Both paths now derive the module name from the input file's base name (stripping `.st`), matching the other entry points.

Strata/Backends/CBMC/GOTO/DefaultSymbols.lean
```diff
@@ -0,0 +1,563 @@
...
+    mkIntSym "__CPROVER_architecture_long_int_width"     cfg.longIntWidth      "11",
+    mkIntSym "__CPROVER_architecture_long_long_int_width" cfg.longLongIntWidth "12",
+    mkIntSym "__CPROVER_architecture_memory_operand_size" cfg.memoryOperandSize "13",
+    mkStrSym "__CPROVER_architecture_os"                 cfg.os                "14",
+    mkIntSym "__CPROVER_architecture_pointer_width"      cfg.pointerWidth      "15",
+    mkIntSym "__CPROVER_architecture_short_int_width"    cfg.shortIntWidth     "16",
+    mkIntSym "__CPROVER_architecture_single_width"       cfg.singleWidth       "17",
+    mkIntSym "__CPROVER_architecture_wchar_t_is_unsigned" cfg.wcharTIsUnsigned "18",
+    mkIntSym "__CPROVER_architecture_wchar_t_width"      cfg.wcharTWidth       "19",
+    mkIntSym "__CPROVER_architecture_word_size"          cfg.wordSize          "20"
```
MikaelMayer, line 332: I have no context to verify these constants so I assume they are correct

copilot_responses.txt
```diff
@@ -0,0 +1,19 @@
+## Response to Comment 1 (ASTtoCST.lean:422 — SNegOverflow mapped to neg_expr)
```
MikaelMayer: This file should not be added.
What helped for me is to instruct the agent to never use git add -A but instead always add files by name.

pr_description.md
```diff
@@ -0,0 +1,57 @@
+## PR Title
```
MikaelMayer: This file should not be part of the PR

Strata/Backends/CBMC/GOTO/DefaultSymbols.lean
tautschnig: These constants were ported from the original `defaults.json` that was generated by CBMC's `--show-symbol-table` on a reference program.

copilot_responses.txt
tautschnig: Ugh, thank you! (And, indeed, I have now updated my steering doc.)

pr_description.md
tautschnig: Done.

PR #577, opened 2026-03-13
Add type constraint preconditions/postcondition for function inputs/output

thanhnguyen-aws:
*Description of changes:* 
This PR adds type constraint pre-conditions for function inputs to ensure consistency with the original Python code. Currently, all function inputs are translated to `Any` type, which loses type information during translation.

Changes

- Add type constraint assertions for function input parameters at the beginning of the function's body.

- Add pre-conditions based on declared input types.

- Add post-conditions based on declared out types, we also translate Python function into Opaque procedures to accommodate this.

- Support Python's type union operator (|) for type constraints.

Limitation: 

- This PR does not add type constraint for class methods. An upcoming separate PR will unify functions translation with class's methods translation.

- This PR does not add postcondition when the function return None.

### DIFF
--- a/Strata/DDM/Integration/Lean/Gen.lean
+++ b/Strata/DDM/Integration/Lean/Gen.lean
@@ -44,26 +44,31 @@ namespace Strata
 namespace Lean
 
 /--
-Prepend the current namespace to the Lean name and convert to an identifier.
-When in a module file and not in a `public section`, uses `mkPrivateName` to
-resolve the `.decl` pre-resolution hint to the private-mangled name.
+Resolve a name within the current scope: prepend the namespace and, when in a
+module file outside a `public section`, apply `mkPrivateName`.
 -/
-def mkScopedIdent (subName : Lean.Name) : CommandElabM Ident := do
+def resolveScopedName (subName : Lean.Name) : CommandElabM Lean.Name := do
   let env ← getEnv
   let scope ← getScope
   let fullName := scope.currNamespace ++ subName
-  let resolvedName :=
-    if !env.header.isModule || scope.isPublic then
-      fullName
-    else
-      Lean.mkPrivateName env fullName
+  if !env.header.isModule || scope.isPublic then
+    return fullName
+  else
+    return Lean.mkPrivateName env fullName
+
+/--
+Prepend the current namespace to the Lean name and convert to an identifier.
+Uses `resolveScopedName` for the `.decl` pre-resolution hint.
+-/
+def mkScopedIdent (subName : Lean.Name) : CommandElabM Ident := do
+  let resolvedName ← resolveScopedName subName
   let rawStr := (toString subName).toRawSubstring
   let preresolution := [.decl resolvedName []]
   return .mk (.ident .none rawStr subName preresolution)
 
 end Lean
 
-open Lean (mkScopedIdent)
+open Lean (mkScopedIdent resolveScopedName)
 
 def arrayLit [Monad m] [Lean.MonadQuotation m] (as : Array Term) : m Term := do
   ``( (#[ $as:term,* ] : Array _) )
@@ -1546,6 +1551,25 @@ partial def generateInhabitedInstances (group : Array (QualifiedIdent × Array D
     pure sm
 
 /--
+Checks that none of the category names that `#strata_gen` is about to
+introduce already exist in the Lean environment.  Reports an error for
+each collision and returns `true` when at least one conflict was found.
+-/
+def checkCategoryNamesAvailable
+    (categories : Array (QualifiedIdent × Array DefaultCtor))
+    : GenM Bool := do
+  let env ← getEnv
+  let mut hasConflict := false
+  for (cat, _) in categories do
+    if cat ∈ declaredCategories then continue
+    let catName ← getCategoryScopedName cat
+    let resolvedName ← resolveScopedName catName
+    if env.contains resolvedName then
+      logError m!"#strata_gen: '{catName}' already exists as '{resolvedName}'."
+      hasConflict := true
+  return hasConflict
+
+/--
 Generates all code for a list of categories: inductive types, Inhabited
 instances, toAst and ofAst functions. Processes categories in topologically
 sorted groups to handle dependencies correctly.
@@ -1553,6 +1577,7 @@ sorted groups to handle dependencies correctly.
 def generateCategoryCode
     (categories : Array (QualifiedIdent × Array DefaultCtor))
     : GenM Unit := do
+  if ← checkCategoryNamesAvailable categories then return
   let mut inhabitedCats : InhabitedSet :=
     Std.HashSet.ofArray
       declaredCategories.keysArray
--- a/Strata/Languages/Laurel/HeapParameterization.lean
+++ b/Strata/Languages/Laurel/HeapParameterization.lean
@@ -470,7 +470,6 @@ def heapParameterization (model: SemanticModel) (program : Program) : Program :=
   let program := { program with
     types := program.types
     staticProcedures := program.staticProcedures }
-  -- Collect instance procedures from composite types for heap analysis
   let instanceProcs := program.types.foldl (fun acc td =>
     match td with
     | .Composite ct => acc ++ ct.instanceProcedures
--- a/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
+++ b/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
@@ -694,6 +694,8 @@ def translateWithLaurel (options: LaurelTranslateOptions) (program : Program): T
   let (program, model) := (result.program, result.model)
   let diamondErrors := validateDiamondFieldAccesses model program
 
+  let (program, nonCompositeDiags) := filterNonCompositeModifies model program
+
   let program := heapParameterization model program
   let result := resolve program (some model)
   let (program, model) := (result.program, result.model)
@@ -720,7 +722,7 @@ def translateWithLaurel (options: LaurelTranslateOptions) (program : Program): T
 
   let initState : TranslateState := {model := model }
   let (coreProgramOption, translateState) := runTranslateM initState (translateLaurelToCore options program)
-  let allDiagnostics := resolutionErrors ++ diamondErrors ++ modifiesDiags ++ constrainedTypeDiags ++ translateState.diagnostics
+  let allDiagnostics := resolutionErrors ++ diamondErrors ++ nonCompositeDiags ++ modifiesDiags ++ constrainedTypeDiags ++ translateState.diagnostics
   let coreProgramOption := if translateState.coreProgramHasSuperfluousErrors then none else coreProgramOption
   (coreProgramOption, allDiagnostics, program)
   where
--- a/Strata/Languages/Laurel/LaurelTypes.lean
+++ b/Strata/Languages/Laurel/LaurelTypes.lean
@@ -102,6 +102,25 @@ def computeExprType (model : SemanticModel) (expr : StmtExprMd) : HighTypeMd :=
   | .All => default -- TODO: implement
   | .Hole _ typeOption => typeOption.getD  ⟨ HighType.Unknown, md ⟩
 
+/-- Classification of a heap-relevant modifies type. -/
+inductive ModifiesTypeKind where
+  | composite    -- a single Composite reference (UserDefined)
+  | compositeSet -- a Set of Composite references (TSet)
+
+/-- Classify a type as heap-relevant for modifies clauses, or `none` for
+non-heap-relevant types. Single source of truth for which types participate
+in modifies clauses and heap parameterization. -/
+def classifyModifiesHighType : HighType → Option ModifiesTypeKind
+  | .UserDefined _ => some .composite
+  | .TSet _        => some .compositeSet
+  | _              => none
+
+/-- Returns `true` when the given `HighType` is heap-relevant (composite or set
+of composite), i.e. the kind of type that appears in modifies clauses and
+triggers heap parameterization. -/
+def isHeapRelevantType (ty : HighType) : Bool :=
+  (classifyModifiesHighType ty).isSome
+
 end Strata.Laurel
 
 end
--- a/Strata/Languages/Laurel/ModifiesClauses.lean
+++ b/Strata/Languages/Laurel/ModifiesClauses.lean
@@ -48,15 +48,26 @@ inductive ModifiesEntry where
   | set (expr : StmtExprMd)          -- a Set Composite expression
 
 /--
+Classify a heap-relevant type into a `ModifiesEntry`, or `none` for
+non-heap-relevant types. Delegates to `classifyModifiesHighType` for the
+type classification.
+-/
+def classifyModifiesType (expr : StmtExprMd) (ty : HighType) : Option ModifiesEntry :=
+  match classifyModifiesHighType ty with
+  | some .composite    => some (.single expr)
+  | some .compositeSet => some (.set expr)
+  | none               => none
+
+/--
 Extract modifies entries from the list of modifies StmtExprs, using the type
 environment and type definitions to distinguish Composite from Set Composite.
+Non-composite types (e.g., global variables of primitive type) are filtered out
+since the frame condition only applies to heap objects.
 -/
 def extractModifiesEntries (model: SemanticModel)
     (modifiesExprs : List StmtExprMd) : List ModifiesEntry :=
-  modifiesExprs.map fun expr =>
-    match (computeExprType model expr).val with
-    | .TSet _ => .set expr
-    | _ => .single expr
+  modifiesExprs.filterMap fun expr =>
+    classifyModifiesType expr (computeExprType model expr).val
 /--
 Build the "obj is not modified" condition for a single modifies entry as a Laurel StmtExpr.
 - For a single Composite `e`: `$obj != e`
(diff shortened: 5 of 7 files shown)


StrataTest/Languages/Python/run_py_analyze_sarif.py
```diff
@@ -20,7 +20,7 @@
 )
 
 BOTH_SKIP = {"test_foo_client_folder", "test_invalid_client_type", "test_unsupported_config"}
-SKIP_TESTS = BOTH_SKIP | {"test_class_field_use", "test_list", "test_subscription", "test_with_statement"}
+SKIP_TESTS = BOTH_SKIP | {"test_class_field_use", "test_list", "test_subscription", "test_with_statement", "test_func_input_type_constraints"}
```
keyboardDrummer: Why is this skipped?

StrataMain.lean
```diff
@@ -607,7 +607,10 @@ def pyAnalyzeLaurelCommand : Command where
                 ("", "")
         | none => ("", "")
       let outcomeStr := vcResult.formatOutcome
-      s := s ++ s!"{locationPrefix}{vcResult.obligation.label}: \
+      let vcLabel := match vcResult.obligation.metadata.getPropertySummary with
```
keyboardDrummer: Consider using `.getD vcResult.obligation.label` to remove the need for the match. Did you run an LLM reviewer on this branch? Seems like it might help find some small options for Lean improvements.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -62,7 +62,7 @@ deriving Inhabited
 structure PythonFunctionDecl where
   name : String
   --args include name, type, default value
-  args : List (String × String × Option (Python.expr SourceRange))
+  args : List (String × MetaData × List String × Option (Python.expr SourceRange))
```
keyboardDrummer: Such a long tuple is hard to read. I think this deserves a structure. Could you add that?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1074,23 +1091,60 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
...
+def getUnionTypeConstraint (var: String) (md: MetaData) (tys: List String) (funcname: String): Option StmtExprMd :=
+  let type_constraints := tys.filterMap (getSingleTypeConstraint var)
+  if type_constraints.isEmpty then none else
+    let md: MetaData := md.withPropertySummary ("(" ++ funcname ++ " requires) Type constraint of " ++ var)
+    some {creatBoolOrExpr type_constraints with md:=md}
+
+def getUnionTypeAssertions (var: String) (md: MetaData) (tys: List String) (funcname: String): Option StmtExprMd :=
+  match getUnionTypeConstraint var md tys funcname with
+  | some constraint =>
+    let md: MetaData := md.withPropertySummary ("(" ++ funcname ++ " assert) Type constraint of " ++ var)
```
keyboardDrummer: Consider reducing use of parenthesis using `$`

StrataTest/Languages/Python/run_py_analyze_sarif.py
thanhnguyen-aws: We are no longer focus on Python->Core, so new tests of Python->Laurel should be add to SKIP_TESTS to avoid CI fail.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1258,23 +1282,60 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
...
+      else if ty.startsWith "List" then mkStmtExprMd (.StaticCall "Any..isfrom_ListAny" [freeVar var])
+      else none
+
+def creatBoolOrExpr (exprs: List StmtExprMd) : StmtExprMd :=
+  match exprs with
+  | [] => mkStmtExprMd (.LiteralBool true)
+  | [expr] => expr
+  | expr::exprs => mkStmtExprMd (.StaticCall "Bool.Or" [expr, creatBoolOrExpr exprs])
+
+def getUnionTypeConstraint (var: String) (md: MetaData) (tys: List String) (funcname: String): Option StmtExprMd :=
```
joehendrix, line 1504: There's a two-track type system here: value-world types (int, str, bool, Dict, List, etc.) map to `Core(Any)` with `isfrom_X` constraints, while heap-world types (user-defined classes) map to `Composite`/`UserDefined` — a fundamentally different Laurel type. The prelude's `Any` type can't hold composites, and `Composite` can't hold value types.

`getSingleTypeConstraint` doesn't have access to `ctx.compositeTypeNames`, so it silently returns `none` for composite types. For `Union[MyClass, str]`, `tys = ["MyClass", "str"]` — the composite `"MyClass"` is filtered out and the constraint becomes just `isfrom_string(x)`, which over-restricts.

But the deeper problem is that `Union[MyClass, str]` can't be soundly represented at all — there's no single Laurel type that can hold both a composite and a value. At line 1328–1332, since `tys.length != 1`, the parameter gets `AnyTy`, but `Any` can't hold `MyClass`.

This should be an error rather than silently generating an incorrect constraint. `getUnionTypeConstraint` (or the caller) should check whether any type in `tys` is in `compositeTypeNames` and report that mixed value/composite unions are not supported. The pyspec pipeline handles this by mapping `.pyClass` atoms to `Composite` and rejecting unions that mix composites with value types.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1258,23 +1282,60 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
...
   | _ => throw (.internalError "Expected FunctionDef")
 
+def getSingleTypeConstraint (var: String) (ty: String): Option StmtExprMd :=
+  match ty with
+  | "str" => mkStmtExprMd (.StaticCall "Any..isfrom_string" [freeVar var])
+  | "int" => mkStmtExprMd (.StaticCall "Any..isfrom_int" [freeVar var])
+  | "bool" => mkStmtExprMd (.StaticCall "Any..isfrom_bool" [freeVar var])
+  | "datetime" => mkStmtExprMd (.StaticCall "Any..isfrom_datetime" [freeVar var])
+  | "None" => mkStmtExprMd (.StaticCall "Any..isfrom_none" [freeVar var])
+  | _ => if ty.startsWith "Dict" then mkStmtExprMd (.StaticCall "Any..isfrom_Dict" [freeVar var])
```
joehendrix: String prefix matching here is fragile — `ty.startsWith "Dict"` would match a user-defined class like `DictionaryHelper`, and similarly `"List"` would match `ListManager`.

The root cause is that `argumentTypeToString` (line 1222) flattens the Python AST type expression into `List String`, discarding the structure that's already there. For example, `List[str]` is a `Subscript` node with the base name and element type as children — but after `argumentTypeToString` it's just the string `"List[str]"` and we're left doing prefix matching to recover what was known.

It would be more robust to keep the Python AST nodes (or a structured intermediate representation like the pyspec pipeline's `SpecAtomType`) and match on those directly. That way `List[str]` vs bare `List` vs a class named `ListManager` are unambiguously distinguishable.

Separately, bare `List` (no element type) maps to `isfrom_ListAny`, but `ListAny` elements are value-world `Any` which can't hold composites. If a bare `List` could contain user-defined classes, the constraint would be unsound.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1204,45 +1214,59 @@ partial def getNestedSubscripts (expr:  Python.expr SourceRange) : List ( Python
...
+          | .mk_arg sr name oty _ =>
+            let md := sourceRangeToMetaData ctx.filePath sr
             match oty.val with
-              | .some ty => return (name.val, ← argumentTypeToString ty, a.snd)
-              | _ => return (name.val, PyLauType.Any, a.snd))
+              | .some ty =>
+                let defaultType := match a.snd.mapM (inferExprType ctx) with
+                  | .ok (some ty) => [ty]
+                  | _ => []
+                return {name:= name.val, md:=md, tys:=(← argumentTypeToString ty) ++ defaultType, default:= a.snd}
```
joehendrix: The inferred type of the default value is appended to the declared annotation types. For the test case `def Mul(y: int | bool = "abc")`, `y.tys` becomes `["int", "bool", "str"]` — the constraint permits `str` even though the annotation says `int | bool`.

This silently accommodates type errors in default values rather than catching them. Was this intentional? If the goal is robustness, a warning when the default type doesn't match the annotation might be more appropriate than widening the constraint.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1258,23 +1282,60 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
...
+  | "str" => mkStmtExprMd (.StaticCall "Any..isfrom_string" [freeVar var])
+  | "int" => mkStmtExprMd (.StaticCall "Any..isfrom_int" [freeVar var])
+  | "bool" => mkStmtExprMd (.StaticCall "Any..isfrom_bool" [freeVar var])
+  | "datetime" => mkStmtExprMd (.StaticCall "Any..isfrom_datetime" [freeVar var])
+  | "None" => mkStmtExprMd (.StaticCall "Any..isfrom_none" [freeVar var])
+  | _ => if ty.startsWith "Dict" then mkStmtExprMd (.StaticCall "Any..isfrom_Dict" [freeVar var])
+      else if ty.startsWith "List" then mkStmtExprMd (.StaticCall "Any..isfrom_ListAny" [freeVar var])
+      else none
+
+def creatBoolOrExpr (exprs: List StmtExprMd) : StmtExprMd :=
```
joehendrix: Nit: typo `creatBoolOrExpr` → `createBoolOrExpr`.

Also — Laurel has `PrimitiveOp.Or` for boolean disjunction, which doesn't depend on name resolution. `StaticCall "Bool.Or"` works (the resolver treats unresolved names as functions, and the Core SMT encoder handles `Bool.Or`), but `PrimitiveOp.Or` is what preconditions use elsewhere. Not a blocker, just a consistency note.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1258,23 +1282,60 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
...
+          { name := arg.name, type := mkHighTypeMd (.UserDefined {text:= arg.tys[0]!}) }
         else
-          { name := name, type := AnyTy})
+          { name := arg.name, type := AnyTy})
     if funcDecl.hasKwargs then
       let paramType ← translateType ctx PyLauType.DictStrAny
       inputs:= inputs ++ [{ name := "kwargs", type := paramType }]
 
-    -- Translate return type
+    let typeConstraintAssertions := getInputTypecheckAssertions funcDecl
```
joehendrix: The PR generates both body assertions and preconditions for every typed parameter. For transparent bodies, the assertion is trivially satisfied by the precondition — the verifier proves the same thing twice.

The PR description says this is "in case we want to inline the procedure call." Could you clarify when inlining would need the assertion but not have access to the precondition? If this is speculative, consider gating assertion generation behind a flag to avoid doubling VCs per parameter.

StrataTest/Languages/Python/tests/test_func_input_type_constraints.py
```diff
@@ -0,0 +1,10 @@
+def Mul(x: int | bool, y: int | bool = "abc") -> int:
+    return x * y 
+
+def Sum(x: Union[int , bool], y: Union[int , bool] = None) -> int:
```
joehendrix: Nit: `Union` is used without `from typing import Union`. The Python AST parser doesn't validate imports, but the test file isn't valid Python as-is. Adding the import would make it runnable for debugging.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I remove assertions because currently we are not inlining procedure calls and procedure call inlining will be updated to automatically turn preconditions to assertions.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I added a check at line 1279, which will throw an error if a composite type appears in a union of types (i.e. we only accept a single composite type or union of non-composite types)

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I removed that with an exception that if default is `None` that `None` is added to the type list. The translator will throw an error when default's type does not match the input types.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I moved it to line 1341.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I fixed that and also added the checking that the List's element type is among those wrapped by Any type.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -807,18 +821,18 @@ partial def combinePositionalAndKeywordArgs
     let kwords := pyKwordsToHashMap kwords
     let unprovidedPosArgs := funcDecl.args.drop posArgs.length
     --every unprovided positional args must have a default value in the function signature or be provided in the kwargs
-    let missingArgs := unprovidedPosArgs.filter fun (name, _, d) =>
-      !(name ∈ kwords.keys) && d.isNone
+    let missingArgs := unprovidedPosArgs.filter fun arg =>
+      !(arg.name ∈ kwords.keys) && arg.default.isNone
     if missingArgs.length > 0 then
       let missingNames := missingArgs.map (·.1)
```
MikaelMayer: Bug: `missingArgs` is now `List PyArgInfo` (a structure), but `(·.1)` is tuple projection syntax. This should be `(·.name)` to match the refactored type.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1418,6 +1485,39 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
...
+  | expr::exprs => mkStmtExprMd (.PrimitiveOp .Or [expr, createBoolOrExpr exprs])
+
+def getUnionTypeConstraint (var: String) (md: MetaData) (tys: List String) (funcname: String): Option StmtExprMd :=
+  let type_constraints := tys.filterMap (getSingleTypeConstraint var)
+  if type_constraints.isEmpty then none else
+    let md: MetaData := md.withPropertySummary $ "(" ++ funcname ++ " requires) Type constraint of " ++ var
+    some {createBoolOrExpr type_constraints with md:=md}
+
+def getReturnTypeEnsure (md: MetaData) (tys: List String) (funcname: String): Option StmtExprMd :=
+  let type_constraints := tys.filterMap (getSingleTypeConstraint PyLauFuncReturnVar)
```
MikaelMayer: `getReturnTypeEnsure` is nearly identical to `getUnionTypeConstraint` — the only differences are the variable name and the metadata summary string. Consider implementing it as:
```lean
def getReturnTypeEnsure (md: MetaData) (tys: List String) (funcname: String): Option StmtExprMd :=
  getUnionTypeConstraint PyLauFuncReturnVar md tys funcname
  |>.map fun c => {c with md := md.withPropertySummary $ "(" ++ funcname ++ " ensures) Return type constraint"}
```
This avoids duplicating the `filterMap`/`createBoolOrExpr` logic.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1418,6 +1485,39 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
...
+  if type_constraints.isEmpty then none else
+    let md: MetaData := md.withPropertySummary $ "(" ++ funcname ++ " ensures) Return type constraint"
+    some {createBoolOrExpr type_constraints with md:=md}
+
+def getInputTypePreconditions (funcDecl : PythonFunctionDecl): List StmtExprMd :=
+  funcDecl.args.filterMap (λ arg => getUnionTypeConstraint arg.name arg.md arg.tys funcDecl.name)
```
MikaelMayer, line 1515: Missing blank line between `getInputTypePreconditions` and the docstring for `translateFunctionBody`. The docstring currently looks like it belongs to `getInputTypePreconditions`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1437,42 +1537,43 @@ def translateFunction (ctx : TranslationContext) (sourceRange: SourceRange) (fun
...
+      match funcDecl.ret.map fun (tys, md) => getReturnTypeEnsure md tys funcDecl.name with
+        | some (some constraint) => [constraint]
+        | _ => []
     -- Declare an output parameter when the function has a return type annotation.
     -- Return statements explicitly assign to LaurelResult and exit $body.
-    let outputs : List Parameter :=
-      match funcDecl.ret with
-      | none => []
-      | some _ => [{ name := "LaurelResult", type := AnyTy }]
+    let outputs : List Parameter := [{ name := PyLauFuncReturnVar, type := AnyTy }]
```
MikaelMayer, line 1554: Stale comment: says "Declare an output parameter when the function has a return type annotation" but the code now unconditionally declares the output. The comment should reflect that all functions now get a `LaurelResult` output.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1363,52 +1386,96 @@ partial def getNestedSubscripts (expr:  Python.expr SourceRange) : List ( Python
...
+    | "Union" =>  match sliceHead with
+        | .Tuple _ tys _ => return (← tys.val.toList.mapM getArgumentTypes).flatten
+        | _ => throw (.internalError s!"Unhandled Expr: {repr arg}")
+    | "List" => return ["ListAny"]
+    | "Dict" => return ["DictStrAny"]
+    | _ =>  throw (.internalError s!"Unhandled Expr: {repr arg}")
+  | .Constant _ _ _ => return ["None"]
+  | .Attribute _ _ _ _ => return [pyExprToString arg]
+  | .BinOp _ _ _ _ => return (← (getUnionTypes arg).mapM getArgumentTypes).flatten
+  | .Slice _ lower upper _ => match lower.val, upper.val with
```
MikaelMayer, line 1431: The `Slice` case is non-obvious — it matches `some (.Name _ _ _), some upper` and discards the lower bound. A brief comment explaining what Python AST pattern this handles (e.g., `Dict[str, Any]` being represented as a Slice node) would help future readers.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1363,52 +1386,96 @@ partial def getNestedSubscripts (expr:  Python.expr SourceRange) : List ( Python
   | .Subscript _ val slice _ => [val] ++ (getNestedSubscripts slice)
   | _ => [expr]
 
-partial def argumentTypeToString (arg: Python.expr SourceRange) : Except TranslationError String :=
+def getUnionTypes (arg: Python.expr SourceRange) : List (Python.expr SourceRange) :=
+  match arg with
+  | .BinOp _ left _ right => getUnionTypes left ++ getUnionTypes right
+  | _ => [arg]
+
```
MikaelMayer, line 1401: The `isOfAnyType` allowlist and `getSingleTypeConstraint` match list are closely related but maintained separately. If a new type is added to one but not the other, it could silently produce incorrect behavior. Consider deriving one from the other, or at least adding a comment linking them.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -54,12 +54,19 @@ inductive UnmodeledFunctionBehavior where
   | havocInputsAndOutputs
 deriving Inhabited
 
+structure PyArgInfo where
```
MikaelMayer, line 57: No action needed: `PyArgInfo` is a clean abstraction that replaces the previous `String × String × Option (...)` tuple. This makes the code much more readable throughout.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: An easy fix is to create a theorem (in a separate file) that proves that `isOfAnyType` implies that `getSingleTypeConstraint` is not none.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1439,42 +1533,44 @@ def translateFunction (ctx : TranslationContext) (sourceRange: SourceRange) (fun
...
     let proc : Procedure := {
       name := funcDecl.name
       inputs := inputs
       outputs := outputs
-      preconditions := []
+      preconditions := typeConstraintPreconditions
       determinism := .deterministic none -- TODO: need to set reads
       decreases := none
-      body := Body.Transparent bodyBlock
+      body := Body.Opaque typeConstraintPostcondition bodyBlock []
```
tautschnig, line 1573: Changing from `Body.Transparent bodyBlock` to `Body.Opaque typeConstraintPostcondition bodyBlock []` means the function body is now opaque with postconditions. This is a significant semantic change — callers will reason about the function through its contract (pre/postconditions) rather than by inlining the body. The PR description mentions this but it's worth confirming: is this the intended behavior for *all* user-defined Python functions, or should some remain transparent?

thanhnguyen-aws: Currently in Laurel, `Body.Transparent` and `Body.Opaque` differ only in that `Body.Transparent` does not support postconditions. In practice, `Body.Transparent` is not inlined at call sites — it is treated the same as Body.Opaque without a postcondition — so there is no significant semantic difference between the two at this time.

PR #580, opened 2026-03-14
Add `with` statement support in Python->Laurel translation

ssomayyajula:
*Description of changes:*

(written by Kiro CLI)

Translate Python with statements to __enter__/__exit__ instance calls. For each withitem, a manager variable is declared from the context expression, __enter__ is called (binding the result if 'as VAR' is present), the body is translated, then __exit__ is called.

Also fix translateType to recognize user-defined composite types and return UserDefined instead of falling through to AnyTy. This was a pre-existing gap that caused type unification errors when composite-typed variables flowed through translateType.

Add test covering with...as, with (no as), and multiple context managers.

### DIFF
--- a/Strata/DDM/AST.lean
+++ b/Strata/DDM/AST.lean
@@ -579,86 +579,71 @@ instance : GetElem? Metadata QualifiedIdent (Array MetadataArg) (fun md i => i �
     | none => none
     | some a => a.args
 
-private def scopeIndex (metadata : Metadata) : Option Nat :=
+private def scopeIndex (metadata : Metadata) : Except String (Option Nat) :=
   match metadata[MetadataAttr.scopeName]? with
-  | none => none
-  | some #[.catbvar idx] => some idx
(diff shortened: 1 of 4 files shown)


StrataTest/Languages/Python/run_py_analyze_sarif.py
```diff
@@ -20,7 +20,7 @@
 )
 
 BOTH_SKIP = {"test_foo_client_folder", "test_invalid_client_type", "test_unsupported_config"}
-SKIP_TESTS = BOTH_SKIP | {"test_class_field_use", "test_list", "test_subscription"}
+SKIP_TESTS = BOTH_SKIP | {"test_class_field_use", "test_list", "test_subscription", "test_with_statement"}
```
MikaelMayer, line 23: Is it skipping the test? I've seen Kiro do this instead of fixing the test in the past.

ssomayyajula: I think June resolved this since SKIP_TESTS is just for the original ->Core pipeline

PR #581, opened 2026-03-15
feat(Core): add loop termination checking via `decreases` measure annotations

shigoel:
Introduces support for `decreases` clauses on `while` loops in the Core
language: concrete syntax, passive encoding, CFG visualization, and verification
condition generation.

Before this PR, Core's AST supported the provision of a measure for a loop, 
but didn't expose it (no concrete syntax) and the analysis didn't check 
loop termination.

- Loop termination VCs (LoopElim.lean)
- `decreases` concrete syntax (DDMTransform)
- New generic abstractions (DL/Imperative layer): `HasIntOrder`, `HasIdent`,
`HasInit`
- CFG visualization of measures (StructuredToUnstructured.lean)
- PrecondElim support (PrecondElim.lean)

Also added a format instance for Core's experssions that shadows the generic
`LExpr` formatter with `Core.formatExprs`.

*Tests updated in Loops.lean*

- New `measureFailExamplePgm`: `decreases n` on a `countUp` loop (wrong
  measure); confirms `measure_decrease_0` fails while other VCs pass.
- `gaussPgm` and `nestedPgm` annotated with measures.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/Strata/Transform/PrecondElim.lean
+++ b/Strata/Transform/PrecondElim.lean
@@ -236,6 +236,9 @@ def transformStmt (s : Statement)
     let measureAsserts := match measure with
       | none => []
       | some m => collectPrecondAsserts F m "loop_measure" md
+    let measureAssertsEnd := match measure with
+      | none => []
+      | some m => collectPrecondAsserts F m "loop_measure_end" md
     let invAsserts := invariant.flatMap (fun inv => collectPrecondAsserts F inv "loop_invariant" md)
     let guardAsserts := collectPrecondAsserts F guard "loop_guard" md
     let guardAssertsEnd := collectPrecondAsserts F guard "loop_guard_end" md
@@ -244,7 +247,7 @@ def transformStmt (s : Statement)
     setFactory savedF
     return (changed || !invAsserts.isEmpty || !guardAsserts.isEmpty || !measureAsserts.isEmpty,
       guardAsserts ++ invAsserts ++ measureAsserts ++
-      [.loop guard measure invariant (body' ++ guardAssertsEnd) md])
+      [.loop guard measure invariant (body' ++ measureAssertsEnd ++ guardAssertsEnd) md])
   | .exit lbl md =>
     return (false, [.exit lbl md])
   | .funcDecl decl md => do
--- a/StrataTest/Languages/Core/Examples/Loops.lean
+++ b/StrataTest/Languages/Core/Examples/Loops.lean
@@ -516,6 +516,10 @@ Obligation: measure_lb_0
 Property: assert
 Result: ✅ pass
 
+Obligation: loop_measure_end_calls_Int.SafeDiv_0
+Property: division by zero check
+Result: ✅ pass
+
 Obligation: arbitrary_iter_maintain_invariant_0_0
 Property: assert
 Result: ✅ pass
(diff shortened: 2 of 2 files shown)


Strata/Languages/Core/StatementSemantics.lean
```diff
@@ -37,6 +37,15 @@ instance : HasFvar Core.Expression where
 instance : HasSubstFvar Core.Expression where
   substFvar := Lambda.LExpr.substFvar
 
+instance : HasIntOrder Core.Expression where
```
joscoh, line 40: Just curious, this means that all `decreases` clauses must be nat-valued, right?

Strata/Transform/PrecondElim.lean
```diff
@@ -233,16 +233,17 @@ def transformStmt (s : Statement)
...
     let invAsserts := invariant.flatMap (fun inv => collectPrecondAsserts F inv "loop_invariant" md)
     let guardAsserts := collectPrecondAsserts F guard "loop_guard" md
     let guardAssertsEnd := collectPrecondAsserts F guard "loop_guard_end" md
     let savedF ← getFactory
     let (changed, body') ← transformStmts body
     setFactory savedF
-    return (changed || !invAsserts.isEmpty || !guardAsserts.isEmpty,
-      guardAsserts ++ invAsserts ++
+    return (changed || !invAsserts.isEmpty || !guardAsserts.isEmpty || !measureAsserts.isEmpty,
+      guardAsserts ++ invAsserts ++ measureAsserts ++
```
joscoh, line 249: I think this needs the same treatment as `guardAsserts` - it must be checked at the beginning and then again after each loop iteration. (e.g. if we have something like `while e decreases (y / x) { x := ...}`)

StrataTest/Languages/Core/Examples/Regex.lean
```diff
@@ -183,7 +183,8 @@ str.in.re("a", bad_re_loop(1))
 Result: Obligation: assert_0
 Property: assert
 Result: 🚨 Implementation Error! SMT Encoding Error! Natural numbers expected as indices for re.loop.
-Original expression: (~Re.Loop (~Re.Range #a #z) #1 %0)
+Original expression: re.loop(re.range("a", "z"), 1, bvar!0) -- Errors: Unsupported construct in lexprToExpr: bvar index out of bounds: 0
```
joscoh, line 186: Why does termination checking change this?

Strata/Languages/Core/StatementSemantics.lean
shigoel: For now, yes.

StrataTest/Languages/Core/Examples/Regex.lean
shigoel: It doesn't, but I slipped in a default format instance for Core's expressions that uses the AST -> CST transform. Now when we print Core's expressions, we print in DDM syntax.

shigoel: Did that to clean up Core -> CFG output, and that pass is affected by the addition of termination checking.

Strata/Transform/PrecondElim.lean
shigoel: Oops, you're right, thanks! I'll fix this.

shigoel: Fixed in https://github.com/strata-org/Strata/pull/581/commits/e0961bc09f4e82852e75c51956504ab9735787a3.

PR #584, opened 2026-03-16
feat(laurel): Add eliminateHoles Laurel→Laurel pass and wire into pipeline

olivier-aws:
### Summary

Holes (`<?>`, `<??>`) represent unknown parts of a program (e.g. from unsupported Python constructs). This PR adds contextual type inference for holes and translates them soundly for both procedures and functions, with support for both deterministic and nondeterministic semantics.

### Deterministic vs nondeterministic holes

| Syntax | Semantics | Translation | Allowed in |
|--------|-----------|-------------|------------|
| `<?>` | Deterministic: same inputs _ same output | Uninterpreted function call | Procedures and functions |
| `<??>` | Nondeterministic: can return different values each time | Havoced local variable | Procedures only |

**Deterministic holes** are replaced by calls to freshly generated uninterpreted functions whose parameters mirror the enclosing callable's inputs. This is sound because uninterpreted functions are deterministic (required for functions) but unconstrained (the SMT solver can assign any interpretation).

**Nondeterministic holes** are lifted to havoced local variables (fresh variable declaration with no initializer) by the existing impure expression lifter. They are rejected in functions since functions must be deterministic.

### Architecture

Three passes handle holes in sequence:

1. **`inferHoleTypes`** (`InferHoleTypes.lean`): Walks the AST and annotates each `Hole` with a type inferred from its surrounding context using the `SemanticModel` and `computeExprType`. Pure type annotation _ no structural changes. Handles both det and nondet holes.

2. **`eliminateHoles`** (`EliminateHoles.lean`): Replaces **deterministic** holes with calls to freshly generated uninterpreted functions. Nondeterministic holes pass through unchanged. After this pass, no deterministic holes remain.

3. **`liftExpressionAssignments`** (`LiftImperativeExpressions.lean`): The existing impure expression lifter, extended to handle **nondeterministic** holes. When `transformExpr` encounters a nondet hole, it generates a fresh variable declaration with no initializer (havoc) and replaces the hole with a variable reference.

Pipeline:
```
inferHoleTypes _ eliminateHoles _ resolve _ liftExpressionAssignments _ translate
```

After all three passes, the program contains no `Hole` nodes.

### Type inference

The type of each hole is inferred from its surrounding context:

| Context | Inferred type |
|---------|--------------|
| `var x: int := 1 + <?>` | `int` (from `computeExprType` on the `Add` expression) |
| `assert <?> > 0` | `int` (from `computeExprType` on the non-hole sibling `0`) |
| `assert <?>` / `assume <?>` | `bool` |
| `if (<?>) ...` / `while(<?>) ...` | `bool` |
| `while(...) invariant <?>` | `bool` |
| `forall(i: int) => <?>` | `bool` |
| `if (true) <?> else 0` | inherits expected type from parent |
| `f(1, <?>)` with known `f(a: int, b: int)` | `int` (from callee parameter type) |
| `return <?>` with `returns(result: int)` | `int` (from procedure output type) |
| `assert n > <?>` with `n: int` | `int` (from resolved variable type via model) |
| Unknown context | `_` (Top) |

### AST change

`StmtExpr.Hole` now carries two fields:
```
| Hole (deterministic : Bool := true) (type : Option HighTypeMd := none)
```

### Grammar

Two new operators in `LaurelGrammar.st`:
```
op hole: StmtExpr => "<?>";
op nondetHole: StmtExpr => "<??>";
```

### Files

| File | Change |
|------|--------|
| `InferHoleTypes.lean` | New: hole type inference pass |
| `EliminateHoles.lean` | New: deterministic hole elimination pass |
| `Laurel.lean` | `Hole` gains `deterministic` and `type` fields |
| `LaurelGrammar.st` | Add `<?>` and `<??>` syntax |
| `ConcreteToAbstractTreeTranslator.lean` | Add `hole` and `nondetHole` cases |
| `LiftImperativeExpressions.lean` | Handle nondet holes; process assert/assume/return/while-invariants for nondet lifting |
| `LaurelToCoreTranslator.lean` | Import new passes, update pipeline, reject remaining holes |
| `LaurelTypes.lean` | `computeExprType` uses annotated hole type |
| `LaurelFormat.lean` | Format `<?>` and `<??>` |
| `Resolution.lean` | Preserve `deterministic` flag |
| `LaurelEval.lean` | Update pattern |
| `LiftHolesTest.lean` | Parser-based tests for both det and nondet holes |

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Strata/Languages/Laurel/Resolution.lean
+++ b/Strata/Languages/Laurel/Resolution.lean
@@ -159,6 +159,9 @@ structure ResolveState where
   typeScopes : TypeScopes := {}
   /-- Diagnostics collected during resolution. -/
   errors : Array DiagnosticModel := #[]
+  /-- When resolving inside an instance procedure, the owning composite type name.
+      Used by `resolveFieldRef` to resolve `self.field` when `self` has type `Any`. -/
+  instanceTypeName : Option String := none
 
 @[expose] abbrev ResolveM := StateM ResolveState
 
@@ -210,21 +213,31 @@ private def targetTypeName (target : StmtExprMd) : ResolveM (Option String) := d
     | none => pure none
   | _ => pure none
 
+/-- Try to resolve a field name via a type scope lookup. Returns `some id` on success. -/
+private def resolveFieldInTypeScope (typeName : String) (fieldName : Identifier) : ResolveM (Option Identifier) := do
+  let s ← get
+  match s.typeScopes.get? typeName with
+  | some typeScope =>
+    match typeScope.get? fieldName.text with
+    | some (defId, _) => return some { fieldName with uniqueId := some defId }
+    | none => return none
+  | none => return none
+
 /-- Resolve a field reference using the target's type to build a qualified lookup key.
-    Falls back to unqualified lookup if the target type cannot be determined. -/
+    Falls back to the instance type name (for `self.field` in instance methods),
+    then to unqualified lookup if the target type cannot be determined. -/
 def resolveFieldRef (target : StmtExprMd) (fieldName : Identifier)
     (md : Imperative.MetaData Core.Expression) : ResolveM Identifier := do
   let typeName? ← targetTypeName target
-  match typeName? with
-  | some typeName =>
-    let s ← get
-    match s.typeScopes.get? typeName with
-    | some typeScope =>
-      match typeScope.get? fieldName.text with
-      | some (defId, _) => return { fieldName with uniqueId := some defId }
-      | none => resolveRef fieldName md
-    | none => resolveRef fieldName md
-  | none => resolveRef fieldName md
+  -- Try type scope from the target's declared type
+  if let some typeName := typeName? then
+    if let some resolved ← resolveFieldInTypeScope typeName fieldName then
+      return resolved
+  -- Fallback: use the owning instance type (handles `self.field` when self has type `Any`)
+  if let some instTypeName := (← get).instanceTypeName then
+    if let some resolved ← resolveFieldInTypeScope instTypeName fieldName then
+      return resolved
+  resolveRef fieldName md
 
 /-- Save and restore scope around a block (for lexical scoping). -/
 def withScope (action : ResolveM α) : ResolveM α := do
@@ -445,12 +458,15 @@ def resolveField (ownerName : Identifier) (field : Field) : ResolveM Field := do
 def resolveInstanceProcedure (typeName : Identifier) (proc : Procedure) : ResolveM Procedure := do
   let procName' ← defineName proc.name (.instanceProcedure typeName proc)
   withScope do
+    let savedInstType := (← get).instanceTypeName
+    modify fun s => { s with instanceTypeName := some typeName.text }
     let inputs' ← proc.inputs.mapM resolveParameter
     let outputs' ← proc.outputs.mapM resolveParameter
     let pres' ← proc.preconditions.mapM resolveStmtExpr
     let det' ← resolveDeterminism proc.determinism
     let dec' ← proc.decreases.mapM resolveStmtExpr
     let body' ← resolveBody proc.body
+    modify fun s => { s with instanceTypeName := savedInstType }
     return { name := procName', inputs := inputs', outputs := outputs',
              isFunctional := proc.isFunctional,
              preconditions := pres', determinism := det', decreases := dec',
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -1275,14 +1275,18 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
           .ok (some (funcDecl))
       | _ => .ok none)
     let ctx := {ctx with functionSignatures:= ctx.functionSignatures ++ classFunDecls}
-    -- Extract fields from __init__ method body
-    let mut fields : List Field := []
+    -- Extract fields from class-level annotations (e.g. `x: int`)
+    let mut fields ← extractClassFields ctx body.val
+    -- Also extract fields from __init__ method body (e.g. `self.x: int = ...`)
     for stmt in body.val do
       match stmt with
       | .FunctionDef _ name _ initBody _ _ _ _ =>
         if name.val == "__init__" then
           let initFields ← extractFieldsFromInit ctx initBody.val
-          fields := fields ++ initFields
+          -- Only add fields not already declared at class level
+          for f in initFields do
+            unless fields.any (fun existing => existing.name.text == f.name.text) do
+              fields := fields ++ [f]
       | _ => pure ()
 
     -- Extract methods from class body


Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -434,5 +434,205 @@ def liftExpressionAssignments (model: SemanticModel) (program : Program) : Progr
...
+  counter : Nat := 0
+
+private abbrev HoleLiftM := StateM HoleLiftState
+
+private def freshHoleVar : HoleLiftM Identifier := do
+  let n := (← get).counter
+  modify fun s => { s with counter := n + 1 }
+  return s!"$hole_{n}"
+
+private def holeType : HighTypeMd := bareType (.TCore "Any")
```
keyboardDrummer: A hole can have different types than Any though. I think the type of Hole should be inferred from its context.

olivier-aws: I updated quite a lot the code to have the semantic model and infer the best possible type

Strata/Languages/Laurel/Grammar/LaurelGrammar.st
```diff
@@ -17,6 +17,7 @@ op literalBool (b: Bool): StmtExpr => b;
 op int(n : Num) : StmtExpr => n;
 op real(d : Decimal) : StmtExpr => d;
 op string (s: Str): StmtExpr => s;
+op hole: StmtExpr => "<?>";
```
MikaelMayer, line 20: Oh interesting, I thought holes already had a representation. Is this syntax everyone in the Laurel team likes? Other languages have `???` or `...` so just double-checking about the design choice in case it was just inferred by A.I.

olivier-aws: This is what is used by the formal method in `LaurelFormal.lean`

keyboardDrummer: `???` or `...` seem better than `<?>` since they exist in other languages (Scala and Python). I don't have a strong opinion on the syntax but and existing one seems better than a new one.

keyboardDrummer:
> I think this will only support holes that occur in a procedure context, although maybe that's fine. Would it also be possible to use a mechanism that also works when holes occur in functions? What about creating a global variable in Core without an initializer and referencing that with the hole?

I'm guessing the above idea would not work in the context of recursive procedures and inlining, since then different invocations of the same function would use the same value for the hole. @MikaelMayer is that correct? Btw, will it even be possible to make procedures transparent (encode their bodies using SMT functions) if they contain uninitialized variables?

Strata/Languages/Laurel/LiftHoles.lean
```diff
@@ -0,0 +1,285 @@
...
+private def emptyMd : Imperative.MetaData Core.Expression := #[]
+private def bare (v : StmtExpr) : StmtExprMd := ⟨v, emptyMd⟩
+private def bareType (v : HighType) : HighTypeMd := ⟨v, emptyMd⟩
+
+/-- State for the hole lifting pass. -/
+structure HoleLiftState where
+  counter : Nat := 0
+  /-- Semantic model for type inference. -/
+  model : SemanticModel
+  /-- All procedures in the program (fallback for unresolved identifiers). -/
```
keyboardDrummer: `(fallback for unresolved identifiers)` why did you add this? It doesn't make sense to me. If resolution failed to resolve a reference then no alternative should succeed either.

Strata/Languages/Laurel/LiftHoles.lean
```diff
@@ -0,0 +1,285 @@
+/-
```
keyboardDrummer: Have you considered having a separate type inference pass, where you assign a type to each hole (we would add a `type: Option HighTypeMd` field to `.Hole`), and doing the lifting of holes in the existing lifting pass? The lifting part seems equivalent for holes and other impure expressions, so having it again in this pass seems duplicative.

Also, a type inference pass could be extended in the future to work for things other than holes.

olivier-aws: I did not consider this alternative. I think this is a good idea, as it avoids some duplication in the lifting phase. It also allows for the front-end to generate a Hole directly with the type, which they might infer more easily. I'll look into this.

Strata/Languages/Laurel/LiftHoles.lean
olivier-aws: This is probably not needed, left over of an old attempt that did not have the SemanticModel. Will clean up

olivier-aws: Removed

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -300,8 +300,9 @@ inductive StmtExpr : Type where
   | Abstract
   /-- Refers to all objects in the heap. Used in reads or modifies clauses. -/
   | All
-  /-- A hole with Top type, useful for partially available programs. -/
-  | Hole
+  /-- A hole representing an unknown expression. The optional type is inferred
+      by the hole type inference pass; `none` means the type has not been inferred yet. -/
+  | Hole (type : Option (WithMetadata HighType) := none)
```
keyboardDrummer: Maybe we could have assigned the default `Top` type right away so this would not have needed to be an `Option`, but this is fine as well : )

keyboardDrummer (changes requested):
Sorry to retract my approval, but I think @MikaelMayer's concerns related to determinism vs non-determinism are valid. Could you add a property `deterministic: Bool` to .Hole, and only eliminate the deterministic ones? The non-det ones should be lifted.

StrataTest/Languages/Laurel/LiftHolesTest.lean
```diff
@@ -0,0 +1,364 @@
...
+import Strata.Languages.Laurel.EliminateHoles
+import Strata.Languages.Laurel.LaurelFormat
+
+open Strata
+open Strata.Elab (parseStrataProgramFromDialect)
+
+namespace Strata.Laurel
+
+/-- Parse a Laurel source string, resolve, eliminate holes, and print all procedures. -/
+private def parseElimAndPrint (input : String) : IO Unit := do
```
keyboardDrummer, line 26: Minor: a test that compares against generated Core is brittle. I think ideally we would have an end-to-end test. Something like:
```
transparent procedure detHoleTest(x: int): int {
  var y: int := <?>
  assert x == y; // error
  assert x != y; // error
  assert exists: z: int :: z == y // pass
  y
}

function detHoleFunctionTest(x: int): int {
  ..same as before...
}
```

and then something related to determinism vs non-det:
```
transparent procedure nondetHole(x: int): int {
  var y: int := <???>
  ... same asserts as before ...
  y
}

procedure testDetVsNonDet() {
  assert detHoleTest() == detHoleTest(); // pass
  assert nondetHole() == nondetHole(); // error
  assert nondetHole() != nondetHole(); // error
}
```

Note that we don't support `transparent procedure` yet so the testDetVsNonDet can't be expressed

olivier-aws: Will do that in a follow-up PR if that's OK.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -306,14 +329,25 @@ def transformStmt (stmt : StmtExprMd) : LiftM (List StmtExprMd) := do
...
   | WithMetadata.mk val md =>
   match val with
-  | .Assert _ =>
-      -- Do not transform assert conditions: impure expressions inside contracts
-      -- must be rejected by the translator, not silently lifted.
-      return [stmt]
+  | .Assert cond =>
+      -- Do not transform assert conditions with assignments — they must be rejected.
+      -- But nondeterministic holes need to be lifted.
+      if containsNondetHole cond && !containsAssignmentOrImperativeCall (← get).model cond then
```
keyboardDrummer, line 335: I think the previous code of `return [stmt]` should be good also for this PR. I think non-det holes in contracts should not be valid, and should not be lifted, since we don't allow non-determinism in contracts.

keyboardDrummer: How come the test "Mixed: det hole eliminated, nondet hole preserved." that contains `assert <??>` does not lift the hole from the assertion?

olivier-aws: It's because the test only checks the `eliminateHoles` pass, not the combination `eliminateHoles` + `liftExpressionAssignments` that handles both.
If that's OK with you I'll complete the testing strategy with a combination of both in a follow-up PR

PR #588, opened 2026-03-16
Change some `panic`s in DDM to use `Except`

joscoh:
*Issue #, if available:*

*Description of changes:* Changes several definitions in the DDM related to `preRegisterTypes` to use the `Except` monad instead of `panic!`. A few `panic`s are added, but only in functions that already have `panic`s.

### DIFF
--- a/Strata/DL/Lambda/LExpr.lean
+++ b/Strata/DL/Lambda/LExpr.lean
@@ -589,9 +589,18 @@ private def formatLExpr (T : LExprParamsT) [ToFormat T.base.IDMeta] [ToFormat T.
     match ty with
     | none => f!"{x}"
     | some ty => f!"({x} : {ty})"
-  | .abs _ _ _ e1 => Format.paren (f!"λ {formatLExpr T e1}")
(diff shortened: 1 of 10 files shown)


Strata/DDM/Elab/LoadedDialects.lean
```diff
@@ -74,21 +74,29 @@ def addDialect! (loader : LoadedDialects) (d : Dialect) : LoadedDialects :=
   | .error msg =>
     @panic _ ⟨loader⟩ s!"Could not add open dialect: {eformat msg |>.pretty}"
   | .ok parsers =>
+    match loader.syntaxElabMap.addDialect d with
+    | .error msg =>
+      @panic _ ⟨loader⟩ s!"Could not add dialect syntax elaborators: {msg}"
```
MikaelMayer, line 79: Is that a disguised way of introducing a panic?

joscoh: This is a panic, and it is not intended to be disguised but just consistent with the other pre-existing panics already returned by this function.

shigoel (approved):

PR #590, opened 2026-03-17
feat(laurel): short-circuit AndThen/OrElse, eager And/Or, forLoop, javaGen improvements

fabiomadge:
### Short-circuit and eager boolean operators
- Add `AndThen`/`OrElse` operations (short-circuit via `ite` in Core)
- Keep `And`/`Or` as eager (`boolAndOp`/`boolOrOp`)
- Grammar: `&&`/`||` map to `AndThen`/`OrElse`, `&`/`|` map to `And`/`Or`
- `Implies` uses short-circuit (`ite a b true`)
- Desugar `AndThen`/`OrElse`/`Implies` to `IfThenElse` before expression lifting when the second operand contains imperative calls, preserving short-circuit semantics
- Tests cover pure path (function with `requires false`, division by zero) and imperative path (procedure with `requires false`) for all three operators

### For loop
- Add `forLoop` grammar op: `for(init; cond; step) invariants body`
- Desugars to `Block[init, While(cond, invariants, Block[body, step])]` in ConcreteToAbstractTreeTranslator
- Test in T17_ForLoop.lean

### javaGen CLI improvements
- Accepts a dialect **name** (e.g. `Laurel`) in addition to a file path
- Integration test for preloaded dialect path

### Solver config
- Extract `laurelVerifyOptions` to deduplicate z3 solver config across Laurel commands

### Windows fix
- Fix SMT solver output parsing on Windows where z3 outputs `\r\n` line endings
- `parseVerdict` now calls `.trimAscii.toString` (regression from #487 which dropped the `.toString`)

### DIFF
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -11,7 +11,10 @@ import Strata.Languages.Core.SarifOutput
 import Strata.Languages.Laurel.Grammar.ConcreteToAbstractTreeTranslator
 import Strata.Languages.Laurel.LaurelToCoreTranslator
 import Strata.Languages.Python.Python
+import Strata.Languages.Python.PySpecPipeline
 import Strata.Languages.Python.Specs
+import Strata.Languages.Python.Specs.DDM
+import Strata.Languages.Python.Specs.IdentifyOverloads
 import Strata.Languages.Python.Specs.ToLaurel
 import Strata.Languages.Laurel.LaurelFormat
 import Strata.Languages.Laurel.Laurel
@@ -26,11 +29,21 @@ import Strata.SimpleAPI
 
 open Strata
 
-open Core (VerifyOptions VerboseMode)
+open Core (VerifyOptions VerboseMode VerificationMode CheckLevel)
 
 def exitFailure {α} (message : String) (hint : String := "strata --help") : IO α := do
   IO.eprintln s!"Exception: {message}\n\nRun {hint} for additional help."
   IO.Process.exit 1
+
+/-- Exit with code 1 for user code errors (detected bugs in the Python source). -/
+def exitUserCodeError {α} (message : String) : IO α := do
+  IO.eprintln s!"❌ {message}"
+  IO.Process.exit 1
+
+/-- Exit with code 2 for internal errors (tool limitations or crashes). -/
+def exitInternalError {α} (message : String) : IO α := do
+  IO.eprintln s!"Exception: {message}"
+  IO.Process.exit 2
 
 def exitCmdFailure {α} (cmdName : String) (message : String) : IO α :=
   exitFailure message (hint := s!"strata {cmdName} --help")
@@ -90,6 +103,30 @@ def buildDialectFileMap (pflags : ParsedFlags) : IO Strata.DialectFileMap := do
   return sp
 
 end ParsedFlags
+
+def parseCheckMode (pflags : ParsedFlags) : IO VerificationMode :=
+  match pflags.getString "check-mode" with
+  | .none => pure .deductive
+  | .some s => match VerificationMode.ofString? s with
+    | .some m => pure m
+    | .none => exitFailure s!"Invalid check mode: '{s}'. Must be {VerificationMode.options}."
+
+def parseCheckLevel (pflags : ParsedFlags) : IO CheckLevel :=
+  match pflags.getString "check-level" with
+  | .none => pure .minimal
+  | .some s => match CheckLevel.ofString? s with
+    | .some l => pure l
+    | .none => exitFailure s!"Invalid check level: '{s}'. Must be {CheckLevel.options}."
+
+def checkModeFlag : Flag :=
+  { name := "check-mode",
+    help := s!"Check mode: {VerificationMode.options}. Default: 'deductive'.",
+    takesArg := .arg "mode" }
+
+def checkLevelFlag : Flag :=
+  { name := "check-level",
+    help := s!"Check level: {CheckLevel.options}. Default: 'minimal'.",
+    takesArg := .arg "level" }
 
 structure Command where
   name : String
@@ -290,6 +327,9 @@ def pyAnalyzeCommand : Command where
       let vcResults ← match options.vcDirectory with
                       | .none => IO.FS.withTempDir runVerification
                       | .some tempDir => runVerification tempDir
+      let mfm : Option (String × Lean.FileMap) := match pySourceOpt with
+        | some (pyPath, srcText) => some (pyPath, .ofString srcText)
+        | none => none
       let mut s := ""
       for vcResult in vcResults do
         -- Build location string based on available metadata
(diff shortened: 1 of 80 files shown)


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -170,9 +170,9 @@ def translateExpr (expr : StmtExprMd)
     match op with
     | .Eq => return .eq () re1 re2
     | .Neq => return .app () boolNotOp (.eq () re1 re2)
-    | .And => return binOp boolAndOp
-    | .Or => return binOp boolOrOp
-    | .Implies => return binOp boolImpliesOp
+    | .And => return .ite () re1 re2 (.boolConst () false)
```
keyboardDrummer: Should we rename these to `LazyAnd`, `LazyOr` and `LazyImplies` ?

StrataMain.lean
```diff
@@ -805,6 +812,7 @@ def laurelAnalyzeCommand : Command where
     match transResult with
     | .error transErrors => exitFailure s!"Translation errors: {transErrors}"
     | .ok laurelProgram =>
+      -- Laurel pipeline uses z3 (cvc5 is the default for Core-only verification)
```
keyboardDrummer: Would be nice to get rid of the duplication between commands.

keyboardDrummer (changes requested):
To prevent accidental unsoundness due to Laurel consumers misinterpreting the meaning of And/Or/Implies, please rename them to make the laziness clear, and please add tests to confirm the laziness. To do the equivalent of an `assert false` in the assert false branches, you can call a function with `requires false` as a precondition.

fabiomadge:
The preloading of Laurel in `buildDialectFileMap` broke `javaGen` when passed the `.st` file path — it threw 'Laurel already loaded'. This was a blocker for JVerify's `make generate-laurel-ast`. The fix follows the same pattern as the `print` command.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
fabiomadge: Went with `AndThen`/`OrElse` following Ada's convention. `And`/`Or` stay eager, exposed as `&`/`|` in the grammar. Added laziness tests using `requires false` and division by zero.

StrataMain.lean
fabiomadge: Done — extracted `laurelVerifyOptions`.

aqjune-aws (dismissed):

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -172,7 +172,9 @@ def translateExpr (expr : StmtExprMd)
     | .Neq => return .app () boolNotOp (.eq () re1 re2)
     | .And => return binOp boolAndOp
     | .Or => return binOp boolOrOp
-    | .Implies => return binOp boolImpliesOp
+    | .AndThen => return .ite () re1 re2 (.boolConst () false)
```
olivier-aws, line 178: This should be done in the `ListImperativeExpression` pass. Otherwise imperative expressions that should be short-circuited would be lifted before the `AndThen`.

fabiomadge: Good catch. Fixed by desugaring to `IfThenElse` before lifting when the second operand is imperative. Pure operands pass through unchanged. Added tests for all three operators.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -474,13 +474,66 @@ def transformProcedure (proc : Procedure) : LiftM Procedure := do
   | .External =>
       pure proc
 
+/--
+Desugar short-circuit operators (AndThen, OrElse, Implies) to IfThenElse
+when the second operand contains imperative calls. This must happen before
+LiftImperativeExpressions to prevent the lifter from hoisting imperative
+calls out of the short-circuited branch.
+-/
+private def desugarShortCircuitExpr (model : SemanticModel) (expr : StmtExprMd) : StmtExprMd :=
```
keyboardDrummer: HMM, sorry but I'd rather not complicate this pass. Why not always desugar the lazy operators to IfThenElse, before this pass? You can then drop support for translating them in LaurelToCoreTranslator.

StrataTest/Languages/Laurel/Examples/Fundamentals/T15_ShortCircuit.lean
```diff
@@ -0,0 +1,79 @@
+/-
+  Copyright Strata Contributors
```
keyboardDrummer, line 2: Nice tests!

Strata/Languages/Laurel/LiftImperativeExpressions.lean
fabiomadge: I tried a non-recursive top-level rewrite inside transformExpr, but Lean's termination checker can't connect subterms of the rewritten expression to the original. I could desugar in ConcreteToAbstractTreeTranslator instead (~60 fewer lines), but that loses the short-circuit operators early in the pipeline, making intermediate representations and laurelPrint output harder to read, and preventing future Laurel passes from using them. Which do you prefer?

olivier-aws: Can't we write a new Laurel -> Laurel pass that just does the desugaring of AndThen/OrElse to if-then-else?

fabiomadge: That's pretty much what this already is, but I can put it into a separate fine, if that's the concern.

olivier-aws: Yes it may be cleaner for this to be its own pass, in a separate file, and invoke it in the pipeline from Laurel to Core (so in the t`ranslate` method of `LaurelToCoreTranslator.lean`) rather than for it to be linked to the liftImperativeStatement pass.

fabiomadge: Done. Moved the desugaring to its own file (`DesugarShortCircuit.lean`) and invoke it from `translate` in `LaurelToCoreTranslator.lean`. `LiftImperativeExpressions` is unchanged from main (only made `containsAssignmentOrImperativeCall` public so the new pass can use it).

keyboardDrummer: Thanks!

keyboardDrummer: I'm hoping that in the future we can create some functions that'll allow us to write these passes without all the boilerplate.

keyboardDrummer (approved):
Looks great, thanks

Strata/Languages/Laurel/LiftImperativeExpressions.lean
fabiomadge: Agreed! I was thinking about that as well.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -666,6 +669,7 @@ def translate (options: LaurelTranslateOptions) (program : Program): Except (Arr
   let (program, model) := (result.program, result.model)
   let program := inferHoleTypes model program
   let program := eliminateHoles program
+  let program := desugarShortCircuit model program
```
olivier-aws, line 672: Nice, thanks.

PR #591, opened 2026-03-17
Remove panics in Laurel

keyboardDrummer:
### Changes
- Bring back the use of `PythonRuntimeLaurelPart`, to enable more resolution in Laurel. This was accidentally broken in https://github.com/strata-org/Strata/pull/515
- Remove some Laurel related dead code, like LaurelEval.lean and isPureExpr
- Remove panics in Laurel. In many cases the panic has been replaced by a default value, so the behavior is not different, but the default value indicates that the author thought this default does not cause unsound behavior.
- Rename `HighType.Top` to `HighType.Unknown` and add a comment

### Caveats
Also after this PR, there are many incompleteness bugs in the Laurel pipeline that it does not inform the user about. Cases like:
```
  | .InstanceCall _ _ _ =>
      dbg_trace "Not yet implemented: computeExprType InstanceCall"
      default
```

A better experience would be for emitted diagnostics to report that a completeness bug was encountered, but I'm leaving that out of scope for this PR.

### Testing
- Refactoring, no additional automated tests needed
- Internal testing scripts returns "Results: 28 passed, 22 failed"

### DIFF
--- a/Strata/Languages/Boole/Verify.lean
+++ b/Strata/Languages/Boole/Verify.lean
@@ -577,7 +577,7 @@ private def lowerPureFuncDef
     (ret : Boole.Type)
     (pres : Array (BooleDDM.SpecElt SourceRange))
     (body : Boole.Expr)
-    (inline : Bool) : TranslateM Core.Decl := do
+    (inline : Bool) : TranslateM Core.Function := do
   withTypeBVars tys do
     let bsList := bindingsToList bs
     let inputs ← bsList.mapM toCoreBinding
@@ -585,7 +585,7 @@ private def lowerPureFuncDef
     let pres ← withBVars inputNames (toCoreSpecElts m n pres)
     let pres := pres.preconditions.map (fun (_, c) => ⟨c.expr, ()⟩)
     let body ← withBVars inputNames (toCoreExpr body)
-    return .func {
+    return {
       name := mkIdent n
       typeArgs := tys
       inputs := inputs
@@ -667,10 +667,12 @@ def toCoreDecls (cmd : BooleDDM.Command SourceRange) : TranslateM (List Core.Dec
       return [.func { name := mkIdent n, typeArgs := tys, inputs := ← (bindingsToList bs).mapM toCoreBinding, output := ← toCoreMonoType ret, body := none, concreteEval := none, attr := #[], axioms := [] }]
   | .command_fndef m ⟨_, n⟩ ⟨_, targs?⟩ bs ret ⟨_, pres⟩ body ⟨_, inline?⟩ =>
     let tys := match targs? with | none => [] | some ts => typeArgsToList ts
-    return [← lowerPureFuncDef m n tys bs ret pres body inline?.isSome]
+    let f ← lowerPureFuncDef m n tys bs ret pres body inline?.isSome
+    return [.func f]
   | .command_recfndef m ⟨_, n⟩ ⟨_, targs?⟩ bs ret ⟨_, pres⟩ body =>
     let tys := match targs? with | none => [] | some ts => typeArgsToList ts
-    return [← lowerPureFuncDef m n tys bs ret pres body false]
+    let f ← lowerPureFuncDef m n tys bs ret pres body false
+    return [.func {f with isRecursive := true}]
   | .command_var _ b =>
     let (id, ty) ← toCoreBind b
     let i := (← get).globalVarCounter
--- a/StrataTest/Languages/Boole/FeatureRequests/abstract_types_and_stubs.lean
+++ b/StrataTest/Languages/Boole/FeatureRequests/abstract_types_and_stubs.lean
@@ -41,6 +41,7 @@ spec {
 };
 #end
 
+#guard_msgs (drop info) in
 #eval Strata.Boole.verify "cvc5" abstractTypesAndStubsSeed
 
 example : Strata.smtVCsCorrect abstractTypesAndStubsSeed := by
--- a/StrataTest/Languages/Boole/FeatureRequests/choose_operator.lean
+++ b/StrataTest/Languages/Boole/FeatureRequests/choose_operator.lean
@@ -44,7 +44,11 @@ spec {
 };
 #end
 
-#eval Strata.Boole.verify "cvc5" chooseOperatorSeed
+/-- info: Obligation: choose_seed_ensures_1_833
+Property: assert
+Result: ✅ pass-/
+#guard_msgs in
+#eval Strata.Boole.verify "cvc5" chooseOperatorSeed (options :=.quiet)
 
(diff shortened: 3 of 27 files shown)


Strata/Languages/Laurel/LaurelEval.lean
```diff
@@ -1,354 +0,0 @@
-/-
```
aqjune-aws, line 1: Why is this file removed?

Strata/Languages/Laurel/HeapParameterizationConstants.lean
```diff
@@ -37,6 +37,7 @@ program Laurel;
 // Composite: datatype with a reference (int)
 datatype Composite { MkComposite(ref: int) }
 
+datatype Top {}
```
aqjune-aws: What is the `Top` type?

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -320,7 +320,7 @@ instance : Inhabited StmtExprMd where
   default := ⟨ .Hole, .empty ⟩
 
 instance : Inhabited HighTypeMd where
-  default := { val := HighType.TVoid, md := default }
+  default := { val := HighType.Top, md := default }
```
aqjune-aws: Why did this change in this pull request (does it help reducing panic)?

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -115,8 +115,8 @@ structure SemanticModel where
 
 def SemanticModel.get (model: SemanticModel) (iden: Identifier): AstNode :=
   match iden.uniqueId with
-  | some key => (model.refToDef.get? key).getD (panic! s!"could not find key {key}")
-  | none => default -- panic! s!"model.get called on identifier {iden.text} without number"
+  | some key => (model.refToDef.get? key).get!
```
shigoel: `get!` will still panic if the lookup returns `none` here.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -61,15 +61,13 @@ def translateType (model : SemanticModel) (ty : HighTypeMd) : LMonoTy :=
   | .TCore s => .tcons s []
   | .TFloat64 => dbg_trace "NOT SUPPORTED YET: Float64"; .tcons "Float64IsNotSupportedYet" []
   | .TReal => LMonoTy.real
-  | .Top => LMonoTy.bool
-  | _ => panic s!"translateType: unsupported type {ToFormat.format ty}"
+  | .Top => .tcons "Top" []
+  | _ => dbg_trace s!"NOT SUPPORTED YET: {repr ty.val}"; .tcons "Float64IsNotSupportedYet" []
```
shigoel: Hadn't noticed this before, but the catch-all can be misleading here. If we want exhaustiveness checks during matching, let's not do this.

Strata/DL/Imperative/MetaData.lean
```diff
@@ -230,14 +230,14 @@ def getFileRange {P : PureExpr} [BEq P.Ident] (md: MetaData P) : Option FileRang
 
 /-- Create a DiagnosticModel from metadata and a message.
     Uses the file range from metadata if available, otherwise uses a default location. -/
-def MetaData.toDiagnostic {P : PureExpr} [BEq P.Ident] (md : MetaData P) (msg : String) : DiagnosticModel :=
+def MetaData.toDiagnostic {P : PureExpr} [BEq P.Ident] (md : MetaData P) (msg : String) (type : DiagnosticType := DiagnosticType.UserError): DiagnosticModel :=
   match getFileRange md with
-  | some fr => DiagnosticModel.withRange fr msg
+  | some fr => DiagnosticModel.withRange fr msg type
   | none => DiagnosticModel.fromMessage msg
```
shigoel: The `type` parameter is dropped on this branch — `fromMessage` always uses `UserError`, so a `StrataBug`-typed diagnostic will silently be reclassified. Fix:
```lean
| none => { DiagnosticModel.fromMessage msg with type := type }
```

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -256,35 +268,37 @@ def translateExpr (expr : StmtExprMd)
...
-  | .Assert _ => panic "assert expression not implemented"
-  | .Assume _ => panic "assume expression not implemented"
-  | .ProveBy value _ => panic "proveBy expression not implemented"
-  | .ContractOf _ _ => panic "contractOf expression not implemented"
-  | .Abstract => panic "abstract expression not implemented"
-  | .All => panic "all expression not implemented"
-  | .InstanceCall target callee args => panic "This expression not implemented"
-  | .PureFieldUpdate _ _ _ => panic "This expression not implemented"
-  | .This => panic "This expression not implemented"
+      throwDiagnostic $ md.toDiagnostic "FieldSelect should have been eliminated by heap parameterization: {Std.ToFormat.format target}#{fieldId.text}" DiagnosticType.StrataBug
```
shigoel:
```suggestion
      throwDiagnostic $ md.toDiagnostic s!"FieldSelect should have been eliminated by heap parameterization: {Std.ToFormat.format target}#{fieldId.text}" DiagnosticType.StrataBug
```

Strata/Languages/Laurel/LaurelEval.lean
keyboardDrummer: It was not used

Strata/Languages/Laurel/HeapParameterizationConstants.lean
keyboardDrummer: I will rename it and add a comment in another PR, but it's only used internally to help avoid incorrect errors when continuing compilation after an initial error occurs.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
joehendrix: I'd like to remove dbg_trace from cluttering stderr when debugging completely unrelated code.

Can you replace with proper logging (e.g., a state monad that contains a list of warnings or a post-processing state that collects all the not-supported-yet constructs and reports).

Strata/Languages/Laurel/Laurel.lean
keyboardDrummer: It's unrelated to panic, but .Top is more correct here.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: > I'd like to remove dbg_trace from cluttering stderr when debugging completely unrelated code.

Makes sense.

PR #592, opened 2026-03-17
Add automated overflow/underflow checks for bitvector arithmetic

tautschnig:
### Motivation

Arithmetic overflow is a major source of bugs in systems code. Signed integer overflow is undefined behavior in C/C++, unsigned overflow silently wraps, and floating-point overflow silently produces infinity. Unlike division-by-zero (added in #510), these errors are easy to miss in code review and hard to trigger in testing.

This PR extends Strata's automated verification to catch arithmetic overflow at every use site — the same way #510 catches every division by zero — so that front-end languages like Laurel (and future C/Java front-ends) get overflow checking for free, without any source-level instrumentation.

### Approach

Following the pattern established by #510 for division-by-zero:

1. **Safe operator variants** (e.g., `Bv32.SafeAdd`) carry built-in preconditions asserting that overflow does not occur.
2. **The existing `PrecondElim` transform** automatically generates verification conditions at each use site.
3. **Front-end translators** emit safe variants instead of unchecked ones — no language-level instrumentation required.

This keeps the architecture uniform: the same mechanism handles division-by-zero, signed overflow, unsigned overflow, and signed division overflow (`INT_MIN / -1`).

### What's covered

**Signed bitvector overflow** (all sizes: bv1, bv2, bv8, bv16, bv32, bv64):
- `SafeAdd/SafeSub/SafeMul/SafeNeg` with preconditions using SMT-LIB's `bvsaddo`/`bvssubo`/`bvsmulo`/`bvnego`

**Signed division overflow** (`INT_MIN / -1`):
- `SafeSDiv/SafeSMod` with dual preconditions: divisor ≠ 0 AND ¬(dividend = INT_MIN ∧ divisor = −1)

**Unsigned bitvector overflow** (configurable via `unsignedOverflowCheck` option):
- `SafeUAdd/SafeUSub/SafeUMul/SafeUNeg` with unsigned overflow predicates

**GOTO backend**:
- Overflow predicates translate to CBMC's native `overflow-+`, `overflow--`, `overflow-*`, `overflow-unary-` expressions
- Safe operators translate to the same GOTO operations as their unchecked counterparts

### Key design decisions

- **Safe variants are opt-in at the Core level** — `Bv32.Add` (wrapping) and `Bv32.SafeAdd` (checked) coexist. Front-end translators choose which to emit.
- **Unsigned overflow is off by default** — wrapping is the intended semantics in most languages. Enable with `unsignedOverflowCheck := true`.
- **Signed overflow is always checked** when using safe variants — this matches C/C++ undefined behavior semantics.
- **The GOTO backend uses CBMC's native overflow expressions** rather than the SMT encoding, so CBMC's solver infrastructure handles them natively.

*Description of changes:*

New `arithmeticOverflow` property type alongside the existing `divisionByZero`, with SARIF classification, display labels, and check mode handling. Overflow predicates and safe operators for all BV sizes generated via Lean4 macros. SMT encoding for signed overflow uses SMT-LIB builtins; unsigned uses custom encodings. Comprehensive unit tests verify factory presence, WF obligation generation, and GOTO translation.


Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -402,6 +402,24 @@ def handleUnaryOps {M} [Inhabited M] (name : String) (arg : CoreDDM.Expr M)
...
+  | "Bv8.SafeUNeg" => pure (.neg_expr default (.bv8 default) arg)
+  | "Bv16.SafeUNeg" => pure (.neg_expr default (.bv16 default) arg)
+  | "Bv32.SafeUNeg" => pure (.neg_expr default (.bv32 default) arg)
+  | "Bv64.SafeUNeg" => pure (.neg_expr default (.bv64 default) arg)
+  -- Signed negation overflow predicates
+  | "Bv1.SNegOverflow" => pure (.neg_expr default (.bv1 default) arg)
+  | "Bv8.SNegOverflow" => pure (.neg_expr default (.bv8 default) arg)
+  | "Bv16.SNegOverflow" => pure (.neg_expr default (.bv16 default) arg)
+  | "Bv32.SNegOverflow" => pure (.neg_expr default (.bv32 default) arg)
+  | "Bv64.SNegOverflow" => pure (.neg_expr default (.bv64 default) arg)
```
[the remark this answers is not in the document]
tautschnig: Fixed in 0f524cc. `SNegOverflow` and `UNegOverflow` are now mapped to `Bool.Not` instead of `neg_expr`, which is type-correct (boolean predicate, not arithmetic negation). These overflow predicates only appear in PrecondElim-generated assertions, not in user-written code, so the CST representation is only used for debug printing.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -439,7 +457,16 @@ def bvBinaryOpMap {M} [Inhabited M] :
   ("SLe", fun ty arg1 arg2 => .bvsle default ty arg1 arg2),
   ("SLt", fun ty arg1 arg2 => .bvslt default ty arg1 arg2),
   ("SGe", fun ty arg1 arg2 => .bvsge default ty arg1 arg2),
-  ("SGt", fun ty arg1 arg2 => .bvsgt default ty arg1 arg2)
+  ("SGt", fun ty arg1 arg2 => .bvsgt default ty arg1 arg2),
+  -- Safe variants map to the same CST constructors
+  ("SafeAdd", fun ty arg1 arg2 => .add_expr default ty arg1 arg2),
+  ("SafeSub", fun ty arg1 arg2 => .sub_expr default ty arg1 arg2),
+  ("SafeMul", fun ty arg1 arg2 => .mul_expr default ty arg1 arg2),
+  ("SafeSDiv", fun ty arg1 arg2 => .bvsdiv default ty arg1 arg2),
```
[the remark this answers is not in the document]
tautschnig: Fixed in 0f524cc. All binary overflow predicates (`SAddOverflow`, `SSubOverflow`, `SMulOverflow`, `SDivOverflow`, `UAddOverflow`, `USubOverflow`, `UMulOverflow`) are now in `bvBinaryOpMap` with a boolean CST approximation. They no longer fall through to the error path.

Strata/Transform/PrecondElim.lean
```diff
@@ -53,6 +53,11 @@ def wfProcName (name : String) : String := s!"{name}{wfSuffix}"
 private def classifyPrecondition (funcName : String) : Option String :=
   if funcName.startsWith "Int.SafeDiv" || funcName.startsWith "Int.SafeMod" then
     some Imperative.MetaData.divisionByZero
+  else if funcName.startsWith "Bv1.Safe" || funcName.startsWith "Bv2.Safe" ||
+          funcName.startsWith "Bv8.Safe" || funcName.startsWith "Bv16.Safe" ||
+          funcName.startsWith "Bv32.Safe" || funcName.startsWith "Bv64.Safe" ||
+          funcName.startsWith "Float64.Safe" then
+    some Imperative.MetaData.arithmeticOverflow
```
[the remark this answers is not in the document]
tautschnig: Fixed in 0f524cc. `classifyPrecondition` now takes a `precondIdx` parameter, and `collectPrecondAsserts` tracks the per-function precondition index. For `SafeSDiv`/`SafeSMod`, precondition 0 (y ≠ 0) is classified as `divisionByZero`, and precondition 1 (¬SDivOverflow) is classified as `arithmeticOverflow`.

Strata/Languages/Core/Options.lean
```diff
@@ -86,6 +86,8 @@ structure VerifyOptions where
   checkMode : VerificationMode
   /-- Check amount: minimal (only necessary checks) or full (both checks for better messages) -/
   checkLevel : CheckLevel
+  /-- Enable unsigned bitvector overflow checks -/
+  unsignedOverflowCheck : Bool
```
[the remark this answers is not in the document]
tautschnig: Addressed in 0f524cc. Added documentation explaining that this is forward-looking infrastructure: the option will be wired when a front-end language (Laurel or a future C front-end) gains bitvector types. Currently no front-end emits BV operations, so there's nothing to wire it to yet. The option documents the design intent that unsigned overflow checking should be opt-in.

Strata/Languages/Core/Factory.lean
```diff
@@ -166,6 +300,48 @@ def realGtFunc : WFLFunc CoreLParams :=
...
+/- Float64 Safe Arithmetic (overflow = result becomes ±infinity from finite inputs) -/
+
+def float64SafeAddFunc : WFLFunc CoreLParams :=
+  binaryFuncUneval "Float64.SafeAdd" mty[float64] mty[float64] mty[float64]
+def float64SafeSubFunc : WFLFunc CoreLParams :=
+  binaryFuncUneval "Float64.SafeSub" mty[float64] mty[float64] mty[float64]
+def float64SafeMulFunc : WFLFunc CoreLParams :=
+  binaryFuncUneval "Float64.SafeMul" mty[float64] mty[float64] mty[float64]
+def float64SafeDivFunc : WFLFunc CoreLParams :=
+  binaryFuncUneval "Float64.SafeDiv" mty[float64] mty[float64] mty[float64]
```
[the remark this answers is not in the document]
tautschnig: Fixed in 0f524cc. `Float64.SafeAdd/Sub/Mul/Div` now carry preconditions requiring `¬Float64.*Overflow(x, y)`. Each is constructed as a `WFLFunc` with an explicit precondition expression referencing the corresponding overflow predicate function. `PrecondElim` now generates overflow VCs at every Float64 safe arithmetic use site.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -167,16 +170,19 @@ def translateExpr (expr : StmtExprMd)
...
     | .Eq => return .eq () re1 re2
     | .Neq => return .app () boolNotOp (.eq () re1 re2)
     | .And => return binOp boolAndOp
     | .Or => return binOp boolOrOp
     | .Implies => return binOp boolImpliesOp
-    | .Add => return binOp (if isReal then realAddOp else intAddOp)
-    | .Sub => return binOp (if isReal then realSubOp else intSubOp)
-    | .Mul => return binOp (if isReal then realMulOp else intMulOp)
-    | .Div => return binOp (if isReal then realDivOp else intSafeDivOp)
+    | .Add => return binOp (if isFloat64 then float64SafeAddOp else if isReal then realAddOp else intAddOp)
```
keyboardDrummer: I think floats in the Laurel source languages, Java and Python, implement IEEE 754 so they do allow overflows for floating point operations, so I think Laurel's translation of floating ops should not use the 'safe' floating point operators in Core.

I think we could define a subset type for `float64` that does not have positive or negative infinity as a value, and then something like `var x: finiteFloat64 := a * b` would   simplify to 
```
var x: float64 := 
  var t := a * b; 
  assert t != Infinity && t != -Infinity; 
  t
```

But it would be up to the user to indicate when they want to use a `finiteFloat64` instead of a `float64`.

tautschnig: Agreed — IEEE 754 defines overflow to ±∞ as valid behavior, so unconditionally inserting overflow assertions for Laurel float ops would produce false positives for standard Java/Python programs.

Fixed in 5cae06b2: the Laurel translator now uses the non-safe float operators by default. I also took the opportunity to unify the approach across all overflow check categories. There's now an `OverflowChecks` structure in `Core.Options`:

```lean
structure OverflowChecks where
  signedBV   : Bool := true   -- C: signed overflow is UB
  unsignedBV : Bool := false  -- C: unsigned wraps (defined)
  float64    : Bool := false  -- IEEE 754: overflow to ±∞ (defined)
```

This replaces the previous (unused) `unsignedOverflowCheck` flag. The config is threaded through `LaurelTranslateOptions` → `TranslateState` and exposed via `--overflow-checks signed,unsigned,float64` on the CLI. Users who want finite-float checking can opt in with `--overflow-checks float64`.

The `finiteFloat64` subset type idea is interesting for a future iteration — it would let users express the constraint at the type level rather than as a global flag.

pr592-reply.md
shigoel, line 1: I think this file accidentally made it into this PR. Let's delete.

Strata/Transform/PrecondElim.lean
```diff
@@ -68,12 +83,19 @@ def collectPrecondAsserts (F : @Lambda.Factory CoreLParams) (e : Expression.Expr
...
-      | none => md
-    Statement.assert
-    s!"{labelPrefix}_calls_{ob.funcName}_{idx}" ob.obligation md'
+  -- Track per-function precondition index for correct classification
+  -- (e.g., SafeSDiv precondition 0 = div-by-zero, precondition 1 = overflow)
+  let (_, _, result) := wfObs.foldl (init := ("", 0, ([] : List Statement)))
+    fun (prevFunc, prevIdx, acc) ob =>
+      let precondIdx := if ob.funcName == prevFunc then prevIdx + 1 else 0
+      let globalIdx := acc.length
+      let md' := match classifyPrecondition ob.funcName precondIdx with
```
shigoel, line 100: The precondition index tracking here assumes obligations for the same function appear consecutively, but `collectWFObligations` returns obligations in depth-first traversal order. For nested calls like `SafeSDiv(SafeSDiv(x, y), z)`, the four obligations arrive as [inner-precond0, inner-precond1, outer-precond0, outer-precond1] with the same funcName throughout, so `precondIdx` increments to 2 and 3 instead of resetting to 0 and 1 for the outer call. This causes the outer call's div-by-zero precondition to be misclassified as `arithmeticOverflow`. Consider tracking the index modulo the function's precondition count (available from `F[ob.funcName]?.map (·.preconditions.length)`) so the index cycles correctly across call sites.

StrataMain.lean
```diff
@@ -1227,11 +1228,21 @@ def verifyCommand : Command where
...
+      | .none => Core.OverflowChecks.mk true false false
+      | .some s => s.splitOn "," |>.foldl (fun acc c =>
+          match c.trimAscii.toString with
+          | "signed"   => { acc with signedBV := true }
+          | "unsigned" => { acc with unsignedBV := true }
+          | "float64"  => { acc with float64 := true }
+          | "none"     => { signedBV := false, unsignedBV := false, float64 := false }
+          | "all"      => { signedBV := true, unsignedBV := true, float64 := true }
+          | _          => acc) {}
     let opts : VerifyOptions :=
```
shigoel: The fold initializer `{}` inherits `OverflowChecks` struct defaults (`signedBV := true`), so `--overflow-checks unsigned` silently enables signed checks too. When the flag is explicitly provided, the seed should be all-false (`{ signedBV := false, unsignedBV := false, float64 := false }`) so that only the listed checks are enabled. Without this, there is no way to request only unsigned checks without the `none,unsigned` workaround.

Strata/Languages/Core/Options.lean
```diff
@@ -98,6 +98,18 @@ inductive CheckLevel where
 instance : Inhabited CheckLevel where
   default := .minimal
 
+/-- Configuration for which arithmetic overflow checks to enable.
+    Defaults match C/IEEE 754 semantics: signed BV overflow is UB (checked),
```
shigoel, line 109: Is IEEE754 relevant anymore here? IIUC, no FP in this PR?

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -408,6 +408,21 @@ def handleUnaryOps {M} [Inhabited M] (name : String) (arg : CoreDDM.Expr M)
   | "Bv32.Neg" => pure (.neg_expr default (.bv32 default) arg)
   | "Bv64.Not" => pure (.bvnot default (.bv64 default) arg)
   | "Bv64.Neg" => pure (.neg_expr default (.bv64 default) arg)
+  -- Safe negation variants (same CST as regular negation)
```
shigoel, line 426: Just occurs to me: this design decision means that any Core program that goes through an AST/CST roundtrip (e.g., via the new `strata transform` command) will lose all safe operator annotations and we won't benefit from property instrumentation then.

shigoel: Maybe it's worth adding dedicated syntax? Or a `safe` modifier to existing operations (though that sounds very noisy, even as I type this out)?

Strata/Languages/Core/Options.lean
```diff
@@ -152,6 +164,8 @@ structure VerifyOptions where
   /-- How many checks to run per VC and how detailed the
       messages should be. -/
   checkLevel : CheckLevel
+  /-- Overflow check configuration: which arithmetic overflow checks to enable. -/
+  overflowChecks : OverflowChecks := {}
```
shigoel, line 178: If PR https://github.com/strata-org/Strata/pull/764 happens to get its comments addressed and merged before this one: adding a note to sync. this new option in StrataMain.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -344,7 +344,9 @@ private def createUnreachableAssertObligations
   asserts.toArray.map
     (fun (label, md) =>
       let propType := match md.getPropertyType with
-        | some s => if s == Imperative.MetaData.divisionByZero then .divisionByZero else .assert
+        | some s => if s == Imperative.MetaData.divisionByZero then .divisionByZero
```
shigoel, line 347: Are we missing a similar classification in CmdEval.lean?
E.g., here: https://github.com/strata-org/Strata/blob/main/Strata/DL/Imperative/CmdEval.lean#L69

Strata/Languages/Core/Verifier.lean
shigoel, line 1: Perhaps out of scope of this PR, but this matching based on property type across functions like `emoji`, `label`, and `verifySingleEnv` is getting a little hard to manage (at least to me). I think the main distinction is between whether they're assert-like or cover-like properties, so I'd feel comfortable matching on `property.passWhenUnreachable` instead of `property`.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -445,7 +460,24 @@ def bvBinaryOpMap {M} [Inhabited M] :
   ("SLe", fun ty arg1 arg2 => .bvsle default ty arg1 arg2),
   ("SLt", fun ty arg1 arg2 => .bvslt default ty arg1 arg2),
   ("SGe", fun ty arg1 arg2 => .bvsge default ty arg1 arg2),
-  ("SGt", fun ty arg1 arg2 => .bvsgt default ty arg1 arg2)
+  ("SGt", fun ty arg1 arg2 => .bvsgt default ty arg1 arg2),
+  -- Safe variants map to the same CST constructors
```
joscoh, line 477: What is the purpose of having safe variants that behave exactly the same as the unsafe ones? Why do we need both?

Strata/Languages/Core/Factory.lean
```diff
@@ -141,11 +162,123 @@ elab "ExpandBVOpFuncDefs" "[" sizes:num,* "]" : command => do
...
+      let overflowFuncName := mkIdent
+        (.str .anonymous s!"bv{s}{spec.overflowPredSuffix}Func")
+      if spec.isUnary then
+        elabCommand (← `(
+          def $funcName : Lambda.WFLFunc CoreLParams :=
+            Lambda.unaryOp (InValTy := BitVec $sizeNum) $opName $(mkIdent spec.opFn)
+              (preconditions := [⟨.app default
+                (Lambda.boolNotFunc (T := CoreLParams)).func.opExpr
+                (.app default ($overflowFuncName).opExpr
+                  (.fvar default "x" (some (.bitvec $sizeNum)))),
```
joscoh: This is brittle: it relies on the name of the variable in `Lambda.unaryOp`. At least this should be marked with a comment.

Strata/Languages/Core/Factory.lean
```diff
@@ -141,11 +162,123 @@ elab "ExpandBVOpFuncDefs" "[" sizes:num,* "]" : command => do
...
+                (.eq default
+                  (.fvar default "y" (some (.bitvec $sizeNum)))
+                  (LExpr.bitvecConst default $sizeNum (0 : BitVec $sizeNum))),
+                default⟩,
+              -- Precondition 2: ¬ SDivOverflow(x, y)
+              ⟨.app default
+                (Lambda.boolNotFunc (T := CoreLParams)).func.opExpr
+                (.app default
+                  (.app default ($overflowFuncName).opExpr
+                    (.fvar default "x" (some (.bitvec $sizeNum))))
```
joscoh: Same here

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -388,10 +388,78 @@ partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Conte
...
+      let name := func.name.name
+      if name.startsWith "Bv" && (name.endsWith "UAddOverflow" || name.endsWith "USubOverflow" ||
+         name.endsWith "UMulOverflow" || name.endsWith "UNegOverflow" ||
+         name.endsWith "SafeUAdd" || name.endsWith "SafeUSub" ||
+         name.endsWith "SafeUMul" || name.endsWith "SafeUNeg") then
+        -- Parse size from "BvN.Op"
+        let sizeStr := (name.splitOn ".").head!.drop 2
+        let n := sizeStr.toNat!
+        let bvTy := TermType.prim (.bitvec n)
+        if name.endsWith "UAddOverflow" then uaddOverflowEnc n ctx
```
joscoh, line 453: Is there a check in the Core concrete syntax to disallow such names? All of these `startWith` and `endWith` checks make a bit nervous that they could be accidentally hit via user-defined functions.

Strata/Transform/PrecondElim.lean
```diff
@@ -68,12 +83,19 @@ def collectPrecondAsserts (F : @Lambda.Factory CoreLParams) (e : Expression.Expr
...
   let wfObs := Lambda.collectWFObligations F e
-  wfObs.mapIdx fun idx ob =>
-    let md' := match classifyPrecondition ob.funcName with
-      | some pt => md.pushElem Imperative.MetaData.propertyType (.msg pt)
-      | none => md
-    Statement.assert
-    s!"{labelPrefix}_calls_{ob.funcName}_{idx}" ob.obligation md'
+  -- Track per-function precondition index for correct classification
+  -- (e.g., SafeSDiv precondition 0 = div-by-zero, precondition 1 = overflow)
+  let (_, _, result) := wfObs.foldl (init := ("", 0, ([] : List Statement)))
```
joscoh, line 94: I am confused why this more complicated `foldl` mechanism is necessary.

Strata/Transform/PrecondElim.lean
```diff
@@ -68,12 +83,19 @@ def collectPrecondAsserts (F : @Lambda.Factory CoreLParams) (e : Expression.Expr
...
+    fun (prevFunc, prevIdx, acc) ob =>
+      let precondIdx := if ob.funcName == prevFunc then prevIdx + 1 else 0
+      let globalIdx := acc.length
+      let md' := match classifyPrecondition ob.funcName precondIdx with
+        | some pt => md.pushElem Imperative.MetaData.propertyType (.msg pt)
+        | none => md
+      let stmt := Statement.assert
+        s!"{labelPrefix}_calls_{ob.funcName}_{globalIdx}" ob.obligation md'
+      (ob.funcName, precondIdx, acc ++ [stmt])
```
joscoh: The original `map`-based traversal was linear, this is now quadratic

StrataTest/Languages/Core/Tests/OverflowCheckTest.lean
```diff
@@ -0,0 +1,74 @@
...
+example := Core.bv32SafeNegOp
+example := Core.bv64SafeAddOp
+
+-- Verify overflow predicate expressions exist
+example := Core.bv32SAddOverflowOp
+example := Core.bv32SSubOverflowOp
+example := Core.bv32SMulOverflowOp
+example := Core.bv32SNegOverflowOp
+
+-- Verify WF obligations are generated for safe add (1 precondition)
```
joscoh, line 33: At least one of these tests should test that the correct obligation gets created IMO.

Strata/Languages/Core/SMTEncoder.lean
shigoel: Yes, I agree. Also, I'm worried about our reliance on string dispatch in general. I've filed an issue to track this (https://github.com/strata-org/Strata/issues/785).

pr592-reply.md
tautschnig: Deleted in f3c6653f.

Strata/Transform/PrecondElim.lean
tautschnig: Fixed in f3c6653f. `collectPrecondAsserts` now computes `rawIdx % precondCount` where `precondCount` is looked up from `F[ob.funcName]?.map (·.preconditions.length)`. For nested `SafeSDiv(SafeSDiv(x,y),z)`, the four obligations now correctly classify as [divByZero, overflow, divByZero, overflow] instead of [divByZero, overflow, overflow, overflow]. Added a test that exercises this exact scenario.

StrataMain.lean
tautschnig: Fixed in f3c6653f. The fold seed is now `{ signedBV := false, unsignedBV := false, float64 := false }` when the flag is explicitly provided, so `--overflow-checks unsigned` only enables unsigned checks.

Strata/Languages/Core/Options.lean
tautschnig: The `float64` option is forward-looking infrastructure: it will be wired when a front-end language gains bitvector/float types that map to BV operations. Currently no front-end emits Float64 safe operations, so there's nothing to wire it to yet. The option documents the design intent that float overflow checking should be opt-in (IEEE 754 defines overflow to ±∞ as valid behavior).

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
tautschnig: Yes - let me know if you want me to do this as part of this PR or in a separate one.

Strata/Languages/Core/StatementEval.lean
tautschnig: Yes, good catch. Fixed in f3c6653f. `CmdEval.lean` now handles `arithmeticOverflow` in addition to `divisionByZero`, matching `StatementEval.lean`.

Strata/Languages/Core/Verifier.lean
tautschnig: Agreed this would simplify the code. Let me know if you want me to do this as part of this PR or in a separate one.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
tautschnig: The safe variants carry preconditions (WF obligations) that generate verification conditions at call sites via `PrecondElim`. The CST representation is only used for debug printing — the actual semantics difference is in the `Lambda.Factory` where safe ops have `preconditions := [...]` while unsafe ops have `preconditions := []`. The CST mapping to the same operator is intentional: the safe op computes the same result, it just additionally asserts that the inputs don't cause overflow.

Strata/Languages/Core/Factory.lean
tautschnig: Fixed in f3c6653f. Extracted `Lambda.unaryParamName`, `Lambda.binaryParam1Name`, and `Lambda.binaryParam2Name` as constants in `IntBoolFactory.lean`. Both `unaryOp`/`binaryOp` definitions and `Factory.lean` precondition expressions now reference these constants instead of hardcoded string literals.

Strata/Languages/Core/Factory.lean
tautschnig: Done, as above.

Strata/Languages/Core/SMTEncoder.lean
tautschnig: Let me know if you want me to make (those much larger) changes in this PR or leave it to follow-up.

Strata/Transform/PrecondElim.lean
tautschnig: The `foldl` tracks the per-function precondition index across consecutive obligations so that multi-precondition functions (like `SafeSDiv` with div-by-zero + overflow) get each precondition classified correctly. A simple `mapIdx` can't do this because the index needs to reset (or cycle) per function. With the modulo fix in f3c6653f, the mechanism is now correct for nested calls too.

Strata/Transform/PrecondElim.lean
tautschnig: Fixed in f3c6653f. Changed from `acc ++ [stmt]` (O(n) per iteration) to `stmt :: acc` with a final `.reverse` (O(1) per iteration, O(n) total).

StrataTest/Languages/Core/Tests/OverflowCheckTest.lean
tautschnig: Fixed in f3c6653f. Added two `#eval` tests that verify the `getPropertyType` metadata on generated assertions: one for a single `SafeSDiv` call (checks divByZero + overflow), and one for nested `SafeSDiv(SafeSDiv(x,y),z)` (checks all four obligations are correctly classified).

Strata/Languages/Core/Options.lean
tautschnig: Fixed in 8299977e.

tautschnig:

tautschnig:

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
shigoel: A separate PR would work. Please feel free to open an issue to track this, unless you plan to work on this immediately.

Strata/Languages/Core/Verifier.lean
shigoel: A separate one, thanks.

Strata/Languages/Core/SMTEncoder.lean
shigoel: A follow-up PR works. Seems like we can bundle that with whatever resolves https://github.com/strata-org/Strata/issues/785.

tautschnig:

Strata/Languages/Core/SMTEncoder.lean
tautschnig: I'm working on a PR to address #785. Will address this issue as part of that PR (which might actually become multiple PRs).

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
tautschnig: Now done as part of this PR, in a1d7001ff.

Strata/Languages/Core/Verifier.lean
tautschnig: Now actually done as part of this PR, in 20dd15e06.

Strata/Backends/CBMC/GOTO/LambdaToCProverGOTO.lean
```diff
@@ -101,6 +101,21 @@ def fnToGotoID (fn : String) : Except Format CProverGOTO.Expr.Identifier :=
...
+    | "SafeSMod" => return .binary .Mod
+    -- Overflow predicates map to CBMC overflow expressions
+    | "SAddOverflow" | "UAddOverflow" => return .binary .PlusOverflow
+    | "SSubOverflow" | "USubOverflow" => return .binary .MinusOverflow
+    | "SMulOverflow" | "UMulOverflow" => return .binary .MultOverflow
+    | "SNegOverflow" | "UNegOverflow" => return .unary .UnaryMinusOverflow
+    -- SDivOverflow is special: (x == INT_MIN) && (y == -1)
+    -- Handled as a function application (expanded in toGotoExpr)
+    | "SDivOverflow" => return .functionApplication fn
     | _ =>
```
shigoel, line 119: `SDivOverflow` is mapped to `.functionApplication fn` with a comment saying it is "expanded in toGotoExpr", but neither `toGotoExpr` nor `toGotoExprCtx` contain such an expansion. The only function-application expansions in those functions are for `Int.EuclideanDiv` and `Int.EuclideanMod` (LambdaToCProverGOTO.lean:226-227 and 306-307). When a `SafeSDiv` precondition containing `SDivOverflow(x, y)` is translated to GOTO, it will be emitted as an uninterpreted function call that CBMC cannot solve, unlike the SMT backend which correctly expands it to `(x == INT_MIN) && (y == -1)` via `sdivOverflowEnc` in SMTEncoder.lean.

joscoh (dismissed):

Strata/Backends/CBMC/GOTO/LambdaToCProverGOTO.lean
tautschnig: Thanks, addressed in [b916831](https://github.com/strata-org/Strata/pull/592/commits/b916831937687cad9055d622b941329be8ba921f).

shigoel (approved):

joscoh (approved):

PR #593, opened 2026-03-17
Feat/class field extraction and resolve fix

oyendrila-dobe:
This PR adds fix for resolveQualifiedFieldName panicking on our benchmarks

*Description of changes:*

#### PythonToLaurel.lean

- **Class-level field extraction:** translateClass now calls extractClassFields on the
class body in addition to extractFieldsFromInit, so annotated fields declared at 
class level (e.g., name: str) are included in the CompositeType. Fields from __init__ 
that duplicate class-level declarations are deduplicated — class-level annotations 
take priority.

#### HeapParameterization.lean

- **Graceful handling of unresolved fields:** resolveQualifiedFieldName no longer 
panics when a field isn't in the semantic model. Falls back to the unqualified field 
name. This handles attribute access on external/dynamic types (e.g., e.response on a 
caught exception) that the resolver can't trace to a known composite type. Sound but 
imprecise — won't disambiguate same-named fields across different classes.

#### Tests

- test_class_field_init.py — new test with annotated class-level field (name: str) and
annotated init-only field (self.size: int = size). Skipped for the non-laurel path.

### Benchmark results
No regression - (18/32) same as compared to mainline it is rebased on 

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -72,25 +72,25 @@ structure TranslationContext where
   /-- List of function signatures -/
   functionSignatures : List PythonFunctionDecl := []
   /-- Map from prelude procedure names to their full signatures -/
-  preludeProcedures : List (String × CoreProcedureSignature) := []
+  preludeProcedures : Std.HashMap String CoreProcedureSignature := {}
   /-- Names of prelude functions (non-procedure callables) -/
   preludeFunctions : List String := []
   /-- Names of user-defined functions -/
   userFunctions : List String := []
   /-- Names of user-defined classes -/
   userClasses : List String := []
-  /-- Map (Classname, Attribute) to its type -/
-  classAttributeType: Std.HashMap (String × String) String := {}
+  /-- Map ClassName -> (FieldName -> HighType) for field access coercions and type inference -/
+  classFieldHighType: Std.HashMap String (Std.HashMap String HighType) := {}
   /-- Names of prelude types -/
-  preludeTypes : List String := []
+  preludeTypes : Std.HashSet String := {}
   /-- Overload dispatch table from PySpec: function name → overloads -/
   overloadTable : Specs.ToLaurel.OverloadTable := {}
   /-- Behavior for unmodeled functions -/
   unmodeledBehavior : UnmodeledFunctionBehavior := .havocOutputs
   /-- File path for source location metadata -/
   filePath : String := ""
-  /-- List of defined composite types (classes) -/
-  compositeTypes : List CompositeType := []
+  /-- Known composite type names (user-defined classes + PySpec types) -/
+  compositeTypeNames : Std.HashSet String := {}
   /-- Track current class during method translation -/
   currentClassName : Option String := none
 deriving Inhabited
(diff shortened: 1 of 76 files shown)


Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -239,7 +239,9 @@ Returns the qualified field name "DeclaringType.fieldName".
 def resolveQualifiedFieldName (model: SemanticModel) (fieldName : Identifier) : String :=
   match model.get fieldName with
     | .field owner _ => owner.text ++ "." ++ fieldName.text
-    | _ => panic! s!"resolveQualifiedFieldName {fieldName} did not resolve to a field"
+    -- Unresolved fields (e.g., attribute access on external/dynamic types):
+    -- fall back to unqualified name. Sound but imprecise.
+    | _ => fieldName.text
```
shigoel: I don't often advocate for this, but perhaps it's worth putting a `dbg_trace` here to know how often we're in the fallback case.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1275,14 +1275,18 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
           .ok (some (funcDecl))
       | _ => .ok none)
     let ctx := {ctx with functionSignatures:= ctx.functionSignatures ++ classFunDecls}
-    -- Extract fields from __init__ method body
-    let mut fields : List Field := []
+    -- Extract fields from class-level annotations and __init__ body, with dedup
+    let classLevelFields ← extractClassFields ctx body.val
```
shigoel, line 1383: `extractClassFields` throws on non-`Name` annotated assignment targets:

```lean
| _ => throw (.unsupportedConstruct "Only simple field names supported" ...)
```

This is inconsistent with `extractFieldsFromInit`, which silently skips non-matching patterns via `| _ => pure ()`. If a class body contains an `AnnAssign` with a non-simple target (e.g. `self.x: int` written at class level), `translateClass` will now fail entirely — a regression from the old behavior where only `__init__`-level fields were extracted and unrecognized patterns were ignored.

Suggest changing to `| _ => pure ()` to match the skip-unknown pattern used elsewhere.

StrataTest/Languages/Python/tests/test_class_field_init.py
```diff
@@ -0,0 +1,14 @@
+import test_helper
+
+class CircularBuffer:
+    name: str
+    def __init__(self, size: int, name: str):
+        self.size: int = size
+        self.name = name
+
+def main():
+    my_buf: CircularBuffer = CircularBuffer(5)
```
shigoel: Shouldn't `CircularBuffer` be called with two arguments?

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -239,7 +239,11 @@ Returns the qualified field name "DeclaringType.fieldName".
 def resolveQualifiedFieldName (model: SemanticModel) (fieldName : Identifier) : String :=
   match model.get fieldName with
     | .field owner _ => owner.text ++ "." ++ fieldName.text
-    | _ => panic! s!"resolveQualifiedFieldName {fieldName} did not resolve to a field"
+    -- Unresolved fields (e.g., attribute access on external/dynamic types):
+    -- fall back to unqualified name. Sound but imprecise.
+    | _ =>
+      dbg_trace s!"resolveQualifiedFieldName fallback: {fieldName} not resolved to a field"
+      fieldName.text
```
keyboardDrummer, line 246: I don't think there should be a fallback here. I've changed this to return an `Option String` in https://github.com/strata-org/Strata/pull/591

andrewmwells-amazon: I'm fine merging this PR and leaving that change for 591

PR #594, opened 2026-03-17
Fix 419: eliminate Core VCG path multiplication across procedure boundaries

shigoel:
Fixes [#419](https://github.com/strata-org/Strata/issues/419)

### Summary

- `Procedure.eval` now always returns a single `(Procedure × Env)` instead of
  one per symbolic path through the procedure body. Multiple results are merged
  via `mergeResults`: deferred proof obligations are unioned.

- Fix obligation duplication: dead-branch obligations are now prepended to only
  the first live-branch result, preventing duplicates when `mergeResults` unions
  multiple paths from nested symbolic ITEs with exits.

- Drop source-order preservation for dead-branch obligations; these always
  appear before live-branch obligations.

- Add tests: dead-branch obligation generation, unreachable annotations
  (`checkLevel := .full`), and no-duplication with nested symbolic ITE + exit.

### DIFF
--- a/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
+++ b/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
@@ -61,7 +61,7 @@ def translateType (model : SemanticModel) (ty : HighTypeMd) : LMonoTy :=
   | .TCore s => .tcons s []
   | .TFloat64 => dbg_trace "NOT SUPPORTED YET: Float64"; .tcons "Float64IsNotSupportedYet" []
   | .TReal => LMonoTy.real
-  | .Top => LMonoTy.bool
+  | .Top => .tcons "Any" []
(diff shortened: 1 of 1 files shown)


StrataTest/Languages/Core/Tests/VCGPathTests.lean
```diff
@@ -0,0 +1,338 @@
...
+---------------------------------------------------------------------
+
+def concreteTrueDeadElse :=
+#strata
+program Core;
+procedure p() returns () {
+  if (true) {
+    assert [live_then]: true;
+  } else {
+    assert [dead_else]: true;
```
joscoh, line 98: Possibly silly question: if this assertion were `false`, then this should still pass, right?

shigoel: Not a silly question; and yes, it should still pass (and does; I checked)

PR #595, opened 2026-03-17
Support for f-strings in Python->Laurel translation

ssomayyajula:
*Description of changes:*

Handle f-strings in Python->Laurel translation
- FormattedValue: convert inner expr to string via to_string_any
- JoinedStr: concatenate all parts using PAdd (same as string +)
- Interpolation/TemplateStr (Python 3.14+ t-strings): fall through to Hole
- inferExprType: return Str for JoinedStr and FormattedValue
- Add f-string test cases to test_strings.py
- ~Bind exception variables from `except ... as e` handlers as LocalVariable declarations in Laurel, so they are in 
scope when referenced in f-strings or other expressions within the handler body.~ Fixed in PR 609
- test_missing_models expected file updated


aqjune-aws:
Could you resolve the conflict?

PR #596, opened 2026-03-17
Adding support for List Slice

thanhnguyen-aws:
*Description of changes:* Adding support for List slice subscription.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -498,7 +498,9 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
   -- TODO: Handle by creating explicit variable declarations
   | .Subscript _ val slice _ =>
     let dictOrList ← translateExpr ctx val
-    let index ← translateExpr ctx slice
+    let index ←  match slice with
+      | .Slice _ start stop step => translateSlice ctx start.val stop.val step.val
+      | _ => translateExpr ctx slice
     return mkStmtExprMd (.StaticCall "Any_get" [dictOrList, index])
 
   -- Attribute access: obj.attr or obj.method
@@ -546,8 +548,6 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
   -- Tuple literal: (1, 2)
   -- Abstract: return havoc'd tuple (sound abstraction)
   | .Tuple .. => return mkStmtExprMd .Hole
-
-  | .Slice _ start stop step => translateSlice ctx start.val stop.val step.val
 
   -- List comprehension: [x for x in items]
   -- Abstract: return havoc'd list (sound abstraction)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -522,6 +523,27 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+            let start := mkStmtExprMd (.StaticCall "Any..as_int!" [start])
+            return mkStmtExprMd (.StaticCall "from_Range" [start, optNone])
+        | none ,some stop =>
+            let start := mkStmtExprMd (.LiteralInt 0)
+            let stop ← translateExpr ctx stop
+            let stop := mkStmtExprMd (.StaticCall "Some" [mkStmtExprMd (.StaticCall "Any..as_int!" [stop])])
+            return mkStmtExprMd (.StaticCall "from_Range" [start, stop])
+        | _ , _ =>
+            let start := mkStmtExprMd (.LiteralInt 0)
+            return mkStmtExprMd (.StaticCall "from_Range" [start, optNone])
```
joehendrix: This is already a massive procedure.  Can you hoist this code into it a separate function (it may need to be in a mutual block)?

Even better if you could have the recursive calls done first and then have most of the logic in a separate function before translateExpr.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -522,6 +523,27 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
   -- Abstract: return havoc'd tuple (sound abstraction)
   | .Tuple .. => return mkStmtExprMd .Hole
 
+  | .Slice _ start stop _ =>
```
joehendrix: For soundness, this should either support step or throw a not-supported error if step is used.

joehendrix: Should Slice really even go in this function?  From some limited testing and the [AST documentation](https://docs.python.org/3/library/ast.html), it appears to only be allowed in subscription positions (e.g., `x[1:]`).

thanhnguyen-aws: When translating subscription, this function (`translateExpr`) is called to translate the Slice (line 477 `let index ← translateExpr ctx slice`).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -357,6 +358,31 @@ partial def translateDictStrAny (ctx : TranslationContext)
...
+            let start ← translateExpr ctx start
+            let stop ← translateExpr ctx stop
+            let start := mkStmtExprMd (.StaticCall "Any..as_int!" [start])
+            let stop := mkStmtExprMd (.StaticCall "Some" [mkStmtExprMd (.StaticCall "Any..as_int!" [stop])])
+            return mkStmtExprMd (.StaticCall "from_Range" [start, stop])
+        | some start, none =>
+            let start ← translateExpr ctx start
+            let start := mkStmtExprMd (.StaticCall "Any..as_int!" [start])
+            return mkStmtExprMd (.StaticCall "from_Range" [start, optNone])
+        | none ,some stop =>
```
joehendrix:
```suggestion
        | none, some stop =>
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -357,6 +358,31 @@ partial def translateDictStrAny (ctx : TranslationContext)
   let keys ← keys.mapM pyOptExprToString
   return  mkStmtExprMd (.StaticCall "from_Dict" [DictStrAny_mk (keys.zip val_trans)])
 
+partial def translateSlice (ctx : TranslationContext) (start stop step: Option (expr SourceRange))
+    : Except TranslationError StmtExprMd := do
+    if step.isSome then
+        throw (.unsupportedConstruct "Expression type not yet supported" (toString (repr step)))
+    else
```
joehendrix: The throw will exit, so the typical approach would be to remove the `else` here and not indent.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -522,6 +548,8 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
   -- Abstract: return havoc'd tuple (sound abstraction)
   | .Tuple .. => return mkStmtExprMd .Hole
 
+  | .Slice _ start stop step => translateSlice ctx start.val stop.val step.val
```
joehendrix: Should this be migrated into some code specific to index expressions `x[1:]` is valid Python, but `1:` by itself is not.  

I did a test and here's what the object is:
```
class Foo:
   def __getitem__(self, key):
     return key
>>> Foo()[1]
1
>>> Foo()[1:]
slice(1, None, None)
```

I think there should be a comment that fromRange is equivalent to the Python slice object (or the fromRange constructor should be renamed).

PR #597, opened 2026-03-17
feat: add --parallel flag for concurrent obligation verification

MikaelMayer:
## Parallel obligation verification via `--parallel N`

Adds a `--parallel N` flag to both `StrataVerify` and `pyAnalyzeLaurel` that runs up to N solver instances concurrently when verifying proof obligations. Without the flag, behavior is unchanged (sequential).

### Problem

Verification of programs with many obligations is bottlenecked by sequential solver invocations. Each obligation spawns a separate solver process, waits for the result, then moves to the next. This affects both Core verification (HeapReasoning: 53 goals) and Python/Laurel verification (27-47 prelude goals per test).

### Solution

When `--parallel N` is specified, obligations are dispatched in batches of N using `IO.asTask`. Each task runs an independent solver process. Results are collected in original obligation order, so output is deterministic and matches the sequential path exactly.

Changes:
- `VerifyOptions` gains a `parallelWorkers` field (default 1)
- `verifySingleEnv` refactored to extract per-obligation logic, with parallel and batched execution paths
- `StrataVerify` parses `--parallel N`
- `pyAnalyzeLaurel` parses `--parallel N`
- `run_examples.sh` reads optional `.flags` files from `expected/` to pass extra CLI flags per benchmark (e.g., `HeapReasoning.core.flags` contains `--parallel 5`)

### Benchmarks

**Core verification** (HeapReasoning.core.st, 53 goals, n=10):

| Mode | Wall time | Speedup |
|------|-----------|---------|
| Sequential | 6.99s ± 0.08s | baseline |
| `--parallel 5` | 3.78s ± 0.03s | 1.85x |
| `--parallel 10` | 3.58s ± 0.79s | 1.95x |

**Python/Laurel verification** (13 tests, 27-47 goals each):

| Mode | Total time | Speedup |
|------|-----------|---------|
| Sequential | 159s | baseline |
| `--parallel 5` | 136s | 1.17x |

Largest individual win: test_subscription 45.6s → 31.7s (1.44x).

### Testing

- All `lake test` guard_msgs tests pass
- All example expected outputs match
- Parallel output is identical to sequential output for both Core and Python paths


MikaelMayer:
We are going to implement this differently.

PR #598, opened 2026-03-17
Fix typing bugs in LaurelToCore

andrewmwells-amazon:
Fix typing bugs in LaurelToCore

### DIFF
--- a/Strata/Languages/Core/StatementType.lean
+++ b/Strata/Languages/Core/StatementType.lean
@@ -77,8 +77,10 @@ def typeCheckCmd (C: LContext CoreLParams) (Env : TEnv Unit) (P : Program) (c : 
         -- Add source location to error messages if not already present.
         .error <| e.withRangeIfUnknown (getFileRange md |>.getD FileRange.unknown)
 
-def typeCheckAux (C: LContext CoreLParams) (Env : TEnv Unit) (P : Program) (op : Option Procedure) (ss : List Statement) :
-  Except DiagnosticModel (List Statement × TEnv Unit × LContext CoreLParams) :=
+def typeCheckAux (C: LContext CoreLParams) (Env : TEnv Unit)
+    (P : Program) (op : Option Procedure) (ss : List Statement) :
+    Except DiagnosticModel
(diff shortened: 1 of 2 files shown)


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -61,7 +61,7 @@ def translateType (model : SemanticModel) (ty : HighTypeMd) : LMonoTy :=
   | .TCore s => .tcons s []
   | .TFloat64 => dbg_trace "NOT SUPPORTED YET: Float64"; .tcons "Float64IsNotSupportedYet" []
   | .TReal => LMonoTy.real
-  | .Top => LMonoTy.bool
+  | .Top => .tcons "Any" []
```
keyboardDrummer, line 64: Neither the before or the after are correct right? The new one seems like a hack specific to Python. The correct behavior is to abort compilation to Core, since .Top can't be represented in Core. I will also rename `top` to `Unknown`.

I think the right solution for Python is to avoid occurrences of the Top type, which we can do by avoiding resolution errors, for which we need to more the Python prelude definitions from Core to Laurel.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -305,12 +305,11 @@ Preserves the expression so it is not silently dropped from the Core output.
...
 private def exprAsUnusedInit (expr : StmtExprMd) (md : Imperative.MetaData Core.Expression)
     : TranslateM (List Core.Statement) := do
-  let model := (← get).model
   let coreExpr ← translateExpr expr
   let id ← freshId
   let ident : Core.CoreIdent := ⟨s!"$unused_{id}", ()⟩
-  let highTy := computeExprType model expr
-  let coreType := LTy.forAll [] (translateType model highTy)
+  let tyVarName := s!"$__ty_unused_{id}"
+  let coreType := LTy.forAll [tyVarName] (.ftvar tyVarName)
```
keyboardDrummer, line 312: What's wrong with the original? Actually I would have expected `let coreType := translateType model highTy`

PR #599, opened 2026-03-17
Add mutually recursive functions to Strata Core

joscoh:
*Issue #, if available:*

*Description of changes:*

This PR adds support for mutually recursive functions over ADTs to Strata Core. The SMT encoding and partial evaluation is almost exactly the same as for singular recursive functions, so the bulk of the changes are related to the concrete syntax. Specifically, the major changes are:

1. This PR adds a `@[preregisterFunctions]` mechanism which is very similar to the `@[preregisterTypes]` mechanism for datatypes. Similarly, the elaboration and translation to the Core AST is heavily based on the corresponding mechanisms for mutually recursive datatypes.
2. The AST now has a `recFuncBlock` `Decl`, enforcing a clear separation between non-recursive function definitions and mutually recursive blocks. Recursive functions are backwards compatible with the prior mechanism, but they are now implemented in the concrete syntax as single-element mutual blocks. This simplifies some of the typechecking logic, as well as the CST/AST translation.
3. The concrete syntax for mutually recursive functions is similar to that of mutually recursive types: blocks are terminated by semicolons, and individual mutually recursive elements within the block are separated by newlines. This is mostly backwards compatible with existing syntax (except for the required semicolons).
4. This PR adds a well-formedness predicate for mutual blocks and a proof in `ProgramWF.lean` that typechecking is sound for this predicate.
5. By unifying single and mutually recursive functions, this PR fixes a prior bug in the AST-to-CST translation of recursive functions.
6. `RecursiveFunctions.md` is updated with information about mutually recursive types and the current implementation.

Testing:
1. `RecursiveFunctionDDMTest.lean` contains a new test case for a mutually recursive function.
2. `TestASTtoCST.lean` contains test cases demonstrating that single and mutually recursive functions are correctly handled by `ASTtoCST`.
3. `MutualRecursiveFunctionErrorTest.lean` tests some error cases (e.g. no `@[cases]` annotation)
4. `MutualRecursiveFunctionTests.lean` contains various tests for evaluation and verification of goals involving mutually recursive functions, including those with preconditions.
5. `TermFormula.lean` contains a more sophisticated example of a deep embedding of terms and formulas, proving some simple theorems about the (mutually recursive) evaluation/denotation function.


Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -714,6 +714,19 @@ def translateTriggers (p: Program) (bindings : TransBindings) (arg : Arg) :
...
+partial def resolveRecFunc (funcs : List Core.Function) (idx : Nat) : TransM Core.Function := do
+  let gctx := (← StateT.get).globalContext
+  match gctx.nameOf? idx with
+  | some name =>
+    match funcs.find? (fun f => f.name.name == name) with
+    | some f => pure f
+    | none => TransM.error s!"function {name} not found in recFuncBlock"
+  | none =>
+    match funcs with
+    | f :: _ => pure f
```
shigoel: The fallback here silently returns the first function in the block when `gctx.nameOf? idx` returns `none`. If the name lookup fails for any reason, the wrong function's `opExpr` will be used in the translation with no diagnostic. This should be a `TransM.error` like the other failure branch, e.g.:
```lean
| none => TransM.error s!"resolveRecFunc: no name for index {idx} in global context"
```

Strata/Languages/Core/ProgramType.lean
```diff
@@ -124,25 +124,41 @@ C are already well-typed.
...
+          .error (DiagnosticModel.withRange fileRange <|
+            f!"recursive function block must contain at least one function")
+        -- Validate: no inline functions in the block
+        let _ ← funcs.foldlM (fun _ func => do
+          if func.attr.any (· == .inline) then
+            .error (DiagnosticModel.withRange fileRange <|
+              f!"recursive function '{func.name}' cannot be marked inline")
+          else pure ()) ()
+        -- Phase 1: Add ALL function signatures as stubs so mutual calls resolve
+        let C' := funcs.foldl (fun C func =>
```
shigoel, line 148: The Phase 1 stub registration uses a plain `foldl` (not `foldlM`), so if two functions in the block share a name the second silently overwrites the first stub in `C'` with no error. `WFRecFuncBlockProp.namesDistinct` requires uniqueness, but there's nothing here that enforces it at type-check time.

Consider either switching to a `foldlM` with an error-propagating variant of `addFactoryFunction`, or adding an upfront uniqueness check analogous to the `Identifiers.addListWithError` pattern used elsewhere:
```lean
let _ ← funcs.foldlM (fun seen func => do
  if seen.contains func.name then
    .error (DiagnosticModel.withRange fileRange
      f!"duplicate function name '{func.name}' in recursive block")
  else pure (seen.insert func.name)) (∅ : Std.HashSet _)
```

joscoh: The uniqueness of names was already checked earlier in the
```
let idents ← C.idents.addListWithError decl.names (fun n =>
      (DiagnosticModel.withRange fileRange f!"Error in {decl.kind} {n}: a declaration of this name already exists."))
```
line. It would be redundant to check again here.

joscoh: And in fact this is proved in Lean: `Program.typeCheck.goWF` proves that if typechecking succeeds, the well-formedness property (including `WFRecFuncBlockProp.namesDistinct`) holds.

Strata/DDM/Elab/Core.lean
```diff
@@ -1325,6 +1303,181 @@ partial def elaborateWithPreRegistration
...
+         expected 1 scope cat arg, got {scopeCat.args.size}"
+    return default
+  let some (_sep, getChildren) := scopeSepFormat scopeCat.name
+    | logInternalError fallbackLoc
+        s!"elaborateWithFunctionPreRegistration: \
+           unsupported scope category {scopeCat.name}"
+      return default
+  let children := getChildren scopeStx
+  -- Phase 1: Pre-register all function signatures
+  assert! tctx0.bindings.size = 0
```
shigoel: `assert!` panics on failure rather than producing a recoverable elaboration error. The surrounding code all uses `logInternalError` + `return default` for invariant violations (and the analogous `assert!` in `elaborateWithPreRegistration` at line 1296 has the same issue). This should follow the same pattern:
```lean
if tctx0.bindings.size ≠ 0 then
  logInternalError fallbackLoc "elaborateWithFunctionPreRegistration: expected empty local bindings"
  return default
```

Strata/Languages/Core/Program.lean
```diff
@@ -81,11 +85,15 @@ def Decl.name (d : Decl) : Expression.Ident :=
   | .distinct n _ _ => n
   | .proc p _       => p.header.name
   | .func f _       => f.name
+  | .recFuncBlock [] _ => ""
```
shigoel, line 90: Returning `""` for an empty `recFuncBlock` is a silent footgun. An empty block shouldn't occur in practice (the grammar requires at least one `recfn_decl` and `WFRecFuncBlockProp.nonempty` rejects it), but it can be constructed programmatically. A `panic!` with a clear message would fail loudly rather than propagating a meaningless empty-string name:
```lean
| .recFuncBlock [] _ => panic! "Decl.name: empty recFuncBlock"
```

joscoh: Do we want to introduce more `panic`s? I was unsure here.

shigoel: How invasive will `Except` be?

shigoel: I checked all references for `Decl.name`, and yeah, it could be pretty invasive.

shigoel: For now, I'd be okay to leave a comment here (along the lines of what Claude said above) so that we're aware of this problem. Another (probably very bad) idea is to return an internal free var (`__$bad_recFuncBlock`).

Strata/Languages/Core/ProgramType.lean
shigoel: Thanks! I think it's worth adding a comment here for easier context retrieval in the future.

Strata/DDM/Elab/Core.lean
```diff
@@ -1325,6 +1303,181 @@ partial def elaborateWithPreRegistration
...
+Analogous to `elaborateWithPreRegistration` for types, but pre-registers
+expression-level bindings (function signatures) instead of type-level bindings.
+
+- **Phase 1**: For each child, partially elaborate name, bindings, and return type
+  to extract the function signature.
+- **Pre-register**: Add all function signatures as expression-level bindings in the
+  GlobalContext so mutual calls resolve.
+- **Phase 2**: Fully elaborate each child with the updated context.
+-/
+partial def elaborateWithFunctionPreRegistration
```
aqjune-aws: It seems this function is pretty similar to `elaborateWithPreRegistration`. Other than the error messages and comments, `extractDatatypeInfo` vs. `extractFunctionInfo` seems to be the only the difference. Do you want to combine these two functions?

Strata/DDM/BuiltinDialects/StrataDDL.lean
```diff
@@ -178,7 +178,7 @@ def StrataDDL : Dialect := BuiltinM.create! "StrataDDL" #[initDialect] do
      Used on constructor args in datatype declarations so that type parameters are available
      as type variables inside constructor field types. -/
   declareMetadata { name := "scopeTVar", args := #[.mk "typeParams" .ident] }
-  declareMetadata { name := "scopeSelf", args := #[.mk "name" .ident, .mk "args" .ident, .mk "type" .ident] }
```
aqjune-aws, line 181: Could you update docs/verso/DDMDoc.lean so that `scopeSelf` is replaced with `preRegisterFunctions`?

Strata/Languages/Core/DDMTransform/Grammar.lean
aqjune-aws, line 330: Could you leave a short comment here suggesting that when `command_fndef` is to be edited, `recfn_decl` might need edits as well?

Strata/Languages/Core/Program.lean
joscoh: Sure, I can add a comment. `Except` would be very invasive. Since `Program.typeCheck` already ensures that these are non-empty, I am not too concerned with returning a default value for a case that is provably impossible for well-typed program (i.e. it is not just that the concrete grammar does not allow this).

Strata/Languages/Core/WF.lean
```diff
@@ -154,6 +154,12 @@ structure WFProcedureProp (p : Program) (d : Procedure) : Prop where
   wfspec : WFSpecProp p d.spec d
 structure WFFunctionProp (p : Program) (f : Function) : Prop where
 
+structure WFRecFuncBlockProp (p : Program) (fs : List Function) : Prop where
+  nonempty : fs ≠ []
+  namesDistinct : (fs.map (·.name)).Nodup
+  noInline : ∀ f ∈ fs, ¬ f.attr.any (· == .inline)
+  wfFuncs : ∀ f ∈ fs, WFFunctionProp p f
```
aqjune-aws, line 161: Can we also have existence of `inlineIfConstr`, or is there a restriction in having the property here?

StrataTest/Languages/Core/Examples/TermFormula.lean
```diff
@@ -0,0 +1,335 @@
+/-
```
aqjune-aws, line 1: This test is super cool. :)

Strata/Languages/Core/WF.lean
joscoh: I don't think that should be part of the typing rules. That is an evaluation-specific invariant, but we will eventually (likely) allow recursive functions that are not just structurally recursive over ADTs.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -986,6 +986,44 @@ def procToCST {M} [Inhabited M] (proc : Core.Procedure) : ToCSTM M (Command M) :
...
+    processInput id ty (casesIdx == some idx)
+  let bindings := results.map (·.1)
+  let paramNames := results.map (·.2)
+  let b : Bindings M := .mkBindings default ⟨default, bindings⟩
+  let r ← lmonoTyToCoreType func.output
+  modify (·.addScopedBoundVars (reverse? := false) paramNames)
+  let preconds ← precondsToSpecElts func.preconditions
+  let bodyExpr ← match func.body with
+    | some body => lexprToExpr body 0
+    | none => pure (.btrue default)  -- shouldn't happen for recursive functions
```
shigoel, line 1023: Makes sense to `logError` here, just in case.

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -337,14 +339,22 @@ op command_fndef (name : Ident,
...
+               typeArgs : Option TypeArgs,
+               @[scope(typeArgs)] b : Bindings,
+               @[scope(typeArgs)] r : Type,
+               @[scope(b)] preconds : Seq SpecElt,
+               @[scope(b)] c : r) : RecFnDecl =>
+  "function " name typeArgs b " : " r indent(2, preconds) "\n{\n  " indent(2, c) "\n}";
+
+@[scope(recfns), preRegisterFunctions(recfns)]
+op command_recfndefs (recfns : NewlineSepBy RecFnDecl) : Command =>
+  "rec " recfns ";\n";
```
shigoel, line 357: I somehow like the `;` after a `rec` block, but it's worth noting that Core's functions don't have a `;`.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
shigoel: (not blocking though; can do in a follow-up if you like)

PR #600, opened 2026-03-17
Fix issue 548 by special casing encoding of SMT types returned by LMonoTy.toSMTType

### LINKED ISSUE #548: Errant (declare-sort int 0) (declare-sort bool 0) in the generated SMTLib files
Found by Shilpi.

---

E.g., this is the SMTLib generated for https://github.com/strata-org/Strata/blob/main/StrataTest/Languages/Core/Examples/TypeDecl.lean#L82.

```
(reset)
(set-logic ALL)
(declare-sort int 0)
(declare-sort bool 0)
(declare-sort Foo 2)
; fooConst1
(declare-const fooConst1 (Foo Int Bool))
; fooVal
(declare-const fooVal (Foo Int Bool))
; fooConst2
(declare-const fooConst2 (Foo Int Bool))
(define-fun t0 () (Foo Int Bool) fooConst1)
(define-fun t1 () (Foo Int Bool) fooVal)
(define-fun t2 () Bool (= t0 t1))
(define-fun t3 () (Foo Int Bool) fooConst2)
(define-fun t4 () Bool (= t3 t1))
(assert t2)
(assert t4)
(define-fun t5 () Bool (= t0 t3))
(define-fun t6 () Bool (not t5))
(assert t6)
; Proof check
(set-info :file "")
(set-info :start 1675)
(set-info :stop 1723)
(set-info :sat-message "Assertion cannot be proven")
(check-sat)
(get-value (fooConst2 fooConst1 fooVal))
```

aqjune-aws:
*Issue #, if available:*

This resolves #548 and another internal issue due to introduction of unnecessary declare-sorts.

*Description of changes:*

### DIFF
--- a/Strata/Languages/Laurel/Resolution.lean
+++ b/Strata/Languages/Laurel/Resolution.lean
@@ -159,6 +159,9 @@ structure ResolveState where
   typeScopes : TypeScopes := {}
   /-- Diagnostics collected during resolution. -/
   errors : Array DiagnosticModel := #[]
+  /-- When resolving inside an instance procedure, the owning composite type name.
+      Used by `resolveFieldRef` to resolve `self.field` when `self` has type `Any`. -/
+  instanceTypeName : Option String := none
 
 @[expose] abbrev ResolveM := StateM ResolveState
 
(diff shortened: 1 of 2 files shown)


Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -190,7 +199,8 @@ def LMonoTy.toSMTType (E: Env) (ty : LMonoTy) (ctx : SMT.Context) (useArrayTheor
   | .tcons "Map" args =>
     -- When using Array theory, convert Map to Array
     let id := if useArrayTheory then "Array" else "Map"
-    let ctx := SMT.Context.addType E id args ctx
+    let ctx := if useArrayTheory then ctx
```
joscoh, line 202: Can SMT arrays have type parameters? If so, we still need to add the `args`, or else if the type arguments are datatypes they may not be emitted.

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -149,6 +149,13 @@ def extractTypeInstantiations (typeVars : List String) (patterns : List LMonoTy)
     Map.empty
 
 
+/--
+Returns true if the given type name is a built-in Core type whose SMT-LIB
+encoding is handled specially and should not be declared via `declare-sort`.
+-/
+def isBuiltinCoreTy (id : String) : Bool :=
```
shigoel, line 156: Is the goal to sync. this with 
https://github.com/strata-org/Strata/blob/main/Strata/Languages/Core/Factory.lean#L32?

shigoel: How are bitvecs handled?

Strata/Languages/Core/SMTEncoder.lean
shigoel: +1

aqjune-aws: `LMonoTys.toSMTType` is doing the registration of the remaining type arguments. `LMonoTys.toSMTType` is returning a list of TermType.

Strata/Languages/Core/SMTEncoder.lean
aqjune-aws: This is to sync with `LMonoTy.toSMTType`. Bitvec doesn't need to be handled because it isn't represented with `.tcons`. On the other hand, int/bool/... of Core types are represented with `.tcons "int" []`, etc.

joscoh (approved):

PR #601, opened 2026-03-17
Fix handling of shadowed fields names in __init__ methods

andrewmwells-amazon:
Fix handling of shadowed fields names in __init__ methods

### DIFF
--- a/Strata/Languages/Core/ProcedureEval.lean
+++ b/Strata/Languages/Core/ProcedureEval.lean
@@ -28,7 +28,39 @@ def fixupError (E : Env) : Env :=
                      pathConditions := E.pathConditions.pop }
   | some _ => E
 
-def eval (E : Env) (p : Procedure) : List (Procedure × Env) :=
+/--
+Merge multiple procedure evaluation results into one.
+
+After `fixupError`, all paths through a procedure have identical variable state
+and path conditions — the procedure scope and its path-condition scope have been
+popped, leaving only the outer (global) scope which is the same on every path.
+The differences across paths are:
+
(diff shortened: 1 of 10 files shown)


Strata/Languages/Laurel/Resolution.lean
```diff
@@ -159,6 +159,9 @@ structure ResolveState where
   typeScopes : TypeScopes := {}
   /-- Diagnostics collected during resolution. -/
   errors : Array DiagnosticModel := #[]
+  /-- When resolving inside an instance procedure, the owning composite type name.
+      Used by `resolveFieldRef` to resolve `self.field` when `self` has type `Any`. -/
```
keyboardDrummer, line 163: This PR doesn't make sense to me. If `self` is a datatype, like the current `Any` in Python, then you can't do a field access on it in Laurel. I think it would be better to revert these changes.

What you might want to do is define 
```java
function readAny(a: Any, key: String): Any 
  requires Any..isFrom_Dict(a) || Any..isFromClassInstance
{
  if (Any..isFrom_Dict(a))
    readDictStrAny(a, key)
  else
    readDictStrAny(Any..fromClassInstance(a).instance_attributes, key)
}

function readDictStrAny(dictionary: DictStrAny, key: String): Any {
  ...
}
```

then when translating a Python field read from containing type `Any`, instead of using a Laurel `FieldSelect`, do a StaticCall to `readAny`.

For a field update, you could define a 
```
composite MutableDictStrAny {
  var value: DictStrAny
}

function readMutableDictStrAny(...

procedure updateMutableDictStrAny(..
```

PR #603, opened 2026-03-18
feat: Add --check-mode and --check-level flags to pyAnalyzeLaurel

MikaelMayer:
Expose `--check-mode` and `--check-level` flags on the `pyAnalyzeLaurel` command, matching the existing flags on `StrataVerify`.

Adds `VerificationMode.ofString?` and `CheckLevel.ofString?` to `Options.lean` as the single source of truth for string parsing and help text. Both `pyAnalyzeLaurel` and `StrataVerify` now use these shared functions instead of hardcoded string matching.

Existing tests pass. No behavioral change for `StrataVerify`.

### DIFF
--- a/Strata/DDM/Elab.lean
+++ b/Strata/DDM/Elab.lean
@@ -36,6 +36,84 @@ inductive Header where
 | program (loc : SourceRange) (name : DialectName)
 deriving Inhabited
 
+private inductive QuantifierSepState where
+| outside
+| inBinder
+| sawColon
+
+/--
+Canonicalize the legacy dotted Unicode quantifier separator to `::` before DDM
+parsing. This keeps `#strata` generic while accepting both `∀ x . P` and
+`∀ x :: P`.
+-/
+private def normalizeUnicodeQuantifierSeparators (src : String) : String :=
+  (src.foldl
+    (init := ("", QuantifierSepState.outside))
+    (fun (st : String × QuantifierSepState) (ch : Char) =>
+      let (acc, qstate) := st
+      match qstate with
+      | .outside =>
+        if ch == '∀' || ch == '∃' then
+          (acc.push ch, .inBinder)
+        else
+          (acc.push ch, .outside)
+      | .inBinder =>
+        if ch == '.' then
+          (acc ++ "::", .outside)
+        else if ch == ':' then
+          (acc.push ch, .sawColon)
+        else
+          (acc.push ch, .inBinder)
+      | .sawColon =>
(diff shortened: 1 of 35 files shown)


StrataVerify.lean
```diff
@@ -70,8 +66,8 @@ def usageMessage : Std.Format :=
   --output-format=sarif       Output results in SARIF format to <file>.sarif{Std.Format.line}  \
   --vc-directory=<dir>        Store VCs in SMT-Lib format in <dir>{Std.Format.line}  \
   --solver <name>             SMT solver executable to use (default: {defaultSolver}){Std.Format.line}  \
-  --check-mode <mode>         Check mode: 'deductive' (default, prove correctness), 'bugFinding' (find bugs), or 'bugFindingAssumingCompleteSpec' (find bugs assuming complete preconditions).{Std.Format.line}  \
-  --check-level <level>       Check level: 'minimal' (default, simple messages), 'minimalVerbose' (detailed messages, one check), or 'full' (both checks, all outcomes)."
+  --check-mode <mode>         Check mode: {Core.VerificationMode.options}. Default: 'deductive'.{Std.Format.line}  \
```
shigoel, line 69: Why are the descriptions removed from the help message? The names in `Core.VerificationMode.options` are not that self-evident...

StrataMain.lean
```diff
@@ -104,6 +104,26 @@ def buildDialectFileMap (pflags : ParsedFlags) : IO Strata.DialectFileMap := do
...
 
+def parseCheckMode (pflags : ParsedFlags) : VerificationMode :=
+  match pflags.getString "check-mode" |>.bind VerificationMode.ofString? with
+  | .some m => m
+  | .none => .deductive
+
+def parseCheckLevel (pflags : ParsedFlags) : CheckLevel :=
+  match pflags.getString "check-level" |>.bind CheckLevel.ofString? with
+  | .some l => l
+  | .none => .minimal
```
shigoel: Both `parseCheckMode` and `parseCheckLevel` silently fall back to the default when an invalid string is passed — e.g. `--check-mode bugfinding` (lowercase typo) will silently produce `.deductive` with no warning. The `none` branch handles both "flag not given" and "bad value" identically.

This is inconsistent with `StrataVerify`, which returns an error for invalid values (`parseOptions` returns `Except Std.Format ...`).

Consider threading an error out of these functions (e.g. returning `IO VerificationMode` or `Except String VerificationMode`) and printing a diagnostic when the flag value is present but unrecognized.

PR #604, opened 2026-03-18
Add polymorphic Sequence type to Strata Core

atomb:
Adds `Sequence a`, a polymorphic indexed sequence abstraction, to the Strata Core dialect with 9 operations: `empty`, `length`, `select`, `append`, `build`, `update`, `contains`, `take`, and `drop`.

# What's included

Factory definitions (`Factory.lean`): Each operation is axiomatized with triggers following the same pattern as the existing Map type. Key axiom properties:

* `length` is non-negative; build increments length by 1; append sums lengths; update preserves length
* `take`/`drop` length axioms are guarded with `0 <= n <= length(s)` to prevent unsound derivations at out-of-bounds indices
* `build`/`update` have read-over-write semantics for select
* `contains` is defined via an existential quantifier over valid indices

The axioms have corresponding models based on the `List` type in `Strata.Languages.Core.SeqModel`.

# DDM grammar and translation

`Grammar.lean`, `ASTtoCST.lean`, `Translate.lean`: All 8 parseable operations are added to the grammar, with corresponding CST↔AST translation cases. Sequence.empty is registered in the Factory for programmatic use but is not yet parseable from .st files because the DDM parser cannot currently handle 0-ary polymorphic functions (documented with TODO comments in all three files).

# Tests

`Seq.lean`: Two test programs exercise all operations through the full pipeline (parse → translate → type-check → VCG → SMT → verify). Includes a deliberate negative assertion (t_length_wrong) to confirm the verifier doesn't vacuously accept. 10 of 11 assertions in `seqOpsPgm` pass; `contains_yes returns unknown` due to the existential quantifier in the axiom — expected behavior with universally quantified axiom-based SMT encoding.

# Existing test updates

* `TypeDecl.lean`: `KnownTypes` list updated
* `PolymorphicDatatypeTest.lean`: renamed Seq → SeqTest to avoid collision
* `ProcedureEvalTests.lean` and `ProgramTypeTests.lean` expected output updated.

# Known limitations

* `Sequence.empty` is not yet parseable from `.st` files (0-ary polymorphic function limitation in DDM).

Written almost entirely by Kiro.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -1160,6 +1160,20 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     | some lbl => return (ctx, [mkStmtExprMdWithLoc (StmtExpr.Exit lbl) md])
     | none => return (ctx, [mkStmtExprMdWithLoc (StmtExpr.Assert (mkStmtExprMd .Hole)) md])
 
+  -- Augmented assignment: x += expr  →  x = x op expr
+  | .AugAssign _ target op value => do
+    let targetExpr ← translateExpr ctx target
+    let valueExpr ← translateExpr ctx value
+    let rhs := match op with
+      | .Add _      => mkStmtExprMd (StmtExpr.StaticCall "PAdd"      [targetExpr, valueExpr])
+      | .Sub _      => mkStmtExprMd (StmtExpr.StaticCall "PSub"      [targetExpr, valueExpr])
+      | .Mult _     => mkStmtExprMd (StmtExpr.StaticCall "PMul"      [targetExpr, valueExpr])
+      | .FloorDiv _ => mkStmtExprMd (StmtExpr.StaticCall "PFloorDiv" [targetExpr, valueExpr])
+      | .Mod _      => mkStmtExprMd (StmtExpr.StaticCall "PMod"      [targetExpr, valueExpr])
+      | _           => mkStmtExprMd .Hole
+    let assignStmt := mkStmtExprMdWithLoc (StmtExpr.Assign [targetExpr] rhs) md
+    return (ctx, [assignStmt])
+
   | _ => throw (.unsupportedConstruct "Statement type not yet supported" (toString (repr s)))
 
 partial def translateStmtList (ctx : TranslationContext) (stmts : List (Python.stmt SourceRange))
--- a/StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
@@ -1,0 +1,34 @@
+
+==== Verification Results ====
+List_get_body_calls_List_get_0: ✅ pass
+List_take_body_calls_List_take_0: ✅ pass
+List_drop_body_calls_List_drop_0: ✅ pass
+List_slice_body_calls_List_drop_0: ✅ pass
+List_slice_body_calls_List_take_1: ✅ pass
+List_set_body_calls_List_set_0: ✅ pass
+DictStrAny_get_body_calls_DictStrAny_get_0: ✅ pass
+Any_get_body_calls_DictStrAny_get_0: ✅ pass
+Any_get_body_calls_List_get_1: ✅ pass
+Any_get!_body_calls_DictStrAny_get_0: ✅ pass
+Any_get!_body_calls_List_get_1: ✅ pass
+Any_set_body_calls_List_set_0: ✅ pass
+Any_set!_body_calls_List_set_0: ✅ pass
+PFloorDiv_body_calls_Int.SafeDiv_0: ✅ pass
+PFloorDiv_body_calls_Int.SafeDiv_1: ✅ pass
+PFloorDiv_body_calls_Int.SafeDiv_2: ✅ pass
+PFloorDiv_body_calls_Int.SafeDiv_3: ✅ pass
+PAnd_body_calls_Any_to_bool_0: ✅ pass
+POr_body_calls_Any_to_bool_0: ✅ pass
+ret_type: ✅ pass (in prelude file)
+ret_type: ✅ pass (in prelude file)
+ret_pos: ✅ pass (in prelude file)
+assert_name_is_foo: ✅ pass (in prelude file)
+assert_opt_name_none_or_str: ✅ pass (in prelude file)
+assert_opt_name_none_or_bar: ✅ pass (in prelude file)
+ensures_maybe_except_none: ✅ pass (in prelude file)
+assert_assert(81)_calls_Any_to_bool_0: ✅ pass (at line 5, col 4)
+assert(81): ✅ pass (at line 5, col 4)
+assert_assert(124)_calls_Any_to_bool_0: ✅ pass (at line 7, col 4)
+assert(124): ✅ pass (at line 7, col 4)
+assert_assert(167)_calls_Any_to_bool_0: ✅ pass (at line 9, col 4)
+assert(167): ✅ pass (at line 9, col 4)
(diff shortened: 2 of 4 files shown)


Strata/Languages/Core/Factory.lean
```diff
@@ -270,6 +275,268 @@ def mapUpdateFunc : WFLFunc CoreLParams :=
...
+      esM[∀ (Sequence %a): -- %2 s
+          (∀ (int): -- %1 i
+            (∀ (%a): -- %0 v
+              {((~Sequence.length : (Sequence %a) → int)
+                ((((~Sequence.update : (Sequence %a) → int → %a → (Sequence %a)) %2) %1) %0))}
+              ((~Sequence.length : (Sequence %a) → int)
+                ((((~Sequence.update : (Sequence %a) → int → %a → (Sequence %a)) %2) %1) %0))
+              ==
+              ((~Sequence.length : (Sequence %a) → int) %2)))],
+      -- select(update(s, i, v), i) == v  (same index)
```
joscoh: This needs a precondition that `i` is within bounds, I believe.

Strata/Languages/Core/Factory.lean
```diff
@@ -270,6 +275,268 @@ def mapUpdateFunc : WFLFunc CoreLParams :=
...
+                    ((~Sequence.length : (Sequence %a) → int) %1)))
+            then
+              ((~Sequence.length : (Sequence %a) → int)
+                (((~Sequence.drop : (Sequence %a) → int → (Sequence %a)) %1) %0))
+              ==
+              (((~Int.Sub : int → int → int)
+                ((~Sequence.length : (Sequence %a) → int) %1))
+                %0)
+            else #true)],
+      -- select(drop(s, n), j) == select(s, j + n)
```
joscoh: I also think we should have some bounds here.

Strata/Languages/Core/Factory.lean
```diff
@@ -270,6 +275,268 @@ def mapUpdateFunc : WFLFunc CoreLParams :=
...
+              {(((~Sequence.select : (Sequence %a) → int → %a)
+                    (((~Sequence.build : (Sequence %a) → %a → (Sequence %a)) %2) %1)) %0)}
+              if (%0 == ((~Sequence.length : (Sequence %a) → int) %2))
+              then
+                (((~Sequence.select : (Sequence %a) → int → %a)
+                    (((~Sequence.build : (Sequence %a) → %a → (Sequence %a)) %2) %1)) %0)
+                == %1
+              else #true))],
+      -- select(build(s, v), i):
+      --   i != length(s) ==> select(build(s,v), i) == select(s, i)
```
joscoh: Does this need to know that `i` is in the correct bound? Of course it should be the case that there is a constant default value, but I don't know if we want to enforce that in the axiomatization?

Strata/Languages/Core/Factory.lean
```diff
@@ -270,6 +275,268 @@ def mapUpdateFunc : WFLFunc CoreLParams :=
...
+      -- select(update(s, i, v), i) == v  (same index)
+      esM[∀ (Sequence %a): -- %2 s
+          (∀ (int): -- %1 i
+            (∀ (%a): -- %0 v
+              {(((~Sequence.select : (Sequence %a) → int → %a)
+                  ((((~Sequence.update : (Sequence %a) → int → %a → (Sequence %a)) %2) %1) %0)) %1)}
+              (((~Sequence.select : (Sequence %a) → int → %a)
+                  ((((~Sequence.update : (Sequence %a) → int → %a → (Sequence %a)) %2) %1) %0)) %1)
+              == %0))],
+      -- select(update(s, i, v), n) == select(s, n)  when n != i
```
joscoh: Same here

Strata/Languages/Core/Factory.lean
```diff
@@ -41,7 +45,8 @@ def KnownLTys : LTys :=
    -- We can simply add the following here.
    t[∀n. bitvec n],
    t[∀a b. %a → %b],
-   t[∀a b. Map %a %b]]
+   t[∀a b. Map %a %b],
+   t[∀a. Sequence %a]]
```
joscoh, line 49: Does `Sequence` need to be special cased like `Array` in https://github.com/strata-org/Strata/pull/600?

Strata/Languages/Core/Factory.lean
atomb: Yep, I think you're right. Good catch. (Here and in the other cases.)

Strata/Languages/Core/Factory.lean
atomb: I think not, since we're not using the SMT sequence theory, unless I'm missing something.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -527,6 +533,13 @@ def handleBinaryOps {M} [Inhabited M] (name : String)
   | "Bool.Equiv" => pure (.equiv default arg1 arg2)
   -- Map operations
   | "select" => pure (.map_get default ty ty arg1 arg2)
+  -- Sequence operations
+  | "Sequence.select" => pure (.seq_get default ty arg1 arg2)
```
shigoel: Nit: Curious why the internal name is `select` instead of `get`. All other ops have the same name.

Strata/Languages/Core/Factory.lean
```diff
@@ -270,6 +275,298 @@ def mapUpdateFunc : WFLFunc CoreLParams :=
...
+      esM[∀ (Sequence %a): -- %0 s
+        {((~Sequence.length : (Sequence %a) → int) %0)}
+        (((~Int.Ge : int → int → bool)
+          ((~Sequence.length : (Sequence %a) → int) %0))
+          #0)]
+    ])
+
+/- An empty `Sequence` constructor with type `∀a. Sequence a`.
+   NOTE: This is registered in the Factory for programmatic use, but is not yet
+   parseable from `.st` files because the DDM grammar cannot currently handle
```
shigoel, line 293: Can we file an issue/feature request from DDM to track this?

Strata/Languages/Core/Factory.lean
```diff
@@ -270,6 +275,298 @@ def mapUpdateFunc : WFLFunc CoreLParams :=
                     ))))]
     ])
 
+/- A `Sequence` length function with type `∀a. Sequence a → int`. -/
+def seqLengthFunc : WFLFunc CoreLParams :=
+  polyUneval "Sequence.length" ["a"]
+    [("s", seqTy mty[%a])] mty[int]
+    (axioms := [
```
shigoel, line 282: I'm not quite sure about using the `axioms` field here (but yes, it's nice that the axioms are validated in Lean). My understanding was that we'll be deprecating that. Also, note that proofs about FactoryWF don't take `axioms` into account.

shigoel: @aqjune-aws @joscoh Thoughts?

shigoel: Oh sorry, I see that Josh and Aaron talked about this already. Sounds good -- so this can be out of scope of this PR, but I'd still like thoughts about FactoryWF proofs.

atomb: My initial thought is that I'd ultimately like for FactoryWF proofs to link axioms to verified models. We're getting close to where we could do that. There might be an even better solution, though.

Strata/Languages/Core/Factory.lean
atomb: Yep, I'll file one once this is merged (so it can link to the specific code where this comes up).

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
atomb: I'll make them all the same.

Strata/Languages/Core/Factory.lean
joscoh: I don't think we will directly want that to be part of FactoryWF, since in my mind, FactoryWF should involve syntactic checks whereas a verified model needs reasoning about typing and semantics (likely relying on the WF invariants). But I agree that this is something we want to do, even if it is a separate predicate.

PR #607, opened 2026-03-18
Add #guard_msgs to Boole output, fix bug in translation of recursive functions

joscoh:
*Issue #, if available:*

*Description of changes:*
1. Remove `guard_msgs` from output. Anything with trivial output uses a `guard_msgs (drop info)`, and everything else uses `Options.quiet`
2. Fix bug where recursive functions were not translated to functions with `isRecursive := true`. This makes the test in `mutual_recursion.lean` fail (since recursive functions over ints are not yet supported), so it is commented out

CC @kondylidou 


StrataTest/Languages/Boole/FeatureRequests/abstract_types_and_stubs.lean
```diff
@@ -41,6 +41,7 @@ spec {
 };
 #end
 
+#guard_msgs (drop info) in
```
shigoel, line 44: It looks like this `verify` is basically used to make sure that type checking succeeds, right? Hence the `drop info`, but not `drop errors`?

shigoel: If so, a comment would help but also okay to leave it out of scope of this PR.

joscoh: I believe so, but maybe @kondylidou can confirm the original purpose of the #evals

kondylidou: I checked: `#guard_msgs (drop info)` here is effectively using `#eval Strata.Boole.verify` as a smoke test that the Boole verification pipeline runs without parse/translation/type-check/internal errors. It intentionally ignores the normal printed VC results. The actual proof that obligations succeed is the following `smtVCsCorrect` theorem/example, not the `#eval` itself. So a short comment to that effect would be reasonable.

PR #608, opened 2026-03-18
Propagate Python assert messages as Laurel property summaries

tautschnig:
Python assert with string message (e.g., assert x > 0, "positive") now flows through as a Laurel property summary. CBMC displays these as property descriptions instead of opaque labels.

Before: [main.1] line 2 assert(38): SUCCESS
After:  [main.1] line 2 sum is reflexive: SUCCESS

Only simple string literals are supported; f-strings and non-string messages are silently ignored (the label is used as fallback).

### DIFF
--- a/StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
@@ -1,14 +1,14 @@
-test_arithmetic.py(7, 4): ✅ pass - assert_assert(102)_calls_Any_to_bool_0
-test_arithmetic.py(7, 4): ✅ pass - assert(102)
-test_arithmetic.py(12, 4): ✅ pass - assert_assert(226)_calls_Any_to_bool_0
-test_arithmetic.py(12, 4): ✅ pass - assert(226)
-test_arithmetic.py(16, 4): ✅ pass - assert_assert(345)_calls_Any_to_bool_0
-test_arithmetic.py(16, 4): ✅ pass - assert(345)
-test_arithmetic.py(20, 4): ✅ pass - assert_assert(458)_calls_Any_to_bool_0
-test_arithmetic.py(20, 4): ✅ pass - assert(458)
+test_arithmetic.py(7, 4): ✅ pass - multiplication implemented incorrectly
+test_arithmetic.py(7, 4): ✅ pass - multiplication implemented incorrectly
+test_arithmetic.py(12, 4): ✅ pass - multiplication implemented incorrectly
+test_arithmetic.py(12, 4): ✅ pass - multiplication implemented incorrectly
+test_arithmetic.py(16, 4): ✅ pass - addition implemented incorrectly
+test_arithmetic.py(16, 4): ✅ pass - addition implemented incorrectly
+test_arithmetic.py(20, 4): ✅ pass - subtraction implemented incorrectly
+test_arithmetic.py(20, 4): ✅ pass - subtraction implemented incorrectly
 test_arithmetic.py(23, 4): ✅ pass - set_quot_calls_PFloorDiv_0
-test_arithmetic.py(24, 4): ✅ pass - assert_assert(567)_calls_Any_to_bool_0
-test_arithmetic.py(24, 4): ✅ pass - assert(567)
+test_arithmetic.py(24, 4): ✅ pass - floor division implemented incorrectly
+test_arithmetic.py(24, 4): ✅ pass - floor division implemented incorrectly
 test_arithmetic.py(26, 0): ✅ pass - ite_cond_calls_Any_to_bool_0
 DETAIL: 12 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success
--- a/StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
@@ -1,8 +1,8 @@
-test_augmented_assign.py(5, 4): ✅ pass - assert_assert(81)_calls_Any_to_bool_0
-test_augmented_assign.py(5, 4): ✅ pass - assert(81)
-test_augmented_assign.py(7, 4): ✅ pass - assert_assert(124)_calls_Any_to_bool_0
-test_augmented_assign.py(7, 4): ✅ pass - assert(124)
-test_augmented_assign.py(9, 4): ✅ pass - assert_assert(167)_calls_Any_to_bool_0
-test_augmented_assign.py(9, 4): ✅ pass - assert(167)
+test_augmented_assign.py(5, 4): ✅ pass - 5 + 3 == 8
+test_augmented_assign.py(5, 4): ✅ pass - 5 + 3 == 8
+test_augmented_assign.py(7, 4): ✅ pass - 8 - 2 == 6
+test_augmented_assign.py(7, 4): ✅ pass - 8 - 2 == 6
+test_augmented_assign.py(9, 4): ✅ pass - 6 * 2 == 12
+test_augmented_assign.py(9, 4): ✅ pass - 6 * 2 == 12
 DETAIL: 6 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success
--- a/StrataTest/Languages/Python/expected_laurel/test_boolean_logic.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_boolean_logic.expected
@@ -1,30 +1,30 @@
-test_boolean_logic.py(5, 4): ✅ pass - assert_assert(73)_calls_PAnd_0
-test_boolean_logic.py(5, 4): ✅ pass - assert_assert(73)_calls_Any_to_bool_1
-test_boolean_logic.py(5, 4): ✅ pass - assert(73)
-test_boolean_logic.py(6, 4): ✅ pass - assert_assert(124)_calls_PAnd_0
-test_boolean_logic.py(6, 4): ✅ pass - assert_assert(124)_calls_Any_to_bool_1
-test_boolean_logic.py(6, 4): ✅ pass - assert(124)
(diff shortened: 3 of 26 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -971,8 +971,12 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     return (ctx, [retStmt])
 
   -- Assert statement
-  | .Assert _ test _msg => do
+  | .Assert _ test msg => do
     let condExpr ← translateExpr ctx test
+    -- Extract assert message as property summary
+    let md := match msg.val with
+      | some (.Constant _ (.ConString _ str) _) => md.withPropertySummary str.val
+      | _ => md
```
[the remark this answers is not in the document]
tautschnig: Done.

StrataTest/Languages/Python/tests/test_property_summary.sh
```diff
@@ -0,0 +1,54 @@
...
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/../../../.." && pwd)"
+STRATA="$PROJECT_ROOT/.lake/build/bin/strata"
+DIALECT="$PROJECT_ROOT/Tools/Python/dialects/Python.dialect.st.ion"
+PYTHON=${PYTHON:-python3}
+
+WORK=$(mktemp -d)
+trap "rm -rf $WORK" EXIT
```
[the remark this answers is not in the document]
tautschnig: Done.

StrataTest/Backends/CBMC/GOTO/test_property_summary_e2e.sh
```diff
@@ -0,0 +1,109 @@
...
+    var y: int := 3;
+    assert x + y == 8 summary "addition equals eight";
+    assert x - y == 2 summary "difference equals two"
+};
+LAUREL
+
+# Step 2: Translate to GOTO
+cd "$WORK"
+"$STRATA" laurelAnalyzeToGoto test.lr.st 2>&1 | grep -q "Written"
+
```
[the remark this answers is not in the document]
tautschnig: Done.

StrataTest/Backends/CBMC/GOTO/test_property_summary_e2e.sh
```diff
@@ -0,0 +1,109 @@
...
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/../../../.." && pwd)"
+STRATA="$PROJECT_ROOT/.lake/build/bin/strata"
+CBMC=${CBMC:-cbmc}
+SYMTAB2GB=${SYMTAB2GB:-symtab2gb}
+
+WORK=$(mktemp -d)
+trap "rm -rf $WORK" EXIT
```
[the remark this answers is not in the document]
tautschnig: Done.

Strata/Languages/Python/PythonToLaurel.lean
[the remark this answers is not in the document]
tautschnig: Done.

StrataTest/Languages/Python/tests/test_property_summary.sh
```diff
@@ -0,0 +1,54 @@
...
+# instruction with the message as the sourceLocation comment.
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/../../../.." && pwd)"
+STRATA="$PROJECT_ROOT/.lake/build/bin/strata"
+DIALECT="$PROJECT_ROOT/Tools/Python/dialects/Python.dialect.st.ion"
+PYTHON=${PYTHON:-python3}
+
```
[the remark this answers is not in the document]
tautschnig: Done.

StrataTest/Languages/Python/tests/test_property_summary.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
keyboardDrummer: I have several concerns with a test like this:
- It interweaves test framework and test contents code
- The test framework part is not reusable. 
- It mixes programming languages
- It's not run on `lake test`
- It's not obvious what the user experience when running Strata on the input program

Can we write this test in Lean so it'll run on `lake test` ?
Can we place the results that the user will see, inline in the test?

tautschnig: Thank you for pushing me to do the right thing, I admittedly had been cutting a corner here. I believe [b1c0e0e](https://github.com/strata-org/Strata/pull/608/commits/b1c0e0e3181961134496c6c672115c206bf7f398) implements your request.

StrataTest/Backends/CBMC/GOTO/test_property_summary_e2e.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+# E2E test: property summary metadata flows from Laurel through GOTO to CBMC output.
+#
+# Verifies that Laurel `assert ... summary "..."` annotations appear in
+# CBMC's verification output as property descriptions.
+
+set -eo pipefail
```
MikaelMayer, line 7: I feel uncomfortable having a bash script to do this kind of testing. First is this script executed anywhere in the CI? Second, as soon as we want to make more end to end tests, it's likely agents will simply want to make a variation of this file, meaning things will get duplicated and hard to maintain.
At the minimum I would like to see the test program and expectations in two different separate files from the machinery that verifies that the current strata converts the test to the expectations correctly.
That way the machinery can ideally be reused as well.

tautschnig: Yes, this is part of CI. If you are uncomfortable with bash, what would you be comfortable with instead?

StrataTest/Languages/Python/PropertySummaryTest.lean
```diff
@@ -0,0 +1,72 @@
...
+Verifies that `assert cond, "message"` in Python flows through as a
+property summary in the Core verification results.
+-/
+
+namespace Strata.Python.PropertySummaryTest
+
+open Strata (pyTranslateLaurel verifyCore)
+
+private meta def testPy : String :=
+"def main(x: int) -> None:
```
keyboardDrummer: IMO we should use the same style of testing as Laurel uses for Python as well, so this test would simplify to a `.lean` file containing

```lean
def program := # strata
program Python;

def main(x: int) -> None:
    assert x == x + 1, "a big lie"
#          ^^^^^^^^^^ error: a big lie does not hold

#end

#guard_msgs(drop info, error) in
#eval testAnnotatedPython program
```

I think we should prioritize enabling our Python tests to be written like that.

tautschnig: Ack, but do you want me to do this as part of this PR, in a separate PR that is then to be merged before this one, or merge this one first and then do that work in a separate PR?

keyboardDrummer: Since there's no rush for this PR, I would suggest the first or middle option, but if you feel strongly we can do the last one.

tautschnig: You have seen it already, but for the sake of completeness: https://github.com/strata-org/Strata/pull/658.

tautschnig: Now using the infrastructure from #658.

StrataTest/Backends/CBMC/GOTO/test_property_summary_e2e.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+# E2E test: property summary metadata flows from Laurel through GOTO to CBMC output.
```
keyboardDrummer, line 2: Sorry but why do we need a bash file to add tests? Can't we add all tests using Lean files so they're run through `lake test` ?

tautschnig: We still need to invoke external tools like symtab2gb, goto-cc, cbmc - surely this can also be done from Lean code, we just haven't been doing this so far.

keyboardDrummer: We've been running some back-ends and their related tools from Lean, like cvc5 and z3, so I think we should feel free to run the tools you mention, from Lean, just as well.

MikaelMayer: The advantage of remaining is lean is multifold. No need to add a line to the CI so that these tests are run. We can configure how the tests are parsed/redacted. We can share the infrastructure with other tests. We can embed both the test and the expected output.

Even my agents complain that Bash is "painful to maintain" and prefer to switch to python. Whatever you choose, I would still prefer to avoid embedding test programs in bash files. If something is written in python, let's create a python file and read it. If something is written in Core, let's create a core.st file and read it.
The only case when feel ok _embedding_ the source code of a test is in a lean test file, and even in that case I would request that there be no more than one helper to process the test, so that we can add more tests easily.

I don't feel comfortable having a bash script that writes a test file on disk just to test it.

tautschnig: I have created https://github.com/strata-org/Strata/issues/771. Yes, this should be done, but I'd rather not spend extended time resolving conflicts for various PRs in flight.

shigoel:
@tautschnig quick note (admittedly haven’t had the chance to review this PR yet). My understanding of Python assert messages is that they only print when the condition is False, so they are typically used to encode messages like “BUG in foo bar”. Will Strata print these messages irrespective of assertion success or failure?

PR #609, opened 2026-03-19
Fix type checking bugs in Python->Laurel translation

thanhnguyen-aws:
*Description of changes:* This PR fix some bugs in Python->Laurel translation which result in type-checking failure.
Changes:
1. Ignore comments in source code: currently those comments are translated to literal string expressions.
2. Translate InstanceCall to Hole: to transform them together with other .Hole expressions before translating to Core
3. Remove .Hole processing in `if`, `while`, and `assert`.
4. Add variable declarations for error variables which have names in TryCatch statement.
5. Add Box constructors and destructors for TCore types: neccessary in case the field type is Any.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -1588,7 +1588,7 @@ def pythonToLaurel' (info : PreludeInfo)
 
     -- FIRST PASS: Collect all class definitions and field type info
     let mut compositeTypes : List CompositeType := [pyErrorTy]
-    let mut compositeTypeNames := info.compositeTypes
+    let mut compositeTypeNames := info.compositeTypes.insert "PythonError"
     let mut classFieldHighType : Std.HashMap String (Std.HashMap String HighType) := {}
     for stmt in body.val do
       match stmt with


Strata/Languages/Python/PythonToLaurel.lean
olivier-aws: This code should be removed as Holes are now properly taken care of (https://github.com/strata-org/Strata/commit/554235c57f928650773478826f4fb7f4fb37b524)

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -761,10 +761,11 @@ partial def translateCall (ctx : TranslationContext)
     else [trans_kwords]
   match f with
   | .Name  _ _ _ =>  return mkStmtExprMd (StmtExpr.StaticCall funcName (trans_args ++ trans_kwords_exprs))
-  | .Attribute _ val attr _ =>
-      let target_trans ← translateExpr ctx val
+  | .Attribute _ val _attr _ =>
+      let _target_trans ← translateExpr ctx val
       if opt_firstarg.isSome then
-        return mkStmtExprMd (StmtExpr.InstanceCall target_trans attr.val (trans_args ++ trans_kwords_exprs))
+        return mkStmtExprMd (.Hole)
```
joehendrix, line 767: **Regression (may be out of scope for this PR):** For `client.get_item(Bucket="b")` where `client: Storage`, `funcName` resolves to `"Storage_get_item"` which has a model (prelude procedure). But this branch returns `Hole`, discarding the modeled behavior entirely. This means all instance method calls on typed objects — even when they have full prelude models — produce unconstrained values in the analysis.

I understand `InstanceCall` was also broken — in `LaurelToCoreTranslator.lean` it panics as an expression (line 286), havocs as an initializer (line 365), and is a no-op as a statement (line 433). So `Hole` avoids the panics, and fixing this properly likely requires two things that are out of scope here: (1) supporting `InstanceCall` in the Core translator, and (2) fixing the PySpec pipeline which currently strips `self` and produces static procedures so instance method calls can't round-trip correctly.

Flagging this so we can track it — could we file an issue for proper `InstanceCall` support? In the meantime, a narrower improvement would be to use `StaticCall funcName` here (same as the `else` branch on line 770) for PySpec methods, since they're already static procedures. That would at least preserve the modeled behavior for the common case.

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -178,7 +178,8 @@ def boxDestructorName (model : SemanticModel) (ty : HighType) : Identifier :=
   | .UserDefined name =>
       if isDatatype model name then s!"Box..{name.text}Val!"
       else "Box..compositeVal!"
-  | _ => dbg_trace "BUG, boxDestructorName bad type "; "boxDestructorNameError"
+  | .TCore name => s!"Box..{name}Val"
```
joehendrix: **Question:** The TCore destructor returns `s!"Box..{name}Val"` without the `!` suffix, but all other destructors use `!` (e.g., `"Box..intVal!"`, `s!"Box..{name.text}Val!"`). Is this intentional? If TCore destructors should follow the same convention, this should be:

```suggestion
  | .TCore name => s!"Box..{name}Val!"
```

**Nit:** The docstrings for `boxDestructorName` and `boxConstructorName` (lines 169-171, 184-186) should mention the new TCore case.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1116,9 +1069,12 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
         match optVars.val with
         | some varExpr =>
           let varName := pyExprToString varExpr
-          let varDecl := mkStmtExprMd (StmtExpr.LocalVariable varName AnyTy (some enterCall))
-          currentCtx := {currentCtx with variableTypes := currentCtx.variableTypes ++ [(varName, PyLauType.Any)]}
-          setupStmts := setupStmts ++ [mgrDecl, varDecl]
+          if varName ∈ currentCtx.variableTypes.unzip.fst then
+            setupStmts := setupStmts ++ [mgrDecl]
```
joehendrix: **Bug:** When the `as` variable already exists (e.g., `f = None; with open("x") as f:`), this branch skips `__enter__()` entirely — so the variable is never updated, and the context manager protocol is broken (`__exit__` still runs at line 1080). The guard correctly avoids a duplicate `LocalVariable`, but should emit an `Assign` instead so `__enter__()` is still called and its result stored:

```suggestion
          if varName ∈ currentCtx.variableTypes.unzip.fst then
            let assignStmt := mkStmtExprMd (StmtExpr.Assign
              [mkStmtExprMd (StmtExpr.Identifier varName)] enterCall)
            setupStmts := setupStmts ++ [mgrDecl, assignStmt]
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1621,8 +1577,15 @@ def pythonToLaurel' (info : PreludeInfo)
...
+      name := {text := "PythonError"}
+      extending := []  -- No inheritance support for now
+      fields := [{name:= "response", isMutable:= false, type:= AnyTy}]
+      instanceProcedures := []
+    }
+
     -- FIRST PASS: Collect all class definitions and field type info
-    let mut compositeTypes : List CompositeType := []
+    let mut compositeTypes : List CompositeType := [pyErrorTy]
     let mut compositeTypeNames := info.compositeTypes
```
joehendrix: **Issue:** `pyErrorTy` is added to `compositeTypes` but `"PythonError"` is never inserted into `compositeTypeNames`. The hardcoded `|| tyStr == "PythonError"` checks on lines 1005 and 1011 work around this, but other code paths (e.g., `hasModel`, `translateType`, `translateAssign`) won't recognize PythonError as composite. Suggest:

```suggestion
    let mut compositeTypes : List CompositeType := [pyErrorTy]
    let mut compositeTypeNames := info.compositeTypes.insert "PythonError"
```

Then the `|| tyStr == "PythonError"` special cases on lines 1005 and 1011 can be removed.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1043,19 +992,23 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
         | .Assign _ targets _ _ => targets.val.toList.head?.map (fun t => (pyExprToString t, PyLauType.Any))
         | _ => none
     let bodyDecls := collectDeclaredNamesAndTypes body.val.toList
+
+    let errorVarDecls: List (String × String) := (handlers.val.toList.filterMap (λ h => match h with
+          | .ExceptHandler _ _ errname _ => errname.val)).map (λ h => (h.val, "PythonError"))
```
joehendrix, line 997: **Regression:** Exception variables typed as `PythonError` (a Composite/UserDefined type) fail Core type checking when used in contexts expecting `Any` — e.g., `str(e)` produces "Impossible to unify". Reproduced locally with:

```python
def except_var_usage() -> bool:
    client: Storage = servicelib.connect("storage")
    try:
        client.put_item(Bucket="b", Key="k", Data="v")
    except Exception as e:
        msg: str = str(e)
        return False
    return True
```

The hoisted error variable gets type `PythonError` but needs to be usable as `Any`. Consider either typing it as `AnyTy` instead, or providing a coercion path from `PythonError` to `Any`.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I agree. I think we should merge this PR first and have a separate PR for InstanceCall handling.

aqjune-aws (approved):

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -270,7 +274,7 @@ where
         if calleeWritesHeap then
           if valueUsed then
             let freshVar ← freshVarName
-            let varDecl := mkMd (.LocalVariable freshVar ⟨.TInt, #[]⟩ none)
+            let varDecl := mkMd (.LocalVariable freshVar (computeExprType model exprMd) none)
```
keyboardDrummer, line 277: Need to add a test for this to the Laurel tests. I'll put that on my TODO.

PR #615, opened 2026-03-19
feat: Translate PySpec preconditions to Laurel assertions for bug finding

MikaelMayer:
Translate PySpec preconditions into Laurel assertions for bug-finding verification.

When a PySpec function has `assert` preconditions (e.g. `assert len(kwargs["name"]) >= 1`),
the generated Laurel procedure now contains `Assert` statements instead of an opaque body.
After inlining at call sites, the solver can prove whether preconditions are satisfied
or always violated — detecting real bugs statically.

Changes:
- `Specs/ToLaurel.lean`: Translate `SpecExpr` to Laurel `StmtExpr` using Core prelude
  functions (`Str.Length`, `Any..as_string!`, `Any..as_int!`, `from_int`,
  `DictStrAny_contains`). Untranslatable preconditions emit `assert true` with a
  diagnostic. `kwargs["field"]` references resolve to expanded parameter names.
- `PySpecPipeline.lean`: Thread type aliases (unprefixed → prefixed class names) so
  Python type annotations resolve to pyspec composite types.
- `PythonToLaurel.lean`: Resolve type aliases for composite type creation and method
  dispatch. Emit `StaticCall` instead of `InstanceCall` when the method resolves to a
  known pyspec procedure.
- `StrataMain.lean`: Add procedure inlining to `pyAnalyzeLaurel` so pyspec assertions
  are checked at call sites with concrete arguments.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -1749,8 +1749,8 @@ def pythonToLaurel' (info : PreludeInfo)
   }
 
   -- Separate functions from other statements
-  let mut procedures : List Procedure := []
-  let mut otherStmts : List (Python.stmt SourceRange) := []
+  let mut procedures : Array Procedure := #[]
+  let mut otherStmts : Array (Python.stmt SourceRange) := #[]
 
   for stmt in body do
     match stmt with
@@ -1758,14 +1758,14 @@ def pythonToLaurel' (info : PreludeInfo)
       let funcDecl ←  pyFuncDefToPythonFunctionDecl ctx stmt
       let proc ← translateFunction ctx stmt.ann funcDecl fbody.val.toList
       ctx := {ctx with functionSignatures:= ctx.functionSignatures ++ [funcDecl]}
-      procedures := procedures ++ [proc.fst]
+      procedures := procedures.push proc.fst
     | .ClassDef _ _ _ _ _ _ _ =>
       pure ()  -- Already processed in first pass
     | _ =>
-      otherStmts := otherStmts ++ [stmt]
+      otherStmts := otherStmts.push stmt
 
   ctx := {ctx with variableTypes:= [("nullcall_ret", PyLauType.Any)]}
-  let (_, bodyStmts) ← translateStmtList ctx otherStmts
+  let (_, bodyStmts) ← translateStmtList ctx otherStmts.toList
   let bodyStmts := prependExceptHandlingHelper bodyStmts
   let bodyStmts := mkStmtExprMd (.LocalVariable "__name__" AnyTy (some <| strToAny "__main__")) :: bodyStmts
   let bodyStmts := (mkStmtExprMd (.LocalVariable "nullcall_ret" AnyTy (some AnyNone))) :: bodyStmts
@@ -1785,7 +1785,7 @@ def pythonToLaurel' (info : PreludeInfo)
   }
 
   let program : Laurel.Program := {
-    staticProcedures := procedures ++ [mainProc]
+    staticProcedures := procedures.push mainProc |>.toList
     staticFields := []
     types := compositeTypes.map TypeDefinition.Composite
     constants := []


Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -301,7 +413,18 @@ def funcDeclToLaurel (procName : String) (func : FunctionDecl)
     | .TVoid => []
     | _ => [{ name := "result", type := retType }]
   if func.preconditions.size > 0 || func.postconditions.size > 0 then
```
MikaelMayer: This condition fires even when only preconditions are present (no postconditions), emitting a misleading "Postconditions not yet supported" warning for functions that only have preconditions — which are now supported. Should be:

```suggestion
  if func.postconditions.size > 0 then
```

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,116 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+/-- Format an Assertion's message parts into a string for diagnostics. -/
+private def formatAssertionMessage (msg : Array MessagePart) : String :=
+  let parts := msg.map fun
+    | .str s => s
+    | .expr _ => "<expr>"
+  String.join parts.toList
+
+/-- Build a procedure body that asserts preconditions and havocs outputs. -/
+def buildSpecBody (preconditions : Array Assertion) (outputs : List Parameter)
+    : ToLaurelM (Body × Nat) := do
```
MikaelMayer: Nit: `translatedCount` is returned but always discarded by the caller (`let (body, _) ← buildSpecBody ...`). Consider either dropping it from the return type or using it for a diagnostic (e.g., logging how many preconditions were successfully translated vs. emitted as `assert true`).

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: Fixed — now only warns when postconditions are present.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: Dropped — buildSpecBody now returns just Body.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,114 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+  for assertion in preconditions do
+    match specExprToLaurel assertion.formula with
+    | some condExpr =>
+      stmts := stmts ++ [mkStmt (.Assert condExpr)]
+    | none =>
+      -- Emit assert true for untranslatable preconditions so they're visible but don't block
+      let msg := formatAssertionMessage assertion.message
+      reportError default s!"Untranslatable precondition (emitting assert true): {msg}"
+      stmts := stmts ++ [mkStmt (.Assert (mkStmt (.LiteralBool true)))]
+  -- Havoc each output by assigning a nondeterministic hole
```
MikaelMayer: We absolutely don't need to do that. Outputs declarations are already nondeterministic.

MikaelMayer: Removed — outputs are already nondeterministic, no havoc needed.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,110 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+  | .forallDict _ _ _ _ => none -- TODO: quantifiers
+
+/-- Format an Assertion's message parts into a string for diagnostics. -/
+private def formatAssertionMessage (msg : Array MessagePart) : String :=
+  let parts := msg.map fun
+    | .str s => s
+    | .expr _ => "<expr>"
+  String.join parts.toList
+
+/-- Build a procedure body that asserts preconditions and havocs outputs. -/
```
MikaelMayer: Docstring says "havocs outputs" but the function only emits `Assert` statements — it doesn't havoc anything. Docstring should say "outputs are already initialized non-deterministically".

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,110 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+             mkStmt (.FieldSelect s (mkId field))
+  | .isInstanceOf _ _ => none -- TODO: isinstance checks
+  | .len subject =>
+    -- len(x) where x is Any: Str.Length(Any..as_string!(x)) wrapped back as from_int
+    subject |> specExprToLaurel |>.map fun s =>
+      let unwrapped := mkStmt (.StaticCall (mkId "Any..as_string!") [s])
+      mkStmt (.StaticCall (mkId "from_int") [mkStmt (.StaticCall (mkId "Str.Length") [unwrapped])])
+  | .intGe subject bound => do
+    -- Compare as ints: Any..as_int!(subject) >= Any..as_int!(bound)
+    let s ← specExprToLaurel subject
```
MikaelMayer: The `intGe`/`intLe` cases are identical except for the comparison operator (`.Geq` vs `.Leq`), and `floatGe`/`floatLe` follow the same pattern without the `Any..as_int!` unwrapping. A small helper like `translateComparison (unwrapInt : Bool) (op : PrimitiveOp) (subject bound : SpecExpr)` would remove the duplication across all four cases.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,110 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+  String.join parts.toList
+
+/-- Build a procedure body that asserts preconditions and havocs outputs. -/
+def buildSpecBody (preconditions : Array Assertion)
+    : ToLaurelM Body := do
+  let mut stmts : List StmtExprMd := []
+  for assertion in preconditions do
+    match specExprToLaurel assertion.formula with
+    | some condExpr =>
+      stmts := stmts ++ [mkStmt (.Assert condExpr)]
```
MikaelMayer: `stmts := stmts ++ [x]` is O(n) per append on `List`, making the loop O(n²). Consider prepending (`x :: stmts`) and reversing at the end, which is the project convention for list building.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -792,8 +800,15 @@ partial def translateCall (ctx : TranslationContext)
   | .Attribute _ val _attr _ =>
       let _target_trans ← translateExpr ctx val
       if opt_firstarg.isSome then
-        return mkStmtExprMd (.Hole)
-        --return mkStmtExprMd (StmtExpr.InstanceCall target_trans attr.val (trans_args ++ trans_kwords_exprs))
+        -- Only emit StaticCall when the caller type was resolved through a
+        -- pyspec type alias, indicating the procedure has precondition assertions.
+        let callerTy := match f with
```
MikaelMayer: The outer match already binds `val` from `.Attribute _ val _attr _`, but this inner match re-destructures `f` to get `v` which is the same value. You can just use `val` directly:
```
let callerTy := (inferExprType ctx val).toOption.getD ""
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1681,16 +1700,26 @@ def pythonToLaurel' (info : PreludeInfo)
     let mut ctx : TranslationContext := match prev_ctx with
     | some prev_ctx => prev_ctx
     | _ =>
+    -- Add unprefixed method names as procedure aliases
+    -- e.g. MessageService_send → same signature as prefix_MessageService_send
+    let preludeProcedures := info.typeAliases.fold (init := info.procedures)
+      fun procs unprefixed prefixed =>
+        procs.fold (init := procs) fun procs' name sig =>
```
MikaelMayer: The inner `procs.fold` iterates over the accumulator `procs` (which grows as aliases are inserted), not the original `info.procedures`. This means aliases inserted for one type could be re-examined when processing the next type alias. It works correctly because the `startsWith` guard is specific enough, but folding over `info.procedures` instead of `procs` would be clearer about intent and avoid the growing-accumulator subtlety.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: Fixed.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: Fixed.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: Fixed.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Fixed.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Fixed.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -91,6 +91,8 @@ structure TranslationContext where
   filePath : String := ""
   /-- Known composite type names (user-defined classes + PySpec types) -/
   compositeTypeNames : Std.HashSet String := {}
+  /-- Maps unprefixed class names to prefixed names (from pyspec). -/
```
MikaelMayer: I don't understand why this typeAliases map is here. This file is about translating python to Laurel. Although similar, PySpec is a separate language on its own. I understand we are re-using the translation context, but could you please explain why type aliases are necessary here?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1546,6 +1560,8 @@ structure PreludeInfo where
   functions : List String := []
   /-- Procedure names (non-function callables) -/
   procedureNames : List String := []
+  /-- Maps unprefixed class names to prefixed names (from pyspec). -/
```
MikaelMayer: Aliases are specific to files, aren't they? Could you please explain by replying to this thread why it seems like a top-level thing?

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,100 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
     | .typedDict _ _ _ => return mkCore "DictStrAny"
 
+/-! ## SpecExpr to Laurel Translation -/
+
+/-- Wrap a StmtExpr with default metadata. -/
+private def mkStmt (e : StmtExpr) : StmtExprMd :=
+  { val := e, md := .empty }
+
+/-- Translate a SpecExpr to a Laurel StmtExpr.
+    All values are assumed to be Core(Any)-typed, matching the Python pipeline.
```
MikaelMayer: Any is not a Core type, it's defined in a python prelude, so at best it's a datatype defined in Laurel.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,100 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+    -- kwargs["fieldName"] → direct reference to the expanded parameter
+    | .var "kwargs" => some (mkStmt (.Identifier (mkId field)))
+    | _ => subject |> specExprToLaurel |>.map fun s =>
+             mkStmt (.FieldSelect s (mkId field))
+  | .isInstanceOf _ _ => none -- TODO: isinstance checks
+  | .len subject =>
+    -- len(x) where x is Any: Str.Length(Any..as_string!(x)) wrapped back as from_int
+    subject |> specExprToLaurel |>.map fun s =>
+      let unwrapped := mkStmt (.StaticCall (mkId "Any..as_string!") [s])
+      mkStmt (.StaticCall (mkId "from_int") [mkStmt (.StaticCall (mkId "Str.Length") [unwrapped])])
```
MikaelMayer, line 301: Isn't there a function `Any..len` ? If not, I'd recommend adding one so that we can support the length of more things in the future rather than hard-coding the length to be a string.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,100 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+  | .intGe subject bound => do
+    let s ← specExprToLaurel subject; let b ← specExprToLaurel bound
+    some (mkStmt (.PrimitiveOp .Geq [mkStmt (.StaticCall (mkId "Any..as_int!") [s]), mkStmt (.StaticCall (mkId "Any..as_int!") [b])]))
+  | .intLe subject bound => do
+    let s ← specExprToLaurel subject; let b ← specExprToLaurel bound
+    some (mkStmt (.PrimitiveOp .Leq [mkStmt (.StaticCall (mkId "Any..as_int!") [s]), mkStmt (.StaticCall (mkId "Any..as_int!") [b])]))
```
MikaelMayer: There are a lot of redundancies between these two cases. Please refactor common parts into a helper defined before the match.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,100 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+      let sStr := mkStmt (.StaticCall (mkId "Any..as_string!") [s])
+      let eqs := values.toList.map fun v =>
+        mkStmt (.PrimitiveOp .Eq [sStr, mkStmt (.LiteralString v)])
+      eqs.foldl (init := mkStmt (.LiteralBool false)) fun acc eq =>
+        mkStmt (.PrimitiveOp .Or [acc, eq])
+  | .containsKey container key =>
+    -- DictStrAny_contains(Any..as_Dict!(container), key)
+    container |> specExprToLaurel |>.map fun c =>
+      let unwrapped := mkStmt (.StaticCall (mkId "Any..as_Dict!") [c])
+      mkStmt (.StaticCall (mkId "DictStrAny_contains") [unwrapped, mkStmt (.LiteralString key)])
```
MikaelMayer: Aren't we supposed to wrap boolean results with `Any..from_bool`? or is this wrappinga lready done in `DictStrAny_contains` ?

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,100 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+  | .regexMatch subject pattern =>
+    -- Assumes future Laurel functions:
+    --   string_to_regex(pattern: string) -> Regex
+    --   regex_matches(regex: Regex, subject: string) -> bool
+    subject |> specExprToLaurel |>.map fun s =>
+      let sStr := mkStmt (.StaticCall (mkId "Any..as_string!") [s])
+      let regex := mkStmt (.StaticCall (mkId "string_to_regex") [mkStmt (.LiteralString pattern)])
```
MikaelMayer: For all the regex methods,please make it so that it will be compatible with the upcoming PR
https://github.com/strata-org/Strata/pull/623

StrataTest/Languages/Python/PreconditionTest.lean
```diff
@@ -0,0 +1,99 @@
...
+
+meta import Strata.SimpleAPI
+meta import Strata.Languages.Python.PySpecPipeline
+meta import StrataTest.Util.Python
+
+/-! ## End-to-end tests for PySpec precondition assertion translation
+
+These tests verify that PySpec `assert` preconditions are translated into
+Core assertions. The pyspec file `MessageService.py` defines `send()` with
+three `len` preconditions. The test checks that the generated Core program
```
MikaelMayer: For an end-to-end test, we should not need to check that the generated core program contains assert statements from the pyspec procedure. That's a microtest that might change in the future if we moved preconditions to expressions preconditions instead of statements.
I expect the end-to-end test only have a failing call and the ability to detect 2 or more failing preconditions. This means that in the pipeline, we should be able to indicate that a procedure should be inlined prior to verifying it.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: typeAliases is needed in TranslationContext because refineFunctionCallExpr resolves method calls like svc.send() to the prefixed pyspec procedure name (e.g. MessageService_send → prefix_MessageService_send). Without it, the translator can't map the Python type annotation to the pyspec composite type. I agree it's a PySpec concern leaking into the Python translator — an alternative would be to resolve aliases in PySpecPipeline before building the context, so TranslationContext only sees the resolved names. Happy to refactor if you prefer that approach.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: You're right, aliases are per-file (per pyspec module). They're merged into PreludeInfo because multiple pyspec files can be loaded (via --pyspec), and the Python translator needs to see all aliases at once. The merge happens in buildPreludeInfo. Each file's aliases are scoped by the module prefix so there's no collision.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: You're right — Any is a Laurel datatype defined in PythonRuntimeLaurelPart.lean, not a Core type. TCore "Any" works because the Laurel→Core translator maps it to .tcons "Any" [] which matches the Core prelude's Any datatype. But the naming is misleading. I should use .UserDefined or reference the Laurel datatype directly. What's the preferred way to reference the Laurel Any type from ToLaurel.lean?

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: No, there's no Any..len in the prelude. List_len exists for ListAny, and Str.Length is a Core built-in for strings. I agree we should add an Any-level len function that dispatches based on the wrapped type (string → Str.Length, list → List_len). I'll add it to the prelude.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: The intGe/intLe cases are now 2 lines each (down from 5), and floatGe/floatLe are also 2 lines. I tried extracting a helper but it can't reference specExprToLaurel (mutual recursion) and Operation isn't in scope inside a local let. The remaining duplication is the do/bind + mkStmt pattern which is inherent to the Option monad. Open to suggestions on how to factor this further.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: DictStrAny_contains returns bool, not Any. Since the Assert condition in Laurel is translated as a pure Core expression that needs to be bool, returning raw bool is correct here — the assert doesn't need Any wrapping. PIn wraps with from_bool because it returns Any for use in Python expressions, but our assert conditions should stay bool.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: PR 623 introduces re_search_str(pattern: string) → regex and uses Str.InRegEx for matching. I'll update regexMatch to use re_search_str for compilation and Str.InRegEx for the match check, matching PR 623's approach.

StrataTest/Languages/Python/PreconditionTest.lean
MikaelMayer: Agreed — the test should verify actual bug detection, not string inspection. The blocker was that ProcedureInlining depends on sorry-using code (StatementWF), which prevents #eval from running. I'll look into using #eval! with the inlining step, or restructuring the test to call the CLI pipeline directly.

Strata/Languages/Python/Specs/ToLaurel.lean
MikaelMayer: I was just thinking about changing the comment that says `Core(Any)`

MikaelMayer: Fixed — updated the docstring to say 'Any-typed (the Python prelude universal type)' instead of 'Core(Any)-typed'.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -259,6 +261,99 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
...
+/-- Wrap a StmtExpr with default metadata. -/
+private def mkStmt (e : StmtExpr) : StmtExprMd :=
+  { val := e, md := .empty }
+
+/-- Translate a SpecExpr to a Laurel StmtExpr.
+    All values are assumed to be Any-typed (the Python prelude's universal type).
+    Returns `none` for unsupported expressions (placeholders).
+    Uses Core prelude function names (Any_len, DictStrAny_contains, etc.)
+    which are resolved after the Core prelude is prepended. -/
+partial def specExprToLaurel (e : SpecExpr) : Option StmtExprMd :=
```
MikaelMayer: There should be a better type that takes multiple errors. There is a monad around `ToLaurelM`

MikaelMayer: Good point — specExprToLaurel returns Option which loses error context. I'll change it to ToLaurelM (Option StmtExprMd) so it can report diagnostics for unsupported expressions while still returning none for the caller to emit assert true.

StrataTest/Languages/Python/Specs/precondition_test/MessageService.py
```diff
@@ -0,0 +1,13 @@
+from typing import TypedDict, Required, NotRequired, Unpack
+
+SendRequest = TypedDict('SendRequest', {
+    'url': Required[str],
+    'body': Required[str],
+    'timeout': NotRequired[int],
+})
+
+class MessageService:
+    def send(self, **kwargs: Unpack[SendRequest]) -> None:
```
MikaelMayer: I've heard that there was another place where we tests such pyspecs integrations, and it might be better to just add assertions there. Moreover, make sure we can detect assertions to always fail, that's the purpose of having assertions

MikaelMayer: The existing pyspec integration tests are in StrataTest/Languages/Python/Specs/dispatch_test/ (tested by AnalyzeLaurelTest.lean). I can add precondition assertions to the existing Storage.py spec there and add a test case that calls with invalid args to detect always-fail. The blocker for detecting always-fail is that ProcedureInlining depends on sorry-using code, so the inlined assertions can't be verified in #eval tests. I'll investigate using the CLI pipeline or #eval! to work around this.

Strata/DDM/Integration/Lean/OfAstM.lean
```diff
@@ -222,6 +222,9 @@ def ofOptionM {α β} [Repr α] [SizeOf α]
       pure { ann := ann, val := none }
     | some v =>
       (fun v => { ann := ann, val := some v }) <$> act v (by decreasing_tactic)
+  -- Treat Init.boolFalse as none (used by some generators to represent absent optional values)
+  | .op ⟨ann, q`Init.boolFalse, #[]⟩ =>
+      pure { ann := ann, val := none }
```
joehendrix: Is there a bug report somewhere?  This seems like a workaround to a bug in a serializer?

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: The typeAliases map in PreludeInfo is per-file: each pyspec file produces its own aliases (e.g. Storage.py → {"Storage": "Storage_Storage"}), and they're merged in buildPySpecLaurel when combining multiple pyspec files. PreludeInfo is the natural place since it already carries per-file procedure signatures and type definitions that get merged the same way. The map is then threaded into TranslationContext so that refineFunctionCallExpr can resolve unprefixed class names to their prefixed procedure names at call sites.

Strata/DDM/Integration/Lean/OfAstM.lean
MikaelMayer: The pyspec ion serializer uses Init.boolFalse to represent absent optional values in TypedDict field declarations (e.g. `mkClassFieldDecl(name, type, constValue)` where `constValue : Option Str`). The serializer emits `Init.boolFalse` instead of the DDM `option(none)` wrapper. I can file a bug against the serializer — this workaround can be removed once it's fixed there.

StrataTest/Languages/Python/PreconditionTest.lean
MikaelMayer: Done — replaced the string-inspection test with a full verification pipeline test in AnalyzeLaurelTest.lean. It runs inlining + SMT verification on `test_precondition_violation.py` (calls `put_item(Bucket="")`) and checks for `✖️ always false if reached (Bucket must not be empty)` via `#guard_msgs`. PreconditionTest.lean and MessageService.py removed.

StrataTest/Languages/Python/Specs/precondition_test/MessageService.py
MikaelMayer: Done — added precondition assertions to the existing Storage.py in dispatch_test/ and added `test_precondition_violation.py` with a `#guard_msgs` verification test. MessageService.py and PreconditionTest.lean removed.

StrataTest/Languages/Python/AnalyzeLaurelTest.lean
```diff
@@ -214,29 +214,28 @@ private meta def runTestCase (dispatchIon tmpDir : System.FilePath)
...
+          | .label "message", .msg s => some s
+          | _, _ => none
+        let msgStr := msg.map (s!" ({·})") |>.getD ""
+        IO.println s!"{r.obligation.label}: {r.formatOutcome}{msgStr}"
+      let alwaysFalse := storageResults.filter fun r =>
+        match r.outcome with
+        | .ok o => o.alwaysFalseAndReachable || o.alwaysFalseReachabilityUnknown
+        | _ => false
+      if alwaysFalse.size == 0 then
+        throw <| IO.userError "Expected ✖️ always false for empty Bucket violation"
```
MikaelMayer: Why did you remove the nice user message errors? You must give a strong reason to do that as the test would now be much less maintainable and understandable

StrataTest/Languages/Python/Specs/dispatch_test/Storage.py
```diff
@@ -24,9 +24,11 @@
 
 class Storage:
     def put_item(self, **kwargs: Unpack[PutItemRequest]) -> None:
-        ...
+        assert len(kwargs["Bucket"]) >= 1, "Bucket must not be empty"
```
MikaelMayer, line 27: I want regular expression tests as well.

MikaelMayer: Meaning having at least one assert with a regular expression test that pyspec can parse and that the end to end workflow can detect is wrong

StrataTest/Languages/Python/AnalyzeLaurelTest.lean
MikaelMayer: You're right — I've restored the human-readable output test. However, the full `#guard_msgs` verification test (with ✖️/✔️ output) is currently blocked by a PR #623 interaction: the Core type checker fails on `re_fullmatch_str` forward declarations when running `verifyCore` in interpreted mode. The CLI pipeline works correctly (it uses the native binary's type checker). For now the test checks that the Core program contains the right assertions (str.len, isfrom_none, Storage_Storage_put_item). Once the PR #623 type checker issue is resolved, we can restore the `#guard_msgs` with full verification output.

StrataTest/Languages/Python/Specs/dispatch_test/Storage.py
MikaelMayer: The regex translation is enabled (`re_search_bool` from PR #623), but the pySpecs generator (our Python tool) doesn't parse `re.search(...)` in assertions — it produces a `placeholder` instead of a `regexMatch` SpecExpr. The botomoog generator handles regex correctly. To add a regex test to Storage.py, we'd need to extend the pySpecs Python generator to recognize `re.search` calls. The `specExprToLaurel` regex translation is tested end-to-end via the S3Vectors/SSM benchmarks which use botomoog-generated pyspec files.

Strata/DDM/Integration/Lean/OfAstM.lean
joehendrix: I can't find any serialization code that could do this, so I'm removing there workaround.

PR #618, opened 2026-03-19
Bug fix: Put the datatypes defined inside Python prelude on the top of Laurel Prelude.

thanhnguyen-aws:
*Description of changes:* This PR put the datatypes defined inside Python prelude on the top of Laurel Prelude. This fixes the type-checking error when adding Box constructor/destructor for Any type.


keyboardDrummer (changes requested):
I don't think this change is right. The current setup allows anything in the Core part of the prelude to refer to the Laurel part of the prelude, but not the other way around. I think that is sufficient.

If there's a definition in the Core part of the prelude that you want to refer to from the Laurel prelude part, you should include that definition in the Laurel prelude as well.

PR #619, opened 2026-03-19
Move new variables declaration out of `if` , `else`, `while` and `for` blocks

thanhnguyen-aws:
*Description of changes:* This PR remove new variables declaration inside `if` , `else`, `while` and `for` blocks and attach them before the those  statements because Python allows using those variables outside of those blocks (they are not local).

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -980,6 +980,8 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
   | .While _ test body _orelse => do
     -- Note: Python while-else not supported yet
     let condExpr ← translateExpr ctx test
+    let newDecls := collectDeclaredNamesAndTypes body.val.toList
+    let (varDecls, ctx) := createVarDeclStmtsAndCtx ctx newDecls
     let breakLabel := s!"loop_break_{test.toAst.ann.start.byteIdx}"
     let continueLabel := s!"loop_continue_{test.toAst.ann.start.byteIdx}"
     let loopCtx := { ctx with loopBreakLabel := some breakLabel, loopContinueLabel := some continueLabel }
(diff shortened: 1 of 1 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -271,6 +271,7 @@ def lookupFieldHighType (ctx : TranslationContext) (className fieldName : String
     | some ty => .ok ty
 
 def NoError : StmtExprMd := mkStmtExprMd (StmtExpr.StaticCall "NoError" [])
+def optNone := mkStmtExprMd (StmtExpr.StaticCall "None" [])
```
joehendrix, line 274: Nit: `optNone` doesn't appear to be referenced anywhere in this file or the repo (the only other `optNone` is an unrelated DDM operator in `StrataTest/DDM/Ion.lean`). Given the three "remove unexpected change" commits, this looks like one that was missed. Consider removing it.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -923,7 +948,7 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
       .ok (some (mkStmtExprMd (StmtExpr.Block elseStmts none)))
```
joehendrix, line 974: Not introduced by this PR, but worth noting: the else branch is translated using `bodyCtx` (context after translating the if-body) rather than the hoisted `ctx`. This means variables assigned in the if-body leak into the else translation context, which doesn't match Python runtime semantics where only one branch executes. Pre-existing from `main` (was line 922). A future cleanup could change this to use the hoisted `ctx` instead.

joehendrix (approved):
Looks good — clean refactoring and the if/else hoisting correctly mirrors the try/except pattern. Two minor non-blocking notes inline.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: Yes. Currently it is not a problem because we have move all new variables' declarations inside both block out of the `if` statement. The `variableTypes` field in `TranslationContext` now is mainly for keeping track of those declarations as all variable types now are either Any or Class (I am thinking of how to unify them). The problem may occurs when the same variable name in both block are of different Class type (or one is Any and one is Class).

andrewmwells-amazon (approved):

PR #621, opened 2026-03-19
feat(Core): add Str.ToLower and Str.ToUpper string operations

shigoel:
## Summary

- Add `str.tolower` and `str.toupper` operations to Strata Core, each with concrete evaluation and SMT axioms
- Extend the `unaryOp` combinator in `IntBoolFactory.lean` with an optional `axioms` parameter
- Skip axiom-bearing factory functions in `ExprEvalTest` to avoid MBQI timeouts on UF-based operations


Strata/Languages/Core/Factory.lean
```diff
@@ -182,6 +182,52 @@ def strSubstrFunc : WFLFunc CoreLParams :=
...
+        == ((~Str.Length : string → int) %0)],
+    -- Concat distributivity: `toUpper(s1 ++ s2) == toUpper(s1) ++ toUpper(s2)`
+    -- `str.concat` is a built-in SMT op so we elide any trigger here.
+    esM[∀ (string): -- %1 s1
+        (∀ (string): -- %0 s2
+          ((~Str.ToUpper : string → string)
+            (((~Str.Concat : string → string → string) %1) %0))
+          == (((~Str.Concat : string → string → string)
+                ((~Str.ToUpper : string → string) %1))
+                ((~Str.ToUpper : string → string) %0)))]
```
aqjune-aws, line 228: Confirmed that all these axioms and their ToLower versions are indeed valid, using CVC5.
I heard `ß` uppercases to `SS`, but probably ToUpper only deals with English alphabets.

aqjune-aws (approved):

StrataTest/Languages/Core/Tests/StrCaseTest.lean
joscoh, line 1: I think we should have a test with an unprovable or false goal to demonstrate that the axioms are not vacuous.

joehendrix (approved):
It's be nice to see if we can reduce some of the per-operator boilerplate wrt DDM integration, but that's clearly a separate PR.

Strata/Languages/Core/Factory.lean
MikaelMayer: Could you verify that idea?

StrataTest/Languages/Core/Tests/StrCaseTest.lean
```diff
@@ -0,0 +1,528 @@
...
+1. **Concrete evaluation** – `str.tolower` / `str.toupper` on literal strings
+   is constant-folded by the evaluator's `concreteEval`; the SMT solver is
+   never invoked.
+
+2. **Symbolic reasoning** – when the argument is a symbolic variable the
+   three axioms registered in `Factory.lean` are the only source of
+   knowledge for the solver:
+   - *Idempotence*: `f(f(s)) == f(s)`
+   - *Length preservation*: `len(f(s)) == len(s)`
+   - *Concat distributivity*: `f(s1 ++ s2) == f(s1) ++ f(s2)`
```
MikaelMayer, line 23: These three axioms can be trivially satisfied by the identity function. This means that having them will make "sat" no longer sound. @aqjune-aws  any insight  we could bring on that?

StrataTest/Languages/Core/Tests/StrCaseTest.lean
```diff
@@ -0,0 +1,528 @@
...
+    assert [already_lower]: str.tolower("abc") == "abc";
+
+    // Digits and punctuation are unchanged
+    assert [symbols_unchanged]: str.tolower("abc123!") == "abc123!";
+
+    // Empty string
+    assert [empty_lower]: str.tolower("") == "";
+
+    // Concrete idempotence (two calls on a literal)
+    assert [concrete_idempotent]:
```
MikaelMayer, line 55: These tests are about concrete values, so they don't make sense since we know the result of these functions. It would be more interesting to see symbolic inputs here to see that the axioms are being picked up

MikaelMayer (approved):
There seem to be redundant tests.

Strata/Languages/Core/Factory.lean
aqjune-aws: Yeah, I could test with this! :)

```
(set-logic QF_SLIA)                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                    
(declare-const result String)                                                                                                                                                                                                                       
(assert (= result (str.to_upper "\u{00DF}")))                                                                                                                                                                                                       
                                                                                                                                                                                                                                                    
(check-sat)                                                                                                                                                                                                                                         
(get-value (result))
```
It prints:
```
sat <- this doesn't matter
((result "\u{df}"))
```
`\u00df` is ß.

StrataTest/Languages/Core/Tests/StrCaseTest.lean
aqjune-aws: It is true that this might lead to spurious counter examples - but I thought it was fine to raise spurious counter examples. There was a discussion suggesting that preventing `cover` from unsat -> sat was too strict.

PR #623, opened 2026-03-19
Regex support in Python→Laurel translation

ssomayyajula:
*Description of changes:*

Add re.compile, re.match, re.search, re.fullmatch for Python->Laurel

- Laurel prelude: from_regex constructor on Any, re_Match composite type with bodiless methods.
- Core prelude: re_to_regex normalizer (str|Pattern→regex), re_fullmatch_bool/re_match_bool/re_search_bool with correct anchoring, re_compile, re_fullmatch, re_match, re_search with specs.
- PyFactory: re_compile_str with concreteEval calling pythonRegexToCore.
- ReToCore: fixed bug — added LMonoTy type annotations to all regex expression builders so concreteEval output survives the SMT encoder (which rejects unannotated ops). The old pipeline never triggered this because the old Python→Core translator panics on re.match and DiffTestCore round-trips through text which re-annotates everything.
- LaurelToCoreTranslator: map Laurel UserDefined "regex" to Core regex sort.
- StrataMain: pass ReFactory to Laurel pipeline's verifyCore.

Test: end-to-end test_regex.py covering fullmatch/match/search with literal and compiled patterns, character classes, quantifiers, alternation, dot, and anchoring differences between modes.


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -56,8 +56,10 @@ def translateType (model : SemanticModel) (ty : HighTypeMd) : LMonoTy :=
   | .TSet elementType => Core.mapTy (translateType model elementType) LMonoTy.bool
   | .TMap keyType valueType => Core.mapTy (translateType model keyType) (translateType model valueType)
   | .UserDefined name =>
-    -- Composite types map to "Composite"; datatypes map to their own name
-    match name.uniqueId.bind model.refToDef.get? with
+    -- Composite types map to "Composite"; datatypes map to their own name.
+    -- "regex" is a Core builtin sort used by the regex prelude; pass it through.
+    if name.text == "regex" then .tcons "regex" []
```
keyboardDrummer: This is incorrect. Use `.TCore "regex"` instead of a `.Userdefined "regex"` type, so you don't need this change.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -93,12 +94,105 @@ datatype DictStrAny () {
...
+// by ReFactory at verification time.
+function re_compile_str(pattern : string) : regex;
+
 type CoreOnlyDelimiter;
 
 // =====================================================================
 // Core-only declarations (not expressed in Laurel)
 // =====================================================================
 
+// /////////////////////////////////////////////////////////////////////////////////////
```
keyboardDrummer, line 119: Have you tried putting all of this in the Laurel prelude instead of the Core one?

ssomayyajula: The initial few functions may have to stay in Core because they call the SMTLIB regex functions, but the functions that call _them_ might be movable to Laurel. Let me look into that...

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
ssomayyajula: Yeah that comment is LLM nonsense; I'll just follow the convention of the other builtin Laurel types that map to Core types by DDM ops for them.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
ssomayyajula: Managed to move them to the Laurel prelude, but getting all asserts with regexes going from pass -> unknown b/c those functions are no longer marked inline. Any existing way to do that in Laurel?

keyboardDrummer: > Any existing way to do that in Laurel?

Not right now. I think I will change it so Laurel will just mark everything inline when using the Python front-end.

Strata/Languages/Python/Regex/ReToCore.lean
```diff
@@ -105,39 +105,48 @@ private def RegexAST.hasNonAnchorContent (r : RegexAST) : Bool :=
     | .complement _ => true
     | _             => false)
 
+-- Type annotations for regex ops.  concreteEval runs after the type-checker,
+-- so expressions it produces must already carry annotations for the SMT encoder.
+private def reTy : Lambda.LMonoTy := .tcons "regex" []
```
shigoel: Thanks for adding these! Nit: You can use `mty[regex]` and `mty[str → regex]`, etc. (where `mty` stands for monotype) to construct these `.tcons` terms. Note you'll need to open `LTy.Syntax`.  

See [Factory.lean](https://github.com/strata-org/Strata/blob/main/Strata/Languages/Core/Factory.lean) for some inspiration, if you like.

Strata/Languages/Python/PyFactory.lean
```diff
@@ -97,9 +97,27 @@ def reCompileFunc : LFunc Core.CoreLParams :=
...
+-- Same expansion as reCompileFunc but without flags or Except wrapping.
+def reCompileStrFunc : LFunc Core.CoreLParams :=
+    { name := "re_compile_str",
+      typeArgs := [],
+      inputs := [("pattern", mty[string])],
+      output := mty[regex],
+      concreteEval := some
+        (fun _ args => match args with
+          | [LExpr.strConst () s] =>
+            let (expr, maybe_err) := pythonRegexToCore s .fullmatch
```
shigoel: This function (also the pre-existing `reCompileFunc`) has a problem in that it hard-codes `.fullmatch` (`pythonRegexToCore` generates different outputs based on the mode). Making this parametric in the mode won't suffice: we'd need to look ahead in the program to see how the output of `re.compile` is used and provide that appropriate mode here. Too messy.

I think we need a different solution where Python's `re.compile` is essentially a no-op semantically. However, Python's `re.search`, `re.match`, and `re.fullmatch` can be implemented using such factory functions, with hard-coded match modes.

Also, could you tell me why `reCompileStrFunc` doesn't return an exception (unlike `reCompileFunc`)? `pythonRegexToCore` does model exceptions -- Python-analog `patternError` and Strata's own `unimplemented` where we can return a havoced regex. It'd be a shame to lose this info. at the Laurel level.

StrataTest/Languages/Python/tests/test_regex_positive.py
shigoel, line 1: This is a nice file to have! However, I'd like more tests with anchors esp. in the `match` and `search` modes. They'd help highlight the issue I raised earlier with the hardcoded `.fullmatch` in `pythonToRegexToCore`.

Strata/Languages/Python/Regex/ReToCore.lean
shigoel: Fixed in https://github.com/strata-org/Strata/pull/623/commits/2376558b6eb5793346ed2736b966fc9dbac41640.

Strata/Languages/Python/PyFactory.lean
shigoel: Fixed in https://github.com/strata-org/Strata/pull/623/commits/b868c65eddc8880a82c1cf9e2c1c5fda1ebee918. Exception checking added in https://github.com/strata-org/Strata/pull/623/commits/caa351a4dbcb1fdb5effd8d9d68767fe1ab07e28 and tests added in https://github.com/strata-org/Strata/pull/623/commits/c5156e468a4cfb8ae65f0439e4347b84d2021fbe

ssomayyajula: I'm understanding that the exact bug before your commits was not just that the regex compiler was always being called in fullmatch mode, but also there's no way for the prelude regex functions to be written to cope with patterns that themselves use anchors when the regex compiler is uniformly called that way. Egg on face for letting that slip by, etc.

And the reason you want `re.compile` to be identity on the nose is b/c you are no longer distinguishing between strings and regexes? I think that's fine for now, but would probably need to be changed once we start wanting to support more of the regex standard library. In particular, when we want to start supporting the instance methods on patterns.

Thanks for the exception handling and updating the ReToCore tests. Do you mind quickly explaining the necessity of `pythonRegexToCoreEraseTypes`?

Strata/Languages/Python/PythonLaurelCorePrelude.lean
ssomayyajula: Got it - is that something you can do on Monday (tomorrow)? Otherwise I could sneak that into this PR

keyboardDrummer: You can add `attr := #[.inline] -- Will be made configurable in PR 638` to 
```
  let f : Core.Function := {
    name := ⟨proc.name.text, ()⟩
    typeArgs := []
    inputs := inputs
    output := outputTy
    body := body
    preconditions := preconditions
    isRecursive := isRecursive
  }
```

But I haven't checked that this doesn't affect existing tests. It could change their results for the better, but it might be annoying to deal with those changes. Feel free to ignore this comment thread and leave your functions in the Core part of the prelude. I will then them in a later PR.

StrataTest/Languages/Python/expected_laurel/test_regex.expected
```diff
@@ -0,0 +1,494 @@
...
+assert(14768): ✅ pass (at line 459, col 4)
+set_m_calls_re_match_0: ✅ pass
+assert_assert(14856)_calls_Any_to_bool_0: ✅ pass (at line 462, col 4)
+assert(14856): ✅ pass (at line 462, col 4)
+set_m_calls_re_match_0: ✅ pass
+assert_assert(14936)_calls_Any_to_bool_0: ✅ pass (at line 465, col 4)
+assert(14936): ✅ pass (at line 465, col 4)
+set_m_calls_re_fullmatch_0: ✅ pass
+assert_assert(15379)_calls_Any_to_bool_0: ✅ pass (at line 474, col 4)
+assert(15379): ❓ unknown (at line 474, col 4)
```
MikaelMayer: I would prefer that negative tests be in a separate file from positive tests.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
ssomayyajula: Got it, thanks!

StrataTest/Languages/Python/expected_laurel/test_regex.expected
ssomayyajula: When you say positive / negative, do you mean at least one match / absence of matches OR pass / (unknown | fail)?

Strata/Languages/Python/PyFactory.lean
```diff
@@ -16,92 +16,111 @@ public section
...
         (fun _ args => match args with
-          | [LExpr.strConst () s, LExpr.intConst () 0] =>
-            -- This function has a concrete evaluation implementation only when
-            -- flags == 0.
-            -- (FIXME): We use `.match` mode below because we support only
-            -- `re.match` for now. However, `re.compile` isn't mode-specific in
-            -- general.
-            let (expr, maybe_err) := pythonRegexToCore s .match
+          | [LExpr.strConst () s] =>
+            let (_, maybe_err) := pythonRegexToCore s .fullmatch
```
MikaelMayer: The mode passed here (`.fullmatch`) is irrelevant for error detection since `pythonRegexToCore` detects parse errors in `parseTop` before the mode is consulted. Worth a brief comment to prevent a future reader from wondering whether this should vary per call site.

```suggestion
            let (_, maybe_err) := pythonRegexToCore s .fullmatch  -- mode is irrelevant: errors are detected in parseTop before mode-specific compilation
```

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
MikaelMayer: Hardcoding `"regex"` as a special case here is pragmatic but fragile — if another Core builtin sort needs the same treatment later, this will need to grow. Consider whether a set of "passthrough" sort names (even if it's a singleton today) would be clearer, or at minimum a note that this list may need extending.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -100,12 +101,103 @@ datatype DictStrAny () {
...
+}
+
+// re.compile is a no-op: returns the pattern string unchanged.
+inline function re_compile(pattern : Any) : Any
+  requires Any..isfrom_string(pattern);
+{
+  pattern
+}
+
+inline function re_fullmatch(pattern : Any, s : Any) : Any
```
MikaelMayer, line 173: The bodies of `re_fullmatch`, `re_match`, and `re_search` are identical except for the `_bool` helper they call. Since these are DDM inline functions (not Lean), extracting a shared helper parameterized by the bool function isn't straightforward. Just flagging this as a known duplication — if DDM ever gains higher-order inline functions, these three would be a natural candidate for consolidation.

Strata/Languages/Python/Regex/ReToCore.lean
```diff
@@ -105,39 +106,48 @@ private def RegexAST.hasNonAnchorContent (r : RegexAST) : Bool :=
     | .complement _ => true
     | _             => false)
 
+-- Type annotations for regex ops.  concreteEval runs after the type-checker,
+-- so expressions it produces must already carry annotations for the SMT encoder.
+private def reTy  := mty[regex]
+private def s2r   := mty[string → regex]
```
MikaelMayer, line 112: Nice set of abbreviations. The names are terse but unambiguous given the local scope (`reTy`, `s2r`, `r2r`, `rr2r`, `ss2r`, `rii2r`). The comment explaining *why* type annotations are needed (concreteEval output must survive the SMT encoder) is exactly the right level of detail.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
ssomayyajula: Wow somehow my fix to this got reverted, adding it back

Strata/Languages/Laurel/Grammar/LaurelGrammar.st
```diff
@@ -9,6 +9,7 @@ op boolType : LaurelType => "bool";
 op realType : LaurelType => "real";
 op float64Type : LaurelType => "float64";
 op stringType : LaurelType => "string";
+op regexType : LaurelType => "regex";
```
keyboardDrummer: Preference: add a more generic "coreType" to the Laurel grammar

ssomayyajula: Do we want all Core types to be directly accessible in Laurel?

PR #624, opened 2026-03-20
Fix Laurel soundness bug related to instance procedures

keyboardDrummer:
This PR does not affect the internal Python support benchmark

### Changes
1. Emit an error diagnostic when any instance procedures are used, since they are not compiled to Core at all.
2. Add grammar for instance procedures
3. Add support for fields of type string
4. Update Python so it now drops instance procedures, since Laurel no longer drops them. Without this change the Python benchmarks would fail to compile more often.
5. Update `run_py_analyze.sh` so it allows filtering the tests

### Testing
- For 1 & 2: Added a test for instance procedures
- For 3: Add test for fields of type string
- Add test for change made in a previous PR here (https://github.com/strata-org/Strata/pull/609/changes#diff-0606e416379a3a38a1be1920d6d5721ab69bfbb1f0ba8316e444399f3d877cbfR277), to test calling heap modifying procedures that return variables of different types.

### DIFF
--- a/Strata/DL/SMT/Encoder.lean
+++ b/Strata/DL/SMT/Encoder.lean
@@ -31,11 +31,11 @@ The encoding pipeline has two layers:
 2. **Encoder layer** (`EncoderM`): Sits on top of `SolverM` and manages
    A-normal form decomposition purely in the `Term` domain:
    - **Term → abbreviated Term cache** (`terms`): Maps each `Term` to its
-     abbreviated `Term.var` reference (e.g., a variable named `t0`, `t1`).
+     abbreviated `Term.var` reference (e.g., a variable named `$__t.0`, `$__t.1`).
      Large terms are broken into small `define-fun` definitions with short
      names, and the Solver handles all `Term → String` conversion.
    - **UF → abbreviated name cache** (`ufs`): Maps uninterpreted functions to
-     their abbreviated identifiers (e.g., `f0`, `f1`).
+     their abbreviated identifiers (e.g., `$__f.0`, `$__f.1`).
 
 The Encoder works purely with `Term` values. The `SolverM` layer handles all
 string conversion and caching when emitting commands.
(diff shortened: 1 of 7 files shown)


Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -175,6 +175,7 @@ def boxDestructorName (model : SemanticModel) (ty : HighType) : Identifier :=
   | .TBool => "Box..boolVal!"
   | .TFloat64 => "Box..float64Val!"
   | .TReal => "Box..realVal!"
+  | .TString => "Box..stringVal"
```
shigoel: `Box..stringVal!`? (note the `!`).

keyboardDrummer: Fixed and added missing test.

StrataTest/Languages/Boole/FeatureRequests/binary_search.lean
```diff
@@ -111,4 +111,5 @@ spec
 
 #end
 
+#guard_msgs(drop info, error) in
```
MikaelMayer: This test was passing on `main` without `#guard_msgs`. Adding `(drop info, error)` here suppresses any errors the `#eval` produces, which could mask a regression introduced by this PR. Is this change intentional, or did something in this PR break the Boole binary search verification? If it's a pre-existing flake unrelated to this PR, it might be better handled in a separate commit with an explanation.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1670,7 +1672,8 @@ def pythonToLaurel' (info : PreludeInfo)
           classFieldHighType := classFieldHighType,
           filePath := filePath
         }
-        let composite ← translateClass initCtx stmt
+        let (composite, _instanceProcedures) ← translateClass initCtx stmt
+        procedures := procedures -- TODO, compile instanceProcedure ++ _instanceProcedures
         compositeTypes := compositeTypes ++ [composite]
```
MikaelMayer, line 1678: This line is a no-op: `procedures := procedures`. The actual concatenation `++ _instanceProcedures` is inside the comment. Consider either removing the assignment entirely (since it does nothing) or writing it as just a standalone `-- TODO: compile instance procedures and append them to `procedures`` comment, so it's clear no mutation is happening here.

PR #630, opened 2026-03-20
Add collectTypeRefs for TCore

thanhnguyen-aws:
*Description of changes:* This PR adds collectTypeRefs for TCore, which fix the bug of unrecognized dependency between Box and Any.
Thanks @keyboardDrummer.

This PR replaces https://github.com/strata-org/Strata/pull/618


Strata/Languages/Laurel/DatatypeGrouping.lean
```diff
@@ -31,6 +31,7 @@ def collectTypeRefs : HighTypeMd → List String
       collectTypeRefs base ++ args.flatMap collectTypeRefs
   | ⟨.Pure base, _⟩ => collectTypeRefs base
   | ⟨.Intersection ts, _⟩ => ts.flatMap collectTypeRefs
+  | ⟨.TCore name, _⟩ => [name]
   | _ => []
```
shigoel, line 35: Out of scope for this PR (indeed, what I'm commenting on doesn't even belong to the diff), but since I feel passionately about this, I'll share: I wish we'd avoid catch-alls in our code.

PR #631, opened 2026-03-20
Denotational semantics for Laurel IR with concrete evaluator and transform preservation tests

olivier-aws:
# Denotational semantics for Laurel IR with concrete evaluator and transform preservation tests

## Summary

This PR adds a fuel-based denotational interpreter for Laurel IR, a concrete program evaluator, a comprehensive test suite (~130 tests), and transform-preservation infrastructure that validates the Laurel→Laurel lowering pipeline preserves semantics. It also fixes a bug in `liftExpressionAssignments` that broke evaluation order.

### Denotational interpreter (`Strata/Languages/Laurel/`)

- **LaurelSemantics.lean**: Shared semantic types (values, stores, heaps, outcomes) and helper functions (`evalPrimOp`, `bindParams`, `getBody`, etc.) used by the interpreter and evaluator.
- **LaurelDenote.lean**: Fuel-based denotational interpreter (`denoteStmt`, `denoteBlock`, `denoteArgs`) — three mutually recursive functions covering all `StmtExpr` constructors exhaustively. Short-circuit operators (`AndThen`, `OrElse`, `Implies`) are handled as special cases before the general `PrimitiveOp` path. `evalPrimOp` uses explicit per-operation fallthrough (no wildcard) so adding a new `Operation` variant forces a build error.
- **LaurelDenoteMono.lean**: Fuel monotonicity proof — if the interpreter succeeds with fuel `n`, it succeeds with any `m ≥ n` giving the same result.
- **LaurelConcreteEval.lean**: Bridges `denoteStmt` to `Laurel.Program` by building `ProcEnv` from static + instance procedures, constructing the initial store from static fields, and running `main`.

### Test suite (`StrataTest/Languages/Laurel/`)

- **ConcreteEval/**: 13 test modules covering primitives, arithmetic, boolean ops (including short-circuit), control flow, variables, procedures, side effects, recursion, aliasing, heap objects, type ops, verification constructs, and edge cases. Shared `TestHelper.lean` with `parseLaurel` (parse + resolve) and programmatic AST helpers.
- **LaurelDenoteTest.lean**, **LaurelDenoteUnitTest.lean**, **LaurelDenoteIntegrationTest.lean**, **LaurelDenotePropertyTest.lean**: Direct tests of the denotational interpreter using both programmatic AST and Plausible property-based testing.
- **LaurelConcreteEvalTest.lean**: End-to-end tests of the concrete evaluator.

### Transform preservation (`ConcreteEval/TransformPreservation.lean`)

Runs all 94 string-based ConcreteEval tests after the full Laurel→Laurel lowering pipeline (the exact pass sequence from `LaurelToCoreTranslator.translate`, stopping before Laurel→Core translation). **82/94 pass** — the remaining 12 failures are all due to `heapParameterization` changing the calling convention for composite types.

### Bug fix: `liftExpressionAssignments`

Fixed two bugs in the `StaticCall` case of `transformExpr`:

1. **Nested call ordering**: `add(mul(2,3), mul(4,5))` — the outer call's temp variable was declared before inner calls' temps, causing the evaluator to reference undeclared variables.

2. **Side-effect evaluation order**: `add({x:=1;x}, {x:=x+10;x})` — the lifted code didn't preserve left-to-right evaluation order. Fixed by isolating each arg's prepends, capturing side-effectful args in temporaries, and emitting prepend groups in the correct order.

### Other improvements

- `LaurelToCoreTranslator.lean`: Extracted `lowerLaurelToLaurel` function that runs all Laurel→Laurel passes, reusable by both `translate` and the test infrastructure.
- Removed `applyLift` parameter from `parseLaurel` — tests validate raw Laurel semantics directly.
- Removed unused `simp` warnings in `LaurelDenoteMono.lean`.

## Testing

- `lake build` — no warnings from new files
- `lake test` — all tests pass
- Transform preservation: 82/94 tests match direct-mode output


Strata/Languages/Laurel/LaurelDenote.lean
```diff
@@ -0,0 +1,398 @@
...
+  match h addr with
+  | some (tag, fields) =>
+      some (fun a => if a == addr then some (tag, fun f => if f == field then some v else fields f) else h a)
+  | none => none
+
+/-! ## Denotational Interpreter -/
+
+mutual
+/-- Evaluate a single statement/expression. -/
+def denoteStmt (δ : LaurelEval) (π : ProcEnv) (fuel : Nat)
```
keyboardDrummer: Seems like the store and heap would fit better in a State monad

Strata/Languages/Laurel/LaurelDenote.lean
```diff
@@ -0,0 +1,398 @@
...
+    (h : LaurelHeap) (σ : LaurelStore) (stmt : StmtExpr)
```
keyboardDrummer: Please avoid the use of single character variable names

Strata/Languages/Laurel/LaurelDenote.lean
```diff
@@ -0,0 +1,398 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Laurel.LaurelSemantics
+
+/-!
+# Fuel-Based Denotational Interpreter for Laurel IR
```
keyboardDrummer: What's denotational about this? An AI summary of "denotational semantics" is "Denotational semantics defines programming language meaning by mapping syntactic constructs directly to mathematical objects (denotations), rather than simulating execution steps", which seems opposite of what this implementation does. [The wiki](https://en.wikipedia.org/wiki/Denotational_semantics) on denotational semantics also doesn't relate to this type of semantics.

Strata/Languages/Laurel/LaurelDenote.lean
keyboardDrummer: If we want to use the semantic definition to verify that the compilation correctly maintains the mapping of source locations, then the Outcome here needs to be more detailed when errors occur, for example by containing a `DiagnosticModel`.

Strata/Languages/Laurel/LaurelDenote.lean
```diff
@@ -0,0 +1,398 @@
...
+        | none => none
+      | _ => none
+
+    | .LocalVariable name _ty none =>
+      match initStore σ name .vVoid with
+      | some σ' => some (.normal .vVoid, σ', h)
+      | none => none
+
+    -- Verification Constructs
+    -- The relational semantics requires assert/assume conditions to be pure
```
keyboardDrummer: Runtime compilation might erase assertions and assumptions, and to do that safely their bodies should not have side-effects. The semantics here should check that if the heap/store changes, `none` is returned.

Strata/Languages/Laurel/LaurelInterpreter.lean
```diff
@@ -0,0 +1,398 @@
...
+    | .PrimitiveOp .Implies [a, b] =>
+      match denoteStmt δ π fuel h σ a.val with
+      | some (.normal (.vBool false), σ₁, h₁) =>
+        some (.normal (.vBool true), σ₁, h₁)
+      | some (.normal (.vBool true), σ₁, h₁) =>
+        denoteStmt δ π fuel h₁ σ₁ b.val
+      | _ => none
+
+    -- Eager Primitive Operations
+    | .PrimitiveOp op args =>
```
keyboardDrummer, line 131: After moving the heap and store into a State monad, do notation can be used for all these cases to make them much nicer to read.

```
.PrimitiveOp op args => do
  let vals <- denoteArgs δ π fuel args
  evalPrimOp op vals
```

Strata/Languages/Laurel/LaurelDenote.lean
```diff
@@ -0,0 +1,398 @@
...
+      | _ => none
+
+    | .AsType target _ty =>
+      match denoteStmt δ π fuel h σ target.val with
+      | some (.normal v, σ₁, h₁) => some (.normal v, σ₁, h₁)
+      | _ => none
+
+    -- Quantifiers (delegated to δ)
+    | .Forall name ty body =>
+      match δ σ (.Forall name ty body) with
```
keyboardDrummer: It seems you're leaving the meaning of quantifiers unspecified. How about you enable `δ` to return a List of all possible values of a particular type? Then you can use the fuel to iterate over these values and check the property.

Strata/Languages/Laurel/LaurelInterpreter.lean
```diff
@@ -0,0 +1,398 @@
...
+      match δ σ (.Forall name ty body) with
+      | some v => some (.normal v, σ, h)
+      | none => none
+
+    | .Exists name ty body =>
+      match δ σ (.Exists name ty body) with
+      | some v => some (.normal v, σ, h)
+      | none => none
+
+    -- Specification Constructs (delegated to δ)
```
keyboardDrummer, line 353: What does delegating to `δ` mean? Shouldn't these all return a not yet implemented error?

Strata/Languages/Laurel/LaurelInterpreter.lean
```diff
@@ -0,0 +1,398 @@
...
+    | .Old val =>
```
keyboardDrummer, line 356: To implement old you'll need access to older versions of the heap, but it can be added later.

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,182 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Laurel.Laurel
```
keyboardDrummer, line 7: The separation between LaurelConcreteEval, LaurelDenote and LaurelSemantics seems arbitrary to me.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -223,24 +223,53 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
-      let seqArgs ← args.reverse.mapM transformExpr
+      -- Process arguments right to left (for substitution mechanism)
+      let seqArgs ← args.reverse.attach.mapM fun ⟨arg, _⟩ => transformExpr arg
       return ⟨.PrimitiveOp op seqArgs.reverse, md⟩
 
   | .StaticCall callee args =>
     let model := (← get).model
-    let seqArgs ← args.reverse.mapM transformExpr
-    let seqCall := ⟨.StaticCall callee seqArgs.reverse, md⟩
+    -- Process arguments right-to-left (for substitution mechanism).
```
keyboardDrummer: This seems awfully complicated. Isn't the change on line 267 and 272 sufficient? If not, can you explain why?

Strata/Languages/Laurel/LaurelDenote.lean
olivier-aws: Agreed this would be cleaner. However, the explicit state threading is currently used directly by the fuel monotonicity proof in `LaurelDenoteMono.lean`, which pattern-matches on the `(Outcome × LaurelStore × LaurelHeap)` tuples. If that's fine with you I'll file this as a follow-up refactor — we'd need to verify the monotonicity proof still works through the monad abstraction.

Strata/Languages/Laurel/LaurelDenote.lean
olivier-aws: Fixed. Renamed `h` → `heap`, `σ` → `store`.

Strata/Languages/Laurel/LaurelDenote.lean
olivier-aws: Fair point. This is a fuel-based recursive interpreter, not denotational semantics in the classical sense (mapping syntax to mathematical domains). It is however very close to how we usually define a denotational semantics, e.g. in https://www.cs.utexas.edu/~bornholt/courses/cs345h-22au/lectures/2-interpreters/, as it defines the meaning (denotation) of an AST construct as a function mapping a state to the outcome value.

For simplicity and clarity though, I renamed throughout:
- `LaurelDenote.lean` → `LaurelInterpreter.lean`
- `LaurelDenoteMono.lean` → `LaurelInterpreterMono.lean`
- All doc comments updated from "denotational" to "interpreter"
- Function names (`denoteStmt`, `denoteBlock`, `denoteArgs`) renamed to `interpStmt`, `interpBlock`, `interpArgs`
- Test files renamed accordingly

Strata/Languages/Laurel/LaurelDenote.lean
olivier-aws: Agree. This would be IMO the next step in the definition of the semantics: a more fine grained handling of the non-normal outcome. If that's fine with you, I would like to defer it to a follow-up addition. I added a TODO comment and will file an issue to track this.

Strata/Languages/Laurel/LaurelDenote.lean
olivier-aws: Agreed in principle. The interpreter already discards the post-condition store/heap and returns the originals, so side effects in assert/assume conditions are invisible to subsequent code. However, we cannot detect impure conditions at runtime because `LaurelStore` and `LaurelHeap` are function types (`String ? Option LaurelValue` and `Nat ? Option (...)`) which have no `BEq` or `DecidableEq` instance in Lean.  There is no way to compare two closures for equality. The purity requirement must be enforced statically (e.g., by a well-formedness check on the AST before interpretation) rather than dynamically. Added documentation clarifying this design constraint.

Strata/Languages/Laurel/LaurelInterpreter.lean
olivier-aws: Agreed, this would be much cleaner with do-notation. Will address together with the State monad refactor (see comment 1).

Strata/Languages/Laurel/LaurelDenote.lean
olivier-aws: Interesting idea. The current `δ` callback was designed as an escape hatch for constructs the interpreter can't handle natively. Enumerating all values of a type using fuel would work for finite types (bool, bounded int ranges) but not for unbounded ones (int, string). I could add a `typeValues : LaurelType → Option (List LaurelValue)` field to `δ` that returns `some` for enumerable types and `none` otherwise.

Happy to add this in a follow-up PR.

Strata/Languages/Laurel/LaurelInterpreter.lean
olivier-aws: Yes, exactly. Implementing `Old` would require threading a "pre-state" snapshot through the interpreter, captured at procedure entry. Added a TODO for when we need postcondition evaluation.

Strata/Languages/Laurel/LaurelSemantics.lean
olivier-aws: `LaurelSemantics` defines shared types and helpers (values, stores, heaps, `evalPrimOp`). `LaurelInterpreter` (formerly `LaurelDenote`) is the fuel-based interpreter that uses those types. `LaurelConcreteEval` bridges the interpreter to `Laurel.Program` by building `ProcEnv` from program declarations and running `main`. The split keeps the interpreter independent of the program-level structure. Added module-level doc comments making this layering explicit.

Happy to merge some of these if you feel strongly about it.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
olivier-aws: Investigated. The simple change (just lines 267/272) handles the basic case where an imperative call's temp variable needs to be declared after its arguments' temps. But it does NOT handle the evaluation-order bug for side-effectful arguments like `add({x:=1;x}, {x:=x+10;x})`. In that case we need to:

1. Isolate each argument's prepended statements (so arg2's side effects don't leak into arg1's scope)
2. Capture side-effectful results in temporaries (so the final call uses snapshot values)
3. Emit the prepend groups in left-to-right order (so side effects execute in program order)

The simple fix only addresses point (1). Points (2) and (3) require the full complexity. Added a detailed comment in the code explaining which test cases require each piece.

Strata/Languages/Laurel/LaurelDenote.lean
keyboardDrummer: > However, we cannot detect impure conditions at runtime because LaurelStore and LaurelHeap are function types

What if you use data instead of functions to represent the store and heap? Out of scope for this PR though.

Strata/Languages/Laurel/LaurelInterpreter.lean
keyboardDrummer: ok

Strata/Languages/Laurel/LaurelSemantics.lean
keyboardDrummer: I don't feel strongly but IMO the "program level structure" is as much a part of the language as any other part of the AST.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
keyboardDrummer: For this example `add({x:=1;x}, {x:=x+10;x})`, could you write down the before this PR and after this PR translation?

I thought the before would already be:
```
x := 1;
var x1 := x;
x := x + 10
add(x1, x);
```

Is something wrong with that, or was it not producing that?

Strata/Languages/Laurel/LaurelDenote.lean
keyboardDrummer: > but not for unbounded ones

Why not? If you run out of fuel then the quantifier holds, so the proof would be of the form: for whatever amount of fuel you start with, the quantifier's value does not change before and after phase P of the compilation.

Strata/Languages/Laurel/LaurelDenote.lean
keyboardDrummer: ok to leave this for a follow-up

keyboardDrummer: Seems easier to write it like you want it to be written the first way around, but happy to go with your preference here and do this in a follow-up

Strata/Languages/Laurel/LaurelSemantics.lean
olivier-aws: I agree and I have a follow-up to this PR that adds an "interpretProgram" method to define the semantics at the program level. I'll update the PR with this part

Strata/Languages/Laurel/LiftImperativeExpressions.lean
olivier-aws: With my change:

``` 
procedure main() returns 
 ()
 deterministic
 {
   var x: int := 0;
   x := 1;
   var $c_1: int;
   $c_1 := x;
   x := x + 10;
   var $c_0: int;
   $c_0 := x;
   var $c_2: void;
   $c_2 := add($c_1, $c_0);
   return $c_2
 }
```

Before my change:
```
procedure main() returns 
()
deterministic
{
 var x: int := 0;
 var $c_0: void;
 $c_0 := add(x, x);
 x := x + 10;
 x := 1;
 return $c_0
}
```

Strata/Languages/Laurel/LaurelSemantics.lean
robin-aws: I feel more strongly about this - I think there should be no distinction between the "interpreter" and the "evaluator" anywhere, they should be labeled as the same concept that could potentially be implemented for any elements of a dialect.

I would vote for using "interpreter" in this case, only because "evaluator" is slightly ambiguous in Lean  since it has `#eval.

Strata/Languages/Laurel/LaurelConcreteEval.lean
```diff
@@ -0,0 +1,122 @@
...
+      interpStmt defaultEval procEnv fuel initHeap initStore body.val
+
+/-! ## User-Friendly Result Type -/
+
+inductive EvalResult where
+  | success (value : LaurelValue) (store : LaurelStore) (heap : LaurelHeap)
+  | returned (value : Option LaurelValue) (store : LaurelStore) (heap : LaurelHeap)
+  | noMain
+  | noBody
+  | stuck (msg : String)
```
robin-aws: It seems simpler to just have the `stuck` alternative, rather than picking out `noMain` and `noBody` as more specific ways to get stuck.

robin-aws: I suppose you might be thinking that `noMain` and `noBody` are program errors, whereas `stuck` is an interpreter error? I might buy that, but in that case it would be better to encode the two categories of errors more explicitly in the result naming.

Strata/Languages/Laurel/LaurelConcreteEval.lean
```diff
@@ -0,0 +1,122 @@
...
+    | .returned (some v) _ _ => s!"returned: {v}"
+    | .returned none _ _ => "returned: void"
+    | .noMain => "error: no 'main' procedure found"
+    | .noBody => "error: 'main' has no body"
+    | .stuck msg => s!"stuck: {msg}"
+    | .fuelExhausted => "error: fuel exhausted"
+
+/-- Run a program and classify the result. Delegates to `evalProgram` for
+    the core evaluation, preserving the `noMain` / `noBody` distinction. -/
+def runProgram (prog : Program) (fuel : Nat := 10000) : EvalResult :=
```
robin-aws: Why not make the default `fuel` something like `2^64 - 1` (or whatever the right value that stops short of expensive bignumber operations)? We only have the fuel so we can still make it technically guaranteed to terminate, but in practice having the interpreter terminate by running out of fuel is never useful information. This number is probably far too low to be a reasonable "timeout" value on large examples, so I think we're better off letting the Lean process run "forever" and we can kill the process if actually necessary.

Strata/Languages/Laurel/laurel-denotational-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+for testing, debugging, and as the foundation for the concrete
+evaluator.
+
+## Motivation
+
+For testing, debugging, and downstream tooling, we need a computable
+interpreter that:
+
+1. Can be `#eval`'d on concrete programs
+2. Is deterministic by construction (it is a function, not a relation)
```
robin-aws, line 23: Apologies for the dumb question if this is known, but is Laurel itself deterministic? I had thought not, by design, in which case how is this point true?

Strata/Languages/Laurel/laurel-denotational-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Fuel Parameter
+
+The interpreter uses a `fuel : Nat` parameter decremented on every
+recursive call. This ensures termination (required by Lean for
+non-`partial` functions) without restricting the class of programs that
+can be evaluated — any terminating program can be evaluated with
+sufficient fuel.
+
+When fuel reaches zero, the interpreter returns `none` (indistinguishable
```
robin-aws, line 36: Here I got confused because `LaurelResult` has `stuck`, but I'm again getting confused between the interpreter and the evaluator. Another point that they should be the same thing IMO.

Strata/Languages/Laurel/laurel-denotational-semantics.md
robin-aws, line 1: This file should probably go under `docs` instead.

Strata/Languages/Laurel/laurel-denotational-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+If the interpreter succeeds with `fuel₁`, it succeeds with any larger
+fuel giving the same result. This is proved by mutual induction on fuel,
+case-splitting on the statement, and applying the IH to sub-calls.
+
+Analogous theorems hold for `denoteBlock` and `denoteArgs`.
+
+## Limitations
+
+1. **Fuel exhaustion is indistinguishable from stuck.** When fuel
```
robin-aws, line 116: Does not seem true of the interpreter, since `LaurelResult` DOES distinguish.

robin-aws (changes requested):
Just top level feedback so far. I'm inclined to look closer after you merge the interpreter and the evaluator, if you agree, as I believe that will cut some of the documentation and datatype complexity in half.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
keyboardDrummer: I think the above code changes are not the correct solution for the bug. I made a fix that I think is correct here: https://github.com/strata-org/Strata/pull/672

Critical question: did all the changes in LiftImperativeExpressions in this PR make sense to you? I think before submitting a PR the answer has to be yes. I looked at the lifting changes in this PR but I couldn't make sense of them that's why I investigated the bug separately.

olivier-aws: The fix you propose is not sufficient in the case of calls with imperative statements as arguments. On the same example, here is what it produces:
```
 procedure main() returns
 ()
 deterministic
 {
   var x: int := 0;
   var $c_0: void;
   $c_0 := add($x_0, x);
   var $x_1: int := x;
   x := 1;
   var $x_0: int := x;
   x := x + 10;
   return $c_0
 }
```
So the order between the call and the prepends is still wrong. The proposed change does make sense to me because I reviewed several iterations, proposed by a coding agent, including some that did more or less what you did. And neither I nor the LLM were able to generate the good ordering before doing something specific for calls as in this change.

It may very well be not the simplest possible change still, and I'm happy if we come up with something simpler, but I was not able to.

keyboardDrummer: What's the Laurel program that's failing for you on https://github.com/strata-org/Strata/pull/672?

I also tried:
```
function add(x: int, y: int): int {
  x + y
};
procedure repeatedBlockExpressions() {
  var x: int := 2;
  var y: int := { x := 1; x } + { x := x + 10; x };
  var z: int := add({ x := 1; x }, { x := x + 10; x });
  assert y == 1 + 11;
  assert z == 1 + 11
};
```

this passes verification, and returns:
```
procedure repeatedBlockExpressions () returns ()
{
  $body: {
    var x : int := 2;
    var $x_1 : int := x;
    x := 1;
    var $x_0 : int := x;
    x := x + 10;
    var y : int := $x_0 + x;
    var $x_3 : int := x;
    x := 1;
    var $x_2 : int := x;
    x := x + 10;
    var z : int := add($x_2, x);
    assert [|assert(2376)|]: y == 1 + 11;
    assert [|assert(2398)|]: z == 1 + 11;
    }
  };
```

which seems good

 > The proposed change does make sense to me because I reviewed several iterations

Sorry for questioning that

olivier-aws: I integrated the change from #672 into this PR after reverting the change on LiftImperative pass. And then the test 

```
procedure add(a: int, b: int) { return a + b };
procedure main() {
  var x: int := 0;
  return add({x := 1; x}, {x := x + 10; x})
};
```
would not produce the correct IR.
Maybe it is because of the `return` statement.

> Sorry for questioning that

Don't be, it is very sound to always look for the simplest solution, especially with LLMs that tend to find too complex solutions.

keyboardDrummer: https://github.com/strata-org/Strata/pull/672 has some issues and I'm working on resolving them. I'll update when resolved and I'll include the above test-case as well.

keyboardDrummer: #672 was still missing the fixes on lines 267/272 from this PR. With those, the program:
```
procedure add(a: int, b: int) { return a + b };
procedure main() {
  var x: int := 0;
  return add({x := 1; x}, {x := x + 10; x})
};
```
passes as well.

olivier-aws: Thanks, one #672 is merged I will update the branch and revert these changes then.

Strata/Languages/Laurel/LaurelInterpreter.lean
```diff
@@ -0,0 +1,518 @@
...
+-/
+
+import Strata.Languages.Laurel.LaurelSemantics
+
+/-!
+# Fuel-Based Interpreter for Laurel IR
+
+A computable interpreter mirroring the relational semantics in
+`LaurelSemantics.lean` (Option A from the design document
+`docs/designs/denotational-semantics-for-laurel-as-an-interprete.md`).
```
robin-aws, line 14: Wrong link (although said elsewhere I think `docs/designs` is the better location at least)

Strata/Languages/Laurel/LaurelInterpreter.lean
```diff
@@ -0,0 +1,518 @@
...
+  | none => none
+
+/-- Initialize a new variable in the store. Returns `none` if the variable already exists. -/
+def initStore (store : LaurelStore) (x : Identifier) (v : LaurelValue) : Option LaurelStore :=
+  match store x.text with
+  | none => some (fun y => if y == x.text then some v else store y)
+  | some _ => none
+
+/-- Upper bound on the address range searched by `findSmallestFree` and `allocHeap`. -/
+def heapSearchBound : Nat := 10000
```
robin-aws, line 55: Probably worth making this a top-level parameter similar to fuel, since similarly it's not actually significant to the semantics. And having a proof similar to the monotonicity of this like fuel too (but fine to cross that bridge when we come to it).

Strata/Languages/Laurel/LaurelInterpreter.lean
```diff
@@ -0,0 +1,518 @@
...
+                  | _ => none
+                | none => none
+              | none => none
+            | none => none
+          | none => none
+        | none => none
+      | _ => none
+
+    | .This =>
+      match store "this" with
```
robin-aws, line 320: This might collide with a `this` variable? Or does Laurel ensure you can't name a variable that?

This would be an excellent point to actually leverage Lean to prove some properties of interpretation, which has a good chance of forcing explicit consideration here.

Strata/Languages/Laurel/LaurelInterpreter.lean
```diff
@@ -0,0 +1,518 @@
...
+      some (fun a => if a == addr then some (tag, fun f => if f == field then some v else fields f) else heap a)
+  | none => none
+
+/-! ## Interpreter -/
+
+mutual
+/-- Evaluate a single statement/expression. -/
+def interpStmt (δ : LaurelEval) (π : ProcEnv) (fuel : Nat)
+    (heap : LaurelHeap) (store : LaurelStore) (stmt : StmtExpr)
+    : Option (Outcome × LaurelStore × LaurelHeap) :=
```
robin-aws, line 88: I think it's right for the semantics to say the result of each of these evaluation functions should just be an `Option`, because that way you aren't unintentionally specifying details of the result of programs that go wrong. But I strongly think the interpreter should use a `Result` type that allows for details of *how* the program went wrong. Otherwise debugging what caused a `none` result of an entire program is going to be a debugging nightmare. :) This would still allow for the central theorem to relate the interpreter to the semantics with `some(x) <==> success(x)`, and `none <==> any error(...)`

Strata/Languages/Laurel/LaurelInterpreter.lean
```diff
@@ -0,0 +1,518 @@
...
+    the initial store from static fields, and sets the heap to empty. -/
+def buildProgramSemantics (prog : Program) : ProgramEnv :=
+  { π := buildProcEnv prog
+    σ₀ := buildInitialStore prog
+    h₀ := fun _ => none }
+
+
+/-! ## User-Friendly Result Type -/
+
+inductive InterpResult where
```
robin-aws, line 475: As per my other comment, I think this should be be the result type of the interpreter everywhere, not just at the top level.

Strata/Languages/Laurel/laurel-denotational-semantics.md
```diff
@@ -0,0 +1,136 @@
...
+
+The interpreter uses a `fuel : Nat` parameter decremented on every
+recursive call. This ensures termination (required by Lean for
+non-`partial` functions) without restricting the class of programs that
+can be evaluated — any terminating program can be evaluated with
+sufficient fuel.
+
+When fuel reaches zero, the interpreter returns `none` (indistinguishable
+from a stuck program). This is a limitation: the interpreter cannot
+distinguish non-termination from insufficient fuel.
```
robin-aws, line 38: I think using LaurelResult throughout the interpreter will address this too, and I think it's very important to not have this limitation.

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,199 @@
...
+abbrev LaurelHeap := Nat → Option (String × (String → Option LaurelValue))
+abbrev LaurelEval := LaurelStore → StmtExpr → Option LaurelValue
+abbrev ProcEnv := Identifier → Option Procedure
+
+/-! ## Outcomes -/
+
+inductive Outcome where
+  | normal : LaurelValue → Outcome
+  | exit   : String → Outcome
+  | ret    : Option LaurelValue → Outcome
```
robin-aws, line 72: I'm not convinced we need the distinction between "normal" evaluation to a value and "ret" evaluation to a return value? If this is the best approach, it's worth mentioning why in the design doc.

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,199 @@
...
+    InitStore σ x v σ'
+
+/-! ## Heap Operations -/
+
+/-- Heap allocation using a bump-allocator (smallest-free-address) model.
+The `alloc` constructor requires `addr` to be the smallest free address:
+all addresses below `addr` must be occupied (`(h a).isSome`).
+This invariant makes allocation deterministic but precludes heap deallocation.
+If Laurel ever needs a `free` operation, this must be relaxed to a free-list
+model, which would invalidate `AllocHeap_deterministic` and downstream proofs. -/
```
robin-aws, line 98: If we don't have `free` yet, then this implementation is making each allocation linear by checking for the lowest free address, which will always be the next one, so we don't get any actual benefit. :) We might as well make this just a growing and never shrinking array for now since it will be faster, and we may not need "GC" for realistic examples for quite a while.

Strata/Languages/Laurel/LaurelSemantics.lean
```diff
@@ -0,0 +1,199 @@
...
+inductive HeapFieldWrite : LaurelHeap → Nat → String → LaurelValue → LaurelHeap → Prop where
+  | write :
+    h addr = .some (tag, fields) →
+    h' addr = .some (tag, fun f => if f == field then some v else fields f) →
+    (∀ a, addr ≠ a → h' a = h a) →
+    HeapFieldWrite h addr field v h'
+
+/-! ## Helpers -/
+
+def catchExit : Option String → Outcome → Outcome
```
robin-aws, line 121: "exit" might need a better name, I initially assumed it was like Java's `System.exit()` :) Again the design doc should talk a little bit about how it implements the semantics of control flow with the `Outcome` type (or something closer to `LaurelResult` as I asked for elsewhere)

robin-aws (changes requested):
Another batch, will fill my gaps about Laurel itself and then come back for remaining comments.

PR #632, opened 2026-03-20
feat(SMT): avoid redundant reset, globally unique quantified var names, more smtReservedKeywords

aqjune-aws:
This pull request adds a few more syntactic constraints to SMT generated code for supporting internal benchmarks.
* Removes redundant `(reset)` calls (to make it consistent, this was applied not only to Core pipeline but also B3 pipeline)
* To avoid `$__bvN` quantified variables having duplicated names across different formula (which is a valid SMT in general but more tailored for internal use case) this adds `bvCounter` to `SMT.Context`.
* Adds more predefined theory symbols to `smtReservedKeywords`.
* To avoid an unfortunate case where "tN" and "fN" may overlap with user-defined Sort, add `$__` to their prefixes.

### DIFF
--- a/Strata/DL/SMT/Encoder.lean
+++ b/Strata/DL/SMT/Encoder.lean
@@ -96,6 +96,8 @@ def smtReservedKeywords : List String :=
   -- Additional keywords not in the parser list
   parserKeywords ++
    ["true", "false", "Int", "Bool", "Real", "Array", "BitVec",
+   -- Symbols from SMT. Note: this must be synchronized with Strata's internal SMT solver which has a denylist of
+   -- names of variables/UFs/sorts.
    -- Core theory symbols
    "abs", "and", "distinct", "/", "=", ">", ">=", "ite", "=>",
    "div", "is_int", "<", "<=", "-", "mod", "*", "not", "or", "+",


Strata/DL/SMT/Encoder.lean
```diff
@@ -94,10 +94,23 @@ def smtReservedKeywords : List String :=
   -- SMT-LIB reserved words from the DDM parser
   let parserKeywords := _root_.Strata.reservedKeywords.map (·.2)
   -- Additional keywords not in the parser list
-  parserKeywords ++ ["select", "store", "and", "or", "not", "ite",
-   "true", "false", "Int", "Bool", "Real", "Array", "BitVec",
-   -- Theory function symbols that cvc5 disallows shadowing
-   "abs", "mod", "div", "to_real", "to_int", "is_int"]
+  parserKeywords ++
+   ["true", "false", "Int", "Bool", "Real", "Array", "BitVec",
+   -- Core theory symbols
```
joscoh, line 101: Is this supposed to be in sync with `Core.Factory`? If so, it does not include sequences. Also if so, it would be better to automatically determine these names via `Core.Factory.map(LFunc.name)` or similar.

atomb (approved):

PR #635, opened 2026-03-22
Irrelevant Axiom Transform

shigoel:
## Summary

Reimplements the irrelevant axiom removal feature (originally drafted in PR #524) with a cleaner architecture and proven termination of helper functions.

**Key changes:**

- **`IrrelevantAxiomsMode` enum** (`Options.lean`): Replaces the old `removeIrrelevantAxioms : Bool` with `.Off`, `.Aggressive`, and `.Precise` modes, giving callers explicit control over the tradeoff between pruning aggressiveness and soundness risk.

- **Caching** (`IrrelevantAxioms.lean`): A `Cache` struct (function call graph + axiom map) is built once per program and reused across all per-goal queries, avoiding redundant recomputation.

- **Builtin exclusion** (`CallGraph.lean`): Builtins are excluded from the function-to-axiom map because they appear in nearly every axiom, which would otherwise collapse the relevance filter.

- **Proven termination**: The fixed-point computation in `computeRelevantAxioms` now has a Lean-verified termination proof instead of relying on a fuel counter.

- **Precise mode correctness fix**: Axiom-labeled path conditions are excluded from the antecedent function extraction. Without this, axiom bodies would seed the relevant set, making everything relevant and removing nothing.

- **Cover obligations never prune**: Removing axioms from cover VCs is unsound (it weakens path conditions), so pruning is skipped for those.

### DIFF
--- a/Strata/Languages/Python/Specs/IdentifyOverloads.lean
+++ b/Strata/Languages/Python/Specs/IdentifyOverloads.lean
@@ -32,8 +32,10 @@ public structure ResolveState where
   modules  : Std.HashSet String := {}
   warnings : Array String := #[]
 
-/-- Monad for the overload-resolution walker. -/
-abbrev ResolveM := StateM ResolveState
+/-- Monad for the overload-resolution walker.
+    Reads an `OverloadTable` from the environment and accumulates
+    resolved modules and warnings in `ResolveState`. -/
+abbrev ResolveM := ReaderT OverloadTable (StateM ResolveState)
 
 /-- Record a warning about an unhandled AST node. -/
 def warn (msg : String) : ResolveM Unit :=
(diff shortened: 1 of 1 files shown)


Strata/Transform/IrrelevantAxioms.lean
```diff
@@ -0,0 +1,60 @@
...
+
+/-- Build the cache from `prog`. -/
+def Cache.build (prog : Program) : Cache :=
+  { funcCG   := prog.toFunctionCG
+    axiomMap := prog.functionImmediateAxiomMap }
+
+/-- Compute the axioms irrelevant to `functions` using precomputed `cache`.
+  Uses the transitive fixed-point algorithm (`computeRelevantAxioms`) and
+  reuses the prebuilt call graph and axiom map for efficiency. -/
+def getIrrelevantAxioms (prog : Program) (cache : Cache) (functions : List String)
```
aqjune-aws, line 47: I think a program transformation should exhibit a high-level function that is Program -> Program (or any sub-component type of AST like Stmt), and ideally have `CoreTransformM` so that monads can be composed. Would it be unnatural if we have such function of the type here? It seems `getIrrelevantAxioms` more works like a program analysis.

MikaelMayer: Added a `run : Program → List String → CoreTransformM Program` function that removes irrelevant axiom declarations from the program. It builds the cache, computes irrelevant axioms, and returns a pruned program — following the same `Program → CoreTransformM Program` pattern as `FilterProcedures.run`.

The existing `getIrrelevantAxioms` and `Cache` are kept as the analysis layer since the Verifier still uses them for per-obligation pruning (different obligations need different axiom subsets). The new `run` function provides the high-level transform entry point for cases where whole-program axiom removal is desired.

Also changed the import from `CallGraph` to `CoreTransform` (which transitively includes `CallGraph`) to get access to `CoreTransformM`.

Strata/Transform/IrrelevantAxioms.lean
```diff
@@ -0,0 +1,75 @@
...
+  allAxiomNames.filter (fun a => a ∉ relevantAxioms)
+
+open Transform in
+/-- Remove irrelevant axiom declarations from `prog` given a list of relevant
+    function names. Returns the pruned program via `CoreTransformM`. -/
+def run (prog : Program) (functions : List String) : CoreTransformM Program := do
+  let cache := Cache.build prog
+  let irrelevant := getIrrelevantAxioms prog cache functions
+  let irrelevantSet := irrelevant.toArray.qsort (· < ·)
+  let isIrrelevant (name : String) :=
```
joscoh: Not hugely important, but given that the only thing we are doing with `irrelevantSet` is looking up set membership, why not use a HashMap instead of sorting and using binary search?

StrataTest/Languages/Python/expected_non_laurel/test_datetime.expected
```diff
@@ -19,4 +19,4 @@ py_assertion: ✅ pass (at line 25, col 0)
 
 py_assertion: ✅ pass (at line 27, col 0)
 
-Assertion failed at line 30, col 0: py_assertion: ❌ fail
+py_assertion: ✅ pass (at line 30, col 0)
```
joscoh, line 22: Why does this go from failing to passing?

shigoel: Previously our irrelevant axiom detection was too imprecise and removed axioms pertinent to this proof. Now, we retain them

Strata/Transform/IrrelevantAxioms.lean
MikaelMayer: Good call — switched to `Std.HashSet.ofList` + `.contains` instead of sorted array + binary search.

StrataTest/Languages/Core/Examples/RemoveIrrelevantAxioms.lean
```diff
@@ -109,65 +109,65 @@ def normalizeModelValues (s : String) : String :=
...
-Obligation: assert_0
+Obligation: a0
 Property: assert
 Result: ✅ pass
 
-Obligation: assert_1
+Obligation: a1
 Property: assert
-Result: ❓ unknown
+Result: ❌ fail
```
aqjune-aws, line 118: Cool! When a and c were false, this must have been failed.

Strata/Transform/IrrelevantAxioms.lean
aqjune-aws: It seems the new `run` isn't used here though - but probably doing refactoring & having a dedicated "Strata/Analysis" directory could be the future work.

PR #636, opened 2026-03-22
Fix Boole old lowering for compound expressions

kondylidou:
*Issue #, if available:*

*Description of changes:*

Fixes Boole’s `old(...)` lowering in `Verify.lean` so it works on compound expressions, not just bare identifiers.

Boole already lowers `old` by mapping free variables to their pre-state names in `Core`. This PR extends that existing lowering to compound expressions needed by the new regression test `stack_array_based`, without changing Core’s representation of old-state values.


kondylidou:
@joscoh can you review this please?:)

Strata/Languages/Boole/Verify.lean
```diff
@@ -227,6 +227,17 @@ def toCoreTypedBin (m : SourceRange) (ty : Boole.Type) (op : String) (a b : Core
   let iop : Core.Expression.Expr := .op () ⟨s!"Int.{op}", ()⟩ none
   return mkCoreApp iop [a, b]
 
+private def oldNamePrefix : String := "old "
```
joscoh: I think it would be better to add a `isOldIdent` function to `Identifiers.lean` so that we don't need to remember to keep this in sync with the string in `CoreIdent.mkOld`.

kondylidou: got it thanks

Strata/Languages/Core/Identifiers.lean
```diff
@@ -34,6 +34,9 @@ def CoreIdent.toPretty (x : CoreIdent) : String := x.name
 /-- Create the `old g` identifier for a global variable named `name`. -/
 def CoreIdent.mkOld (name : String) : CoreIdent := ⟨"old " ++ name, ()⟩
 
+/-- Check whether an identifier is already an `old`-prefixed global name. -/
+def CoreIdent.isOldIdent (ident : CoreIdent) : Bool := ident.name.startsWith "old "
```
joscoh: I think `"old "` should be explicitly named e.g. as `def oldPrefix := "old "` and then used in `mkOld` and `isOldIdent`. This way strings are hard-coded only once.

PR #637, opened 2026-03-22
Fix Boole lowering bug and refresh Boole test coverage

kondylidou:
*Issue #, if available:*

*Description of changes:*

This PR fixes a Boole lowering bug where `returns ()` procedures were silently dropped because nested `return []` statements escaped `toCoreDecls` instead of producing empty local lists.

It also includes a small Boole frontend cleanup and test refresh:
    - add Boole string type lowering to Core
    - normalize a few remaining nested branch returns in `Verify.lean` from `return` to `pure`
    - update Boole regression tests whose expected obligations changed once unit-return procedures were actually lowered and verified
    - stabilize several feature-request seeds and update their comments/docs to reflect current implementation priorities
    - added a regression test for the bug fix

The net effect is that Boole procedures with no outputs are lowered correctly, the relevant regression tests are now meaningful again, and the Boole test suite is back in sync with the frontend behavior.

### DIFF
--- a/StrataTest/Languages/Boole/FeatureRequests/binary_search.lean
+++ b/StrataTest/Languages/Boole/FeatureRequests/binary_search.lean
@@ -1,3 +1,9 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
 import Strata.MetaVerifier
 
 -- Exercise 2.3-5 solution
--- a/StrataTest/Languages/Boole/FeatureRequests/string_operators.lean
+++ b/StrataTest/Languages/Boole/FeatureRequests/string_operators.lean
@@ -1,3 +1,9 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
 import Strata.MetaVerifier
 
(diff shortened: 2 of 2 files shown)


Strata/Languages/Boole/Verify.lean
```diff
@@ -629,18 +630,23 @@ private def initFVarIsOp (p : Boole.Program) : Array Bool :=
 
 def toCoreDecls (cmd : BooleDDM.Command SourceRange) : TranslateM (List Core.Decl) := do
   match cmd with
-  | .command_procedure m ⟨_, n⟩ ⟨_, targs?⟩ ins ⟨_, outs?⟩ ⟨_, spec?⟩ ⟨_, body?⟩ =>
+  | .command_procedure m nameAnn targsAnn ins outsAnn specAnn bodyAnn =>
+    let n := nameAnn.val
```
joscoh, line 634: I am confused how this is different from the previous version.

kondylidou: They’re effectively the same if you only need the wrapped values. The new version just binds the whole annotation first (nameAnn) and then reads nameAnn.val, instead of unpacking it directly in the pattern with ⟨_, n⟩. That keeps the annotation metadata available if needed.

PR #638, opened 2026-03-23
Resolve Laurel in Python pipeline

keyboardDrummer:
Note: it would be possible to extract the Laurel changes in this PR, the top 5 bullets, into a separate PR. They're also easy to review in isolation though since they only affect files in Laurel folders.

#### Laurel
- Change grammar for `if/then/else` in Laurel to match that of Core expressions (but not statements) and Lean. `{` and `}` is now only for blocks
- Add an optional `invokeOn` property to Laurel procedures, which acts like a Verus `broadcast` and creates an axiom for the procedure's postcondition
- Support (mutually)recursive Laurel functions
- Add support for constrained types to Laurel's Resolution phase
- Report Laurel resolution errors on the incoming Laurel program, before the Laurel->Laurel transformations, instead of after.

#### Python
- Move almost all definitions from PythonLaurelCorePrelude to PythonRuntimeLaurelPart
- In the Python pipeline, let all non-rec Laurel functions use the inline attribute when translating to Core
- Turn on Laurel resolution error reporting in the Python pipeline

### Caveats
- Core currently requires all recursive functions to specify a `cases` argument. Laurel will simply use the first datatype argument as the cases argument, which could be incorrect. In a future PR, I will use the decreases clauses from Laurel to determine the cases argument. Initially, we will only support a decreases clause that contains just a single datatype parameter, because of the restrictions of Core.

- Currently, Core library/factory functions need to have their signature be redefined in a Laurel prelude file, which creates some duplication. In a future PR, we should enable Laurel to automatically import these Core functions. Example of the code duplication:
```
function re_pattern_error(s: string): Error
  external;

function Str.InRegEx(s: string, r: Core regex): bool external;
```
- `PythonLaurelCorePrelude.lean` still exists, because it contains two functions that have a signature with function types, that Laurel can't express yet. I will enable and move those functions in a follow-up PR.

### Testing
- Tests updated for new if/then/else grammar
- Added a test-file for invokeOn
- Added a test file for recursive Laurel functions

### DIFF
--- a/Strata/DL/Util/List.lean
+++ b/Strata/DL/Util/List.lean
@@ -29,6 +29,23 @@ def dedup {α : Type} [DecidableEq α] : List α → List α
     if a ∈ as then as else a :: as
 
 /--
+Tail-recursive worker for `dedup`. Walks the input left-to-right,
+skipping elements that still appear later, and collects kept elements
+in reverse order.
+-/
+def dedupTR.go {α : Type} [DecidableEq α] :
+    List α → List α → List α
+  | [], acc => acc.reverse
+  | a :: as, acc =>
+    if a ∈ as then dedupTR.go as acc else dedupTR.go as (a :: acc)
+
+/--
+Tail-recursive implementation of `dedup`.
+-/
+def dedupTR {α : Type} [DecidableEq α] (l : List α) : List α :=
+  dedupTR.go l []
+
+/--
 A deduplicated list satisfies `Nodup`.
 -/
 theorem nodup_dedup {α : Type} [DecidableEq α] (l : List α) :
@@ -109,6 +126,27 @@ theorem mem_of_dedup {α : Type} [DecidableEq α]
   apply Iff.intro
   exact fun h => mem_of_mem_dedup l a h
   exact fun h => mem_dedup_of_mem l a h
+
+theorem dedupTR.go_eq {α : Type} [DecidableEq α]
+    (l acc : List α) :
+    dedupTR.go l acc = acc.reverse ++ l.dedup := by
+  induction l generalizing acc with
+  | nil => simp [dedupTR.go, dedup]
+  | cons a as ih =>
+    simp only [dedupTR.go, dedup]
+    by_cases h : a ∈ as
+    · have h' : a ∈ as.dedup := mem_of_mem_dedup as a h
+      simp [h, h', ih]
+    · have h' : a ∉ as.dedup := by
+        intro hc; exact h (mem_dedup_of_mem as a hc)
+      simp [h, h', ih]
+
+/--
+`List.dedup` is equivalent to `dedupTR` at compile time.
+-/
+@[csimp] theorem dedup_eq_dedupTR : @List.dedup = @dedupTR := by
+  funext α _ l
+  simp [dedupTR, dedupTR.go_eq]
 
 theorem length_dedup_cons_of_mem {α : Type} [DecidableEq α] (a : α) (l : List α)
   (h : a ∈ l) : (a :: l).dedup.length = l.dedup.length := by


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -528,11 +529,136 @@ def translateProcedure (proc : Procedure) : TranslateM Core.Procedure := do
   let spec : Core.Procedure.Spec := { modifies, preconditions, postconditions }
   return { header, spec, body }
 
+/--
+Collect all `StaticCall` callee names referenced anywhere in a `StmtExpr`.
+Used to determine which functions an invokeOn axiom depends on.
+-/
+partial def collectStaticCallNames (expr : StmtExpr) : List String :=
```
joscoh: Is there a reason this is `partial`?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -543,20 +627,39 @@ def translateProcedureToFunction (proc : Procedure) : TranslateM Core.Decl := do
     let checkExpr ← translateExpr precondition [] true
     return { expr := checkExpr, md := () })
 
+  -- For recursive functions, infer the @[cases] parameter index: the first input
+  -- whose type is a user-defined datatype (has constructors). This is the argument
```
joscoh, line 619: The comment should mention a TODO to use the decreases clause or something else in the future (I know it is in the PR description, but I think it should be in the code too)

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -590,7 +690,11 @@ def translate (options: LaurelTranslateOptions) (program : Program): TranslateRe
     staticProcedures := coreDefinitionsForLaurel.staticProcedures ++ program.staticProcedures
   }
 
+  -- dbg_trace "=== Initial Laurel program ==="
```
joscoh, line 689: I think these should be removed

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -650,17 +750,94 @@ def translate (options: LaurelTranslateOptions) (program : Program): TranslateRe
...
+          | some calleeIdx => g.addEdge! calleeIdx callerIdx
+          | none => g) g
+        (g', callerIdx + 1)) (Strata.OutGraph.empty n, 0) |>.1
+
+    -- Run Tarjan's SCC algorithm. Results are in reverse topological order
+    -- (a node appears after all nodes that have paths to it).
+    let sccs := Strata.OutGraph.tarjan graph
+
+    -- For each SCC, determine if it is purely functional or contains procedures.
+    -- Procedures can't call functions (only functions can call functions), so an SCC
```
joscoh: What about a function call in a procedure contract or assertion? How do you ensure this is well-formed? Unless you mean additionally that functions cannot call procedures.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -650,17 +750,94 @@ def translate (options: LaurelTranslateOptions) (program : Program): TranslateRe
...
+            | none => pure []
+            | some trigger => do
+              let axDecl? ← translateInvokeOnAxiom proc trigger
+              pure axDecl?.toList
+          let procDecl ← translateProcedure proc
+          return [Core.Decl.proc procDecl .empty] ++ axiomDecls
+
+    -- Topological sort: tarjan already gives reverse-topological order (dependencies first).
+    -- We additionally want procedure SCCs to come before function SCCs at the same level,
+    -- so that invokeOn axioms (which reference functions) are placed after the functions
```
joscoh: Why would axioms come after the functions they depend on? And I am confused why procedures would come first, considering functions cannot call procedures but procedures can call functions in Core.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -650,17 +750,94 @@ def translate (options: LaurelTranslateOptions) (program : Program): TranslateRe
...
+    -- they depend on. Since tarjan output is already dependency-first, we just stable-sort
+    -- to move procedure-containing SCCs before function-only SCCs within the same position.
+    -- The tarjan output is already correct topological order; we emit it as-is since
+    -- procedure SCCs naturally depend on function SCCs (functions are called by procedures),
+    -- so functions will already appear before procedures in the tarjan output.
```
joscoh: Why does this say functions appear before, and the previous sentence says they appear after?

StrataTest/Languages/Laurel/Examples/Fundamentals/T19_InvokeOn.lean
```diff
@@ -0,0 +1,62 @@
...
+
+procedure invokeB(x: int, y :real) {
+  assert B(y)
+//^^^^^^^^^^^ error: assertion could not be proved
+};
+
+function R(x: int): bool;
+procedure badPostcondition(x: int)
+  invokeOn R(x)
+  ensures R(x)
```
joscoh, line 60: Why does this fail?

StrataTest/Languages/Laurel/Examples/Fundamentals/T19_InvokeOn.lean
```diff
@@ -0,0 +1,62 @@
...
+open Strata
+
+namespace Strata.Laurel
+
+def program := r#"
+function P(x: int): bool;
+function Q(x: int): bool;
+
+procedure PAndQ(x: int)
+  invokeOn P(x)
```
joscoh, line 25: Can an `invokeOn` clause involve compound expressions (e.g. `P(Q(x))` or `p(x) /\ q(x)`, etc)? If so, might be good to have a test for this

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: The axiom needs to come after the procedure, so the procedure can not use it to cheat at proving its postcondition. The axiom also needs to come before functions that need it, so that's why certain procedures will be before certain functions.

StrataTest/Languages/Laurel/Examples/Fundamentals/T19_InvokeOn.lean
keyboardDrummer: Because `R(x)` is not proven to hold. R(x) only holds outside of this procedure because of the `invokeOn`

joscoh: I guess I am just confused why the previous case (e.g. `procedure PAndQ`) succeeds but this fails.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Since functions can't call procedures, once you have a function in the SCC there's no way to call a procedure any more, so SCC won't have both functions and procedures

StrataTest/Languages/Laurel/Examples/Fundamentals/T19_InvokeOn.lean
keyboardDrummer: That one does not have a body, so the body is assumed. Maybe the syntax should change so instead of no body it should say "assumed"

StrataTest/Languages/Laurel/Examples/Fundamentals/T19_InvokeOn.lean
keyboardDrummer: It supports whatever Core supports for quantifier patterns, which likely won't include anything that's interpreted like `/\`, but will include `P(Q(x))`

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: It's a confusing comment. The procedures should appear as early as possible, so after any functions that they refer to and before any other functions. I've refactored the code and added a different comment so now it should be much clearer.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Me too, but I would prefer to do that in another PR. In this PR they've just been moved from lines 643

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Removed it but to use the default decreases tactic I had to make the code a bit less idiomatic

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -574,20 +615,42 @@ def translateProcedureToFunction (proc : Procedure) : TranslateM Core.Decl := do
...
     typeArgs := []
     inputs := inputs
     output := outputTy
     body := body
     preconditions := preconditions
-  } proc.md
+    isRecursive := isRecursive
+    attr := attr
+  }
+  return .func f proc.md
```
joscoh, line 653: Just to make sure, this `.func` is temporary for recursive functions, right? (And they are added to `.recFuncBlock` instead in `translateLaurelToCore` (line 763)) I realize this could expose a bug in Core: we don't currently ensure that `.func` has `isRecursive = false` at typechecking time.

StrataTest/Languages/Laurel/Examples/Fundamentals/T19_InvokeOn.lean
joscoh: Ah ok thanks, I missed the distinction between no body and empty body.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Yes that's correct.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -139,15 +163,811 @@ function re_Match_lastindex  (self : re_Match) : int;
...
+  exception(UnimplementedError ("Pow operator is not supported"))
+};
+
+function PMod (v1: Any, v2: Any) : Any
+{
+  exception(UnimplementedError ("Mod operator is not supported"))
+};
+
+// /////////////////////////////////////////////////////////////////////////////////////
+// Modelling of Python Boolean operations And and Or
```
MikaelMayer: This section comment "Modelling of Python Boolean operations And and Or" is a duplicate of the one at line 854. The section between these two comments (lines 883-886) is empty — it looks like a leftover from the migration. Consider removing this duplicate separator.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -339,15 +326,15 @@ public def splitProcNames (prog : Core.Program)
     removed. -/
 public def translateCombinedLaurel (combined : Laurel.Program)
     : (Option Core.Program × List DiagnosticModel) :=
-  let (coreOption, errors) := Laurel.translate { emitResolutionErrors := false } combined
-  (coreOption.map prependPrelude, errors)
+  let (coreOption, errors) := Laurel.translate { inlineFunctionsWhenPossible := true } combined
```
MikaelMayer: The docstring above (lines 320-326) says "Resolution errors are suppressed" but this code uses `Laurel.translate { inlineFunctionsWhenPossible := true }` which defaults `emitResolutionErrors` to `true`. The docstring is outdated and should be updated to reflect the current behavior.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -339,15 +326,15 @@ public def splitProcNames (prog : Core.Program)
...
+  (coreOption.map appendCorePartOfRuntime, errors)
 
 /-- Like `translateCombinedLaurel` but also returns the lowered Laurel program
     (after all Laurel-to-Laurel passes, before translation to Core). -/
 public def translateCombinedLaurelWithLowered (combined : Laurel.Program)
     : (Option Core.Program × List DiagnosticModel × Laurel.Program) :=
   let (coreOption, errors, lowered) := Laurel.translateWithLaurel { emitResolutionErrors := false } combined
```
MikaelMayer: `translateCombinedLaurelWithLowered` uses `{ emitResolutionErrors := false }` (defaulting `inlineFunctionsWhenPossible` to `false`), while `translateCombinedLaurel` on line 329 uses `{ inlineFunctionsWhenPossible := true }` (defaulting `emitResolutionErrors` to `true`). These two sibling functions have inconsistent options. Is this intentional? If `translateCombinedLaurelWithLowered` is meant to behave like `translateCombinedLaurel` plus returning the lowered program, it should probably also set `inlineFunctionsWhenPossible := true`.

StrataTest/Languages/Python/expected_laurel/test_regex_positive.expected
```diff
@@ -22,429 +28,436 @@ PFloorDiv_body_calls_Int.SafeDiv_2: ✅ pass (in prelude file)
...
+ret_pos: ✅ pass (in prelude file)
+assert_name_is_foo: ✅ pass (in prelude file)
+assert_opt_name_none_or_str: ✅ pass (in prelude file)
+assert_opt_name_none_or_bar: ✅ pass (in prelude file)
+ensures_maybe_except_none: ✅ pass (in prelude file)
 set_m_calls_re_fullmatch_0: ✅ pass
 assert_assert(215)_calls_Any_to_bool_0: ✅ pass (at line 8, col 4)
-assert(215): ✅ pass (at line 8, col 4)
+assert(215): ❓ unknown (at line 8, col 4)
 set_m_calls_re_fullmatch_0: ✅ pass
```
MikaelMayer, line 6: Many assertions in this test regressed from `✅ pass` to `❓ unknown` (e.g. `assert(215)`, `assert(308)`, `assert(440)`, etc.). The PR description doesn't mention this regression. Is this an expected consequence of the pipeline changes, or a bug introduced by the refactoring?

Strata/Languages/Laurel/CoreGroupingAndOrdering.lean
```diff
@@ -0,0 +1,204 @@
...
+
+/--
+Collect all `StaticCall` callee names referenced anywhere in a `StmtExpr`.
+Used to build the call graph for SCC-based procedure ordering.
+-/
+def collectStaticCallNames (expr : StmtExprMd) : List String :=
+  match expr with
+  | WithMetadata.mk val _ =>
+  match val with
+  | .StaticCall callee args =>
```
MikaelMayer, line 82: Nit (no action needed): Well-designed separation of `collectStaticCallNames` and `computeSccDecls` — the call-graph construction is clean and the invokeOn-first partitioning strategy is a nice touch for axiom ordering.

StrataTest/Languages/Laurel/Examples/Fundamentals/T15_ShortCircuit.lean
```diff
@@ -27,11 +27,15 @@ procedure mustNotCallProc(): int
 
 procedure testAndThenFunc() {
   var b: bool := false && mustNotCallFunc(0) > 0;
+//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
+// TODO caused by a bug in Core.
```
atomb: Is this bug documented somewhere?

atomb: Kiro gives me more on this:

`AndThen`/`OrElse` are now translated to non-short-circuit `boolAndOp`/`boolOrOp`:
   ```lean
   | .AndThen => return binOp boolAndOp
   | .OrElse => return binOp boolOrOp
   ```
   The Laurel AST explicitly documents these as short-circuit operators (`"Short-circuit logical conjunction. Only evaluates the second argument if the first is true."`). The previous `ite`-based translation preserved short-circuit semantics. The new translation evaluates both operands unconditionally, which is a semantic change. The `T15_ShortCircuit.lean` test now marks several previously-passing assertions as expected failures with `// TODO caused by a bug in Core.` — this confirms the regression. If this is intentional (trading short-circuit correctness for simplicity while a Core-level fix is pending), the TODO comments should be more specific about what the Core bug is and link to a tracking issue.

Strata/Languages/Laurel/CoreGroupingAndOrdering.lean
```diff
@@ -0,0 +1,204 @@
...
+  let dtsArr := dts.toArray
+  OutGraph.tarjan g |>.toList.filterMap fun comp =>
+    let members := comp.toList.filterMap fun idx =>
+      dtsArr[idx]? |>.bind fun dt => ldtMap.get? dt.name.text
+    if members.isEmpty then none else some members
+
+end Strata.Laurel
+
+namespace Strata.Laurel
+
```
atomb: Could the previous three lines be removed?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1603,7 +1603,7 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
     for stmt in body do
       if let .FunctionDef .. := stmt then
         let proc ← translateMethod ctx className stmt
-        instanceProcedures := instanceProcedures.push proc
+        instanceProcedures := instanceProcedures.push { proc with body := .Opaque [] .none [] }
```
atomb, line 1665: Kiro says:

Instance procedure bodies are now unconditionally replaced with `.Opaque [] .none []`:
   ```lean
   ```
   This discards the translated method body. The comment at line 1615 says "Laurel does not yet support instance procedures, so treat them as if they were static", but the body erasure means any postconditions from the method are also lost. If this is intentional (methods are treated as abstract stubs), it should be documented more explicitly.

StrataTest/Languages/Laurel/Examples/Fundamentals/T15_ShortCircuit.lean
joscoh: >Is this bug documented somewhere?

Created issue https://github.com/strata-org/Strata/issues/697 (it was on my TODO list already)

Strata/Languages/Python/PythonToLaurel.lean
keyboardDrummer: Before this PR those methods were skipped entirely, not just the body. Now they're included to prevent resolution errors, but without a body. I've added a TODO now

StrataTest/Languages/Laurel/Examples/Fundamentals/T15_ShortCircuit.lean
keyboardDrummer: Linked to the issue

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -13,7 +13,7 @@ namespace Python
 
 /--
 Python prelude declarations expressed in Laurel grammar.
-Converted from PythonLaurelCorePrelude.lean (Core dialect) to Laurel dialect.
+Converted from PythonRuntimeCorePart.lean (Core dialect) to Laurel dialect.
```
MikaelMayer, line 16: The docstring says "Converted from PythonRuntimeCorePart.lean (Core dialect) to Laurel dialect" but there is no file named `PythonRuntimeCorePart.lean`. This should reference `PythonLaurelCorePrelude.lean`.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -139,15 +163,805 @@ function re_Match_lastindex  (self : re_Match) : int;
...
+  else from_none()
+};
+
+function Any_get_or_none (dict: Any, key: Any) : Any
+  requires Any..isfrom_Dict(dict) && Any..isfrom_string(key)
+{
+  DictStrAny_get_or_none(Any..as_Dict!(dict), Any..as_string!(key))
+};
+
+function DictStrAny_insert (/* */ d : DictStrAny, key: string, val: Any) : DictStrAny
```
MikaelMayer: Stray `/* */` comment in the parameter list: `DictStrAny_insert (/* */ d : DictStrAny, ...)`. Looks like a leftover from development.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -626,7 +686,11 @@ def translateWithLaurel (options: LaurelTranslateOptions) (program : Program): T
     staticProcedures := coreDefinitionsForLaurel.staticProcedures ++ program.staticProcedures
   }
 
+  -- dbg_trace "=== Initial Laurel program ==="
```
MikaelMayer, line 689: Three commented-out `dbg_trace` lines. These were moved from a different location in the function. Consider removing them entirely — they can always be re-added when debugging.

Strata/Languages/Laurel/CoreGroupingAndOrdering.lean
```diff
@@ -0,0 +1,200 @@
...
+Collect all `StaticCall` callee names referenced anywhere in a `StmtExpr`.
+Used to build the call graph for SCC-based procedure ordering.
+-/
+def collectStaticCallNames (expr : StmtExprMd) : List String :=
+  match expr with
+  | WithMetadata.mk val _ =>
+  match val with
+  | .StaticCall callee args =>
+      callee.text :: args.flatMap (fun a => collectStaticCallNames a)
+  | .PrimitiveOp _ args => args.flatMap (fun a => collectStaticCallNames a)
```
MikaelMayer, line 84: 👍 No action needed. The `collectStaticCallNames` function is thorough — it covers all `StmtExpr` variants and uses `termination_by sizeOf` for the recursive traversal. Well done.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1661,7 +1661,7 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
...
```
MikaelMayer, line 1665: The `body := .Opaque [] .none []` override discards the translated method body. The comment on line 1672 says "Laurel does not yet support instance procedures, so treat them as if they were static" — but the body is being thrown away here, not made static. A brief comment on this line explaining *why* the body is replaced with an opaque stub would help readers understand the intent.

StrataTest/Languages/Python/expected_laurel/test_missing_models.expected
```diff
@@ -2,15 +2,9 @@
...
-callElimAssert_req_name_is_foo_4: ✅ pass (at line 15, col 4)
-callElimAssert_req_opt_name_none_or_str_5: ✅ pass (at line 15, col 4)
-callElimAssert_req_opt_name_none_or_bar_6: ✅ pass (at line 15, col 4)
+Assertion failed at line 9, col 4: callElimAssert_requires_0_22: ❌ fail
+callElimAssert_requires_1_23: ✅ pass (at line 9, col 4)
+callElimAssert_requires_2_24: ✅ pass (at line 9, col 4)
 
-DETAIL: 8 passed, 1 failed, 2 inconclusive
+DETAIL: 2 passed, 1 failed, 2 inconclusive
 RESULT: Failures found
```
MikaelMayer, line 10: This expected output went from 11 verification results to 5 — the `callElimAssert` results for lines 12 and 15 are gone entirely. Is this because those call sites are no longer reachable after resolution is turned on, or is this a regression where verification obligations are silently dropped?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -559,11 +559,52 @@ def translateProcedure (proc : Procedure) : TranslateM Core.Procedure := do
...
+      : Core.Expression.Expr :=
+    match params with
+    | [] => body
+    | [p] =>
+      LExpr.allTr () p.name.text (some (translateType model p.type)) trigger body
+    | p :: rest =>
+      LExpr.all () p.name.text (some (translateType model p.type))
+        (buildQuants model rest body trigger)
+
+structure LaurelTranslateOptions where
```
MikaelMayer, line 599: Minor: `emitResolutionErrors` defaults to `true` and `inlineFunctionsWhenPossible` defaults to `false`. The only caller that sets `inlineFunctionsWhenPossible := true` is `translateCombinedLaurelWithLowered` in the Python pipeline. Consider whether this option belongs in `LaurelTranslateOptions` (which is Laurel-general) or should be a Python-pipeline-specific concern.

StrataTest/Languages/Laurel/Examples/Fundamentals/T15_ShortCircuit.lean
```diff
@@ -27,11 +27,15 @@ procedure mustNotCallProc(): int
 
 procedure testAndThenFunc() {
   var b: bool := false && mustNotCallFunc(0) > 0;
+//^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
```
MikaelMayer, line 30: Four new `error:` annotations were added to tests that were previously passing. The TODO comments reference issue #697 (a Core bug). Since these are pre-existing Core bugs now surfaced by the new resolution, the annotations are appropriate — just flagging for visibility.

StrataTest/Languages/Python/expected_laurel/test_missing_models.expected
keyboardDrummer: The except cases are not being handled correctly. Not before this PR either. There are two assertion failures in the except handlers, but they don't show up in the .expected file,.

The reason is that missing models are replaced with an assignment of hole to the LHS of the assign, but maybe_except is left unchanged, so logically the except cases never trigger. I don't understand how assertions from the except branches were showing before this PR.

If I fix the Python bug then the assertions come back, but because of an existing bug in Core that causes VCs to get duplicated in particular branches scenarios, the two failing assertions in the except branches show up twice each, so 5 failing assertions total, and some of the passing assertions also get duplicated.

What would be your preference? I suggest we leave it for a follow-up

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: I've considered

StrataTest/Languages/Python/expected_laurel/test_missing_models.expected
MikaelMayer: It seems to me that assertions from the except branches and more broadly from any place in the code should always appear in the results, whether proved, satisfiable, disproved, or unknown.

If you are confident that these assertions are still somewhere else in the output and not silently dropped, I'm good with it.

keyboardDrummer: The assertions are in the Core code, but they're unreachable due to a compilation bug in PythonToLaurel

PR #639, opened 2026-03-23
Move all variable declarations to the beginning of the function body

thanhnguyen-aws:
*Description of changes:* This PR move all variable declarations to the beginning of the function body.

This PR continues the local variable bug fix from this PR https://github.com/strata-org/Strata/pull/619. 
PR #619 moves the variable declarations inside the body blocks of `if`, `while`, and `for` statements and attach them on the top of those statements. However, in case those `if`, `while`, and `for` statements are inside a sub-block, those variables are not accessible outside of that block.
This PR resolves that issue by collecting all variables that appear inside a function body and making declarations for them at the top of the function body.

Note: To accommodate this task, this PR also adds overloaded types to compositeTypeNames so that isCompositeType recognizes them during type inference.

### DIFF
--- a/Strata/SimpleAPI.lean
+++ b/Strata/SimpleAPI.lean
@@ -290,6 +290,7 @@ def pySpecsDir (sourceDir strataDir dialectFile : System.FilePath)
     (events : Std.HashSet String := {})
     (skipNames : Array String := #[])
     (warningOutput : WarningOutput := .detail)
+    (pythonCmd : String := "python")
     : EIO String Unit := do
   -- Create output dir
   match ← IO.FS.createDirAll strataDir |>.toBaseIO with
@@ -347,7 +348,7 @@ def pySpecsDir (sourceDir strataDir dialectFile : System.FilePath)
     match ← Strata.Python.Specs.translateFile
         dialectFile strataDir pythonFile sourceDir
         (events := events) (skipNames := skipIdents)
-        (moduleName := mod) |>.toBaseIO with
+        (moduleName := mod) (pythonCmd := pythonCmd) |>.toBaseIO with
     | .error msg =>
       Python.Specs.baseLogEvent events "import" s!"Failed {mod}: {msg}"
       failures := failures.push (toString mod, msg)
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -284,7 +284,10 @@ def pyAnalyzeCommand : Command where
   args := [ "file" ]
   flags := [{ name := "verbose", help := "Enable verbose output." },
             { name := "sarif", help := "Write results as SARIF to <file>.sarif." },
-            { name := "unique-bound-names", help := "Use globally unique names for quantifier-bound variables." }]
+            { name := "unique-bound-names", help := "Use globally unique names for quantifier-bound variables." },
+            { name := "vc-directory",
+              help := "Store VCs in SMT-Lib format in <dir>.",
+              takesArg := .arg "dir" }]
   help := "Verify a Python Ion program. Translates to Core, inlines procedures, and runs SMT verification."
   callback := fun v pflags => do
     let verbose := pflags.getBool "verbose"
@@ -311,13 +314,15 @@ def pyAnalyzeCommand : Command where
         IO.print newPgm
       let solverName : String := "Strata/Languages/Python/z3_parallel.py"
       let verboseMode := VerboseMode.ofBool verbose
+      let vcDir : Option System.FilePath := pflags.getString "vc-directory" |>.map (⟨·⟩)
       let options :=
               { VerifyOptions.default with
                 stopOnFirstError := false,
                 verbose := verboseMode,
                 removeIrrelevantAxioms := .Precise,
                 solver := solverName,
-                uniqueBoundNames := uniqueBoundNames }
+                uniqueBoundNames := uniqueBoundNames,
+                vcDirectory := vcDir }
       let runVerification tempDir :=
           EIO.toIO
             (fun f => IO.Error.userError (toString f))
--- a/StrataTest/Languages/Python/AnalyzeLaurelTest.lean
+++ b/StrataTest/Languages/Python/AnalyzeLaurelTest.lean
@@ -29,13 +29,14 @@ private meta def testDir : System.FilePath :=
 /-- Compile a Python source file to a `.python.st.ion` Ion file.
     Returns the path to the generated Ion file. -/
 private meta def compilePython
+    (pythonCmd : System.FilePath)
     (dialectFile : System.FilePath) (pyFile : System.FilePath)
     (outDir : System.FilePath) : IO System.FilePath := do
   let some stem := pyFile.fileStem
     | throw <| .userError s!"No stem for {pyFile}"
   let ionPath := outDir / s!"{stem}.python.st.ion"
   let spawnArgs : IO.Process.SpawnArgs := {
-    cmd := "python"
+    cmd := toString pythonCmd
     args := #["-m", "strata.gen", "py_to_strata",
               "--dialect", dialectFile.toString,
               pyFile.toString, ionPath.toString]
@@ -55,29 +56,32 @@ private meta def compilePython
 
 /-- Set up the test fixture: compile all servicelib modules and return the
     spec directory.  The dispatch and pyspec modules are resolved by name. -/
-private meta def setupFixture (_pythonCmd : System.FilePath)
+private meta def setupFixture (pythonCmd : System.FilePath)
     (outDir : System.FilePath) : IO Unit := do
   IO.FS.withTempFile fun _handle dialectFile => do
     IO.FS.writeBinFile dialectFile Python.Python.toIon
     -- Compile all servicelib modules (dispatch + individual services)
     match ← pySpecsDir testDir outDir dialectFile
         (modules := #["servicelib", "servicelib.Storage", "servicelib.Messaging"])
-        (warningOutput := .none) |>.toBaseIO with
+        (warningOutput := .none)
+        (pythonCmd := toString pythonCmd) |>.toBaseIO with
     | .ok () => pure ()
     | .error msg => throw <| IO.userError s!"pySpecsDir failed: {msg}"
 
 /-- Compile a test Python file to Ion format. -/
-private meta def compileTestScript (pyFile : System.FilePath)
+private meta def compileTestScript (pythonCmd : System.FilePath)
+    (pyFile : System.FilePath)
     (outDir : System.FilePath) : IO System.FilePath := do
   IO.FS.withTempFile fun _handle dialectFile => do
     IO.FS.writeBinFile dialectFile Python.Python.toIon
-    compilePython dialectFile pyFile outDir
+    compilePython pythonCmd dialectFile pyFile outDir
 
 /-- Run pyAnalyzeLaurel on a test script within the shared fixture. -/
 private meta def runAnalyze
+    (pythonCmd : System.FilePath)
     (tmpDir : System.FilePath) (scriptName : String)
     : IO (Except String Core.Program) := do
-  let testIon ← compileTestScript (testDir / scriptName) tmpDir
+  let testIon ← compileTestScript pythonCmd (testDir / scriptName) tmpDir
   let laurel ←
     match ← Strata.pyAnalyzeLaurel testIon.toString
         (dispatchModules := #["servicelib"])
@@ -94,9 +98,10 @@ private meta def runAnalyze
 
 /-- Run pyAnalyzeLaurel with inlining and verification, returning the formatted results. -/
 private meta def runAnalyzeAndVerify
+    (pythonCmd : System.FilePath)
     (tmpDir : System.FilePath) (scriptName : String)
     : IO (Except String (Array Core.VCResult)) := do
-  let testIon ← compileTestScript (testDir / scriptName) tmpDir
+  let testIon ← compileTestScript pythonCmd (testDir / scriptName) tmpDir
   let laurel ←
     match ← Strata.pyAnalyzeLaurel testIon.toString
         (dispatchModules := #["servicelib"])
@@ -186,9 +191,9 @@ private meta def testCases : List (String × Expected) := [
 ]
 
 /-- Run a single test case and return an error message on failure, or `none` on success. -/
-private meta def runTestCase (tmpDir : System.FilePath)
+private meta def runTestCase (pythonCmd : System.FilePath) (tmpDir : System.FilePath)
     (scriptName : String) (expected : Expected) : IO (Option String) := do
-  let result ← runAnalyze tmpDir scriptName
+  let result ← runAnalyze pythonCmd tmpDir scriptName
   match expected, result with
   | .success, .ok _ => return none
   | .success, .error msg =>
@@ -204,9 +209,9 @@ private meta def runTestCase (tmpDir : System.FilePath)
     if msg.startsWith pfx then return none
     else return some s!"{scriptName}: Expected error starting with:\n  {pfx}\nGot:\n  {msg}"
 
-#eval withPython fun _pythonCmd => do
+#eval withPython fun pythonCmd => do
   IO.FS.withTempDir fun tmpDir => do
-    setupFixture _pythonCmd tmpDir
+    setupFixture pythonCmd tmpDir
     -- Launch all tests concurrently, checking for duplicate filenames
     let mut seen : Std.HashSet String := {}
     let mut tasks : Array (String × Task (Except IO.Error (Option String))) := #[]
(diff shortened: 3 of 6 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -924,12 +924,57 @@ partial def translateAssign  (ctx : TranslationContext)
...
-    | .Assign _ targets _ _ => targets.val.toList.head?.map (fun t => (pyExprToString t, PyLauType.Any))
-    | _ => none
+
+partial def getWithItemsVars (withItems: List (Python.withitem SourceRange))
+    :List (String × String) :=
+  withItems.filterMap fun item => match item with
+    | .mk_withitem _ _ var =>
+      match var.val with
+        | some var => some (pyExprToString var, PyLauType.Any)
+        | _ => none
```
joehendrix, line 1024: I'm always suspicious when I see a filterMap that there's a risk some error is being hidden.  Can you add a docstring to comment that helps me understand why this is ok?

joehendrix:
I can approve if this is blocking, but I think some additional comments on the filterMap (or error checks) would help convince me this isn't going to hide errors.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -960,12 +960,63 @@ partial def translateAssign  (ctx : TranslationContext)
...
+/-- Extracts loop variables from a `for` loop target expression.
+    Handles single var or tuple/list unpacking targets. -/
+partial def getForLoopVars (targetIter: Python.expr SourceRange) :List (String × String) :=
+  match targetIter with
+    | .Name _ n _ => [(n.val, PyLauType.Any)]  -- `for x in ...`
+    | .Tuple _ tup _ | .List _ tup _ =>
+        tup.val.toList.flatMap fun n => getForLoopVars n -- `for x, y in ...` or `for [x, y] in ...`
+    | _ => []
+
+partial def collectDeclaredNamesAndTypes (ctx : TranslationContext) (stmts : List (Python.stmt SourceRange)) : List (String × String) :=
```
joehendrix, line 1042: I'm concerned about the defaulting to just return an empty list on AST nodes not understood, and relying on translateExpr in the body.

We could merge and fixup errors, but it'd be nice to have an approach that computed this purely from the Python AST, and tried to cover cases completely.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -960,12 +960,63 @@ partial def translateAssign  (ctx : TranslationContext)
...
+      let ty := match value.val with
+        | some value =>
+          match translateExpr ctx value with
+          | .ok {val := .New classname, ..}  => classname.text
+          | .ok {val := .StaticCall funcname _ , ..} =>
+            if isCompositeType ctx funcname.text then funcname.text else PyLauType.Any
+          | _ => pyExprToString annoTy
+        | _ => pyExprToString annoTy
+      [(pyExprToString lhs, ty)]
+    | .If _ _ body elsebody => body.val.toList.flatMap go ++ elsebody.val.toList.flatMap go
```
shigoel, line 1054: This collects AnnAssign targets even when there is no initializer value (the `| _ => pyExprToString annoTy` branch). That causes `createVarDeclStmtsAndCtx` to emit a hoisted `LocalVariable x AnyTy Hole` at the function top. But the existing `AnnAssign none` handler in `translateStmt` (around line 1007) unconditionally emits a second `LocalVariable x intTy AnyNone` without checking `ctx.variableTypes` first. The old code only had this duplication for bare annotations inside compound-statement bodies; with function-level hoisting it now happens for every bare annotation at any nesting depth. The `AnnAssign none` handler should check whether the variable is already in `ctx.variableTypes` and emit an `Assign` instead of a `LocalVariable` when it is.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -960,12 +960,63 @@ partial def translateAssign  (ctx : TranslationContext)
...
+    | .For _ targetIter _ body _ _ => getForLoopVars targetIter ++ (body.val.toList.flatMap go)
+    | .While _ _ body _ => body.val.toList.flatMap go
+    | .Try _ body handlers _orelse finalbody =>
+        let handlers_bodies := handlers.val.toList.map (λ h => match h with
+          | .ExceptHandler _ _ _ body => body.val.toList)
+        let error_var := (handlers.val.toList.filterMap (λ h => match h with
+          | .ExceptHandler _ _ errname _ => errname.val)).map (λ h => (h.val, "PythonError"))
+        error_var ++ (body.val.toList.flatMap go) ++ (handlers_bodies.flatten.flatMap go) ++ (finalbody.val.toList.flatMap go)
+    | .With _ items body _ => (getWithItemsVars items.val.toList) ++ (body.val.toList.flatMap go)
```
shigoel: This line traverses `finalbody` (the `finally` clause) but `_orelse` (the `else` clause) is skipped. Since `translateStmt` for `Try` (pattern `| .Try _ body handlers _ _ =>`) does not translate either clause, the two should be treated consistently: variables from `finally` blocks get phantom declarations (hoisted but never assigned) while variables from `else` blocks are silently dropped. Either traverse both or neither.

thanhnguyen-aws: I added the variables collecting for `orelse` body.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I added the check (line 1068).

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I add more case handlings. I think we almost cover all the places where variable declaration may occur.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -982,32 +982,90 @@ partial def translateAssign  (ctx : TranslationContext)
...
-        else AnyTy
-      mkStmtExprMd (StmtExpr.LocalVariable (name : String) ty (some (mkStmtExprMd .Hole)))
+  let hoistedDecls : List StmtExprMd ←  newDecls.mapM fun (name, tyStr) => do
+      if isCompositeType ctx tyStr then
+        let ty := mkHighTypeMd (.UserDefined tyStr)
+        pure $ mkStmtExprMd (StmtExpr.LocalVariable (name : String) ty (some (mkStmtExprMd .Hole)))
+      else if isOfAnyType tyStr then
+        pure $ mkStmtExprMd (StmtExpr.LocalVariable (name : String) AnyTy (some (mkStmtExprMd .Hole)))
+      else
+        throw (.userPythonError .none s!"'{tyStr}' is not a type")
```
MikaelMayer: This is a behavioral regression compared to the old code. Previously, `createVarDeclStmtsAndCtx` would fall back to `AnyTy` for any type that wasn't composite or `PythonError`. Now it throws an error for types not in the `isOfAnyType` hardcoded list.

For example, if `pyExprToString annoTy` returns `"Optional[int]"`, `"List[str]"`, `"Tuple[int, str]"`, or any other Python type annotation string, this will throw `'...' is not a type` instead of gracefully defaulting to `AnyTy`.

Note that `translateType` itself already handles unknown types by falling back to `AnyTy` (line ~247). Consider replacing the `isOfAnyType` check + error with a simple `else` branch that uses `AnyTy`, matching the old behavior:
```lean
      else
        pure $ mkStmtExprMd (StmtExpr.LocalVariable (name : String) AnyTy (some (mkStmtExprMd .Hole)))
```
If the intent is to catch genuinely invalid types, this should be validated against the test suite with Python files that use complex type annotations.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -982,32 +982,90 @@ partial def translateAssign  (ctx : TranslationContext)
...
-  stmts.filterMap fun s => match s with
-    | .AnnAssign _ target annotation _ _ =>
-      some (pyExprToString target, pyExprToString annotation)
-    | .Assign _ targets _ _ => targets.val.toList.head?.map (fun t => (pyExprToString t, PyLauType.Any))
-    | _ => none
+def isCompositeType (ctx : TranslationContext) (ty: String) : Bool := match ctx.importedSymbols[ty]? with
+  | some (ImportedSymbol.compositeType _) => true
+  | _ => false
+
+def isOfAnyType (ty: String): Bool := ty ∈ ["None", "bool", "int", "str", "float", "datetime", "ListAny", "DictStrAny", "Any"]
```
MikaelMayer: This list is fragile — it must be kept in sync with any new primitive types added to the system. If the suggestion above (removing the error and always falling back to `AnyTy` for non-composite types) is adopted, this function becomes unnecessary. Otherwise, consider deriving this list from the existing `PyLauType` definitions or `translateType` logic to avoid drift.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -982,32 +982,90 @@ partial def translateAssign  (ctx : TranslationContext)
...
+    | .Name _ n _ => [(n.val, PyLauType.Any)]  -- `for x in ...`
+    | .Tuple _ tup _ | .List _ tup _ =>
+        tup.val.toList.flatMap fun n => getForLoopVars n -- `for x, y in ...` or `for [x, y] in ...`
+    | _ => []
+
+partial def collectDeclaredNamesAndTypes (ctx : TranslationContext) (stmts : List (Python.stmt SourceRange)) : List (String × String) :=
+  let rec go (s : Python.stmt SourceRange) : List (String × String) :=
+    match s with
+    | .Assign _ lhs value _ =>
+      let ty := match translateExpr ctx value with
```
MikaelMayer: The type-inference-from-`translateExpr` pattern (matching `.New` and `.StaticCall`) is duplicated between the `.Assign` branch (lines 1015–1019) and the `.AnnAssign` branch (lines 1025–1029). Consider extracting a helper like:
```lean
def inferTypeFromExpr (ctx : TranslationContext) (value : Python.expr SourceRange) : String :=
  match translateExpr ctx value with
  | .ok {val := .New classname, ..} => classname.text
  | .ok {val := .StaticCall funcname _, ..} =>
      if isCompositeType ctx funcname.text then funcname.text else PyLauType.Any
  | _ => PyLauType.Any
```
The `.AnnAssign` branch would then only need to add its `.Hole` and annotation-fallback cases around this helper.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1321,6 +1359,16 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
     }
   | _ => throw (.internalError "Expected FunctionDef")
 
+def translateFunctionBody (ctx : TranslationContext) (inputTypes : List (String × String)) (body: List (Python.stmt SourceRange))
```
MikaelMayer, line 1382: This function could use a docstring — it's the central orchestrator for translating a function body (collecting declarations, hoisting them, prepending exception handling). Something brief like:
```lean
/-- Translate a Python function body: collect all variable declarations, hoist them
    to the top, and translate the remaining statements. -/
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1321,6 +1359,16 @@ def pyFuncDefToPythonFunctionDecl  (ctx : TranslationContext) (f : Python.stmt S
...
+  : Except TranslationError (StmtExprMd × TranslationContext) := do
+    let ctx := {ctx with variableTypes:= ("nullcall_ret", PyLauType.Any)::inputTypes}
```
MikaelMayer, line 1384: The caller in `pythonToLaurel'` (line ~1841) sets `ctx.variableTypes := [("nullcall_ret", PyLauType.Any)]` right before calling `translateFunctionBody ctx [] ...`, but `translateFunctionBody` immediately overwrites `variableTypes` with `("nullcall_ret", PyLauType.Any)::inputTypes`. The caller's assignment is redundant for that code path. Not a bug, but worth noting for clarity — the function is self-contained in setting up `variableTypes`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1726,9 +1770,17 @@ def pythonToLaurel' (info : PreludeInfo)
     instanceProcedures := []
   }
 
+    let overloadCompositeType := Std.HashSet.ofList $
```
MikaelMayer: This block adds overloaded types to `compositeTypeNames` so that `isCompositeType` recognizes them during type inference. While it's needed for the new `collectDeclaredNamesAndTypes` logic, it's a separate concern from hoisting variable declarations. Consider noting this dependency in the PR description, or splitting it out if it can stand alone.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I removed the redundant code at (line ~1841) .

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I updated the PR description.

aqjune-aws (dismissed):
Indeed, without this patch I confirmed that this error appears:
```
> DETAIL: test_variable_in_nested_block.py(91-114) ❌ Type checking error.
> [assert [assert_assert(91)_calls_Any_to_bool_0] ((~Bool.Or : (arrow bool (arrow bool bool)))
>  ((~Bool.Or : (arrow bool (arrow bool bool)))
>   ((~Bool.Or : (arrow bool (arrow bool bool)))
>    ((~Bool.Or : (arrow bool (arrow bool bool)))
>     ((~Bool.Or : (arrow bool (arrow bool bool)))
>      ((~Any..isfrom_bool : (arrow Any bool)) (~PEq (label : Composite) (~from_string #large)))
>      ((~Any..isfrom_none : (arrow Any bool)) (~PEq (label : Composite) (~from_string #large))))
>     ((~Any..isfrom_string : (arrow Any bool)) (~PEq (label : Composite) (~from_string #large))))
>    ((~Any..isfrom_int : (arrow Any bool)) (~PEq (label : Composite) (~from_string #large))))
>   ((~Any..isfrom_Dict : (arrow Any bool)) (~PEq (label : Composite) (~from_string #large))))
>  ((~Any..isfrom_ListAny : (arrow Any bool))
>   (~PEq (label : Composite) (~from_string #large))))] No free variables are allowed here!
> Free Variables: [label, label, label, label, label, label]
> RESULT: Internal error
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -230,22 +230,13 @@ def isKnownType (ctx : TranslationContext) (typeStr : String) : Bool :=
...
-  | _ =>
-    -- Check if it's a Python type that maps to Core
-    match pythonTypeToCoreType typeStr with
-    | some coreType => .ok (mkCoreType coreType)
-    | none =>
-      -- Check if it matches a known composite type (user-defined or PySpec)
-      match ctx.importedSymbols[typeStr]? with
-      | some (ImportedSymbol.compositeType laurelName) =>
+  -- Check if it matches a known composite type (user-defined or PySpec)
+  match ctx.importedSymbols[typeStr]? with
```
shigoel, line 234: The PR removes the `"int"`, `"bool"`, `"str"` match arms and the `pythonTypeToCoreType` lookup from `translateType`, replacing the whole body with an `importedSymbols`/`preludeTypes` check. For `createVarDeclStmtsAndCtx` this is harmless because the old code already used `AnyTy` for non-composite types anyway. But `translateType` is also called in the `AnnAssign` handler (`let varType ← translateType ctx varType`) and in `translateFunction` for each parameter type annotation. Both sites now receive `AnyTy` for `"int"`, `"bool"`, `"str"`, and Core-mapped types like `"datetime"` that were previously handled by `pythonTypeToCoreType`. Function parameters annotated `int` or `bool` and bare annotations like `x: int` now generate `AnyTy` declarations instead of `TInt`/`TBool`/`TString`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1225,31 +1242,27 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+            setupStmts := setupStmts ++ [mgrDecl, assignStmt]
           else
             -- New variable — declare outside the block so it's visible after
             let varDecl := mkStmtExprMd (StmtExpr.LocalVariable varName AnyTy (some enterCall))
             currentCtx := {currentCtx with variableTypes := currentCtx.variableTypes ++ [(varName, PyLauType.Any)]}
-            declStmts := declStmts ++ [mgrDecl, varDecl]
+            setupStmts := setupStmts ++ [mgrDecl, varDecl]
         | none =>
-          declStmts := declStmts ++ [mgrDecl]
           setupStmts := setupStmts ++ [enterCall]
```
shigoel: In the `| none =>` branch (a `with` item that has no `as` clause, e.g. `with lock:`), the PR removes `declStmts := declStmts ++ [mgrDecl]` but does not add `mgrDecl` to `setupStmts`. `mgrDecl` is the `mgr_N = <expr>` assignment that binds the context-manager expression to `mgrName`. `enterCall` is built as `mkInstanceMethodCall mgrTy "__enter__" mgrRef []` where `mgrRef = Identifier mgrName` (base file line 1222), so it can only execute correctly if `mgrDecl` has run first. In both `| some varExpr =>` branches above, `mgrDecl` was correctly moved into `setupStmts`; the `| none =>` case needs the same fix: `setupStmts := setupStmts ++ [mgrDecl, enterCall]`.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: All variables of "actual" types of `int`, `bool`, `str`, `datetime` are translated into Any, so the output Core program remain the same. For AnnAssign the variable types that is added to `TranslationContext` (`variableTypes` list) is obtained by `pyExprToString annotation` (line 982 and 1105), not `translateType`, so we won't loose those information in the `TranslationContext`

PR #641, opened 2026-03-24
Sound equality checking for Lambda

joscoh:
*Issue #, if available:*

*Description of changes:* Previously, Lambda's equality checking was unsound (when compared against an in-progress denotational semantics). This was partially fixed in https://github.com/strata-org/Strata/pull/633, but it involved an awkward mechanism that prevented equality checking under lambdas. It was also still unsound, since we cannot erase types when comparing equality (e.g. `@Nil int` is not semantically equal to `@Nil bool`).

This PR adds a more principled equality check that is sound but incomplete, giving up if it cannot prove equality of inequality. It tries syntactic equality (which can prove equality, not inequality), reasons about constants, proves equality and inequality of lambdas extensionally, and reasons about the (in)equality of datatype constructors by using injectivity+disjointness properties. Tests demonstrating its behavior are in `LExprEqTests.lean`.

### DIFF
--- a/Strata/DL/Lambda/LExprEval.lean
+++ b/Strata/DL/Lambda/LExprEval.lean
@@ -55,7 +55,7 @@ Returns `some true` if provably equal, `some false` if provably not equal,
 
 This test is relatively conservative:
 - Terms are equal if syntactically equal
-- Syntactically different constants are not equal
+- Syntactically different, non-real constants are not equal
 - Closed lambdas of the same type are compared extensionally (i.e.
 syntactically after substituting a fresh variable for the body). Note that this
 does not evaluate the body, which may not be a canonical value.


Strata/DL/Lambda/LExprEval.lean
```diff
@@ -35,14 +35,81 @@ inductive EvalProvenance
...
 /--
-Canonical values of `LExpr`s.
+Semantic equality check for two expressions `e1` and `e2`.
+
+Returns `some true` if provably equal, `some false` if provably not equal,
+`none` if inconclusive.
+
+This test is relatively conservative:
+- Terms are equal if syntactically equal
+- Syntactically different constants are not equal
```
andrewmwells-amazon: I checked that `Rat` reduces fractions, but this still feels dangerous. If we do this, can we document this requirement on `LConst`?

joscoh: Good point. I think it might be better to just make real constants incomparable (this would also help if we ever interpret them correctly instead of as rationals).

Strata/DL/Lambda/LExprEval.lean
```diff
@@ -35,14 +35,84 @@ inductive EvalProvenance
...
```
andrewmwells-amazon:
```suggestion
- Syntactically different, non-real constants are not equal
```

Strata/DL/Lambda/LExprEval.lean
```diff
@@ -35,14 +35,84 @@ inductive EvalProvenance
...
+- Syntactically different, non-real constants are not equal
+- Closed lambdas of the same type are compared extensionally (i.e.
+syntactically after substituting a fresh variable for the body). Note that this
+does not evaluate the body, which may not be a canonical value.
+- Datatype constructor applications are compared using known injectivity
+and disjointness properties of datatypes.
+-/
+def eql (F : @Factory T.base) (e1 e2 : LExpr T) : Option Bool :=
+  -- If syntactic equality holds, so does semantic equality
+  if eqModuloMeta e1 e2 then some true
```
aqjune-aws, line 67: During a separate discussion on the slack channel, I became to have a curiosity - does this fast-forwarded path actually still allow `Nil` without any type annotation to still compare equal? If the two `Nil`s are annotated with different types, does it imply that type annotation may change evaluation result?

joscoh: Yes, two `Nil` without type annotations are considered equal and yes, if they are annotated with different types, the evaluation result may change. I think this is fine. Our semantics are defined only after type inference, when the annotation is uniquely determined. Essentially, we make no guarantees about the behavior of the evaluator (except that it is consistent with `Step`) on untyped terms.

PR #642, opened 2026-03-24
fix(laurel): local variable no longer shadows output parameter

### LINKED ISSUE #567: Variable shadowing in Laurel -> Strata Core translation
It seems lowering the `return` statement can sometimes result in assignment to a shadowed variable. In the attached code, the Laurel code is lowered a Core procedure like this:
```
procedure double (x : int) returns (result : int)
{
  $body: {
    var result : int := x + x;
    result := result;
    exit $body;
    }
  };
```

https://gist.github.com/aqjune-aws/744847acf57f62b576521de2441a97f7

fabiomadge:
Fixes #567.

## Problem

When a Laurel procedure declares a local variable with the same name as an output parameter (e.g. `var result: int := x + x` in a procedure with `returns (result: int)`), the generated Core emits `init result` inside the `$body` block, which shadows the output parameter. The `return` statement then assigns the local to itself, and the output parameter is never written.

## Fix

In `translateStmt` for `LocalVariable`: when the variable name matches an output parameter, emit `set`/`havoc` instead of `init`, since the variable already exists in Core as the output parameter.

- `init name type (some expr)` → `set name expr`
- `init name type none` → `havoc name`
- `init` before `call` → dropped (the `call` writes to the existing variable directly)

## Before

```
procedure double (x : int) returns (result : int)
{
  $body: {
    var result : int := x + x;   // shadows output param
    result := result;            // self-assignment on local
    exit $body;
  }
};
```

## After

```
procedure double (x : int) returns (result : int)
{
  $body: {
    result := x + x;             // assigns to output param
    result := result;            // harmless self-assignment
    exit $body;
  }
};
```

## Tests

New test `T18_LocalVarShadowsOutput` covers five cases:
1. Expression initializer with auto-generated `result` name
2. Expression initializer with user-named output param
3. No initializer (havoc path)
4. Procedure call as initializer
5. Output param used as scratch, then overwritten by return

### DIFF
--- a/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
+++ b/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
@@ -358,33 +358,45 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
       let coreMonoType := translateType model ty
       let coreType := LTy.forAll [] coreMonoType
       let ident := ⟨id.text, ()⟩
+      -- When a local variable has the same name as an output parameter, the
+      -- variable already exists in Core.  Emit `set`/`havoc` instead of `init`
+      -- to avoid shadowing the output parameter (see issue #567).
+      let isOutputParam := outputParams.any (fun p => p.name.text == id.text)
+      let intro (e : Option Core.Expression.Expr) : List Core.Statement :=
+        if isOutputParam then
+          match e with
+          | some v => [Core.Statement.set ident v md]
+          | none   => [Core.Statement.havoc ident md]
+        else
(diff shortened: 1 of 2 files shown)


keyboardDrummer (changes requested):
This PR changes Laurel in a way that I think is confusing.

What happens if I do:
```
procedure double(x: int) returns (result: int)
  ensures result == 1
{
  result := 1;
  var result: int := 23;
  return;
};
```
Would that fail?

Instead of what this PR does, I'd rather add a feature to Laurel that enables anonymous return variables, so we can replace the code that creates a "result" return variable.

We could also consider adding an option that disallows shadowing of return parameters.

PR #645, opened 2026-03-24
Change pythonToLaurel and pythonToCore to take Array stmt

joehendrix:
Accept an Array of Python statements instead of a Command/Program wrapper so callers control how modules are unwrapped. This removes the hidden assumption that translation always starts from a single Module command and makes the API more composable for upcoming pipeline changes.

- `pythonToLaurel'` and `pythonToLaurel` now take `Array (stmt SourceRange)` instead of `Python.Command SourceRange`, eliminating the internal `match` on `Module` and the dead `| _ => throw` branch.
- `pythonToCore` now takes `Array (stmt SourceRange)` instead of `Strata.Program`; the `toPyCommands` and `unwrapModule` helpers are removed from `PythonToCore.lean`.
- `signaturesToLaurel` `modulePrefix` parameter is now required (no default) so callers explicitly choose the right prefix rather than silently getting `""`.
- Unwrapping logic moves to call sites (`StrataMain`, `PySpecPipeline`, `IdentifyOverloadsTest`) where the module structure context is clear.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -486,6 +486,12 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
       let first ← translateExpr ctx values.val[0]!
       return first
 
+  | .IfExp _ cond thenb elseb =>
+    let condExpr ← translateExpr ctx cond
+    let thenExpr ← translateExpr ctx thenb
+    let elseExpr ← translateExpr ctx elseb
+    return mkStmtExprMd (StmtExpr.IfThenElse (Any_to_bool condExpr) thenExpr elseExpr)
+
   | .Call _ f args kwargs => translateCall ctx f args.val.toList kwargs.val.toList
 
   -- Subscript access: dict['key'] or list[0]
--- a/StrataTest/Languages/Python/expected_laurel/test_ifexpr.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_ifexpr.expected
@@ -1,0 +1,33 @@
+
+==== Verification Results ====
+List_get_body_calls_List_get_0: ✅ pass
+List_take_body_calls_List_take_0: ✅ pass
+List_drop_body_calls_List_drop_0: ✅ pass
+List_slice_body_calls_List_drop_0: ✅ pass
+List_slice_body_calls_List_take_1: ✅ pass
+List_set_body_calls_List_set_0: ✅ pass
+DictStrAny_get_body_calls_DictStrAny_get_0: ✅ pass
+Any_get_body_calls_DictStrAny_get_0: ✅ pass
+Any_get_body_calls_List_get_1: ✅ pass
+Any_get_body_calls_List_slice_2: ✅ pass
+Any_get_body_calls_List_drop_3: ✅ pass
+Any_get!_body_calls_DictStrAny_get_0: ✅ pass
+Any_get!_body_calls_List_get_1: ✅ pass
+Any_set_body_calls_List_set_0: ✅ pass
+Any_set!_body_calls_List_set_0: ✅ pass
+PFloorDiv_body_calls_Int.SafeDiv_0: ✅ pass
+PFloorDiv_body_calls_Int.SafeDiv_1: ✅ pass
+PFloorDiv_body_calls_Int.SafeDiv_2: ✅ pass
+PFloorDiv_body_calls_Int.SafeDiv_3: ✅ pass
+PAnd_body_calls_Any_to_bool_0: ✅ pass
+POr_body_calls_Any_to_bool_0: ✅ pass
+ret_type: ✅ pass (in prelude file)
+ret_type: ✅ pass (in prelude file)
+ret_pos: ✅ pass (in prelude file)
+assert_name_is_foo: ✅ pass (in prelude file)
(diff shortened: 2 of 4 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1625,119 +1625,111 @@ def PreludeInfo.merge (a b : PreludeInfo) : PreludeInfo where
...
+    match stmt with
+    | .ClassDef _ _ _ _ _ _ _ =>
+      let initCtx : TranslationContext := {
+        preludeProcedures := info.procedures,
+        preludeTypes := info.types,
+        compositeTypeNames := compositeTypeNames,
+        classFieldHighType := classFieldHighType,
+        filePath := filePath
+      }
+      let (composite, _instanceProcedures) ← translateClass initCtx stmt
```
MikaelMayer: The old code had a TODO comment here:
```
-- TODO uncomment this line and resolve compilation issues
-- procedures := procedures ++ _instanceProcedures
```
This TODO was silently dropped during the refactoring. Was this intentional? If the TODO is still relevant, it should be preserved (perhaps near the `procedures` declaration at line 1690).

StrataMain.lean
```diff
@@ -204,12 +204,10 @@ def diffCommand : Command where
     | _, _ =>
       exitFailure "Cannot compare dialect def with another dialect/program."
 
-def readPythonStrata (strataPath : String) : IO Strata.Program := do
+def readPythonStrata (strataPath : String) : IO (Array (Strata.Python.stmt SourceRange)) := do
```
MikaelMayer: This `readPythonStrata` function now delegates to `Strata.Python.readPythonStrataBytes` for parsing, which is good. It remains separate from `Strata.Python.readPythonStrata` because it uses `readBinInputSource` (stdin support via `-`). A short comment explaining why this wrapper exists (stdin support) would help future readers understand why it's not just calling `Strata.Python.readPythonStrata` directly.

StrataMain.lean
```diff
@@ -686,7 +677,7 @@ def pySpecToLaurelCommand : Command where
       match ← Strata.Python.Specs.readDDM ionFile |>.toBaseIO with
```
MikaelMayer: Nit: the empty string `""` passed as `modulePrefix` is correct given the API change, but it's not immediately obvious what it means. A named argument (`modulePrefix := ""`) would be self-documenting.

Strata/Languages/Python/ReadPython.lean
```diff
@@ -138,23 +138,12 @@ def pythonToStrata (dialectFile pythonFile : System.FilePath)
     runWithOptions options s!"parsing {pythonFile}"
       (runPyToStrata pythonCmd options.extraPythonArgs
         dialectFile pythonFile strataFile)
-    runWithOptions options s!"reading {pythonFile}" (readParseStrataFile strataFile)
+    runWithOptions options s!"reading {pythonFile}" (readPythonStrata strataFile.toString)
   finally
     match ← IO.FS.removeFile strataFile |>.toBaseIO with
     | .ok () => pure ()
     | .error msg => throw s!"Internal: Error deleting temp file {strataFile}: {msg}"
```
MikaelMayer: Nit: there's an extra blank line here (two blank lines before `end Strata.Python`), likely left over from the deletion of the old `readPythonStrata` that used to live at the bottom of the file.

StrataMain.lean
```diff
@@ -686,7 +677,7 @@ def pySpecToLaurelCommand : Command where
...
       | .ok t => pure t
       | .error msg => exitFailure s!"Could not read {ionFile}: {msg}"
-    let result := Strata.Python.Specs.ToLaurel.signaturesToLaurel pythonFile sigs
+    let result := Strata.Python.Specs.ToLaurel.signaturesToLaurel pythonFile sigs ""
```
shigoel, line 680: Always nice to see a default value removed from a function (i.e., `signaturesToLaurel`). 

However, I'm wondering if the empty string is the right default here, or indeed, if any default makes sense. Should pySpecToLaurelCommand derive a meaningful prefix (e.g., from the source file name) rather than passing in an empty string? Or is the empty string ok because we're only dealing with a single file and somehow there are guaranteed to be no collisions within it??

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1631,119 +1631,111 @@ def PreludeInfo.merge (a b : PreludeInfo) : PreludeInfo where
...
+      compositeTypes := compositeTypes ++ [composite]
```
joscoh, line 1672: This was already the case before, so I won't block on this, but there are several mutable lists here that are constructed by appending to the end in each iteration rather than using `::` and a single `reverse` at the end. Will these lists be large?

joehendrix: They may be -- I have no idea.  This is just an indention change though so I think optimizing the function is intentionally out of scope for the PR.  My other option would be to leave indention alone to shrink the diff.

StrataMain.lean
joehendrix: Judging from the context, this is a Strata subcommand to test converting PySpecToLaurel and reporting any errors.

We could add a prefix, but the consumer is likely just a person wanting to look at the Laurel and know what errors there are.

PR #647, opened 2026-03-24
Support Python package layout and module-name-based PySpec resolution

joehendrix:
PySpec commands can now work with an entire directory of PySpec files using standard Python conventions: `__init__.py` for package modules and relative imports (`from .Foo import Bar`). The CLI and API take dotted module names (e.g., `servicelib.Storage`) instead of individual Ion file paths, resolved against a spec directory.

- Previously, PySpec only understood absolute `from X import Y` imports. Relative imports, bare imports, and `__init__.py` packages all produced errors. This PR adds support for all of these so that real-world Python codebases with typical project layouts can be translated.
- When an imported module itself contains imports, those are now resolved relative to the imported file's own directory, not the top-level search path. This matches Python's import system and is essential for packages where `__init__.py` imports from sibling modules.
- `translateFile` now takes an explicit `searchPath` parameter instead of deriving it implicitly, giving callers control over where imports are resolved from.
- The existing `dispatch_test` fixture is converted from a flat `servicelib.py` to a proper Python package (`servicelib/__init__.py` with sibling modules), so the test suite actually exercises package-level imports end-to-end.
- A new `RelativeImportTest` suite covers basic relative imports, aliased imports, mixed absolute+relative, bare imports, dotted package paths, `__init__.py` resolution, and negative cases (unsupported level-2 imports, missing modules).

## SimpleAPI Changes

**Breaking:** `pyAnalyzeLaurel` and `pyTranslateLaurel` now take **module names** instead of Ion file paths.

| Parameter | Before | After |
|-----------|--------|-------|
| `dispatchPaths : Array String` | Ion file paths | `dispatchModules : Array String` — dotted module names (e.g., `"servicelib"`) |
| `pyspecPaths : Array String` | Ion file paths | `pyspecModules : Array String` — dotted module names (e.g., `"servicelib.Storage"`) |
| _(new)_ | — | `specDir : System.FilePath := "."` — directory containing compiled PySpec Ion files |

Module names are resolved against `specDir` using `ModuleName.specIonPath`, which tries canonical paths (`specDir/servicelib/Storage.pyspec.st.ion`) then falls back to `__init__` layout (`specDir/servicelib/__init__.pyspec.st.ion`).

`resolveAndBuildLaurelPrelude` has the same parameter changes (`dispatchPaths` → `dispatchModules`, `pyspecPaths` → `pyspecModules`).

## Strata Exe Changes

**Breaking:** The `--dispatch` and `--pyspec` flags now take **module names** instead of Ion file paths for these commands:
- `pyAnalyzeLaurel`
- `pyTranslateLaurel`
- `pyAnalyzeLaurelToGoto`

New flag: `--spec-dir <dir>` specifies where compiled PySpec Ion files live (defaults to `.`). The command exits with an error if the directory does not exist.

**Before:**
```sh
strata pyAnalyzeLaurel --dispatch path/to/dispatch.pyspec.st.ion --pyspec path/to/Storage.pyspec.st.ion file.ion
```

**After:**
```sh
strata pyAnalyzeLaurel --spec-dir path/to/pyspec --dispatch servicelib --pyspec servicelib.Storage file.ion
```


Strata/Languages/Python/Specs.lean
```diff
@@ -106,20 +111,61 @@ def findInPath (mod : ModuleName) (searchPath : System.FilePath) : EIO String Sy
...
+  -- Fail both
+  throw s!"{file} not found (also no {initFile})."
 
 /-- Generates the output filename for a module's spec file. -/
 public def strataFileName (mod : ModuleName) : String := s!"{mod.fileRoot}.pyspec.st.ion"
 
+/-- Construct the `.pyspec.st.ion` file path for a module under the given root.
+    Mirrors subdirectories for multi-component module names, e.g.
+    module `service.module` with root `/specs` → `/specs/service/module.pyspec.st.ion`. -/
+public def ionPath (mod : ModuleName) (root : System.FilePath) : System.FilePath :=
```
MikaelMayer: `ionPath` is `public` but has no callers anywhere in the codebase. If it's intended for future use, consider leaving it non-public until needed; if it was meant to replace the inline `mod.foldlDirs (init := ...) (· / ·) / mod.strataFileName` at line 1455, it could be used there.

StrataTest/Languages/Python/Specs/RelativeImportTest.lean
```diff
@@ -0,0 +1,105 @@
...
+namespace Strata.Python.Specs.RelativeImportTest
+
+open Strata.Python.Specs (translateFile)
+
+private meta def testDir : System.FilePath :=
+  "StrataTest/Languages/Python/Specs/import_test"
+
+/-- Check if `needle` is a substring of `haystack`. -/
+private meta def containsSubstr (haystack needle : String) : Bool :=
+  (haystack.splitOn needle).length != 1
```
MikaelMayer: Minor: `containsSubstr` is duplicated from `SpecsTest.lean` (line 252). Could be extracted to a shared test utility.

Strata/Languages/Python/Specs.lean
```diff
@@ -1439,7 +1487,24 @@ partial def resolveModule (loc : SourceRange) (modName : String) :
...
+  -- (e.g., __init__.py doing `from .module import Foo`).
+  let parentCtx ← read
+  let childSearchPath := pythonFile.parent.getD pythonFile
+  let childModuleReader : ModuleReader := fun (childMod : ModuleName) => do
+    let childPath ← childMod.findInPath childSearchPath
+    baseLogEvent parentCtx.eventSet "findFile"
+      s!"Found {childMod} as {childPath} in {childSearchPath}"
+    match ← IO.FS.readFile childPath |>.toBaseIO with
+    | .ok _contents =>
+      pure childPath
```
shigoel: The childModuleReader reads each nested file's contents but discards them (_contents), so the file's FileMap is never registered in fileMapsRef. The original moduleReader in translateModule (line 1603) explicitly calls `fileMapsRef.modify fun m => m.insert pythonPath fm` to store each file's map. As a result, any compilation error arising from a module two or more levels deep in an import chain (e.g., __init__.py → Messaging.py → some further import) will be reported without source-location context (no line/column), because `fmm[e.file]?` in the error-reporting loop will return none for those files.

joehendrix: This is a good catch.  I have added a regression test to catch this and fixed the logic.

Strata/Languages/Python/Specs.lean
```diff
@@ -1457,19 +1564,92 @@ partial def resolveModule (loc : SourceRange) (modName : String) :
...
+      -- types so that downstream references don't produce unknown-identifier
+      -- errors.
+      for a in names do
+        let name := a.name
+        let asname := a.asname.getD name
+        let source : PythonIdent := { pythonModule := mod, name := name }
+        let tpv : SpecValue := .typeValue (.ident loc source)
+        setNameValue asname tpv
+        pushSignature (.externTypeDecl asname source)
+  | none =>
```
MikaelMayer, line 1643: The body of this `none` branch (lines 1637–1652) is nearly identical to `translateImport` (lines 1593–1606) — both resolve each name as a module and register exports under `"{asname}.{name}"`, with the same fallback to opaque extern types. The only difference is the `resolveRelativeModuleName` call. Consider extracting a shared helper like `resolveAndRegisterModule(loc, mod, asname)` to reduce duplication.

Strata/SimpleAPI.lean
```diff
@@ -204,78 +206,148 @@ inductive WarningOutput where
...
+          | .error e => throw s!"Invalid module name '{m}': {e}"
+        let (path, _) ←
+          match ← ModuleName.findInPath mod sourceDir |>.toBaseIO with
+          | .ok r => pure r
+          | .error e => throw s!"Module '{m}' not found in {sourceDir}: {e}"
+        result := result.push (mod, path)
+      pure result
+
+  -- Translate each module
+  let mut totalWarnings : Nat := 0
```
MikaelMayer: `totalWarnings` is accumulated (line 334) but never read after the loop — it's dead code. Either remove it or use it (e.g., in a summary message at the end).

StrataTest/Languages/Python/Specs/RelativeImportTest.lean
```diff
@@ -0,0 +1,153 @@
...
+
+namespace Strata.Python.Specs.RelativeImportTest
+
+open Strata.Python.Specs (translateFile ModuleName)
+
+private meta def testDir : System.FilePath :=
+  "StrataTest/Languages/Python/Specs/import_test"
+
+/-- Check if `needle` is a substring of `haystack`. -/
+private meta def containsSubstr (haystack needle : String) : Bool :=
```
MikaelMayer: Minor: `containsSubstr` is duplicated from `SpecsTest.lean` (line 252). Could be extracted to a shared test utility.

PR #652, opened 2026-03-24
Implement more functions in `SimpleAPI`

atomb:
This PR implements several previously-opaque functions in the `Strata/SimpleAPI.lean` and consolidates the verification entry point.

### Changes

Implemented Simple API functions:
- `genericToCore` — translates a generic `Strata.Program` to` Core.Program` using
`Core.getProgram`
- `genericToLaurel` — translates a generic `Strata.Program` to `Laurel.Program` using
`Laurel.parseProgram`
- `laurelToCore` — translates a `Laurel.Program` to `Core.Program` using the existing
Laurel translator
- `Core.inlineProcedures` — inlines procedure calls with a configurable predicate
controlling which calls to inline
- `Core.loopElimUsingContract` — eliminates loops by replacing them with invariant-
based assertions/assumptions
- `Core.callElimUsingContract` — eliminates procedure calls by replacing them with
contract-based assertions/assumptions
- `Core.verifyProgram` — runs deductive verification on a Core program (replaces
the standalone `verifyCore`)

Structural changes:
- `Core.InlineTransformOptions` is now a concrete structure with an optional
`doInline` predicate, rather than an opaque type
- Removed the standalone `verifyCore` function; its logic is now in
`Core.verifyProgram` under the `Core` namespace
- Renamed `Core.loopElimWithContract` → `Core.loopElimUsingContract` and
`Core.callElimWithContract` → `Core.callElimUsingContract`
- Trimmed unnecessary imports from `SimpleAPI.lean` and `StrataMain.lean`
- Fixed a docstring typo (Core dialect → Laurel dialect for `genericToLaurel`)

Transform module workarounds:
- Added top-level aliases `coreCallElimCmd` and `coreInlineCallCmd` in `CallElim.lean`
and `ProcedureInlining.lean` to work around the Core module/dialect name collision

Dogfooding in `StrataMain`:
- Updated `pyAnalyzeCommand` to use `Core.inlineProcedures` instead of calling the
transform directly
- Updated `pyAnalyzeLaurelCommand` to use `Core.verifyProgram` instead of
`Strata.verifyCore`

### Motivation

The Simple API was originally sketched with opaque declarations to define the
intended surface. This PR makes the majority of those functions concrete, moving
the API closer to being usable by external clients. The remaining
`noncomputable opaque` functions depend on DDM functionality that is not yet
available.


Strata/SimpleAPI.lean
```diff
@@ -160,39 +156,80 @@ dialect into the dialect-specific AST for the Core dialect. This can fail with
...
 /--
 Transform a Core program to inline some or all procedure calls.
 -/
-noncomputable opaque Core.inlineProcedures : Core.Program → Core.InlineTransformOptions → Core.Program
+def Core.inlineProcedures (p : Core.Program) (opts : Core.InlineTransformOptions)
+    : Except String Core.Program :=
+  let pred := opts.doInline.getD (fun _ _ => true)
+  match Core.Transform.run p (fun prog => do
+    let (_, prog) ← Core.Transform.runProgram (coreInlineCallCmd (doInline := pred)) prog
+    return prog) with
```
joscoh: Is this pattern match needed?

Strata/SimpleAPI.lean
```diff
@@ -160,39 +156,80 @@ dialect into the dialect-specific AST for the Core dialect. This can fail with
...
 
 /--
 Transform a Core program to replace each procedure call with assertions and
 assumptions about its contract.
 -/
-noncomputable opaque Core.callElimWithContract : Core.Program → Core.Program
+def Core.callElimUsingContract (p : Core.Program) : Except String Core.Program :=
+  match Core.Transform.run p (fun prog => do
+    let (_, prog) ← Core.Transform.runProgram coreCallElimCmd prog
+    return prog) with
```
joscoh: Same here

Strata/SimpleAPI.lean
```diff
@@ -160,39 +156,80 @@ dialect into the dialect-specific AST for the Core dialect. This can fail with
...
 supported.
 -/
-noncomputable opaque laurelToCore : Laurel.Program → Except String Core.Program
+def laurelToCore (p : Laurel.Program) : Except String Core.Program :=
+  let (coreOpt, diags) := Laurel.translate { emitResolutionErrors := true } p
+  match coreOpt with
+  | some core => .ok core
+  | none => .error s!"Laurel to Core translation failed: {diags.map (·.message)}"
 
 /-! ### Transformation of Core programs -/
```
joscoh, line 167: Just wondering, is the API going to have a function for each transformation?

Strata/SimpleAPI.lean
```diff
@@ -137,7 +125,12 @@ Translate a program in the generic AST for Strata into the dialect-specific AST
 for Core. This can fail with an error message if the input is not a
 well-structured instance of the Core dialect.
 -/
-noncomputable opaque genericToCore : Strata.Program → Except String Core.Program
+def genericToCore (p : Strata.Program) : Except String Core.Program :=
+  let (program, errors) := Core.getProgram p
+  if errors.isEmpty then
+    .ok program
+  else
+    .error s!"Core DDM translation errors: {errors.toList}"
```
shigoel: Nit, but it'd be nice to use `String.intercalate` with newlines or some such for a more readable output

shigoel (approved):

Strata/SimpleAPI.lean
atomb: Oh, you're right. The original and final monads are superficially different, but no match is actually necessary.

Strata/SimpleAPI.lean
atomb: Same answer :)

PR #655, opened 2026-03-25
Fix BoogieToStrata: handle cross-nesting and backward gotos

tautschnig:
Cross-nesting forward gotos: recursively collect goto targets from nested if/while bodies so wrapper blocks are created at the correct nesting level. Track enclosing wrapper labels to avoid emitting duplicate labeled blocks.

Backward gotos (loops): translate back-edges to while(true) loops instead of extending wrapper closeAt indices (which broke Strata's exit semantics). For each detected back-edge, the affected blocks are wrapped in while(true) with the back-edge target label wrapping the entire loop body, so 'exit <label>' acts as continue. Forward targets within the loop body use inner wrappers; forward exits to labels outside the loop propagate through the while as normal.

Add CrossNestingExit regression test covering:
- CrossNestingForward: goto from inside if-branch to sibling block
- BackwardGotoBuggy: demonstrates the old approach incorrectly passed a buggy assertion (g <= 2 after a doubling loop), while the new while-loop encoding correctly detects the bug
- BackwardGotoLoop: simple backward goto forming a loop
- CrossNestingInLoop: cross-nesting goto inside a loop body


Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -1075,15 +1115,56 @@ private void EmitWithExitWrappers(
...
+    /// The synthetic "_exit" label (procedure exit) is excluded since it represents
+    /// procedure return, not a loop target.
+    /// Returns a dictionary mapping back-edge target label to the loop end index (inclusive).
+    /// </summary>
+    private static Dictionary<string, int> DetectBackEdges(
+        Dictionary<string, int> labelToIndex,
+        Dictionary<string, int> maxSourceForTarget) {
+        var backEdges = new Dictionary<string, int>();
+        foreach (var (label, maxSource) in maxSourceForTarget) {
+            if (label == "_exit") continue;
```
atomb: Nit: I'd rather have this be a named constant holding a string value, to ensure that it's the same everywhere. A typo here could be a source of subtle bugs.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -1160,6 +1171,73 @@ private static Dictionary<string, int> DetectBackEdges(
...
+                                     IList<BigBlock> bigBlocks) {
+        for (var i = 0; i < siblings.Count; i++) {
+            var existing = siblings[i];
+            if (newRegion.start == existing.start) {
+                // Same start: merge
+                existing.labels.AddRange(newRegion.labels);
+                existing.end = Math.Max(existing.end, newRegion.end);
+                return;
+            }
+            if (newRegion.start > existing.start && newRegion.end <= existing.end) {
```
atomb, line 1216: `InsertRegion` doesn't check if `newRegion` fully contains `existing`.**

   The nesting check only handles `newRegion` inside `existing`:
   ```csharp
   if (newRegion.start > existing.start && newRegion.end <= existing.end) {
       // Properly nested inside existing: recurse into children
   ```
   But if `newRegion` is wider than `existing` (i.e., `newRegion.start < existing.start` or `newRegion.end > existing.end` while still overlapping), the code falls through to the overlap check and throws an irreducible-flow exception. Since `sorted` processes by ascending start then descending end, the wider region should come first for the same start (handled by merge). But consider two regions: `[1,5]` already inserted, then `[0,6]` arrives — this would not match any branch and would be appended as a sibling, creating two overlapping top-level regions. This could happen if back-edges are discovered in an order where a wider outer loop is processed after a narrower inner one with a different start.

   The sort `OrderBy(start).ThenByDescending(end)` should prevent this for same-start cases, but for different-start cases where the new region fully contains an existing one, the existing region should become a child of the new region. Consider adding a check for `newRegion.start <= existing.start && newRegion.end >= existing.end`.

Tools/BoogieToStrata/Source/StrataGenerator.cs
atomb: This isn't entirely new to the PR, but might be relevant.

`EmitStmtList` inner target classification uses linear scan over all regions for every target.**

   ```csharp
   for (var r = 0; r < loopRegions.Count; r++) {
       var (start, end, _) = loopRegions[r];
       if (closeAt > start && closeAt <= end) {
   ```
   This only checks top-level regions, not children. If a forward target's `closeAt` falls inside a child region (nested loop), it gets assigned to the parent region's `innerTargets` but may need to be further delegated to the child. The `EmitLoopRegion` method does handle this delegation (lines 1430-1438), so this may work in practice, but the two-level dispatch is fragile — a target that falls inside a grandchild region would need to be passed through two levels of `innerTargetLabels`.

Tools/BoogieToStrata/Source/StrataGenerator.cs
tautschnig: Good catch — fixed in 0f0a1bbc. Replaced all 7 occurrences with a `private const string ExitLabel` field.

Tools/BoogieToStrata/Source/StrataGenerator.cs
tautschnig: You're right that the sort order alone isn't a sufficient guarantee. Fixed in 00d350c5: `InsertRegion` now reparents existing siblings as children when the new region fully contains them, and rejects partial overlaps as irreducible.

Tools/BoogieToStrata/Source/StrataGenerator.cs
tautschnig: Agreed this deserves documentation. Added a comment explaining the delegation chain in 91d380f4, plus a `TripleNestedLoops` test with three nesting levels to exercise the grandchild case end-to-end.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -1075,15 +1131,142 @@ private void EmitWithExitWrappers(
...
+    private static void CollectChildBackEdgeLabels(LoopRegion region, HashSet<string> result) {
+        foreach (var child in region.children) {
+            foreach (var l in child.labels) result.Add(l);
+            CollectChildBackEdgeLabels(child, result);
+        }
+    }
+
     private void EmitStmtList(StmtList stmtList) {
         var bigBlocks = stmtList.BigBlocks;
+        // Collect goto targets from direct children AND nested structures
```
MikaelMayer, line 1263: The `CollectGotoTargets` call on the line above is redundant: `CollectNestedGotoTargets` already collects from each BigBlock's `tc` (GotoCmd/ReturnCmd) in addition to recursing into nested `ec` structures. Since both write to the same `HashSet`, the first call contributes nothing that the second doesn't already cover. You could replace both lines with just `CollectNestedGotoTargets(bigBlocks, gotoTargets)` (or rename it to make the intent clearer).

MikaelMayer (commented):
Solid implementation of cross-nesting and backward goto handling. The decomposition into `ComputeMaxSourceForTarget`, `DetectBackEdges`, `BuildLoopRegions`, `InsertRegion`, and `EmitLoopRegion` keeps the complexity manageable. The irreducible control-flow detection is a nice safety net. Test coverage is thorough — the `BackwardGotoBuggy` test demonstrating that the old approach missed a real bug is a great regression test.

A few observations below, nothing blocking.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -7,6 +7,17 @@ internal class StrataConversionException(IToken tok, string s) : Exception {
...
 }
 
+internal class LoopRegion(int start, int end, List<string> labels) {
+    public int start = start;
+    public int end = end;
+    public List<string> labels = labels;
+    public List<LoopRegion> children = [];
+
+    public void Deconstruct(out int s, out int e, out List<string> l) {
+        s = start; e = end; l = labels;
```
MikaelMayer: The `Deconstruct` is only used once (line 1332), and there only `start` and `end` are used while `labels` is discarded. Consider removing `Deconstruct` and accessing `region.start` / `region.end` directly — it would be clearer at the call site what's being read.

Tools/BoogieToStrata/Source/StrataGenerator.cs
tautschnig: Replaced the redundant `CollectGotoTargets` + `CollectNestedGotoTargets` pair with justCollectNestedGotoTargets` (which already covers the top-level `tc` targets).

Tools/BoogieToStrata/Source/StrataGenerator.cs
tautschnig: Removed `LoopRegion.Deconstruct` and replaced the single call site with direct `loopRegions[r].start` / `loopRegions[r].end` access.

PR #656, opened 2026-03-25
Add Laurel tests for Java machine integers as constrained types

tautschnig:
Model Java primitive integer types (byte, short, char, int, long) as Laurel constrained types over mathematical int with range predicates.

Tests cover:
- Range constraints on inputs and outputs
- Valid and invalid returns (overflow/underflow detection)
- Cross-type widening (byte→int safe, int→byte unsafe)
- Arithmetic within range
- Opaque procedure contracts with range guarantees
- Quantifiers over constrained types


StrataTest/Languages/Laurel/Examples/Fundamentals/T18_JavaMachineIntegers.lean
```diff
@@ -0,0 +1,129 @@
...
+import StrataTest.Util.TestDiagnostics
+import StrataTest.Languages.Laurel.TestExamples
+
+open StrataTest.Util
+
+namespace Strata
+namespace Laurel
+
+def program := r"
+// Java machine integer types as constrained types over mathematical int.
```
keyboardDrummer: What is this test adding that the existing tests for constrained types did not add yet? I don't think we need Java specific tests in the Laurel codebase: those can live on the JVerify side.

tautschnig: None of the existing tests exercised use of constrained types that are equivalent to bitvector types. (I have a follow-up PR that will introduce an option to translate such types to Core's bitvector types.)

keyboardDrummer: > None of the existing tests exercised use of constrained types that are equivalent to bitvector types. (I have a follow-up PR that will introduce an option to translate such types to Core's bitvector types.)

A PR that would translate constrained types with specific constraints to BV types? That seems like too implicit behavior. Why not make BV types explicit in Laurel?

tautschnig: No, it won't be implicit behaviour, it will be controlled via an option. The goal is to be able to compare the different encoding approaches, so making BV types explicit in Laurel could very well be done, but is somewhat orthogonal to that.

keyboardDrummer: Is there a use-case for automatically translating constrained types to bitvectors? If a front-end wants to use bitvectors, why not use them directly (once they exist in Laurel) ?

I imagine we'll have options that allow configuring how to verify Laurel bitvectors, but I don't see a need for an option to convert constrained types to bvs.

Also, translating constrained types into bitvectors seems complicated. What if you have a constrained type like:
```
constrained nat32 := i: int | i >= 0 && i < ...
```

Are you going to translate that into 
```
constrained nat32 := i: i32 | i >= 0
```
?

I'd rather ask the front-end to immediately write the second version.

tautschnig: My take so far is a very simple one: only use bitvectors when the constrained type directly maps to a bitvector. So, yes, I can do a PR that introduces bitvectors to Laurel, and then future front-ends can use those directly. My proposed constrained-types-to-bitvectors path should then become one available as a Core-to-Core translation (and perhaps one that should be available in both directions). That would then provide flexibility for verification as you suggest.

keyboardDrummer: > My proposed constrained-types-to-bitvectors path should then become one available as a Core-to-Core translation

I don't understand what you mean. Core doesn't have a concept of constrained types so how could such a path exist as a Core to Core translation?

But the heart of my feedback remains the same. If explicit bv types are available, then I don't see a use-case for a translation that maps certain constrained types to bitvectors. Do you?

tautschnig: Quantifiers and bitvectors don't play well together in SMT solvers, so I do think it is reasonable that we have a verification path that uses constrained types rather than bitvectors. But, yes, this should be a decision to be made with respect to a specific back-end rather than prematurely by a front-end.

tautschnig: https://github.com/strata-org/Strata/pull/701 now adds native BV types to Laurel.

atomb: I like adding bit vectors, as in #701, because I'd argue that, even if we use a bit vector type in a Strata dialect, we don't necessarily need to use a bit vector type in the back end solver. We just need to make sure that the semantics are consistent with bit vectors. Said differently, bit vectors in Laurel or Core could be translated to bounded integers in SMT without trouble. And, to me, doing so seems preferable. It means modeling the types as accurately as possible -- Java integer types _are_ bit vectors -- but doesn't need to constrain the type of reasoning we do later in the pipeline.

keyboardDrummer: > Quantifiers and bitvectors don't play well together in SMT solvers, so I do think it is reasonable that we have a verification path that uses constrained types rather than bitvectors. But, yes, this should be a decision to be made with respect to a specific back-end rather than prematurely by a front-end.

Then it's still unclear to me what the use-case of the test in this PR is.

tautschnig: It was meant to provide example of what Java types one day might look like in Laurel. Now it seems that #701 should be the preferred approach, so I'm ok closing this PR.

keyboardDrummer: > Now it seems that https://github.com/strata-org/Strata/pull/701 should be the preferred approach, so I'm ok closing this PR.

It seems like that to me as well. Shall we close this then?

PR #660, opened 2026-03-25
Define Hoare triple / validity of asserts / correct transform, prove correctness of DetToNondet with loops

aqjune-aws:
**Specification of Transformation**: Strata/Transform/Specification.lean

- Define two notions of assertion validity (`AssertValid` via reachability, `Hoare.Triple` via partial-correctness triples) and prove them equivalent (`hoareTriple_implies_assertValid`, `assertValid_implies_hoareTriple`).
- Define `Sound` and `Overapproximates` predicates for transformation correctness with `WellFormedSemanticEvalBool`/`Val` preconditions.
- Prove `overapproximates_triple`, composition, identity, and block-level `overapproximates_stmts` (using `noFuncDecl` for eval preservation).
- Add structural Hoare rules: `cmd`, `seq_cons`, `ite`, `skip`, block lifting.
- Add `Lang P` structure abstracting statement/config types and step relation for generality.

**The new `hasFailure` flag in a program state, and small-step of Imperative**

- Add `Env P` record bundling store, evaluator, and cumulative `hasFailure` flag. Refactor `Config`, `StepStmt`, `EvalStmt`, `EvalBlock` to use `Env`.
- Properly amend small-step semantics of Imperative; copied verbatim from https://github.com/strata-org/Strata/pull/509, @MikaelMayer

**Correctness proof**: DetToNondetCorrect.lean
- Prove `detToNondet_overapproximates` using the `Overapproximates` predicate defined above.
- Defined small-step semantics of NondetStmt. This also supports loops.
- Fix DetToNondet pass to optionally return `none` if funcDecl or exit is met.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Strata/Languages/Python/PySpecPipeline.lean
+++ b/Strata/Languages/Python/PySpecPipeline.lean
@@ -313,6 +313,23 @@ private def prependPrelude (coreFromLaurel : Core.Program) : Core.Program :=
     fun d => !coreOnlyNames.contains (toString d.name)
   { decls := filteredPrelude ++ coreOnly ++ userDecls }
 
+/-- Split procedure names in a Core program into prelude names
+    (before `FIRST_END_MARKER`) and user names (after it).
+    If `FIRST_END_MARKER` is absent, nothing is considered prelude. -/
+public def splitProcNames (prog : Core.Program)
+    : Std.HashSet String × List String :=
+  let (before, rest) := prog.decls.span (fun d => toString d.name != "FIRST_END_MARKER")
+  let (preludeDecls, userDecls) := match rest with
+    | _ :: tl => (before, tl)  -- marker found: before is prelude, after is user
+    | [] => ([], before)       -- no marker: everything is user
+  let preludeNames := preludeDecls.foldl (init := ({} : Std.HashSet String)) fun s d =>
+    match d.getProc? with
+    | some p => s.insert (Core.CoreIdent.toPretty p.header.name)
+    | none => s
+  let userProcNames := userDecls.filterMap fun d =>
+    d.getProc?.map (Core.CoreIdent.toPretty ·.header.name)
+  (preludeNames, userProcNames)
+
 /-- Translate a combined Laurel program to Core and prepend the full
     runtime prelude.  Resolution errors are suppressed because PySpec
     Laurel procedures reference names defined in the Core prelude
--- a/Strata/SimpleAPI.lean
+++ b/Strata/SimpleAPI.lean
@@ -215,10 +215,11 @@ def Core.verifyProgram
     (program : Core.Program)
     (options : Core.VerifyOptions)
     (moreFns : @Lambda.Factory Core.CoreLParams := Lambda.Factory.default)
+    (proceduresToVerify : Option (List String) := none)
     : EIO String Core.VCResults := do
   let runVerification (tempDir : System.FilePath) : IO Core.VCResults :=
     EIO.toIO (IO.Error.userError ∘ toString)
-      (Core.verify program tempDir .none options moreFns)
+      (Core.verify program tempDir proceduresToVerify options moreFns)
   let ioAction := match options.vcDirectory with
     | .some vcDir => IO.FS.createDirAll vcDir *> runVerification vcDir
     | .none => IO.FS.withTempDir runVerification
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -529,6 +529,9 @@ def pyAnalyzeLaurelCommand : Command where
       IO.println "\n==== Core Program ===="
       IO.print coreProgram
 
+    -- Split prelude / user procedure names at FIRST_END_MARKER
+    let (preludeNames, userProcNames) := Strata.splitProcNames coreProgram
+
     if let some dir := keepDir then
       let path := s!"{dir}/{baseName}.core"
       IO.FS.writeFile path (toString coreProgram)
@@ -538,12 +541,6 @@ def pyAnalyzeLaurelCommand : Command where
     let pyspecFiles := pflags.getRepeated "pyspec"
     let coreProgram ←
       if pyspecFiles.size > 0 then
-        -- Collect prelude procedure names to avoid inlining them
-        let mut preludeNames : Std.HashSet String := {}
-        for d in coreProgram.decls do
-          if toString d.name == "FIRST_END_MARKER" then break
-          if let some p := d.getProc? then
-            preludeNames := preludeNames.insert (Core.CoreIdent.toPretty p.header.name)
         match Core.Transform.runProgram (targetProcList := .none)
               (Core.ProcedureInlining.inlineCallCmd
                 (doInline := λ name _ => name ≠ "__main__" && !preludeNames.contains name))
@@ -574,7 +571,8 @@ def pyAnalyzeLaurelCommand : Command where
         | none => baseOptions
     let vcResults ←
       match ← Core.verifyProgram coreProgram options
-                (moreFns := Strata.Python.ReFactory) |>.toBaseIO with
+                (moreFns := Strata.Python.ReFactory)
+                (proceduresToVerify := some userProcNames) |>.toBaseIO with
       | .ok r => pure r
       | .error msg => exitPyAnalyzeInternalError msg
 
--- a/StrataTest/Languages/Python/AnalyzeLaurelTest.lean
+++ b/StrataTest/Languages/Python/AnalyzeLaurelTest.lean
@@ -107,13 +107,9 @@ private meta def runAnalyzeAndVerify
   let coreProgram ← match coreProgramOption with
     | none => return .error "Laurel to Core translation failed"
     | some core => pure core
-  -- Inline all non-main procedures
+  -- Split prelude / user procedure names at FIRST_END_MARKER
+  let (preludeNames, userProcNames) := Strata.splitProcNames coreProgram
   -- Inline all non-main, non-prelude procedures
-  let mut preludeNames : Std.HashSet String := {}
-  for d in coreProgram.decls do
-    if toString d.name == "FIRST_END_MARKER" then break
-    if let some p := d.getProc? then
-      preludeNames := preludeNames.insert (Core.CoreIdent.toPretty p.header.name)
   let coreProgram ← match Core.Transform.runProgram (targetProcList := .none)
         (Core.ProcedureInlining.inlineCallCmd
           (doInline := λ name _ => name ≠ "__main__" && !preludeNames.contains name))
@@ -126,7 +122,8 @@ private meta def runAnalyzeAndVerify
       stopOnFirstError := false, verbose := .quiet, solver := "z3",
       checkMode := .bugFinding, checkLevel := .full }
   match ← Core.verifyProgram coreProgram options
-      (moreFns := Strata.Python.ReFactory) |>.toBaseIO with
+      (moreFns := Strata.Python.ReFactory)
+      (proceduresToVerify := some userProcNames) |>.toBaseIO with
   | .ok results => return .ok results
   | .error msg => return .error (toString msg)
 
--- a/StrataTest/Languages/Python/PreludeVerifyTest.lean
+++ b/StrataTest/Languages/Python/PreludeVerifyTest.lean
@@ -1,0 +1,151 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Python.PySpecPipeline
+import Strata.Languages.Python.PyFactory
+import Strata.Languages.Core.Verifier
+
+/-! # Prelude Verification Test
+
+Verify that all prelude procedures pass verification.
+This ensures the Python runtime prelude is well-formed
+after PrecondElim generates WF-checking procedures. -/
+
+namespace Strata.Python.PreludeVerifyTest
+
+/-- Build the full Core prelude program (Laurel-translated + Core-only parts). -/
+private def preludeProgram : Core.Program :=
+  let (coreOption, _) := Strata.translateCombinedLaurel pythonRuntimeLaurelPart
+  match coreOption with
+  | some prog => prog
+  | none => { decls := [] }
+
+private def verifyPrelude : IO Core.VCResults := do
+  IO.FS.withTempDir fun tempDir => do
+    let r ← EIO.toIO (IO.Error.userError ∘ toString)
+      (Core.verify preludeProgram tempDir
+        (options := .quiet)
+        (moreFns := Strata.Python.ReFactory))
+    return r
+
+/--
+info:
+Obligation: List_get_body_calls_List_get_0
+Property: assert
+Result: ✅ pass
+
+Obligation: List_take_body_calls_List_take_0
+Property: assert
+Result: ✅ pass
+
+Obligation: List_drop_body_calls_List_drop_0
+Property: assert
+Result: ✅ pass
+
+Obligation: List_slice_body_calls_List_drop_0
+Property: assert
+Result: ✅ pass
+
+Obligation: List_slice_body_calls_List_take_1
+Property: assert
+Result: ✅ pass
+
+Obligation: List_set_body_calls_List_set_0
+Property: assert
+Result: ✅ pass
+
+Obligation: DictStrAny_get_body_calls_DictStrAny_get_0
+Property: assert
+Result: ✅ pass
+
+Obligation: Any_get_body_calls_DictStrAny_get_0
+Property: assert
+Result: ✅ pass
+
+Obligation: Any_get_body_calls_List_get_1
+Property: assert
+Result: ✅ pass
+
+Obligation: Any_get_body_calls_List_slice_2
+Property: assert
+Result: ✅ pass
+
+Obligation: Any_get_body_calls_List_drop_3
+Property: assert
+Result: ✅ pass
+
+Obligation: Any_get!_body_calls_DictStrAny_get_0
+Property: assert
+Result: ✅ pass
+
+Obligation: Any_get!_body_calls_List_get_1
+Property: assert
+Result: ✅ pass
+
+Obligation: Any_set_body_calls_List_set_0
+Property: assert
+Result: ✅ pass
+
+Obligation: Any_set!_body_calls_List_set_0
+Property: assert
+Result: ✅ pass
+
+Obligation: PFloorDiv_body_calls_Int.SafeDiv_0
+Property: division by zero check
+Result: ✅ pass
+
+Obligation: PFloorDiv_body_calls_Int.SafeDiv_1
+Property: division by zero check
+Result: ✅ pass
+
+Obligation: PFloorDiv_body_calls_Int.SafeDiv_2
+Property: division by zero check
+Result: ✅ pass
+
+Obligation: PFloorDiv_body_calls_Int.SafeDiv_3
+Property: division by zero check
+Result: ✅ pass
+
+Obligation: PAnd_body_calls_Any_to_bool_0
+Property: assert
+Result: ✅ pass
+
+Obligation: POr_body_calls_Any_to_bool_0
+Property: assert
+Result: ✅ pass
+
+Obligation: ret_type
+Property: assert
+Result: ✅ pass
+
+Obligation: ret_type
+Property: assert
+Result: ✅ pass
+
+Obligation: ret_pos
+Property: assert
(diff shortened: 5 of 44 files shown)


Strata/Transform/Specification.lean
aqjune-aws, line 1: This file contains the most important definitions in this pull request.

Strata/DL/Imperative/NondetStmtSemantics.lean
```diff
@@ -5,60 +5,176 @@
...
+
+A configuration is either executing a `NondetStmt`, sequencing two parts
+(left config + right continuation), or terminated.
+-/
+
+/-- Configurations for small-step execution of `NondetStmt`. -/
+inductive NondetConfig (P : PureExpr) (CmdT : Type) : Type where
+  /-- Executing a single nondeterministic statement. -/
+  | stmt : NondetStmt P CmdT → Env P → NondetConfig P CmdT
+  /-- Sequencing: execute the left config, then continue with the right stmt. -/
```
joscoh, line 28: Why do we need this sequencing operator in the configurations?

Strata/DL/Imperative/SemanticsProps.lean
```diff
@@ -17,76 +17,229 @@ public section
...
+/-! ### hasFailure monotonicity and irrelevance
+
+`hasFailure` is never consulted by any `EvalStmt` / `EvalBlock` premise,
+so it is both *monotone* (once `true`, stays `true`) and *irrelevant*
+(changing only `hasFailure` in the input env yields an execution with the
+same `store` and `eval` in the output).
+-/
+
+mutual
+theorem EvalStmt_hasFailure_monotone
```
joscoh, line 80: I thought the big-step semantics for statements are going to be deprecated and replaced only with the small-step semantics? Is there a reason that these theorems are stated in terms of the big-step semantics rather than the small-step (star) version?

Strata/DL/Imperative/StmtSemanticsSmallStep.lean
```diff
@@ -20,289 +20,1233 @@ public section
...
+The proof proceeds in two parts:
+1. A helper lemma (`seq_inner_star`) showing that multi-step execution of
+   the inner config lifts to multi-step execution of the enclosing `.seq`.
+2. The main theorem (`stmts_cons_step`) composing the pieces.
+-/
+
+/-- Transitivity of `ReflTrans`: if `r* x y` and `r* y z` then `r* x z`.
+    This is a local helper that avoids the opaque `Transitive` wrapper
+    from `Relations.lean` (which becomes opaque across module boundaries). -/
+theorem reflTrans_trans {r : Relation A}
```
joscoh: Isn't it better to just expose `ReflTrans_Transitive` rather than proving it again?

Strata/DL/Imperative/StmtSemanticsSmallStep.lean
```diff
@@ -20,289 +20,1233 @@ public section
...
+      subst htgt; cases hrest with | step _ _ _ h _ => cases h
+    | step_block_exit_mismatch =>
+      subst htgt
+      cases hrest with
+      | refl => exact ⟨_, .refl _⟩
+      | step _ _ _ h _ => cases h
+
+/-! ## Store/eval simulation and hasFailure irrelevance -/
+
+/-- Two configs agree on store/eval (may differ on hasFailure). -/
```
joscoh, line 581: This comment is not quite correct: it also checks that configs have the same "shape" (e.g. `.stmt s p1` and `.terminal p2` may agree on store and eval but are not `ConfigSE`).

Strata/DL/Imperative/StmtSemanticsSmallStep.lean
```diff
@@ -20,289 +20,1233 @@ public section
...
+  | _, _ => False
+
+/-- Single-step simulation: if two configs agree on store/eval and one steps,
+    the other can take the same step with store/eval preserved. -/
+private def step_simulation
+    (c₁ c₁' c₂ : Config P CmdT)
+    (hstep : StepStmt P EvalCmd extendEval c₁ c₁')
+    (heq : ConfigSE P c₁ c₂) :
+    ∃ c₂', StepStmt P EvalCmd extendEval c₂ c₂' ∧ ConfigSE P c₁' c₂' := by
+  cases hstep with
```
joscoh, line 598: There has to be a nicer way to prove this, though I don't know enough Lean to give a concrete suggestion :(

Strata/Transform/Specification.lean
```diff
@@ -0,0 +1,703 @@
...
+
+2. **`Hoare.Triple` (Hoare-triple-based)** — a partial-correctness triple
+   `{Pre} s {Post}` holds when, for every `ρ₀` satisfying `Pre` with a
+   well-formed evaluator and no prior failure, if `s` terminates at `ρ'`
+   then `Post ρ'` holds and `hasFailure` is still false.  Since assert
+   failure is recorded in `hasFailure`, the postcondition
+   `ρ'.hasFailure = false` captures that all asserts passed.
+
+The two are shown equivalent by `hoareTriple_implies_assertValid` and
+`assertValid_implies_hoareTriple`.
```
joscoh, line 38: Very nice!

Strata/Transform/Specification.lean
```diff
@@ -0,0 +1,703 @@
...
+
+2. **`Overapproximates`** — states that the set of reachable terminal/exiting
+   environments in the source is a subset of those reachable in the target.
+   This is a semantic simulation condition.
+
+Both predicates are *bilingual*: they relate two (possibly different) `Lang P`
+values, so they can express cross-language transformations such as
+deterministic-to-nondeterministic.
+
+It is proven that both specifications imply `AssertValid`:
```
joscoh: I assume this means they both imply `AssertValid` of the original program?

Strata/Transform/Specification.lean
```diff
@@ -0,0 +1,703 @@
...
+    (h₁ : Triple (Lang.imperative P CmdT evalCmd extendEval isAtAssertFn) Pre s Mid)
+    (h₂ : TripleBlock (Lang.imperative P CmdT evalCmd extendEval isAtAssertFn) Mid ss Post)
+    (hnofd : Stmt.noFuncDecl s = true)
+    (hnoesc : Stmt.exitsCoveredByBlocks [] s) :
+    TripleBlock (Lang.imperative P CmdT evalCmd extendEval isAtAssertFn) Pre (s :: ss) Post := by
+  intro ρ₀ ρ' hpre hwfb hf₀ hdone
+  have hwfb_preserved : ∀ ρ₁, StepStmtStar P evalCmd extendEval (.stmt s ρ₀) (.terminal ρ₁) →
+      WellFormedSemanticEvalBool ρ₁.eval := by
+    intro ρ₁ hterm
+    have := smallStep_noFuncDecl_preserves_eval P evalCmd extendEval s ρ₀ ρ₁ hnofd hterm
```
joscoh: Nit: should name hypotheses instead of referring to `this` IMO

Strata/Transform/Specification.lean
```diff
@@ -0,0 +1,703 @@
...
+end StandardConnection
+
+end Hoare
+
+
+namespace Transform
+
+/-- A transformation is *sound* if it preserves assertion validity.
+    Bilingual: source and target may live in different languages. -/
+def Sound (L₁ L₂ : Lang P) (T : L₁.StmtT → Option L₂.StmtT) : Prop :=
```
joscoh, line 501: This definition does not handle monadic transforms (e.g. those in `CoreTransformM`). Is that intentional?

Strata/Transform/Specification.lean
```diff
@@ -0,0 +1,703 @@
...
+
+omit [HasVal P] in
+/-- Empty block is skip. -/
+theorem skip (l : String) (md : MetaData P) (Pre : Env P → Prop) :
+    Triple (Lang.imperative P CmdT evalCmd extendEval isAtAssertFn) Pre (.block l [] md) Pre :=
+  (skip_block evalCmd extendEval isAtAssertFn Pre).toTriple
+
+omit [HasVal P] in
+/-- If-then-else rule. -/
+theorem ite {c : P.Expr} {tss ess : List (Stmt P CmdT)} {md : MetaData P}
```
joscoh, line 311: Why don't you need a rule for loops?

Strata/Transform/Specification.lean
```diff
@@ -0,0 +1,703 @@
...
+    cases hexit with
+    | step _ _ _ h _ => cases h with
+      | step_stmts_nil => rename_i r; cases r with | step _ _ _ h _ => cases h
+
+omit [HasVal P] in
+/-- A single command. -/
+theorem cmd (c : CmdT) (Pre Post : Env P → Prop)
+    (h : ∀ ρ₀ σ' f, Pre ρ₀ → WellFormedSemanticEvalBool ρ₀.eval → ρ₀.hasFailure = false →
+      evalCmd ρ₀.eval ρ₀.store c σ' f →
+      Post { ρ₀ with store := σ', hasFailure := f } ∧ f = false) :
```
joscoh, line 209: Not necessary for this PR, but it seems like we should extend the notion of hoare triples to commands. The command/statement distinction is pretty arbitrary, and usually there are specific proof rules for variable assignments, `havoc`, procedure calls, etc.

Strata/Transform/DetToNondetCorrect.lean
```diff
@@ -6,284 +6,526 @@
...
+theorem nondet_seq_terminal
+    (s1 s2 : NondetStmt P (Cmd P)) (ρ ρ₁ ρ' : Env P)
+    (h1 : StepNondetStar P (EvalCmd P) (.stmt s1 ρ) (.terminal ρ₁))
+    (h2 : StepNondetStar P (EvalCmd P) (.stmt s2 ρ₁) (.terminal ρ')) :
+    StepNondetStar P (EvalCmd P) (.stmt (.seq s1 s2) ρ) (.terminal ρ') :=
+  .step _ _ _ .step_seq (reflTrans_trans
+    (reflTrans_trans (nondet_seq_inner_star _ _ s2 h1)
+      (.step _ _ _ .step_seq_done (.refl _))) h2)
+
+omit [HasFvar P] [HasVal P][HasBool P] [HasNot P] in
```
joscoh, line 74: Would be nice to not need all these `omit ... in` if possible

Strata/Transform/DetToNondetCorrect.lean
```diff
@@ -6,284 +6,526 @@
...
+    have ⟨ρ₂, h_rest', h_s, hlen₂⟩ := ih ρ₁ h_rest hcov.2
+    exact ⟨ρ₂,
+      reflTrans_trans
+        (stmts_cons_step P (EvalCmd P) extendEval s' rest' ρ₀ ρ₁ (reflTransT_to_prop h_s'))
+        h_rest',
+      h_s, by omega⟩
+
+/-! ## Loop simulation -/
+
+private noncomputable def loop_sim
```
joscoh: Could you write a comment explaining what this function is doing? Is this effectively a proof that if a loop steps (star) to a terminal value, then there is a proof that this nondeterministic loop also steps to this value? I guess mostly I am confused why you need the `ReflTransT` in `Type` when this function returns a `Prop`-valued expression.

Strata/Transform/DetToNondetCorrect.lean
```diff
@@ -6,284 +6,526 @@
...
+          (.loop (.seq (.cmd (.assume "guard" g md)) b)) h_iter)
+        (.step _ _ _ .step_seq_done nondet_loop))
+  termination_by hstarT.len
+  decreasing_by
+    simp_all [ReflTransT.len]
+    omega
+
+/-! ## Core simulation by strong induction on statement/block size -/
+
+private theorem simulation
```
joscoh, line 322: I think this could be proven directly via mutual recursion on `Stmt` and `Block` (like in `EvalStmtRefinesContract` and `EvalBlockRefinesContract`) which would avoid the need for strong induction. But this is a matter of style.

Strata/DL/Imperative/NondetStmtSemantics.lean
aqjune-aws: It is a style of this definition of configuration (same for configuration of Imperative.Stmt). When a sequence of statements `s1; s2` is met:
- The initial configuration is `.stmt [s1; s2] \empty`
- After the first small-step, the next configuration focuses on `s1` and leaves `s2` as the next statement to run. This is represented as `.seq (.stmt s1 \empty) s2`.
- After several small-steps, `s1` is fully evaluated, and the output configuration is `.seq (.terminal <some Env>) s2`.
- Finally, the `step_seq_done` rule shifts this to `.stmt s2 <some Env>`.

An alternative style of this is to define a program counter, make configuration keep the current program counter and extract the statement to know what's coming next. It seems either design has its benefit, but this sequencing style seems more lightweight.

Strata/DL/Imperative/SemanticsProps.lean
aqjune-aws: Eventually we will remove big-step semantics. The updates were to show that this patch does not at least break existing theorems for big-step, especially the new introduction of `Env` data structure and `hasFailure` flag in Cmd.assert could have affected these :)

Strata/DL/Imperative/StmtSemanticsSmallStep.lean
aqjune-aws: It seems Claude couldn't see this existing lemma. Will fix in the upcoming commit

Strata/DL/Imperative/StmtSemanticsSmallStep.lean
aqjune-aws: Let me try some more tweaks... I think casing (hstep, c_2) at once might be easier to simplify.

Strata/Transform/Specification.lean
aqjune-aws: Yes

Strata/Transform/Specification.lean
aqjune-aws: It was intentional because `CoreTransformM` was too specific to `Core`. Since the AST of `NondetStmt` is a different type from `Stmt`, `CoreTransformM`could not be used to describe `DetToNondet` pass.

Strata/Transform/Specification.lean
aqjune-aws: It became quite lengthy so I chose to omit in this PR.... I left TODO but it was removed during edits. Next commit will have that

Strata/Transform/Specification.lean
aqjune-aws: Indeed, it will be pretty nice.

Strata/Transform/Specification.lean
joscoh: Does this mean that we will need a separate definition of soundness for monadic transformations?

Strata/Transform/DetToNondetCorrect.lean
```diff
@@ -6,284 +6,530 @@
...
+    | s :: rest =>
+      have ⟨s', r', hs, hr, _⟩ := block_transform_some s rest ns ht
+      simp [Block.noFuncDecl]
+      exact ⟨stmtToNondet_some_noFuncDecl s s' hs, blockHelper rest r' hr⟩
+
+/-! ## Block exitsCovered from successful transform (empty labels) -/
+
+ omit [HasFvar P] [HasVal P] [HasBoolVal P] in
+private theorem blockToNondet_some_exitsCovered_nil
+    (bss : List (Stmt P (Cmd P))) (ns : NondetStmt P (Cmd P))
```
MikaelMayer: `blockToNondet_some_exitsCovered_nil` is identical to `stmtToNondet_some_exitsCovered.blockHelper` instantiated with `labels := []`. Consider removing this and using the `where` helper directly, e.g. `stmtToNondet_some_exitsCovered.blockHelper [] bss ns ht`.

Strata/Transform/DetToNondetCorrect.lean
```diff
@@ -6,284 +6,530 @@
...
+  | _, _ => False
+
+abbrev Lang.nondet : Lang P where
+  StmtT := NondetStmt P (Cmd P)
+  CfgT := NondetConfig P (Cmd P)
+  star := StepNondetStar P (EvalCmd P)
+  stmtCfg := .stmt
+  stmtsCfg := fun ss ρ => match ss with
+    | [] => .terminal ρ
+    | s :: _ => .stmt (ss.foldl (init := s) fun acc s' => .seq acc s') ρ
```
MikaelMayer: The `stmtsCfg` implementation double-counts the first element: for `ss = [a, b, c]`, `foldl (init := s) f [a, b, c]` produces `seq (seq (seq a a) b) c` since `s = a` and `a` is also the first element of the list being folded. This isn't exercised by any current proof, but it would be incorrect if `TripleBlock` were ever used with `Lang.nondet`. A correct version would be `| s :: rest => .stmt (rest.foldl (init := s) fun acc s' => .seq acc s') ρ`.

Strata/Transform/Specification.lean
```diff
@@ -0,0 +1,705 @@
...
+    (ht : TripleBlock (Lang.imperative P CmdT evalCmd extendEval isAtAssertFn) (fun ρ => Pre ρ ∧ ρ.eval ρ.store c = some HasBool.tt) tss Post)
+    (he : TripleBlock (Lang.imperative P CmdT evalCmd extendEval isAtAssertFn) (fun ρ => Pre ρ ∧ ρ.eval ρ.store c = some HasBool.ff) ess Post) :
+    Triple (Lang.imperative P CmdT evalCmd extendEval isAtAssertFn) Pre (.ite c tss ess md) Post := by
+  intro ρ₀ ρ' hpre hwfb hf₀ hstar
+  cases hstar with
+  | step _ _ _ h1 r1 => cases h1 with
+    | step_ite_true hc _ => exact ht ρ₀ ρ' ⟨hpre, hc⟩ hwfb hf₀ (.inl r1)
+    | step_ite_false hc _ => exact he ρ₀ ρ' ⟨hpre, hc⟩ hwfb hf₀ (.inl r1)
+
+/- TODO: the WHILE rule -/
```
MikaelMayer, line 322: Noting the TODO: the WHILE Hoare rule would complete the structural rule set. Not blocking, just flagging for visibility.

Strata/Transform/DetToNondet.lean
```diff
@@ -10,39 +10,48 @@ public import Strata.DL.Imperative.Stmt
...
+/-- Deterministic-to-nondeterministic transformation for a block.
+    Returns `none` if any statement is unsupported. -/
 def BlockToNondetStmt {P : Imperative.PureExpr} [Imperative.HasBool P] [HasNot P]
   (ss : Imperative.Block P (Cmd P)) :
-  Imperative.NondetStmt P (Cmd P) :=
+  Option (Imperative.NondetStmt P (Cmd P)) :=
   match ss with
-  | [] => (.assume "skip" Imperative.HasBool.tt .empty)
-  | s :: ss => .seq (StmtToNondetStmt s) (BlockToNondetStmt ss)
+  | [] => some (.assert "skip" Imperative.HasBool.tt .empty)
```
MikaelMayer: The label `"skip"` could collide with a user-defined assert label, making `isAtNondetAssert` match unexpectedly. Consider using a label that cannot collide with user labels (e.g. a reserved prefix).

Strata/DL/Util/Relations.lean
```diff
@@ -10,7 +10,7 @@ section Relation
 
 @[expose] def Relation (A: Type) := A → A → Prop
 def Reflexive (r: Relation A) : Prop := ∀ x, r x x
-def Transitive (r: Relation A) : Prop := ∀ x y z, r x y → r y z → r x z
+abbrev Transitive (r: Relation A) : Prop := ∀ x y z, r x y → r y z → r x z
```
MikaelMayer, line 13: No action needed: Good change from `def` to `abbrev` for `Transitive` — this avoids the opacity issue that required the now-removed `reflTrans_trans` workaround in `StmtSemanticsSmallStep.lean`.

Strata/Transform/DetToNondetCorrect.lean
aqjune-aws: Makes sense, thanks - I digged further and actually the field `stmtsCfg` was only used in Hoare rules for Imperative.Stmt, which didn't apply to NondetStmt! I removed this `stmtsCfg` field.

Strata/Transform/Specification.lean
aqjune-aws: For Core-specific one, yes - I think we need a statement about CachedAnalyses. :)

PR #663, opened 2026-03-25
Pass source metadata to subexpressions in Python/Specs/ToLaurel

andrewmwells-amazon:
Pass source metadata to subexpressions in Python/Specs/ToLaurel

### DIFF
--- a/Strata/Languages/Python/Specs/ToLaurel.lean
+++ b/Strata/Languages/Python/Specs/ToLaurel.lean
@@ -249,7 +249,7 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
 
 /-! ## SpecExpr to Laurel Translation -/
 
-/-- Wrap a StmtExpr with metadata (empty by default). -/
+/-- Wrap a StmtExpr with metadata. -/
 private def mkStmt (e : StmtExpr) (md : Imperative.MetaData Core.Expression) : StmtExprMd :=
   { val := e, md := md }
 


Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -249,9 +249,17 @@ def specTypeToLaurelType (ty : SpecType) : ToLaurelM HighTypeMd := do
 
 /-! ## SpecExpr to Laurel Translation -/
 
-/-- Wrap a StmtExpr with default metadata. -/
-private def mkStmt (e : StmtExpr) : StmtExprMd :=
-  { val := e, md := .empty }
+/-- Wrap a StmtExpr with metadata (empty by default). -/
+private def mkStmt (e : StmtExpr) (md : Imperative.MetaData Core.Expression := .empty) : StmtExprMd :=
```
joehendrix: Is having a default ok here?

Strata/Languages/Python/PythonToCore.lean
```diff
@@ -877,7 +879,7 @@ def pythonToCore (signatures : Python.Signatures) (insideMod : Array (Python.stm
   | .ClassDef _ _ _ _ _ _ _ => false
   | _ => true)
 
-  let globals := [(.var "__name__" (.forAll [] mty[string]) (some (.strConst () "__main__")))]
+  let globals := [(.var "__name__" (.forAll [] mty[string]) (some (.strConst () "__main__")) .empty)]
```
joehendrix, line 882: I'm also wondering about the .empty here.  We may want comments or dedicated combinators.

Strata/Languages/Python/Specs/ToLaurel.lean
aqjune-aws: Let me fix this

Strata/Languages/Python/PythonToCore.lean
aqjune-aws: PythonToCore will be replaced with Python -> Laurel -> Core, so let's keep as it is.

shigoel (dismissed):

Strata/Languages/Python/Specs/ToLaurel.lean
aqjune-aws: Fixed :) (even if it became explicitly `.empty` from the caller side).

aqjune-aws (approved):

keyboardDrummer:
Love this!

PR #664, opened 2026-03-25
Instance call support: @ separator, void-returns-Any, hybrid field type system, and class method emission

joehendrix:
## Summary

Three interrelated changes to the Python→Laurel→Core pipeline that enable correct handling of instance methods, void procedures, and user-defined class fields:

1. **Unambiguous `@` separator for instance methods** — `Storage@put_item` instead of `Storage_put_item`, eliminating collisions with underscored names
2. **Void procedures return Any** — all PySpec procedures now have exactly one output (void returns `Any` via `from_none`), removing special-case void handling at call sites
3. **Hybrid two-kind field type system** — builtin types (int, str, bool, float) become `Any` on Composite fields; class types (PySpec services, user classes) stay `Composite`, matching parameter semantics at call boundaries
4. **User-defined class method emission** — class method definitions (`Counter@__init__`, `Counter@increment`) are now included in the Laurel program as static procedures

## PythonToLaurel changes

- `@` separator applied across PythonToLaurel, PythonToCore, PySpecPipeline, and Specs/ToLaurel
- New `translateFieldType` helper maps builtins→Any, composites→Composite for field declarations
- `self` parameter typed as `Composite` (not `Any`) in `__init__` and user methods
- Removed `wrapFieldInAny` for `self.field` reads in method bodies — builtin fields are already Any-typed
- `unsupportedConstruct` error thrown when Composite→Any coercion is attempted (e.g., returning a Storage field as str)

## LaurelToCoreTranslator changes

- Statement-position calls now synthesize throwaway `$unused_N` variables to match output count, fixing Core arity checks for void-returns-Any procedures
- Single-target `Assign` with `StaticCall` also synthesizes extra LHS variables for multi-output procedures

## Testing

- New dispatch tests for void assign/init, non-void discard, instance call result, user class construct, class field assign, and class method return
- Two negative tests for Composite↔Any kind mismatches: `test_class_composite_as_any` (Composite field returned as Any) and `test_class_any_as_composite` (Any value assigned to Composite field). The latter requires explicit pyspec paths since `IdentifyOverloads` resolves modules from complete programs with `main`, not isolated class snippets.
- 7 regenerated expected_laurel files reflecting the new field type system

## Known limitations

- Composite→Any coercion (e.g., returning a `Storage` field as `str`) is explicitly rejected with a user-facing error. Coercion functions are being developed separately.
- Procedure names used as type annotations (e.g., `Storage_put_item` as a type) silently fall through to `Any` instead of producing an error or warning.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT licenses.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -160,10 +160,12 @@ def mkStmtExprMd (expr : StmtExpr) : StmtExprMd :=
 def mkStmtExprMdWithLoc (expr : StmtExpr) (md : Imperative.MetaData Core.Expression) : StmtExprMd :=
   { val := expr, md := md }
 
-/-- Build a StaticCall for an instance method: ClassName_methodName(self, args...) -/
+/-- Build a StaticCall for an instance method: ClassName@methodName(self, args...).
+    For Any-typed receivers (no model available), returns a Hole instead. -/
 def mkInstanceMethodCall (className : String) (methodName : String)
     (self : StmtExprMd) (args : List StmtExprMd) : StmtExprMd :=
-  mkStmtExprMd (StmtExpr.StaticCall (className ++ "@" ++ methodName) (self :: args))
+  if className == "Any" then mkStmtExprMd .Hole
+  else mkStmtExprMd (StmtExpr.StaticCall (className ++ "@" ++ methodName) (self :: args))
 
 /-- Extract string representation from Python expression (for type annotations) -/
 partial def pyExprToString (e : Python.expr SourceRange) : String :=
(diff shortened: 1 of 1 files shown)


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -369,7 +376,9 @@ def translateStmt (outputParams : List Parameter) (stmt : StmtExprMd)
             let coreArgs ← args.mapM (fun a => translateExpr a)
             let defaultExpr := defaultExprForType model ty
             let initStmt := Core.Statement.init ident coreType (some defaultExpr) md
-            let callStmt := Core.Statement.call [ident] callee.text coreArgs md
+            -- Void procedures have no outputs; emit call without assignment target
+            let lhs := if isVoidProcedure model callee then [] else [ident]
+            let callStmt := Core.Statement.call lhs callee.text coreArgs md
```
keyboardDrummer: I think this is adding code to handle what is already a typing error, namely a program `ident := callThatReturnsNothing(..)`

Why was the faulty code produced to begin with?

joehendrix: Agreed.  This PR was accidentally marked as non-draft when it was extracted out of #613 to make smaller PRs.  I'll remove these changes and work upstream to fix PythonToLaurel.

PR #666, opened 2026-03-26
feat: parametric ResultClassifier for pyAnalyzeLaurel success/failure

shigoel:
## Summary

- Add `ResultClassifier` structure with pluggable `isSuccess`/`isFailure` predicates (defaults preserve existing behavior) so callers can control what counts as a failure without touching `printPyAnalyzeSummary`.
- In bug-finding mode, narrow `isFailure` to `alwaysFalseAndReachable` only — the outcome where validity confirms the property is always false on a reachable path. Other failure modes become inconclusive.
- Apply the classifier consistently: both the per-VC `"Assertion failed"` prefix and the summary counts/exit-code now use the same `classifier.isFailure`.
- `nInconclusive` is now the remainder (total minus success, failure, unreachable, impl-error), so narrowing `isFailure` automatically widens inconclusive.
- Add `VCOutcome.hasSMTError` / `VCResult.hasSMTError` to `Verifier.lean` alongside the other nine-case predicates, and use them to simplify `nImplError` in `printPyAnalyzeSummary`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,7 +238,7 @@ jobs:
               sudo cp z3-4.13.4-x64-glibc-2.35/bin/z3 /usr/local/bin/
             fi
     - name: Run PySpec and dispatch tests
-      run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest StrataTest.Languages.Python.AnalyzeLaurelTest StrataTest.Languages.Python.Specs.IdentifyOverloadsTest
+      run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest StrataTest.Languages.Python.AnalyzeLaurelTest StrataTest.Languages.Python.Specs.IdentifyOverloadsTest StrataTest.Languages.Python.VerifyPythonTest
     - name: Run test script
       run: FAIL_FAST=1 ./scripts/run_cpython_tests.sh ${{ matrix.python_version }}
       working-directory: Tools/Python
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -162,7 +162,8 @@ private partial def coreStmtsToGoto
                 .Integer
             CProverGOTO.Expr.symbol name ty
           | [] => CProverGOTO.Expr.symbol "" .Empty
-        let calleeExpr := CProverGOTO.Expr.symbol procName .Empty
+        let calleeExpr := CProverGOTO.Expr.symbol procName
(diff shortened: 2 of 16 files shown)


StrataMain.lean
```diff
@@ -400,29 +410,24 @@ private def exitPyAnalyzeKnownLimitation {α} (message : String) : IO α := do
...
-                 | .err _, _ | _, .err _ => true
-                 | _,      _             => false
-      | _     => false)                                            |>.size
+  let nImplError   := vcResults.filter (fun r => r.isImplementationError ||
+                                                 r.hasSMTError)          |>.size
+  -- Inconclusive is the remainder: everything not classified above.
+  -- This means narrowing `isFailure` in the classifier automatically widens
+  -- inconclusive without requiring changes here.
+  let nInconclusive := vcResults.size - nSuccess - nFailure
+                                      - nUnreachable - nImplError
```
MikaelMayer: With `Nat` subtraction, if the classifier's `isSuccess`/`isFailure` overlap with `isUnreachable` or `hasSMTError || isImplementationError`, this silently truncates to 0 instead of going negative. The overlap check on line 430 catches it, but the error message says "Overlapping VC result predicates" — which is accurate for the new remainder-based computation but could be confusing since the overlap might come from a custom classifier rather than the built-in predicates.

Note: the default `isSuccess` (`VCResult.isSuccess` → `VCOutcome.isPass`) returns true for `(unsat, unsat)` which is also `unreachable`. This overlap exists in the old code too and doesn't trigger in practice, but the remainder-based computation makes it worth documenting as a precondition on `ResultClassifier`: its `isSuccess` and `isFailure` must be disjoint from `isUnreachable` and `isImplementationError || hasSMTError`.

StrataMain.lean
```diff
@@ -377,6 +377,16 @@ All output uses two structured lines on stdout:
...
 A successful run exits 0 with `RESULT: Analysis success` or `RESULT: Inconclusive`. -/
 
+/-- Determines which VC results count as successes and which count as failures
+    for the purposes of the `pyAnalyzeLaurel` summary and exit code.
+    Unreachable and implementation-error results are always handled separately
+    and are not affected by this classifier.
+    `nInconclusive` is computed as the remainder, so narrowing `isFailure`
+    (e.g. to only `alwaysFalseAndReachable`) automatically widens inconclusive. -/
+structure ResultClassifier where
+  isSuccess : Core.VCResult → Bool := (·.isSuccess)
```
MikaelMayer, line 388: [No action needed] Good docstring — clearly states the contract that `nInconclusive` is a remainder and that unreachable/impl-error are handled separately.

StrataMain.lean
MikaelMayer: I would suggest doing the classification not as a premature count, but as successive list partitioning, so that we are guaranteed no overlap.

shigoel: Addressed (hopefully).

PR #668, opened 2026-03-26
feat: support postconditions on Laurel functions

fabiomadge:
Adds support for function postconditions (both `ensures` clauses and constrained return types) via postcondition axioms and a new `OrderedProgram` intermediate representation.

## Approach

Function postconditions are made available to callers via **axioms** rather than assume statements. This means postconditions work everywhere — including inside function bodies and quantifiers — without any call-site lifting.

A `$check` procedure is generated per function with postconditions to verify the body satisfies the postcondition.

## Pipeline

Introduces `OrderedProgram` as an intermediate IR between `Laurel.Program` and `Core.Program`:

```
Program → toLaurelOrdered → generateAxioms → orderedToCore → Core.Program
```

- `toLaurelOrdered` — structural analysis: SCC ordering, grouping, diagnostics
- `generateAxioms` — populates postcondition and invokeOn axioms (can be skipped for a future testing mode)
- `orderedToCore` — thin translation pass, no structural logic

## Grammar

- `function f(): int ensures result > 0 { ... }` — transparent, body visible to callers
- `opaque function f(): int ensures result > 0 { ... }` — body hidden, callers only see postcondition via axiom
- `function f(): nat { ... }` — constrained return type becomes a postcondition

## AST change

- `Body.Transparent` now carries a postconditions list

## Key changes

- `OrderedProgram.lean` / `ToOrderedProgram.lean` — new IR and structural analysis pass
- `FunctionPostcondCheck.lean` — generates `$check` procedures (replaces old `FunctionPostcondElim`)
- `DeclGrouping.lean` — renamed from `CoreGroupingAndOrdering`, no longer depends on Core types
- `LaurelToCoreTranslator.lean` — `orderedToCore` replaces monolithic `translateLaurelToCore`; `op` nodes annotated with function types at creation
- All Laurel passes: preserve new `Transparent` postconditions field
- Tests: new cases for opaque-calls-opaque, quantifier postconds, result shadowing

## Known trade-off

Opaque function callers get "could not be proved" instead of counterexamples, since the symbolic evaluator cannot instantiate axioms as ground facts. The right fix is in the Core evaluator, which would benefit all frontends.


keyboardDrummer (changes requested):
Looking at the implementation:
> 5. Inserting `assume f$post(args, result)` at call sites

I don't think that translation is executable. In general I think Laurel should not generate axioms or assume statements.

I think it's better if function postconditions are supported at the Core level, that way the Laurel->Core translation can remain back-end independent, and also then Core can handle everything related to postconditions, both for procedures and functions. There's an AWS internal ticket for adding them to Core.

fabiomadge:
> When the back-end has been decided, which is after translating to Core, I think that is the right thing to do.

As of now, Core doesn't support postconditions on functions and there is no concrete plan on how to add them.

> Why not do the right thing right away? Why would it be easier to add them to Laurel than to Core? I think adding it to Laurel like you did it now breaks our executability promise.

The kind of statement lifting this PR does could also work as an early Core pass, but I think it would be more consistent with the rest of the Core implementation to hold onto them, which makes it harder. I'm not sure how assuming something we've proven elsewhere is breaking any executability promises. I'd expect someone trying to execute the final Core program to ignore these assumptions, just like they would the postconditions.

> Do you have a particular end-target in mind?

You mean what I would like to enable? Constrained types as return values in functions. You can see examples in the tests I added.

PR #672, opened 2026-03-26
Fix bug in handling of blocks that occur in expresssions

keyboardDrummer:
### Changes
- Fix bug in handling of blocks that occur in expressions. The heart of the problem was that `transformStmt`, which does not update the substitution map, was being called from inside `transformExpr` when it was handling a block, leading to missing substitutions. A second but smaller bug was a statement ordering bug that occurred when handling calls.

### Testing
- Added test-case


Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -104,9 +104,42 @@ private def freshCondVar : LiftM Identifier := do
...
-private def addPrepend (stmt : StmtExprMd) : LiftM Unit :=
+private def prepend (stmt : StmtExprMd) : LiftM Unit :=
   modify fun s => { s with prependedStmts := stmt :: s.prependedStmts }
 
+private def onlyKeepSideEffectStmtsAndLast (stmts : List StmtExprMd) : LiftM (List StmtExprMd) := do
+  match stmts with
+  | [] => return []
+  | _ =>
+    let last := stmts.getLast!
+    let nonLast ← stmts.dropLast.flatMapM (fun s =>
```
joscoh, line 115: Doesn't every case return `pure []`? Why return a list at all?

keyboardDrummer: True but it's only because of the hacks, otherwise the localvars, assert/assumes would fill the list.

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
```diff
@@ -97,6 +97,36 @@ procedure imperativeCallInConditionalExpression(b: bool) {
...
+
+procedure addProcCaller(): int {
+  var x: int := 0;
+  var y: int := addProc({x := 1; x}, {x := x + 10; x});
+  assert y == 11
+
+  // The next statement is not translated correctly.
+  // I think it's a bug in the handling of StaticCall
+  // Where a reference is substituted when it should not be
+  // var z: int := addProc({x := 1; x}, {x := x + 10; x}) + (x := 3);
```
olivier-aws, line 127: Is the bug only affecting the case `addProc`, or also with `add`?

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
```diff
@@ -22,16 +22,6 @@ open Strata.Elab (parseStrataProgramFromDialect)
...
-composite Box {
-  var value: int
-}
-
-procedure heapUpdateInBlockExpr(b: Box)
-{
-  var x: int := { b#value := b#value + 1; b#value };
-  assert x == b#value
-};
-
```
olivier-aws, line 34: Why removing this test?

keyboardDrummer: This file only applies a single phase from the Laurel pipeline, and that phase does not support the program in this test, since it uses field access which should've already been lowered at that point. Not sure how it was working before.

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
keyboardDrummer: ~I think anything with `.StaticCall`, so both~ I'm not sure

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -104,9 +104,42 @@ private def freshCondVar : LiftM Identifier := do
...
+      match s.val with
+      | .LocalVariable .. => do
+          -- This addPrepend is a hack to work around Core not having let expressions
```
MikaelMayer, line 118: Stale reference: this comment says `addPrepend` but the function was renamed to `prepend` in this PR.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -104,9 +104,42 @@ private def freshCondVar : LiftM Identifier := do
...
+
+      /-
+      Any other impure StmtExpr, like .Assign, .Exit or .Return,
+      should already have been processed by translateExpr,
+      so we can assume this StmtExpr is pure and can be dropped.
+      TODO: currently .Exit and .Return are not processed by translateExpr, this is a bug
+      -/
+      | _ => pure []
+    )
+    return nonLast ++ [last]
```
MikaelMayer, line 141: Since every branch of the `flatMapM` returns `pure []`, `nonLast` is always `[]`, making the return value always `[last]`. The `nonLast ++` part is dead code. Consider simplifying to just iterate with `forM` (for the prepend side effects) and return `[last]`, which would make the intent clearer.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -235,12 +268,13 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
-      -- so that when reversed the var declaration comes before the call.
       let callResultVar ← freshCondVar
       let callResultType ← computeType expr
-      addPrepend (⟨.Assign [bare (.Identifier callResultVar)] seqCall, md⟩)
-      addPrepend (bare (.LocalVariable callResultVar callResultType none))
+      let liftedCall := [
+        ⟨ (.LocalVariable callResultVar callResultType none), md ⟩,
+        ⟨.Assign [bare (.Identifier callResultVar)] seqCall, md⟩
+      ]
+      modify fun s => { s with prependedStmts := s.prependedStmts ++ liftedCall}
```
MikaelMayer, line 277: This uses `s.prependedStmts ++ liftedCall` (append to end) while every other call site uses `prepend` (cons to front). The `prependedStmts` field is documented as "in reverse order — newest first", so appending to the end breaks that invariant. It works because `takePrepends` consumers don't rely on the reverse-order invariant uniformly, but this is fragile. A comment explaining why append is used here (the call must come after all arg-processing prepends) would help future readers.

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
```diff
@@ -22,16 +22,6 @@ open Strata.Elab (parseStrataProgramFromDialect)
 namespace Strata.Laurel
 
 def blockStmtLiftingProgram : String := r"
```
MikaelMayer, line 24: The `heapUpdateInBlockExpr` test was removed. This tested `PureFieldUpdate` (`b#value := ...`) inside a block expression, which is a scenario the new `transformExpr`-based block handling doesn't explicitly cover (`PureFieldUpdate` falls through to `| _ => return expr`). Was this removed because heap field updates in block expressions are no longer supported, or because the test needs to be rewritten? If the latter, it would be good to keep it (possibly updated) to avoid losing coverage.

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
```diff
@@ -97,6 +97,36 @@ procedure imperativeCallInConditionalExpression(b: bool) {
...
+};
+
+procedure addProcCaller(): int {
+  var x: int := 0;
+  var y: int := addProc({x := 1; x}, {x := x + 10; x});
+  assert y == 11
+
+  // The next statement is not translated correctly.
+  // I think it's a bug in the handling of StaticCall
+  // Where a reference is substituted when it should not be
```
MikaelMayer, line 126: [No action needed] The commented-out test with the TODO about a known `StaticCall` substitution bug is useful documentation of a known limitation. Good to have it here as a reminder.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
keyboardDrummer: Previous comment was similar: https://github.com/strata-org/Strata/pull/672#discussion_r2996425653

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
keyboardDrummer: Previous comment was similar: https://github.com/strata-org/Strata/pull/672#discussion_r3000846825

Strata/Languages/Laurel/LiftImperativeExpressions.lean
keyboardDrummer: Do you really want me to add a comment saying that the lifted call statement should come after the statements related to its arguments? Can this be a [nit] ?

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
keyboardDrummer: Previous comment was similar: https://github.com/strata-org/Strata/pull/672#discussion_r3000846825

Strata/Languages/Laurel/LiftImperativeExpressions.lean
keyboardDrummer: Can this be a [nit]?

PR #673, opened 2026-03-26
Make pyAnalyzeLaurel only check non-prelude asserts.

andrewmwells-amazon:
- Make pyAnalyzeLaurel only check non-prelude asserts
- Add check of prelude asserts as build-time check
- Add check for missing source location to CI

### DIFF
--- a/Strata/Languages/Python/PySpecPipeline.lean
+++ b/Strata/Languages/Python/PySpecPipeline.lean
@@ -342,6 +342,13 @@ public def translateCombinedLaurel (combined : Laurel.Program)
   let (coreOption, errors) := Laurel.translate { emitResolutionErrors := false } combined
   (coreOption.map prependPrelude, errors)
 
+/-- Like `translateCombinedLaurel` but also returns the lowered Laurel program
+    (after all Laurel-to-Laurel passes, before translation to Core). -/
+public def translateCombinedLaurelWithLowered (combined : Laurel.Program)
+    : (Option Core.Program × List DiagnosticModel × Laurel.Program) :=
+  let (coreOption, errors, lowered) := Laurel.translateWithLaurel { emitResolutionErrors := false } combined
+  (coreOption.map prependPrelude, errors, lowered)
+
 /-- Errors from the pyAnalyzeLaurel pipeline. -/
 public inductive PipelineError where
   /-- The Python source contains invalid code (bad method name, wrong arguments, etc.). -/
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -6,7 +6,6 @@
 
 -- Executable with utilities for working with Strata files.
 import Strata.DDM.Integration.Java.Gen
-import Strata.Languages.Core.Options
 import Strata.Languages.Core.SarifOutput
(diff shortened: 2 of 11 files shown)


StrataMain.lean
```diff
@@ -523,6 +523,16 @@ def pyAnalyzeLaurelCommand : Command where
           pure inlined
       else pure coreProgram
 
+    -- Collect user procedure names (those after FIRST_END_MARKER) for selective verification
```
joscoh: Is there a way to avoid repeating this twice (in `AnalyzeLaurelTest` and here)? Especially asking because it has a fixed string name in it.

StrataMain.lean
```diff
@@ -534,11 +544,26 @@ def pyAnalyzeLaurelCommand : Command where
...
       | .none => baseOptions
-    let vcResults ←
+    let allVcResults ←
       match ← Strata.verifyCore coreProgram options
-                (moreFns := Strata.Python.ReFactory) |>.toBaseIO with
+                (moreFns := Strata.Python.ReFactory)
+                (proceduresToVerify := some userProcNames) |>.toBaseIO with
       | .ok r => pure r
       | .error msg => exitPyAnalyzeInternalError msg
+    -- Filter out prelude VCs (those with empty file path in metadata).
```
joscoh: This feels like a little bit of a hack, is it possible for someone to create a procedure without metadata separately?

andrewmwells-amazon: Filtering these out wasn't sound, we need to verify preconditions that transitively come from the prelude.

StrataMain.lean
andrewmwells-amazon: Done

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -313,6 +313,21 @@ private def prependPrelude (coreFromLaurel : Core.Program) : Core.Program :=
...
+public def splitProcNames (prog : Core.Program)
+    : Std.HashSet String × List String :=
+  let (preludeDecls, rest) := prog.decls.span (fun d => toString d.name != "FIRST_END_MARKER")
+  let preludeNames := preludeDecls.foldl (init := ({} : Std.HashSet String)) fun s d =>
+    match d.getProc? with
+    | some p => s.insert (Core.CoreIdent.toPretty p.header.name)
+    | none => s
+  -- rest starts with FIRST_END_MARKER (or is empty); skip it
+  let userDecls := match rest with | _ :: tl => tl | [] => []
+  let userProcNames := userDecls.filterMap fun d =>
```
shigoel, line 329: If `FIRST_END_MARKER` is absent from the program, `List.span` puts every declaration into `preludeDecls` and `rest` is `[]`. Then `userProcNames` is `[]`, and the downstream call `proceduresToVerify := some []` tells the verifier to skip all procedures, silently returning "Analysis success" with 0 obligations. The old code was safe here: it also relied on `FIRST_END_MARKER` to identify prelude names for inlining, but `proceduresToVerify` was always `none` (verify all), so a missing marker never caused a vacuous result. Consider returning an error or logging a warning when `rest` is empty but `prog.decls` is non-empty.

PR #674, opened 2026-03-26
Add an option to keep intermediate files

olivier-aws:
*Description of changes:*

Adds the `--keep-all-files dir` to pyAnalyzeLaurel that creates the dir directory and stores a text representation of the Laurel and Core programs in it, plus all the SMT queries.

Manually tested.


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -625,11 +630,12 @@ def translate (options: LaurelTranslateOptions) (program : Program): TranslateRe
   let (program, model) := (result.program, result.model)
 
     let initState : TranslateState := {model := model }
+  let loweredLaurel := program
```
joscoh: This let-binding seems unecessary

StrataMain.lean
```diff
@@ -439,6 +434,14 @@ private def printPyAnalyzeSummary (vcResults : Array Core.VCResult)
   else
     printPyAnalyzeResult (if nInconclusive > 0 then "Inconclusive" else "Analysis success") counts
 
+private def deriveBaseName (file : String) : String :=
+  let name := System.FilePath.fileName file |>.getD file
+  if name.endsWith ".python.st.ion" then (name.dropEnd 14).toString
```
joscoh: I would prefer to see the string names separated into a separate variable and the length calculated to make sure the lengths are consistent.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
MikaelMayer: Removed the binding, now using `program` directly in the return tuple.

StrataMain.lean
MikaelMayer: Refactored to collect suffixes in a list and use `.length` to compute the drop count, so the string and length are always consistent.

PR #676, opened 2026-03-26
Support **dict expansion at resolved method call sites

MikaelMayer:
Support `**dict` expansion at resolved method call sites

Python code often builds a parameter dictionary and passes it to a method
call using `**`, e.g. `client.start_job_run(**params)`. When the target
method had a known signature (via pyspec), the translator would fail with
`'method' called with unknown keyword arguments: []` because it couldn't
match the unnamed `**` keyword to individual named parameters.

This PR detects `**dict` at call sites of resolved methods and expands the
dictionary into individual arguments: required parameters use
`DictStrAny_get` (preserving the precondition that the key must exist), and
optional parameters use a new prelude function `DictStrAny_get_or_none`
(returns `None` if the key is absent).

Two end-to-end tests were added to the dispatch test suite: one covering
all-required parameters, one covering a mix of required and optional
parameters where the dictionary omits the optional keys. All existing tests
continue to pass.

### DIFF
--- a/StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
@@ -35,5 +35,5 @@ assert_assert(567)_calls_Any_to_bool_0: ✅ pass (at line 24, col 4)
 assert(567): ✅ pass (at line 24, col 4)
 ite_cond_calls_Any_to_bool_0: ✅ pass (at line 26, col 0)
 
-DETAIL: 41 passed, 0 failed, 0 inconclusive
+DETAIL: 34 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success
--- a/StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
@@ -29,5 +29,5 @@ assert(124): ✅ pass (at line 7, col 4)
 assert_assert(167)_calls_Any_to_bool_0: ✅ pass (at line 9, col 4)
 assert(167): ✅ pass (at line 9, col 4)
 
-DETAIL: 35 passed, 0 failed, 0 inconclusive
+DETAIL: 28 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success
--- a/StrataTest/Languages/Python/expected_laurel/test_boolean_logic.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_boolean_logic.expected
@@ -51,5 +51,5 @@ assert_assert(825)_calls_Any_to_bool_0: ✅ pass (at line 37, col 4)
 assert(825): ✅ pass (at line 37, col 4)
 ite_cond_calls_Any_to_bool_0: ✅ pass (at line 39, col 0)
 
-DETAIL: 57 passed, 0 failed, 0 inconclusive
+DETAIL: 50 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success
--- a/StrataTest/Languages/Python/expected_laurel/test_break_continue.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_break_continue.expected
@@ -26,5 +26,5 @@ loop_guard_calls_Any_to_bool_0: ✅ pass (at line 3, col 4)
 loop_guard_calls_Any_to_bool_0: ✅ pass (at line 8, col 4)
 loop_guard_end_calls_Any_to_bool_0: ✅ pass (at line 8, col 4)
 
-DETAIL: 32 passed, 0 failed, 0 inconclusive
+DETAIL: 25 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success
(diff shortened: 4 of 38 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -304,6 +304,17 @@ def DictStrAny_empty:= mkStmtExprMd (StmtExpr.StaticCall "DictStrAny_empty" [])
 
 def DictStrAny_mk (kv: List (String × StmtExprMd)) := DictStrAny_mk_aux kv DictStrAny_empty
 
+/-- Generate `DictStrAny_get(dict, key)` to extract a value from a dictionary. -/
+def DictStrAny_get_expr (dict : StmtExprMd) (key : String) : StmtExprMd :=
+  mkStmtExprMd (.StaticCall "DictStrAny_get" [dict, mkStmtExprMd (.LiteralString key)])
```
MikaelMayer: `DictStrAny_get_expr` is unused — `DictStrAny_get_param` with `isOptional = false` produces the same call. Consider removing it to avoid dead code.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -796,6 +807,26 @@ partial def translateCall (ctx : TranslationContext)
...
+    let trans_dictArgs := remainingParams.map fun (argName, _, dflt) =>
+      DictStrAny_get_param trans_dict argName dflt.isSome
+    let allArgs := trans_posArgs ++ trans_dictArgs
+    let kwargsArg := if funcDecl.hasKwargs then [trans_dict] else []
+    let mkCall (name : String) := mkStmtExprMd (StmtExpr.StaticCall name (allArgs ++ kwargsArg))
+    match f with
+    | .Name  _ _ _ => return mkCall funcName
+    | .Attribute _ val _attr _ =>
+        let _target_trans ← translateExpr ctx val
+        if opt_firstarg.isSome then return mkStmtExprMd (.Hole)
```
MikaelMayer: When `opt_firstarg.isSome`, this unconditionally returns `.Hole`. The existing non-`**dict` path (line 845) has a special case that emits a `StaticCall` for transparent procedures (`ImportedSymbol.procedure _ _ true`). That check is missing here, so `**dict` expansion on a transparent procedure method call would silently produce a `.Hole` instead of the expanded call.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -796,6 +804,30 @@ partial def translateCall (ctx : TranslationContext)
...
+    | .Name  _ _ _ => return mkCall funcName
+    | .Attribute _ val _attr _ =>
+        let _target_trans ← translateExpr ctx val
+        if opt_firstarg.isSome then
+          if let some (ImportedSymbol.procedure _ _ true) := ctx.importedSymbols[funcName]? then
+            return mkCall funcName
+          else
+            return mkStmtExprMd (.Hole)
+        else return mkCall funcName
+    | _ => throw (.unsupportedConstruct "Invalid call construct" (toString (repr f)))
```
MikaelMayer, line 820: The `match f with` dispatch block here (lines 819–829) is nearly identical to the one in the `else` branch (lines 840–852) — both do `.Name` → return call, `.Attribute` → translate val, check `opt_firstarg` / `importedSymbols`, return call or `Hole`, `_` → throw. The only difference is the argument list.

Consider extracting a shared helper like:
```
let emitCall (callArgs : List StmtExprMd) := do
  let mkCall (name : String) := mkStmtExprMd (StmtExpr.StaticCall name callArgs)
  match f with
  | .Name _ _ _ => return mkCall funcName
  | .Attribute _ val _ _ =>
      let _ ← translateExpr ctx val
      if opt_firstarg.isSome then
          return mkCall funcName
        else return mkStmtExprMd .Hole
      else return mkCall funcName
  | _ => throw (.unsupportedConstruct "Invalid call construct" (toString (repr f)))
```
Then both branches just compute their argument list and call `emitCall`.

MikaelMayer: Done — extracted `emitCall` helper that both the `**dict` and named-kwargs paths now call with their respective argument lists.

StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
```diff
@@ -7,6 +7,7 @@ List_slice_body_calls_List_drop_0: ✅ pass (in prelude file)
 List_slice_body_calls_List_take_1: ✅ pass (in prelude file)
 List_set_body_calls_List_set_0: ✅ pass (in prelude file)
 DictStrAny_get_body_calls_DictStrAny_get_0: ✅ pass (in prelude file)
+DictStrAny_get_or_none_body_calls_DictStrAny_get_0: ✅ pass (in prelude file)
```
MikaelMayer, line 10: This is a prelude assertion and hopefully one day we don't display them.

MikaelMayer: Agreed — the prelude assertions are noise in the expected outputs. Tracked separately.

PR #677, opened 2026-03-26
Support composite types in Any-typed contexts

MikaelMayer:
The pipeline crashes with a type unification error when a composite value flows into an `Any`-typed position (e.g., `str(e)` on an exception object in a `try/except` handler).

This PR fixes the issue by:

1. Generating `$composite_to_string_<type>` and `$composite_to_string_any_<type>` functions for each composite type. These take a composite parameter, so heap parameterization naturally adds a `Heap` dependency — preventing the verifier from unsoundly assuming referential transparency for `str()` calls across heap mutations.

2. When `str()` is called on a composite-typed variable (e.g., `str(e)` where `e: PythonError`), the translation emits `$composite_to_string_any_<Type>(e)` instead of `to_string_any(e)`, keeping the composite type and routing through the heap-aware function.

Existing tests pass. New test added for try/except with `str(e)` on PythonError.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -249,6 +249,10 @@ def translateType (ctx : TranslationContext) (typeStr : String) : Except Transla
 def AnyTy := mkCoreType PyLauType.Any
 def compositeToStringName (typeName : String) : String := "$composite_to_string_" ++ typeName
 def compositeToStringAnyName (typeName : String) : String := "$composite_to_string_any_" ++ typeName
+
+def isCompositeType (ctx : TranslationContext) (typeName : String) : Bool :=
+  typeName != PyLauType.Any && (ctx.importedSymbols[typeName]?.any fun s =>
+    match s with | .compositeType _ => true | _ => false)
 def strToAny (s: String) := mkStmtExprMd (.StaticCall "from_string" [mkStmtExprMd (StmtExpr.LiteralString s)])
 def intToAny (i: Int) := mkStmtExprMd (.StaticCall "from_int" [mkStmtExprMd (StmtExpr.LiteralInt i)])
 def boolToAny (b: Bool) := mkStmtExprMd (.StaticCall "from_bool" [mkStmtExprMd (StmtExpr.LiteralBool b)])
@@ -817,8 +821,7 @@ partial def translateCall (ctx : TranslationContext)
           if funcName == "to_string_any" && args.length == 1 then
             match inferExprType ctx args[0]! with
             | .ok argType =>
-              if argType != PyLauType.Any && (ctx.importedSymbols[argType]?.any fun s =>
-                match s with | .compositeType _ => true | _ => false) then
+              if isCompositeType ctx argType then
                 compositeToStringAnyName argType
               else funcName
             | .error _ => funcName
@@ -1825,9 +1828,9 @@ def pythonToLaurel' (info : PreludeInfo)
   -- for each composite type. These take a composite, so heap parameterization
   -- will add a Heap parameter, ensuring the verifier does not assume referential
   -- transparency across heap mutations.
-  let compositeToStringFns : List Procedure := compositeTypes.flatMap fun ct =>
+  for ct in compositeTypes do
     let selfParam : Parameter := { name := "self", type := mkHighTypeMd (.UserDefined ct.name.text) }
-    let toStr : Procedure :=
+    procedures := procedures.push
       { name := { text := compositeToStringName ct.name.text }
         inputs := [selfParam]
         outputs := [{ name := "result", type := mkHighTypeMd .TString }]
@@ -1837,7 +1840,7 @@ def pythonToLaurel' (info : PreludeInfo)
         body := .Opaque [] none []
         md := default
         isFunctional := false }
-    let toStrAny : Procedure :=
+    procedures := procedures.push
       { name := { text := compositeToStringAnyName ct.name.text }
         inputs := [selfParam]
         outputs := [{ name := "result", type := AnyTy }]
@@ -1847,10 +1850,9 @@ def pythonToLaurel' (info : PreludeInfo)
         body := .Opaque [] none []
         md := default
         isFunctional := false }
-    [toStr, toStrAny]
 
   let program : Laurel.Program := {
-    staticProcedures := (procedures.push mainProc |>.toList) ++ compositeToStringFns
+    staticProcedures := (procedures.push mainProc).toList
     staticFields := []
     types := compositeTypes.map TypeDefinition.Composite
     constants := []


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -805,7 +805,20 @@ partial def translateCall (ctx : TranslationContext)
...
+  | .Name  _ _ _ =>
+      -- If calling str() on a composite-typed variable, use composite_to_string_<type>
+      -- so that heap parameterization adds the heap dependency.
+      let funcName' :=
+        if funcName == "to_string_any" && args.length == 1 then
+          match inferExprType ctx args[0]! with
+          | .ok argType =>
+            if argType != PyLauType.Any && (ctx.importedSymbols[argType]?.any fun s =>
+              match s with | .compositeType _ => true | _ => false) then
+              s!"composite_to_string_{argType}"
```
MikaelMayer: I think that this function name should not be writable by a user.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -284,7 +284,6 @@ inline function Any_to_bool (v: Any) : bool
   if (Any..isfrom_string(v)) then !(Any..as_string!(v) == "") else
   if (Any..isfrom_int(v)) then !(Any..as_int!(v) == 0) else
   false
-  //WILL BE ADDED
```
MikaelMayer, line 287: This comment has no reason to be removed.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1095,7 +1108,7 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     let bodyDecls := collectDeclaredNamesAndTypes body.val.toList
 
     let errorVarDecls: List (String × String) := (handlers.val.toList.filterMap (λ h => match h with
-          | .ExceptHandler _ _ errname _ => errname.val)).map (λ h => (h.val, "PythonError"))
+          | .ExceptHandler _ _ errname _ => errname.val)).map (λ h => (h.val, "Any"))
```
MikaelMayer: We should still return PythonError, the to_string should now handle this gracefully by invoking composite_to_string_PythonError.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1111,10 +1124,17 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     let mut handlerStmts : List StmtExprMd := []
     for handler in handlers.val do
       match handler with
-      | .ExceptHandler _ _ _ handlerBody =>
+      | .ExceptHandler _ _ errname handlerBody =>
+        -- Assign exception variable from maybe_except: e := exception(maybe_except)
```
MikaelMayer: WHy is this needed?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1784,8 +1804,23 @@ def pythonToLaurel' (info : PreludeInfo)
...
+    let selfParam : Parameter := { name := "self", type := mkHighTypeMd (.UserDefined ct.name.text) }
+    { name := { text := s!"composite_to_string_{ct.name.text}" },
+      inputs := [selfParam],
+      outputs := [{ name := "result", type := mkHighTypeMd .TString }],
+      preconditions := [],
+      determinism := .deterministic none,
+      decreases := none,
+      body := .External,
+      md := default,
+      isFunctional := true }
```
MikaelMayer: Can you remove this isFunctional? __toString__ might modify the heap.

MikaelMayer:
We should have a proper test that demonstrates the try/catch that you are solving where e: PythonError is provided to the str() function to demonstrate we are supporting it without type checking issue.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1784,8 +1804,23 @@ def pythonToLaurel' (info : PreludeInfo)
     isFunctional := false
   }
 
+  -- Generate composite_to_string_<type> functions for each composite type.
+  -- These take a composite, so heap parameterization will add a Heap parameter,
+  -- ensuring the verifier does not assume referential transparency across heap mutations.
+  let compositeToStringFns : List Procedure := compositeTypes.map fun ct =>
```
MikaelMayer: This generates `composite_to_string_PythonError` (since `PythonError` is in `compositeTypes`). That seems intentional, but worth confirming — `str(e)` on a caught exception would hit this path if the variable were typed as `PythonError` rather than `Any`. Since exception variables are now typed as `Any` (line 1111), the `inferExprType` check at line 812 would return `"Any"` for them and skip the redirect, so the `PythonError` variant would go unused. Not a bug, just dead code for now.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Prefixed generated function names with `$` to prevent user name clashes.

MikaelMayer:
Added test in `VerifyPythonTest.lean`: try/except with `str(e)` on PythonError, verifying no type checking errors.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1806,8 +1819,36 @@ def pythonToLaurel' (info : PreludeInfo)
...
+        inputs := [selfParam]
+        outputs := [{ name := "result", type := mkHighTypeMd .TString }]
+        preconditions := []
+        determinism := .deterministic none
+        decreases := none
+        body := .Opaque [] none []
+        md := default
+        isFunctional := false }
+    let toStrAny : Procedure :=
+      { name := { text := "$composite_to_string_any_" ++ ct.name.text }
```
joscoh: Nit: it would be nicer to avoid hard-coding in this name in multiple places.

MikaelMayer: excellent idea.

MikaelMayer:
Extracted `compositeToStringName` and `compositeToStringAnyName` helpers to avoid hardcoding the name in multiple places.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -808,7 +810,20 @@ partial def translateCall (ctx : TranslationContext)
...
-    | .Name  _ _ _ => return mkCall funcName
+    | .Name  _ _ _ =>
+        -- If calling str() on a composite-typed variable, use $composite_to_string_any_<type>
+        -- so that heap parameterization adds the heap dependency.
+        let funcName' :=
+          if funcName == "to_string_any" && args.length == 1 then
+            match inferExprType ctx args[0]! with
+            | .ok argType =>
+              if argType != PyLauType.Any && (ctx.importedSymbols[argType]?.any fun s =>
+                match s with | .compositeType _ => true | _ => false) then
```
joehendrix: This predicate seems like it should be lifted to it's own function.  Something like isCompositeType?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1806,8 +1821,36 @@ def pythonToLaurel' (info : PreludeInfo)
...
+        determinism := .deterministic none
+        decreases := none
+        body := .Opaque [] none []
+        md := default
+        isFunctional := false }
+    [toStr, toStrAny]
+
   let program : Laurel.Program := {
-    staticProcedures := procedures.push mainProc |>.toList
+    staticProcedures := (procedures.push mainProc |>.toList) ++ compositeToStringFns
```
joehendrix: This should be a for loop over compositeTypes that appends to the mutable procedures array. Kiro seems to like to append to the end of potentially large linked lists.

PR #678, opened 2026-03-26
feat: emit obligation metadata in generated SMTLib

shigoel:
## Summary
- Appends `(set-info :final-message "...")` at the end of each generated SMTLib file, containing the "message" metadata field from the obligation. Emitted once per obligation, after all check-sat commands.
- Promotes MetaData field definitions from `def` to `abbrev` with `@[match_pattern]` so they can be used in pattern matches.
- Adds `MetaData.message` field definition.

### DIFF
--- a/Strata/Languages/Python/PySpecPipeline.lean
+++ b/Strata/Languages/Python/PySpecPipeline.lean
@@ -313,6 +313,23 @@ private def prependPrelude (coreFromLaurel : Core.Program) : Core.Program :=
     fun d => !coreOnlyNames.contains (toString d.name)
   { decls := filteredPrelude ++ coreOnly ++ userDecls }
 
+/-- Split procedure names in a Core program into prelude names
+    (before `FIRST_END_MARKER`) and user names (after it).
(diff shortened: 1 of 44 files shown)


Strata/Languages/Core/Verifier.lean
```diff
@@ -100,6 +100,13 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
       Solver.assert (← encodeTerm False (Factory.not obligationTerm) |>.run estate).1
       let _ ← Solver.checkSat ids
 
+  -- Emit the "message" metadata field at the very end (once per obligation).
+  match md.findElem Imperative.MetaData.message with
+  | some elem =>
+    let msg := toString (Std.format elem.value) |>.replace "\\" "\\\\" |>.replace "\"" "\\\""
```
joscoh, line 106: I assume this has to do with escape characters?

shigoel: Yep.

PR #680, opened 2026-03-26
Fix with-block variable scoping in Python-to-Laurel translation

MikaelMayer:
## Fix `with` statement variable scoping in Python-to-Laurel translation

Python `with` blocks (e.g., `with open(...) as f:`) were placing variable declarations inside a Laurel `Block`, which scopes them in Core. When a function contained multiple `with` blocks reusing the same variable name, the second block would generate a `havoc` for a variable that wasn't declared at the outer Core scope, causing a "Cannot havoc undeclared variable" type-checking error.

The fix hoists the manager and `as`-variable declarations out of the `Block`, matching Python's scoping semantics where `with`-block variables remain visible after the block ends.

New test added in `VerifyPythonTest.lean` exercising multiple `with` blocks with the same variable name.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -247,6 +247,12 @@ def translateType (ctx : TranslationContext) (typeStr : String) : Except Transla
           .ok (mkCoreType PyLauType.Any)
 
 def AnyTy := mkCoreType PyLauType.Any
+def compositeToStringName (typeName : String) : String := "$composite_to_string_" ++ typeName
+def compositeToStringAnyName (typeName : String) : String := "$composite_to_string_any_" ++ typeName
+
+def isCompositeType (ctx : TranslationContext) (typeName : String) : Bool :=
+  typeName != PyLauType.Any && (ctx.importedSymbols[typeName]?.any fun s =>
+    match s with | .compositeType _ => true | _ => false)
 def strToAny (s: String) := mkStmtExprMd (.StaticCall "from_string" [mkStmtExprMd (StmtExpr.LiteralString s)])
 def intToAny (i: Int) := mkStmtExprMd (.StaticCall "from_int" [mkStmtExprMd (StmtExpr.LiteralInt i)])
 def boolToAny (b: Bool) := mkStmtExprMd (.StaticCall "from_bool" [mkStmtExprMd (StmtExpr.LiteralBool b)])
@@ -808,7 +814,19 @@ partial def translateCall (ctx : TranslationContext)
   let emitCall (callArgs : List StmtExprMd) : Except TranslationError StmtExprMd := do
     let mkCall (name : String) := mkStmtExprMd (StmtExpr.StaticCall name callArgs)
     match f with
-    | .Name  _ _ _ => return mkCall funcName
+    | .Name  _ _ _ =>
+        -- If calling str() on a composite-typed variable, use $composite_to_string_any_<type>
+        -- so that heap parameterization adds the heap dependency.
+        let funcName' :=
+          if funcName == "to_string_any" && args.length == 1 then
+            match inferExprType ctx args[0]! with
+            | .ok argType =>
+              if isCompositeType ctx argType then
+                compositeToStringAnyName argType
+              else funcName
+            | .error _ => funcName
+          else funcName
+        return mkCall funcName'
     | .Attribute _ val _attr _ =>
         let _target_trans ← translateExpr ctx val
         if opt_firstarg.isSome then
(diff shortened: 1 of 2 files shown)


StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -101,4 +101,19 @@ open Strata.Parser (stringInputContext)
     unless expectedErrors.any (fun exp => matchesDiagnostic d exp) do
       throw <| .userError s!"Unexpected diagnostic: line {d.start.line}, {d.message}"
 
+-- Multiple `with` blocks reusing the same variable name should not crash.
+-- Previously, the `as` variable leaked out of the with-block scope, causing
```
MikaelMayer: I can't accept stateful comments.

MikaelMayer: Nit: the wording "leaked out of the with-block scope" is slightly misleading — the actual bug was the opposite: the variable *name* was tracked in the context (via `bodyCtx`) but the *declaration* was trapped inside the `Block`, so the second `with` block saw the name in `variableTypes`, took the "already declared" path, and emitted an assign/havoc for a variable that didn't exist at the outer Core scope.

Something like "the `as` variable was declared inside the Block but its name was tracked in the outer context" would be more precise. Minor — up to you.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1185,19 +1186,22 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
             currentCtx := {currentCtx with variableTypes := currentCtx.variableTypes ++ [(varName, PyLauType.Any)]}
-            setupStmts := setupStmts ++ [mgrDecl, varDecl]
+            declStmts := declStmts ++ [mgrDecl, varDecl]
         | none =>
-          setupStmts := setupStmts ++ [mgrDecl, enterCall]
+          declStmts := declStmts ++ [mgrDecl]
+          setupStmts := setupStmts ++ [enterCall]
         cleanupStmts := cleanupStmts ++ [mkStmtExprMd (StmtExpr.InstanceCall mgrRef "__exit__" [])]
-    let (bodyCtx, bodyStmts) ← translateStmtList currentCtx body.val.toList
+    let (_bodyCtx, bodyStmts) ← translateStmtList currentCtx body.val.toList
```
MikaelMayer, line 1220: Not blocking — just flagging a potential follow-up: unlike `If`, `While`, and `For` (which call `collectDeclaredNamesAndTypes` on their bodies to hoist inner declarations), the `With` handler doesn't pre-scan its body. This means variables first assigned inside a `with` block won't be visible after it, which doesn't match Python semantics.

Returning `currentCtx` here is more correct than the old `bodyCtx` (which created a context/declaration mismatch), but a complete fix would add `collectDeclaredNamesAndTypes body.val.toList` hoisting like the other compound statements do.

StrataTest/Languages/Python/VerifyPythonTest.lean
MikaelMayer: Fixed — rewrote the comment to be stateless.

MikaelMayer: Updated to the suggested wording — more precise.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Acknowledged — good follow-up candidate for body hoisting with collectDeclaredNamesAndTypes.

StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -101,4 +101,20 @@ open Strata.Parser (stringInputContext)
     unless expectedErrors.any (fun exp => matchesDiagnostic d exp) do
       throw <| .userError s!"Unexpected diagnostic: line {d.start.line}, {d.message}"
 
+-- Multiple `with` blocks reusing the same variable name should not crash.
+-- The `as` variable was declared inside the Block but its name was tracked
```
joscoh: I don't think we need to mention the previous behavior here

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1185,19 +1186,22 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
         | some varExpr =>
           let varName := pyExprToString varExpr
           if varName ∈ currentCtx.variableTypes.unzip.fst then
-            let assignStmt := mkStmtExprMd (StmtExpr.Assign
-              [mkStmtExprMd (StmtExpr.Identifier varName)] enterCall)
-            setupStmts := setupStmts ++ [mgrDecl, assignStmt]
+            -- Variable already declared — just reassign inside the block
+            declStmts := declStmts ++ [mgrDecl]
```
joscoh, line 1208: I don't like appending to the end of a list in a for-loop (`n^2` vs `n` with cons+reverse), but since it was there before, it can be fixed in a future PR

StrataTest/Languages/Python/VerifyPythonTest.lean
MikaelMayer: Removed — comment is now just the one-liner.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: Agreed, pre-existing issue. Not touching it here.

PR #687, opened 2026-03-27
Stop re-verifying prelude in pyAnalyzeLaurel

MikaelMayer:
## Summary

The prelude was being re-verified for every Python test file, wasting time and cluttering test output with 22 identical prelude lines per file.

## Changes

**Verifier** (`Strata/Languages/Core/Verifier.lean`): When `proceduresToVerify` is set, after `CallElim` inlines contracts at call sites, drop every procedure not in the target list. Previously the second `FilterProcedures` used a stale call-graph closure and kept prelude procedures, causing them to be re-verified every time.

**Expected files**: All 39 `expected_laurel` test files updated — 22 prelude lines removed from each, DETAIL counts adjusted.

Prelude correctness is verified once by the dedicated `PreludeVerifyTest`.

## Testing

- `lake build StrataMain` passes
- `PreludeVerifyTest` passes
- Core verification tests pass (`FailingAssertion`, `Axioms`, `AlwaysRunSMT`, `RemoveIrrelevantAxioms`)

### DIFF
--- a/Strata/Languages/Python/PySpecPipeline.lean
+++ b/Strata/Languages/Python/PySpecPipeline.lean
@@ -36,6 +36,8 @@ public structure PySpecLaurelResult where
   functionSignatures : List Python.PythonFunctionDecl := []
   /-- Maps unprefixed class names to prefixed names for type resolution. -/
   typeAliases : Std.HashMap String String := {}
+  /-- Classes whose spec is considered exhaustive (lists all methods). -/
+  exhaustiveClasses : Std.HashSet String := {}
 
 /-! ### Private Helpers -/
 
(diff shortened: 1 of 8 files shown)


Strata/Languages/Core/Verifier.lean
```diff
@@ -762,8 +762,17 @@ def verify (program : Program)
...
+        -- After all transforms, keep only the target procedures.
+        -- PrecondElim has already inserted call-site precondition
+        -- checks into user procedure bodies and generated WF
+        -- procedures for prelude functions; we discard the latter.
+        let procsSet := Std.HashSet.ofList
+          (procs ++ procs.map PrecondElim.wfProcName)
+        let prog := { prog with decls := prog.decls.filter fun d =>
+          match d with
+          | .proc p _ => procsSet.contains (CoreIdent.toPretty p.header.name)
+          | _ => true }
```
aqjune-aws, line 780: Looks good to me - what do you think about calling `FilterProcedures.run prog ( procs ++ procs.map PrecondElim.wfProcName)` once again instead? This will help the `CachedAnalyses` data structure in CoreTransformM monad stay valid.

MikaelMayer: Excellent idea, much better.

MikaelMayer: Tried using FilterProcedures.run — unfortunately it uses the call-graph closure (getAllCalleesClosure), which pulls prelude functions back in as dependencies of the WF procedures. We need a strict name-based filter here. Added a comment in the code explaining why.

PR #690, opened 2026-03-27
feat(python): Add support for ** (power) operator

MikaelMayer:
The Python-to-Laurel pipeline rejected programs using the `**` operator with "Binary operator not yet supported: Pow", causing an exit-4 crash.

This PR adds support by:
- Mapping `Pow` to the existing `PPow` prelude function in `PythonToLaurel.lean`
- Constant-folding `**` with integer literal operands at translation time (e.g. `1024 ** 3` becomes `1073741824`)
- For non-constant operands, `PPow` delegates to an uninterpreted `int_pow` function, handling int×int, bool×int, and int×bool combinations
- Replacing the `PPow` stub (which threw `UnimplementedError`) with the real implementation

Tested with a new `test_power.py` test that verifies `8 ** 2 == 64` and `2 ** 10 == 1024` through the full pipeline. Existing tests pass.


Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -850,11 +850,24 @@ inline function POr (v1: Any, v2: Any) : Any
 
 
 // /////////////////////////////////////////////////////////////////////////////////////
-// Modelling of other Python operations, currrently unsupported
+// Integer exponentiation (evaluable for constants, UF in SMT otherwise)
+// /////////////////////////////////////////////////////////////////////////////////////
+function int_pow (base: int, exp: int) : int;
```
joscoh: What happens when `exp` is negative?

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -850,11 +850,24 @@ inline function POr (v1: Any, v2: Any) : Any
...
+
+// /////////////////////////////////////////////////////////////////////////////////////
+// Modelling of other Python operations
 // /////////////////////////////////////////////////////////////////////////////////////
 inline function PPow (v1: Any, v2: Any) : Any
 {
-  exception(UnimplementedError ("Pow operator is not supported"))
+  if Any..isexception(v1) then v1 else if Any..isexception(v2) then v2
```
joehendrix: This may be out of scope, but a good a place to ask as any.  I see this pattern of Any..isexception and throwing exception in other operators, but it doesn't seem to make it to PythonToLaurel checks.

Am I overlooking something?  Maybe we should consider adding exception information?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -457,6 +457,16 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
 
   -- Binary operations
   | .BinOp _ left op right => do
+    -- Constant-fold Pow for integer literals
```
joscoh: It seems strange to me that this is done here and not as a `Factory` `concreteEval` but that would be a larger change (not blocking).

Strata/Languages/Python/PythonLaurelCorePrelude.lean
MikaelMayer: You might have identified an issue indeed, not related to this PR but indeed if we don't check that things are exceptions at the statement level, it could defeat that purpose.
I'm not sure however that it could harm us however, it's like having an uninterpreted value for non-supported cases.  It would require more investigation.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: No it's definitely something I want to avoid. Any constant folding should be done in Core using concreteEval so that we have one place to check soundness.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
thanhnguyen-aws: I will create a PR to handle this issue.

Strata/Languages/Core/ProgramEval.lean
```diff
@@ -80,8 +80,14 @@ def eval (E : Env) : List (Program × Env) :=
       go rest declsE
 
     | .func func _ =>
+      -- Skip bodyless functions that already exist in the factory
```
MikaelMayer: Regular expression methods were not doing that kind of hack and it did work. Figure out why.
Doing that causes a risk of defining a method whose concrete evaluation is accidentally captured as a built in function.
At minimum there should be an boolean parameter in the original Laurel AST to allow that.

MikaelMayer: Reverted the hack. The regex pattern works because re_fullmatch_str uses the regex type which fails in handleUnaryOps during Laurel-to-Core translation, so it never reaches Core. The factory provides it without conflict. int_pow uses int which translates fine, causing the duplicate. Constant folding stays in PythonToLaurel for now.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -849,25 +849,12 @@ inline function POr (v1: Any, v2: Any) : Any
...
-  if Any..isexception(v1) then v1 else if Any..isexception(v2) then v2
-  else if (Any..isfrom_int(v1) && Any..isfrom_int(v2)) then
-    from_int(int_pow(Any..as_int!(v1), Any..as_int!(v2)))
-  else if (Any..isfrom_bool(v1) && Any..isfrom_int(v2)) then
-    from_int(int_pow(bool_to_int(Any..as_bool!(v1)), Any..as_int!(v2)))
-  else if (Any..isfrom_int(v1) && Any..isfrom_bool(v2)) then
-    from_int(int_pow(Any..as_int!(v1), bool_to_int(Any..as_bool!(v2))))
-  else
-    exception(UndefinedError ("Operand Type is not defined"))
+  exception(UnimplementedError ("Pow operator is not supported"))
```
MikaelMayer: The problem of this implementation is that it's not sound.
We will be able to _prove_ that PPow never returns an integer and thus that the following code is valid:
```
try:
  a = 2 ** b
  a = a / 0 // Proved unreachable
catch:
  ...
```

MikaelMayer: This is already fixed in the latest commit. PPow now calls int_pow (which is provided by the factory with concreteEval). For constant operands, concreteEval computes the result. For symbolic operands, int_pow stays uninterpreted in SMT — which is sound (SAT results become unknown, not incorrectly valid).

Strata/Languages/Python/PythonLaurelCorePrelude.lean
MikaelMayer: Case handled.

Strata/Languages/Python/PyFactory.lean
```diff
@@ -114,12 +114,40 @@ def rePatternErrorFunc : LFunc Core.CoreLParams :=
...
+          | [b, e] => match LExpr.denoteInt b, LExpr.denoteInt e with
+            | some bv, some ev =>
+              if ev ≥ 0 then .some (LExpr.intConst md (bv ^ ev.toNat)) else .none
+            | _, _ => .none
+          | _ => .none)
+      }
+
+-- Float exponentiation (uninterpreted). Used for negative integer exponents
+-- where Python returns a float (e.g. 2 ** -3 = 0.125).
+def floatPowFunc : LFunc Core.CoreLParams :=
```
joscoh, line 137: Not completely sound, e.g. `3^(-1)` is not representable as a float, but this is as good as we can do until we have proper float support, I believe.

MikaelMayer: Mmmh in general no but in python it is. So perhaps something we will do in the future is to rename these functions so that we know they correspond to a certain implementation. Integers are unbounded ,but floats are bounded.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -844,11 +846,23 @@ inline function POr (v1: Any, v2: Any) : Any
...
+  else if (Any..isfrom_int(v1) && Any..isfrom_int(v2)) then
+    if Any..as_int!(v2) >= 0 then from_int(int_pow(Any..as_int!(v1), Any..as_int!(v2)))
+    else from_float(float_pow(int_to_real(Any..as_int!(v1)), int_to_real(Any..as_int!(v2))))
+  else if (Any..isfrom_bool(v1) && Any..isfrom_int(v2)) then
+    if Any..as_int!(v2) >= 0 then from_int(int_pow(bool_to_int(Any..as_bool!(v1)), Any..as_int!(v2)))
+    else from_float(float_pow(bool_to_real(Any..as_bool!(v1)), int_to_real(Any..as_int!(v2))))
+  else if (Any..isfrom_int(v1) && Any..isfrom_bool(v2)) then
+    from_int(int_pow(Any..as_int!(v1), bool_to_int(Any..as_bool!(v2))))
+  else
+    exception(UndefinedError ("Operand Type is not defined"))
```
joehendrix: What about the following to reflect we are mssing a few cases (like float and operator overloading). 

```suggestion
    exception(UnimplementedError("Pow is not defined on these input types"))
```

StrataTest/Languages/Python/tests/test_power.py
```diff
@@ -0,0 +1,10 @@
+def main():
+    # Integer exponentiation
+    x: int = 8 ** 2
+    assert x == 64, "8 ** 2 should be 64"
+
+    y: int = 2 ** 10
+    assert y == 1024, "2 ** 10 should be 1024"
```
joehendrix, line 7: It'd be good to cover special cases (bool, negative exponents, maybe float if supported).

Strata/Languages/Python/PyFactory.lean
joscoh: I'm confused why this being Python makes a difference? Floats are always bounded

Strata/Languages/Python/PythonLaurelCorePrelude.lean
MikaelMayer: Applied the suggestion — changed to UnimplementedError.

StrataTest/Languages/Python/tests/test_power.py
MikaelMayer: Added bool base cases (True**5, False**3). Negative exponents return from_float(float_pow(...)) which is uninterpreted — can't assert on the result in a test, but the path is exercised. Float base operands fall through to UnimplementedError.

MikaelMayer: If we add negative exponents or floats, we should see inconclusive results. Let me work on that too.

Strata/Languages/Python/PyFactory.lean
MikaelMayer: `3^(-1) is not representable as a float`
In python `3 ** (-1)` is a float that approximates `1/3`.

MikaelMayer: Right, 3^(-1) = 0.333... isn't exactly representable as IEEE 754 float. But Python still returns a float (0.3333333333333333). In our model, we use `real` (mathematical real) via `float_pow`, which is more precise than Python's actual float — so it's a sound over-approximation.

joscoh: It is still not sound, because real numbers satisfy properties that floats do not.
E.g. associativity:
`(x+y)+z = x +(y+z)` is true for all reals
But
`(1 - 1) + 10 ** (-30) == 1 + (-1 + 10 ** (-30))` is false in floats and in Python. If we encode as an uninterpreted real function, the above is true.

joscoh: To be a sound over approximation, floats have to be uninterpreted, not reals.

MikaelMayer: Ok let me just pause for a moment, What I mean is that `float_pow` is currently not named `real_pow` for the reason you mention. Whereas `real_pow` can have tons of algebraic properties, `float_pow` does not have them. We shouldn't emit an associativity axiom on `float_pow` that allows us to prove `(1 - 1) + 10 ** (-30) == 1 + (-1 + 10 ** (-30))` (and any other axiom that is true typically for reals, such as commutativity or associativity).

joscoh: Yes, but could the SMT solver use its knowledge of real numbers to use such properties implicitly? Basically, I am saying, suppose the SMT solver knows, for any reals, associativity holds. Then, even if `float_pow` is completely uninterpreted, the SMT solver may still apply associativity "knowing" it is real-valued.

MikaelMayer: For the SMT point of view `float_pow` is an uninterpreted function over two reals so any solver would have absolutely zero knowledge on it. An SMT solver cannot add constraints, it can suppose constraints but only to finally disprove them, which makes it possible to invalidate constraints that were supposed. So at best, if we provided the assumption about concrete values of `float_pow`, an SMT solver would be able to invalidate the associativity.

Strata/Languages/Python/PyFactory.lean
```diff
@@ -114,12 +114,40 @@ def rePatternErrorFunc : LFunc Core.CoreLParams :=
...
+        (fun md args => match args with
+          | [b, e] => match LExpr.denoteInt b, LExpr.denoteInt e with
+            | some bv, some ev =>
+              if ev ≥ 0 then .some (LExpr.intConst md (bv ^ ev.toNat)) else .none
+            | _, _ => .none
+          | _ => .none)
+      }
+
+-- Float exponentiation (uninterpreted). Used for negative integer exponents
+-- where Python returns a float (e.g. 2 ** -3 = 0.125).
```
MikaelMayer, line 135:
```suggestion
-- Float exponentiation (uninterpreted). Used for negative integer exponents
-- where Python returns a float (e.g. 2 ** -3 = 0.125).
-- This function should NOT be mapped to any real-based power functions in solvers, since float power is imprecise.
```

MikaelMayer: Applied.

PR #691, opened 2026-03-27
Migrate `StrataVerify` functionality to `strata verify`

atomb:
Migrates the standalone `StrataVerify` executable into a `verify` subcommand of the `strata` CLI (`StrataMain.lean`), deletes the old `StrataVerify.lean`, and enhances the shared argument parser to support `--flag=value` syntax (with validation that values don't contain `=`). All references across CI, C# integration tests, documentation, and example files are updated.

### Features

- The `verifyCommand` faithfully reproduces all flags and behavior from the original `StrataVerify.lean`, including backward compatibility with `--output-format=sarif` via the new `--flag=value` parsing.
- The `--flag=value` enhancement to `parseArgs` is well-implemented: it splits on `=`, validates that the value doesn't itself contain `=` (clear error message), and correctly handles both `.arg` and `.repeat` flag types with fallback to the next positional argument when no inline value is present.
- The diff is clean: no stale files, no dead code. `StrataVerify.lean` is deleted, no untracked files are accidentally included.
- The steering doc now lists `strata` with its subcommands instead of the removed `StrataVerify` entry.
- The C# integration test correctly updates both the binary path and prepends `verify` to the arguments.
- All documentation (README, GettingStarted, CI, example comments) is consistently updated.
- Build succeeds and smoke tests pass for `--sarif`, `--output-format=sarif`, and the `=`-in-value rejection.

### Testing

Existing tests that previously used `StrataVerify` now use `strata verify`.

### DIFF
--- a/Strata/SimpleAPI.lean
+++ b/Strata/SimpleAPI.lean
@@ -290,6 +290,7 @@ def pySpecsDir (sourceDir strataDir dialectFile : System.FilePath)
     (events : Std.HashSet String := {})
     (skipNames : Array String := #[])
     (warningOutput : WarningOutput := .detail)
+    (pythonCmd : String := "python")
     : EIO String Unit := do
   -- Create output dir
   match ← IO.FS.createDirAll strataDir |>.toBaseIO with
@@ -347,7 +348,7 @@ def pySpecsDir (sourceDir strataDir dialectFile : System.FilePath)
     match ← Strata.Python.Specs.translateFile
         dialectFile strataDir pythonFile sourceDir
         (events := events) (skipNames := skipIdents)
-        (moduleName := mod) |>.toBaseIO with
+        (moduleName := mod) (pythonCmd := pythonCmd) |>.toBaseIO with
     | .error msg =>
       Python.Specs.baseLogEvent events "import" s!"Failed {mod}: {msg}"
       failures := failures.push (toString mod, msg)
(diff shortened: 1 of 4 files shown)


StrataMain.lean
```diff
@@ -1180,20 +1305,35 @@ private def parseArgs (cmdName : String)
...
           parseArgs cmdName flagMap acc (pflags.insertFlag flagName Option.none) cmdArgs
         | .arg _ =>
-          let value :: cmdArgs := cmdArgs
-            | exitCmdFailure cmdName s!"Expected value after {arg}."
-          parseArgs cmdName flagMap acc (pflags.insertFlag flagName (some value)) cmdArgs
+          match inlineValue with
+          | some value =>
+            parseArgs cmdName flagMap acc (pflags.insertFlag flagName (some value)) cmdArgs
+          | none =>
+            let value :: cmdArgs := cmdArgs
```
joscoh, line 1331: I'm confused by this part. Why does it matter if `cmdArgs` is `Nil` here or not, if there is no value after the `=`. In what circumstances would this not be an error?

atomb: The logic here is to allow both `--flag=arg` and `--flag arg`. If the `splitOn` above includes a value (`inlineValue` is `.some`), then it's the former case. If not, it's the latter case, and we take the next element from `cmdArgs`.

PR #693, opened 2026-03-27
PythonToLaurel: add method arguments to context during translation, resolve modifies error 

aqjune-aws:
This patch resolves two failures:
- Add method arguments to context when translating Python to Laurel
- Rename the internal variable name for the argument to add 'in_', which will resolve the 'modifies'-type checking error.
These two fixes are needed to make the new test pass.

### DIFF
--- a/Strata/SimpleAPI.lean
+++ b/Strata/SimpleAPI.lean
@@ -266,18 +266,21 @@ private partial def discoverModules (sourceDir : System.FilePath)
 /-- Derive the output path for a Python file by mirroring the source directory
     structure and replacing `.py` with `.pyspec.st.ion`. -/
 def pySpecOutputPath (sourceDir strataDir pythonFile : System.FilePath)
-    : Option System.FilePath :=
+    : Option System.FilePath := Id.run do
   let sourceDirStr := sourceDir.toString
-  let srcPrefix := if sourceDirStr.endsWith "/" then sourceDirStr else sourceDirStr ++ "/"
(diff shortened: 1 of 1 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1512,15 +1512,35 @@ def translateMethod (ctx : TranslationContext) (className : String)
     let outputs : List Parameter := [{name := "LaurelResult", type := AnyTy}]
 
     -- Translate method body with class context
-    let ctxWithClass := {ctx with currentClassName := some className}
+    -- Translate method body with class context
```
shigoel, line 1514:
```suggestion
```
Duplicate comment?

StrataTest/Languages/Python/tests/test_method_param_reassign.py
```diff
@@ -0,0 +1,12 @@
+class Validator:
+    def __init__(self):
+        pass
+
+    def process(self, account_id: Any) -> Any:
+        """Reassigning a method parameter inside try must not produce
```
andrewmwells-amazon, line 6: I'm worried this won't be easily caught if the `.expected` file changes. Could we also add some `guard_msgs` for the Laurel itself or similar?

aqjune-aws: It seems Kiro tries to create a big test code for this.... Good idea, but skipping for now.

PR #694, opened 2026-03-27
Add documentation for DDM-generated Java code

atomb:
Adds `docs/DDMJavaCodeGen.md`, a reference for the DDM Java code generator. Covers CLI usage (`strata javaGen`), Lean API, generated file structure, type mapping, Ion serialization format, name disambiguation, and limitations. Documentation only — no code changes.


Strata/DDM/Integration/Java/Gen.lean
```diff
@@ -151,35 +147,72 @@ partial def syntaxCatToQualifiedName (cat : SyntaxCat) : Option QualifiedIdent :
...
+  | q`Init.Str => some "serializeStrlit"
+  | q`Init.Num => some "serializeNum"
+  | q`Init.Decimal => some "serializeDecimal"
+  | q`Init.Bool => some "serializeBool"
+  | q`Init.ByteArray => some "serializeBytes"
+  | _ => .none
+
+/-- Get the serializer method reference for a SyntaxCat's inner type (used in Option/List). -/
+partial def serializerFnRef (c : SyntaxCat) : String :=
+  match primitiveSerializerMethod c.name with
```
MikaelMayer, line 165: `serializerFnRef` falls through to `$s::serialize` for abstract categories, dialect-defined categories, and any unrecognized `Init.*` category. This is correct because `serialize(Node)` handles all `Node` subtypes, but a brief docstring note that this intentionally delegates to the `Node`-based overload for non-primitive types would help readers understand why there's no abstract-category check here (unlike `serializeFieldExpr` which has one).

Strata/DDM/Integration/Java/Gen.lean
```diff
@@ -270,59 +323,46 @@ def assignAllNames (d : Dialect) : NameAssignments :=
...
+      -- For single-op categories where the op name matches the category, use "Of"
+      let base := if ops.size == 1 && base.toLower == catName.toLower then "Of" else base
+      let (name, newLocalUsed) := disambiguate base localUsed
+      (opNames.insert (op.category, op.name) name, newLocalUsed)
+    opNames
 
   -- Assign stub names (referenced types not in this dialect's categories)
   let stubInit : Std.HashMap QualifiedIdent String × Std.HashSet String := ({}, used)
-  let (stubNames, used) := refs.toArray.foldl (init := stubInit) fun (map, used) ref =>
+  let (stubNames, _used) := refs.toArray.foldl (init := stubInit) fun (map, used) ref =>
```
MikaelMayer: The variable is named `_used` (suggesting it's unused) but is actually referenced on line 356 for `disambiguate`. Consider naming it `used'` or `usedAfterStubs` to avoid the misleading `_` prefix convention.

Strata/DDM/Integration/Java/Gen.lean
```diff
@@ -233,11 +252,45 @@ def generateNodeInterface (package : String) (categories : List String) : String
 def generateStubInterface (package : String) (name : String) : String × String :=
   (s!"{name}.java", s!"package {package};\n\npublic non-sealed interface {name} extends Node \{}\n")
```
MikaelMayer, line 253: No action needed — the `generateSerializer` function is now a clean one-liner after removing the separator map injection. Nice simplification.

Strata/DDM/Integration/Java/templates/IonSerializer.java
```diff
@@ -6,145 +6,102 @@
...
-        }
-        if (value instanceof java.util.Optional<?> opt) {
-            return serializeOption(opt);
-        }
-        if (value instanceof java.util.List<?> list) {
-            return serializeSeq(list, sep != null ? sep : "seq");
-        }
-        throw new java.lang.IllegalArgumentException("Unsupported type: " + type);
+    public IonValue serializeIdent(java.lang.String s) {
+        IonSexp sexp = ion.newEmptySexp();
```
MikaelMayer, line 42: The `serializeIdent` and `serializeStrlit` methods (lines 42-55) have identical structure — they differ only in the symbol name (`"ident"` vs `"strlit"`). A private helper like `serializePrimString(String symbol, String s)` could reduce this duplication. Same pattern applies to `serializeNum`, `serializeDecimal`, and `serializeBytes` (lines 57-82) which all follow the `(symbol null value)` pattern with different symbol names and value creation.

docs/DDMJavaCodeGen.md
```diff
@@ -0,0 +1,377 @@
...
+
+When a **multi-operator category** has an operator whose name matches the
+category name, the record gets a `_` suffix:
+
+```java
+public sealed interface Procedure extends Node permits Procedure.Procedure_, Procedure.Function {
+    public record Procedure_(...) implements Procedure { ... }
+    public record Function(...) implements Procedure { ... }
+}
+```
```
MikaelMayer, line 192: The multi-operator collision example shows `Procedure_` as the disambiguated name, which is correct per the `disambiguate` function. However, the example uses `Procedure` as both the category and operator name — it might be worth noting that this applies whenever `toPascalCase(opName).toLower == catName.toLower`, not just exact string matches, since the comparison is case-insensitive.

docs/DDMJavaCodeGen.md
```diff
@@ -0,0 +1,377 @@
...
+### `Node`
+
+```java
+import com.amazon.ion.IonSexp;
+
+public sealed interface Node permits Stmt, Expr, ... {
+    SourceRange sourceRange();
+    java.lang.String operationName();
+    IonSexp toIon(IonSerializer $s);
+}
```
MikaelMayer, line 119: The `Node` interface snippet shows `permits Stmt, Expr, ...` — the `...` is clear enough as pseudocode, but since the actual `permits` clause also includes stub interfaces, it might be worth showing that explicitly (e.g., `permits Stmt, Expr /* stubs */`) to match the description in the paragraph below.

Strata/DDM/Integration/Java/Gen.lean
fabiomadge: I think the function is simple enough that the fallback is self-evident from the pattern match. `serializeFieldExpr` is more complex because it handles `Option`/`List` wrapping — the difference isn't a missing check, it's a different concern.

Strata/DDM/Integration/Java/Gen.lean
fabiomadge: Good catch — fixed in 7bcbcf5c.

Strata/DDM/Integration/Java/templates/IonSerializer.java
fabiomadge: Done in 7bcbcf5c — extracted a `newTagged` private helper. The 5 primitive serializers are now one-liners.

docs/DDMJavaCodeGen.md
fabiomadge: Added in 7bcbcf5c.

docs/DDMJavaCodeGen.md
fabiomadge: The paragraph right below the snippet already explains that the `permits` clause includes stubs. I think the `...` is clear enough without duplicating that in the code example.

MikaelMayer (dismissed):

docs/DDMJavaCodeGen.md
```diff
@@ -0,0 +1,377 @@
...
+1. **Java reserved words** — a trailing `_` is appended (e.g., `class` → `class_`).
+   Applied after stripping invalid characters.
+2. **Base class names** — `Node`, `SourceRange`, and `IonSerializer` are reserved.
+3. **Cross-dialect collisions** — when two categories share the same short name,
+   the dialect name is prefixed in PascalCase (e.g., `LambdaExpr` vs `CoreExpr`).
+4. **Operator/category collisions** — for single-operator categories where the
+   operator name matches the category name, the record is named `Of`. For
+   multi-operator categories, a `_` suffix is added, then `_2`, `_3`, etc.
+   for further collisions.
+5. **Invalid characters** — non-alphanumeric characters (except `_`) are stripped.
```
joscoh, line 354: Curious, does this mean that two names can become equal after this preprocessing when there were no collisions before?

Strata/DDM/Integration/Java/Gen.lean
```diff
@@ -151,35 +147,72 @@ partial def syntaxCatToQualifiedName (cat : SyntaxCat) : Option QualifiedIdent :
...
+  | q`Init.Ident => some "serializeIdent"
+  | q`Init.Str => some "serializeStrlit"
+  | q`Init.Num => some "serializeNum"
+  | q`Init.Decimal => some "serializeDecimal"
+  | q`Init.Bool => some "serializeBool"
+  | q`Init.ByteArray => some "serializeBytes"
+  | _ => .none
+
+/-- Get the serializer method reference for a SyntaxCat's inner type (used in Option/List). -/
+partial def serializerFnRef (c : SyntaxCat) : String :=
```
joscoh: Why is this `partial`, it is not even recursive AFAIK?

Strata/DDM/Integration/Java/Gen.lean
```diff
@@ -151,35 +147,72 @@ partial def syntaxCatToQualifiedName (cat : SyntaxCat) : Option QualifiedIdent :
...
+  match primitiveSerializerMethod c.name with
+  | some method => s!"$s::{method}"
+  | none => "$s::serialize"
+
+/-- Generate the serialization expression for a single field. -/
+partial def serializeFieldExpr (kind : ArgDeclKind) (fieldName : String) : String :=
```
joscoh: Same here

StrataTest/DDM/Integration/Java/TestGen.lean
```diff
@@ -320,6 +341,16 @@ elab "#testRoundtrip" : command => do
     if cmd.name != (⟨"Simple", "block"⟩ : Strata.QualifiedIdent) then Lean.logError "Expected block command"; return
     if let .seq _ _ stmts := cmd.args[0]! then
       if stmts.size != 4 then Lean.logError s!"Expected 4 statements, got {stmts.size}"
+      -- Verify print's msg arg is strlit (not ident) — catches Init.Str serialization bug
+      if stmts.size < 2 then Lean.logError "Expected at least 2 statements"; return
+      match stmts[1]! with
```
joscoh, line 346: Since you already check `stmts.size < 2` before, you can use the safe indexing function here. This avoids introducing a `panic`.

Strata/DDM/Integration/Java/Gen.lean
fabiomadge: You're right, neither this nor `serializeFieldExpr` are recursive. Removed `partial` from both — builds clean.

Strata/DDM/Integration/Java/Gen.lean
fabiomadge: (Addressed above.)

StrataTest/DDM/Integration/Java/TestGen.lean
fabiomadge: Agreed in principle, but the `if ... return` guard doesn't propagate the size proof into the `do` continuation, so `stmts[1]'(by omega)` doesn't work here. `stmts[1]?` works but adds nesting and a redundant `none` branch. Which approach did you have in mind?

docs/DDMJavaCodeGen.md
fabiomadge: Yes — e.g. `foo?` and `foo` both become `foo` after stripping. But `disambiguate` runs after stripping and appends `_`, `_2`, etc. to resolve any resulting collisions, so the final names are always unique.

StrataTest/DDM/Integration/Java/TestGen.lean
tautschnig: A pragmatic option would be `(stmts.get? 1).get!` - still panics on `none` but makes the intent clearer. Or just leave it as-is with a comment noting the guard.

tautschnig (approved):

StrataTest/DDM/Integration/Java/TestGen.lean
joehendrix: I introduce a `decideProp` utility function  specifically for this.  With that the code would look something like:

```
  let .isTrue stmtsSizeP := decideProp (stmts.size >= 2)
    | Lean.logError "Expected at least 2 statements"; return
```  

With this, the proposition would be brought into scope for use later.

joscoh: I don't think this is blocking, I had hoped there was a simple way to do this. I think Joe's solution is the nicest, but I am fine if you want to keep the current version.

PR #695, opened 2026-03-27
Add class method signatures to TransalationContext 

thanhnguyen-aws:
*Description of changes:* This PR add Class method signatures into TransalationContext, so inputs (positional & keyword args) of the calls of those methods are parsed correctly.

### DIFF
--- a/Strata/DL/Util/List.lean
+++ b/Strata/DL/Util/List.lean
@@ -29,6 +29,23 @@ def dedup {α : Type} [DecidableEq α] : List α → List α
     if a ∈ as then as else a :: as
 
 /--
+Tail-recursive worker for `dedup`. Walks the input left-to-right,
+skipping elements that still appear later, and collects kept elements
+in reverse order.
+-/
+def dedupTR.go {α : Type} [DecidableEq α] :
+    List α → List α → List α
+  | [], acc => acc.reverse
+  | a :: as, acc =>
+    if a ∈ as then dedupTR.go as acc else dedupTR.go as (a :: acc)
+
+/--
+Tail-recursive implementation of `dedup`.
+-/
+def dedupTR {α : Type} [DecidableEq α] (l : List α) : List α :=
+  dedupTR.go l []
+
+/--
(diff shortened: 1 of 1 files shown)


joehendrix (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1564,7 +1566,11 @@ def translateMethod (ctx : TranslationContext) (className : String)
           match arg with
           | .mk_arg _ paramName _paramAnnotation _ =>
             inputs := inputs ++ [{name := paramName.val, type := AnyTy}]
-
+    match args with
+      | .mk_arguments _ _ _ _ _ _ kwargs _ =>
+          if kwargs.val.isSome then
+            let paramType := mkCoreType PyLauType.DictStrAny
```
shigoel: The kwargs parameter is always added with the hardcoded name "kwargs" regardless of the actual Python parameter name. In the test file the parameter is **args, so if a method body referenced args it would be inaccessible — there is no local variable copy created for it (kwargs is excluded from nonSelfParams). This works for the current test only because both methods ignore their kwargs parameter entirely.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1859,12 +1866,14 @@ def pythonToLaurel' (info : PreludeInfo)
         (init := info.importedSymbols) fun m name =>
           m.insert name (ImportedSymbol.compositeType name)
       let initCtx : TranslationContext := {
+        functionSignatures := info.functionSignatures
         preludeTypes := info.types,
         importedSymbols := localSymbols,
```
shigoel, line 1876: allClassFuncDecls is only appended to functionSignatures in the prev_ctx = none branch. When prev_ctx is some, the class method signatures collected during the first pass of this call are silently discarded, so keyword-arg calls to class methods from those classes would be unresolvable in any incremental or multi-pass translation scenario.

PR #699, opened 2026-03-27
Fix unsupported regex patterns causing cvc5 theory-combination errors

shigoel:
## Summary

- **Root cause:** Patterns using unsupported features (`\S`, `\d`, `\w`, `\t`, non-greedy quantifiers, lookaheads, etc.) left `re_*_str` (RegLan-typed) as uninterpreted UFs in the SMT encoding. cvc5 rejects these with `"Regular expression terms are not supported in theory combination"`, which aborted all verification results with an internal error.

- **Fix 1 — `re_*_bool` as factory functions:** Promoted `re_fullmatch_bool`, `re_match_bool`, `re_search_bool` from `inline` prelude functions to factory functions. Their `concreteEval` returns `Str.InRegEx(s, compiled_regex)` on success, or `.none` on failure — leaving an uninterpreted **Bool** UF instead of a RegLan UF. An uninterpreted Bool UF produces `unknown` gracefully (sound over-approximation).

- **Fix 2 — `parseCharClass` escape handling:** Backslash escape sequences inside character classes (e.g. `[a-zA-Z_\.\-/0-9:]`) were misread: `\-` triggered a spurious range check, causing a `patternError` and leaving the pattern uninterpreted. Added escape handling to `parseCharClass` mirroring the top-level atom parser.

### DIFF
--- a/Strata/Languages/Python/PyFactory.lean
+++ b/Strata/Languages/Python/PyFactory.lean
@@ -82,7 +82,7 @@ private def mkModeBoolFunc (name : String) (mode : MatchMode) :
           | [LExpr.strConst () pattern, sExpr] =>
             let (regexExpr, maybe_err) := pythonRegexToCore pattern mode
             match maybe_err with
-            | none => .some (LExpr.mkApp () (.op () "Str.InRegEx" (some mty[bool])) [sExpr, regexExpr])
+            | none => .some (LExpr.mkApp () (.op () "Str.InRegEx" (some mty[string → (regex → bool)])) [sExpr, regexExpr])
             | some _ => .none
           | _ => .none)
       }


Strata/Languages/Python/PythonLaurelCorePrelude.lean
```diff
@@ -108,6 +108,13 @@ function re_fullmatch_str(pattern : string) : regex;
 function re_match_str(pattern : string) : regex;
 function re_search_str(pattern : string) : regex;
```
MikaelMayer: The `re_fullmatch_str`, `re_match_str`, and `re_search_str` forward declarations (lines 107–109) are no longer referenced by any code — the inline functions that used them have been removed, and the factory no longer registers them. These appear to be dead declarations now. The comment on lines 104–106 ("needed so the inline functions after CoreOnlyDelimiter can reference these") is also stale.

Separately, the architecture comment block further down (lines 136–142) still describes the old design: "The mode-specific factory functions re_fullmatch_str, re_match_str, re_search_str each compile a pattern string to a regex" and "The _bool helpers call the mode-specific factories". Neither statement is true anymore — the `_bool` functions *are* the factories now. Worth updating to match the current design.

Strata/Languages/Python/PyFactory.lean
```diff
@@ -66,28 +73,29 @@ open Lambda LTy.Syntax LExpr.SyntaxMono
 -- Mode-specific regex compilation.  Each function compiles a Python regex
 -- string with the correct MatchMode so that anchors (^/$) are handled
 -- properly.
-private def mkModeCompileFunc (name : String) (mode : MatchMode) :
+-- Bool-valued factory.  See architecture comment above.
```
MikaelMayer, line 73: Nit: the comment on lines 73–75 is a leftover from the old `mkModeCompileFunc` — it describes "mode-specific regex compilation" returning a regex. It's not wrong per se, but it reads as describing a different function. Since line 76 already says "Bool-valued factory. See architecture comment above.", the older comment could be trimmed or merged to avoid the impression of two separate concerns.

shigoel: Good call. Fixed in https://github.com/strata-org/Strata/pull/699/commits/3a52d992f1547e3e0345ffb8100737cb0158bdbf.

Strata/Languages/Python/PythonLaurelCorePrelude.lean
shigoel: Thanks! Fixed in https://github.com/strata-org/Strata/pull/699/commits/3a52d992f1547e3e0345ffb8100737cb0158bdbf.

Strata/Languages/Python/PyFactory.lean
```diff
@@ -66,28 +73,29 @@ open Lambda LTy.Syntax LExpr.SyntaxMono
...
+      output := mty[bool],
       concreteEval := some
         (fun _ args => match args with
-          | [LExpr.strConst () s] =>
-            let (expr, maybe_err) := pythonRegexToCore s mode
+          | [LExpr.strConst () pattern, sExpr] =>
+            let (regexExpr, maybe_err) := pythonRegexToCore pattern mode
             match maybe_err with
-            | none => .some expr
+            | none => .some (LExpr.mkApp () (.op () "Str.InRegEx" (some mty[bool])) [sExpr, regexExpr])
```
joscoh: Shouldn't the type of the `.op` here be `string -> string -> bool`, not `bool`?

Side note: we should really be typechecking the `concreteEval`.

shigoel: Whew, good catch and thanks for the diagnosis. I'll open an issue to type check `concreteEval`.

shigoel: Issue filed. https://github.com/strata-org/Strata/issues/708

shigoel: Comment addressed in https://github.com/strata-org/Strata/pull/699/commits/a2db3dcc57186cd3ff1c551c4721962be2d93810.

Strata/Languages/Python/Regex/ReParser.lean
```diff
@@ -101,6 +127,25 @@ def parseCharClass (s : String) (pos : String.Pos.Raw) : Except ParseError (Rege
...
+          let iAfterDash := iNext.next s
+          if !iAfterDash.atEnd s && iAfterDash.get? s != some ']' then
+            let some c2 := iAfterDash.get? s | throw (.patternError "Invalid character in range" s iAfterDash)
+            if escapedChar > c2 then
+              throw (.patternError s!"Invalid character range [{escapedChar}-{c2}]: \
+                                      start character '{escapedChar}' is greater than end character '{c2}'" s i)
+            result := some (match result with | none => .range escapedChar c2 | some prev => .union prev (.range escapedChar c2))
+            i := iAfterDash.next s
+            continue
+      result := some (match result with | none => r | some prev => RegexAST.union prev r)
```
MikaelMayer, line 146: Minor: the `result := some (match result with | none => X | some prev => .union prev X)` pattern appears 4 times in this function (lines 143, 146, 160, 165). A small local helper like:
```lean
let addToResult (r : RegexAST) := match result with | none => r | some prev => .union prev r
```
would reduce the repetition and make each branch's intent clearer. Not blocking.

shigoel: Thanks for the suggestion. The helper as written would capture the initial value of result (none) rather than the current value on each iteration, so it would need to take the accumulator as an explicit parameter: addToResult (acc : Option RegexAST) (r : RegexAST). With that fix the call site becomes result := some (addToResult result r), which isn't much shorter and repeats result twice — so I think the inline match stays clearer. Leaving it as is for now.

MikaelMayer (approved):

PR #701, opened 2026-03-28
Add bitvector types (bv8, bv16, bv32, bv64) to Laurel

tautschnig:
Introduce explicit bitvector types in Laurel so that front-ends can use them directly rather than encoding them as constrained integer types.

- Add TBv(size) to HighType with equality support
- Add bv8, bv16, bv32, bv64 grammar rules in LaurelGrammar.st
- Translate to Core's LMonoTy.bitvec in LaurelToCoreTranslator
- Add formatting and fix exhaustive match in ToLaurelTest


Strata/Languages/Laurel/Grammar/LaurelGrammar.st
```diff
@@ -9,6 +9,10 @@ op boolType : LaurelType => "bool";
 op realType : LaurelType => "real";
 op float64Type : LaurelType => "float64";
 op stringType : LaurelType => "string";
+op bv8Type : LaurelType => "bv8";
+op bv16Type : LaurelType => "bv16";
+op bv32Type : LaurelType => "bv32";
+op bv64Type : LaurelType => "bv64";
```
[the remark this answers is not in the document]
tautschnig: The fixed set matches Core's predefined widths and is the right level of abstraction for now.

Strata/Languages/Laurel/LaurelFormat.lean
```diff
@@ -52,6 +52,7 @@ def formatHighTypeVal : HighType → Format
   | .TReal => "real"
   | .TString => "string"
   | .THeap => "Heap"
+  | .TBv n => s!"bv{n}"
```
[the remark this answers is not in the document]
tautschnig: The current formatting is fine. Unsupported widths can only arise programmatically, and the formatter is primarily for debugging/display, not round-tripping.

Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
```diff
@@ -92,6 +92,10 @@ partial def translateHighType (arg : Arg) : TransM HighTypeMd := do
     | q`Laurel.float64Type, _ => return mkHighTypeMd .TFloat64 md
     | q`Laurel.realType, _ => return mkHighTypeMd .TReal md
     | q`Laurel.stringType, _ => return mkHighTypeMd .TString md
+    | q`Laurel.bv8Type, _ => return mkHighTypeMd (.TBv 8) md
+    | q`Laurel.bv16Type, _ => return mkHighTypeMd (.TBv 16) md
+    | q`Laurel.bv32Type, _ => return mkHighTypeMd (.TBv 32) md
+    | q`Laurel.bv64Type, _ => return mkHighTypeMd (.TBv 64) md
     | q`Laurel.coreType, #[.ident _ name] => return mkHighTypeMd (.TCore name) md
```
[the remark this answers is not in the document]
tautschnig: Fixed — replaced the hardcoded type list with a generic message that won't go stale as new types are added.

Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
```diff
@@ -92,6 +92,10 @@ partial def translateHighType (arg : Arg) : TransM HighTypeMd := do
     | q`Laurel.float64Type, _ => return mkHighTypeMd .TFloat64 md
     | q`Laurel.realType, _ => return mkHighTypeMd .TReal md
     | q`Laurel.stringType, _ => return mkHighTypeMd .TString md
+    | q`Laurel.bv8Type, _ => return mkHighTypeMd (.TBv 8) md
```
keyboardDrummer: Could we abstract over the amount of bits? Also, I think the right default SMT encoding is to use an `int` with constraints, not an SMT bitvector. Until Core supports that, can we really define this with the correct verification behavior in Laurel? Right now I think the front-ends can't use these types because of the default SMT bitvector behavior, so it seems confusing to add them.

Also, should we think about this end-to-end? How will users indicate what SMT encoding they want to use for particular variables? Can we infer the probably best encoding based on which operators the variables are applied to? My guess would be that a good heuristic is that if no bit specific operators are used, use an int and otherwise a bv.

tautschnig: > Could we abstract over the amount of bits?

In principle we could, but then Core has hard-coded bit widths only (I don't know why). So it would require a change in Core first.

> Also, I think the right default SMT encoding is to use an `int` with constraints, not an SMT bitvector.

Why do you believe this to be the case?

> Until Core supports that, can we really define this with the correct verification behavior in Laurel? Right now I think the front-ends can't use these types because of the default SMT bitvector behavior, so it seems confusing to add them.

I have a changed queued up to add transforms between constrained types and bitvectors, but then it only makes sense to add that if we have tests (#656) and language support (this PR) in the first place.

> Also, should we think about this end-to-end? How will users indicate what SMT encoding they want to use for particular variables? Can we infer the probably best encoding based on which operators the variables are applied to? My guess would be that a good heuristic is that if no bit specific operators are used, use an int and otherwise a bv.

I'd love to conduct all these experiments, but it seems impossible to do them when we don't have the necessary infrastructure in place first.

keyboardDrummer: > > Could we abstract over the amount of bits?
> 
> In principle we could, but then Core has hard-coded bit widths only (I don't know why). So it would require a change in Core first.

Okay. What I see from this PR is that Core has `LMonoTy.bitvec n`. In what way does it only support hard-coded bit widths?

> > Also, I think the right default SMT encoding is to use an `int` with constraints, not an SMT bitvector.
> 
> Why do you believe this to be the case?

I think the common case is one where the user-program wants to prove that the integer operations are not overflowing. After knowing that, it wants to prove other properties about these now effectively unbounded integers. To reason about unbounded integers, it seems simplest to use lemma's that operate on these instead of converting the integers to bit arrays and using lemma's for those.

For cases where you want to rely on bit level properties like bit shifting or bit-wise logical operators, I'm guessing that using bv SMT theory works better. For cases where the user wants to rely on overflows, I'm not sure what the best approach is, bvs or NLA with `%`.

> > Until Core supports that, can we really define this with the correct verification behavior in Laurel? Right now I think the front-ends can't use these types because of the default SMT bitvector behavior, so it seems confusing to add them.
> 
> I have a changed queued up to add transforms between constrained types and bitvectors, but then it only makes sense to add that if we have tests (#656) and language support (this PR) in the first place.

The pipeline I think we should have is:
1. Front-ends use Laurel BV types for any integers that fit them. They use constrained types only for properties that BV types can't handle, like a number being non-negative.
2. Somewhere in the Core->SMT pipeline, an analysis determines whether Core BV types should use SMT integer or BV theory.

Without (2), I don't see how we can do (1). Also, I don't see a use-case for translating Laurel constrained integers to BV types, since those constrained types should only be used for things that can't be expressed using a BV type anyways, like `nat`.

atomb: I can provide some context here. The DDM makes it difficult to define a type that's parameterized by a natural (at least if we're letting the DDM do type checking). So at the syntactic level of a single dialect, we need to choose a specific set of widths. However, I'd hope we might be able to relax that DDM restriction eventually. And, in anticipation of that, and to make all the internal code simpler, I'd prefer to move to a parameterized form as quickly as possible. So I'd support arbitrary-width bit vectors in Laurel's types, even if there's no syntax for most of them right now.

keyboardDrummer: I think on the Laurel side using the DDM for parsing `bv n` wouldn't be a problem. Laurel only uses `category` and `op` and not `type` or `function` from the DDM.

atomb (dismissed):
It'd be nice to have arbitrary bitvector sizes at the Laurel concrete syntax level, but we could add that later, as needed. The fact that arbitrary widths are supported internally is more important.

keyboardDrummer:
It's still unclear to me why we couldn't have arbitrary bit width types at the Laurel level. It seems like at least the Core types support it. @keyboardDrummer-bot can you comment on whether the above is possible? Don't update the PR but feel free to try out this change locally.

tautschnig:
@keyboardDrummer-bot Could you please implement what you experimented with and push this change onto the branch? Thank you!

keyboardDrummer:
@keyboardDrummer-bot please resolve the merge conflicts as well

tautschnig:
@keyboardDrummer-bot It seems there are test failures after your merge from main.

keyboardDrummer (approved):

atomb (approved):
Love it!

PR #704, opened 2026-03-30
Defining invariant that CachedAnalysis must hold, update PrecondElim

### LINKED ISSUE #702: Defining invariant that CachedAnalysis must hold, update PrecondElim
Once PrecondElim is updated, we can make https://github.com/strata-org/Strata/pull/687 use FilterProcedure too.

MikaelMayer:
Fixes #702

## Problem

`PrecondElim` generates WF (well-formedness) checking procedures but did not update the cached call graph. This meant any downstream transform relying on the call graph would see stale data after `PrecondElim`.

## Solution

- Document the `CachedAnalyses` invariant: when the call graph is present, it must reflect the current program's procedure call structure. Transforms must update it or invalidate it.
- Add `CallGraph.addLeafNode` for inserting nodes with no callees (updates both `callees` and `callers` maps to keep their key sets consistent).
- Update `PrecondElim` to register generated precondition-WF procedures as leaf nodes in the call graph (they contain only assert/assume statements and make no procedure calls).
- Replace the manual post-transform `keepSet` filtering in the selective verification pipeline with a second `FilterProcedures.run` pass. Since `CallElim` maintains the call graph precisely (via `decrementEdge`) and `PrecondElim` now registers WF procs as leaf nodes with `noFilter := true`, the call-graph closure is accurate after transforms and `FilterProcedures` works correctly — no manual filtering needed.
- Extend `filterProceduresPipelinePhase` with a `respectNoFilter` parameter so both the initial and post-transform filter phases reuse the same helper.

## Testing

All existing tests pass, including selective verification, NoFilterWFProc, CallElim, and InlineAssertionMetadata tests.


MikaelMayer:
Acknowledged — the PR description now accurately scopes the work to `PrecondElim` only and notes that `CallElim` maintaining the invariant is a follow-up. No further changes needed.

Strata/Transform/PrecondElim.lean
```diff
@@ -277,6 +277,14 @@ end
 
 /-! ## Main transformation -/
 
+/-- Add a WF procedure as a leaf node in the cached call graph.
+WF procedures contain only assert/assume statements and make no procedure calls. -/
+private def addWFProcToCallGraph (name : String) : CoreTransformM Unit :=
+  modify fun σ => match σ.cachedAnalyses.callGraph with
+  | .some cg => { σ with cachedAnalyses := { σ.cachedAnalyses with
+      callGraph := .some (cg.addLeafNode name) } }
```
MikaelMayer, line 297: The `.proc` and `.func` cases both call `addWFProcToCallGraph` for their generated WF procedures, but the `recFuncBlock` case (around line 342) does not. It generates WF procedures via `funcs.filterMap (mkFuncWFProc F' · md)` and adds them to the program, but never registers them in the call graph. This violates the invariant documented in `CachedAnalyses`.

The fix would be to iterate over the generated WF decls and call `addWFProcToCallGraph` for each, e.g.:
```lean
for f in funcs do
  if (mkFuncWFProc F' f md).isSome then
    addWFProcToCallGraph (wfProcName (CoreIdent.toPretty f.name))
```

MikaelMayer (commented):
Clean, well-scoped PR. The invariant documentation is clear and the `addLeafNode` / `addWFProcToCallGraph` pattern is easy to follow. One issue: the `recFuncBlock` case is missing the call graph update.

Strata/Transform/PrecondElim.lean
MikaelMayer: Good catch. Added `addWFProcToCallGraph` calls in the `recFuncBlock` case by switching from `filterMap` to `filterMapM`, calling `addWFProcToCallGraph` for each function that produces a WF procedure.

MikaelMayer:
Hey, do you have a test that fails once FilterProcedures.run prog (procs ++ procs.map PrecondElim.wfProcName) is used instead of it....? Your comment said CallElim has to be updated to correctly update call graph as well, but CallElim's one is (supposed to be) precise, modulo any bug introduced by human because it hasn't been verified.

MikaelMayer:
Acknowledged. The FilterProcedures.run replacement is in place (commit e024ab6b), the build passes, and all tests pass. No further code changes needed — the CI failures were indeed unrelated DDM clang errors. Updating the PR description to reflect the final state.

MikaelMayer:
The HEAD had been reverted to manual keepSet filtering (commit 019a6198) after your confirmation. I've re-applied `FilterProcedures.run` — build passes and all relevant tests pass (SelectiveVerification, NoFilterWFProc, InlineAssertionMetadataTest, CallElimWithOldGlobals, CallElim). Committing and pushing now.

MikaelMayer:
Where are we now on that? Can we now use FilterProcedure instead of keepset?

MikaelMayer:
Yes! Removed `keepSetFilterPipelinePhase` entirely and replaced it with a second `filterProceduresPipelinePhase` pass after `CallElim` and `PrecondElim`. This works because:

- `CallElim` maintains the call graph precisely via `decrementEdge`
- `PrecondElim` registers WF procedures as leaf nodes with `noFilter := true`
- `FilterProcedures.run` respects `noFilter` and uses call-graph closure

All tests pass (SelectiveVerification, NoFilterWFProc, CallElim, CallElimWithOldGlobals, InlineAssertionMetadataTest).

Strata/Languages/Core/Verifier.lean
```diff
@@ -681,13 +678,17 @@ def corePipelinePhases (procs : Option (List String) := none)
...
     | some f => [precondElimPipelinePhase f]
     | none => []
-  let keepSetPhase := match procs with
-    | some ps => [keepSetFilterPipelinePhase ps]
+  let postFilterPhases := match procs with
+    | some ps =>
+      let targets := ps ++ ps.map PrecondElim.wfProcName
+      [modelPreservingPipelinePhase "FilterProcedures" fun prog => do
+        let filtered ← FilterProcedures.run prog targets (respectNoFilter := false)
+        return (true, filtered)]
```
MikaelMayer: I'm puzzled by this, because just above we have `[filterProceduresPipelinePhase ps]` and here we have a complicated
` [modelPreservingPipelinePhase "FilterProcedures" fun prog => do
        let filtered ← FilterProcedures.run prog targets (respectNoFilter := false)
        return (true, filtered)]`
I wasn't expecting such inline code to be complicated. I thought we could just reuse the same phase? If not why not what is failing? If yes please reuse the same phase.

MikaelMayer: Good catch. I added a `respectNoFilter` parameter to `filterProceduresPipelinePhase` and now the post-filter phase reuses it:

```lean
[filterProceduresPipelinePhase targets (respectNoFilter := false)]
```

instead of the inline `modelPreservingPipelinePhase` block.

Strata/Transform/PrecondElim.lean
MikaelMayer: 🤖 Confirmed — the `recFuncBlock` case already uses `filterMapM` and calls `addWFProcToCallGraph` for each generated WF procedure (lines 359-363). No additional code changes needed.

Strata/Languages/Core/CallGraph.lean
```diff
@@ -64,6 +64,10 @@ def CallGraph.decrementEdge (cg : CallGraph) (caller : String) (callee : String)
     callers := new_callers
   }
 
+/-- Add a node with no callees to the call graph. -/
+def CallGraph.addLeafNode (cg : CallGraph) (name : String) : CallGraph :=
+  { cg with callees := cg.callees.insert name {} }
```
aqjune-aws: Could you update `callers` to have an empty entry of `name` as well? Keys of both of the two HashMaps should contain all procedures/functions. :)

Strata/Transform/PrecondElim.lean
```diff
@@ -288,6 +288,14 @@ end
 
 /-! ## Main transformation -/
 
+/-- Add a WF procedure as a leaf node in the cached call graph.
+WF procedures contain only assert/assume statements and make no procedure calls. -/
```
atomb: It's a minor nit, but I think this isn't the only notion of well-formed that it makes sense to have for procedures. It's useful for deductive verification, but other analyses would want to consider many procedures containing calls to be well-formed.

Strata/Languages/Core/CallGraph.lean
MikaelMayer: 🤖 Good catch! Updated `addLeafNode` to also insert an empty entry in the `callers` map, so both HashMaps maintain the same key set.

Strata/Transform/PrecondElim.lean
MikaelMayer: 🤖 Fair point. Updated the docstring to say "precondition-WF procedure" to clarify this is specifically about precondition well-formedness for deductive verification, not a universal notion.

aqjune-aws (approved):

atomb (approved):

PR #705, opened 2026-03-30
Optimize List.dedup

### LINKED ISSUE #703: Optimize List.dedup
List.dedup needs to be rewritten as a tail-recursive or iterative implementation to handle large lists.
And we should prove that the new implementation does the same deduplication.

MikaelMayer:
Fixes #703

`List.dedup` was implemented with non-tail recursion, causing stack overflow on large lists. This PR adds a tail-recursive implementation (`dedupTR`) that uses an accumulator to process elements iteratively.

The original `dedup` definition is preserved unchanged so all existing proofs continue to work. The tail-recursive version is wired in at runtime via `@[csimp]`. An equivalence proof (`dedupTR_eq_dedup`) verifies that the new implementation produces the same result as the original.

Tested: existing tests pass, no new test failures introduced.

### DIFF
--- a/DiffTestCore.lean
+++ b/DiffTestCore.lean
@@ -104,12 +104,12 @@ def checkMatch (pyRegex testStr : String) (mode : MatchMode)
       let vcResults ← verify pgm inputCtx none .quiet
       match vcResults[0]? with
       | none    => return .smtError "no VCs generated"
-      | some vc => return match vc.outcome with
-        | .ok o => 
-          if o.isPass then .match 
-          else if o.alwaysFalseAndReachable || o.canBeTrueOrFalseAndIsReachable then .noMatch 
-          else .smtError "unknown"
-        | .error msg => .smtError s!"impl: {msg}"
+      | some vc =>
+          if vc.isSuccess then return .match
+          else if vc.isFailure then return .noMatch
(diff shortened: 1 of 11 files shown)


Strata/DL/Util/List.lean
```diff
@@ -110,6 +127,29 @@ theorem mem_of_dedup {α : Type} [DecidableEq α]
...
+      simp [h, h', ih]
+
+/--
+`dedupTR` is equivalent to `dedup`.
+-/
+private theorem dedupTR_eq_dedup {α : Type} [DecidableEq α]
+    (l : List α) : dedupTR l = l.dedup := by
+  simp [dedupTR, dedupTR.go_eq]
+
+attribute [implemented_by dedupTR] List.dedup
```
joscoh: I know this is still a draft, but I think we should use `@[csimp]` which is safer than `implemented_by`.

MikaelMayer: Switched from `implemented_by` to `@[csimp]`. This required making `dedupTR`, `dedupTR.go`, and `dedupTR.go_eq` public since `@[csimp]` requires the theorem to be public. The equivalence theorem now has the form `@List.dedup = @dedupTR` as required by `csimp`.

PR #707, opened 2026-03-30
Inlining assertions should keep metadata

### LINKED ISSUE #706: Inlining assertions should keep metadata
Currently in procedure inlining, we keep the metadata of assertions.
However, we need to combine the two metadata, one for the original source, and one for the original place where the procedure was called.
That way, the reported error can be primarily at the procedure call, and the secondary position would be the original assert metadata. The procedure call metadata is the most important to report. There are already a few tests that use procedure inlining in python with pyspec. We should update their output so that we demonstrate that we display the metadata of the call site first but with the error message of the original assertion. We can then display related positions.

MikaelMayer:
Fixes #706

During procedure inlining, assertions in the callee body now carry the call site metadata as the primary file range, with the original assertion location stored as a related file range. This allows error reporting to show the procedure call location first, with the original assertion as a related position.

- Added `relatedFileRange` metadata field and `setCallSiteFileRange` helper to combine call site and original assertion metadata
- Updated `ProcedureInlining.inlineCallCmd` to set call site metadata on inlined assertions, assumes, and covers
- Display related positions in `pyAnalyze` and `pyAnalyzeLaurel` output
- Added `RelatedLocation` to SARIF output for related positions
- Updated SARIF test expected outputs for the new `relatedLocations` field

Tested: existing tests pass, SARIF test outputs updated.

### DIFF
--- a/Strata/Util/Sarif.lean
+++ b/Strata/Util/Sarif.lean
@@ -103,7 +103,9 @@ structure Result where
   level : Level
   message : Message
   locations : Array SarifLocation := #[]
-  /-- Related locations (e.g., original assertion location when inlined) -/
+  /-- Related locations (e.g., original assertion location when inlined).
+      Order follows the call stack: the innermost (most deeply inlined) location
+      comes first. Each element's `id` field is 1-indexed and matches its position. -/
   relatedLocations : Array RelatedLocation := #[]
   /-- Tool-specific properties (SARIF property bag) -/
   properties : PropertyBag := {}


Strata/DL/Imperative/MetaData.lean
```diff
@@ -250,6 +250,29 @@ def MetaData.formatFileRangeD {P : PureExpr} [BEq P.Ident] (md : MetaData P) (fi
   | some fr => fr.format fileMap includeEnd?
   | none => f!""
 
+/-- Metadata field for a related file range (e.g., the original assertion location
+    when the primary file range points to the call site after inlining). -/
+def MetaData.relatedFileRange : MetaDataElem.Field P := .label "relatedFileRange"
```
MikaelMayer, line 257: There might be more than one inlining, so related file ranges should be ordered. For every inlining, the related file ranges should shift as well.

Strata/DL/Imperative/MetaData.lean
```diff
@@ -250,6 +250,29 @@ def MetaData.formatFileRangeD {P : PureExpr} [BEq P.Ident] (md : MetaData P) (fi
...
+
+/-- Get the related file range from metadata, if present. -/
+def getRelatedFileRange {P : PureExpr} [BEq P.Ident] (md: MetaData P) : Option FileRange := do
```
MikaelMayer: There might be more than one related file range.

StrataTest/Languages/Core/Tests/SarifOutputTests.lean
```diff
@@ -293,23 +293,23 @@ def makeVCResult (label : String) (outcome : VCOutcome)
...
   let vcResults : VCResults := #[makeVCResult "test_unknown" (VCOutcome.mk .unknown .unknown)]
   let sarif := vcResultsToSarif .deductive files vcResults
   Strata.Sarif.toJsonString sarif
 
-/-- info: "{\"runs\":[{\"results\":[{\"level\":\"error\",\"locations\":[],\"message\":{\"text\":\"Unknown (solver timeout or incomplete)\"},\"properties\":{\"propertyType\":\"assert\"},\"ruleId\":\"test_error\"}],\"tool\":{\"driver\":{\"informationUri\":\"https://github.com/strata-org/Strata\",\"name\":\"Strata\",\"version\":\"0.1.0\"}}}],\"schema\":\"https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json\",\"version\":\"2.1.0\"}" -/
+/-- info: "{\"runs\":[{\"results\":[{\"level\":\"error\",\"locations\":[],\"message\":{\"text\":\"Unknown (solver timeout or incomplete)\"},\"properties\":{\"propertyType\":\"assert\"},\"relatedLocations\":[],\"ruleId\":\"test_error\"}],\"tool\":{\"driver\":{\"informationUri\":\"https://github.com/strata-org/Strata\",\"name\":\"Strata\",\"version\":\"0.1.0\"}}}],\"schema\":\"https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json\",\"version\":\"2.1.0\"}" -/
```
MikaelMayer, line 311: Nice to update this but I would like to see a test that displays related locations. Look into the python specification tests, there is a place where we inline a procedure. Look at StrataTest/Languages/Python/AnalyzeLaurelTest.lean  and messages like `Storage_Storage_put_item_assert(0)_9: ✔️ always true if reached (Required parameter 'Bucket' is missing)`.
First you should update the messages to include line and column number and remove the assertion names that are not human-readable. Then verify that with your approach, it indicates the assertion's line and column number to be at the call site.

Strata/DL/Imperative/MetaData.lean
MikaelMayer: Updated `setCallSiteFileRange` to support multiple inlinings. It now collects existing related file ranges, clears them, prepends the current primary range as the first related range, then appends the previously existing related ranges. This preserves ordering: the most recent inlining's original location comes first, followed by earlier ones.

Strata/DL/Imperative/MetaData.lean
MikaelMayer: Changed `getRelatedFileRange` (returning `Option FileRange`) to `getRelatedFileRanges` (returning `Array FileRange`), which collects all metadata elements with the `relatedFileRange` field. Updated all callers: SARIF output now emits all related locations with sequential IDs, and the CLI output formats all related positions.

StrataTest/Languages/Core/Tests/SarifOutputTests.lean
MikaelMayer: Added a SARIF test that constructs metadata with `setCallSiteFileRange` (simulating inlining) and verifies the related location appears in the SARIF JSON output with the correct line/column.

Regarding the Python spec test updates (updating messages to include line/column, removing non-human-readable assertion names, and verifying call site location through the full pipeline) — this is a larger change that touches the Python pipeline output formatting. I can tackle that as a follow-up if you'd like, or address it in this PR with more guidance on the expected output format.

MikaelMayer: The change would likely touch only one file which is the one containing the line I told you above. I'm not talking about changing the output of every python file. What i would like is to see the above transformed to
```
file(line,column): ✔️ always true if reached (Required parameter 'Bucket' is missing). Related locations: file(line,column) file(line,column)
```
etc.

MikaelMayer: We forgot to address this comment.

StrataTest/Languages/Core/Tests/SarifOutputTests.lean
```diff
@@ -323,4 +330,18 @@ def makeVCResult (label : String) (outcome : VCOutcome)
...
+#guard_msgs in
+#eval
+  -- Simulate inlining: original assertion at byte 0 (line 1), call site at byte 13 (line 2)
+  let origMd := makeMetadataAt "/test/caller.st" 0
+  let callSiteMd := makeMetadataAt "/test/caller.st" 13
+  let inlinedMd := origMd.setCallSiteFileRange callSiteMd
+  let files := makeFilesMap "/test/caller.st"
+  let vcResults : VCResults := #[makeVCResult "inlined_assert" (VCOutcome.mk .unsat (.sat [])) inlinedMd]
+  let sarif := vcResultsToSarif .deductive files vcResults
+  Strata.Sarif.toJsonString sarif
```
MikaelMayer, line 345: We can keep this unit test, but what I want to see is an actual end-to-end test that will ACTUALLY perform inlining of an assertion and verify that the result is printed WITH the relative location. Not SARIF output, regular output.

MikaelMayer: Added an end-to-end test that:
1. Defines a Core program with `callee` (containing `assert false`) and `caller` (which calls `callee`)
2. Translates via `Core.getProgram`, inlines via `Core.inlineProcedures`, verifies via `Core.verifyProgram`
3. Checks that the VCResult for the inlined assertion has non-empty `relatedFileRanges` metadata, confirming the original assertion location is preserved through the full pipeline

The test output shows:
```
willFail: ❌ fail
  Related location: original assertion in callee
```

StrataTest/Languages/Core/Tests/SarifOutputTests.lean
```diff
@@ -344,4 +345,40 @@ def makeVCResult (label : String) (outcome : VCOutcome)
...
+    EIO.toIO (fun e => IO.Error.userError e)
+      (Strata.Core.verifyProgram inlined
+        { Core.VerifyOptions.default with verbose := .quiet }
+        (proceduresToVerify := some ["caller"]))
+  let mut output := ""
+  for vcr in vcResults do
+    output := output ++ s!"{vcr.obligation.label}: {vcr.formatOutcome}"
+    let related := Imperative.getRelatedFileRanges vcr.obligation.metadata
+    if !related.isEmpty then
+      output := output ++ s!"\n  Related location: original assertion in callee"
```
MikaelMayer: This is not a Sarif test. Find out where to place it elsewhere.
Also, we don't see the offset. I think you should compute the offset difference and display it at least so that we know that there are two reporting locations relative to each other in this test.

MikaelMayer: Moved the end-to-end inlining test to a new file `InlineAssertionMetadataTest.lean`. The test now displays both the call site offset and the original assertion offset along with their delta, so we can see the two distinct reporting locations are at different positions (e.g. call site at 622, original assertion at 559, delta 63).

StrataTest/Languages/Core/Tests/InlineAssertionMetadataTest.lean
```diff
@@ -0,0 +1,57 @@
...
+program Core;
+procedure callee() returns () {
+  assert [willFail]: false;
+};
+procedure caller() returns () {
+  call callee();
+};
+#end
+
+/-- info: "willFail: ❌ fail\n  Call site offset: 622, Original assertion offset: 559 (delta: 63)" -/
```
MikaelMayer: I want this to be
```suggestion
/-- info: "willFail: ❌ fail\n Assertion is 63 characters after the related location" -/
```

MikaelMayer: Updated the output format to show `Assertion is 63 characters after the related location` instead of raw offsets. Build passes.

Strata/Util/Sarif.lean
```diff
@@ -89,13 +89,22 @@ structure PropertyBag where
...
+
 /-- SARIF result representing a single verification result -/
 structure Result where
   /-- Stable identifier of the rule that was evaluated to produce the result --/
   ruleId : String
   level : Level
   message : Message
   locations : Array SarifLocation := #[]
+  /-- Related locations (e.g., original assertion location when inlined) -/
+  relatedLocations : Array RelatedLocation := #[]
```
aqjune-aws, line 109: Could you leave a short comment explaining whether the order of elements in this array is important, and if true, how is the order determined?

Strata/DL/Imperative/MetaData.lean
```diff
@@ -250,6 +250,40 @@ def MetaData.formatFileRangeD {P : PureExpr} [BEq P.Ident] (md : MetaData P) (fi
   | some fr => fr.format fileMap includeEnd?
   | none => f!""
 
+/-- Metadata field for a related file range (e.g., the original assertion location
+    when the primary file range points to the call site after inlining). -/
+def MetaData.relatedFileRange : MetaDataElem.Field P := .label "relatedFileRange"
```
aqjune-aws, line 257: Let's leave a comment here mentioning that there can be multiple `MetaData.relatedFileRange` fields due to multiple inlinings.

Strata/DL/Imperative/MetaData.lean
```diff
@@ -250,6 +250,40 @@ def MetaData.formatFileRangeD {P : PureExpr} [BEq P.Ident] (md : MetaData P) (fi
...
+
+/-- Get all related file ranges from metadata, in order. -/
```
aqjune-aws: Could you comment that the returned Array's order is determined by the call stack and where the innermost call is placed in the array?

Strata/DL/Imperative/MetaData.lean
MikaelMayer: Added a comment to the docstring mentioning that there can be multiple `relatedFileRange` fields due to multiple levels of inlining.

Strata/DL/Imperative/MetaData.lean
MikaelMayer: Added a comment to the docstring explaining that the array order is determined by the call stack, with the innermost (most deeply inlined) call first.

Strata/Util/Sarif.lean
MikaelMayer: Added a docstring comment explaining that the order follows the call stack (innermost/most deeply inlined location first) and that each element's `id` is 1-indexed matching its position.

PR #711, opened 2026-03-30
Fix #697, improve handling of precondition obligations

### LINKED ISSUE #697: Improve handling of precondition checking
Currently, the precondition checking in expressions follows the following rules:
1. given `p => e`, all precondition-assertions in `e` can assume `p`
2. given `if p then e1 else e2`, all precondition assertions in `e1` can assume `p` and all in `e2` can assume `not p`

We would like to add:
3. given `p || e`, all precondition assertions in `e` can assume `not p`
4. given `p /\ e`, all precondition assertions in `e` can assume `p`

And likely (using a lambda-based encoding for `let`):
5. given `let x := e1 in e`, preconditions in `e1` are added only at the point where `x` is used in `e` (which can be in a context including some of the assumptions from the previous rules)

joscoh:
*Issue #, if available:* #697

*Description of changes:* 
When generating function precondition proof obligations, Strata Core now:
- Short circuits by generating assumptions from boolean operators (not just "implies" as previously)
- Handles let-bindings lazily, including assumptions. This means that e.g. `let x := 3 / 0 in true` generates no proof obligations, and `let x := 3 / 0 in true || (x > 0)` generates a trivial one.

Includes unit tests in `PreconditionsTests.lean` and a short-circuiting example in `FunctionPreconditions.lean`.


Strata/DL/Lambda/Preconditions.lean
```diff
@@ -64,16 +66,21 @@ For each call to a function with preconditions:
...
 where
   /-- Wrap an obligation with accumulated implications -/
   wrapImplications (implications : List (T.Metadata × LExpr T.mono))
       (ob : LExpr T.mono) : LExpr T.mono :=
     implications.foldr (fun (md, lhs) acc =>
       .app md (.app md (@boolImpliesFunc T).opExpr lhs) acc) ob
 
+  /-- Shift all keys in the let-obligations map up by 1 (for entering a binder) -/
+  shiftLetObs (m : Std.HashMap Nat (List (WFObligation T))) : Std.HashMap Nat (List (WFObligation T)) :=
+    m.fold (fun acc k v => acc.insert (k + 1) v) {}
```
aqjune-aws, line 92: Let's leave a comment stating that the key of `m` is De Bruijn index.

Strata/DL/Lambda/Preconditions.lean
aqjune-aws, line 61: Let's briefly describe what kind of guarantee can we get if we discharge the `WFObligation` lists returned by `collectWFObligations`. :)
Something like "for any function call `f(x)` that is a subexpression of `e`, we know that preconditions of `e` is valid, or there is a reduction strategy to not evaluate `f(x)` at all"?

aqjune-aws (dismissed):

Strata/DL/Lambda/Preconditions.lean
```diff
@@ -32,6 +32,8 @@ structure WFObligation (T : LExprParams) where
   callSiteMetadata : T.Metadata
   /-- Metadata from the precondition definition -/
   precondMetadata : T.Metadata
+  /-- Whether this obligation was propagated from a let-bound argument -/
+  fromLetArg : Bool := false
```
MikaelMayer, line 37: `fromLetArg` is internal bookkeeping — it's always `false` in the output of `collectWFObligations` (reset on line 170, or never set). Exposing it on the public `WFObligation` structure means callers see a field that's meaningless to them. Consider either making it a private wrapper type used only inside `go`/the `where` block, or adding a docstring noting it's always `false` in the returned list.

Strata/DL/Lambda/Preconditions.lean
```diff
@@ -89,43 +108,77 @@ where
...
+        (go F body implications (shiftLetObs letObs)).map fun ob =>
           { ob with obligation := .quant md .all name ty trigger ob.obligation }
-      /- If we are on the RHS of an implication, add assumption
-        E.g. y > 0 ==> x / y = 1 should produce
-        y > 0 ==> y != 0 -/
       | .app md (.app _ (.op _ opName _) lhs) rhs =>
+        /- p ==> e: preconditions in e can assume p
+           E.g. y > 0 ==> x / y > 0 should produce
+           y > 0 ==> y != 0 -/
         if opName == (@boolImpliesFunc T).name then
```
MikaelMayer, line 132: The `==>` and `&&` branches are identical in behavior (both add `lhs` as the assumption for the RHS). You could merge them into a single condition:
```
if opName == (@boolImpliesFunc T).name || opName == (@boolAndFunc T).name then
```
with a combined comment explaining both cases. The `||` branch remains distinct since it negates `lhs`. Up to you whether the separate comments justify the duplication.

Strata/DL/Lambda/Preconditions.lean
```diff
@@ -89,43 +108,77 @@ where
...
+           implication context at the usage site of `x`).
+         - Body obligations from calls in the body are wrapped in the let.
+           E.g. `let x := safeDiv(a, b) in safeDiv(c, x)` produces
+           `(let x := safeDiv(a, b) in x != 0)` since the body call
+           `safeDiv(c, x)` generates `x != 0` which references `x`.
+       -/
       | .app md (.abs amd name ty body) arg =>
-        let argObs := go F arg implications
-        let bodyObs := (go F body implications).map fun ob =>
+        let argObs := collectWFObligations F arg
```
MikaelMayer, line 167: Noting a subtlety: `collectWFObligations F arg` drops the current `implications` and `letObs`, so arg obligations are collected in a fresh context. This is correct for the lazy semantics (obligations are re-emitted at usage sites with local assumptions), but it also means that if the arg itself contains a function call whose obligation doesn't flow through a bvar (e.g. `let x := (safeDiv(a, b) + safeDiv(c, d)) in x`), both `b != 0` and `d != 0` are collected without outer implications and then re-wrapped at usage. Might be worth a brief inline comment noting that the fresh context is intentional (the docstring above covers it, but a one-liner at the call site would help future readers).

MikaelMayer (commented):
Well-designed PR. The lazy let-binding semantics and short-circuit assumption propagation are cleanly implemented. The test coverage is thorough, covering `||`, `&&`, `==>`, unused let-bound variables, nested lets with body calls, and the bvar shifting logic. A few minor observations below.

Strata/DL/Lambda/Preconditions.lean
joscoh: I think I would prefer to keep separate for readability, but I can change if you feel strongly.

tautschnig (approved):

MikaelMayer (changes requested):
Here is my last assessment and recommendation of the situation.
I think we should not have Core expressions (even with preconditions) creating proof obligations. Core expressions are always total and so are Core functions. The precondition mechanism we have is a way to only specify what the value of a function is if the precondition is satisfied, but we already assign an uninterpreted value if the precondition is not satisfied. So there is no error introduction.

This also means that the phase that removes function preconditions to introduce proof obligations should be done in the part of the pipeline where it's ok to consider preconditions as errors semantically, and where it's possible to insert assertions directly just before the function call. That part seems to be most likely Laurel, which combines statements and expressions seamlessly.

I know that Core leverages these function preconditions in at least two places now: division by zero, datatype destructor ability. Both define two operators functions, one with preconditions and one without.
The problem is that the semantics of Core currently does not trigger errors or assertions on failing preconditions, so the path forward is to either go full mode and accept that expressions can fail, and that should be reflected in the semantics, so that we can prove the soundness of these potential failures when eliminating function preconditions; or my now slightly preferred that we move all these precondition checks to Laurel which is more suited than Core to do a small transformation that eliminate the function preconditions.

You can dismiss my review if others are convinced otherwise, but I think we are all in a state right now where we aren't sure of the current path we are taking, so I wanted to write down my own thoughts.

joscoh:
Agree with Mikael that more discussion about the semantics is needed. Marking as a draft for now until we discuss.

keyboardDrummer:
Hey, I'm not sure I follow all the details in the discussion, but still I think that 

```
function A(): Bool
function B(): Bool requires A

const x := if A() then B() else false
```

should behave the same as:

```
function A(): Bool
function B(): Bool requires A

const x := A() && B()
```

And currently that's not the case for Core.

Would it be possible to only include the first part of this PR, `Short circuits by generating assumptions from boolean operators (not just "implies" as previously)`, @MikaelMayer ?

A resolution of this issue is blocking the correct implementation of short-circuiting `&&` and `||` in Laurel, which we need to replace Dafny for JVerify.

PR #712, opened 2026-03-30
fix: reduce stdout clutter in pyAnalyzeLaurel without --verbose

### LINKED ISSUE #710: stdout is cluttered even without --verbose
For a command like this:
```
lake exe strata pyAnalyzeLaurel \
    --check-mode bugFinding --check-level full \
    --no-solve --vc-directory <PATH_TO_SMTLIB_DUMP> \
    --spec-dir <SPEC_DIR> \
    --dispatch "something" \
    somefile.python.st.ion
```

we get a extremely verbose output.
```
==== Laurel Program ====
...the entire program reprinted...
```

Also I see errors like
```
Untranslatable precondition (emitting nondeterministic assert): Expected len(something) >= 1, got <expr>
```
which does not help. The `<expr>` is not a placeholder, it's literally emitted like that.

MikaelMayer:
Fixes #710

## Problem

`pyAnalyzeLaurel` printed verbose per-VC verification results and the full `==== Verification Results ====` block to stdout even without `--verbose`. Additionally, untranslatable precondition error messages showed the literal string `<expr>` instead of the actual expression, and the SARIF "written to" message went to stdout.

## Solution

- Gate per-VC verification results behind `--verbose`; only the machine-readable `DETAIL:` and `RESULT:` lines appear on stdout by default
- Move Laurel translation errors and SARIF informational messages to stderr
- Add a `ToString` instance for `SpecExpr` using the DDM pretty-printer
- Show per-VC results by default (suppressed only in SARIF mode) with `fileName(line, col): outcome - message` format
- Use assertion name as fallback when no property summary message exists
- Propagate source location metadata to constructor calls (`__init__`), with-statement calls (`__enter__`/`__exit__`), and regular function calls so `callElimAssert` entries display file locations instead of "unknown location"
- Standardize `FileRange.format` to use consistent `(line, col)` spacing with space after comma
- Use `unknown location:` prefix for assertions that still lack source location

## Testing

All existing Lean and Python tests pass. Expected output files updated to match the new output format and resolved source locations.

### DIFF
--- a/Strata/DL/Lambda/LExprEval.lean
+++ b/Strata/DL/Lambda/LExprEval.lean
@@ -178,20 +178,6 @@ instance [ToFormat T.TypeType]: ToFormat (Except Strata.DiagnosticModel (LExpr T
               | .error err => f!"{err.message}"
 
 /--
-Embed `core` in an abstraction whose depth is `arity`. Used to implement
-eta-expansion.
-
-E.g., `mkAbsOfArity 2 core` will give `λxλy ((core y) x)`.
--/
-def mkAbsOfArity (arity : Nat) (core : LExpr T) : (LExpr T) :=
-  go 0 arity core
-  where go (bvarcount arity : Nat) (core : LExpr T) : (LExpr T) :=
-  match arity with
-  | 0 => core
-  | n + 1 =>
-    go (bvarcount + 1) n (.abs core.metadata "" .none (.app core.metadata core (.bvar core.metadata bvarcount)))
-
-/--
 A metadata merger. It will be invoked 'subst s e' is invoked, to create a new
 metadata.
 -/
@@ -309,16 +295,13 @@ def evalApp (n' : Nat) (σ : LState TBase) (e e1 e2 : LExpr TBase.mono) : LExpr 
       let newMeta := mergeMetadataForSubst mAbs e2'.metadata metaReplacementVar
       replaceMetadata1 newMeta e2') e1'
     if eqModuloMeta e e' then e else eval n' σ e'
-  | .op m fn _ =>
-    match σ.config.factory.getFactoryLFunc fn.name with
-    | none => LExpr.app m e1' e2'
-    | some lfunc =>
-      let e' := LExpr.app m e1' e2'
-      -- In `e'`, we have already supplied one input to `fn`.
-      -- Note that we can't have 0-arity Factory functions at this point.
-      let e'' := @mkAbsOfArity TBase.mono (lfunc.inputs.length - 1) (e' : LExpr TBase.mono)
-      eval n' σ e''
-  | _ => .app e.metadata e1' e2'
+  | _e =>
+    -- Re-evaluate when subexpressions changed (e.g. fvar resolved to .op),
+    -- so that `callOfLFunc` in `eval` can recognise the rebuilt expression
+    -- as a factory function call.  When nothing changed, `eqModuloMeta`
+    -- short-circuits and we return immediately.
+    let e' := .app e.metadata e1' e2'
+    if eqModuloMeta e e' then e else eval n' σ e'
 end
 
 instance : Traceable EvalProvenance Unit where
--- a/StrataTest/DL/Lambda/LExprEvalTests.lean
+++ b/StrataTest/DL/Lambda/LExprEvalTests.lean
@@ -254,7 +254,21 @@ private def testBuiltIn : @Factory TestParams :=
       inputs := [("x", mty[int]), ("y", mty[int])],
       output := mty[int],
       body := some esM[((~Int.Add x) y)]
-    }]
+    },
+
+    { name := "Int.Add3",
+      inputs := [("x", mty[int]), ("y", mty[int]), ("z", mty[int])],
+      output := mty[int],
+      concreteEval := some (fun _e args => match args with
+                        | [e1, e2, e3] =>
+                          let e1i := LExpr.denoteInt e1
+                          let e2i := LExpr.denoteInt e2
+                          let e3i := LExpr.denoteInt e3
+                          match e1i, e2i, e3i with
+                          | some x, some y, some z =>
+                            .some (.intConst e1.metadata (x + y + z))
+                          | _, _, _ => .none
+                        | _ => .none) }]
 
 private def testState : LState TestParams :=
   let ans := LState.addFactory LState.init testBuiltIn
@@ -656,6 +670,34 @@ example: stuck test25 := by
     rename_i a a2 a3 he
     cases a3
     cases a2; unfold denoteInt at he; contradiction
+
+
+-- Ternary function applied through a state variable.
+
+private def testStateFV : LState TestParams :=
+  { testState with state := [[("f", (none, esM[~Int.Add3]))]] }
+
+def test_ternary_fv := TestCase.mk
+  testStateFV
+  esM[((((f : int → int → int → int) #10) #20) #30)]
+  esM[#60]
+
+/-- info: true -/
+#guard_msgs in
+#eval check test_ternary_fv
+
+example: steps_well test_ternary_fv := by
+  unfold steps_well Scopes.toEnv test_ternary_fv testStateFV
+  take_step; apply Step.reduce_1
+  · inhabited_metadata
+  · apply Step.reduce_1
+    · inhabited_metadata
+    · apply Step.reduce_1
+      · inhabited_metadata
+      · apply Step.expand_fvar; rfl
+  take_step; apply Step.eval_fn <;> try rfl
+  · inhabited_metadata
+  take_refl
 
 
 end EvalTest
--- a/StrataTest/DL/Lambda/Lambda.lean
+++ b/StrataTest/DL/Lambda/Lambda.lean
@@ -53,7 +53,7 @@ info: Annotated expression:
 ((~Int.Div : (arrow int (arrow int int))) ((~Int.Add : (arrow int (arrow int int))) #2 #1))
 
 ---
-info: (λ ((~Int.Div : (arrow int (arrow int int))) #3 %0))
+info: ((~Int.Div : (arrow int (arrow int int))) #3)
 -/
 #guard_msgs in
 #eval format $ typeCheckAndPartialEval TypeFactory.default  (IntBoolFactory : @Factory TestParams)


StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
```diff
@@ -1,17 +1,2 @@
-
-==== Verification Results ====
-assert_assert(102)_calls_Any_to_bool_0: ✅ pass (at line 7, col 4)
```
MikaelMayer, line 3: Actually, since you are changing that, I would like to still see these detailed verification results by default, unless the SARIF mode is used. And while we are at it, I would like to hide the assertion name (which is a machine generated name that will be used mostly for denotational semantics) and instead display fileName(7,4): ✅ pass
and add ` - <message>` if a message was attached in the metadata of the assertion.

MikaelMayer: Done. Verification results are now shown by default (suppressed only in SARIF mode). The format uses `fileName(line, col): outcome` instead of machine-generated assertion names. Property summary messages are appended as ` - <message>` when available in the metadata. Updated all 43 expected test files and the README documentation.

StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
```diff
@@ -1,17 +1,14 @@
...
-assert_assert(345)_calls_Any_to_bool_0: ✅ pass (at line 16, col 4)
-assert(345): ✅ pass (at line 16, col 4)
-assert_assert(458)_calls_Any_to_bool_0: ✅ pass (at line 20, col 4)
-assert(458): ✅ pass (at line 20, col 4)
-set_quot_calls_PFloorDiv_0: ✅ pass
-assert_assert(567)_calls_Any_to_bool_0: ✅ pass (at line 24, col 4)
-assert(567): ✅ pass (at line 24, col 4)
-ite_cond_calls_Any_to_bool_0: ✅ pass (at line 26, col 0)
-
+test_arithmetic.py(7, 4): ✅ pass
```
MikaelMayer: Ok let's do the following. If the assertion has no message associated to it, we should reuse the assertion name instead of the message.

StrataTest/Languages/Python/expected_laurel/test_datetime.expected
```diff
@@ -1,18 +1,15 @@
...
-assert_assert(673)_calls_Any_to_bool_0: ✅ pass (at line 25, col 0)
-assert(673): ✅ pass (at line 25, col 0)
-assert_assert(758)_calls_Any_to_bool_0: ✅ pass (at line 27, col 0)
-assert(758): ✅ pass (at line 27, col 0)
-assert_assert(859)_calls_Any_to_bool_0: ✅ pass (at line 30, col 0)
-assert(859): ✅ pass (at line 30, col 0)
-
+✅ pass
+✅ pass
+✅ pass
```
MikaelMayer: A pass without any feedback indicates we are missing information. Investigate why we don't have line and column and file numbers. Since the original error indicated call elim assert, I suspect that they are preconditions, check why metadata is dropped.

StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
MikaelMayer: Done. When `getPropertySummary` returns no message, the assertion name (`obligation.label`) is now used as the fallback suffix (e.g. `test_arithmetic.py(7, 4): ✅ pass - assert(102)`). Updated all 43 expected test files.

StrataTest/Languages/Python/expected_laurel/test_datetime.expected
MikaelMayer: Investigated and fixed. The root cause was in `translateAssign` in `PythonToLaurel.lean`: the `Assign` and `LocalVariable` statements were created using `mkStmtExprMd` (which uses `defaultMetadata` with `FileRange.unknown`) instead of `mkStmtExprMdWithLoc ... md`. This meant the source location from the Python AST was dropped before reaching the Laurel-to-Core translator, so the `call` statements (and their resulting `callElimAssert` precondition checks) had no file range metadata.

Fixed by replacing `mkStmtExprMd` with `mkStmtExprMdWithLoc ... md` for all `Assign`/`LocalVariable` statements in the `Name` and `Subscript` cases of `translateAssign`. The `Attribute` case was already correct.

The fix also surfaced source locations for other previously-missing entries (e.g. `set_*_calls_*` in regex, arithmetic, loops tests). All 43 Python laurel tests pass with updated expected files.

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -351,6 +351,29 @@ inductive SpecExpr where
 | forallDict (dict : SpecExpr) (keyVar : String) (valVar : String) (body : SpecExpr)
 deriving Inhabited
 
+partial def SpecExpr.toString : SpecExpr → String
```
MikaelMayer: The DDM for PySpec already has pretty-printing capabilities. Cant' we use it instead of defining our own and be out of sync one day?

StrataTest/Languages/Python/README.md
```diff
@@ -150,14 +150,12 @@ lake exe strata pyAnalyzeLaurel --verbose \
...
-corresponding `.py` source file is adjacent to the `.python.st.ion` file,
-results include line and column numbers:
+Verification results are printed by default (suppressed in SARIF mode).
+Each line shows the source location and outcome:
 
 ```
-==== Verification Results ====
-Assertion failed at line 12, col 4: assert_result_positive: fail
-check_return_type: pass (at line 15, col 8)
+test_arithmetic.py(7, 4): ✅ pass
```
MikaelMayer: This documentation should also display assertion names

Strata/Languages/Python/Specs/Decls.lean
MikaelMayer: Replaced the custom `SpecExpr.toString` with a `ToString SpecExpr` instance in `DDM.lean` that uses the DDM pretty-printer via `SpecExpr.toDDM` + `mformat`. The old hand-written `toString` and its `ToString` instance have been removed from `Decls.lean`. `ToLaurel.lean` now imports `DDM.lean` to pick up the instance.

StrataTest/Languages/Python/README.md
MikaelMayer: Updated the README output example to include assertion names, matching the actual output format (e.g. `test_arithmetic.py(7, 4): ✅ pass - assert(102)`).

StrataTest/Languages/Python/expected_laurel/test_class_decl.expected
```diff
@@ -1,7 +1,4 @@
-
-==== Verification Results ====
-callElimAssert_requires_4: ✅ pass
-ite_cond_calls_Any_to_bool_0: ✅ pass (at line 13, col 0)
-
+✅ pass - callElimAssert_requires_4
+test_class_decl.py(13, 0): ✅ pass - ite_cond_calls_Any_to_bool_0
```
joscoh, line 2: Just wondering, why do some have file info and others do not? (I know that was also the case before this PR)

MikaelMayer: It's likely because we also need to merge this PR:
https://github.com/strata-org/Strata/pull/707

StrataMain.lean
```diff
@@ -648,46 +648,31 @@ def pyAnalyzeLaurelCommand : Command where
             | _     => false }
       | _ => {}
```
aqjune-aws: It seems this classifier variable isn't used anymore other than as a parameter to `printPyAnalyzeSummary` - maybe inline the classifier to `printPyAnalyzeSummary`, or even inline `printPyAnalyzeSummary` to this place because it is only used in this place?

StrataMain.lean
```diff
@@ -648,46 +648,31 @@ def pyAnalyzeLaurelCommand : Command where
...
 
-    -- Print results
+    -- Print translation errors (always on stderr)
     if !laurelTranslateErrors.isEmpty then
-      IO.println "\n==== Errors ===="
+      IO.eprintln "\n==== Errors ===="
```
aqjune-aws, line 683: +1

StrataTest/Languages/Python/expected_laurel/test_multi_function.expected
```diff
@@ -1,17 +1,14 @@
...
-callElimAssert_requires_1_10: ✅ pass (at line 26, col 4)
-callElimAssert_requires_2_11: ✅ pass (at line 26, col 4)
-ite_cond_calls_Any_to_bool_0: ✅ pass (at line 28, col 0)
-
+test_multi_function.py(9, 4): ❌ fail - ite_cond_calls_PNotIn_0
+test_multi_function.py(9, 4): ✅ pass - ite_cond_calls_Any_to_bool_1
+test_multi_function.py(11, 4): ❓ unknown - ite_cond_calls_PNotIn_0
+test_multi_function.py(11, 4): ✅ pass - ite_cond_calls_Any_to_bool_1
+test_multi_function.py(18, 4): ❓ unknown - ite_cond_calls_Any_to_bool_0
+❓ unknown - set_LaurelResult_calls_Any_get_0
```
aqjune-aws: What about having `unknown location: ` as prefix so that this line becomes:
```
unknown location: ❓ unknown - set_LaurelResult_calls_Any_get_0
```
It is because I think it will be pretty useful to always have `:` as a separator. It will be easier to e.g., write a bash script that collects assert result locations.

StrataMain.lean
MikaelMayer: Inlined the classifier into `printPyAnalyzeSummary` — it now takes `checkMode` directly and builds the classifier internally. The `classifier` variable at the call site is removed.

StrataTest/Languages/Python/expected_laurel/test_multi_function.expected
MikaelMayer: Added `unknown location: ` prefix when no source location is available. Updated all affected expected files.

StrataTest/Languages/Python/expected_laurel/test_class_field_any.expected
```diff
@@ -1,8 +1,5 @@
-
-==== Verification Results ====
-callElimAssert_requires_3: ✅ pass
-assert_assert(113)_calls_Any_to_bool_0: ✅ pass (at line 6, col 0)
-assert(113): ❓ unknown (at line 6, col 0)
-
+unknown location: ✅ pass - callElimAssert_requires_3
+test_class_field_any.py(6, 0): ✅ pass - assert_assert(113)_calls_Any_to_bool_0
```
MikaelMayer, line 2: What makes the most sense, (6,0) or (6, 0) ? asking to make sure we are standard. Be honest.

StrataTest/Languages/Python/expected_laurel/test_with_statement.expected
```diff
@@ -1,23 +1,20 @@
...
+unknown location: ✅ pass - callElimAssert_requires_9
+unknown location: ✅ pass - callElimAssert_requires_5
+unknown location: ✅ pass - callElimAssert_requires_2
+test_with_statement.py(19, 4): ✅ pass - assert_assert(368)_calls_Any_to_bool_0
+test_with_statement.py(19, 4): ❌ fail - assert(368)
+unknown location: ✅ pass - callElimAssert_requires_19
+unknown location: ✅ pass - callElimAssert_requires_15
+test_with_statement.py(25, 8): ✅ pass - assert_assert(500)_calls_Any_to_bool_0
+test_with_statement.py(25, 8): ❌ fail - assert(500)
+unknown location: ✅ pass - callElimAssert_requires_12
```
MikaelMayer: These unknown locations should now have been resolved since the PR that preserves metadata for assertions just got merged. Please investigate why we don't have a location here. We should have a location.

StrataTest/Languages/Python/expected_laurel/test_class_field_use.expected
```diff
@@ -1,8 +1,5 @@
-
-==== Verification Results ====
-callElimAssert_requires_7: ✅ pass
-assert_assert(285)_calls_Any_to_bool_0: ✅ pass (at line 14, col 4)
-Assertion failed at line 14, col 4: assert(285): ❌ fail
-
+unknown location: ✅ pass - callElimAssert_requires_7
```
MikaelMayer: This unknown location might be the easiest to fix as per my previous comment.

StrataTest/Languages/Python/expected_laurel/test_class_field_any.expected
MikaelMayer: `(6, 0)` with a space is the right choice — it matches Python tuple formatting and is more readable. The `includeEnd?=false` case in `FileRange.format` already used `(line, column)` with a space, but the `includeEnd?=true` cases were inconsistent (no space). Standardized all cases in `FileRange.format` to use a space after the comma.

StrataTest/Languages/Python/expected_laurel/test_with_statement.expected
MikaelMayer: Investigated and fixed. The root cause was that `mkInstanceMethodCall` (used for `__enter__` and `__exit__` calls in the `With` statement translation) created nodes with `defaultMetadata` (which has `FileRange.unknown`). Added an optional `md` parameter to `mkInstanceMethodCall` and now pass the statement's source location metadata to all `__enter__`/`__exit__` calls. Also fixed `emitCall` in `translateCall` to propagate the call-site source location to all emitted `StaticCall` nodes. Expected test files will be updated once CI shows the actual output (cannot build the binary locally due to GLIBC version mismatch).

StrataTest/Languages/Python/expected_laurel/test_class_field_use.expected
MikaelMayer: Fixed in the same commit. The `callElimAssert_requires_7` here comes from the `CircularBuffer(5)` constructor call, which triggers `mkInstanceMethodCall` for `__init__`. Now passes the statement's source location metadata so the assertion gets the correct file location.

aqjune-aws (approved):

PR #714, opened 2026-03-30
Improve exception logic for missing Python models

keyboardDrummer:
### Changes
- Improve handling of missing models by also assigning a hole to `maybe_except`
- Improve Laurel's type inference for holes

### Testing
- Update expect file for `test_missing_models`


MikaelMayer (dismissed):
Yes this is a good soundness fix.

Strata/Languages/Laurel/InferHoleTypes.lean
```diff
@@ -120,7 +120,15 @@ private def inferStmt (stmt : StmtExprMd) : InferHoleM StmtExprMd := do
   match val with
   | .LocalVariable name ty (some initExpr) =>
       return ⟨.LocalVariable name ty (some (← inferExpr initExpr ty)), md⟩
-  | .Assign targets value => return ⟨.Assign targets (← inferExpr value defaultHoleType), md⟩
+  | .Assign targets value =>
+      -- Infer the value type from the first target's type when possible
+      let targetType : HighTypeMd := match targets with
+        | [tgt] => computeExprType model tgt
```
atomb: The comment suggests that this pattern should be `tgt :: _` instead of `[tgt]`. Is that right?

atomb (commented):
I'm not sure I understand the translation of exceptions from Python well enough to be sure whether this is correct, but I noticed one place where the code and comment seem inconsistent.

Strata/Languages/Laurel/InferHoleTypes.lean
keyboardDrummer: The comment is incorrect. @keyboardDrummer-bot could you update the comment to match the implementation?

Also, given that the fallback type for a hole is .Unknown, 
```
      let inferredType := match targetType.val with
        | .Unknown => defaultHoleType
        | _ => targetType
```

can be replaced with just `let inferredType := targetType` and then `inferredType` can be inlined.

atomb:
@keyboardDrummer-bot, can you fix the merge conflicts?

atomb (dismissed):

keyboardDrummer:
@keyboardDrummer-bot please resolve the merge conflict

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1019,13 +1019,20 @@ partial def translateAssign  (ctx : TranslationContext)
       stmts := stmts ++ eltStmts
     return (curCtx, stmts)
   let rhs_trans ←  translateExpr ctx rhs
+  -- When an unmodeled call produces a Hole, also havoc maybe_except since
+  -- the call is a black box that could throw any exception.
+  let rhsIsCall := match rhs with | .Call _ _ _ _ => true | _ => false
```
tautschnig, line 1060: The `exceptHavoc` pattern (check if RHS is a Call, emit `Assign [maybeExceptVar] Hole`) is duplicated in `translateStmt` (line 1339 as `holeExceptHavoc`). Consider extracting a helper:

```lean
private def mkExceptHavocIfCall (rhs : Python.expr SourceRange) (md : MetaData) : List StmtExprMd :=
  match rhs with
  | .Call _ _ _ _ => [mkStmtExprMdWithLoc (StmtExpr.Assign [maybeExceptVar] (mkStmtExprMd (.Hole false none))) md]
  | _ => []
```

StrataTest/Languages/Python/expected_laurel/test_missing_models.expected
```diff
@@ -1,7 +1,31 @@
...
 test_missing_models.py(9, 4): ✅ pass - (Origin_test_helper_procedure_Requires)req_opt_name_none_or_bar
-DETAIL: 2 passed, 0 failed, 3 inconclusive
+test_missing_models.py(12, 4): ✅ pass - (Origin_test_helper_procedure_Requires)req_name_is_foo
+test_missing_models.py(12, 4): ✅ pass - (Origin_test_helper_procedure_Requires)req_opt_name_none_or_str
+test_missing_models.py(12, 4): ✅ pass - (Origin_test_helper_procedure_Requires)req_opt_name_none_or_bar
+test_missing_models.py(15, 4): ✅ pass - (Origin_test_helper_procedure_Requires)req_name_is_foo
+test_missing_models.py(15, 4): ✅ pass - (Origin_test_helper_procedure_Requires)req_opt_name_none_or_str
+test_missing_models.py(15, 4): ✅ pass - (Origin_test_helper_procedure_Requires)req_opt_name_none_or_bar
+DETAIL: 22 passed, 0 failed, 7 inconclusive
 RESULT: Inconclusive
```
tautschnig, line 31: The output grew from 5 to 29 lines, with `req_name_is_foo` appearing 9 times (was 1). This suggests the `maybe_except` havoc is creating multiple execution paths that each independently check the same preconditions. Is this expected, or is the havoc being inserted at a point that causes path explosion?

keyboardDrummer: >  Is this expected, or is the havoc being inserted at a point that causes path explosion?

It's a known bug in Core. The havoc causes more existing paths to become alive instead of dead, and those paths trigger the duplication. What Core should do is group all VC results that originate from the same assertion, so that we can produce a single diagnostic per assertion.

Strata/Languages/Python/PythonToLaurel.lean
keyboardDrummer: I'm not inclined to have a helper for such a small amount of code, but if you prefer it I'm happy to add it. What would you like?

tautschnig: I think the main question is how much this code will evolve over time and whether there will be more places where the same pattern will appear. Evolution is the biggest risk: if someone only changes one of the uses and forgets about the other, we get into trouble.

StrataTest/Languages/Python/expected_laurel/test_missing_models.expected
tautschnig: I know there was #419, but that's been fixed - is there another issue where this is documented?

keyboardDrummer: It's a separate but related problem. I don't think there was an issue for it but I've created one: https://github.com/strata-org/Strata/issues/832

keyboardDrummer:

tautschnig:
@keyboardDrummer-bot please resolve merge conflicts.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1385,6 +1398,7 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
             else
               return (ctx, exceptionCheck ++ [expr])
         | _ => return (ctx, exceptionCheck ++ [expr])
```
tautschnig, line 1400: When `expr.val` is `.Hole`, `exceptionCheck` (computed on line 1382) is silently dropped. This is intentionally correct (an unmodeled call can't produce meaningful exception checks), but it's non-obvious to a reader scanning the match arms — every other arm includes `exceptionCheck` in its return.

Consider adding a brief comment:
```lean
    -- Unmodeled call: skip exception checks (no model to check against),
    -- but havoc maybe_except since the call could throw.
    | .Hole => return (ctx, [expr] ++ holeExceptHavoc)
```

StrataTest/Languages/Python/PreludeVerifyTest.lean
```diff
@@ -194,15 +194,15 @@ Obligation: postcondition
 Property: assert
 Result: ✅ pass
 
-Obligation: assert(43079)
+Obligation: assert(43081)
```
tautschnig: Nit: These byte-offset-based obligation names (`assert(43081)`) are fragile — any upstream change to the prelude or translation logic shifts them. This isn't introduced by this PR, but it's worth noting that these tests will break on unrelated refactors. Is there a way to use stable names (e.g., based on the assertion label text) instead of byte offsets?

keyboardDrummer: > Is there a way to use stable names (e.g., based on the assertion label text) instead of byte offsets?

Updated the test so it shows results more like what the user sees, which I think is what our tests generally should do. The user will by default only see failing results, and this test should have none, so it removes the entire expected output.

tautschnig (approved):

MikaelMayer (approved):
Havoc-ing `maybeExcept` is a rightful soundness fix.

PR #715, opened 2026-03-30
Add checking for functions in Prelude that may return Any of exception type

thanhnguyen-aws:
*Issue #, if available:*  Some functions in Python prelude that returns an exception as a value of Any type, such as PAdd, PMul,... Those exceptions are not currently caught and checked in Python to Laurel translation.

*Description of changes:*
- Add a suffix "!AnyMaybeException" to the names of those functions to recognize them.
- Make sure that those functions propagate the exceptions of their arguments (check and return Exceptions if any of their inputs are exception).
- Add an assertion : `!Any..isexception(expr)` for each outermost `expr` that is a StaticCall of a function whose name ends with "!AnyMaybeException".

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -526,7 +526,7 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
         | .IsNot _ => match comparators.val[0]'hComp with
             | .Constant _ (.ConNone _) _ => .ok "PNEq"
             | _ => throw (.unsupportedConstruct "`is not` is only supported with None" (toString (repr e)))
-      return mkStmtExprMd (StmtExpr.StaticCall preludeOpnames [leftExpr, rightExpr])
+      return mkStmtExprMdWithLoc (StmtExpr.StaticCall preludeOpnames [leftExpr, rightExpr]) md
 
   -- Boolean operations
   | .BoolOp _ op values => do
--- a/StrataTest/Languages/Python/expected_laurel/test_is_none.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_is_none.expected
@@ -8,9 +8,11 @@ test_is_none.py(19, 4): ✅ pass - z should be None
 test_is_none.py(19, 4): ✅ pass - z should be None
 test_is_none.py(23, 4): ✅ pass - w should not be None
 test_is_none.py(23, 4): ✅ pass - w should not be None
+test_is_none.py(27, 11): ✅ pass - Check PNot exception
 test_is_none.py(27, 4): ✅ pass - int is not None
 test_is_none.py(27, 4): ✅ pass - int is not None
+test_is_none.py(31, 11): ✅ pass - Check PNot exception
 test_is_none.py(31, 4): ✅ pass - None is not 'not None'
 test_is_none.py(31, 4): ✅ pass - None is not 'not None'
-DETAIL: 14 passed, 0 failed, 0 inconclusive
+DETAIL: 16 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1146,6 +1150,23 @@ def createVarDeclStmtsAndCtx (ctx : TranslationContext) (newDecls : List (String
         if isCompositeType ctx ty then (n, ty) else (n, PyLauType.Any)) }
   return (hoistedDecls, hoistedCtx)
 
+def isMaybeExceptAnyFunc (funcName: String) : Bool := funcName.endsWith "!AnyMaybeExcept"
```
tautschnig: The `!AnyMaybeExcept` suffix convention works but is fragile. Consider:
- A `@[maybeException]` metadata annotation on the Laurel function declaration, or
- A `HashSet` of known exception-propagating function names (similar to `knownIdentTypes` in PR #736)

Either would be more robust than string suffix matching and wouldn't require encoding semantic information in function names.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1146,6 +1150,23 @@ def createVarDeclStmtsAndCtx (ctx : TranslationContext) (newDecls : List (String
...
+
+partial def getMaybeExceptionExprs (e : StmtExprMd) : List StmtExprMd :=
+  match e.val with
+  | .StaticCall funcname args =>
+    if isMaybeExceptAnyFunc funcname.text then
+      [{e with md:= e.md.withPropertySummary $ "Check " ++ funcname.text ++ " exception"}]
```
tautschnig, line 1203: The recursive descent into `StaticCall` args means nested `!AnyMaybeExcept` calls each generate their own assertion. For `PAdd!AnyMaybeExcept(PAdd!AnyMaybeExcept(a, b), c)`, this produces two assertions:
1. `!isexception(PAdd!AnyMaybeExcept(a, b))`
2. `!isexception(PAdd!AnyMaybeExcept(PAdd!AnyMaybeExcept(a, b), c))`

Assertion 2 subsumes assertion 1 because `PAdd!AnyMaybeExcept` propagates exceptions from its inputs. The inner assertion is redundant. Consider only asserting on the outermost `!AnyMaybeExcept` call to reduce VC count.

thanhnguyen-aws: This function does not descent into args when the outer function is of "AnyMaybeExcept", so in case of `PAdd`, it only produces one assertion for the outermost one.

tautschnig (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1157,6 +1163,23 @@ def createVarDeclStmtsAndCtx (ctx : TranslationContext) (newDecls : List (String
...
+
+partial def getMaybeExceptionExprs (ctx : TranslationContext) (e : StmtExprMd) : List StmtExprMd :=
+  match e.val with
+  | .StaticCall funcname args =>
+    if isMaybeExceptAnyFunc ctx funcname.text then
+      [{e with md:= e.md.withPropertySummary $ "Check " ++ funcname.text ++ " exception"}]
+    else args.flatMap $ getMaybeExceptionExprs ctx
+  | .IfThenElse cond thenBranch elseBranch =>
+      ([cond, thenBranch] ++ elseBranch.toList).flatMap $ getMaybeExceptionExprs ctx
+  | _ => []
```
ssomayyajula, line 1208: Agreed with Michael on this: can you verify that it's OK not to recurse through any other expression form?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -587,7 +592,7 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
     let index ←  match slice with
       | .Slice _ start stop step => translateSlice ctx start.val stop.val step.val
       | _ => translateExpr ctx slice
-    return mkStmtExprMd (.StaticCall "Any_get" [dictOrList, index])
+    return mkStmtExprMdWithLoc (.StaticCall "Any_get" [dictOrList, index]) md
```
ssomayyajula, line 609: Is it an issue that `Any_get!` is never emitted in the translation?

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -985,9 +985,18 @@ procedure print(msg : Any) returns ();
 /--
 Parse the Laurel DDM prelude into a Laurel Program.
 -/
+
+def AnyMaybeExceptionList := ["Any_get!", "Any_set!", "Any_sets!", "PNeg", "PNot", "PAdd", "PSub", "PMul",
```
ssomayyajula, line 993: Is it intentional that modeled operators like `PEq`, `PNEq`, `PAnd`, `POr`, `PIn`, and `PNotIn` don't consume (by indicating in their preconditions) or propagate exceptional values?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -87,6 +87,8 @@ structure TranslationContext where
   functionSignatures : List PythonFunctionDecl := []
   /-- Names of user-defined functions -/
   userFunctions : List String := []
+  /-- Function that may returns Any of exception -/
```
MikaelMayer: Nit: grammar — "Function that may returns" → "Functions that may return". Same docstring appears on `PreludeInfo.maybeExceptionfunctions` further down.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -87,6 +87,8 @@ structure TranslationContext where
...
+  maybeExceptionfunctions : List String := []
```
MikaelMayer: Nit: the naming convention in the codebase uses camelCase (`userFunctions`, `userClasses`). This field should be `maybeExceptionFunctions` (capital F) for consistency.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1383,7 +1415,7 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
       | .Mod _      => mkStmtExprMd (StmtExpr.StaticCall "PMod"      [targetExpr, valueExpr])
       | _           => mkStmtExprMd .Hole
     let assignStmt := mkStmtExprMdWithLoc (StmtExpr.Assign [targetExpr] rhs) md
-    return (ctx, [assignStmt])
+    return (ctx, (getExceptionAssertions ctx rhs) ++ [assignStmt])
```
MikaelMayer, line 1478: The `rhs` arms above all use `mkStmtExprMd` (no location), so the generated exception-check assertions end up with `defaultMetadata` — which is why the test expected outputs show `unknown location: ✅ pass - Check PAdd exception` for augmented assignments. Using `mkStmtExprMdWithLoc ... md` for each arm would give proper source locations, consistent with how `BinOp`, `UnaryOp`, and `Compare` now propagate `md`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1168,12 +1191,19 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     -- For now, only support single target
     if targets.val.size != 1 then
       throw (.unsupportedConstruct "Multiple assignment targets not yet supported" (toString (repr s)))
-    translateAssign ctx targets.val[0]! none value md
+    let (newctx, stmts) ← translateAssign ctx targets.val[0]! none value md
+    let rhs_exprs:= stmts.flatMap (λ s => match s.val with |.Assign _ value => [value] | _=> [])
+    let exceptionCheck := rhs_exprs.flatMap $ getExceptionAssertions ctx
```
MikaelMayer: This 3-line pattern (extract rhs from Assign stmts, flatMap getExceptionAssertions, prepend) is duplicated verbatim in the `AnnAssign` case below. Consider extracting a small helper, e.g.:
```lean
def withExceptionChecks (ctx : TranslationContext)
    (result : TranslationContext × List StmtExprMd)
    : TranslationContext × List StmtExprMd :=
  let (newctx, stmts) := result
  let rhs_exprs := stmts.flatMap fun s =>
    match s.val with | .Assign _ value => [value] | _ => []
  let exceptionCheck := rhs_exprs.flatMap $ getExceptionAssertions ctx
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1157,6 +1163,23 @@ def createVarDeclStmtsAndCtx (ctx : TranslationContext) (newDecls : List (String
         if isCompositeType ctx ty then (n, ty) else (n, PyLauType.Any)) }
   return (hoistedDecls, hoistedCtx)
 
+def isMaybeExceptAnyFunc (ctx : TranslationContext) (funcName: String) : Bool := funcName ∈ ctx.maybeExceptionfunctions
+
+partial def getMaybeExceptionExprs (ctx : TranslationContext) (e : StmtExprMd) : List StmtExprMd :=
+  match e.val with
+  | .StaticCall funcname args =>
+    if isMaybeExceptAnyFunc ctx funcname.text then
```
MikaelMayer, line 1202: When the `StaticCall` is a maybe-exception function, the recursion stops — arguments are not traversed. This means for `PAdd(PMul(x, y), z)`, only `PAdd` gets an exception assertion, not `PMul`. This is sound because these functions propagate exceptions from their arguments (e.g. `PAdd` checks `isexception` on its inputs), so if `PMul` returned an exception, `PAdd` would too and the outer assertion catches it. A brief comment here explaining this design choice would help future readers.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
MikaelMayer, line 993: A docstring on `AnyMaybeExceptionList` explaining the criteria for inclusion would help future maintainers know when to add new entries. Something like: "Prelude functions that may return an exception value as `Any` (no `requires` clause guarding against invalid inputs)."

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I added one more case: PrimitiveOp. Besides the 3 forms that are handled by this function, there are 2 forms that can be returned by `translateExpr`: `.Identifier` and `.Hole` and they cannot contain this kind of exception.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: `Any_get!` is just a support function to define `Any_sets!`. Up to this PR, we avoid defining and using functions may return an exception value as Any whenever it is possible and convenient to use function precondition. This may change when we need to know the type of exceptions (for Try-Catch statement).

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
thanhnguyen-aws: I added `PAnd` and `POr` to the list because they can avoid some redundant checking. For other functions, I don't think we should propagate exceptional values at Core level at this point as the performance impact should be higher than checking exceptions for each input argument individually (just my hunt, please correct me if you think it is wrong).

ssomayyajula: Fine with me!

PR #716, opened 2026-03-31
Fix #653 and other bugs with iterated substitution

### LINKED ISSUE #653: Soundness Bug in Function Substitution
This assertion succeeds:
```
def functionAssertPgm : Program :=
#strata
program Core;

inline function foo(x : int, $__y0 : int) : int { x + $__y0 }

procedure TestFoo() returns ()
{
  var y : int;
  assert [fooEq]: (foo(y, 2) == 4);
};

#end

#eval verify functionAssertPgm
```

The reason is that in the evaluation of Factory functions, arguments are substituted for the free variables sequentially, so one can be captured by a future substitution. Here we first substitute `x -> $__y0` and then substitute `$__y0 -> 2`.  This particular example relies on giving an "illegal" variable name, but since functions and expressions can be constructed via an API, name clashes are possible.

The most straightforward fix is to do simultaneous substitution rather than sequential substitution when evaluating Factory functions.

joscoh:
*Issue #, if available:* #653

*Description of changes:* The use of `substFvars` for repeated substitution is unsound whenever either (a) the substituted expressions are not guaranteed to use only `fvars` that do not appear in the substitution list or (b) the substituted expressions contain bound variables. This PR instead adds simulatenous substitution operators that safely substitute multiple free variables at once, with both `bvar`-safe and standard versions. It classifies each place where iterated substitution was used as safe or requiring one of the new functions, adding an appropriate comment for each to make the assumptions explicit (for example, for some cases, the substituted terms are closed, so it safe to use a non-lifting substitution function). 
It includes tests that formerly failed illustrating soundness issues with the prior approach: for the Lambda partial evaluator, for `substFvarsFromState`, for the precondition generation, closure capture, and procedure call capture.

### DIFF
--- a/Strata/DL/Lambda/LExprWF.lean
+++ b/Strata/DL/Lambda/LExprWF.lean
@@ -352,20 +352,20 @@ in a single pass, avoiding variable capture between substitutions.
 Does NOT lift de Bruijn indices when going under binders. Safe only when all
 replacement expressions contain no bvars.
 -/
-def substMultiFvars [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (sm : Map T.Identifier (LExpr ⟨T, GenericTy⟩))
+def substFvars [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (sm : Map T.Identifier (LExpr ⟨T, GenericTy⟩))
   : LExpr ⟨T, GenericTy⟩ :=
-  if sm.isEmpty then e else substMultiFvarsAux e sm
+  if sm.isEmpty then e else substFvarsAux e sm
 where
-  substMultiFvarsAux (e : LExpr ⟨T, GenericTy⟩) (sm : Map T.Identifier (LExpr ⟨T, GenericTy⟩))
+  substFvarsAux (e : LExpr ⟨T, GenericTy⟩) (sm : Map T.Identifier (LExpr ⟨T, GenericTy⟩))
     : LExpr ⟨T, GenericTy⟩ :=
     match e with
     | .const _ _ => e | .bvar _ _ => e | .op _ _ _ => e
     | .fvar _ name _ => match sm.find? name with | some to => to | none => e
-    | .abs m name ty e' => .abs m name ty (substMultiFvarsAux e' sm)
-    | .quant m qk name ty tr' e' => .quant m qk name ty (substMultiFvarsAux tr' sm) (substMultiFvarsAux e' sm)
-    | .app m fn e' => .app m (substMultiFvarsAux fn sm) (substMultiFvarsAux e' sm)
-    | .ite m c t e' => .ite m (substMultiFvarsAux c sm) (substMultiFvarsAux t sm) (substMultiFvarsAux e' sm)
-    | .eq m e1 e2 => .eq m (substMultiFvarsAux e1 sm) (substMultiFvarsAux e2 sm)
+    | .abs m name ty e' => .abs m name ty (substFvarsAux e' sm)
+    | .quant m qk name ty tr' e' => .quant m qk name ty (substFvarsAux tr' sm) (substFvarsAux e' sm)
+    | .app m fn e' => .app m (substFvarsAux fn sm) (substFvarsAux e' sm)
+    | .ite m c t e' => .ite m (substFvarsAux c sm) (substFvarsAux t sm) (substFvarsAux e' sm)
+    | .eq m e1 e2 => .eq m (substFvarsAux e1 sm) (substFvarsAux e2 sm)
 
 /--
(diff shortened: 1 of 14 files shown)


Strata/DL/Lambda/LExprWF.lean
```diff
@@ -345,14 +345,74 @@ where
...
     | .eq m e1 e2 => .eq m (go e1 depth) (go e2 depth)
 
+/--
+Iterated substitution of multiple free variables with lifting (sequential `foldl`).
+**Warning:** This is NOT simultaneous substitution — earlier substitutions can
+introduce variables that are then captured by later ones. Only use when you can
+guarantee that the replacement expressions do not contain fvars appearing later
+in the substitution list. Prefer `substMultiFvarsLifting` for safe simultaneous
+substitution.
+-/
```
andrewmwells-amazon: I think we should just remove this

Strata/DL/Lambda/LExprWF.lean
```diff
@@ -345,14 +345,74 @@ where
...
   List.foldl (fun e (var, s) => substFvarLifting e var s) e sm
 
+/--
+Iterated substitution of multiple free variables (sequential `foldl`).
+**Warning:** This is NOT simultaneous substitution — earlier substitutions can
+introduce variables that are then captured by later ones. Only use when you can
+guarantee that the replacement expressions do not contain fvars appearing later
+in the substitution list, and that the replacement expressions contain no bvars.
+Prefer `substMultiFvars` for safe simultaneous substitution.
+-/
```
andrewmwells-amazon: Same here

Strata/Transform/CallElim.lean
```diff
@@ -79,8 +79,10 @@ def callElimCmd (cmd: Command)
             | _ => none
         let oldSubst := createOldVarsSubst oldTrips ++ unmodifiedOldSubst
 
+        -- Iterated substitution is safe here: keys are "old g" identifiers (prefixed
```
MikaelMayer: The comment says "Iterated substitution is safe here" but the code now uses `substMultiFvars` (simultaneous). Since the whole point of this PR is to move away from iterated substitution, this comment is confusing — it reads as if the old approach is still in use. Consider rewording, e.g.: "Simultaneous substitution is used. (Note: even iterated substitution would be safe here because keys are 'old g' identifiers and values are fresh variables that cannot appear as keys.)"

StrataTest/Languages/Core/Examples/SubstFvarsCaptureTests.lean
```diff
@@ -0,0 +1,152 @@
...
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.Languages.Core.Verifier
+
+/-! # Simultaneous substitution tests (Issue 653)
+
+Tests verifying that simultaneous substitution (`substMultiFvars` /
+`substMultiFvarsLifting`) avoids variable capture that occurrs with the
```
MikaelMayer: Nit: typo — `occurrs` → `occurs`.

Strata/DL/Lambda/LExprWF.lean
```diff
@@ -345,13 +345,49 @@ where
     | .ite m c t f => .ite m (go c depth) (go t depth) (go f depth)
     | .eq m e1 e2 => .eq m (go e1 depth) (go e2 depth)
 
-def substFvarsLifting [BEq T.IDMeta] (e : LExpr ⟨T, GenericTy⟩) (sm : Map T.Identifier (LExpr ⟨T, GenericTy⟩))
+/--
```
MikaelMayer, line 348: No action needed: the docstrings on `substMultiFvars` and `substMultiFvarsLifting` clearly state the contract ("Does NOT lift de Bruijn indices" / "Properly lifts de Bruijn indices") and when to use each variant. This makes it easy for callers to pick the right one.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -49,12 +49,11 @@ private def callConditions (proc : Procedure)
   let names := List.map
                (fun k => s!"(Origin_{proc.header.name.name}_{condType}){k}")
                conditions.keys
+  -- The replacement expressions must be closed (no dangling bvars).
+  let sm := subst.map (fun (x, v) => (x.fst, v))
```
MikaelMayer, line 53: Nice refactor — building the substitution map first and then calling `substMultiFvars` once is cleaner than the old `foldl` that interleaved map construction with substitution.

Strata/Transform/CallElim.lean
joscoh: Thanks, this is an outdated comment. Will fix.

PR #719, opened 2026-03-31
Add --profile and --quiet flags to pyAnalyzeLaurel

joehendrix:
Add per-step timing to pyAnalyzeLaurel so we can identify performance
bottlenecks across the pipeline without modifying benchmark scripts.

## Command Line Flags Added

* `--profile` — Print elapsed time in milliseconds for each pipeline step
  to stdout. Covers Python-to-Laurel sub-steps, Laurel-to-Core, PySpec
  inlining, and SMT verification internals (program transformations,
  type check/PE, VC discharge with preprocess/encoding/solver breakdown).
* `--quiet` — Suppress stderr warnings from PySpec and dispatch module
  resolution, which are noisy during batch benchmarking.

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license and the MIT license.

### DIFF
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -670,9 +670,9 @@ def verifySingleEnv (pE : Program × Env) (options : VerifyOptions)
               [DEBUG] Evaluated program: {Core.formatProgram p}\n\n"
   | _ =>
     let mut results := (#[] : VCResults)
-    let mut preprocessMs : Nat := 0
-    let mut smtEncodeMs : Nat := 0
-    let mut solverMs : Nat := 0
+    let mut preprocessNs : Nat := 0
+    let mut smtEncodeNs : Nat := 0
+    let mut solverNs : Nat := 0
     let mut peResolvedCount : Nat := 0
     for obligation in E.deferred do
       -- Determine which checks to perform based on metadata or check mode/amount
@@ -692,10 +692,10 @@ def verifySingleEnv (pE : Program × Env) (options : VerifyOptions)
           | .bugFinding, .minimalVerbose, .assert | .bugFinding, .minimalVerbose, .divisionByZero => (true, false) -- Same checks as minimal
           | .bugFinding, .minimal, .cover => (true, false)  -- Cover uses satisfiability
           | .bugFinding, .minimalVerbose, .cover => (true, false)  -- Same checks as minimal
-      let t0 ← (IO.monoNanosNow : BaseIO Nat)
+      let t0 ← IO.monoNanosNow
       let (obligation, peSatResult?, peValResult?) ← preprocessObligation obligation p options satisfiabilityCheck validityCheck axiomCache
-      let t1 ← (IO.monoNanosNow : BaseIO Nat)
-      preprocessMs := preprocessMs + (t1 - t0) / 1000000
+      let t1 ← IO.monoNanosNow
+      preprocessNs := preprocessNs + (t1 - t0)
       -- If PE resolved both checks, we're done, unless we always want to generate SMT queries
       if not options.alwaysGenerateSMT then
         if let (some peSat, some peVal) := (peSatResult?, peValResult?) then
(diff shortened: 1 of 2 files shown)


Strata/Languages/Core/Verifier.lean
```diff
@@ -661,13 +662,18 @@ def verifySingleEnv (pE : Program × Env) (options : VerifyOptions)
...
   match E.error with
   | some err =>
     .error <| DiagnosticModel.fromFormat s!"🚨 Error during evaluation!\n\
               {format err}\n\n\
               [DEBUG] Evaluated program: {Core.formatProgram p}\n\n"
   | _ =>
     let mut results := (#[] : VCResults)
+    let mut preprocessMs : Nat := 0
+    let mut smtEncodeMs : Nat := 0
+    let mut solverMs : Nat := 0
```
MikaelMayer: The manual timing accumulation here (and the `smtEncodeMs`/`solverMs` blocks below) duplicates the pattern that `profileStep` already abstracts. Consider wrapping each timed section with `profileStep` instead — it would remove ~12 lines of boilerplate and guarantee the output format stays consistent with the rest of the pipeline.

If the concern is that `profileStep` prints per-invocation while you want an aggregate across the loop, a small `profileAccum` variant (returns elapsed alongside the result, caller sums) would keep the code DRY.

Strata/Languages/Core/Verifier.lean
```diff
@@ -734,6 +750,11 @@ def verifySingleEnv (pE : Program × Env) (options : VerifyOptions)
             let prog := f!"\n\n[DEBUG] Evaluated program:\n{Core.formatProgram p}"
             dbg_trace f!"\n\nResult: {result}\n{prog}"
           if options.stopOnFirstError then break
+    if profile then
```
MikaelMayer, line 753: The indentation here uses 5 spaces (`"[profile]     Preprocess..."`) while the `profileStep` calls in `verify` use 2-space-indented names (`"  Program transformations"`), producing output like `"[profile]   Program transformations: 42ms"`. The nesting intent is good, but the two schemes don't align visually. Consider using the same indentation depth (e.g., 4 spaces in the name) so the hierarchy reads clearly.

Strata/Util/Profile.lean
```diff
@@ -0,0 +1,16 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+/-- Run an action, printing its elapsed time in milliseconds to stdout when `profile` is true. -/
+public def profileStep {m α} [Monad m] [MonadLiftT BaseIO m]
```
MikaelMayer, line 11: 👍 No action needed — this is a well-factored utility. The `[MonadLiftT BaseIO m]` constraint keeps it usable across `IO`, `EIO`, etc., and the early-return when `!profile` avoids any overhead in the common case.

Strata/Languages/Core/Verifier.lean
joehendrix: Thanks for the suggestion. I considered a `profileAccum` variant but it only saves ~3 lines since we still need the mutable accumulators and the summary print block. The manual approach also handles the early-exit `continue` path (PE-resolved obligations) more naturally.

I did clean up the boilerplate: removed the unnecessary `(... : BaseIO Nat)` type annotations, introduced an `nsToMs` helper with rounding, and switched to accumulating in nanoseconds to avoid intermediate rounding errors. See 2270473.

Strata/Languages/Core/Verifier.lean
joehendrix: Good catch — the indentation levels are intentional to show hierarchy (top-level steps at 0 indent, verify sub-steps at 2 spaces, VC discharge sub-steps at 4 spaces) but I agree the spacing was inconsistent. The 5-space indent was a typo. Fixed in 2270473.

Strata/Util/Profile.lean
joehendrix: Thanks!

PR #720, opened 2026-03-31
feat: Emit set-info metadata and user_errors.txt on user error exit

olivier-aws:
## What

When pyAnalyzeLaurelCommand encounters a PipelineError.userCode error, we now emit structured SMT-LIB set-info metadata both to stdout and to a user_errors.txt file, before the existing DETAIL/RESULT lines.

## Why

Downstream consumers need machine-readable error information (source file, byte offsets, error message) rather than having to parse the human-readable DETAIL line. Writing to user_errors.txt additionally gives file
-based consumers a stable artifact to read.

## Changes

- **StrataMain.lean**: In the PipelineError.userCode handler of pyAnalyzeLaurelCommand:
  - Emit (set-info :file ...), (set-info :start ...), (set-info :stop ...), and (set-info :error-message ...) to stdout via IO.println
  - :start/:stop are omitted when range.isNone
  - Write the same metadata lines to user_errors.txt
  - Uses Strata.escapeSMTStringLit for SMT-LIB 2.7 string quoting

- **New test**: test_user_error_metadata.py — a Python file that calls an unknown method, triggering a user error with a known byte range
- **New expected output**: test_user_error_metadata.expected — validates the set-info lines appear before DETAIL/RESULT

## Testing

- New integration test with expected output files

### DIFF
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -28,6 +28,7 @@ import Strata.Backends.CBMC.CollectSymbols
 import Strata.Backends.CBMC.GOTO.CoreToGOTOPipeline
 
 import Strata.SimpleAPI
+import Strata.Util.Profile
 
 open Strata
 
@@ -496,6 +497,8 @@ def pyAnalyzeLaurelCommand : Command where
   args := [ "file" ]
   flags := [{ name := "verbose", help := "Enable verbose output." },
             { name := "no-solve", help := "Generate SMT-Lib files but do not invoke the solver." },
+            { name := "profile", help := "Print elapsed time for each pipeline step." },
+            { name := "quiet", help := "Suppress warnings on stderr." },
             checkModeFlag, checkLevelFlag,
             { name := "spec-dir",
               help := "Directory containing compiled PySpec Ion files.",
@@ -517,6 +520,8 @@ def pyAnalyzeLaurelCommand : Command where
   help := "Verify a Python Ion program via the Laurel pipeline. Translates Python to Laurel to Core, then runs SMT verification."
   callback := fun v pflags => do
     let verbose := pflags.getBool "verbose"
+    let profile := pflags.getBool "profile"
+    let quiet := pflags.getBool "quiet"
     let outputSarif := pflags.getBool "sarif"
     let filePath := v[0]
     let pySourceOpt ← tryReadPythonSource filePath
@@ -536,7 +541,9 @@ def pyAnalyzeLaurelCommand : Command where
       | some (pyPath, srcText) => some (pyPath, .ofString srcText)
       | none => none
     let combinedLaurel ←
-      match ← Strata.pyAnalyzeLaurel filePath dispatchModules pyspecModules sourcePath (specDir := specDir) |>.toBaseIO with
+      match ← Strata.pyAnalyzeLaurel filePath dispatchModules pyspecModules sourcePath
+                (specDir := specDir) (profile := profile)
+                (quiet := quiet) |>.toBaseIO with
       | .ok r => pure r
       | .error (.userCode range msg) =>
         let location := if range.isNone then "" else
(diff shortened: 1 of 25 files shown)


StrataMain.lean
```diff
@@ -545,6 +545,22 @@ def pyAnalyzeLaurelCommand : Command where
...
+        -- Also write the set-info metadata to user_errors.txt.
+        let mut lines := #[
+          s!"(set-info :file {Strata.escapeSMTStringLit filePath'})"
+        ]
+        unless range.isNone do
+          lines := lines.push s!"(set-info :start {range.start})"
+          lines := lines.push s!"(set-info :stop {range.stop})"
+        lines := lines.push s!"(set-info :error-message {Strata.escapeSMTStringLit msg})"
+        IO.FS.writeFile "user_errors.txt" (String.intercalate "\n" lines.toList ++ "\n")
         exitPyAnalyzeUserError s!"{msg}{location}"
```
MikaelMayer, line 568: The set-info lines are constructed identically twice: once for `IO.println` (lines 550–554) and once for the `lines` array (lines 557–562). Consider building the `lines` array first and then iterating over it for both printing and writing:

```lean
let mut lines := #[
  s!"(set-info :file {Strata.escapeSMTStringLit filePath'})"
]
unless range.isNone do
  lines := lines.push s!"(set-info :start {range.start})"
  lines := lines.push s!"(set-info :stop {range.stop})"
lines := lines.push s!"(set-info :error-message {Strata.escapeSMTStringLit msg})"
for line in lines do
  IO.println line
IO.FS.writeFile "user_errors.txt" (String.intercalate "\n" lines.toList ++ "\n")
```

This eliminates the duplication and ensures stdout and the file can never drift apart.

StrataTest/Languages/Python/run_py_analyze.sh
```diff
@@ -77,6 +77,23 @@ for test_file in tests/test_*.py; do
             else
                 echo "Test passed: " $base_name
             fi
+
+            # Check user_errors.txt if a .user_errors.expected file exists
+            user_errors_expected="${expected_dir}/${base_name}.user_errors.expected"
```
MikaelMayer, line 82: The `user_errors.txt` check block runs even when `$update -eq 1`, but there's no update logic for `.user_errors.expected` files (unlike the main `.expected` files which get overwritten on `--update`). This means `--update` won't regenerate the user_errors expected file. Consider adding an update branch here, similar to the one above:

```bash
if [ $update -eq 1 ] && [ -f "$user_errors_file" ]; then
    cp "$user_errors_file" "$user_errors_expected"
    echo "Updated: $user_errors_expected"
    rm -f "$user_errors_file"
elif [ -f "$user_errors_expected" ]; then
    # existing check logic...
fi
```

StrataTest/Languages/Python/run_py_analyze.sh
```diff
@@ -77,6 +77,23 @@ for test_file in tests/test_*.py; do
...
+                    echo "ERROR: user_errors.txt not generated for $base_name"
+                    failed=1
+                elif ! diff -q "$user_errors_expected" "$user_errors_file" > /dev/null; then
+                    echo "ERROR: user_errors.txt content for $base_name does not match expected"
+                    diff "$user_errors_expected" "$user_errors_file"
+                    failed=1
+                else
+                    echo "Test passed:  ${base_name} (user_errors.txt)"
+                fi
+                rm -f "$user_errors_file"
```
MikaelMayer: `user_errors.txt` is only cleaned up when a `.user_errors.expected` file exists. If a user-error test creates the file and a subsequent non-user-error test runs, the stale file persists in the repo root. Consider adding an unconditional `rm -f "$user_errors_file"` after the main diff check (outside the `if [ -f "$user_errors_expected" ]` block), or at the start of each test iteration.

PR #722, opened 2026-03-31
feat: VCOutcome uses unknown instead of sat for unvalidated SMT results

### LINKED ISSUE #721: VCOutcome should use "unknown" instead of "sat" until we have proper counter-example validation
If the SMT solver returns "sat", it does not necessarily mean that the assertion is actually "sat". Examples:
```
procedure f() returns (i: int) {
  return 1;
}
procedure g() {
  var x := f();
  assert x == 2;
}
```
Here the assertion is satisfiable in SMT but clearly not satisfiable in the original program.
Similarly, every time we have a computable function whose body cannot be given to SMT, we translate it to an uninterpreted function.
The problem is, SMT solvers can then assign any interpretation to this function, including interpretations that give values to the functions that the function cannot take.

It turns out, validating SAT is not sufficient at the core level. Every pipeline will have the responsibility of validating (or invalidating) a counter-example, because every step or phase in the pipeline might introduce computable but uninterpreted functions.

Because we want "sat" to be sound, here is what we should immediately do
* Add the possibility for the (unknown) case to have an optional model (in the Core language if possible)
* Keep the solver's outcome (sat + model, unsat, unknown) in the logs of the VCOutcome, e.g. in a separate field like "solverLog", which could actually consist of an ordered list of ("phase", sat + model / unknown + model / unsat), meaning that by the time we were done with the phase or step, the assertion has the satisfiability result, so that at a glance we can know which phase wasn't able to validate the counter-example. For now, there would be only one phase, "SMT"
* For now, every time the SMT solver returns SAT + model, we store the outcome to be UNKNOWN + model until we will have validation of the model. The field `solverLog` will preserve the SMT solver output as a log only.
* We can envision CEGAR-style iteration with the solver (counter-example guided abstraction refinement) but that's a future step.

MikaelMayer:
When the SMT solver returns "sat", it does not necessarily mean the property is truly satisfiable — uninterpreted functions may admit spurious models. This PR makes VCOutcome sound by treating unvalidated SMT "sat" results as "unknown" until counter-example validation is implemented.

Changes:
- `Result.unknown` now carries an optional candidate model (`candidateModel : Option (CounterEx Ident)`), distinguishing between an empty model (`some []`) and no model available (`none`). This allows unknown results to preserve counterexample data for future validation while clarifying that the model is unverified.
- `SolverPhaseLog` is a structure with `phase : String` and `result : SMT.Result` fields, so each log entry is labeled with the pipeline phase that produced it. `VCOutcome` gains a `solverLog : List SolverPhaseLog` field — an ordered list recording both the raw solver results (labeled `solver.sat` / `solver.val`) and per-phase adjusted results, for future diagnostic and traceability tooling.
- In the verification pipeline, every SMT "sat" result that we can't reliably validate is converted to "unknown" (with the model attached). The ways we know a SMT "sat" is preserved is that
  - There was no procedure eliminated from the entry of the procedure to the assertion point
  - There was no loop eliminated from the entry of the procedure to the assertion point.
  - For front-end, by default "sat" is not preserved until they implement a validation procedure, as front-ends might encode functions with uninterpreted functions.
- `PipelinePhase` structure couples each program-to-program transformation with its `AbstractedPhase` (model validation), ensuring transforms and their soundness annotations stay in sync. Each `AbstractedPhase` carries a `name` field that flows into the solver log. `corePipelinePhases` is the single parameterized function that drives both the pipeline and the validation — it accepts the target procedure list and factory, and returns the full pipeline including filtering, call/loop elimination, precondition elimination, and the final keep-set filter. `coreAbstractedPhases` is derived from it via `.map (.phase)`.
- All filter phases (FilterProcedures, keep-set filter) and PrecondElim are explicit pipeline phases marked as model-preserving, since they only remove procedures or add new assertions without introducing over-approximations.
- Pipeline phase types (`ModelValidation`, `AbstractedPhase`, `PipelinePhase`, `obligationHasLabelPrefix`, `modelPreservingPipelinePhase`) are defined in `Strata/Languages/Core/PipelinePhase.lean`. Each transform pass defines its own pipeline phase next to its implementation: `callElimPipelinePhase` in `CallElim.lean`, `loopElimPipelinePhase` in `LoopElim.lean`, `filterProceduresPipelinePhase` in `FilterProcedures.lean`, and `precondElimPipelinePhase` in `PrecondElim.lean`.
- Label prefixes used by call elimination (`callElimAssertPrefix`, `callElimAssumePrefix`) are defined in `CallElim.lean`. `createAsserts`/`createAssumes` accept a `labelPrefix` parameter, decoupling the generic helpers from the call-elimination-specific prefixes.
- `loopElimPipelinePhase` is placed last in the pipeline because loop elimination happens during evaluation (not as a program-to-program pass), making it the closest phase to SMT.
- Phase validation is non-cascading: each phase with `modelToValidate` independently validates the model, regardless of whether the current result is `.sat` or `.unknown`. A phase can demote `.sat m` to `.unknown (some m)` (when the model fails validation) or promote `.unknown (some m)` back to `.sat m` (when the model passes validation against the pre-phase semantics). Phases without a validator (`modelPreserving`) pass results through unchanged. When no candidate model is available (`.unknown none`), validation is skipped.
- In bug-finding modes (`bugFinding` and `bugFindingAssumingCompleteSpec`), unreachable assertions are now errors instead of warnings. Dead code in bug-finding contexts signals a specification or implementation problem. Added `docs/VerificationModes.md` documenting the three verification modes and their error classification, referenced from the README (in the "Running Analyses" section, after `lake exe strata verify` is introduced).
- Extracted `unreachableMsg` helper to deduplicate the check-mode/property match pattern in `VCOutcome.label` and `VCOutcome.emoji`.
- Per-pass obligation tests live next to their transforms: `StrataTest/Transform/CallElim.lean` and `StrataTest/Transform/LoopElim.lean`.

Test expectations updated to reflect the new behavior. Note that in many cases in Core, we keep "sat" unchanged. It's only changing what the front-end receive.

Fixes #721

### DIFF
--- a/Strata/DDM/Elab/Core.lean
+++ b/Strata/DDM/Elab/Core.lean
@@ -979,13 +979,14 @@ partial def inferType (tctx : TypingContext) (e : Expr) : ElabM TypeExpr := do
       some a.val[fnArgCount - i - 1]!
     let .ok tp := mtp
         | return panic! "Unexpected expandMacros failure."
-    let tp := Id.run <| tp.instTypeM fun _ i =>
+    let tp ← tp.instTypeM fun ann i => do
         assert! i < fnArgCount
         let lvl := fnArgCount - i - 1
         match a.val[lvl]! with
-        | .type tp => tp
-        | arg =>
-           panic! s!"Cannot instantiate type {repr tp} with args {repr a}"
+        | .type tp => pure tp
+        | _ =>
+           logError ann s!"Could not infer type parameter {i} for {ident}"
+           pure default
     return resultType! tctx tp (a.val.size - fnArgCount)
   | .app _ f a => panic! "Invalid app in result of Expr.hnf"
 
--- a/StrataTest/Languages/Core/Tests/PolymorphicDatatypeTest.lean
+++ b/StrataTest/Languages/Core/Tests/PolymorphicDatatypeTest.lean
@@ -532,3 +532,61 @@ Result: ✅ pass
 #eval verify nonDatatypeWithDatatypeArgPgm (options := .quiet)
 
 end Strata.PolymorphicDatatypeTest
+
+---------------------------------------------------------------------
+-- Regression test for issue #650: inferType panic with nested
+-- polymorphic datatypes and Sequence
+---------------------------------------------------------------------
+
+namespace Strata.InferTypePanicTest
+
+-- Verify that the program does not panic during type inference (issue #650).
+-- The program has type errors that should be reported gracefully.
+/--
+error: Could not infer type parameter 2 for Core.seq_select
+---
+error: Encountered .|| expression when MethodSetting expected.
+-/
+#guard_msgs in
+def issue650Pgm : Program :=
+#strata
+program Core;
+
+datatype Option (a : Type) { None(), Some(val: a) };
+
+datatype MethodSetting () {
+  MethodSetting_Cons(LoggingLevel: Option string)
+};
+
+datatype Stage () {
+  Stage_Cons(MethodSettings: Option (Sequence MethodSetting))
+};
+
+function method_ok(ms: MethodSetting): bool {
+  (MethodSetting..LoggingLevel(ms) == Some("ERROR"))
+  || (MethodSetting..LoggingLevel(ms) == Some("INFO"))
+}
+
+function stage_ok(stage: Stage): bool {
+  forall i: int ::
+    (Stage..MethodSettings(stage) != None()
+     && 0 <= i
+     && i < Sequence.length(Option..val(Stage..MethodSettings(stage))))
+    ==>
+    method_ok(Sequence.select(Option..val(Stage..MethodSettings(stage)), i))
+}
+
+const s: Stage;
+const grouped: Sequence Stage;
+axiom Sequence.length(grouped) == 1;
+axiom Sequence.select(grouped, 0) == s;
+
+procedure Check() returns ()
+{
+  assert [check]:
+    forall i: int :: (0 <= i && i < Sequence.length(grouped)) ==>
+      stage_ok(Sequence.select(grouped, i));
+};
+#end
+
+end Strata.InferTypePanicTest


Strata/Languages/Core/Verifier.lean
```diff
@@ -209,6 +216,9 @@ Unreachable covers display as ❌ (error) instead of ⛔ (warning).
 structure VCOutcome where
   satisfiabilityProperty : SMT.Result
   validityProperty : SMT.Result
+  /-- Ordered log of solver results per verification phase, preserving the
+      raw solver output before any soundness adjustments (e.g. sat→unknown). -/
+  solverLog : List SolverPhaseLog := []
```
MikaelMayer, line 262: No there should be one solverLog per SMT.Result. That way, `SolverPhaseLog` is just `SMT` for now.

Strata/Languages/Core/Verifier.lean
```diff
@@ -642,7 +662,10 @@ def verifySingleEnv (pE : Program × Env) (options : VerifyOptions)
       let (obligation, peSatResult?, peValResult?) ← preprocessObligation obligation p options satisfiabilityCheck validityCheck
       -- If PE resolved both checks, we're done
       if let (some peSat, some peVal) := (peSatResult?, peValResult?) then
-        let outcome := VCOutcome.mk peSat peVal
+        let outcome : VCOutcome := {
+          satisfiabilityProperty := peSat,
+          validityProperty := peVal,
+          solverLog := [{ phase := "PE", satisfiabilityResult := peSat, validityResult := peVal }] }
```
MikaelMayer: Here too, if we haven't validated the model, we should make sure sat is mapped to unknown.

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -47,7 +47,7 @@ Result: ✅ pass
 
 Obligation: reachable_cover
 Property: cover
-Result: ✅ pass
+Result: ❓ unknown
```
MikaelMayer: I will not compromise on soundness, but how hard would it be to actually try to evaluate the models to still have pass here?

StrataTest/Languages/Core/Examples/Cover.lean
```diff
@@ -172,15 +172,15 @@ Result: ❌ fail (❗path unreachable)
 
 Obligation: reach_assert_pass
 Property: assert
-Result: ✅ always true and is reachable from declaration entry
+Result: ✔️ always true if reached
```
MikaelMayer: I realized something that I want to make sure this PR is aware of: We should do this rewriting ONLY if there are things that are over approximated.
Hence, in all the benchmarks that contain only one procedure, and all functions used have SMT bodies or SMT interpretations, we should NOT transform SAT to unknown but keep SAT. Can you please do that?

Strata/Languages/Core/Verifier.lean
MikaelMayer: Simplified `SolverPhaseLog` to an abbreviation for `SMT.Result`, removing the `phase` field. Each SMT check produces its own log entry (one for satisfiability, one for validity), and the structure is now just `SMT.Result`.

Strata/Languages/Core/Verifier.lean
MikaelMayer: `satToUnknown` is now applied conditionally in the PE path as well, using the same `hasOverApprox` guard. Both `peSat` and `peVal` are adjusted through `satToUnknown` when over-approximations are present.

Strata/Languages/Core/Verifier.lean
```diff
@@ -614,6 +630,7 @@ def getObligationResult (assumptionTerms : List Term) (obligationTerm : Term)
     (obligation : ProofObligation Expression) (p : Program)
     (options : VerifyOptions) (counter : IO.Ref Nat)
     (tempDir : System.FilePath) (satisfiabilityCheck validityCheck : Bool)
+    (hasOverApprox : Bool)
```
MikaelMayer: Instead of hasOverApprox: Bool, I think I would like to pass a list of abstracted phases that the pipeline went through.
Each of these phases is described by
- Name
- If it has over approximations or not
- If it has over-approximations, a function that takes a model and say whether it's a valid model or not.

I will want some unit tests for that, and for now we will use only {name:="VerificationPipeline", modelValidation := ModelPreserving}
or
{name:="VerificationPipeline", modelValidation := ModelToValidate(model => /-TODO-/ false)}

so that architecturally everything is ready to map SAT to UNKNOWN if we can't validate the model.

MikaelMayer: Replaced `hasOverApprox: Bool` with `phases: List AbstractedPhase`. Each `AbstractedPhase` has a `name` and a `ModelValidation` which is either `modelPreserving` (sat results are sound) or `modelToValidate` (with a validation function that returns true when the model is valid).

`Program.abstractedPhases` builds the phases list: when over-approximations exist, it uses `modelToValidate (fun _ => false)` (TODO validator that always rejects); otherwise `modelPreserving`.

`SMT.Result.adjustForPhases` replaces `satToUnknown` — it validates sat models against all phases and converts to unknown only when validation fails.

Added unit tests in VCOutcomeTests covering: preserving phases keep sat, rejecting phases convert sat→unknown, accepting phases keep sat, mixed phases, unsat/unknown unchanged, empty phases.

Strata/Languages/Core/Verifier.lean
```diff
@@ -195,6 +195,37 @@ open Strata
...
+def AbstractedPhase.needsValidation (phases : List AbstractedPhase) : Bool :=
+  phases.any fun p => match p.modelValidation with
+    | .modelToValidate _ => true
+    | .modelPreserving => false
+
+/-- Validate a model against all phases. Returns true only if every
+    `modelToValidate` phase accepts the model. -/
+def AbstractedPhase.validateModel (phases : List AbstractedPhase)
+    (model : Imperative.SMT.CounterEx Expression.Ident) : Bool :=
+  phases.all fun p => match p.modelValidation with
```
MikaelMayer: Since the phases are recorded top-down, we should reverse the phases to start testing them from the last one.

Strata/Languages/Core/Verifier.lean
```diff
@@ -215,11 +215,12 @@ def AbstractedPhase.needsValidation (phases : List AbstractedPhase) : Bool :=
...
 
-/-- Validate a model against all phases. Returns true only if every
-    `modelToValidate` phase accepts the model. -/
+/-- Validate a model against all phases. Phases are recorded top-down,
+    so we reverse them to validate from the last (innermost) phase first.
+    Returns true only if every `modelToValidate` phase accepts the model. -/
 def AbstractedPhase.validateModel (phases : List AbstractedPhase)
     (model : Imperative.SMT.CounterEx Expression.Ident) : Bool :=
-  phases.all fun p => match p.modelValidation with
+  phases.reverse.all fun p => match p.modelValidation with
```
MikaelMayer: Good step.
But now I would like to go farther.
The goal is not to produce a boolean. The goal is to produce a final result + a list of intermediate solverLog, so that we can determine which phase wasn't able to validate the counter example. And that too should be unit tested.
It makes sense that the solverLog is a list where the deepest phases, the closest to SMT, is at the end of the list, so that by going through the solverLog list, we can retrace the why-s.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Already implemented — `validateModel` uses `phases.reverse.foldl` to process phases from innermost first.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Done. `validateModel` now returns `(SMT.Result, List SMT.Result)` — the adjusted result plus a log of intermediate results per phase. The log is ordered outermost-first with the deepest phase (closest to SMT) at the end, so walking the list retraces the validation chain. `adjustForPhases` propagates this tuple, and both the solver and PE paths append the phase logs to `solverLog`. Tests updated with log assertions.

Strata/Languages/Core/Verifier.lean
```diff
@@ -699,7 +774,17 @@ def verifySingleEnv (pE : Program × Env) (options : VerifyOptions)
       -- If PE resolved both checks, we're done, unless we always want to generate SMT queries
       if not options.alwaysGenerateSMT then
         if let (some peSat, some peVal) := (peSatResult?, peValResult?) then
-          let outcome := VCOutcome.mk peSat peVal
+          let phases := p.abstractedPhases
+          let (adjSat, satPhaseLog) := peSat.adjustForPhases phases
```
MikaelMayer: PE results are sound by construction (they come from syntactic evaluation, not the SMT solver), so calling `adjustForPhases` on them seems incorrect. For example, if `obligation.obligation.isFalse && obligation.assumptions.isEmpty`, PE returns `.sat []` — this is a genuine counterexample regardless of uninterpreted functions. Converting it to `.unknown` would lose a sound result.

The PR description says "PE-determined results are left unchanged since they are sound," but the code does adjust them here. Should this skip the phase adjustment and use `peSat`/`peVal` directly?

Strata/Languages/Core/Verifier.lean
```diff
@@ -195,6 +195,45 @@ open Strata
...
+  | modelToValidate (validate : Imperative.SMT.CounterEx Expression.Ident → Bool)
+
+/-- A phase in the verification pipeline, recording its name and
+    whether its models need validation. -/
+structure AbstractedPhase where
+  name : String
+  modelValidation : ModelValidation := .modelPreserving
+
+/-- True when any phase requires model validation. -/
+def AbstractedPhase.needsValidation (phases : List AbstractedPhase) : Bool :=
```
MikaelMayer: `needsValidation` is defined but never used in production code — only in tests. If it's intended for future use, a brief comment noting that would help. Otherwise, consider removing it and keeping only the test-level helpers.

Strata/Languages/Core/Verifier.lean
```diff
@@ -219,6 +258,9 @@ Unreachable covers display as ❌ (error) instead of ⛔ (warning).
 structure VCOutcome where
   satisfiabilityProperty : SMT.Result
   validityProperty : SMT.Result
+  /-- Ordered log of solver results per verification phase, preserving the
+      raw solver output before any soundness adjustments (e.g. sat→unknown). -/
+  solverLog : List SolverPhaseLog := []
```
MikaelMayer, line 262: `solverLog` is populated in both the PE and SMT paths but never read anywhere. Is there a consumer planned for a follow-up PR? If so, a brief note in the docstring (e.g. "consumed by ...") would help readers understand why it exists. If not, carrying this data through every `VCOutcome` adds complexity without benefit.

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +640,34 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+  | .sat _ => AbstractedPhase.validateModel phases r
+  | other => (other, [])
+
+/-- Build the list of abstracted phases for a program. When the program
+    has over-approximations, the phase uses a TODO validator that always
+    rejects the model. -/
+def Program.abstractedPhases (p : Program) : List AbstractedPhase :=
+  if p.hasOverApproximation then
+    [{ name := "VerificationPipeline",
+       modelValidation := .modelToValidate (fun _ => /- TODO -/ false) }]
```
MikaelMayer: The `TODO` comment on the validator is clear about intent. Minor note: the `name` field of `AbstractedPhase` is never read anywhere — it's set to `"VerificationPipeline"` in both branches but nothing consumes it. Consider dropping it until it's needed, or adding a comment about its intended use.

Strata/Languages/Core/Verifier.lean
```diff
@@ -641,9 +705,20 @@ def getObligationResult (assumptionTerms : List Term) (obligationTerm : Term)
...
     .error <| DiagnosticModel.fromFormat e
   | .ok (satResult, validityResult, estate) =>
-    -- Mask SMT results based on requested checks
-    let outcome := maskOutcome (VCOutcome.mk satResult validityResult) satisfiabilityCheck validityCheck
-    -- Extract counterexample model from sat results
+    -- Convert unvalidated sat results to unknown when phases require validation
+    let (adjSat, satPhaseLog) := satResult.adjustForPhases phases
+    let (adjVal, valPhaseLog) := validityResult.adjustForPhases phases
+    -- Build solver log: raw solver results followed by phase validation logs
+    let smtLog : List SolverPhaseLog :=
```
MikaelMayer: The solver log construction pattern (raw results + phase logs) is duplicated between this SMT path and the PE path at line 780. Consider extracting a helper like `buildSolverLog` to reduce duplication and ensure both paths stay in sync.

Strata/Languages/Core/Verifier.lean
```diff
@@ -195,6 +195,45 @@ open Strata
...
+  let (finalResult, revLog) := phases.reverse.foldl (init := (result, [])) fun (r, log) p =>
+    let r' := match r, p.modelValidation with
+      | .sat m, .modelToValidate f => if f m then .sat m else .unknown m
+      | _, _ => r
+    (r', r' :: log)
+  -- Reverse log so outermost is first, deepest is last
+  (finalResult, revLog.reverse)
+
+/-- Alias for solver log entries. Each entry records a single SMT result. -/
+abbrev SolverPhaseLog := SMT.Result
```
MikaelMayer: 👍 No action needed. `SolverPhaseLog` as a type alias is a nice touch for readability — it makes the intent of `List SolverPhaseLog` clearer than `List SMT.Result` even though they're the same type.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Fixed. PE results are now used directly without calling `adjustForPhases`. The PE path passes empty phase logs to `buildSolverLog`, preserving the sound PE results as-is.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Removed `needsValidation` from production code and moved it to the test file as a test-local helper, since it's only used there.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Added a note to the docstring: "Consumed by future diagnostic and traceability tooling." This clarifies the field's purpose for readers.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Removed the `name` field from `AbstractedPhase`. It wasn't consumed anywhere, so it's cleaner to add it back when there's an actual consumer.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Extracted a `buildSolverLog` helper that both the SMT path and PE path now use. This ensures both paths stay in sync for log construction.

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +634,41 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
         pure { obligation with assumptions := newAssumptions }
   return (obligation, peSatResult, peValResult)
 
+/-- True when the program contains at least one non-constructor function
+    without an SMT body, meaning the SMT encoding uses uninterpreted
+    functions and sat results may be spurious. -/
+def Program.hasOverApproximation (p : Program) : Bool :=
```
MikaelMayer: Ok, I realize something imprecise we are doing here.
Currently we only check for over-approximations in a Core program. However, if there are bodiless functions in a Core program, that does not constitute an over-approximation. The over-approximation is whatever translated something to Core. There must have been a phase before for the over-approximation to happen.
So over approximations happen in Core _only_ when 1) there are procedures _that are not inlined_ in the path of verifying the assertion from the beginning of the procedure where they are and 2) when there is a loop starter before the assertion before it got eliminated.

Since we have a brand new mechansm of SolverPhase, I'd like to start using it right away in this PR to represent all the other phases we are doing (filtering procedures, call elimination, loop elimination, perhaps others), and each phase would have a validator that can accurately determine if the assertion was in the path of something that is abstracted away. We will not validate the counter-example at this stage. We will only know it's validated if it's not in the path of something that was abstracted away.

All the phases OUTSIDE of Core Program pipeline (the front-ends) SHOULD  add a phase mechanism that for now simply transforms any SAT to UNKNOWN until we can do better like validate the model or determine that it does not need to be validated.

But at least, we can get precise results in Core.
Go ahead. It's a hard task but it's worth it as we lay a blueprint for making SAT not only sound (which it is now) but also possible down the road.

Strata/Languages/Core/Verifier.lean
MikaelMayer: That initial comment was wrong sorry. PE HAS to still adjust for phases, especially if we one day add phases that add constraints. For example, if a phase adds a CEGAR-based constraint `assume x == 1`, PE might one day evaluate `x == 1` to `false`, but going back this phase makes the model no a valid model anymore.
Revert to make sure we adjust for phases even in the case of PE.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Implemented per-obligation phase validation:

- `AbstractedPhase.getValidation` now takes a `ProofObligation` and returns the appropriate `ModelValidation` for that specific obligation.
- **`callElimPhase`**: checks if the obligation's path conditions contain `callElimAssume_` labels (from call elimination postcondition assumes). If present, SAT→UNKNOWN; otherwise model-preserving.
- **`loopElimPhase`**: checks for `assume_invariant_` or `assume_guard_` labels from loop elimination. Same logic.
- **Bodiless functions** in Core are no longer treated as Core over-approximations — removed `hasOverApproximation`. That responsibility belongs to the front-end.
- **`frontEndPhase`**: blanket SAT→UNKNOWN phase for front-ends (Boole, C_Simp, Laurel, Python) that translates to Core. Passed via the new `externalPhases` parameter on `Core.verify` and `Strata.verify`.
- `Core.verify`, `Core.verifyProgram`, and `Strata.verify` all accept `externalPhases` (default empty).
- Added unit tests for per-obligation precision: `callElimObligation`, `loopElimObligation`, `cleanObligation`, and `frontEndPhase` tests.
- Existing tests updated: obligations unaffected by abstractions now show definitive results.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Restored `adjustForPhases` on the PE path. PE results now go through the same phase validation as SMT results, using `externalPhases ++ p.abstractedPhases`.

Strata/Languages/Core/Verifier.lean
```diff
@@ -647,28 +647,31 @@ private def obligationHasLabelPrefix (obligation : ProofObligation Expression)
...
       .modelToValidate (fun _ => /- TODO -/ false)
     else .modelPreserving
 
-/-- Build the list of Core pipeline phases. Each phase checks per-obligation
-    whether the assertion is in the path of something abstracted. -/
-def Program.abstractedPhases (_p : Program) : List AbstractedPhase :=
+/-- The list of Core pipeline phases. Each phase checks per-obligation
+    whether the assertion is in the path of something abstracted.
+    This list must stay in sync with the transforms applied in `verify`. -/
+def coreAbstractedPhases : List AbstractedPhase :=
```
MikaelMayer: No, what I mean is that the problem is that this list is decoupled from the actual place where the phases are being applied to the program.

The phases that have to be applied to the program should be in a list that contain both the Program to Program transformations paired with their model validation, so that the list `coreAbstractedPhases` can be obtained by extracting the modelValidation field of that global list whereas the pipeline extracts the transformation field.

MikaelMayer: Introduced a `PipelinePhase` structure that pairs a `Program → CoreTransformM (Bool × Program)` transform with its `AbstractedPhase`. `corePipelinePhases` is the single list that drives both the pipeline (extracting `.transform`) and the validation (extracting `.phase` via `coreAbstractedPhases`). The targeted verification path now iterates over `corePipelinePhases` instead of manually calling `CallElim.callElim'`.

Strata/Languages/Core/Verifier.lean
```diff
@@ -888,14 +908,17 @@ def verify (program : Program)
...
-       -- replace all calls with contracts.
+       -- Verify specific procedures. Apply the pipeline transforms from
+       -- `corePipelinePhases` to inline contracts and eliminate loops.
+       -- Each pipeline phase pairs its transform with its model validation,
+       -- ensuring they stay in sync.
       let passes := fun prog => do
-        let prog ← FilterProcedures.run prog procs
-        let (_changed,prog) ← CallElim.callElim' prog
-        let prog ← runPrecondElim prog
+        let filtered ← FilterProcedures.run prog procs
```
MikaelMayer: All the filter procedures should also be phases that are accounted for in the pipeline. Determine if these phases change the SAT-ness. I don't think they do since they only remove information.

MikaelMayer: Done. FilterProcedures, PrecondElim, and the keep-set filter are now explicit `PipelinePhase` entries in `corePipelinePhases`. All three are model-preserving — they only remove procedures or add new assertions without introducing over-approximations that could produce spurious models.

`corePipelinePhases` and `coreAbstractedPhases` are now parameterized by the target procedure list and factory, so the full pipeline (including filter phases) is the single source of truth for both transforms and model validation. The targeted verification path in `verify` now iterates over this complete pipeline instead of manually applying each step.

Strata/Languages/Core/Verifier.lean
```diff
@@ -219,6 +256,10 @@ Unreachable covers display as ❌ (error) instead of ⛔ (warning).
 structure VCOutcome where
   satisfiabilityProperty : SMT.Result
   validityProperty : SMT.Result
+  /-- Ordered log of solver results per verification phase, preserving the
+      raw solver output before any soundness adjustments (e.g. sat→unknown).
+      Consumed by future diagnostic and traceability tooling. -/
```
MikaelMayer: The docstring says the log preserves "the raw solver output before any soundness adjustments" but `buildSolverLog` appends both the raw solver results *and* the phase validation logs (which are post-adjustment). Consider clarifying that the log contains both the original solver results and the per-phase adjusted results, so consumers know what to expect.

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +639,132 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+  phase.getValidation obligation :=
+    if obligationHasLabelPrefix obligation Transform.callElimAssumePrefix then
+      .modelToValidate (fun _ => /- TODO -/ false)
+    else .modelPreserving
+
+/-- Loop-elimination pipeline phase: the transform is applied during
+    evaluation (not as a program-to-program pass), so the transform here
+    is the identity. If the obligation's path includes labels from loop
+    elimination, the loop was replaced by an invariant-based encoding,
+    which is an over-approximation. Uses `loopElimInvariantPrefix` and
```
MikaelMayer: Since the loop-elimination transform happens during evaluation (not as a program-to-program pass), the identity transform `p => pure (false, p)` is the right choice here. The docstring explains this well — just noting this is a good design decision that keeps the `PipelinePhase` abstraction uniform while correctly handling the asymmetry.

Strata/Languages/Core/Verifier.lean
```diff
@@ -195,6 +195,43 @@ open Strata
...
+    (result : SMT.Result)
+    (obligation : ProofObligation Expression)
+    : SMT.Result × List SMT.Result :=
+  -- Process phases innermost-first; accumulate log entries in reverse
+  let (finalResult, revLog) := phases.reverse.foldl (init := (result, [])) fun (r, log) p =>
+    let validation := p.getValidation obligation
+    let r' := match r, validation with
+      | .sat m, .modelToValidate f => if f m then .sat m else .unknown m
+      | _, _ => r
+    (r', r' :: log)
```
MikaelMayer: Once a phase converts `.sat m` to `.unknown m`, all subsequent phases see `.unknown m` and pass it through unchanged (the `| _, _ => r` branch). This means the log entries after the first rejection all show `.unknown m` rather than what each phase would have decided independently. This is sound, but worth documenting — a consumer reading the log might expect each entry to reflect that phase's independent decision.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Fixed the docstring. It now reads: "Ordered log of solver results: the raw solver results followed by per-phase adjusted results (e.g. sat→unknown when a phase cannot validate the model)."

Strata/Languages/Core/Verifier.lean
MikaelMayer: Thanks — agreed, the identity transform keeps the PipelinePhase abstraction uniform while correctly reflecting that loop elimination happens during evaluation.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Documented this in the `validateModel` docstring. It now explicitly notes that validation is cascading — once a phase converts `.sat m` to `.unknown m`, subsequent phases pass it through unchanged, so log entries after the first rejection reflect the propagated `.unknown` rather than independent decisions.

MikaelMayer: Actually, I realize that it should be possible for a phase to transform unknown back to sat, if that phase has the ability to test the model pre-phase. Please take this into account now that we realize it.
Each phase can independently validate a model that was either SAT or unknown.

MikaelMayer: Done. `validateModel` now handles `.unknown m` the same way as `.sat m` — when a phase has `modelToValidate`, it independently validates the model regardless of the current status. If the model passes, `.unknown m` is promoted back to `.sat m`; if it fails, it stays `.unknown m`. `adjustForPhases` now passes both `.sat` and `.unknown` results through `validateModel`.

Updated the docstring to reflect that phases are not cascading — each validating phase makes its own decision. Added tests for unknown→sat promotion (accepting validator) and unknown staying unknown (rejecting validator and preserving phase).

Strata/Languages/Core/Verifier.lean
MikaelMayer: Already addressed — all filter procedures (`filterProceduresPipelinePhase`, `keepSetFilterPipelinePhase`, `precondElimPipelinePhase`) are `PipelinePhase` entries in `corePipelinePhases`, marked as `modelPreserving` since they only remove or add information without introducing over-approximations that could produce spurious models.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Already implemented — `validateModel` handles both `.sat m` and `.unknown m` independently when a phase has `modelToValidate`. Each phase can promote `.unknown m` → `.sat m` (model passes validation) or demote `.sat m` → `.unknown m` (model fails). Tests at lines 239–249 of VCOutcomeTests.lean verify unknown→sat promotion, unknown staying unknown with rejecting validator, and unknown staying unknown with preserving phase.

docs/VerificationModes.md
```diff
@@ -0,0 +1,39 @@
+# Verification Modes
```
MikaelMayer, line 1: This file is stand-alone. Could you please reference it somewhere?

MikaelMayer: Added a reference to `docs/VerificationModes.md` in the README intro paragraph, alongside the existing Architecture and Getting Started links.

README.md
```diff
@@ -11,8 +11,10 @@ developers to customize additional features to their needs.
 This README will help you get started with using and extending
 Strata. Also see the [Architecture](docs/Architecture.md) document
 that introduces some terminology and describes Strata's components,
-and a [Getting Started](docs/GettingStarted.md) guide that describes
-how to create a new dialect and analysis using existing features.
+a [Getting Started](docs/GettingStarted.md) guide that describes
+how to create a new dialect and analysis using existing features,
+and [Verification Modes](docs/VerificationModes.md) for details on
+the deductive and bug-finding verification modes.
```
aqjune-aws: I think it is more natural for mentioning `Verification Modes` to appear after `lake exec strata verify` is introduced because VerificationModes.md starts with explaining `--check-mode`. What about moving the new clause to '## Running Analyses on Existing Strata Programs'?

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -50,7 +50,7 @@ inductive Result (Ident : Type) where
   -- Also see Strata.SMT.Decision.
   | sat (cex : CounterEx Ident)
   | unsat
-  | unknown
+  | unknown (model : CounterEx Ident := [])
```
aqjune-aws: I am hesitant to have the model field of `unknown` means here because this can be interpreted by users as "Strata can use a special SMT solver that can also return a model even if the result of SMT solver was unknown" (since the file name is SMTUtils.lean). But I think we still agree that SMT solvers are not special, and if they returned `unknown`, the model is not available (as SMT-LIB states).

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +652,134 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+  let filterPhases := match procs with
+    | some ps => [filterProceduresPipelinePhase ps]
+    | none => []
+  let precondPhase := match factory with
+    | some f => [precondElimPipelinePhase f]
+    | none => []
+  let keepSetPhase := match procs with
+    | some ps => [keepSetFilterPipelinePhase ps]
+    | none => []
+  filterPhases ++ [callElimPipelinePhase, loopElimPipelinePhase] ++ precondPhase ++ keepSetPhase
```
aqjune-aws: Why is loopElimPipelinePhase added to the core pipeline?

Strata/Transform/CoreTransform.lean
```diff
@@ -22,6 +22,9 @@ open LabelGen
 def oldVarPrefix (id : String) : String := s!"old_{id}"
 def tmpVarPrefix (id : String) : String := s!"tmp_{id}"
 
+/-- Label prefix for call-elimination assume statements. -/
+def callElimAssumePrefix : String := "callElimAssume_"
```
aqjune-aws: It seems the existence of `"callElimAssume_"` in `createAssumes` (as well as is `"callElimAssert_"` in `createAsserts`) is unfortunate. Can we move `callElimAssumePrefix` to the CallElim pass file, and let `createAsserts`/`createAssumes` receive additional argument for their prefixes?

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +652,134 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+    labels from call elimination, the callee body was replaced by its
+    contract, which is an over-approximation. Uses
+    `Transform.callElimAssumePrefix` as the single source of truth for the
+    label prefix. -/
+def callElimPipelinePhase : PipelinePhase where
+  transform := CallElim.callElim'
+  phase.getValidation obligation :=
+    if obligationHasLabelPrefix obligation Transform.callElimAssumePrefix then
+      .modelToValidate (fun _ => /- TODO -/ false)
+    else .modelPreserving
```
aqjune-aws: What do you think about moving the `ModelValidation` information to individual pass files? Like we have `loopElimInvariantPrefix` in LoopElim.lean .
This will be helpful to link the soundness proof of the passes to what the pass explicitly states about approximation or etc. For example, the current design will allow `CallElimCorrect` to prove that it is preserving whereas `callElimPipelinePhase` was mistakenly describing as not-preserving. Even if it is sound, it loses precision.

Strata/DL/Imperative/SMTUtils.lean
aqjune-aws: I am aware that this is datatype is useful, but since "unknown" from SMT solver is conceptually different from "unknown" from Strata's verifier, what about just keeping the definition of this datastructure as it is and adjusting the interpretation of 'sat cex' at places which use this data structure.

MikaelMayer: Actually I think adding a model to unknown is a right move. SMT solvers can provide models even if the result is "unknown", these are models it couldn't disprove, and more than once I've seen them actually identify correct counter-examples in Dafny. It's just that now we have the ability to validate them.
I don't think "unknown" from SMT has to be different than "unknown" from Strata. I would prefer both to have the same meaning: We were not able to determine SAT or UNSAT, but we have a candidate model if possible that could be SAT.
SAT brings a soundness claim that UNKNOWN does not, so if you are ok with it, I prefer to keep unknown with a model.

atomb: In practice, some solvers _can_ produce a model when the result is `unknown` (even if SMT-LIB states that there isn't one). And very frequently it's a valid model. It will require validating, but that's the purpose of this PR altogether. So I'm in favor of keeping this, both for conversions from `sat` to `unknown` and for models retrieved from `uknown` results. But perhaps it should be named something like `potentialModel` or `candidateModel` to make it clear that it might not be a real model.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Good question. I would have preferred that it be a phase on its own, but it's currently done on the fly in the partial evaluator. Since the model validation goes back through all the transformations that were applied until SMT, I added this phase that does nothing from a standpoint of transforming Core but has the ability to validate counter-examples on the way back. Maybe I should move it to the end since it's a phase that is done just before SMT?

MikaelMayer: Moved `loopElimPipelinePhase` to the end of `corePipelinePhases`. Since loop elimination happens during evaluation (not as a program-to-program pass), it's the closest phase to SMT and belongs at the end of the pipeline. Updated the docstring to document this ordering rationale.

README.md
MikaelMayer: Done. Moved the Verification Modes reference from the intro paragraph to the "Running Analyses" section, after `lake exe strata verify` and the description of what it does.

Strata/Transform/CoreTransform.lean
MikaelMayer: Done. Moved `callElimAssumePrefix` to `CallElim.lean` and added `callElimAssertPrefix` there too. `createAsserts` and `createAssumes` now take a `labelPrefix` parameter (defaulting to `"assert_"` and `"assume_"` respectively), and `CallElim.lean` passes the call-elimination-specific prefixes explicitly.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Done. Created `Strata/Languages/Core/PipelinePhase.lean` with the shared types (`ModelValidation`, `AbstractedPhase`, `PipelinePhase`, `obligationHasLabelPrefix`, `modelPreservingPipelinePhase`). Moved `callElimPipelinePhase` to `CallElim.lean` and `loopElimPipelinePhase` to `LoopElim.lean`, so each pass file declares its own model validation behavior next to the transform implementation. This way correctness proofs like `CallElimCorrect` can reference the same `callElimPipelinePhase` that the pipeline uses.

Strata/DL/Imperative/SMTUtils.lean
aqjune-aws: OK... It sounds like a useful feature of practical SMT solvers then.
Then, I am in favor of clarifying that the model is candidate (`candidateModel`), and also want to make the argument optional `Option (CounterEx Ident)` - this will help distinguish between (1) an empty model `[]` and (2) the model unavailable `None` because the solver was not capable of building one.

Strata/Transform/LoopElim.lean
```diff
@@ -183,4 +189,21 @@ def Stmt.removeLoops
 
 end -- public section
 
+public section
```
aqjune-aws: Let's combine `end` + `public section` to a nil string :)

Strata/DL/Imperative/SMTUtils.lean
shigoel: +1 to June's comment above, FWIW

Strata/Languages/Core/PipelinePhase.lean
```diff
@@ -0,0 +1,64 @@
...
+
+public section
+
+/-- Describes whether a pipeline phase preserves models or requires validation. -/
+inductive ModelValidation where
+  /-- The phase preserves models — sat results are sound. -/
+  | modelPreserving
+  /-- The phase may introduce spurious models. The function returns true
+      when the model is valid. -/
+  | modelToValidate (validate : Imperative.SMT.CounterEx Expression.Ident → Bool)
```
shigoel, line 30: Out of scope for this PR, but I'd prefer to change the name of `Imperative.SMT.CounterEx` to `Imperative.SMT.Model` soon. Not all models from solvers are counterexamples (e.g., for cover)

StrataTest/Languages/Core/VCOutcomeTests.lean
```diff
@@ -182,4 +182,119 @@ info: === Outcome Table (assert) ===
...
+#guard (satResult.adjustForPhases [] dummyObligation).1 == satResult
+#guard (satResult.adjustForPhases [] dummyObligation).2 == []
+
+/-! ### Per-obligation phase validation tests -/
+
+/-- Obligation with call-elimination labels in path conditions. -/
+private def callElimObligation : Imperative.ProofObligation Core.Expression :=
+  { label := "test_callElim", property := .assert,
+    assumptions := [[("callElimAssume_post", .true ())]],
+    obligation := .true (), metadata := {} }
```
aqjune-aws, line 262: Let's move this test to StrataTest/Transform/CallElim.lean (and similarly for `loopElimObligation`). Ideally we will want each pass to have this `obligationHasLabelPrefix` calling test.

Strata/Languages/Core/Verifier.lean
aqjune-aws: Actually I am in favor of directly using `SMT.Result` then - it needs less 'go to definition' clicks :)

Strata/Languages/Core/Verifier.lean
```diff
@@ -324,8 +359,10 @@ def label (o : VCOutcome) (property : Imperative.PropertyType)
     (checkLevel : CheckLevel) (checkMode : VerificationMode) : String :=
   -- Unreachable is detected when both checks ran (via fullCheck annotation or full level)
   if o.unreachable then
-    if property.passWhenUnreachable then "pass (❗path unreachable)"
-    else "fail (❗path unreachable)"
+    match checkMode with
+    | .deductive => if property.passWhenUnreachable then "pass (❗path unreachable)"
+                    else "fail (❗path unreachable)"
+    | _ => "fail (❗path unreachable)"
```
aqjune-aws: It seems this match pattern appears frequently. Should we write a polymorphic wrapper that receives `checkMode`, `property.passWhenUnreachable` and the two possible messages, and print the message accordingly?

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +637,87 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+    let filtered ← FilterProcedures.run prog procs
+    return (true, filtered)
+
+/-- PrecondElim pipeline phase: generates well-formedness checks for
+    partial-function preconditions. Model-preserving because it only adds
+    new assertions and procedures without abstracting existing ones. -/
+def precondElimPipelinePhase
+    (factory : @Lambda.Factory CoreLParams) : PipelinePhase :=
+  modelPreservingPipelinePhase fun prog => do
+    PrecondElim.precondElim prog factory
```
aqjune-aws: What about moving `filterProceduresPipelinePhase` and `precondElimPipelinePhase` to the corresponding files as well?

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +637,87 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+    PrecondElim). Model-preserving because it only removes procedures. -/
+def keepSetFilterPipelinePhase (procs : List String) : PipelinePhase :=
+  modelPreservingPipelinePhase fun prog => do
+    let keepSet := Std.HashSet.ofList
+      (procs ++ procs.map PrecondElim.wfProcName)
+    let result := { prog with decls := prog.decls.filter fun d =>
+      match d with
+      | .proc p _ => keepSet.contains (CoreIdent.toPretty p.header.name)
+      | _ => true }
+    return (true, result)
```
aqjune-aws, line 663: Did you manage to find a counterexample that makes behavior of `Keep-set filter pipeline` and `ProcedureElim` diverge, after your patch..?

MikaelMayer: I haven't looked into that, is that question for https://github.com/strata-org/Strata/pull/704 instead?

Strata/DL/Imperative/SMTUtils.lean
MikaelMayer: Done. Renamed the field from `model` to `candidateModel` and changed its type to `Option (CounterEx Ident)` (defaulting to `none`). This distinguishes between an empty model `some []` and no model available `none`. The `validateModel` function now only attempts validation when a candidate model is present (`.unknown (some m)`); `.unknown none` passes through unchanged.

Strata/Languages/Core/Verifier.lean
MikaelMayer: I realize I forgot to label these SMT results according to the phase.
Let's do the following: Each phase should have a name, a transformation phase, and a model validation.
The SolverPhaseLog should be a structure that contains a phase name + an SMT result, so that it's clearler that the SMT result is the one for after the phase name.

Strata/Transform/LoopElim.lean
MikaelMayer: Done. Removed the redundant `end` + `public section` pair — the definitions are now in the same section as the preceding code.

StrataTest/Languages/Core/VCOutcomeTests.lean
MikaelMayer: Done. Moved the per-pass obligation tests: callElim tests to `StrataTest/Transform/CallElim.lean` and loopElim tests to a new `StrataTest/Transform/LoopElim.lean`. The combined-phase and frontEnd tests remain in VCOutcomeTests.lean since they test cross-phase integration.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Done. Extracted a polymorphic `unreachableMsg` helper that takes `checkMode`, `passWhenUnreachable`, and the two possible messages. Both `label` and `emoji` now use it instead of duplicating the match pattern.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Done. Moved `filterProceduresPipelinePhase` to `Strata/Transform/FilterProcedures.lean` and `precondElimPipelinePhase` to `Strata/Transform/PrecondElim.lean`, next to their respective transform implementations.

Strata/Languages/Core/Verifier.lean
aqjune-aws: Aha yes. It seems the agent agreed using FilterProcedure is fine. :) Probably we can update the pull request after this one is finished.

Strata/Languages/Core/Verifier.lean
aqjune-aws: Sounds like a great idea :)

MikaelMayer: `SolverPhaseLog` is now a structure with `phase : String` and `result : SMT.Result` fields, so each log entry is labeled with the pipeline phase that produced it. `AbstractedPhase` has a `name` field again, and all phases carry their names: `FilterProcedures`, `CallElim`, `PrecondElim`, `KeepSetFilter`, `LoopElim`, `FrontEnd`. Raw SMT results in `buildSolverLog` use `SMT.sat` / `SMT.val` as phase labels. Tests updated.

Strata/Transform/CallElim.lean
```diff
@@ -127,6 +136,19 @@ def callElim' (p : Program) : CoreTransformM (Bool × Program) :=
...
+/-- Call-elimination pipeline phase: the transform replaces procedure calls
+    with assert/havoc/assume sequences. If the obligation's path includes
+    labels from call elimination, the callee body was replaced by its
+    contract, which is an over-approximation. -/
+def callElimPipelinePhase : PipelinePhase where
+  transform := CallElim.callElim'
+  phase.name := "CallElim"
+  phase.getValidation obligation :=
+    if obligationHasLabelPrefix obligation CallElim.callElimAssumePrefix then
+      .modelToValidate (fun _ => /- TODO -/ false)
```
atomb, line 149: Does this imply that we'll never be able to validate counterexamples from any program that contains procedure calls?

Strata/Transform/LoopElim.lean
```diff
@@ -181,6 +187,20 @@ def Stmt.removeLoops
...
+    is the identity. If the obligation's path includes labels from loop
+    elimination, the loop was replaced by an invariant-based encoding,
+    which is an over-approximation. -/
+def loopElimPipelinePhase : PipelinePhase where
+  transform p := pure (false, p)
+  phase.name := "LoopElim"
+  phase.getValidation obligation :=
+    if obligationHasLabelPrefix obligation loopElimInvariantPrefix
+       || obligationHasLabelPrefix obligation loopElimGuardPrefix then
+      .modelToValidate (fun _ => /- TODO -/ false)
```
atomb, line 201: Same comment as for CallElim.

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +647,70 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+private def buildSolverLog (satResult valResult : SMT.Result)
+    (satisfiabilityCheck validityCheck : Bool)
+    (satPhaseLog valPhaseLog : List SolverPhaseLog) : List SolverPhaseLog :=
+  (if satisfiabilityCheck then [{ phase := "SMT.sat", result := satResult }] else []) ++
+  (if validityCheck then [{ phase := "SMT.val", result := valResult }] else []) ++
+  satPhaseLog ++ valPhaseLog
+
+/-- Adjust an SMT result through pipeline phase validation. A `.sat` result
+    may be demoted to `.unknown` if a phase cannot validate the model, and
+    an `.unknown` result may be promoted back to `.sat` if a phase can
```
atomb, line 706: This second property is very nice. It allows us to make solid use of models that come along with `unknown` results from the solver.

shigoel (approved):

Strata/Transform/CallElim.lean
MikaelMayer: Nothing in this PR is doing any validation of models yet, we only conservatively evaluate whether an assertion's model could still be valid in the light of the code that was abstracted before, which as of today makes assertions after procedure calls and while loops not checked.

In the future, I envision that each phase will actively try to validate the counterexample (or model for covers or reachability checks) by doing some kind of interpretation.

Strata/Languages/Core/Verifier.lean
```diff
@@ -195,6 +195,41 @@ open Strata
...
+    (result : SMT.Result)
+    (obligation : ProofObligation Expression)
+    : SMT.Result × List SolverPhaseLog :=
+  -- Process phases innermost-first; accumulate log entries in reverse
+  let (finalResult, revLog) := phases.reverse.foldl (init := (result, [])) fun (r, log) p =>
+    let validation := p.getValidation obligation
+    let r' := match r, validation with
+      | .sat m, .modelToValidate f => if f m then .sat m else .unknown (some m)
+      | .unknown (some m), .modelToValidate f => if f m then .sat m else .unknown (some m)
+      | _, _ => r
```
tautschnig, line 229: Nit: The catch-all silently handles `.unknown none` (no candidate model) by passing it through unchanged. This is correct, but since the docstring above explicitly calls out this case ("When no candidate model is available, validation is skipped"), consider making that explicit here:
```suggestion
      -- .unknown none or .modelPreserving: pass through unchanged
      | _, _ => r
```

Strata/Languages/Core/Verifier.lean
```diff
@@ -456,20 +505,13 @@ structure VCResult where
     also determined satisfiability, we mask it to `.unknown`. -/
 def maskOutcome (outcome : VCOutcome) (satisfiabilityCheck validityCheck : Bool) : VCOutcome :=
   if satisfiabilityCheck && validityCheck then
-    -- Both checks requested: return outcome as-is
     outcome
   else if validityCheck && !satisfiabilityCheck then
-    -- Only validity requested: mask satisfiability to unknown
-    { satisfiabilityProperty := .unknown,
-      validityProperty := outcome.validityProperty }
+    { outcome with satisfiabilityProperty := .unknown }
```
tautschnig, line 512: Nit: `maskOutcome` writes `.unknown` (which defaults to `candidateModel := none`). If `adjustForPhases` already ran and produced `.unknown (some m)`, the candidate model is discarded here. This is fine since the masked property wasn't requested, and the `solverLog` field preserves the pre-mask data. But a brief comment noting this would help future readers:
```suggestion
    -- Mask to .unknown none — the solverLog preserves the original result
    { outcome with satisfiabilityProperty := .unknown }
```

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +647,70 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+def coreAbstractedPhases (procs : Option (List String) := none)
+    (factory : Option (@Lambda.Factory CoreLParams) := none) : List AbstractedPhase :=
+  (corePipelinePhases procs factory).map (·.phase)
+
+/-- Build the solver log from raw results and phase validation logs. -/
+private def buildSolverLog (satResult valResult : SMT.Result)
+    (satisfiabilityCheck validityCheck : Bool)
+    (satPhaseLog valPhaseLog : List SolverPhaseLog) : List SolverPhaseLog :=
+  (if satisfiabilityCheck then [{ phase := "SMT.sat", result := satResult }] else []) ++
+  (if validityCheck then [{ phase := "SMT.val", result := valResult }] else []) ++
```
tautschnig: Nit: The phase labels `"SMT.sat"` and `"SMT.val"` could be confused with the `SMT.Result.sat` constructor when reading the log. Consider renaming to make it clearer these are the raw solver results before phase adjustment:
```suggestion
  (if satisfiabilityCheck then [{ phase := "solver.sat", result := satResult }] else []) ++
  (if validityCheck then [{ phase := "solver.val", result := valResult }] else []) ++
```

Strata/Languages/Core/Verifier.lean
```diff
@@ -605,6 +647,70 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
+  let filterPhases := match procs with
+    | some ps => [filterProceduresPipelinePhase ps]
+    | none => []
+  let precondPhase := match factory with
+    | some f => [precondElimPipelinePhase f]
+    | none => []
+  let keepSetPhase := match procs with
+    | some ps => [keepSetFilterPipelinePhase ps]
+    | none => []
+  filterPhases ++ [callElimPipelinePhase] ++ precondPhase ++ keepSetPhase ++ [loopElimPipelinePhase]
```
tautschnig, line 689: Question: `corePipelinePhases` always appends `loopElimPipelinePhase` (with its identity transform) even when `procs` is `none` (full program verification). This means the non-targeted verification path also runs the identity transform and includes loop-elim validation in the abstracted phases. This is correct for soundness, but was this intentional for the non-targeted path too, or should `loopElimPipelinePhase` only be included when `procs` is `some`?

Strata/Languages/Core/Verifier.lean
```diff
@@ -850,18 +969,28 @@ def Core.getProgram
   (ictx : InputContext := Inhabited.default) : Core.Program × Array String :=
   TransM.run ictx (translateProgram p)
 
+/-- Front-end phase: any translation from a source language to Core may
+    introduce over-approximations. Until front-ends can validate models or
+    determine that an assertion is unaffected, all sat results are converted
+    to unknown. -/
+def frontEndPhase : Core.AbstractedPhase where
```
tautschnig, line 978: Nit: `frontEndPhase` is defined here in the `Strata` namespace inside `Core/Verifier.lean`. If other front-ends (C_Simp, Dyn, Laurel) later need their own phases with different validation logic, this location may not scale well. Consider adding a comment noting this is a shared default for all front-ends, or whether it should eventually move to a more general location.

Strata/Languages/Core/Verifier.lean
MikaelMayer: It was intentional. loopElimPipelinePhase should always be included.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Added the comment to the catch-all case as suggested.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Added the comment noting that the candidate model is intentionally discarded and the solverLog preserves the original result.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Renamed the phase labels from `"SMT.sat"`/`"SMT.val"` to `"solver.sat"`/`"solver.val"` to avoid confusion with the `SMT.Result.sat` constructor.

Strata/Languages/Core/Verifier.lean
MikaelMayer: I think the comment already states that.

Strata/Languages/Core/PipelinePhase.lean
```diff
@@ -0,0 +1,67 @@
...
+-/
+module
+
+public import Strata.Transform.CoreTransform
+public import Strata.DL.Imperative.SMTUtils
+public import Strata.DL.Imperative.EvalContext
+
+/-! # Pipeline Phase Definitions for Model Validation
+
+This module defines the types used to describe how verification pipeline
```
shigoel, line 14: This mechanism is super nice!

PR #723, opened 2026-03-31
Filter Laurel prelude to only transitively needed declarations

joehendrix:
Reduce verification time by filtering the Laurel prelude and restricting verification to user procedures. 

### 1. Filter pyspec procedures from verification targets (`splitProcNames`)

Based on a fix by June Lee (commit b6f90973). `splitProcNames` was classifying pyspec procedures (e.g. service stubs) as user procedures because they carry non-empty file range metadata from their `.pyspec.st.ion` source files. This caused all pyspec procedures to be passed as verification targets, generating hundreds of unnecessary VCs.

Add a `userSourcePaths` parameter so callers can positively identify user source files. Procedures from the user's Python file are inlined and verified; pyspec procedures are still inlined (so their preconditions are checked at call sites) but not independently verified. The legacy heuristic is preserved when no paths are supplied.

### 2. Filter Laurel prelude to transitively needed declarations (`FilterPrelude`)

The prelude contains all PySpec-generated procedures and types, but a typical user program only references a small fraction. The new `Strata.Languages.Laurel.FilterPrelude` module:

1. **Collects names** from the user program — `StaticCall`/`InstanceCall` targets and `UserDefined`/`New`/`extending` type references.

2. **Builds a dependency map** over the prelude, with alias entries for datatype constructors, testers, and destructors. Procedures with `invokeOn` clauses create reverse dependencies so axiom procedures (e.g. `List_len_pos`) are automatically included when their target functions are needed.

3. **Computes a BFS transitive closure** from the user's referenced names, then filters the prelude's `staticProcedures` and `types` to the reachable set.

### 3. Harden FilterPrelude and de-duplicate alias naming

Addresses review feedback:

- **Reads clause**: `collectProcDeps` now traverses the `reads` clause from `proc.determinism`, ensuring names referenced there are captured.
- **Duplicate-name detection**: `buildDependencyMap` returns `Except String` and uses a `DepM`/`insertNew` pattern that errors on duplicate name bindings instead of silently overwriting. This caught a real bug — the tester alias `"isTypeError"` (unqualified) collided with an explicit helper procedure of the same name.
- **Fixed tester alias**: Changed from `"is" ++ ctor.name` to `"D..isCtor"` (qualified form), matching how `Resolution.resolveTypeDefinition` registers testers.
- **Shared naming helpers**: Extracted `DatatypeDefinition.testerName`, `.destructorName`, and `.unsafeDestructorName` into `Laurel.lean` so both `FilterPrelude` and `Resolution` use the same canonical name construction.

## Expected file changes

The smaller prelude produces simpler SMT formulas, allowing the solver to find concrete counterexamples where it previously timed out (`unknown` → `fail`). All 6 changes represent more precise results, not regressions.

| File | Line | Before | After | Reason |
|------|------|--------|-------|--------|
| test_class_field_any.py | 6 | unknown | fail | Field typed as `Any`; solver proves equality doesn't hold generically |
| test_for_loop.py | 15 | unknown | fail | `items: Any` makes loop contents indeterminate; solver finds counterexample |
| test_if_elif.py | 6 | unknown | fail | `elif` condition reachability; smaller formula lets solver disprove |
| test_variable_reassign.py | 16 | unknown | fail | Loop sum requires invariant inference; solver finds counterexample |
| test_while_loop.py | 7 | unknown | fail | Loop sum same pattern; solver finds counterexample |
| test_pin_any.py | 3-4 | unknown | concrete | `PIn` and `Any_to_bool` on `Any` parameter resolve to definite outcomes |

### DIFF
--- a/StrataTest/Languages/Python/expected_laurel/test_arithmetic.user_errors.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_arithmetic.user_errors.expected
@@ -1,4 +1,0 @@
-(set-info :file "/local/home/joehx/features/StrataInternal/kiro_tests_annotated/test_bedrock_guardrails.py")
-(set-info :start 352)
-(set-info :stop 376)
-(set-info :error-message "'create_guardrail' called with missing required arguments: [blockedInputMessaging, blockedOutputsMessaging]")


Strata/Languages/Laurel/FilterPrelude.lean
```diff
@@ -0,0 +1,242 @@
...
+  | .External => pure ()
+
+/-- Collect all names referenced by a procedure (signature + body). -/
+private partial def collectProcDeps (proc : Procedure) : CollectM Unit := do
+  proc.inputs.forM  fun p => collectHighTypeNames p.type
+  proc.outputs.forM fun p => collectHighTypeNames p.type
+  proc.preconditions.forM collectExprNames
+  match proc.decreases with | some d => collectExprNames d | none => pure ()
+  match proc.invokeOn with | some t => collectExprNames t | none => pure ()
+  collectBodyNames proc.body
```
MikaelMayer, line 149: `collectProcDeps` traverses inputs, outputs, preconditions, decreases, invokeOn, and the body — but skips the `reads` clause inside `proc.determinism`. A `Determinism.deterministic (some reads)` carries a `StmtExprMd` that could reference procedure or type names.

In practice the reads clause in the prelude likely only mentions identifiers/field-selects that are already pulled in by other paths, so this is unlikely to cause a missing-declaration bug today. But for completeness:

```lean
match proc.determinism with
| .deterministic (some reads) => collectExprNames reads
| _ => pure ()
```

Strata/Languages/Laurel/FilterPrelude.lean
```diff
@@ -0,0 +1,242 @@
...
+  let mut m : Std.HashMap String (Std.HashSet String) := {}
+  for proc in prog.staticProcedures do
+    m := m.insert proc.name.text (runCollect (collectProcDeps proc)).allNames
+  for td in prog.types do
+    let name := td.name.text
+    let deps := (runCollect (collectTypeDefDeps td)).allNames
+    m := m.insert name deps
+    if let .Datatype dt := td then
+      for c in dt.constructors do
+        m := m.insert c.name.text (deps.insert name)
```
MikaelMayer: Nit: if two datatypes ever share a constructor name (or a constructor name collides with a procedure name), `m.insert` silently overwrites the earlier entry's dependency set. Currently the runtime datatypes have unique constructor names so this is fine, but a defensive approach would merge rather than replace:
```lean
let existing := m[c.name.text]?.getD {}
m := m.insert c.name.text (existing.union (deps.insert name))
```
Low priority — just flagging for future-proofing.

joehendrix: This could hide an error too, so I've changed the code to catch name collisions and report errors if they ever occur.  We should get a heads up in testing now.

In fact the checking actually caught a naming mistake where FilterPrelude was generating incorrect names for constructor tests.  That's been fixed and I made a common naming function that Resolution and Filtering both use.

Strata/Languages/Laurel/FilterPrelude.lean
joehendrix: Thanks, I've added this and audited the code to make sure we are catching all names.  This uncovered some unchecked assumptions that constants and other Laurel program features are empty.  Those are now checked.

Strata/Languages/Laurel/FilterPrelude.lean
```diff
@@ -0,0 +1,277 @@
...
+    match trigger with | some t => collectExprNames t | none => pure ()
+    collectExprNames body
+  | .Assert cond | .Assume cond => collectExprNames cond
+  | .Return val => match val with | some v => collectExprNames v | none => pure ()
+  | .Old val | .Fresh val | .Assigned val => collectExprNames val
+  | .ProveBy val proof => collectExprNames val; collectExprNames proof
+  | .ContractOf _ func => collectExprNames func
+  | .ReferenceEquals lhs rhs => collectExprNames lhs; collectExprNames rhs
+  | .Hole _ ty => match ty with | some t => collectHighTypeNames t | none => pure ()
+  | _ => pure ()
```
MikaelMayer: A default handler is a dangerous thing if soundness depends on collecting all names. Could you please expand this case exhaustively?

shigoel: +1. We've had issues with this in the past too, so I'm also advocating for using `_` conservatively, even if it comes at the expense of code bloat.

Strata/Languages/Laurel/FilterPrelude.lean
```diff
@@ -0,0 +1,277 @@
...
+  | .UserDefined name => addTypeName name.text
+  | .TCore _ => pure ()
+  | .TTypedField vt => collectHighTypeNames vt
+  | .TSet et => collectHighTypeNames et
+  | .TMap kt vt => collectHighTypeNames kt; collectHighTypeNames vt
+  | .Applied base args =>
+    collectHighTypeNames base; args.forM collectHighTypeNames
+  | .Pure base => collectHighTypeNames base
+  | .Intersection types => types.forM collectHighTypeNames
+  | _ => pure ()
```
shigoel: Same comment here about `_`.

StrataTest/Languages/Python/expected_laurel/test_arithmetic.user_errors.expected
```diff
@@ -0,0 +1,4 @@
+(set-info :file "/local/home/joehx/features/StrataInternal/kiro_tests_annotated/test_bedrock_guardrails.py")
+(set-info :start 352)
+(set-info :stop 376)
+(set-info :error-message "'create_guardrail' called with missing required arguments: [blockedInputMessaging, blockedOutputsMessaging]")
```
olivier-aws: You should remove this file, seems to not correspond to test_arithmetic.py

joehendrix: Yeah, it somehow was added in a merge.

shigoel (approved):

PR #724, opened 2026-03-31
Fix a bug in LExpr's evaluator that generates wrong de bruijn index

aqjune-aws:
This patch fixes a bug in LExpr.evalApp that invokes mkAbsOfArity which generates wrong De Bruijn indices when the number of arguments of Factory op is larger than 2.

`mkAbsOfArity` receives the number of arguments `n` and returns nested abstractions with applications inside, where abs and app alternate. For example, when the input was a Factory op f and first argument x and n = 2 (which means f is a ternary op), `mkAbsOfArity` returns:
```
\. (\. (f x) 0) 1    <--- index 1 doesn't make sense.
```

This bug is not easily observable, unless we have a free variable `x` that has a function type, or there is a partial application of a Factory app. However, it can still happen in the case like attached in LExprEvalTests.lean:

```
-- '"Int.Add3 x y z"' is a ternary op registered to Factory, returning (x + y + z).

...

private def testStateFV : LState TestParams :=
  { testState with state := [[("f", (none, esM[~Int.Add3]))]] }

...

/-
This fails, instead it prints:
LExpr.app () (LExpr.app () (LExpr.app ()
  (LExpr.op () { name := "Int.Add3", metadata := () } none)
  (LExpr.const () Lambda.LConst.intConst 10))
  (LExpr.bvar () 1))                            <---- Dangling De Bruijn index appears
  (LExpr.const () Lambda.LConst.intConst 30)
-/
 #eval Lambda.LExpr.eval 100 test_ternary_fv.σ (esM[((((f : int → int → int → int) #10) #20) #30)])
```

This patch simply removes `mkAbsOfArity`, and invokes `eval n' (.app e1' e2')` whenever there was reduction in the subexpressions of `.app`.

### DIFF
--- a/Tools/BoogieToStrata/Source/StrataGenerator.cs
+++ b/Tools/BoogieToStrata/Source/StrataGenerator.cs
@@ -7,6 +7,13 @@ internal class StrataConversionException(IToken tok, string s) : Exception {
     public string Msg { get; } = $"{tok.filename}({tok.line},{tok.col}): {s}";
 }
 
+internal class LoopRegion(int start, int end, List<string> labels) {
+    public int start = start;
+    public int end = end;
+    public List<string> labels = labels;
+    public List<LoopRegion> children = [];
+}
+
 public class FieldTypeCollector : ReadOnlyVisitor {
     private readonly HashSet<Type> _usedTypes = [];
 
@@ -21,7 +28,17 @@ public class FieldTypeCollector : ReadOnlyVisitor {
 }
 
 public class StrataGenerator : ReadOnlyVisitor {
+    /// <summary>
+    /// Synthetic label representing procedure exit. Used as a goto target for
+    /// return statements and excluded from back-edge detection.
+    /// </summary>
+    private const string ExitLabel = "_exit";
     private readonly Stack<string> _breakLabels = new();
+    /// <summary>
+    /// Set of block labels that are valid exit targets at the current nesting level.
+    /// Updated by EmitWithExitWrappers as wrapper blocks are opened/closed.
+    /// </summary>
+    private readonly HashSet<string> _enclosingLabels = new();
     private readonly VCGenOptions _options;
     private readonly Program _program;
     private readonly Dictionary<Type, HashSet<string>> _uniqueConstants = new();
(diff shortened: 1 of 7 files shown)


Strata/DL/Lambda/LExprEval.lean
```diff
@@ -309,16 +295,13 @@ def evalApp (n' : Nat) (σ : LState TBase) (e e1 e2 : LExpr TBase.mono) : LExpr
...
-    match σ.config.factory.getFactoryLFunc fn.name with
-    | none => LExpr.app m e1' e2'
-    | some lfunc =>
-      let e' := LExpr.app m e1' e2'
-      -- In `e'`, we have already supplied one input to `fn`.
-      -- Note that we can't have 0-arity Factory functions at this point.
-      let e'' := @mkAbsOfArity TBase.mono (lfunc.inputs.length - 1) (e' : LExpr TBase.mono)
-      eval n' σ e''
-  | _ => .app e.metadata e1' e2'
+  | _ =>
```
shigoel: Let's not use wildcards in matches, please.

StrataTest/DL/Lambda/Lambda.lean
```diff
@@ -53,7 +53,7 @@ info: Annotated expression:
 ((~Int.Div : (arrow int (arrow int int))) ((~Int.Add : (arrow int (arrow int int))) #2 #1))
 
 ---
-info: (λ ((~Int.Div : (arrow int (arrow int int))) #3 %0))
+info: ((~Int.Div : (arrow int (arrow int int))) #3)
```
shigoel, line 56: Oof, bug hiding in plain sight here.

shigoel (commented):
Thanks! I'd love for the wildcard to go away, and happy to approve after that.

StrataTest/DL/Lambda/LExprEvalTests.lean
```diff
@@ -658,6 +672,34 @@ example: stuck test25 := by
...
+def test_ternary_fv := TestCase.mk
+  testStateFV
+  esM[((((f : int → int → int → int) #10) #20) #30)]
+  esM[#60]
+
+/-- info: true -/
+#guard_msgs in
+#eval check test_ternary_fv
+
+example: steps_well test_ternary_fv := by
```
joscoh, line 689: This is checking the result via the `Step` relation, not the partial evaluator, right? Did the `Step` semantics have a similar problem?

aqjune-aws: No, `Steps` didn't have the problem. :) The `mkAbsOfArity` function was only used in the partial evaluator.

Strata/DL/Lambda/LExprEval.lean
```diff
@@ -309,16 +295,13 @@ def evalApp (n' : Nat) (σ : LState TBase) (e e1 e2 : LExpr TBase.mono) : LExpr
...
-    match σ.config.factory.getFactoryLFunc fn.name with
-    | none => LExpr.app m e1' e2'
-    | some lfunc =>
-      let e' := LExpr.app m e1' e2'
-      -- In `e'`, we have already supplied one input to `fn`.
-      -- Note that we can't have 0-arity Factory functions at this point.
-      let e'' := @mkAbsOfArity TBase.mono (lfunc.inputs.length - 1) (e' : LExpr TBase.mono)
-      eval n' σ e''
-  | _ => .app e.metadata e1' e2'
+  | _e =>
```
shigoel, line 298: I did say I'd rather we don't use wildcards, but I should've been more explicit given `_e` is not much better. Can we have match against  `.fvar` here? IIUC, no other kind of expression should be seen here?

aqjune-aws: I can list the cases that match to constructors of LExpr here - but I want to understand the motivation behind it.

I think .fvar can appear if the fuel is insufficient or LState has a missing mapping from the variable to a value, but in other normal conditions it wouldn't appear.

Since eval is a smart function that depends on fuel, I think many kind of expressions can appear.

shigoel: I understand June's point (thanks also for offline disucssion).

PR #725, opened 2026-04-01
Add optional tz parameter to datetime_now and reject extra positional args

tautschnig:
### Problem

`datetime.now(timezone.utc)` caused an arity mismatch because the prelude's `datetime_now` declared 0 parameters while the Python call passed 1 argument. This surfaced as a "call [end_time] := datetime_now($hole_12($heap_in))" error in check_storage_costs.

### Approach

Rather than silently dropping extra positional arguments (which could mask real bugs), we model the missing optional parameter in the prelude and add an arity check that rejects extra positional arguments for all functions.

**Prelude changes:**
- Add `tz: Any` parameter to `datetime_now` in the Laurel prelude
- Add `tz: AnyOrNone` parameter to `datetime_now` in the Core prelude and FunctionSignatures

**Arity checking:**
`combinePositionalAndKeywordArgs` now raises a user error when a call passes more positional arguments than the function signature declares, for both user-defined and prelude functions alike

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -93,6 +93,8 @@ structure TranslationContext where
   classFieldHighType: Std.HashMap String (Std.HashMap String HighType) := {}
   /-- Names of prelude types -/
   preludeTypes : Std.HashSet String := {}
+  /-- Prelude procedure signatures (for multi-output detection) -/
+  preludeProcedures : Std.HashMap String CoreProcedureSignature := {}
   /-- Overload dispatch table from PySpec: function name → overloads -/
   overloadTable : OverloadTable := {}
   /-- Behavior for unmodeled functions -/
@@ -960,12 +962,17 @@ Translate Python statements to Laurel StmtExpr nodes.
 These functions are mutually recursive.
 -/
 
+private def hasErrorOutput (sig : CoreProcedureSignature) : Bool :=
+  sig.outputs.length > 0 && sig.outputs.getLast! == "Error"
+
 def withException (ctx : TranslationContext) (funcname: String) : Bool :=
   match ctx.importedSymbols[funcname]? with
   | some (ImportedSymbol.function _) => false
-  | some (ImportedSymbol.procedure _ sig _) =>
-    sig.outputs.length > 0 && sig.outputs.getLast! == "Error"
-  | _ => false
+  | some (ImportedSymbol.procedure _ sig _) => hasErrorOutput sig
+  | _ =>
+    match ctx.preludeProcedures[funcname]? with
+    | some sig => hasErrorOutput sig
+    | none => false
 
 def freeVar (name: String) := mkStmtExprMd (.Identifier name)
 def maybeExceptVar := freeVar "maybe_except"
@@ -2022,6 +2029,7 @@ def pythonToLaurel' (info : PreludeInfo)
     currentClassName := none,
     functionSignatures := info.functionSignatures ++ allClassFuncDecls
     preludeTypes := info.types,
+    preludeProcedures := info.procedures,
     userFunctions := userFunctions,
     classFieldHighType := classFieldHighType,
     overloadTable := overloadTable,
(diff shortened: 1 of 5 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -805,6 +805,9 @@ partial def combinePositionalAndKeywordArgs
       throwUserError callRange
         s!"'{name}' called with unknown keyword arguments: {extraNames}"
     let kwords := pyKwordsToHashMap kwords
+    -- Drop extra positional args beyond the signature (e.g., timezone arg
+    -- in datetime.now(timezone.utc) when the prelude models 0 params).
+    let posArgs := posArgs.take funcDecl.args.length
```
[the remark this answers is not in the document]
tautschnig: Fixed in [88de436](https://github.com/strata-org/Strata/pull/725/commits/88de4367776e484f9578ce7f9fb1d2572412dac4).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -805,6 +806,12 @@ partial def combinePositionalAndKeywordArgs
       throwUserError callRange
         s!"'{name}' called with unknown keyword arguments: {extraNames}"
     let kwords := pyKwordsToHashMap kwords
+    -- For user-defined functions, extra positional args are a real arity error.
+    -- For prelude/external functions, silently drop unmodeled trailing args.
+    if posArgs.length > funcDecl.args.length && funcDecl.name ∈ userFunctions then
+      throwUserError callRange
+        s!"'{name}' called with too many positional arguments: expected at most {funcDecl.args.length}, got {posArgs.length}"
+    let posArgs := posArgs.take funcDecl.args.length
```
olivier-aws: Would be nice to add a test for this error, with the same pattern as the one from #720

tautschnig: Test added.

tautschnig:
> I'm surprised this is sound in general. Another solution would be to modify our Prelude functions to have the optional argument, and add "holes" when one argument is missing. Any reason why we shouldn't use this approach?

That's a great point that we are using a different approach elsewhere. I have now adopted that approach here as well.

StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -130,4 +130,38 @@ open Strata.Parser (stringInputContext)
...
+
+def main() -> None:
+    x: str = greet(\"alice\", \"extra\")
+"
+  try
+    let _ ← processPythonFile pythonCmd (stringInputContext "test.py" program)
+    throw <| IO.userError "Expected pipeline error for too many positional arguments"
+  catch e =>
+    let msg := toString e
+    unless Strata.Python.containsSubstr msg "too many positional arguments" do
```
MikaelMayer: Nit: `containsSubstr` is already opened at line 20, so the fully qualified `Strata.Python.containsSubstr` is redundant — plain `containsSubstr` would suffice (and is what `RelativeImportTest.lean` does).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -806,6 +806,10 @@ partial def combinePositionalAndKeywordArgs
       throwUserError callRange
         s!"'{name}' called with unknown keyword arguments: {extraNames}"
     let kwords := pyKwordsToHashMap kwords
+    -- Extra positional args beyond the signature are an arity error.
+    if posArgs.length > funcDecl.args.length then
+      throwUserError callRange
```
MikaelMayer, line 844: Worth noting (no action needed): when `isVarKwargs` is true (lines ~898-907 above), the call takes a separate path that bypasses `combinePositionalAndKeywordArgs` entirely, so extra positional args wouldn't be caught in that case. That's pre-existing behavior and out of scope for this PR, but might be worth a follow-up.

StrataTest/Languages/Python/VerifyPythonTest.lean
tautschnig: Done.

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Actually amending that pre-existing behaviour just required an extra four lines (plus a test), so that's done as well.

PR #726, opened 2026-04-01
Fix multi-output procedure handling: Unknown fallback and prelude withException detection

tautschnig:
### Problem

`timedelta(days=7)` caused a Core type checking error "Impossible to unify bool with Any" because:

1. `computeExprType` returned `TVoid` for multi-output procedures (e.g., `timedelta_func` returning `(delta: Any, maybe_except: Error)`). `TVoid` translates to `bool` in Core, causing the confusing error.

2. `withException` in PythonToLaurel only checked `importedSymbols` for multi-output procedures, missing prelude procedures like `timedelta_func`. This caused the translator to generate a single-target assignment instead of the required 2-target assignment `call [delta, maybe_except] := timedelta_func(...)`.

### Fix

**computeExprType (LaurelTypes.lean):** Return `Unknown` instead of `TVoid` for zero-output and multi-output procedures. `Unknown` translates to `Any` in Core, which avoids the type mismatch. Single-output procedures continue to return their declared output type.

**withException (PythonToLaurel.lean):** Add `preludeProcedures` field to `TranslationContext` and extend `withException` to check prelude procedure signatures. This ensures multi-output prelude procedures
like `timedelta_func` correctly generate 2-target assignments.


Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -46,8 +46,8 @@ def computeExprType (model : SemanticModel) (expr : StmtExprMd) : HighTypeMd :=
     | .datatypeConstructor t _ => ⟨ .UserDefined t, md, ⟩
     | .parameter p => p.type
     | .staticProcedure proc => match proc.outputs with
-      | [singleOutput] => singleOutput.type
-      | _ => { val := .TVoid, md := default }
+      | firstOutput :: _ => firstOutput.type
+      | [] => { val := .TVoid, md := default }
```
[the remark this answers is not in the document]
tautschnig: Comment updated.

Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -41,13 +41,13 @@ def computeExprType (model : SemanticModel) (expr : StmtExprMd) : HighTypeMd :=
...
   | .PureFieldUpdate target _ _ => computeExprType model target
-  -- Calls — we don't track return types here, so fall back to TVoid
+  -- Calls — return the declared output type when available, fall back to TVoid otherwise
   | .StaticCall callee _ => match model.get callee with
     | .datatypeConstructor t _ => ⟨ .UserDefined t, md, ⟩
     | .parameter p => p.type
     | .staticProcedure proc => match proc.outputs with
-      | [singleOutput] => singleOutput.type
-      | _ => { val := .TVoid, md := default }
+      | firstOutput :: _ => firstOutput.type
```
keyboardDrummer: Why would the first output be special? That seems like a bug to me. Laurel doesn't have a concept of multi-valued expressions, so a call to a multi-output value procedure is only valid it if is immediately assigned where the LHS has the right number of argument.

What I think is the heart of the problem here is that PythonToLaurel incorrectly translates the call to the multi-arg procedure, by leaving out the required assignment. Laurel then fails by not reporting a resolution error specifically about this issue, and by using `.TVoid` as a fallback type for the broken call, while `.Unknown` should be used as a fallback, leading to a cryptic Core error.

keyboardDrummer (changes requested):

Strata/Languages/Laurel/LaurelTypes.lean
tautschnig: Thank you for your insights, I have completely reworking the implementation based on your feedback.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -946,7 +948,11 @@ def withException (ctx : TranslationContext) (funcname: String) : Bool :=
   | some (ImportedSymbol.function _) => false
   | some (ImportedSymbol.procedure _ sig _) =>
     sig.outputs.length > 0 && sig.outputs.getLast! == "Error"
-  | _ => false
+  | _ =>
+    -- Also check prelude procedures for multi-output with Error
+    match ctx.preludeProcedures[funcname]? with
+    | some sig => sig.outputs.length > 0 && sig.outputs.getLast! == "Error"
```
MikaelMayer: Nit: the expression `sig.outputs.length > 0 && sig.outputs.getLast! == "Error"` is duplicated from line 950. Consider extracting a small helper like:
```lean
private def hasErrorOutput (sig : CoreProcedureSignature) : Bool :=
  sig.outputs.length > 0 && sig.outputs.getLast! == "Error"
```
This would make `withException` a bit easier to read and avoids the duplication.

StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -130,4 +130,21 @@ open Strata.Parser (stringInputContext)
   if diags.size ≠ 0 then
     throw <| .userError s!"Expected 0 diagnostics, got {diags.size}"
 
+-- Multi-output procedures (e.g., timedelta_func) return Unknown type
```
MikaelMayer: I don't see a statement with multi-outputs being tested. Is this PR supposed to cover that too?

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Done.

StrataTest/Languages/Python/VerifyPythonTest.lean
tautschnig: That actually is a multi-output prelude procedure, but I've added a comment clarifying that.

PR #727, opened 2026-04-01
Emit Hole for Composite return values assigned to Any

tautschnig:
### Problem

When a Python function returns a Composite-typed value (e.g., a service client) but the Laurel return variable `LaurelResult` has type `Any`, Core type checking fails with "Impossible to unify Any with Composite".

### Approach

Per design discussion with @keyboardDrummer and @joehendrix: emit a Hole (unconstrained value) when a Composite-typed expression is returned from a function with Any return type. This avoids the type unification
crash while being honest about the limitation — the verifier treats the return value as unconstrained, which limits bug-finding but doesn't produce unsound results.

Auto-generated Composite -> Any coercions can be revisited once the broader Composite/Any typing design is resolved.

**Changes:**
- Add `coerceToAny` helper that replaces Composite-typed expressions with a Hole when assigned to an Any-typed context
- Apply coercion in return statement translation before assigning to `LaurelResult`
- Add regression tests in `VerifyPythonTest.lean` and CBMC pipeline


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -717,6 +717,22 @@ If the function_name is a class, add __init__ into it
...
+  if isCompositeType ctx ty then
+    let fields := (ctx.classFieldHighType[ty]?.getD {}).toList
+    let dict ← fields.foldlM (fun acc (fname, fty) =>
+      return mkStmtExprMd (.StaticCall "DictStrAny_cons"
+        [mkStmtExprMd (.LiteralString fname),
+         ← wrapFieldInAny fty (mkStmtExprMd (.FieldSelect translated fname)), acc]))
+      (mkStmtExprMd (.StaticCall "DictStrAny_empty" []))
+    pure <| mkStmtExprMd (.StaticCall "from_ClassInstance"
+      [mkStmtExprMd (.LiteralString ty), dict])
+  else pure translated
```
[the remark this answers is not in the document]
tautschnig: Added TODO.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -717,6 +717,22 @@ If the function_name is a class, add __init__ into it
...
+    Composite values are wrapped via from_ClassInstance with their fields. -/
+partial def coerceToAny (ctx : TranslationContext) (expr : Python.expr SourceRange)
+    (translated : StmtExprMd) : Except TranslationError StmtExprMd := do
+  let ty ← inferExprType ctx expr
+  if isCompositeType ctx ty then
+    let fields := (ctx.classFieldHighType[ty]?.getD {}).toList
+    let dict ← fields.foldlM (fun acc (fname, fty) =>
+      return mkStmtExprMd (.StaticCall "DictStrAny_cons"
+        [mkStmtExprMd (.LiteralString fname),
+         ← wrapFieldInAny fty (mkStmtExprMd (.FieldSelect translated fname)), acc]))
```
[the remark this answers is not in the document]
tautschnig: Added TODO.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1184,6 +1204,8 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     let stmts ← match value.val with
       | some expr => do
         let e ← translateExpr ctx expr
+        -- Coerce Composite return values to Any for LaurelResult : Any
```
keyboardDrummer, line 1221: What about when passing an expression of type composite to a function that takes an Any, do you then also need to coerse? The overal design of how Any and composite types, after having been translated from Python to Laurel, interact, is still unclear to me.

I think what we discussed on Slack is that we want to have at some point, is for every Python expression, when translated to Laurel, to be of type `Any`. That way, there would be no coercions between Any and composite types at call-sites. @joehendrix commented on this idea with:

> The main issue with Any everywhere is method resolution.  For example, we want self.client.service_call(..) to resolve the class of client and which service call is being inlined.

I think this means that at the Python level, we want to know types from specific than Any, but once we get to Laurel each expression can still be an Any. @joehendrix does that make sense to you?

If the answer is yes, then I think we can update `from_ClassInstance (classname : string, instance_attributes: DictStrAny)` to `from_ClassInstance (classname : string, instance_attributes: MutableDictStrAny)` with

```
composite MutableDictStrAny {
  var value: DictStrAny
}

// field reads translate to this
function readAny(a: Any, field: String): Any 
  requires Any..isfrom_ClassInstance(a)
{
  readDictStrAny(Any..as_ClassData(a)#value, field)
}

// field writes translate to this
procedure updateAny(a: Any, field: String, value: Any)
  requires Any..isfrom_ClassInstance(a)
  modifies Any..as_ClassData(a)
  ensures Any..as_ClassData(a)#value == updateDicStrAny(old(Any..as_ClassData(a)), field, value)
```

and update the translation of Python's field read and assign expressions to use the above function/procedure.

tautschnig: It seems consensus is the following (proposal by @joehendrix):
> My inclination is that, for now, we emit a hole (and a warning) when needing to translate between the composite and Any, and just accept that limits our ability to find bugs.

> We could look into auto-generating coercisions once the typing issues have been resolved (and thus can compare tradeoffs a little more precisely).  Coercisions could make more code analyzable but result in confusing messages (e.g., when trying to figure out why a composite is Any or needing to coerce back.

Will implement that.

tautschnig: Done.

StrataTest/Languages/Python/tests/test_composite_return.py
```diff
@@ -0,0 +1,9 @@
+class MyService:
+    name: str
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+def create_service() -> Any:
```
joehendrix, line 9: `Any` is used here without importing it. The inline Lean test includes `from typing import Any` — this file should too for consistency and to ensure the CBMC pipeline resolves the return type correctly.

```suggestion
from typing import Any

class MyService:
    name: str

    def __init__(self, name: str) -> None:
        self.name = name

def create_service() -> Any:
    svc: MyService = MyService("test")
    return svc
```

StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -130,4 +130,23 @@ open Strata.Parser (stringInputContext)
...
+    name: str
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+def create_service() -> Any:
+    svc: MyService = MyService(\"test\")
+    return svc
+"
+  let _ ← processPythonFile pythonCmd (stringInputContext "test.py" program)
```
joehendrix: This discards the diagnostics without checking. The other tests in this file explicitly assert on `diags.size` — doing the same here would confirm the translation produces no unexpected diagnostics, not just that it does not crash.

```suggestion
  let diags \u2190 processPythonFile pythonCmd (stringInputContext "test.py" program)
  if diags.size \u2260 0 then
    throw <| .userError s!"Expected 0 diagnostics, got {diags.size}"
```

StrataTest/Languages/Python/tests/cbmc_expected.txt
```diff
@@ -31,3 +31,4 @@ test_class_with_methods.py.ion  SKIP
 test_module_level.py.ion        SKIP
 test_if_elif.py.ion             SKIP
 test_variable_reassign.py.ion   SKIP
+test_composite_return.py.ion     PASS
```
joehendrix: Nit: extra space before PASS — the other entries align the status column one character earlier.

```suggestion
test_composite_return.py.ion    PASS
```

tautschnig: Done.

StrataTest/Languages/Python/VerifyPythonTest.lean
tautschnig: Thanks, done.

StrataTest/Languages/Python/tests/test_composite_return.py
tautschnig: Thanks, done.

PR #728, opened 2026-04-01
Improve #strata_gen error messages

### LINKED ISSUE #136: Improve #strata_gen error messages
If #strata_gen tried to introduce a datatype, it doesn't check whether the name is already defined.  We should add a check and better error messages for it.

MikaelMayer:
`#strata_gen` did not check whether the datatype names it was about to introduce already existed in the Lean environment, leading to confusing errors from the elaborator.

This adds a pre-generation check that resolves each category name (respecting the current namespace and private name mangling) and verifies it is not already defined. When a conflict is found, a clear error message is reported and code generation is aborted.

The name-resolution logic is factored into a shared `resolveScopedName` helper used by both `mkScopedIdent` and the new availability check.

Tested: existing tests pass, new test added for the duplicate name scenario.

Fixes #136


Strata/DDM/Integration/Lean/Gen.lean
```diff
@@ -1545,6 +1545,29 @@ partial def generateInhabitedInstances (group : Array (QualifiedIdent × Array D
...
+  let env ← getEnv
+  let scope ← getScope
+  let mut hasConflict := false
+  for (cat, _) in categories do
+    if cat ∈ declaredCategories then continue
+    let catName ← getCategoryScopedName cat
+    let fullName := scope.currNamespace ++ catName
+    let resolvedName :=
+      if !env.header.isModule || scope.isPublic then fullName
+      else Lean.mkPrivateName env fullName
```
MikaelMayer: The name-resolution logic (namespace ++ subName, then conditionally `mkPrivateName`) is duplicated from `mkScopedIdent` (lines 54–59). Consider extracting a small helper like `resolveScopedName : Name → CommandElabM Name` that both call sites can share, so future changes to the resolution strategy only need to be made in one place.

MikaelMayer: Extracted `resolveScopedName : Name → CommandElabM Name` and updated both `mkScopedIdent` and `checkCategoryNamesAvailable` to use it.

PR #729, opened 2026-04-01
Track pyspec procedure names for targeted inlining

MikaelMayer:
## Track pyspec procedure names for targeted inlining

Previously, pyspec procedure inlining was gated on whether `--pyspec` flags were passed on the CLI, and when triggered, it inlined all non-main, non-prelude procedures indiscriminately. Tests also had to manually construct the inlining predicate.

This PR tracks which procedure names originate from pyspec files through the pipeline and provides `inlinePySpecProcedures` so callers don't need to know the inlining details.

### Changes

- `PySpecLaurelResult` gains a `pyspecProcedureNames` field, populated during `buildPySpecLaurel`
- `pyAnalyzeLaurel` returns a `PyAnalyzeResult` struct containing both the Laurel program and the pyspec procedure names
- `inlinePySpecProcedures` encapsulates the inlining logic — callers just pass the Core program and the pyspec names
- Inlining is now automatic (no CLI flag gate) and precise (only pyspec procedures)

### Testing

Existing `AnalyzeLaurelTest` tests pass — these exercise the full pyspec dispatch + inlining + verification pipeline.

### DIFF
--- a/Strata/DL/Imperative/SMTUtils.lean
+++ b/Strata/DL/Imperative/SMTUtils.lean
@@ -22,10 +22,13 @@ public section
 ---------------------------------------------------------------------
 
 /--
-A counterexample derived from an SMT solver is a map from an identifier
+A model derived from an SMT solver is a map from an identifier
 to an `SMT.Term`.
 -/
-@[expose] abbrev CounterEx (Ident : Type) := Map Ident Strata.SMT.Term
+@[expose] abbrev Model (Ident : Type) := Map Ident Strata.SMT.Term
+
+/-- Backward-compatible alias. -/
+@[deprecated Model (since := "2026-04-03")] abbrev CounterEx := @Model
 
 /-- Render an `SMT.Term` to a string via the SMTDDM translation. -/
 private def termToString (t : Strata.SMT.Term) : String :=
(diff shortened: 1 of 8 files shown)


StrataMain.lean
```diff
@@ -588,6 +588,8 @@ def pyAnalyzeLaurelCommand : Command where
         exitPyAnalyzeKnownLimitation msg
       | .error (.internal msg) =>
         exitPyAnalyzeInternalError msg
+    let combinedLaurel := analyzeResult.laurelProgram
+    let pyspecProcNames := analyzeResult.pyspecProcedureNames
```
MikaelMayer, line 577: `preludeNames` (line 621) is now unused — the old inlining predicate that referenced it was removed in favour of `pyspecProcNames`. Should be `_preludeNames` like in the test file to silence any unused-variable warning.

Strata/SimpleAPI.lean
```diff
@@ -387,11 +387,11 @@ def pyTranslateLaurel
...
-  let laurel ←
+  let result ←
     match ← pyAnalyzeLaurel pythonIonPath dispatchModules pyspecModules (specDir := specDir) |>.toBaseIO with
     | .ok r => pure r
     | .error err => throw (toString err)
-  let (coreOption, laurelTranslateErrors) := translateCombinedLaurel laurel
+  let (coreOption, laurelTranslateErrors) := translateCombinedLaurel result.laurelProgram
   match coreOption with
   | none => throw s!"Laurel to Core translation failed: {laurelTranslateErrors}"
   | some core => pure (core, laurelTranslateErrors)
```
MikaelMayer: `pyTranslateLaurel` now receives a `PyAnalyzeResult` but discards `pyspecProcedureNames`, so callers like `pyTranslateLaurelCommand` and `pyAnalyzeLaurelToGotoCommand` won't get pyspec inlining. If inlining is meant to be automatic for all pyspec pipelines, this function should either call `inlinePySpecProcedures` on the Core result, or return the pyspec names so callers can do it themselves.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -159,7 +162,8 @@ public def buildPySpecLaurel (pyspecEntries : Array (String × String))
   }
   return { laurelProgram := combinedLaurel, overloads := allOverloads
            functionSignatures := funcSigs, typeAliases := allTypeAliases
-           exhaustiveClasses := allExhaustiveClasses }
+           exhaustiveClasses := allExhaustiveClasses
+           pyspecProcedureNames := seenProcs.fold (init := {}) fun s name _ => s.insert name }
```
MikaelMayer, line 168: Minor: `seenProcs` is a `HashMap String String` where the value (source file) is only used for collision error messages. The fold here extracts just the keys into a `HashSet`. Consider adding a brief comment (e.g. `-- extract procedure names as a set`) so the intent is clear at a glance.

StrataMain.lean
MikaelMayer: Fixed — renamed to `_preludeNames`.

Strata/SimpleAPI.lean
MikaelMayer: Fixed — `pyTranslateLaurel` now calls `inlinePySpecProcedures` on the Core result.

Strata/Languages/Python/PySpecPipeline.lean
MikaelMayer: Added comment: `-- Extract procedure names as a set for targeted inlining`

PR #731, opened 2026-04-01
fix: prevent infinite loop when modifies clause references non-composite type (#490)

### LINKED ISSUE #490: laurelAnalyze: infinite loop when procedure has modifies clause referencing a global variable
## Summary

`strata laurelAnalyze` enters an infinite loop (spewing thousands of PANIC messages and consuming unbounded CPU/memory) when a Laurel program contains a procedure with a `modifies` clause that references a global variable.

## Reproducer

Save as `repro.lr.st`:

```laurel
var g: int;

procedure inc(x: int) returns (r: int)
  modifies g
{
  r := x + 1;
  g := r;
}

procedure main() {
  var a: int := 42;
  assert a > 0;
}
```

Run either:

```bash
strata laurelAnalyze repro.lr.st
```

## Observed behavior

The process enters an infinite loop, emitting hundreds of PANIC messages per second:

```
PANIC at Strata.DialectMap.decl! Strata.DDM.AST:1672:12: Unknown dialect
PANIC at Strata.Elab.elabOperation Strata.DDM.Elab.Core:1013:11: Missing position info ...
Unknown operation .
```

In a 3-second CPU time window, ~840 PANIC messages were emitted. Without a CPU time limit, the process consumed 14+ GB of memory before being killed.

## Expected behavior

The command should either:
- Successfully translate the program, or
- Fail with a clear error message

## Trigger conditions

The issue requires **both**:
1. A global variable declaration (`var g: int`)
2. A `modifies` clause referencing that global (`modifies g`)

The following variations do **not** trigger the bug:
- Procedure with `requires`/`ensures` but no `modifies` (works fine)
- Procedure with `if/else` but no `modifies` on a global (works fine)
- Procedure with `modifies` on an output parameter only (different error, but no infinite loop)

Adding `if/else` or `ensures` to the reproducer does not change the behavior — the infinite loop is triggered by `modifies` + global variable alone.

## Environment

- macOS (arm64)
- Branch: `tautschnig/ToCProverGOTO-Stmt` (also likely affects `main`)

## Workaround

Use `ulimit -t <seconds>` to bound CPU time when invoking `strata`:

```bash
bash -c 'ulimit -t 20; strata laurelAnalyze repro.lr.st'
```

MikaelMayer:
## Problem

`strata laurelAnalyze` enters an infinite loop (emitting PANIC messages and consuming unbounded memory) when a Laurel program contains a procedure with a `modifies` clause referencing a non-composite type, such as a parameter of type `int`.

The root cause is that the heap parameterization pass treats any procedure with a non-empty `modifies` list as heap-modifying, regardless of the types in the list. When the modifies list contains a non-composite type, the modifies clause transformation generates invalid frame conditions, leading to downstream failures.

## Fix

1. **Single source of truth**: `ModifiesTypeKind` enum and `classifyModifiesHighType` in `LaurelTypes.lean` serve as the canonical type classification for modifies clauses. Both `isHeapRelevantType` and `classifyModifiesType` delegate to it.

2. **Separation of concerns**: Non-composite modifies filtering and diagnostics live in `ModifiesClauses.lean` (`filterNonCompositeModifies`), keeping the heap parameterization phase agnostic to modifies clauses. The filtering runs as a pre-pass before heap parameterization in the pipeline.

3. **Error reporting**: Non-composite modifies entries produce a diagnostic error (e.g. `modifies clause entry has non-composite type 'int' and will be ignored`) instead of being silently filtered.

## Testing

All existing tests pass. Added a regression test (`T8_NonCompositeModifies`) that verifies procedures with non-composite modifies entries terminate and produce the expected diagnostic errors.

Fixes #490


Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -470,6 +470,22 @@ def heapParameterization (model: SemanticModel) (program : Program) : Program :=
...
+    | _ => false
+  let filterBodyModifies (body : Body) : Body :=
+    match body with
+    | .Opaque posts impl mods =>
+      .Opaque posts impl (mods.filter (isCompositeModifies model))
+    | other => other
+  let filterProcModifies (proc : Procedure) : Procedure :=
+    { proc with body := filterBodyModifies proc.body }
+  let program := { program with
+    staticProcedures := program.staticProcedures.map filterProcModifies }
```
MikaelMayer: The filtering only applies to `staticProcedures`, but `instanceProcedures` (collected from `program.types` at line 491) are passed unfiltered to `computeReadsHeap`/`computeWritesHeap`. If an instance procedure had a non-composite modifies entry, it would still incorrectly mark it as heap-reading/writing.

This may not be a practical issue today (instance procedures might not have non-composite modifies clauses in practice), but it would be more robust to also filter instance procedures — or at least add a comment explaining why it's not needed.

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -470,6 +470,22 @@ def heapParameterization (model: SemanticModel) (program : Program) : Program :=
   let program := { program with
     types := program.types
     staticProcedures := program.staticProcedures }
+  -- Filter out non-composite modifies entries so that global variables
+  -- of primitive type do not incorrectly trigger heap parameterization.
```
MikaelMayer: The type-check logic here (`UserDefined => true | TSet => true | _ => false`) duplicates the logic in `extractModifiesEntries` in `ModifiesClauses.lean`. Consider extracting a shared helper like `isHeapRelevantType` to keep the two passes in sync. If one is updated to handle a new type, the other could easily be missed.

Strata/Languages/Laurel/HeapParameterization.lean
MikaelMayer: Done — instance procedures are now filtered through the same `filterProcModifies` before being passed to `computeReadsHeap`/`computeWritesHeap`.

Strata/Languages/Laurel/HeapParameterization.lean
MikaelMayer: Done — extracted `isHeapRelevantType` in `LaurelTypes.lean` (where `computeExprType` lives, imported by both files). Both `HeapParameterization.lean` and `ModifiesClauses.lean` now use it.

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -470,10 +470,22 @@ def heapParameterization (model: SemanticModel) (program : Program) : Program :=
   let program := { program with
     types := program.types
     staticProcedures := program.staticProcedures }
-  -- Collect instance procedures from composite types for heap analysis
+  -- Filter out non-composite modifies entries so that global variables
```
keyboardDrummer: > -- Filter out non-composite modifies entries

Shouldn't those produce errors instead of silently being filtered out?

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -50,13 +50,18 @@ inductive ModifiesEntry where
...
-    match (computeExprType model expr).val with
-    | .TSet _ => .set expr
-    | _ => .single expr
+  modifiesExprs.filterMap fun expr =>
+    let ty := (computeExprType model expr).val
+    if !isHeapRelevantType ty then none
+    else match ty with
+    | .TSet _ => some (.set expr)
+    | .UserDefined _ => some (.single expr)
+    | _ => none
```
tautschnig: Hmm, this still weirdly duplicates what isHeapRelevantType does?

Strata/Languages/Laurel/HeapParameterization.lean
MikaelMayer: Done — non-composite modifies entries now produce a diagnostic error (e.g. `modifies clause entry has non-composite type 'int' and will be ignored`) instead of being silently filtered. The error is emitted during heap parameterization and propagated through the diagnostic pipeline. Added a regression test in `T8_NonCompositeModifies.lean` that verifies the errors are reported.

Strata/Languages/Laurel/ModifiesClauses.lean
MikaelMayer: Agreed — refactored to eliminate the duplication. Introduced `classifyModifiesType` which is the single source of truth for both filtering and classifying modifies entries. `extractModifiesEntries` now delegates to it, and `isHeapRelevantType` is only used in `HeapParameterization` for the filtering/diagnostic pass.

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -47,16 +47,26 @@ inductive ModifiesEntry where
   | single (expr : StmtExprMd)       -- a single Composite reference
   | set (expr : StmtExprMd)          -- a Set Composite expression
 
+/--
+Classify a heap-relevant type into a `ModifiesEntry`, or `none` for
+non-heap-relevant types. This is the single source of truth for which types
+produce modifies entries and how they are classified.
+-/
```
tautschnig, line 54: Hmm, but it's still to be the same as isHeapRelevantType?

MikaelMayer:
Re: [duplication in `classifyModifiesType`](https://github.com/strata-org/Strata/pull/731#discussion_r3029494986)

You're right — the duplication was still there. Refactored: introduced `ModifiesTypeKind` enum and `classifyModifiesHighType` in `LaurelTypes.lean` as the single source of truth. `isHeapRelevantType` is now derived from it (`(classifyModifiesHighType ty).isSome`), and `classifyModifiesType` in `ModifiesClauses.lean` delegates to it for the type classification, only adding the expr-wrapping logic on top.

Strata/Languages/Laurel/ModifiesClauses.lean
MikaelMayer: Addressed in a620e1cf — introduced `ModifiesTypeKind` enum and `classifyModifiesHighType` in `LaurelTypes.lean` as the single source of truth. `isHeapRelevantType` is now `(classifyModifiesHighType ty).isSome`, and `classifyModifiesType` in `ModifiesClauses.lean` delegates to `classifyModifiesHighType` for the type classification, only adding the expr-wrapping into `ModifiesEntry`. No duplication remains.

MikaelMayer:
The regression test `T8_NonCompositeModifies.lean` (added in 90f2a3d) covers this. It runs the full `laurelAnalyze` pipeline (`processLaurelFile` → `verifyToDiagnostics` → `verifyToVcResults`) on procedures with non-composite modifies entries. Before the fix, this would have caused an infinite loop; now it terminates and produces the expected diagnostic errors. If the bug were reintroduced, this test would hang and the build would time out.

Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -102,6 +102,25 @@ def computeExprType (model : SemanticModel) (expr : StmtExprMd) : HighTypeMd :=
...
+in modifies clauses and heap parameterization. -/
+def classifyModifiesHighType : HighType → Option ModifiesTypeKind
+  | .UserDefined _ => some .composite
+  | .TSet _        => some .compositeSet
+  | _              => none
+
+/-- Returns `true` when the given `HighType` is heap-relevant (composite or set
+of composite), i.e. the kind of type that appears in modifies clauses and
+triggers heap parameterization. -/
+def isHeapRelevantType (ty : HighType) : Bool :=
```
keyboardDrummer, line 121: Seems like this does not protect against sets of non-heap types, but that can be another PR

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -466,14 +466,43 @@ def heapTransformProcedure (model: SemanticModel) (proc : Procedure) : Transform
...
+  -- parameterization.
+  let collectBodyDiags (body : Body) : List DiagnosticModel :=
+    match body with
+    | .Opaque _ _ mods => mods.filterMap fun e =>
+        let ty := (computeExprType model e).val
+        if isHeapRelevantType ty then none
+        else some (e.md.toDiagnostic s!"modifies clause entry has non-composite type '{formatHighTypeVal ty}' and will be ignored")
+    | _ => []
+  let collectProcDiags (proc : Procedure) : List DiagnosticModel :=
+    collectBodyDiags proc.body
```
keyboardDrummer: Inline `collectBodyDiags` ?

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -466,14 +466,43 @@ def heapTransformProcedure (model: SemanticModel) (proc : Procedure) : Transform
...
-def heapParameterization (model: SemanticModel) (program : Program) : Program :=
+def heapParameterization (model: SemanticModel) (program : Program) : Program × List DiagnosticModel :=
   let program := { program with
     types := program.types
     staticProcedures := program.staticProcedures }
-  -- Collect instance procedures from composite types for heap analysis
+  -- Collect diagnostics for non-composite modifies entries and filter them out
+  -- so that global variables of primitive type do not incorrectly trigger heap
+  -- parameterization.
+  let collectBodyDiags (body : Body) : List DiagnosticModel :=
```
keyboardDrummer: Can you put this in a where clause and combine it with filterBodyModifies? Might be nice to collect the diagnostics with a monad.

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -466,14 +466,43 @@ def heapTransformProcedure (model: SemanticModel) (proc : Procedure) : Transform
     -- This procedure doesn't read or write the heap - no changes needed
     return proc
 
-def heapParameterization (model: SemanticModel) (program : Program) : Program :=
+def heapParameterization (model: SemanticModel) (program : Program) : Program × List DiagnosticModel :=
   let program := { program with
     types := program.types
     staticProcedures := program.staticProcedures }
-  -- Collect instance procedures from composite types for heap analysis
+  -- Collect diagnostics for non-composite modifies entries and filter them out
```
keyboardDrummer: Can we leave the HeapParameterization phase agnostic to modifies clauses, and move this code into the modifies clauses phase?

keyboardDrummer (dismissed):

Strata/Languages/Laurel/LaurelTypes.lean
MikaelMayer: Good catch — you're right that `TSet` of a non-heap type (e.g. `Set[Int]`) would still be classified as `compositeSet`. We could recurse into the element type here. Happy to address in a follow-up PR as you suggest.

Strata/Languages/Laurel/HeapParameterization.lean
MikaelMayer: Done. Moved the non-composite modifies filtering and diagnostics into a new `filterNonCompositeModifies` function in `ModifiesClauses.lean`. It runs as a separate pre-pass before heap parameterization in the pipeline. `heapParameterization` no longer knows about modifies clauses and now returns just a `Program` instead of `Program × List DiagnosticModel`.

Strata/Languages/Laurel/HeapParameterization.lean
MikaelMayer: Addressed as part of the move to `ModifiesClauses.lean`. The new `filterBodyNonCompositeModifies` function combines filtering and diagnostics collection in a single pass over the modifies entries, returning `Body × List DiagnosticModel`.

Strata/Languages/Laurel/HeapParameterization.lean
MikaelMayer: Moot — the entire modifies filtering/diagnostics code (including `collectBodyDiags`) has been moved out of `HeapParameterization.lean` into `ModifiesClauses.lean` per the other comment. The new `filterBodyNonCompositeModifies` is used directly without a wrapper.

PR #732, opened 2026-04-01
Fix doc generation for SSH git clones

### LINKED ISSUE #427: Attempting to generate docs in a git@github.com clone fails
Running `./generate.sh` in `docs/verso/` I got:
```
[...]
✖ [321/554] Running Strata:srcUri.github
error: Could not interpret Git remote uri git@github.com:strata-org/Strata as a Github source repo.
See README on source URIs for more details.
[...]
Some required targets logged failures:
- Strata:srcUri.github
- Strata:srcUri.github
- Strata:srcUri.github
- Strata:srcUri.github
- Strata:srcUri.github
- Strata:srcUri.github
error: build failed
```

I don't even know whether this is a Strata or a Lean problem, but it would be great if documentation could also be generated from such a clone, and not just `https://github.com` ones.

MikaelMayer:
Fixes #427

## Problem

Running `./generate.sh` in `docs/verso/` fails when the repo is cloned via SSH (`git@github.com:strata-org/Strata`) because doc-gen4 only recognizes HTTPS GitHub URLs when constructing source links.

## Solution

The `generate.sh` script now detects SSH-format git remotes and temporarily converts them to HTTPS format before running the doc build. The original remote URL is restored on exit via a shell trap, even if the build fails.

## Testing

The change is a shell script only — the Lean build and existing tests are unaffected.

### DIFF
--- a/Strata/Languages/Laurel/FilterPrelude.lean
+++ b/Strata/Languages/Laurel/FilterPrelude.lean
@@ -1,0 +1,287 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+public import Strata.Languages.Laurel.Laurel
+
+/-! ### Prelude Filtering
+
+Restrict the Laurel prelude to only the `staticProcedures` and `types`
+transitively needed by the user program, reducing the Core program size
+for SMT verification.
+
+#### Name collection
+
+We collect two categories of names from the user program:
+
+- **procNames**: targets of `StaticCall` and `InstanceCall`. In practice
+  this includes true procedure calls (e.g. `print`, `Any_get`) as well as
+  datatype constructor calls (`NoError`, `from_string`), testers (`isError`),
+  and destructors (`Any..as_string!`), since the Laurel IR uses `StaticCall`
+  as the uniform calling convention for all of these.
+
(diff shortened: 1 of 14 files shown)


docs/verso/generate.sh
```diff
@@ -1,5 +1,14 @@
 set -ex
 
+# Normalize SSH git remote to HTTPS so doc-gen4 can parse it (see #427).
+repo_root=$(git -C "$(dirname "$0")/../.." rev-parse --show-toplevel)
+orig_url=$(git -C "$repo_root" remote get-url origin 2>/dev/null || true)
+if echo "$orig_url" | grep -q '^git@github\.com:'; then
+  https_url=$(echo "$orig_url" | sed 's|^git@github\.com:|https://github.com/|; s|\.git$||')
+  git -C "$repo_root" remote set-url origin "$https_url"
+  trap 'git -C "$repo_root" remote set-url origin "$orig_url"' EXIT
```
MikaelMayer, line 9: 👍 Setting the trap only when the URL is actually changed is a nice touch — avoids a no-op restore on HTTPS clones and makes the intent clear.

MikaelMayer: Thanks! Exactly the intent — no reason to set up a restore trap if we didn't touch the URL.

PR #733, opened 2026-04-01
Migrate more StrataMain commands to SimpleAPI

shigoel:
## Summary

- Add `readLaurelTextFile` to SimpleAPI, replacing duplicated 8-line
  Laurel parsing boilerplate in 4 commands (laurelParse, laurelAnalyze,
  laurelAnalyzeToGoto, laurelToCore)
- Add `readLaurelIonFiles` / `readLaurelIonProgram` to SimpleAPI,
  replacing the local `deserializeIonToLaurelFiles` helper
- Add `readPythonIon` / `pythonDirectToCore` to SimpleAPI, replacing
  the local `readPythonStrata` helper used by 3 commands
- Add `Laurel.verifyProgram` / `Laurel.analyzeToDiagnosticModels` to
  SimpleAPI for Laurel analysis
- Route `pyAnalyzeCommand` through `Core.verifyProgram` instead of
  manually managing temp directories
- Remove 7 imports from StrataMain now provided via SimpleAPI

### DIFF
--- a/Strata/SimpleAPI.lean
+++ b/Strata/SimpleAPI.lean
@@ -121,9 +121,8 @@ Read a Laurel source file in textual format and parse it into
 a `Laurel.Program`. Handles dialect loading, parsing, and
 AST translation in one step.
 -/
-def readLaurelTextFile (path : System.FilePath)
+def parseLaurelText (path : System.FilePath) (content : String)
     : IO Laurel.Program := do
-  let content ← IO.FS.readFile path
   let input := Strata.Parser.stringInputContext path content
   let dialects :=
     Strata.Elab.LoadedDialects.ofDialects!
@@ -137,6 +136,11 @@ def readLaurelTextFile (path : System.FilePath)
   | .ok program => pure program
   | .error errors =>
     throw (IO.userError s!"Laurel translation errors: {errors}")
+
+def readLaurelTextFile (path : System.FilePath)
+    : IO Laurel.Program := do
+  let content ← IO.FS.readFile path
+  parseLaurelText path content
 
 /--
 Deserialize Laurel Ion bytes (possibly containing multiple files)
@@ -274,7 +278,7 @@ def Core.callElimUsingContract (p : Core.Program) : Except String Core.Program :
 /-! ### Analysis of Core programs -/
 
 /--
-Analyze a Core program, including any external solver invocation
+Verify a Core program, including any external solver invocation
 that is necessary.
 -/
 def Core.verifyProgram
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -907,16 +907,16 @@ def laurelAnalyzeCommand : Command where
   callback := fun v _ => do
     let laurelProgram ← Strata.readLaurelTextFile v[0]
     let (vcResultsOption, errors) ← Strata.Laurel.verifyProgram laurelProgram { VerifyOptions.default with solver := "z3" }
-      if !errors.isEmpty then
-        IO.println s!"==== ERRORS ===="
-      for err in errors do
-        IO.println s!"{err.message}"
-      match vcResultsOption with
-      | none => return
-      | some vcResults =>
-        IO.println s!"==== RESULTS ===="
-        for vc in vcResults do
-          IO.println s!"{vc.obligation.label}: {match vc.outcome with | .ok o => repr o | .error e => e}"
+    if !errors.isEmpty then
+      IO.println s!"==== ERRORS ===="
+    for err in errors do
+      IO.println s!"{err.message}"
+    match vcResultsOption with
+    | none => return
+    | some vcResults =>
+      IO.println s!"==== RESULTS ===="
+      for vc in vcResults do
+        IO.println s!"{vc.obligation.label}: {match vc.outcome with | .ok o => repr o | .error e => e}"
 
 def laurelAnalyzeToGotoCommand : Command where
   name := "laurelAnalyzeToGoto"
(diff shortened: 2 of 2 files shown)


Strata/SimpleAPI.lean
```diff
@@ -112,6 +116,68 @@ writing directly to a file.
 def writeStrataIon : Strata.Program → ByteArray
 | pgm => pgm.toIon
 
+/--
+Read a Laurel source file in textual format and parse it into
+a `Laurel.Program`. Handles dialect loading, parsing, and
+AST translation in one step.
+-/
+def readLaurelTextFile (path : System.FilePath)
```
atomb: Although I think it makes sense to have common pipelines like this in SimpleAPI, ideally they'd be simple compositions of existing steps. So this could be, roughly, `readStrataText` followed by `genericToLaurel`. However, I think that refactoring won't change the type signature of this function, so I'm okay adding this as-is and refactoring later.

Strata/SimpleAPI.lean
```diff
@@ -208,8 +274,8 @@ def Core.callElimUsingContract (p : Core.Program) : Except String Core.Program :
 /-! ### Analysis of Core programs -/
 
 /--
-Do deductive verification of a Core program, including any external solver
-invocation that is necessary.
+Analyze a Core program, including any external solver invocation
```
atomb: What did you have in mind with this change? I do think that this specific API call is limited to deductive verification.

Strata/SimpleAPI.lean
```diff
@@ -225,6 +291,55 @@ def Core.verifyProgram
...
+/--
+Analyze a Laurel program by translating to Core and running
+verification. Returns VC results (if translation succeeded)
+and any translation diagnostics.
+-/
+def Laurel.verifyProgram
+    (program : Laurel.Program)
+    (options : Core.VerifyOptions := .default)
+    : IO (Option Core.VCResults × List DiagnosticModel) :=
+  Strata.Laurel.verifyToVcResults program options
```
atomb, line 309: This is another case that I'd love to exist as a composition of `laurelToCore` and `Core.verifyProgram`.

Strata/SimpleAPI.lean
```diff
@@ -225,6 +291,55 @@ def Core.verifyProgram
...
+    (options : Core.VerifyOptions := .default)
+    : IO (Array DiagnosticModel) :=
+  Strata.Laurel.verifyToDiagnosticModels program options
+
+/-! ### Python direct-to-Core pipeline -/
+
+/--
+Read Python statements from a Strata Ion file.
+-/
+def readPythonIon (path : String)
```
atomb, line 326: Another case that I think should be a composition: `readStrataIon` and `genericToPython`... even if the latter doesn't exist yet :-)

Strata/SimpleAPI.lean
shigoel: The `options` field allows selecting a `checkMode`, which can be bug finding or deductive verification.

Strata/SimpleAPI.lean
shigoel: Agreed.

Strata/SimpleAPI.lean
shigoel: Agreed.

StrataMain.lean
```diff
@@ -950,35 +897,16 @@ def laurelParseCommand : Command where
...
-
-    let uri := Strata.Uri.file path.toString
-    let transResult := Strata.Laurel.TransM.run uri (Strata.Laurel.parseProgram strataProgram)
-    match transResult with
-    | .error transErrors => exitFailure s!"Translation errors: {transErrors}"
-    | .ok laurelProgram =>
-      let (vcResultsOption, errors) ← Strata.Laurel.verifyToVcResults laurelProgram { VerifyOptions.default with solver := "z3" }
+    let laurelProgram ← Strata.readLaurelTextFile v[0]
+    let (vcResultsOption, errors) ← Strata.Laurel.verifyProgram laurelProgram { VerifyOptions.default with solver := "z3" }
       if !errors.isEmpty then
```
MikaelMayer: The `if`, `for`, and `match` here are indented at 6 spaces (as if still inside the old `match` arm), but they're now top-level statements in the `do` block. They should be at 4-space indent to match the surrounding code. Lean accepts this because they're still more indented than the `let`, but it's misleading to read.

StrataMain.lean
```diff
@@ -997,16 +925,8 @@ def laurelAnalyzeToGotoCommand : Command where
   callback := fun v _ => do
     let path : System.FilePath := v[0]
     let content ← IO.FS.readFile path
```
MikaelMayer, line 927: `content` is read here, and then `readLaurelTextFile` on the next line reads the same file again internally. Consider either:
- Adding a variant of `readLaurelTextFile` that accepts already-read content, or
- Having `readLaurelTextFile` return the raw content alongside the parsed program.

This would avoid the double read and keep the API composable for callers that need both the raw text and the parsed program.

Strata/SimpleAPI.lean
```diff
@@ -208,8 +274,8 @@ def Core.callElimUsingContract (p : Core.Program) : Except String Core.Program :
 /-! ### Analysis of Core programs -/
 
 /--
-Do deductive verification of a Core program, including any external solver
-invocation that is necessary.
+Analyze a Core program, including any external solver invocation
+that is necessary.
 -/
```
MikaelMayer, line 283: Nit: the docstring was changed from "Do deductive verification" to "Analyze", but the function is still named `verifyProgram`. The name and doc are slightly at odds — "Analyze" is vague, while the function specifically runs verification (which may or may not be deductive depending on options). Something like "Verify a Core program, including any external solver invocation that is necessary" would be more precise and consistent with the function name.

shigoel: Addressed in https://github.com/strata-org/Strata/pull/733/commits/82ae3c4dbcc06cecf5ba835f4e7945ac26df545a

StrataMain.lean
shigoel: Addressed in https://github.com/strata-org/Strata/pull/733/commits/82ae3c4dbcc06cecf5ba835f4e7945ac26df545a

StrataMain.lean
shigoel: Addressed in https://github.com/strata-org/Strata/pull/733/commits/82ae3c4dbcc06cecf5ba835f4e7945ac26df545a

MikaelMayer (approved):

atomb (approved):

PR #734, opened 2026-04-01
Resolve tvars in inferType's fvar case to fix panic on polymorphic accessor chains

joehendrix:
Resolve tvars in `inferType`'s `.fvar` case to fix a panic when template-generated datatype accessors are passed as arguments to polymorphic functions.

### 1. Add tvar utility functions to `TypeExprF` (`AST.lean`)

`hasTVar`, `substTVars`, `eqNoAnn`, and `matchTVars` provide first-order pattern matching of types containing tvars against concrete types, producing a consistent tvar-to-type substitution.

### 2. Resolve tvars in `inferType`'s `.fvar` case (`Core.lean`)

Template-generated datatype accessors (e.g., `Option..val`) store their types with unresolved tvars in the GlobalContext. When these accessors appeared as arguments to polymorphic functions (e.g., `Sequence.select`), the tvars flowed through to `instTypeM` which expected concrete Type args, causing a panic.

The new `resolveFVarTVars` helper peels the fvar's arrow type, matches each parameter type against the inferred argument type to build a tvar substitution, then substitutes the result type. Two guards preserve the existing tvar pass-through behavior:

- If the substitution is empty, the original type is returned unchanged.
- If the result type still contains unresolved tvars after substitution (e.g., constructors where some tvars don't appear in any parameter), the original type is returned unchanged.

For `perField`-generated accessors specifically, full resolution is structurally guaranteed: the sole parameter is always the full datatype `DataType(tvar_1, ..., tvar_n)`, so every result tvar necessarily appears in the parameter and will be resolved by `matchTVars`.

### 3. Account for reversed fvar arg order (`Core.lean`)

`flattenTypeApp` stores fvar type arguments in reverse (rightmost-first) order, but template-generated accessor types use forward (declaration) order. The inferred type's top-level fvar args are reversed before matching to align the two conventions. This only affects fvar types with 2+ type parameters (e.g., `Either`); single-parameter types like `Option` are unaffected by reversal.

## Tests

- **`PolyUnifTest`**: Updated test 2 — the fix causes the type mismatch to be caught earlier during elaboration rather than at a later typecheck stage. The error is more precise (source location points to the offending expression) but the text changed.
- **`TVarInference.lean`** (new): Regression test with a minimal inline dialect covering the accessor-into-polymorphic-function pattern that triggered the original panic, plus simpler cases for accessor-only and polymorphic-fn-only usage.

## Potential risks

- **Performance**: `resolveFVarTVars` calls `inferType` on each argument, duplicating work the caller will also do. This only affects fvar calls whose types contain tvars (i.e., template-generated accessors), which are a small fraction of all calls.
- **`args.reverse` fragility**: The reversal assumes inferred types always come from `flattenTypeApp` (reversed) while pattern types come from templates (forward). If a future code path produces inferred fvar types without reversal, `matchTVars` would produce an incorrect substitution — resulting in a type mismatch error (not a panic), but potentially confusing to debug.
- **User-defined polymorphic fvars**: Functions registered via `@[scopeTVar]` can have result tvars that don't appear in any parameter (e.g., `weird : tvar A → tvar B`). The partial-resolution guard returns the original type in this case, so the existing tvar pass-through behavior is preserved. If such a bare tvar then flows into a polymorphic `.fn` call, the original panic path is still reachable. This is an uncommon pattern — it requires a user-defined fvar with an uncovered result tvar composed with a polymorphic function — but could be addressed in a follow-up by converting the `.fn` panic to a reported error.


Strata/DDM/Elab/Core.lean
```diff
@@ -989,6 +1007,51 @@ partial def inferType (tctx : TypingContext) (e : Expr) : ElabM TypeExpr := do
...
+      | .arrow _ pty rest =>
+        match arg with
+        | .expr e =>
+          let argType ← inferType tctx e
+          -- fvar/ident args are stored in de Bruijn (reverse) order by
+          -- flattenTypeApp, but accessor type patterns from
+          -- mkDatatypeTypeRef use forward (declaration) order.
+          -- Reverse the top-level args to align with the pattern.
+          let argType := match argType with
+            | .fvar loc i args => .fvar loc i args.reverse
```
tautschnig: A potential hardening: instead of reversing at the call site, could `flattenTypeApp` itself be changed to return args in forward order? That would eliminate the convention mismatch at the source. Acknowledging that this is a larger change that could affect other callers.

tautschnig (commented):
The PR description's "Potential risks" section is refreshingly honest. The `args.reverse` fragility is the main concern - see inline comment.

Strata/DDM/AST.lean
```diff
@@ -181,6 +181,98 @@ termination_by t
...
+    false
+| .arrow _ a1 r1, .arrow _ a2 r2 =>
+  a1.eqNoAnn a2 && r1.eqNoAnn r2
+| _, _ => false
+termination_by a => a
+
+/-- First-order pattern matching of a type with tvars (pattern) against a
+    concrete type (target). Produces a consistent tvar substitution.
+    Returns the updated substitution on success, or `none` on structural mismatch. -/
+protected def matchTVars {α}
```
joscoh, line 235: This makes me think the DDM is slowly evolving into having a separate type unification and inference algorithm.

StrataTest/Languages/Core/Tests/PolymorphicDatatypeTest.lean
```diff
@@ -541,12 +541,8 @@ end Strata.PolymorphicDatatypeTest
...
 
 -- Verify that the program does not panic during type inference (issue #650).
--- The program has type errors that should be reported gracefully.
-/--
-error: Could not infer type parameter 2 for Core.seq_select
----
-error: Encountered .|| expression when MethodSetting expected.
--/
+-- The tvar resolution fix (PR #734) also resolves the type inference
+-- failure here, so this program now elaborates without errors.
```
joscoh: Nit: comments should be stateless

StrataTest/Languages/Core/Tests/PolyUnifTest.lean
```diff
@@ -45,6 +45,10 @@ function OptString_len (o : (Option string)) : int {
 -- Test 2: Incorrect polymorphic instantiation caught by type checking
 ---------------------------------------------------------------------
 
+/--
+error: Encountered int expression when string expected.
```
joscoh, line 51: It is no longer possible to test the Core type inference with concrete syntax?

joscoh (dismissed):

StrataTest/Languages/Core/Tests/PolyUnifTest.lean
joehendrix: This would catch a type error.  If we want the DDM to intentionally allow certainly clases of errors through, I'll need to revise the design with that feature planned.

Strata/DDM/AST.lean
joehendrix: Maybe, I had originally intended for the DDM to have a type checker for a fairly simple type system since it seems like an IR may want some typechecking (but optimized for performance, gradual ingestion, etc)

Maybe Core should migrate to not use the DDM Expr/Type code so it can better isolate everything into its own type checker?  It sounds like there's a preference among dialect writers to have their own typecheckers.

PR #736, opened 2026-04-02
Improve PySpec-to-Laurel type translation and fix subclass resolution errors

joehendrix:
Improve PySpec-to-Laurel type translation and fix subclass resolution errors that affect all boto3 pyspecs with nested exception classes.

### 1. Translate nested subclasses in `classDefToLaurel`

The pyspec compiler correctly emits nested classes as `subclasses` in the Ion format, but `classDefToLaurel` ignored the `cls.subclasses` field entirely. Fields referencing the subclass type (e.g. `exceptions: class(_Exceptions)`) produced `UserDefined` references to types that were never created, causing resolution failures like `servicelib_Database__Exceptions is not defined`.

Added a recursive loop over `cls.subclasses` at the end of `classDefToLaurel`. The existing `prefixName` logic produces the correct name (e.g. `servicelib_Database__Exceptions`) matching what `specTypeToLaurelType` generates for `class(...)` field type references. The function is marked `partial` since Lean cannot prove termination through `Array ClassDef`.

### 2. Replace if-chain with `knownIdentTypes` HashMap

The `specTypeToLaurelType` function had a chain of `if nm == PythonIdent.X then return ...` comparisons. Replaced with a single `Std.HashMap PythonIdent HighTypeMd` lookup. This also adds previously missing mappings: `builtins.dict` → `DictStrAny`, `builtins.bytearray` → `TString`, `builtins.complex` → `TReal`, `builtins.Exception` → `Error`.

### 3. Use `UserDefined` instead of `TCore` for prelude types

All prelude type references (`Any`, `Error`, `DictStrAny`, `ListStr`, `StrOrNone`, `IntOrNone`, `BoolOrNone`) now use `UserDefined` instead of `TCore`. These types are defined in the Laurel prelude (`PythonLaurelCorePrelude.lean`) and should participate in Laurel resolution rather than bypassing it as opaque pass-through names.

### 4. Lift type constants to named defs

Extracted all inline `mkTy`/`mkUserDefined` calls into named constants (`tyBool`, `tyInt`, `tyAny`, `tyDictStrAny`, etc.) to reduce string-literal duplication and catch typos at compile time.

## Testing

- New regression test with a mock `Database` pyspec reproducing the `_Exceptions` nested-class pattern. Calls `Laurel.resolve` on the combined program after `pyAnalyzeLaurel` and asserts zero resolution errors.
- Updated `#guard_msgs` expectations in `ToLaurelTest` for the `TCore` → `UserDefined` change.
- Verified zero `_Exceptions` resolution errors across all 50 kiro benchmarks (previously affected 12).

### DIFF
--- a/Strata/Languages/Python/Specs/ToLaurel.lean
+++ b/Strata/Languages/Python/Specs/ToLaurel.lean
@@ -123,7 +123,7 @@ private def tyBoolOrNone  : HighTypeMd := mkUserDefined "BoolOrNone"
 mutual
 
 /-- Convert a SpecAtomType to a string for error messages. -/
-partial def atomTypeToString (a : SpecAtomType) : String :=
+def atomTypeToString (a : SpecAtomType) : String :=
   match a with
   | .ident nm args =>
     if nm == PythonIdent.noneType && args.isEmpty then "None"
@@ -139,14 +139,21 @@ partial def atomTypeToString (a : SpecAtomType) : String :=
   | .intLiteral v => s!"Literal[{v}]"
   | .stringLiteral v => s!"Literal[\"{v}\"]"
   | .typedDict _ _ _ => "TypedDict"
+termination_by sizeOf a
 
 /-- Convert a SpecType to a string for error messages. -/
-partial def specTypeToString (t : SpecType) : String :=
-  if t.atoms.size == 1 then
-    atomTypeToString t.atoms[0]!
+def specTypeToString (t : SpecType) : String :=
+  if h : t.atoms.size = 1 then
+    atomTypeToString t.atoms[0]
   else
     let strs := t.atoms.map atomTypeToString
     String.intercalate " | " strs.toList
+termination_by sizeOf t
+decreasing_by
+  · cases t
+    decreasing_tactic
+  · cases t
+    decreasing_tactic
 
 end
 
@@ -318,7 +325,7 @@ private def mkStmtWithLoc (e : StmtExpr) (loc : SourceRange) (msg : String := ""
     Returns `none` for unsupported expressions (placeholders).
     Uses Core prelude function names (Any_len, DictStrAny_contains, etc.)
     which are resolved after the Core prelude is prepended. -/
-partial def specExprToLaurel (e : SpecExpr) (md : Imperative.MetaData Core.Expression)
+def specExprToLaurel (e : SpecExpr) (md : Imperative.MetaData Core.Expression)
   : ToLaurelM (Option StmtExprMd) :=
   match e with
   | .placeholder => do
@@ -517,7 +524,7 @@ def funcDeclToLaurel (procName : String) (func : FunctionDecl)
   }
 
 /-- Convert a class definition to Laurel types and procedures. -/
-partial def classDefToLaurel (cls : ClassDef) : ToLaurelM Unit := do
+def classDefToLaurel (cls : ClassDef) : ToLaurelM Unit := do
   let prefixedName ← prefixName cls.name
   -- Register alias from unprefixed to prefixed name for type resolution
   if prefixedName != cls.name then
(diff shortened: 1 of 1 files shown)


tautschnig (dismissed):

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -94,12 +94,36 @@ private def mkTy (ty : HighType) : HighTypeMd :=
 private def mkCore (s : String) : HighTypeMd :=
   { val := .TCore s, md := default }
```
MikaelMayer: `mkCore` is now unused — it can be removed.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -161,22 +185,22 @@ def detectOptionalType (ty : SpecType) : ToLaurelM (Option HighTypeMd) := do
 
   -- All non-None string literals → StrOrNone
   if otherAtoms.all (fun a => match a with | .stringLiteral _ => true | _ => false) then
-    return some (mkCore "StrOrNone")
+    return some tyStrOrNone
```
MikaelMayer, line 191: The docstring on `detectOptionalType` (above this hunk) still lists `TCore` (e.g. `Union[None, str] → TCore "StrOrNone"`). These are now `UserDefined` after this PR.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -189,66 +213,79 @@ def detectOptionalType (ty : SpecType) : ToLaurelM (Option HighTypeMd) := do
...
+    | .typedDict _ _ _ => return some tyDictStrAny
+    | .intLiteral _ => return some tyIntOrNone
     | _ => return none
   else
     return none
 
+/-- Known PythonIdent → Laurel type mappings for single-atom ident types.
+    - `bytes`/`bytearray` → TString (closest string-like approximation)
+    - `complex` → TReal (no complex type in SMT; real is the closest numeric type)
+    - `Exception` → Core "Error" (matches CorePrelude's Error datatype)
```
MikaelMayer: Docstring says `Exception → Core "Error"` but `tyError` is `mkUserDefined "Error"` (i.e. `UserDefined`, not `TCore`). The other bullet for `typing.Any` correctly says `UserDefined`.

StrataTest/Languages/Python/Specs/dispatch_test/servicelib/Database.py
```diff
@@ -0,0 +1,24 @@
...
+    'Key': Required[str],
+})
+
+class DatabaseError(Exception):
+    response: dict
+
+class Database:
+    # Nested exception container — mimics boto3 service client pattern.
+    # The pyspec compiler emits _Exceptions as a subclass in the Ion file,
+    # but ToLaurel does not translate subclasses, so the field type
```
MikaelMayer: This comment describes the old bug ("ToLaurel does not translate subclasses") rather than the current behavior. After this PR, `classDefToLaurel` does recurse into subclasses, so the comment is misleading. Consider updating it to describe the purpose of the fixture (e.g. "exercises the nested-subclass translation path in `classDefToLaurel`").

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -484,7 +517,7 @@ def funcDeclToLaurel (procName : String) (func : FunctionDecl)
   }
 
 /-- Convert a class definition to Laurel types and procedures. -/
-def classDefToLaurel (cls : ClassDef) : ToLaurelM Unit := do
+partial def classDefToLaurel (cls : ClassDef) : ToLaurelM Unit := do
```
MikaelMayer: I'm not too sure about this addition of partial. Is proving termination too hard? There are a lot of helpers around here.

joehendrix: Thanks. It was subclasses and Claude/myself being lazy.  An easy fix.

tautschnig (approved):

PR #737, opened 2026-04-02
Remove big step semantics

atomb:
Removes the big-step operational semantics (`EvalStmt`/`EvalBlock`) from the Imperative dialect, making the small-step semantics (`StepStmt`/`EvalStmtSmall`/`EvalStmtsSmall`) the sole statement semantics. The `Env` and `ExtendEval` definitions are relocated from the deleted `StmtSemantics.lean` into `StmtSemanticsSmallStep.lean`. All proofs in `SemanticsProps.lean` and `StatementSemanticsProps.lean` are rewritten to work against the small-step relation. A mutual inductive `CoreStepStar`/`EvalCommand` is introduced to satisfy strict positivity for `call_sem`.

Along the way, the CFG semantics are updated to track a failure flag, and `Config` is renamed to `CFGConfig` to avoid collision with the small-step `Config`. This module was already broken, and undetected because it wasn't imported anywhere.

### DIFF
--- a/Strata/Languages/Core/StatementSemantics.lean
+++ b/Strata/Languages/Core/StatementSemantics.lean
@@ -56,6 +56,7 @@ instance : HasBool Core.Expression where
   tt := Core.true
   ff := Core.false
   tt_is_not_ff := by unfold Core.true Core.false; unfold Lambda.LExpr.boolConst; simp
+  boolTy := .forAll [] (.tcons "bool" [])
 
 instance : HasNot Core.Expression where
   not
--- a/Examples/NondetCond.core.st
+++ b/Examples/NondetCond.core.st
@@ -1,0 +1,14 @@
+// Test non-deterministic conditions in if and while statements
+procedure nondet_if()
+  modifies x;
+  ensures x >= 0;
+{
+  var x : int;
+  x := 0;
+  if * {
+    x := 1;
+  } else {
+    x := 2;
+  }
+  assert x >= 0;
+}
--- a/Strata/Backends/CBMC/CollectSymbols.lean
+++ b/Strata/Backends/CBMC/CollectSymbols.lean
@@ -125,11 +125,11 @@ private def collectGlobalSymbols (pgm : Core.Program) :
       let gty ← Lambda.LMonoTy.toGotoType ty.toMonoTypeUnsafe
       let tyJson := CProverGOTO.tyToJson gty
       let valueJson ← match e with
-        | some expr =>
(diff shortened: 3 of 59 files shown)


Strata/DL/Imperative/CFGSemantics.lean
```diff
@@ -88,18 +90,22 @@ inductive EvalNondetBlock
...
 
   | step_goto_some :
-    EvalCmds P EvalCmd δ σ cs σ' →
+    EvalCmds P EvalCmd δ σ cs σ' failed →
     lt ∈ ls →
     EvalNondetBlock P EvalCmd extendEval
-      σ ⟨ cs, .goto ls _ ⟩ (.cont lt σ')
+      σ ⟨ cs, .goto ls _ ⟩ (.cont lt σ' failed)
+
+def updateFailure : CFGConfig l P → Bool → CFGConfig l P
```
MikaelMayer, line 110: `updateFailure` could use a docstring — it's used in `StepCFG.eval_next` and the semantics of OR-ing the incoming failure flag with the block's failure flag is worth documenting briefly.

Strata/Languages/Core/StatementSemantics.lean
```diff
@@ -295,8 +321,9 @@ inductive EvalCommand (π : String → Option Procedure) (φ : CoreEval → Pure
     (∀ pre, (Procedure.Spec.getCheckExprs p.spec.preconditions).contains pre →
       isDefinedOver (HasVarsPure.getVars) σAO pre ∧
       δ σAO pre = .some HasBool.tt) →
-    @Imperative.EvalBlock Expression Command (EvalCommand π φ) (EvalPureFunc φ) _ _ _ _ _ _ _
-      ⟨σAO, δ, false⟩ p.body ρ' →
+    CoreStepStar π φ
```
MikaelMayer, line 314: The NOTE comment above `call_sem` (around line 290) still references `Imperative.EvalBlock`, which no longer exists after this PR. The error scenario described there may no longer apply since the mutual inductive now uses `CoreStepStar` instead. Consider updating or removing that comment.

Strata/DL/Imperative/CFGSemantics.lean
atomb: Agreed!

Strata/Languages/Core/StatementSemantics.lean
atomb: Yes, good point.

PR #738, opened 2026-04-02
Prove that `LExpr.eval` is sound w.r.t. small-step semantics Step

aqjune-aws:
Main theorem: `eval_StepStar`: for every expression `e`, there exists an `e'` reachable from `e` by zero or more `Step`s such that `e'.eraseMetadata = (eval n σ e).eraseMetadata`. This establishes soundness of the fuel-bounded evaluator with respect to the small-step reduction relation, modulo metadata.

```
theorem eval_StepStar
    ...
    (hWF : FactoryWF σ.config.factory)
    (hEval : e2 = LExpr.eval n σ e) :
    ∃ (e' : LExpr Tbase.mono),
      StepStar σ.config.factory (Scopes.toEnv σ.state) e e' ∧
      e'.eraseMetadata = e2.eraseMetadata := by
...
```

This patch also contains two additional important theorems:
- `canonical_value_not_step`: canonical values are normal forms: no `Step` rule can fire on them. This ensures the step relation does not reduce already-evaluated results.
- `eval_eraseMetadata_invariant`: `LExpr.eval` is invariant under metadata erasure — expressions that agree modulo metadata evaluate to results that also agree modulo metadata.

The Step relation is relaxed from strict call-by-value: `beta` no longer requires the argument to be a canonical value, and `reduce_2` / `eq_reduce_rhs` no longer require the left operand to be canonical. To match what `evalCore` does via `substFvarsFromState` does, three new under-binder rules (`abs_subst_fvar`, `quant_subst_fvar_body`, `quant_subst_fvar_trigger`) allow free-variable substitution beneath lambda/quantifier binders. These rules seem to exist in lambda calculus with explicit substitution rules, even though they don't appear in the most generic lambda calculus style. This is purely to support the current behavior of LExpr.lambda.

### DIFF
--- a/.github/workflows/cbmc.yml
+++ b/.github/workflows/cbmc.yml
@@ -106,6 +106,10 @@ jobs:
         run: |
           pip install ./Tools/Python
           ./StrataTest/Languages/Python/run_py_cbmc_tests.sh
+      - name: Run property summary tests
+        shell: bash
+        run: |
+          ./StrataTest/Backends/CBMC/GOTO/test_property_summary_e2e.sh
       - name: Run Laurel CBMC pipeline tests
         shell: bash
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,7 +242,7 @@ jobs:
               sudo cp z3-4.13.4-x64-glibc-2.35/bin/z3 /usr/local/bin/
             fi
     - name: Run PySpec and dispatch tests
-      run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest StrataTest.Languages.Python.AnalyzeLaurelTest StrataTest.Languages.Python.Specs.IdentifyOverloadsTest StrataTest.Languages.Python.VerifyPythonTest
+      run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest StrataTest.Languages.Python.AnalyzeLaurelTest StrataTest.Languages.Python.Specs.IdentifyOverloadsTest StrataTest.Languages.Python.VerifyPythonTest StrataTest.Languages.Python.PropertySummaryTest
     - name: Run test script
       run: FAIL_FAST=1 ./scripts/run_cpython_tests.sh ${{ matrix.python_version }}
       working-directory: Tools/Python
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -1241,10 +1241,32 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     return (ctx, stmts)
 
   -- Assert statement
-  | .Assert _ test _msg => do
+  | .Assert _ test msg => do
     let condExpr ← translateExpr ctx test
-    let assertStmt := mkStmtExprMdWithLoc (StmtExpr.Assert (Any_to_bool condExpr)) md
-    return (ctx, [assertStmt])
+    -- Extract assert message as property summary
+    let md' := match msg.val with
+      | some (.Constant _ (.ConString _ str) _) => md.withPropertySummary str.val
+      | _ => md
+    -- Check if condition contains a Hole - if so, hoist to variable
+    let (condStmts, finalCondExpr, condCtx) :=
+      match condExpr.val with
+      | .Hole =>
+        let freshVar := s!"assert_cond_{test.toAst.ann.start.byteIdx}"
+        let varType := mkHighTypeMd .TBool
+        let varDecl := mkStmtExprMd (StmtExpr.LocalVariable freshVar varType (some condExpr))
+        let varRef := mkStmtExprMd (StmtExpr.Identifier freshVar)
+        ([varDecl], varRef, { ctx with variableTypes := ctx.variableTypes ++ [(freshVar, "bool")] })
+      | _ => ([], condExpr, ctx)
+
+    let assertStmt := mkStmtExprMdWithLoc (StmtExpr.Assert (Any_to_bool finalCondExpr)) md'
+
+    -- Wrap in block if we hoisted condition
+    let result := if condStmts.isEmpty then
+      assertStmt
+    else
+      mkStmtExprMdWithLoc (StmtExpr.Block (condStmts ++ [assertStmt]) none) md
+
+    return (condCtx, [result])
 
   --Ignore comments in source code
   | .Expr _ (.Constant _ (.ConString _ _) _) => return (ctx, [])
--- a/StrataTest/Backends/CBMC/GOTO/test_property_summary_e2e.sh
+++ b/StrataTest/Backends/CBMC/GOTO/test_property_summary_e2e.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+# E2E test: property summary metadata flows from Laurel through GOTO to CBMC output.
+#
+# Verifies that Laurel `assert ... summary "..."` annotations appear in
+# CBMC's verification output as property descriptions.
+
+set -eo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/../../../.." && pwd)"
+LAUREL_TO_CBMC="$PROJECT_ROOT/StrataTest/Languages/Laurel/laurel_to_cbmc.sh"
+
+WORK=$(mktemp -d)
+trap 'rm -rf "$WORK"' EXIT
+
+# Create Laurel program with property summaries
+cat > "$WORK/test.lr.st" << 'LAUREL'
+procedure main() {
+    var x: int := 5;
+    var y: int := 3;
+    assert x + y == 8 summary "addition equals eight";
+    assert x - y == 2 summary "difference equals two"
+};
+LAUREL
+
+# Run the full pipeline (strata → symtab2gb → goto-cc → goto-instrument → cbmc)
+cbmc_out=$("$LAUREL_TO_CBMC" "$WORK/test.lr.st" 2>&1 || true)
+
+# Verify CBMC output contains property summaries
+for summary in "addition equals eight" "difference equals two"; do
+  if echo "$cbmc_out" | grep -q "$summary"; then
+    echo "CBMC output: '$summary' found"
+  else
+    echo "FAIL: '$summary' not in CBMC output"
+    echo "$cbmc_out"
+    exit 1
+  fi
+done
+
+# Verify CBMC says SUCCESSFUL
+if echo "$cbmc_out" | grep -q "VERIFICATION SUCCESSFUL"; then
+  echo "CBMC: VERIFICATION SUCCESSFUL"
+else
+  echo "FAIL: CBMC did not report VERIFICATION SUCCESSFUL"
+  echo "$cbmc_out"
+  exit 1
+fi
+
+echo "PASS: property summaries flow end-to-end from Laurel to CBMC"
--- a/StrataTest/Languages/Python/PropertySummaryTest.lean
+++ b/StrataTest/Languages/Python/PropertySummaryTest.lean
@@ -1,0 +1,62 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import StrataTest.Util.Python
+import Strata.SimpleAPI
+import Strata.Languages.Python.PyFactory
+
+/-! ## Test: Python assert messages propagate as property summaries
+
+Verifies that `assert cond, "message"` in Python flows through as a
+property summary in the Core verification results.
+-/
+
+namespace Strata.Python.PropertySummaryTest
+
+open Strata (pyTranslateLaurel)
+
+/-- Compile a Python string to Ion, translate to Core, verify, and return
+    the property summaries from the VCResults. -/
+private def getPropertySummaries (pythonCmd : System.FilePath) (source : String)
+    : IO (Array String) := do
+  IO.FS.withTempDir fun tmpDir => do
+    let pyFile := tmpDir / "test.py"
+    IO.FS.writeFile pyFile source
+    let dialectFile := tmpDir / "dialect.ion"
+    IO.FS.writeBinFile dialectFile Python.Python.toIon
+    let ionFile := tmpDir / "test.python.st.ion"
+    let child ← IO.Process.spawn {
+      cmd := pythonCmd.toString
+      args := #["-m", "strata.gen", "py_to_strata",
+                "--dialect", dialectFile.toString,
+                pyFile.toString, ionFile.toString]
+      inheritEnv := true
+      stdin := .null, stdout := .null, stderr := .piped
+    }
+    let stderr ← child.stderr.readToEnd
+    let exitCode ← child.wait
+    if exitCode ≠ 0 then
+      throw <| .userError s!"py_to_strata failed (exit code {exitCode}): {stderr}"
+    let (core, _diags) ← match ← pyTranslateLaurel ionFile.toString |>.toBaseIO with
+      | .ok r => pure r
+      | .error msg => throw <| .userError s!"pyTranslateLaurel failed: {msg}"
+    let results ← match ← Core.verifyProgram core Core.VerifyOptions.quiet
+        (moreFns := Strata.Python.ReFactory) |>.toBaseIO with
+      | .ok r => pure r
+      | .error msg => throw <| .userError s!"verifyCore failed: {msg}"
+    return results.filterMap fun vcr =>
+      vcr.obligation.metadata.getPropertySummary
+
+#guard_msgs in
+#eval withPython (warnOnSkip := false) fun pythonCmd => do
+  let summaries ← getPropertySummaries pythonCmd
+    "def main(x: int) -> None:\n    assert x == x, \"reflexivity\"\n    assert x + 0 == x, \"additive identity\"\n"
+  let expected := #["reflexivity", "additive identity"]
+  for msg in expected do
+    unless summaries.any (· == msg) do
+      throw <| .userError s!"FAIL: \"{msg}\" not found in summaries: {summaries}"
+
+end Strata.Python.PropertySummaryTest
--- a/StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_arithmetic.expected
@@ -1,14 +1,14 @@
-test_arithmetic.py(7, 4): ✅ pass - assert_assert(102)_calls_Any_to_bool_0
-test_arithmetic.py(7, 4): ✅ pass - assert(102)
-test_arithmetic.py(12, 4): ✅ pass - assert_assert(226)_calls_Any_to_bool_0
-test_arithmetic.py(12, 4): ✅ pass - assert(226)
-test_arithmetic.py(16, 4): ✅ pass - assert_assert(345)_calls_Any_to_bool_0
-test_arithmetic.py(16, 4): ✅ pass - assert(345)
-test_arithmetic.py(20, 4): ✅ pass - assert_assert(458)_calls_Any_to_bool_0
-test_arithmetic.py(20, 4): ✅ pass - assert(458)
+test_arithmetic.py(7, 4): ✅ pass - multiplication implemented incorrectly
+test_arithmetic.py(7, 4): ✅ pass - multiplication implemented incorrectly
+test_arithmetic.py(12, 4): ✅ pass - multiplication implemented incorrectly
+test_arithmetic.py(12, 4): ✅ pass - multiplication implemented incorrectly
+test_arithmetic.py(16, 4): ✅ pass - addition implemented incorrectly
+test_arithmetic.py(16, 4): ✅ pass - addition implemented incorrectly
+test_arithmetic.py(20, 4): ✅ pass - subtraction implemented incorrectly
+test_arithmetic.py(20, 4): ✅ pass - subtraction implemented incorrectly
 test_arithmetic.py(23, 4): ✅ pass - set_quot_calls_PFloorDiv_0
-test_arithmetic.py(24, 4): ✅ pass - assert_assert(567)_calls_Any_to_bool_0
-test_arithmetic.py(24, 4): ✅ pass - assert(567)
+test_arithmetic.py(24, 4): ✅ pass - floor division implemented incorrectly
+test_arithmetic.py(24, 4): ✅ pass - floor division implemented incorrectly
 test_arithmetic.py(26, 0): ✅ pass - ite_cond_calls_Any_to_bool_0
 DETAIL: 12 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success
--- a/StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_augmented_assign.expected
@@ -1,8 +1,8 @@
-test_augmented_assign.py(5, 4): ✅ pass - assert_assert(81)_calls_Any_to_bool_0
-test_augmented_assign.py(5, 4): ✅ pass - assert(81)
-test_augmented_assign.py(7, 4): ✅ pass - assert_assert(124)_calls_Any_to_bool_0
-test_augmented_assign.py(7, 4): ✅ pass - assert(124)
-test_augmented_assign.py(9, 4): ✅ pass - assert_assert(167)_calls_Any_to_bool_0
-test_augmented_assign.py(9, 4): ✅ pass - assert(167)
+test_augmented_assign.py(5, 4): ✅ pass - 5 + 3 == 8
+test_augmented_assign.py(5, 4): ✅ pass - 5 + 3 == 8
+test_augmented_assign.py(7, 4): ✅ pass - 8 - 2 == 6
+test_augmented_assign.py(7, 4): ✅ pass - 8 - 2 == 6
+test_augmented_assign.py(9, 4): ✅ pass - 6 * 2 == 12
+test_augmented_assign.py(9, 4): ✅ pass - 6 * 2 == 12
 DETAIL: 6 passed, 0 failed, 0 inconclusive
 RESULT: Analysis success
--- a/StrataTest/Languages/Python/expected_laurel/test_boolean_logic.expected
+++ b/StrataTest/Languages/Python/expected_laurel/test_boolean_logic.expected
@@ -1,30 +1,30 @@
-test_boolean_logic.py(5, 4): ✅ pass - assert_assert(73)_calls_PAnd_0
-test_boolean_logic.py(5, 4): ✅ pass - assert_assert(73)_calls_Any_to_bool_1
-test_boolean_logic.py(5, 4): ✅ pass - assert(73)
-test_boolean_logic.py(6, 4): ✅ pass - assert_assert(124)_calls_PAnd_0
-test_boolean_logic.py(6, 4): ✅ pass - assert_assert(124)_calls_Any_to_bool_1
-test_boolean_logic.py(6, 4): ✅ pass - assert(124)
-test_boolean_logic.py(9, 4): ✅ pass - assert_assert(201)_calls_POr_0
-test_boolean_logic.py(9, 4): ✅ pass - assert_assert(201)_calls_Any_to_bool_1
-test_boolean_logic.py(9, 4): ✅ pass - assert(201)
-test_boolean_logic.py(10, 4): ✅ pass - assert_assert(251)_calls_POr_0
-test_boolean_logic.py(10, 4): ✅ pass - assert_assert(251)_calls_Any_to_bool_1
-test_boolean_logic.py(10, 4): ✅ pass - assert(251)
-test_boolean_logic.py(13, 4): ✅ pass - assert_assert(320)_calls_Any_to_bool_0
-test_boolean_logic.py(13, 4): ✅ pass - assert(320)
-test_boolean_logic.py(14, 4): ✅ pass - assert_assert(365)_calls_Any_to_bool_0
-test_boolean_logic.py(14, 4): ✅ pass - assert(365)
+test_boolean_logic.py(5, 4): ✅ pass - True and True should be True
+test_boolean_logic.py(5, 4): ✅ pass - True and True should be True
+test_boolean_logic.py(5, 4): ✅ pass - True and True should be True
+test_boolean_logic.py(6, 4): ✅ pass - True and False should be False
+test_boolean_logic.py(6, 4): ✅ pass - True and False should be False
+test_boolean_logic.py(6, 4): ✅ pass - True and False should be False
+test_boolean_logic.py(9, 4): ✅ pass - True or False should be True
+test_boolean_logic.py(9, 4): ✅ pass - True or False should be True
+test_boolean_logic.py(9, 4): ✅ pass - True or False should be True
+test_boolean_logic.py(10, 4): ✅ pass - False or False should be False
+test_boolean_logic.py(10, 4): ✅ pass - False or False should be False
+test_boolean_logic.py(10, 4): ✅ pass - False or False should be False
+test_boolean_logic.py(13, 4): ✅ pass - not False should be True
+test_boolean_logic.py(13, 4): ✅ pass - not False should be True
+test_boolean_logic.py(14, 4): ✅ pass - not not True should be True
+test_boolean_logic.py(14, 4): ✅ pass - not not True should be True
 test_boolean_logic.py(21, 4): ✅ pass - ite_cond_calls_PAnd_0
 test_boolean_logic.py(21, 4): ✅ pass - ite_cond_calls_Any_to_bool_1
-test_boolean_logic.py(25, 4): ✅ pass - assert_assert(570)_calls_Any_to_bool_0
-test_boolean_logic.py(25, 4): ✅ pass - assert(570)
+test_boolean_logic.py(25, 4): ✅ pass - combined and condition failed
+test_boolean_logic.py(25, 4): ✅ pass - combined and condition failed
 test_boolean_logic.py(27, 4): ✅ pass - ite_cond_calls_POr_0
 test_boolean_logic.py(27, 4): ✅ pass - ite_cond_calls_Any_to_bool_1
-test_boolean_logic.py(31, 4): ✅ pass - assert_assert(700)_calls_Any_to_bool_0
-test_boolean_logic.py(31, 4): ✅ pass - assert(700)
+test_boolean_logic.py(31, 4): ✅ pass - combined or condition failed
+test_boolean_logic.py(31, 4): ✅ pass - combined or condition failed
 test_boolean_logic.py(33, 4): ✅ pass - ite_cond_calls_Any_to_bool_0
-test_boolean_logic.py(37, 4): ✅ pass - assert_assert(825)_calls_Any_to_bool_0
-test_boolean_logic.py(37, 4): ✅ pass - assert(825)
+test_boolean_logic.py(37, 4): ✅ pass - not condition failed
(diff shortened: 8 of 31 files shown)


Strata/DL/Lambda/LExprEval.lean
```diff
@@ -68,7 +68,6 @@ def eql (F : @Factory T.base) (e1 e2 : LExpr T) : Option Bool :=
   else
   -- Disproving equality is harder, we have several special cases
   match _he: e1, e2 with
-  -- Case 1: Syntactic inequality of (non-real) constants implies semantic inequality
```
joscoh, line 71: Why is this comment removed?

Strata/DL/Lambda/LExprEval.lean
```diff
@@ -89,26 +88,26 @@ def eql (F : @Factory T.base) (e1 e2 : LExpr T) : Option Bool :=
...
-      else if f1.name.name != f2.name.name then some false
-      else
-      -- If all arguments are provably equal, constructor app is equal
-      -- If any are not equal, then they are not equal by injectivity
-      -- Otherwise, incomparable
-      List.foldl (fun acc (⟨a1, _⟩, a2) =>
-        eqlCombine acc (eql F a1 a2)
-      ) (some true) (args1.attach.zip args2)
-    | _, _ => none
+    match _h1: Factory.callOfLFunc F e1 false with
```
joscoh: Is there a reason the simultaneous match needs to become a nested one?

Strata/DL/Lambda/LExprWF.lean
```diff
@@ -391,6 +391,558 @@ where
...
+      simp only [LExpr.substFvars.substFvarsAux, LExpr.eraseMetadata, LExpr.replaceMetadata]
+      exact congr (congrArg _ ihl) ihr
+    | quant m qk n ty tr b iht ihb =>
+      simp only [LExpr.substFvars.substFvarsAux, LExpr.eraseMetadata, LExpr.replaceMetadata]
+      exact congr (congrArg _ iht) ihb
+    | ite m c t f ihc iht ihf =>
+      simp only [LExpr.substFvars.substFvarsAux, LExpr.eraseMetadata, LExpr.replaceMetadata]
+      exact congr (congr (congrArg _ ihc) iht) ihf
+
+/-- If `Map.find? sm x = some v` and `v` is closed, then
```
joscoh: Why does `v` have to be closed here? This theorem should hold unconditionally, unless I am missing something.

Strata/DL/Lambda/LExprWF.lean
```diff
@@ -391,6 +391,558 @@ where
...
+    (sm : Map (Identifier T.IDMeta) (LExpr T.mono))
+    (h_find : Map.find? sm x = none) :
+    LExpr.substFvars (.fvar m_meta x ty) sm = .fvar m_meta x ty := by
+  simp only [LExpr.substFvars]
+  split
+  · rfl
+  · simp [LExpr.substFvars.substFvarsAux, h_find]
+
+/-- substFvars preserves "not free": if k ∉ freeVars e and all values in sm are
+closed, then k ∉ freeVars (substFvars e sm). -/
```
joscoh: Not blocking, but these kinds of results seem more naturally stated via subsets, i.e. if all `sm.values` are closed, the free variables of `substFvars e sm` are a subset of the free variables of `e`. 
(As an aside, the generalized version would be that `k` is free in `(LExpr.substFvars e sm)` iff `k` is free in `e` and does not appear in `sm` or there exists `v` free in `e` such that `sm[k] = e'` and `k` is free in `e'`. Then the above result is derivable from this).

Strata/DL/Lambda/LExprWF.lean
```diff
@@ -391,6 +391,558 @@ where
...
+    simp only [LExpr.substFvars.substFvarsAux, LExpr.freeVars, List.map_append, List.mem_append, not_or] at hk ⊢
+    exact ⟨ih1 hk.1, ih2 hk.2⟩
+  | quant _ _ _ _ _ _ ih1 ih2 =>
+    simp only [LExpr.substFvars.substFvarsAux, LExpr.freeVars, List.map_append, List.mem_append, not_or] at hk ⊢
+    exact ⟨ih1 hk.1, ih2 hk.2⟩
+  | ite _ _ _ _ ih1 ih2 ih3 =>
+    simp only [LExpr.substFvars.substFvarsAux, LExpr.freeVars, List.map_append, List.mem_append, not_or] at hk ⊢
+    exact ⟨⟨ih1 hk.1.1, ih2 hk.1.2⟩, ih3 hk.2⟩
+
+/-- substFvars on const is identity. -/
```
joscoh: Suggestion: instead of all of these lemmas, you could just prove that
```
LExpr.substFvars e sm = match e with
    | .const _ _ => e | .bvar _ _ => e | .op _ _ _ => e
    | .fvar _ name _ => match sm.find? name with | some to => to | none => e
    | .abs m name ty e' => .abs m name ty (substFvars e' sm)
    | .quant m qk name ty tr' e' => .quant m qk name ty (substFvars tr' sm) (substFvars e' sm)
    | .app m fn e' => .app m (substFvars fn sm) (substFvars e' sm)
    | .ite m c t e' => .ite m (substFvars c sm) (substFvars t sm) (substFvars e' sm)
    | .eq m e1 e2 => .eq m (substFvars e1 sm) (substFvars e2 sm)
```
Because the `isEmpty` case is the identity. Then rewriting with this lemma gives you the following 8 lemmas for free.

Strata/DL/Lambda/LState.lean
```diff
@@ -160,10 +160,134 @@ instance : ToFormat (T.Identifier × LState T) where
...
+    intro tsm htsm; induction tsm with
+    | nil => rfl
+    | cons p rest ih =>
+      obtain ⟨k, ⟨_, _⟩⟩ := p; simp only [Map.find?] at htsm
+      by_cases hk : k = x
+      · rw [if_pos hk] at htsm; exact absurd htsm (by simp)
+      · rw [if_neg hk] at htsm; simp only [List.map, Map.find?, if_neg hk]; exact ih htsm
+  exact LExpr.substFvars_fvar_none m_meta x ty _ (h_find_none _ h_tsm)
+
+set_option checkBinderAnnotations false in
```
joscoh: Why do you need this?

Strata/DL/Lambda/LState.lean
```diff
@@ -160,10 +160,134 @@ instance : ToFormat (T.Identifier × LState T) where
...
+/-- If `Maps.find? σ.state x = some (ty_val, v)` and `v` is closed, then
+    `substFvarsFromState σ (.fvar m x ty) = v`. -/
```
joscoh: I am confused why the "`v` is closed" is assumption necessary here?

Strata/DL/Util/Relations.lean
```diff
@@ -70,5 +70,11 @@ theorem reflTransT_to_prop {A : Type} {r : A → A → Prop} {a b : A} :
   | .refl _ => 0
   | .step _ _ _ _ rest => 1 + rest.len
 
+theorem ReflTrans.trans {r : Relation A} {x y z : A}
```
joscoh: How is this different from `ReflTrans_Transitive`?

Strata/DL/Lambda/Semantics.lean
```diff
@@ -62,20 +81,27 @@ inductive Step (F:@Factory Tbase) (rf:Env Tbase)
     rf x = .some e →
     Step F rf (.fvar m x ty) e
 
-/-- Call-by-value semantics: beta reduction. -/
+/-- Beta reduction. The argument `e2` need not be a canonical value;
+this relaxation (compared to strict call-by-value) is necessary because `LExpr.eval`
+evaluates both sub-expressions before performing substitution and `LExpr.eval`
+might run out of fuel, making the reduced argument still non-value. -/
```
joscoh: Very minor nit: I don't think we should give the reason as the evaluator running out of fuel. I think it is better to say something like "we want the semantics to be flexible enough to handle various evaluation strategies, including partial evaluation".

Strata/DL/Lambda/Semantics.lean
```diff
@@ -129,11 +155,14 @@ inductive Step (F:@Factory Tbase) (rf:Env Tbase)
     Step F rf e1 e1' →
     Step F rf (.eq m e1 e2) (.eq m' e1' e2)
 
-/-- Evaluation of the right-hand side of an equality. -/
+/-- Evaluation of the right-hand side of an equality.
+Note: this rule does NOT require the LHS to be a canonical value.
+This relaxes the strict call-by-value discipline, analogous to how
+`reduce_2` was relaxed for application arguments. -/
```
joscoh: I don't think this should mention things being relaxed. Comments should be stateless.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -159,6 +188,44 @@ order. -/
...
 
+/-- Substitute a free variable under an abstraction binder. This is analogous to
+the closure rule in lambda calculi with explicit substitutions: it substitutes
+all occurrences of a free variable in the body of an abstraction, even though the
+substitution is not represented as an explicit syntactic term.
+
+This rule is needed because `evalCore` substitutes free variables under binders (via
+`substFvarsFromState`), and the step relation must be able to replicate that in
+order to produce an expression whose post-substitution `eraseMetadata` matches
+the eval result.
```
joscoh: Again, I don't think we should explain the semantics by saying they are necessary to prove the evaluator correct. That reasoning feels a bit circular, even if it is what happened. (By this I mean I don't think we need this paragraph at all, except maybe again to say we want to allow flexible evaluation strategies)

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
   intro H
   contradiction
 
+-- Note: step_abs_stuck and step_quant_stuck are no longer true because of
```
joscoh: Again, comments should be stateless

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+@[expose] def StepStar (F:@Factory Tbase) (rf:Env Tbase)
   : LExpr Tbase.mono → LExpr Tbase.mono → Prop :=
   ReflTrans (Step F rf)
 
+---------------------------------------------------------------------
+
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] in
+theorem StepStar_ite_cond (F : @Factory Tbase) (rf : Env Tbase)
+    (c c' t f : LExpr Tbase.mono) (m : Tbase.Metadata)
+    (h : ReflTrans (Step F rf) c c') :
```
joscoh: Is there a reason this refers to `ReflTrans Step` and not `StepStar`?

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+
+---------------------------------------------------------------------
+
+-- Metadata-preserving ite/eq congruence lemmas (the existing StepStar_ite_*
+-- return ∃ m', but we need metadata preserved for substFvar composition).
+
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] in
+private theorem StepStar_ite_cond_pres (F : @Factory Tbase) (rf : Env Tbase)
+    (c c' t f : LExpr Tbase.mono) (m : Tbase.Metadata)
+    (h : ReflTrans (Step F rf) c c') :
```
joscoh: Same in these lemmas

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+    (h : ReflTrans (Step F rf) e2 e2') :
+    ReflTrans (Step F rf) (.eq m e1 e2) (.eq m e1 e2') := by
+  induction h with
+  | refl => exact ReflTrans.refl _
+  | step x y z hxy _ ih =>
+    exact ReflTrans.step _ (.eq m e1 y) _
+      (Step.eq_reduce_rhs (m' := m) e1 x y hxy) ih
+
+---------------------------------------------------------------------
+-- Structural lemma: getLFuncCall decomposes e into (op, args), and
```
joscoh: Is this comment cut off?

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+  | .fvar _ _ _ => simp only [getLFuncCall.go] at h; cases h; rfl
+  | .abs _ _ _ _ => simp only [getLFuncCall.go] at h; cases h; rfl
+  | .quant _ _ _ _ _ _ => simp only [getLFuncCall.go] at h; cases h; rfl
+  | .ite _ _ _ _ => simp only [getLFuncCall.go] at h; cases h; rfl
+  | .eq _ _ _ => simp only [getLFuncCall.go] at h; cases h; rfl
+  termination_by e.sizeOf
+
+
+---------------------------------------------------------------------
+-- getLFuncCall round-trips with mkApp for .op heads.
```
joscoh: Are these comments supposed to go somewhere else?

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+-- getLFuncCall.go on .op just returns (op, acc), so for
+-- mkApp m (.op ..) args = app .. (app (op) a1) .. an,
+-- getLFuncCall gives (.op .., args).
+
+-- getLFuncCall.go commutes with mkApp for "op-headed" expressions.
+-- An expression is op-headed if it's .op or .app _ (op-headed) _.
+-- For such expressions, getLFuncCall.go (mkApp m e rest) acc = getLFuncCall.go e (rest ++ acc).
+private inductive OpHeaded {T : LExprParamsT} : LExpr T → Prop where
+  | op : ∀ m name ty, OpHeaded (.op m name ty)
+  | app_op : ∀ m m' name ty a, OpHeaded (.app m (.op m' name ty) a)
```
joscoh: This case follows from the first and third cases.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+
+
+---------------------------------------------------------------------
+-- Helpers for canonical_value_not_step
+
+-- Helper: if callOfLFunc with partial app returns (op, args, f) with the
+-- canonical condition (isConstr || blt), and callOfLFunc with full arity
+-- also succeeds, then isConstr must be true.
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] [DecidableEq Tbase.IDMeta] [Inhabited Tbase.IDMeta] in
+private theorem callOfLFunc_partial_full_isConstr
```
joscoh, line 825: Wondering: is this theorem equivalent to saying that if `F.callOfLFunc e (allowPartialApp := false)` and with `allowPartialApp := true`, then the length of the returned `args` is `f.inputs.length`? I don't think this actually depends on anything about `isConstr` in particular

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+        | inl h => exact h
+        | inr h =>
+          simp [Nat.blt] at h
+          simp at h_eq
+          omega
+
+-- Helper: for e = .app m e1 e2, if callOfLFunc (with any allowPartialApp)
+-- succeeds with nonempty args, then e2 is the last arg and thus e2 ∈ args.
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] [DecidableEq Tbase.IDMeta] [Inhabited Tbase.IDMeta] in
+private theorem callOfLFunc_app_arg_mem
```
joscoh, line 860: Maybe you prove this elsewhere, but it seems generally useful to have a lemma like: `F.callOfLFunc e = some (op, args, f) -> e = mkApp op args` (or something, I don't remember the exact syntax. Then this lemma follows by inversion.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+-- For other e1, the first component is not .op in either case.
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] [DecidableEq Tbase.IDMeta] [Inhabited Tbase.IDMeta] in
+private theorem getLFuncCall_go_app_acc
+    (e1 e2 : LExpr Tbase.mono) (acc : List (LExpr Tbase.mono))
+    (op_m : Tbase.mono.base.Metadata)
+    (op_n : Identifier Tbase.mono.base.IDMeta)
+    (op_t : Option Tbase.mono.TypeType)
+    (args : List (LExpr Tbase.mono))
+    (m : Tbase.Metadata)
+    (h : getLFuncCall.go e1 ([e2] ++ acc) = (LExpr.op op_m op_n op_t, args)) :
```
joscoh, line 916: The formulation stated in the comment (`go (.app m e1 e2) acc = go e1 ([e2] ++ acc)`) seems more natural to me (e.g. for rewriting). What is the reason for stating the theorem this way?

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+      simp [LExpr.freeVars, ih1 h1, ih2 h2]
+    · simp at hc
+  | ite =>
+    simp [LExpr.isCanonicalValue, Factory.callOfLFunc, getLFuncCall, getLFuncCall.go] at hc
+  | eq =>
+    simp [LExpr.isCanonicalValue, Factory.callOfLFunc, getLFuncCall, getLFuncCall.go] at hc
+
+-- Canonical values are normal forms: no Step rule can fire on them.
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] in
+theorem canonical_value_not_step
```
joscoh, line 1115: Yay!

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+    have hi_args : i < args.length := by simp [stepped_args] at hi1; exact hi1
+    exact h_per_eM i hi_args
+  rw [hA, hB]
+
+---------------------------------------------------------------------
+-- getLFuncCall.go is eraseMetadata-invariant: if e₁.eM = e₂.eM and acc₁.map eM = acc₂.map eM,
+-- then getLFuncCall.go e₁ acc₁ and getLFuncCall.go e₂ acc₂ produce results with the same eM.
+set_option maxHeartbeats 800000 in
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] [DecidableEq Tbase.IDMeta] [Inhabited Tbase.IDMeta] in
+private theorem getLFuncCall_go_eraseMetadata_congr
```
joscoh, line 1438: Not blocking, but the many, many theorems about things being erase-metadata invariant make me think that we should just be using an equivalence relation (like `EMEquiv`) with appropriate congruence lemmas. I don't know if Lean has a way to rewrite with arbitrary equivalence relations like Rocq does though.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+private theorem eql_eraseMetadata_eq
+    {Tbase : LExprParams}
+    [DecidableEq Tbase.IDMeta] [Inhabited Tbase.IDMeta]
+    (F : @Factory Tbase)
+    (e₁ e₂ e₁' e₂' : LExpr Tbase.mono)
+    (h₁ : e₁.eraseMetadata = e₁'.eraseMetadata)
+    (h₂ : e₂.eraseMetadata = e₂'.eraseMetadata) :
+    LExpr.eql F e₁ e₂ = LExpr.eql F e₁' e₂' := by
+  -- Helper: eqModuloMeta is eraseMetadata-invariant
+  have eqmod_pres : ∀ {a b a' b' : LExpr Tbase.mono},
```
joscoh: This seems like it should be a separate lemma

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+  -- e'.eraseMetadata = (.const mc c).eraseMetadata means e' must be a .const (by inspection).
+  simp [LExpr.eraseMetadata, LExpr.replaceMetadata] at h_eM
+  cases e' <;> simp [LExpr.replaceMetadata] at h_eM <;> try contradiction
+  rename_i mc' c'; subst c'
+  exact ⟨_, hstep⟩
+
+---------------------------------------------------------------------
+-- Bridging lemma: simultaneous substFvars with cons = sequential substFvar then substFvars.
+-- Requires v to be closed (no free variables).
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] [Inhabited Tbase.IDMeta] in
```
joscoh: I think that the rules in `Step` should use `substFvars` so that this lemma is not necessary (and probably neither is the value restriction). We can discuss further.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+      hv_l.eraseMetadata_eq hv_r.eraseMetadata_eq ih
+  | ite hv_c hv_t hv_e =>
+    simp only [LExpr.evalCore]
+    exact evalIte_eraseMetadata_congr n' σ _ _ _ _ _ _ _ _
+      hv_c.eraseMetadata_eq hv_t.eraseMetadata_eq hv_e.eraseMetadata_eq ih
+
+
+---------------------------------------------------------------------
+
+-- Helper for the factory-function branch of eval_eraseMetadata_invariant.
```
joscoh: Can you add a comment saying what this lemma is proving? I do not understand the theorem statement.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -168,12 +235,3178 @@ theorem step_const_stuck:
...
+  obtain ⟨e_final, h_step_final, h_eM_final⟩ := ih res_s
+  refine ⟨e_final, ReflTrans.trans h_step_e (ReflTrans.step _ _ _ h_step_eval h_step_final), ?_⟩
+  rw [h_eM_final]
+  exact eval_eraseMetadata_invariant σ res_s e'_ceval n hWF h_res_eM.symm
+
+
+---------------------------------------------------------------------
+-- Main theorem
+
+theorem eval_StepStar
```
joscoh, line 2896: Yay !!!!!

joscoh (commented):
Looks very good overall! Just some mostly minor comments

Strata/DL/Lambda/LExprEval.lean
aqjune-aws: It was a mistake; reverted

Strata/DL/Lambda/LExprEval.lean
aqjune-aws: It was related to the agent's usage of `fun_induction` while proving `eql_eraseMetadata_eq`. `fun_induction` is a tactic that works on a recursive function with match statements discriminating the arguments of the function.
The agent sticked to using `fun_induction` while proving the theorem, but it could not properly proceed at this match of pair case. I wanted to confirm whether the existence of label `_h1:` in the pair was problematic (`_h2` could be removed simply), but without `_h1:` proving termination of this function became hard.
I switched to this nested form and agent immediately proved this. :/ So this was to help writing/maintaining proofs by machines. I couldn't exactly understand why the agent found hardness in proving this case.

Strata/DL/Lambda/LExprWF.lean
aqjune-aws: Yeah, it was using non-membership.

In parallel I found a few lemmas that were unused after refactoring the Step rules to use `LExpr.substFvars` and this theorem was included. I simply removed this theorem for maintenance.

Strata/DL/Lambda/LExprWF.lean
aqjune-aws: Makes sense. :)

Strata/DL/Lambda/LState.lean
aqjune-aws: Hmm, I have no idea. :/ I removed it.

Strata/DL/Lambda/LState.lean
aqjune-aws: It was because `LExpr.substFvars_fvar_find` was used; This lemma became unused after the latest update, so I erased it.

Strata/DL/Lambda/Semantics.lean
aqjune-aws: I removed this (It is callOfLFunc_mkApp_op, but the theorem statement itself seems readable without this comment)

Strata/DL/Lambda/Semantics.lean
aqjune-aws: Yeah, this was oversight.

Strata/DL/Lambda/Semantics.lean
aqjune-aws: I am not familiar with the general rewriting tactics, but it sounds like something useful for this kind of stuffs.

Strata/DL/Lambda/Semantics.lean
aqjune-aws: The more generic property that relates when `allowPartialApp := true` and `allowPartialApp := false` is added and this lemma is updated to use that. This lemma is used deriving the particular `isConstr = true` fact.

Strata/DL/Lambda/Semantics.lean
aqjune-aws: The inversed version is `callOfLFunc_mkApp_op`, but indeed there isn't one yet. :) Might appear during the next type safety target...

Strata/DL/Lambda/Semantics.lean
aqjune-aws: There is `getLFuncCall_go_app_opHeaded` (I think the comment had to be placed over the theorem). This theorem is more readily applicable because it doesn't need `OpHeaded e`. Slightly updated the comment. :)

Strata/DL/Lambda/Semantics.lean
aqjune-aws: Removed. :)

Strata/DL/Util/Relations.lean
aqjune-aws: Removed

StrataTest/DL/Lambda/LExprEvalTests.lean
```diff
@@ -424,28 +492,7 @@ def test15 := TestCase.mk
...
-  case expand_fn =>
-    rename_i a a2 a3
-    cases a2
-    contradiction
-  case eval_fn =>
-    rename_i a a2 a3 he
-    cases a3
-    cases a2; unfold denoteInt at he; contradiction
--/
+-- TODO: stuck test15 — (~Int.Neg x) can't be evaluated since x is unresolvable
```
aqjune-aws, line 460: I couldn't write proofs for these stuck test cases. :( Marked as TODO...

Strata/DL/Lambda/LExprEval.lean
joscoh: I guess this is a limitation of Lean's `fun_induction` tactic then?

Strata/DL/Lambda/LExprWF.lean
joscoh: Can you update the comment?

Strata/DL/Lambda/Semantics.lean
joscoh: Unfortunately, it looks like Lean does not support this (yet?) https://leanprover-community.github.io/archive/stream/270676-lean4/topic/Rewriting.20congruent.20relations.html. I still think some of the proofs might be nicer with more widespread use of `EMEquiv`, but this is not blocking.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -1996,335 +1943,110 @@ private theorem step_to_const_via_IH
...
-        (StepStar_to_substFvarsFromState σ c hEnv hEnvAll))
-      (ReflTrans.trans
-        (StepStar_ite_then_pres _ _ (LExpr.substFvars c _) t (LExpr.substFvars t _) f m
-          (StepStar_to_substFvarsFromState σ t hEnv hEnvAll))
-        (StepStar_ite_else_pres _ _ (LExpr.substFvars c _) (LExpr.substFvars t _) f (LExpr.substFvars f _) m
-          (StepStar_to_substFvarsFromState σ f hEnv hEnvAll)))
+    (σ : LState Tbase) (m : Tbase.Metadata) (qk : QuantifierKind)
+    (name : String) (ty : Option LMonoTy)
+    (tr body : LExpr Tbase.mono) :
+    ReflTrans (Step σ.config.factory (Scopes.toEnv σ.state))
```
joscoh: This is still using `ReflTrans Step` rather than `StepStar`.

Strata/DL/Lambda/LExprEval.lean
aqjune-aws: I think it is a combination of how `fun_induction` case splits on a pair of discriminants & agents are familiar with that pattern. If `fun_induction` considered the match with multiple discriminants as simply sequentially nested matches each of which has only one discriminant, the problem would not have appeared.

Strata/DL/Lambda/Semantics.lean
joscoh: It might be possible to use quotients to do this, though I haven't used Lean quotients myself.

aqjune-aws: It sounds like something related too. I wonder Lean people have expected the quotient type could have been applied to the programming language world.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -160,21 +187,3081 @@ order. -/
...
 -/
-def StepStar (F:@Factory Tbase) (rf:Env Tbase)
+@[expose] def StepStar (F:@Factory Tbase) (rf:Env Tbase)
   : LExpr Tbase.mono → LExpr Tbase.mono → Prop :=
   ReflTrans (Step F rf)
 
+---------------------------------------------------------------------
+
+omit [DecidableEq Tbase.Metadata] in
+theorem StepStar_ite_cond (F : @Factory Tbase) (rf : Env Tbase)
```
MikaelMayer: `StepStar_ite_cond`, `StepStar_ite_then`, `StepStar_ite_else`, `StepStar_eq_lhs`, and `StepStar_eq_rhs` (the `∃ m'` versions) are never referenced anywhere — the `_pres` variants (which preserve metadata exactly) are strictly stronger and are the ones actually used. Consider either removing the `∃ m'` versions or making the `_pres` versions public and dropping the weaker ones, to avoid maintaining dead code.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -160,21 +187,3081 @@ order. -/
...
+
+omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] in
+private theorem eval_StepStar_factory_terminal
+    {Tbase : LExprParams}
+    [DecidableEq Tbase.Metadata] [DecidableEq Tbase.IDMeta]
+    [Inhabited Tbase.IDMeta] [ToFormat Tbase.Metadata] [ToFormat Tbase.IDMeta]
+    [Traceable LExpr.EvalProvenance Tbase.Metadata]
+    (σ : LState Tbase) (e : LExpr Tbase.mono) (n : Nat)
+    (_hWF : FactoryWF σ.config.factory)
+    (op_expr : LExpr Tbase.mono)
```
MikaelMayer, line 1382: `_hWF` is unused in this theorem (the underscore prefix confirms it). If it's not needed for the proof, it could be removed to avoid a redundant assumption.

Strata/DL/Lambda/Semantics.lean
```diff
@@ -160,21 +187,3081 @@ order. -/
...
+syntactic term.
 
-omit [DecidableEq Tbase.Metadata] [DecidableEq Tbase.Identifier] in
+The `x ∈ freeVars body` witness ensures the step only fires when the body
+actually has free variables. This preserves `canonical_value_not_step`:
+`isCanonicalValue` returns `true` for an abs only when it is *closed*
+(`freeVars e = []`), so a canonical abs has no free variables for this rule
+to fire on.
+-/
+| abs_subst_fvars:
```
MikaelMayer, line 202: The `rf = Scopes.toEnv σ.state` premise ties the abstract environment to a concrete `LState`. This is a pragmatic choice that makes the soundness proof go through, but it means `Step` is not purely defined over an abstract `Env` — it requires the environment to be the `toEnv` projection of some state. The PR description mentions this is "to support the current behavior of LExpr.lambda", which is fair. Just flagging for awareness: if the `Step` relation is ever used in contexts where `rf` is not a `Scopes.toEnv`, these rules won't fire. No action needed — this is a design note.

StrataTest/DL/Lambda/LExprEvalTests.lean
```diff
@@ -424,30 +457,9 @@ def test15 := TestCase.mk
...
```
MikaelMayer, line 460: The three `stuck` proofs (test15, test24, test25) that existed on `main` are replaced with TODO comments. This is justified because the relaxed `Step` rules (e.g., `reduce_2` no longer requiring the function to be canonical) mean these expressions are no longer stuck. Worth tracking these TODOs — the `stuck` property may still hold for the *fully reduced* forms (e.g., `((~Int.Add x) #-30)` for test15).

Strata/DL/Lambda/IntBoolFactory.lean
```diff
@@ -41,22 +41,80 @@ automatically.
...
+  cevalTy_eraseMetadata :
+    ∀ (T : LExprParams) (e1 e2 : LExpr T.mono),
+      LExpr.eraseMetadata e1 = LExpr.eraseMetadata e2 →
+      cevalTy T e1 = cevalTy T e2
+  /-- `mkConst` after `eraseMetadata` is independent of the metadata argument. -/
+  mkConst_eraseMetadata :
+    ∀ (T : LExprParams) (md1 md2 : T.Metadata) (v : ValTy),
+      LExpr.eraseMetadata (mkConst T md1 v) = LExpr.eraseMetadata (mkConst T md2 v)
+
+private theorem cevalTy_of_eraseMetadata_eq
```
MikaelMayer, line 55: Nit: `cevalTy_of_eraseMetadata_eq` is a useful generic helper. Its name suggests it's about `cevalTy` specifically, but it's actually a general lemma about any function `f : LExpr → Option V` that only inspects the `const` constructor. A name like `option_fn_of_eraseMetadata_eq` or a docstring would make its generality clearer.

Strata/DL/Lambda/Semantics.lean
aqjune-aws: Fixed

Strata/DL/Lambda/IntBoolFactory.lean
aqjune-aws: Indeed, it was named `cevalTy_of_eraseMetadata_eq` because this lemma is used to fill in the `cevalTy_eraseMetadata` field. :) Keeping the name...

StrataTest/DL/Lambda/LExprEvalTests.lean
aqjune-aws: I think it is right to be stuck because `x` doesn't exist in the register file. Indeed, these stuck cases are left as TODO.... Could not prove them. :/

Strata/DL/Lambda/Semantics.lean
aqjune-aws: Makes sense, this introduces usage of concrete `LState`. The `rf = Scopes.toEnv σ.state` still links the `rf` parameter with `σ`, but in the future we can even make `Step` only rely on `LState` for simpler rules (at the expense of using it slightly more complicated for PureExpr).

PR #740, opened 2026-04-02
Add type substitution for Factory function evaluation

joscoh:
*Issue #, if available:*

*Description of changes:* It was unsound to only substitute free variables without substituting type arguments when evaluating a Factory function body (in `Step` and `LExpr.eval`). This PR adds the type substitution mechanism to both `Step` and `eval` and includes a test demonstrating why this is necessary.


Strata/DL/Lambda/Factory.lean
```diff
@@ -257,6 +259,52 @@ theorem Factory.callOfLFunc_smaller {T} {F : @Factory T.base} {e : LExpr T} {op
...
+    -- Try the instantiated type on the .op node first
+    let opConstraints := match callee with
+      | .op _ _ (some instTy) =>
+        let genericTy := LMonoTy.mkArrow' fn.output (fn.inputs.values.reverse)
+        [(instTy, genericTy)]
+      | _ => []
+    -- Also unify argument types against formal parameter types
+    let argTys := args.filterMap (fun e => match e with
+      | .fvar _ _ ty => ty | .op _ _ ty => ty | .const _ c => some c.ty | _ => none)
+    let argConstraints := argTys.zip fn.inputs.values
```
MikaelMayer: `filterMap` drops arguments whose types can't be extracted, then `zip` pairs the remaining types positionally against `fn.inputs.values`. This can misalign if an argument in the middle is dropped.

Example: `args = [app_expr, const_true]`, `fn.inputs.values = [ty_a, ty_b]` → `argTys = [bool]` → `argConstraints = [(bool, ty_a)]`, incorrectly pairing argument 2's type with formal parameter 1.

Consider zipping `args` with `fn.inputs.values` first, then extracting types from the paired arguments:
```lean
let argConstraints := (args.zip fn.inputs.values).filterMap (fun (e, formalTy) =>
  match e with
  | .fvar _ _ (some ty) => some (ty, formalTy)
  | .op _ _ (some ty) => some (ty, formalTy)
  | .const _ c => some (c.ty, formalTy)
  | _ => none)
```
This preserves the positional correspondence.

Strata/DL/Lambda/LExprEval.lean
```diff
@@ -223,9 +223,14 @@ def eval (n : Nat) (σ : LState TBase) (e : (LExpr TBase.mono))
...
-          let new_e := substFvarsLifting body input_map
-          eval n' σ new_e
+          -- Apply type substitution to instantiate polymorphic type variables.
+          match LFunc.computeTypeSubst lfunc op_expr args with
+          | some tySubst =>
+            let body := body.applyTySubst tySubst
+            let input_map := lfunc.inputs.keys.zip args
+            let new_e := substFvarsLifting body input_map
+            eval n' σ new_e
+          | none => e -- cannot happen in well-typed terms
```
MikaelMayer, line 233: Silently returning the unevaluated expression on `none` could make debugging difficult if this case is ever reached unexpectedly. Consider logging a trace message here (e.g., `dbg_trace`) so that unexpected failures during evaluation are visible, rather than silently producing a stuck term.

Strata/DL/Lambda/Factory.lean
```diff
@@ -257,6 +259,52 @@ theorem Factory.callOfLFunc_smaller {T} {F : @Factory T.base} {e : LExpr T} {op
     intros op_eq args_eq F_eq
     subst op args F'; exact (getLFuncCall_smaller Hfunc)
 
+/--
+Apply type substitution `S` to all type annotations in an `LExpr`.
+-/
+def LExpr.applyTySubst {T : LExprParams} (e : LExpr T.mono) (S : Subst) : LExpr T.mono :=
```
aqjune-aws: Can we use `applySubst` in Strata/DL/Lambda/LExprT.lean instead, or is the behavior different from LExpr.applyTySubst?

joscoh: Good catch, I should be using that. I might need to move it around to avoid dependency issues though.

Strata/DL/Lambda/Factory.lean
```diff
@@ -257,6 +259,52 @@ theorem Factory.callOfLFunc_smaller {T} {F : @Factory T.base} {e : LExpr T} {op
...
+  else
+    -- Try the instantiated type on the .op node first
+    let opConstraints := match callee with
+      | .op _ _ (some instTy) =>
+        let genericTy := LMonoTy.mkArrow' fn.output (fn.inputs.values.reverse)
+        [(instTy, genericTy)]
+      | _ => []
+    -- Also unify argument types against formal parameter types
+    let argTys := args.filterMap (fun e => match e with
+      | .fvar _ _ ty => ty | .op _ _ ty => ty | .const _ c => some c.ty | _ => none)
```
joscoh: Note: this part is wrong, I need to fix it.

Strata/DL/Lambda/LExprEval.lean
joscoh: I can add a `dbg_trace` if you prefer, but it's not a failure so much as evaluation being unable to continue. It's just like how the evaluator just returns the input when trying to evaluate `.app .true .true` or something.

Strata/DL/Lambda/Factory.lean
joscoh: Fixed

Strata/DL/Lambda/Factory.lean
```diff
@@ -257,6 +259,59 @@ theorem Factory.callOfLFunc_smaller {T} {F : @Factory T.base} {e : LExpr T} {op
...
+Returns `none` if the type substitution cannot be derived.
+-/
+def LFunc.computeTypeSubst {T : LExprParams} (fn : LFunc T) (callee : LExpr T.mono)
+    (args : List (LExpr T.mono)) : Option Subst :=
+  if fn.typeArgs.isEmpty then some Subst.empty
+  else
+    -- Try the instantiated type on the .op node first
+    let opConstraints := match callee with
+      | .op _ _ (some instTy) =>
+        let genericTy := LMonoTy.mkArrow' fn.output (fn.inputs.values.reverse)
```
tautschnig:
```suggestion
        let genericTy := LMonoTy.mkArrow' fn.output fn.inputs.values
```
`mkArrow'` takes arguments in left-to-right order and builds `arg1 -> arg2 -> ... -> ret`. Since `fn.inputs.values` is already in declaration order (matching the curried application order from `getLFuncCall`), the reverse produces a generic type with swapped parameter types.

Add a test with distinct type parameters to catch this. For example, a function `f<a, b>(x: a, y: b) : bool` called as `f<int, string>(42, "hello")`.

joscoh: Thanks, fixed

Strata/DL/Lambda/Factory.lean
aqjune-aws: Let me dive into this further, because this is adding another trusted function to `Step`.

joscoh: For now yes, but this will not need to be trusted when we have the denotational semantics.

PR #741, opened 2026-04-02
Call elimination: requires violation diagnostic should point to call site

### LINKED ISSUE #531: Call elimination: requires violation diagnostic should point to call site
## Problem

When a `requires` precondition is violated at a call site, the diagnostic points to the `requires` expression (the callee's definition), not the call site where the violation occurs.

Laurel example:
```
procedure takesPositive(n: int) returns (r: int)
  requires n > 0
{
  return n;
}
procedure caller() returns (r: int) {
  var x: int := takesPositive(-1);
  return x;
}
```

The error points to `n > 0` on line 2 (the `requires` clause) rather than `takesPositive(-1)` on line 7.

## Root cause

Core's call elimination (`CallElim`) inlines the contract at each call site by generating `assert requires_expr`. The metadata on the generated assert comes from the `requires` expression, so the diagnostic points to where the requires was defined, not where it was violated.

## Proposed fix

In `CallElim`, when generating the assert for a requires check at a call site, use the call site's metadata instead of (or in addition to) the requires expression's metadata.

MikaelMayer:
Fixes #531

## Problem

During call elimination, the generated asserts for `requires` preconditions used only the call site metadata (`md`), discarding the requires clause's own metadata. This meant origin-based labels and the requires clause's file range were lost, producing generic diagnostic names like `callElimAssert_requires_0_4` and no link back to the original requires clause location.

## Fix

In `createAsserts` and `createAssumes`, use `check.md.setCallSiteFileRange md` instead of `md` alone. This:
- Preserves the requires/ensures clause's metadata (including origin labels) as the base
- Sets the call site as the primary file range (which was already the reported location)
- Adds the original clause file range as a related file range for richer SARIF diagnostics

## Testing

All existing tests pass. The correctness proofs (`CallElimCorrect`) compile without changes. A SARIF output test verifies that related locations are emitted correctly.


Strata/Transform/CoreTransform.lean
```diff
@@ -265,7 +268,8 @@ def createAssumes
     conds.mapM (fun (l, check) => do
       let newLabel ← genIdent l (fun s => s!"callElimAssume_{s}")
       -- Non-lifting: the replacement expressions must be closed (no dangling bvars).
-      return Statement.assume newLabel.toPretty (Lambda.LExpr.substFvars check.expr subst) md)
+      let assumeMd := check.md.setCallSiteFileRange md
```
MikaelMayer, line 275: Minor: `createAsserts` has a two-line comment explaining why `setCallSiteFileRange` is used (lines 256–257), but `createAssumes` doesn't. Since both functions apply the same transformation for the same reason, it would be nice to add the same comment here for consistency — or alternatively, move the explanation to the docstrings of both functions so the intent is visible at a higher level.

MikaelMayer: Good catch. Added the matching two-line comment to `createAssumes`, using "ensures clause" instead of "requires clause" since assumes correspond to postconditions.

StrataTest/Languages/Python/expected_laurel/test_class_methods.expected
```diff
@@ -5,9 +5,9 @@ test_class_methods.py(24, 4): ✅ pass - assert_assert(539)_calls_Any_to_bool_0
 test_class_methods.py(24, 4): ❌ fail - assert(539)
 test_class_methods.py(28, 4): ✅ pass - assert_assert(654)_calls_Any_to_bool_0
 test_class_methods.py(28, 4): ❌ fail - assert(654)
-test_class_methods.py(30, 4): ✅ pass - callElimAssert_requires_0_4
-test_class_methods.py(30, 4): ✅ pass - callElimAssert_requires_1_5
-test_class_methods.py(30, 4): ✅ pass - callElimAssert_requires_2_6
+test_class_methods.py(30, 4): ✅ pass - (Origin_test_helper_procedure_Requires)req_name_is_foo
```
MikaelMayer, line 8: I'm puzzled because this PR claims it moves the main error location of requirements from the requires clause to the call location and add the requirement location as related metadata, but the line and column number of this notification haven't changed.

MikaelMayer: You're right — the primary file range was already at the call site before this PR. The old code used `md` (the call site metadata) directly, so the line/column was already correct.

What this PR actually changes is the *base metadata*: instead of using `md` (call site metadata only), it now uses `check.md.setCallSiteFileRange md`, which:
1. Starts from `check.md` (the requires clause metadata, carrying origin labels and other metadata)
2. Replaces its file range with the call site's file range (which was already the location being used)
3. Preserves the requires clause's original file range as a related file range

The visible effect in these tests is the label names changing from generic `callElimAssert_requires_0_4` to origin-based `(Origin_test_helper_procedure_Requires)req_name_is_foo`. The related file range shows up in SARIF output (tested in `SarifOutputTests.lean`).

I'll update the PR description to accurately reflect this.

MikaelMayer: Ok much better. Now here is something new: The error messages with the assertion labels are much less useful than stating that there is another location at which the original requires clause is placed.
I wish we have something like this:
```
test_class_methods.py(30, 4): ✅ pass - inlined from test_class_methods.py(24,8)
```
or whatever correct line and column number it should be.

PR #742, opened 2026-04-03
Replace Factory linear-scan with O(1) HashMap lookup

joehendrix:
Replace Factory's linear-scan lookup with O(1) HashMap-backed structure, eliminating `getFactoryLFunc` which profiling showed consuming ~53% of execution time during partial evaluation.

- Change `Factory` from a transparent `Array (LFunc T)` type alias (with
  `Array.find?` for lookup) to an opaque structure wrapping an `Array`
  plus a `Std.HashMap String Nat` index, with invariants `toArrayDefined`
  and `nameMapValid` ensuring the map is consistent with the array.
- Provide `get`, `get?`, `push`, `pushIfNew`, `appendAll`, `tryPush`
  (replacing `addFactoryFunc`), and `ofArray` as the new Factory API;
  hook into Lean's `GetElem?` typeclass for `F[name]?` syntax.
- Prove key lemmas: `name_nodup` (function names are unique, via
  `toArrayDefined` injectivity), `push_mem_match`, `ofArray_mem`, and
  `getElem?_is_some_implies_mem`.
- Update all downstream call sites across type checking, partial
  evaluation, SMT encoding, precondition elimination, and factory
  construction to use the new API.
- Remove the `addFactoryFunc_wf`/`addFactory_wf` proofs from
  `FactoryWF.lean` (they referenced the old API); the simpler
  `WFLFactory.toFactory_wf` and `FactoryWF` structure (now just
  `lfuncs_wf`) cover the needed well-formedness guarantees.
- Simplify `Core.Factory_wf` from a manual proof to `by simp [Factory]`.

## Test changes

- Update test files to use the new API (`Factory.ofArray`, `.pushIfNew`,
  `.tryPush`, `Factory.toArray`, `.getFunctionNames`).
- Disable `steps_well` and `stuck` proofs in `LExprEvalTests.lean` that
  relied on unfolding the old transparent `Factory` definition; these
  need rework to use the opaque structure. Marked with `FIXME`.

### DIFF
--- a/Strata/DL/Imperative/SMTUtils.lean
+++ b/Strata/DL/Imperative/SMTUtils.lean
@@ -22,10 +22,13 @@ public section
 ---------------------------------------------------------------------
 
 /--
-A counterexample derived from an SMT solver is a map from an identifier
+A model derived from an SMT solver is a map from an identifier
 to an `SMT.Term`.
 -/
-@[expose] abbrev CounterEx (Ident : Type) := Map Ident Strata.SMT.Term
+@[expose] abbrev Model (Ident : Type) := Map Ident Strata.SMT.Term
+
+/-- Backward-compatible alias. -/
+@[deprecated Model (since := "2026-04-03")] abbrev CounterEx := @Model
 
 /-- Render an `SMT.Term` to a string via the SMTDDM translation. -/
 private def termToString (t : Strata.SMT.Term) : String :=
@@ -33,24 +36,24 @@ private def termToString (t : Strata.SMT.Term) : String :=
   | .ok s => s
   | .error _ => repr t |>.pretty
 
-def CounterEx.format {Ident} [ToFormat Ident] (cex : CounterEx Ident) : Format :=
-  match cex with
+def Model.format {Ident} [ToFormat Ident] (m : Model Ident) : Format :=
+  match m with
   | [] => ""
   | [(id, v)] => f!"({id}, {termToString v})"
   | (id, v) :: rest =>
-    (f!"({id}, {termToString v}) ") ++ CounterEx.format rest
+    (f!"({id}, {termToString v}) ") ++ Model.format rest
 
-instance {Ident} [ToFormat Ident] : ToFormat (CounterEx Ident) where
-  format := CounterEx.format
+instance {Ident} [ToFormat Ident] : ToFormat (Model Ident) where
+  format := Model.format
 
 /--
 Result from an SMT solver.
 -/
 inductive Result (Ident : Type) where
   -- Also see Strata.SMT.Decision.
-  | sat (cex : CounterEx Ident)
+  | sat (model : Model Ident)
   | unsat
-  | unknown (candidateModel : Option (CounterEx Ident) := none)
+  | unknown (candidateModel : Option (Model Ident) := none)
   | err (msg : String)
   deriving DecidableEq, Repr
 
@@ -174,7 +177,7 @@ value in the model.
 private def processModel {P : PureExpr} [ToFormat P.Ident]
     (typedVarToSMTFn : P.Ident → P.Ty → Except Format (String × Strata.SMT.TermType))
     (vars : List P.TypedIdent) (pairs : List (String × Strata.SMT.Term))
-    (E : Strata.SMT.EncoderState) : Except Format (CounterEx P.Ident) := do
+    (E : Strata.SMT.EncoderState) : Except Format (Model P.Ident) := do
   match vars with
   | [] => return []
   | (var, ty) :: vrest =>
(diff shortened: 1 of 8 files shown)


StrataTest/DL/Lambda/LExprEvalTests.lean
```diff
@@ -286,13 +286,14 @@ def test8 := TestCase.mk
 #guard_msgs in
 #eval check test8
 
+/- FIXME: Disabled when Factory made private
```
tautschnig, line 289: 16 proofs are disabled here with `FIXME: Disabled when Factory made private`. These `steps_well` and `stuck` proofs verified that the small-step operational semantics correctly evaluates/gets stuck on specific test cases. They relied on unfolding the old transparent `Factory` definition.

Since `Factory` is now an opaque structure, these proofs need a different approach — likely using the `Factory.get?`/`push_mem_match` lemmas instead of `simp`/`unfold`. Could you either:
- Restore them using the new API, or
- File a tracking issue so they don't get forgotten?

Strata/DL/Lambda/Factory.lean
```diff
@@ -147,43 +143,205 @@ IDMeta)` -- lambdas are our only tool. `Factory` gives us a way to add
...
-instance : Inhabited (@Factory T) where
-  default := @Factory.default T
-
-instance : Membership (LFunc T) (@Factory T) where
-  mem x f := Array.Mem x f
-
-
-def Factory.getFunctionNames (F : @Factory T) : Array T.Identifier :=
-  F.map (fun f => f.name)
+structure Factory (T : LExprParams) where
```
tautschnig, line 148: `FactoryWF` (in `FactoryWF.lean`) dropped the `name_nodup` field since it's now a theorem on `Factory` itself, leaving just `lfuncs_wf`. Is there a reason to keep `FactoryWF` as a separate structure rather than folding `lfuncs_wf` into `Factory`? The current split means callers need both a `Factory` and a `FactoryWF` proof.

tautschnig (changes requested):
**1. Development artifacts committed.** `do` (empty file), `feature_plan.md`, `fix_resolution.md`, and `run.sh` are in the branch (introduced in merge commit 97535653). These contain local paths (`/local/home/joehx/...`) and should be removed before merging.

**2. 16 proofs disabled with FIXME.** `LExprEvalTests.lean` has 16 `steps_well`/`stuck` proofs commented out with `FIXME: Disabled when Factory made private`. These are small-step operational semantics proofs that verified the evaluator's correctness. Disabling them weakens the test coverage. A plan to restore them (or an argument for why they're no longer needed) should be provided. See inline comment.

**3. `LFunc.computeTypeSubst` and `LExpr.typeOf` removed.** These were deleted from `Factory.lean` and not moved elsewhere. `computeTypeSubst` was used in `LExprEval.lean` to apply type substitutions when inlining polymorphic function bodies — that call is also removed. Is this safe because the Lambda evaluator only operates on already-monomorphized terms? If so, a brief comment in `LExprEval.lean` (around line 226) explaining why type substitution is unnecessary would help future readers.

StrataTest/DL/Lambda/LExprEvalTests.lean
joehendrix: I'm not quite sure the best solution, but I don't think we can go back to just using alists for this if our goal is an industrially relevant system.

@aqjune-aws is looking into fixing the proofs and have more to say.

Strata/DL/Lambda/Factory.lean
joehendrix: I think this is a good observation.  However, the current split predates this PR, and I'd prefer not to increase the scope of this PR to refactor that as well.

StrataTest/DL/Lambda/LExprEvalTests.lean
tautschnig: Ack, perhaps filing a tracking issue is just the right way to go!

Strata/DL/Lambda/Factory.lean
tautschnig: Would you mind creating a tracking issue so that we can get that done at some later point? (Unless you think it should just not be refactored and be left as is.)

joehendrix: I don't have a strong opinion as I'm not super familiar with this code.  In addition to the FactoryWF type, there's a separate class called `WFLFactory` that bundles the well-formedness proofs and also specializes it to `LExprParams`.  I created that since I do think bundling the Factory and well-formedness proofs makes sense in most use cases.

My gut instinct is to say this code is overly generic and could be specialized to be more concrete, but at the moment, I'm not familiar with the design and more concerned with performance.

tautschnig: I'd vote for creating an issue that links to this conversation so that we can get back to the topic whenever we can make it a priority.

StrataTest/DL/Lambda/LExprEvalTests.lean
aqjune-aws: Filed a tracking issue here: https://github.com/strata-org/Strata/issues/751 Since #738 also touches these tests, I will include the fix in there.

tautschnig (approved):

Strata/DL/Lambda/Factory.lean
aqjune-aws: To me keeping `FactoryWF` seems useful in terms of brevity when stating theorems that assume this. The unrolled form `∀ (lf:LFunc T), lf ∈ fac → LFuncWF lf` needs scrutiny because there could be typo, say `lf'` in the RHS of implication instead of `lf`. Since `FactoryWF` is typically used as an assumption of theorem, reducing possibility of typo is useful.

PR #748, opened 2026-04-03
LoopElim should be a phase, not built in into the partial evaluator

### LINKED ISSUE #747: LoopElim should be a phase, not built in into the partial evaluator.
Currently loop elimination is a phase that happens during partial evaluation. This has the following disadvantages:

- It is harder to prove that it is sound
- It is not composable - perhaps we want to make more optimizations on it like path pruning, but since the PE is the last phase we can't insert new steps.
- It is not replaceable. We will want to sometimes unroll a loop instead of abstracting it, as it's another valid proof strategy that does not require invariants if the loop is bounded.
- PR #722 relies on each phase being independent to ensure the soundness of SAT. Because of this tight coupling, it has a hack that introduces a phantom transformation to validate models, which is error prone as the ordering of phases likely matter.

Ideally, partial evaluation should be able to ignore anything beyond and including a loop if we can guarantee we eliminated loops before

MikaelMayer:
Loop elimination was tightly coupled with the partial evaluator (`Procedure.eval`), making it harder to prove soundness, compose with other optimizations, or replace with alternative strategies (e.g., loop unrolling).

This PR extracts loop elimination into a standalone program transformation phase (`Core.loopElim`) that runs after other transforms but before type checking and partial evaluation. The phase is integrated into the `PipelinePhase` system so that:

- `loopElimPipelinePhase.transform` performs the actual loop elimination
- `loopElimPipelinePhase.phase.getValidation` demotes SAT models to unknown for obligations whose path includes loop-elimination labels, since loop elimination is an over-approximation
- The transform and its soundness annotation are coupled in the same `PipelinePhase`

Both verification paths (all procedures and targeted verification) now use the same `corePipelinePhases` list to run transforms, eliminating duplicated manual transform logic. Each phase determines its own applicability via the phase list construction.

The partial evaluator no longer needs to know about loops — it expects them to be eliminated beforehand.

All existing tests pass.

Fixes #747


Strata/Languages/Core/Verifier.lean
```diff
@@ -811,6 +820,8 @@ def verify (program : Program)
   let axiomCache? ← profileStep profile "  Build axiom relevance cache" do
     pure (if options.removeIrrelevantAxioms == .Off then .none
           else .some (IrrelevantAxioms.Cache.build finalProgram))
+  -- Eliminate loops as a separate phase before partial evaluation.
```
MikaelMayer: The axiom relevance cache is built on `finalProgram` *before* `loopElim` runs (line 824). Since `loopElim` only rewrites procedure bodies and doesn't add or remove declarations, the cache remains valid — but it may be slightly over-conservative: call edges inside loop bodies that get replaced by assertions/assumptions will still appear in the cached call graph, potentially keeping axioms that are no longer referenced.

This is safe (never unsound), but if precision matters for axiom pruning, consider swapping the order so the cache is built after loop elimination.

Strata/Languages/Core/Verifier.lean
```diff
@@ -811,6 +820,8 @@ def verify (program : Program)
...
+  let finalProgram := loopElim finalProgram
```
MikaelMayer: Every other phase in `verify` is wrapped in `profileStep` (program transformations, axiom cache, type check + PE, VC discharge), but `loopElim` runs outside any profiling block. Consider wrapping it for consistency, e.g.:
```lean
let finalProgram ← profileStep profile "  Loop elimination" do
  pure (loopElim finalProgram)
```

Strata/Languages/Core/Verifier.lean
MikaelMayer: Good point — since `loopElim` only rewrites procedure bodies without adding or removing declarations, the axiom relevance cache built on `finalProgram` before `loopElim` remains valid. No change needed here.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Wrapped `loopElim` in `profileStep profile "  Loop elimination"` for consistency with all other phases in `verify`.

Strata/Languages/Core/Verifier.lean
```diff
@@ -757,6 +758,14 @@ def verifySingleEnv (pE : Program × Env) (options : VerifyOptions)
       let _ ← (IO.println s!"[profile]     Obligations: {E.deferred.size} total, {peResolvedCount} resolved by PE" |>.toBaseIO)
     return results
 
+/-- Eliminate loops in all procedures of a Core program by replacing each loop
+with assertions and assumptions about its invariants. -/
+def loopElim (p : Program) : Program :=
+  { decls := p.decls.map fun d => match d with
```
tautschnig: Nit: `loopElim` is a program-level transformation that could live in `Strata/Transform/LoopElim.lean` alongside the statement-level `removeLoopsM` it wraps. Placing it in `Verifier.lean` couples the transform definition to the verification pipeline. That said, it's only 4 lines and has exactly one other caller (`SimpleAPI.lean` and `MetaVerifier.lean` both reference it), so this is a minor organizational point — fine either way.

Strata/MetaVerifier.lean
```diff
@@ -56,6 +56,7 @@ def genVCsSingleENV (pE : Program × Env) : Option coreVCs := do
   | _ => return E.deferred.toList.map (fun ob => (E, ob))
 
 def genVCs (program : Program) (options : VerifyOptions := .default) : Option coreVCs := do
+  let program := loopElim program
```
tautschnig: Good catch adding `loopElim` here — `genVCs` calls `typeCheckAndPartialEval` directly (bypassing `verify`), so without this the meta-verifier would panic on programs with loops.

Strata/Languages/Core/Verifier.lean
```diff
@@ -674,9 +674,9 @@ def keepSetFilterPipelinePhase (procs : List String) : PipelinePhase :=
     All filter phases are model-preserving since they only remove
     information without introducing over-approximations.
 
-    `loopElimPipelinePhase` is placed last because loop elimination happens
-    during evaluation (not as a program-to-program pass), making it the
-    closest phase to SMT. -/
+    `loopElimPipelinePhase` is placed last because loop elimination
```
MikaelMayer: I think we should no longer justify loop elimination being placed last with such an argument. Later, we will have phases ensure that program satisfy some predicates, but for now there is no reason to justify loop elimination being placed last.

Strata/Languages/Core/Verifier.lean
```diff
@@ -912,12 +904,13 @@ def verify (program : Program)
   let factory ← EIO.ofExcept (Core.Factory.addFactory moreFns)
   let phases := coreAbstractedPhases (procs := proceduresToVerify) (factory := some factory)
   let finalProgram ← profileStep profile "  Program transformations" do
-    let runPrecondElim := fun prog => do
-      let (_changed, prog) ← PrecondElim.precondElim prog factory
-      return prog
     match proceduresToVerify with
     | none =>
-      match Transform.run program runPrecondElim with
+      let runTransforms := fun prog => do
```
MikaelMayer: We should have a list of phases and run them. The phases themselves can determine if they can apply or not. That way we can later do profiling, reordering as needed.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Removed the justification paragraph from the docstring. The comment no longer argues for loop elimination placement.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Both code paths now use `corePipelinePhases` to build and run the phase list. The `verify` function no longer has separate transform logic for the `none` vs `some procs` cases — it iterates over the pipeline phases uniformly. `callElimPipelinePhase` is now conditional on `procs` being provided, since call elimination only applies to targeted verification.

Strata/Transform/LoopElim.lean
```diff
@@ -187,13 +188,24 @@ def Stmt.removeLoops
...
+/-- Eliminate loops in all procedures of a Core program by replacing each loop
+    with assertions and assumptions about its invariants. -/
+def loopElim (p : Program) : Program :=
+  { decls := p.decls.map fun d => match d with
+    | .proc proc md =>
+      .proc { proc with body := (StateT.run (Block.removeLoopsM proc.body) 0).fst } md
+    | other => other }
+
+/-- Loop elimination as a `CoreTransformM` pass suitable for the pipeline. -/
+def loopElim' (p : Program) : Transform.CoreTransformM (Bool × Program) :=
```
tautschnig, line 200: `loopElim'` unconditionally returns `true` for the changed flag. If no procedures contain loops, the program is unchanged but downstream phases can't detect this. Consider returning `false` when no loops were found — `loopElim` could return a `(Bool × Program)` pair. Minor optimization, not blocking.

MikaelMayer: I think that the only reason for returning a changed booleans is to make sure to use the newest program. I'll tackle this as an optimization later.

PR #750, opened 2026-04-03
Fix Python->Laurel translation for Try statement

thanhnguyen-aws:
## Summary

Fixes the Python→Laurel translation for try/except statements and makes several soundness improvements to the Python runtime model. Previously, exception handlers merged all handler bodies into a single block; this PR replaces that with type-based dispatch. It also changes how exceptions propagate outside try blocks, removes unsound `requires` preconditions from runtime functions, and adds string indexing/slicing support.

## Exception handling: try/except translation

**Before:** All exception handler bodies were concatenated into one block, ignoring which exception type each handler catches.

**After:** Exception handlers are translated into a nested if-else chain that checks the exception type and dispatches to the correct handler:
```
if Error..isTypeError(maybe_except) then { handler1 }
else if Error..isIndexError(maybe_except) then { handler2 }
else assert Error..isNoError(maybe_except)
```

- `errTyToGuard` maps Python exception type names to Laurel Error tester predicates
- `exceptHandlerToBlock` / `exceptHandlersToIfStmt` build the dispatch chain
- The `as e` variable in `except TypeError as e:` is now bound (via `e.coreError := maybe_except`)
- `PythonError` composite type extended with a `coreError` field for this binding
- `else` and `finally` blocks throw `unsupportedConstruct` errors instead of being silently ignored

## Exception propagation: return instead of assert

**Before:** Outside try blocks, exception checks were `assert !Any..isexception(expr)` — a proof obligation that the verifier must discharge.

**After:** Exception checks are `if Any..isexception(expr) then return expr` — the exception value propagates to the caller. Inside try blocks, exceptions are captured into `maybe_except` for handler dispatch.

This is controlled by `ctx.isInsideTryBlock`: outside try → `getExceptionReturns`; inside try → `getExceptionAssign`.

### Removed unsound `requires` preconditions
- `List_get`, `List_get_non_neg`: removed bounds-check preconditions; now return `exception(IndexError(...))` for out-of-bounds access
- `DictStrAny_get`: removed `DictStrAny_contains` precondition; now returns `exception(IndexError(...))` for missing keys
- `Any_to_bool`: removed type-check precondition; fallback changed from `false` to `<?>` (sound over-approximation)
- `PFloorDiv`, `PMod`: removed division-by-zero preconditions; now return `exception(ZeroDivisionError())`
- `PAnd`, `POr`: removed type-check preconditions; now return `exception(UndefinedError(...))` for unhandled types

### New Error constructors
- `ZeroDivisionError()` — for division/modulo by zero
- `UnknownError()` — catch-all for unrecognized exception types

### String indexing and slicing
- `str_get(s, i)` — character access with negative index support
- `str_slice(s, start, stop)` — substring with negative index normalization
- `Str.Substr` declared as external function
- `Any_get!` extended to handle string subscript (`s[i]`)
- `Any_get_slice!` extended to handle string slicing (`s[start:stop]`)

### Error type corrections
- `UndefinedError` → `TypeError` for type mismatch errors in arithmetic/comparison operators (matches Python's actual exception hierarchy)

### Tuple translation
- `Tuple` literals now translate to `ListAny` (previously returned `Hole`)

## Test changes

- New `test_try_catch_multiple_exceptions.py` with three try/except scenarios
- Promoted 8 tests from `pending/` to active: `test_bool_shortcircuit_and/or`, `test_negative_index_string`, `test_str_index`, `test_str_slice`, `test_try_except_catch`, `test_try_except_multi`, `test_while_true_break`


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1146,8 +1152,70 @@ def createVarDeclStmtsAndCtx (ctx : TranslationContext) (newDecls : List (String
...
+    mkStmtExprMd $ .Assign [maybeExceptVar] (mkStmtExprMd $ .StaticCall "Any..get_error!" [mbe])))
+  (createIfStmt condBlock none).toList
+
+partial def getExceptionCatch (ctx: TranslationContext) (e : StmtExprMd): List StmtExprMd :=
+  let maybeExceptExprs := getMaybeExceptionExprs e
+  if ctx.isInsideTryBlock then getExceptionAssign maybeExceptExprs else getExceptionAssertions maybeExceptExprs
+
+def errTyToGuard (errTy: String) : StmtExprMd := match errTy with
+  | "TypeError" => mkStmtExprMd (.StaticCall "Error..isTypeError" [maybeExceptVar])
+  | "ZeroDivisionError" => mkStmtExprMd (.StaticCall "Error..isUndefinedError" [maybeExceptVar])
```
ssomayyajula: Can we add a TODO comment here to make sure we don't miss translating this properly the next time around?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1146,8 +1152,70 @@ def createVarDeclStmtsAndCtx (ctx : TranslationContext) (newDecls : List (String
...
+  | "Exception" => mkStmtExprMd (.StaticCall "isError" [maybeExceptVar])
+  | _ => mkStmtExprMd (.StaticCall "Error..isUnknownError" [maybeExceptVar])
+
 mutual
 
+partial def exceptHandlerToBlock
+        (ctx: TranslationContext)
+        (h : Python.excepthandler SourceRange) : Except TranslationError (StmtExprMd × StmtExprMd) := do
+  match h with
+  | .ExceptHandler sr errorTy _ body =>
```
ssomayyajula: Just occurred to me that the variable bound by an exception handler is not assigned to an exceptional value (`maybeExceptVar` ?) - I am guessing you are leaving that for later?

ssomayyajula (commented):
Thanks for the PR. Some questions:

- Should we emit some kind of diagnostic if a try-else or try-else-finally block is hit? 
- Can you explain the difference of use between `Any_get` and `Any_get!AnyMaybeExcept`?

tautschnig:

StrataTest/Languages/Python/run_py_analyze_sarif.py
```diff
@@ -55,6 +55,7 @@
     "test_regex_positive",
     "test_return_types",
     "test_subscription",
+    "test_try_catch_multiple_exceptions"
```
tautschnig: **Bug:** Missing trailing comma. In Python, adjacent string literals are implicitly concatenated, so:
```python
"test_try_catch_multiple_exceptions"
"test_try_except",
```
becomes the single string `"test_try_catch_multiple_exceptionstest_try_except"`, which matches neither test name. This silently removes `test_try_except` from the skip list.

```suggestion
    "test_try_catch_multiple_exceptions",
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1271,22 +1272,66 @@ partial def getMaybeExceptionExprs (ctx : TranslationContext) (e : StmtExprMd) :
...
+    mkStmtExprMd $ .Assign [maybeExceptVar] (mkStmtExprMd $ .StaticCall "Any..get_error!" [mbe])))
+  (createIfStmt condBlock none).toList
+
+partial def getExceptionCatch (ctx: TranslationContext) (e : StmtExprMd): List StmtExprMd :=
+  let maybeExceptExprs := getMaybeExceptionExprs ctx e
+  if ctx.isInsideTryBlock then getExceptionAssign maybeExceptExprs else getExceptionAssertions maybeExceptExprs
+
+def errTyToGuard (errTy: String) : StmtExprMd := match errTy with
+  | "TypeError" => mkStmtExprMd (.StaticCall "Error..isTypeError" [maybeExceptVar])
+  | "ZeroDivisionError" => mkStmtExprMd (.StaticCall "Error..isUndefinedError" [maybeExceptVar])
```
tautschnig: **Incorrect mapping:** `ZeroDivisionError` maps to `Error..isUndefinedError`. Python's `ZeroDivisionError` is a subclass of `ArithmeticError`, not related to `UndefinedError`. The `Error` datatype doesn't have a `ZeroDivisionError` constructor, so this is an approximation — but it should at least map to `UnknownError` rather than `UndefinedError`, which has completely different semantics.

Also, this addresses ssomayyajula's unresolved Thread 1: the wildcard `_` case silently maps any unrecognized exception type to `UnknownError`. Many common Python exceptions are missing (`ValueError`, `KeyError`, `RuntimeError`, `AttributeError`, `NameError`, `FileNotFoundError`, etc.). At minimum, add a `-- TODO: extend Error datatype with more Python exception types` comment.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1271,22 +1272,66 @@ partial def getMaybeExceptionExprs (ctx : TranslationContext) (e : StmtExprMd) :
...
+  let exceptionCheck := rhs_exprs.flatMap $ getExceptionCatch ctx
   (newctx, exceptionCheck ++ stmts)
 
 mutual
 
+partial def exceptHandlerToBlock
+        (ctx: TranslationContext)
+        (h : Python.excepthandler SourceRange) : Except TranslationError (StmtExprMd × StmtExprMd) := do
+  match h with
+  | .ExceptHandler sr errorTy _ body =>
```
tautschnig: **Unresolved feedback (ssomayyajula Thread 2):** The `name` field (the `as e` variable in `except TypeError as e:`) is matched with `_` and ignored. This means the handler body can reference `e` but it won't be bound to the exception value — any use of `e` in the handler will be an unresolved identifier.

At minimum, add a `-- TODO: bind the exception variable name to the caught exception value` comment. Better: bind it to `maybeExceptVar` (or the appropriate error message field) so handler bodies that reference the exception variable work correctly.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1421,17 +1466,10 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
   | .Try _ body handlers _ _ => do
```
tautschnig: The `else` and `finally` blocks are silently ignored (matched with `_ _`). The PR description acknowledges this limitation, but there's no diagnostic emitted when they're present. A user writing `try: ... except: ... finally: cleanup()` would get silently incorrect behavior.

Consider emitting an `unsupportedConstruct` error when `else` or `finally` is non-empty, similar to how other unsupported constructs are handled. This is safer than silent omission.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -488,6 +489,27 @@ function Any_get_slice (list: Any, index: Any): Any
...
+  else if (Any..isfrom_ListAny(dictOrList) && Any..isfrom_Slice(index) && Any..start!(index) >= 0 && Any..start!(index) < List_len(Any..as_ListAny!(dictOrList))) then
+    if OptionInt..isOptNone(Any..stop!(index)) then
+      from_ListAny(List_drop(Any..as_ListAny!(dictOrList), Any..start!(index)))
+    else if (OptionInt..isOptSome(Any..stop!(index))) &&  OptionInt..unwrap!(Any..stop!(index)) >= 0 &&
+            OptionInt..unwrap!(Any..stop!(index)) <= List_len(Any..as_ListAny!(dictOrList)) && Any..start!(index) <= OptionInt..unwrap!(Any..stop!(index)) then
+      from_ListAny(List_slice(Any..as_ListAny!(dictOrList), Any..start!(index), OptionInt..unwrap!(Any..stop!(index))))
+    else exception (IndexError("Invalid slice index"))
+  else
+    exception (IndexError("Invalid subscription"))
+};
```
tautschnig: The new `Any_get!` function with slice support is a significant addition that duplicates much of the logic from the original `Any_get!` (now renamed to `Any_get_no_slice!`). The two functions share the dict and list-by-index paths but differ in slice handling.

Consider whether the old `Any_get!` could simply be extended with the slice case rather than creating a separate function. The current approach means two nearly-identical functions that must be kept in sync.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1334,22 +1313,72 @@ partial def getMaybeExceptionExprs (ctx : TranslationContext) (e : StmtExprMd) :
...
   | _ => []
 
-partial def getExceptionAssertions (ctx : TranslationContext) (e : StmtExprMd) : List StmtExprMd :=
+partial def getExceptionReturns (maybeExceptExprs: List StmtExprMd) : List StmtExprMd :=
+  maybeExceptExprs.map (λ mbe => mkStmtExprMdWithLoc
+      (.IfThenElse
+        (mkStmtExprMd $ .StaticCall "Any..isexception" [mbe])
+        (mkStmtExprMd $ .Return mbe)
+        none)
+      mbe.md)
```
tautschnig: **Significant semantic change:** Exception handling outside try blocks changed from asserting no exception (`getExceptionAssertions`) to returning the exception value (`getExceptionReturns`). This models Python's exception propagation more accurately — if an operation raises, the function returns the exception to the caller.

This is a good change, but it's a major behavioral shift that affects every function in the pipeline. The PR description doesn't mention this change. It would be helpful to document the rationale, especially since it changes what the verifier checks (no more "assert no exception" obligations outside try blocks).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1511,20 +1540,17 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
   | .Import _ _ | .ImportFrom _ _ _ _ |.Pass _ => return (ctx, [])
 
   -- Try/except - wrap body with exception checks and handlers
-  | .Try _ body handlers _ _ => do
+  | .Try _ body handlers orelse final => do
+    if orelse.val.size > 0 then
+      throw (.unsupportedConstruct "Unsupport or_else block in Try statement" (toString (repr s)))
```
tautschnig: Nit: typo — "Unsupport" should be "Unsupported".

```suggestion
    if orelse.val.size > 0 then
      throw (.unsupportedConstruct "Unsupported or_else block in Try statement" (toString (repr s)))
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1334,22 +1313,72 @@ partial def getMaybeExceptionExprs (ctx : TranslationContext) (e : StmtExprMd) :
...
+  let exceptionCheck := rhs_exprs.flatMap $ getExceptionCatch ctx
   (newctx, exceptionCheck ++ stmts)
 
 mutual
 
+partial def exceptHandlerToBlock
+        (ctx: TranslationContext)
+        (h : Python.excepthandler SourceRange) : Except TranslationError (StmtExprMd × StmtExprMd) := do
+  match h with
+  | .ExceptHandler sr errorTy _ body =>
```
tautschnig: **Still unresolved (ssomayyajula Threads 1 & 2, my Thread 5):** The `name` field (the `as e` variable) is still matched with `_` and ignored. At minimum, add a `-- TODO: bind exception variable name to caught exception value` comment here so this doesn't get lost.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I updated the PR description.

tautschnig:
> 11 pending test(s) now pass. Move them from tests/pending/ to tests/ and generate expected files with --update.

@thanhnguyen-aws Can you please address those remaining ones?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1511,20 +1543,17 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
   | .Import _ _ | .ImportFrom _ _ _ _ |.Pass _ => return (ctx, [])
 
   -- Try/except - wrap body with exception checks and handlers
-  | .Try _ body handlers _ _ => do
+  | .Try _ body handlers orelse final => do
+    if orelse.val.size > 0 then
+      throw (.unsupportedConstruct "Unsupported or_else block in Try statement" (toString (repr s)))
+    if final.val.size > 0 then
+      throw (.unsupportedConstruct "Unsupport finalize block in Try statement" (toString (repr s)))
```
tautschnig: Nit: the `orelse` typo was fixed but the `finalize` one was missed — "Unsupport" should be "Unsupported".

```suggestion
      throw (.unsupportedConstruct "Unsupported finalize block in Try statement" (toString (repr s)))
```

thanhnguyen-aws:
> > 11 pending test(s) now pass. Move them from tests/pending/ to tests/ and generate expected files with --update.
> 
> @thanhnguyen-aws Can you please address those remaining ones?

I added the relevant ones. Other pending tests pass due to some soundness issues of the translation: https://github.com/strata-org/Strata/issues/962, https://github.com/strata-org/Strata/issues/962, https://github.com/strata-org/Strata/issues/977. I already created 2 PRs to fix them. Adding those pending tests now will cause more problems later as they are not actually solved by this PR.

tautschnig (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1721,6 +1758,7 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     let stmts ← match value.val with
       | some expr => do
         let e ← translateExpr ctx expr
+        let exceptionCheck := getExceptionCatch ctx e
```
tautschnig: **Dead code.** `let exceptionCheck := getExceptionCatch ctx e` is computed but never referenced — the `stmts` value built below uses `preamble` (from line 1762) and does not splice `exceptionCheck`.

Compare the `.Expr` arm at line 1805 where the analogous binding *is* used (`getMaybeExceptHavoc ++ exceptionCheck ++ [assignStmt] ++ ...`). Either:

1. Splice `exceptionCheck` into the returned `stmts` — seems more consistent with the other assignment-emitting arms; a returned expression that contains a potentially-exception-producing subcall should emit the exception-catch assignment/return on the caller's side.
2. Delete the binding if the `getExceptionCheckPreamble` call below already handles it (it returns a `preamble` that includes `getExceptionCatch` via the `containsUserCall` branch of its own logic — worth double-checking).

I think (1) is the right answer — today if `value` is `some_call() + 5` and `some_call()` throws, the current code returns the exception via the preamble path *only if `containsUserCall ctx e`*. For other exception-producing expressions (e.g. `a[i]` with out-of-bounds `i`), the catch is missed.

```suggestion
  -- Return statement: assign to the LaurelResult output parameter, then exit $body.
  | .Return _ value => do
    let stmts ← match value.val with
      | some expr => do
        let e ← translateExpr ctx expr
        let exceptionCheck := getExceptionCatch ctx e
        let (preamble, eRef) := getExceptionCheckPreamble ctx e s!"$ret_exc_{expr.toAst.ann.start.byteIdx}"
        -- Coerce Composite return values to Any for LaurelResult : Any
        let eRef ← coerceToAny ctx expr eRef
        let assign := mkStmtExprMdWithLoc (StmtExpr.Assign [mkVariableMd (.Local PyLauFuncReturnVar)] eRef) md
        .ok $ exceptionCheck ++ preamble ++ [assign, mkStmtExprMdWithLoc (StmtExpr.Exit "$body") md]
      | none => .ok [mkStmtExprMdWithLoc (StmtExpr.Exit "$body") md]
    return (ctx, stmts)
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1576,6 +1556,38 @@ partial def getMaybeExceptionExprs (ctx : TranslationContext) (e : StmtExprMd) :
...
+
+def errTyToGuard (errTy: String) : StmtExprMd := match errTy with
+  | "TypeError" => mkStmtExprMd (.StaticCall "Error..isTypeError" [maybeExceptVarExpr])
+  | "ZeroDivisionError" => mkStmtExprMd (.StaticCall "Error..isZeroDivisionError" [maybeExceptVarExpr])
+  | "IndexError" => mkStmtExprMd (.StaticCall "Error..isIndexError" [maybeExceptVarExpr])
+  | "AssertionError" => mkStmtExprMd (.StaticCall "Error..isAssertionError" [maybeExceptVarExpr])
+  | "AttributeError" => mkStmtExprMd (.StaticCall "Error..isAttributeError" [maybeExceptVarExpr])
+  | "Exception" => mkStmtExprMd (.StaticCall "isError" [maybeExceptVarExpr])
+  | _ => mkStmtExprMd (.StaticCall "Error..isUnknownError" [maybeExceptVarExpr])
+  -- TODO: extend Error datatype with more Python exception types
```
tautschnig: **Python exception hierarchy is not modeled.** This flat string match misses:

- `ArithmeticError` (superclass of `ZeroDivisionError`, `OverflowError`, `FloatingPointError`)
- `LookupError` (superclass of `IndexError`, `KeyError`)
- `OSError`, `FileNotFoundError`, `PermissionError`, ... (full OSError hierarchy)
- `NameError`, `ValueError`, `RuntimeError`, `StopIteration`, ... (common user-visible exceptions with no mapping here)
- `BaseException` (super of `Exception` — also catches `KeyboardInterrupt`, `SystemExit`; CPython users sometimes use this deliberately)

Today, a handler like `except ArithmeticError as e:` falls through the `_ =>` arm and maps to `Error..isUnknownError`, which no runtime function raises — so the handler never fires, and the `assert Error..isNoError` at the end of the if-chain fails. Valid Python code verifying a division by zero being caught by `ArithmeticError` will not verify.

Concrete fix sketch: route Python exception names through a table that encodes subclass relationships, then `or`-combine the matching runtime-error predicates. E.g.

```lean
/-- Python exception name → list of runtime `Error..isX` predicates that
    should route to this handler. Encodes CPython's class hierarchy. -/
def pythonErrorToPredicates (ty : String) : List String :=
  match ty with
  | "TypeError"           => ["Error..isTypeError"]
  | "ZeroDivisionError"   => ["Error..isZeroDivisionError"]
  | "ArithmeticError"     => ["Error..isZeroDivisionError" /- + OverflowError when added -/]
  | "IndexError"          => ["Error..isIndexError"]
  | "KeyError"            => [/- TODO: KeyError constructor when added -/]
  | "LookupError"         => ["Error..isIndexError" /- + isKeyError -/]
  | "AssertionError"      => ["Error..isAssertionError"]
  | "AttributeError"      => ["Error..isAttributeError"]
  | "Exception"           => ["isError"]
  | "BaseException"       => ["isError"]
  | _                     => []   -- unknown/unsupported: empty means never fires

def errTyToGuard (errTy : String) : StmtExprMd :=
  let preds := pythonErrorToPredicates errTy
  match preds with
  | []       => mkStmtExprMd (.Const (.ConBool false))  -- or emit a diagnostic at translation time
  | [p]      => mkStmtExprMd (.StaticCall p [maybeExceptVarExpr])
  | p :: ps  => ps.foldl (init := mkStmtExprMd (.StaticCall p [maybeExceptVarExpr]))
                  fun acc q => mkStmtExprMd (.PrimitiveOp .Or
                    [acc, mkStmtExprMd (.StaticCall q [maybeExceptVarExpr])])
```

(Exact `PrimitiveOp` / `StaticCall` spelling to be matched to Laurel's Bool OR — adjust for surface.) The table makes the hierarchy declarative and makes it obvious which Python exceptions the translator claims to support.

Test to add: `test_try_except_arithmetic_superclass.py` — `try: 1/0 except ArithmeticError: …`, which should reach the handler.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1576,6 +1556,38 @@ partial def getMaybeExceptionExprs (ctx : TranslationContext) (e : StmtExprMd) :
...
+  if ctx.isInsideTryBlock then getExceptionAssign maybeExceptExprs else getExceptionReturns maybeExceptExprs
+
+def errTyToGuard (errTy: String) : StmtExprMd := match errTy with
+  | "TypeError" => mkStmtExprMd (.StaticCall "Error..isTypeError" [maybeExceptVarExpr])
+  | "ZeroDivisionError" => mkStmtExprMd (.StaticCall "Error..isZeroDivisionError" [maybeExceptVarExpr])
+  | "IndexError" => mkStmtExprMd (.StaticCall "Error..isIndexError" [maybeExceptVarExpr])
+  | "AssertionError" => mkStmtExprMd (.StaticCall "Error..isAssertionError" [maybeExceptVarExpr])
+  | "AttributeError" => mkStmtExprMd (.StaticCall "Error..isAttributeError" [maybeExceptVarExpr])
+  | "Exception" => mkStmtExprMd (.StaticCall "isError" [maybeExceptVarExpr])
+  | _ => mkStmtExprMd (.StaticCall "Error..isUnknownError" [maybeExceptVarExpr])
```
tautschnig: **Silent swallow.** The `_ =>` arm maps every unrecognised Python exception name (`KeyError`, `ValueError`, `NameError`, `FileNotFoundError`, `RuntimeError`, `StopIteration`, user-defined exceptions, …) to `Error..isUnknownError`. No runtime function raises `UnknownError()`, so this handler path is *unreachable* by construction — the user wrote a handler that cannot fire, and we give them no warning.

Pick one of:

1. **Fail loudly** — `throw (.unsupportedConstruct s!"Unsupported exception type '{errTy}' in except clause" …)`. Matches how `else` / `finally` are handled elsewhere in this PR. Best for catching user errors early.
2. **Warn via `dbg_trace`** — still route to `isUnknownError`, but emit a trace so the user sees "this handler is inert".
3. **Gate with a known-good set** — maintain an explicit allow-list and only fall through to `UnknownError` for names in the list; anything outside the list is a translation error.

I'd prefer (1); the `unsupportedConstruct` machinery already exists and surfaces cleanly to the user. Concretely:

```suggestion
  | _ =>
    -- Python exception name we do not yet map — dbg_trace to surface to users
    -- so they know the handler will not fire. Alternatively, `throw` via the
    -- caller (needs making `errTyToGuard` Except-valued).
    dbg_trace s!"unsupported except type '{errTy}'; handler will not fire"
    mkStmtExprMd (.StaticCall "Error..isUnknownError" [maybeExceptVarExpr])
```

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -447,9 +447,8 @@ function DictStrAny_contains (d : DictStrAny, key: string) : bool
 };
 
 function DictStrAny_get (d : DictStrAny, key: string) : Any
-  requires DictStrAny_contains(d, key)
 {
```
tautschnig, line 461: **Wrong exception type for missing dict key.** CPython:
```python
>>> d = {'a': 1}
>>> del d['missing']   # or d['missing']
KeyError: 'missing'
```
This function returns `IndexError`, which propagates up to `Any_get!` / `Any_get_no_string!` and further up to handler dispatch. Two consequences:

1. `try: d["missing"] except KeyError: pass` does **not** catch in our translation — `KeyError` isn't in `errTyToGuard`, routes to `isUnknownError`, which is never true since the runtime raised `IndexError`.
2. `try: d["missing"] except IndexError: pass` **does** catch in our translation but would not in CPython.

Both divergences silently; no fixture exercises either case.

Fix needs two pieces:

1. Add a `KeyError` constructor to the `Error` datatype at line 40 (you already add `ZeroDivisionError` / `UnknownError` as part of this PR, so this is a one-line addition).
2. Change this line to `exception(KeyError(key))` — passing the missing key string as payload matches CPython's `KeyError` shape (`repr(key)` as `str(e)`).

```suggestion
function DictStrAny_get (d : DictStrAny, key: string) : Any
{
  if  DictStrAny..isDictStrAny_empty(d) then exception(KeyError(key))
  else if DictStrAny..key!(d) == key then DictStrAny..val!(d)
  else DictStrAny_get(DictStrAny..tail!(d), key)
};
```

And correspondingly add `"KeyError" => mkStmtExprMd (.StaticCall "Error..isKeyError" [maybeExceptVarExpr])` to `errTyToGuard`.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -349,15 +351,13 @@ function List_extend (l1 : ListAny, l2: ListAny) : ListAny
...
-  if ListAny..isListAny_nil(l) then from_None()
+  if (i < 0 || ListAny..isListAny_nil(l)) then exception(IndexError("Index out of bound"))
   else if  i == 0 then ListAny..head!(l)
   else List_get(ListAny..tail!(l), i - 1)
 };
 
 function List_get (l : ListAny, i : int) : Any
-  requires i >= - List_len(l) && i < List_len(l)
 {
   if i >= 0 then List_get_non_neg(l, i)
```
tautschnig, line 373: **Proof-coverage follow-up.** The `requires i >= 0 && i < List_len(l)` / `requires i >= -List_len(l) && i < List_len(l)` preconditions are replaced here by exception-returning arms. This is a genuine soundness improvement — previous callers' precondition obligations are now explicit paths in the result.

Worth stating as `@[simp]`-shaped post-conditions so the refactor doesn't silently lose happy-path provability:

```lean
theorem List_get_non_neg_in_bounds_no_exception (l : ListAny) (i : Int)
    (h0 : 0 ≤ i) (h1 : i < List_len l) :
    ¬ Any..isexception (List_get_non_neg l i)

theorem List_get_in_bounds_no_exception (l : ListAny) (i : Int)
    (h0 : -List_len l ≤ i) (h1 : i < List_len l) :
    ¬ Any..isexception (List_get l i)

-- analogous for DictStrAny_get, str_get, str_slice, PFloorDiv, PMod
```

Each is the exact inverse of the `requires` clause that was removed. Proving them locks in that callers that previously had to discharge the `requires` now don't suddenly hit the exception path. Structurally small, `rfl`/`simp`-shaped once `Any..isexception` reduces on the `from_...` / `exception` constructors. Not blocking this PR — a follow-up with the four theorems at worst.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1613,21 +1625,46 @@ def getExceptionCheckPreamble (ctx : TranslationContext) (e : StmtExprMd) (varNa
...
+        mkStmtExprMd $ .Assign [mkVariableMd (.Field (freeVarExpr e.val) "coreError")] maybeExceptVarExpr)
+    let bodyStmts := errorVarSet ++ bodyStmts
+    return (errTyToGuard errorTy, mkStmtExprMdWithLoc (StmtExpr.Block bodyStmts none) md)
+
+partial def exceptHandlersToIfStmt
+        (md: Option FileRange)
+        (ctx: TranslationContext)
+        (h : List (Python.excepthandler SourceRange)) : Except TranslationError (Option StmtExprMd) := do
+  let exceptionHandler ←  h.mapM $ exceptHandlerToBlock ctx
+  let assertUnCaughtMdCond := {condition:= mkStmtExprMd (.StaticCall "Error..isNoError" [maybeExceptVarExpr]), summary:= "Assert No UnCaught Exception"}
```
tautschnig: Nit (purely cosmetic): `"Assert No UnCaught Exception"` uses CamelCase mid-word; other summaries in the file are sentence case (`"caught exception"`, `"modified first"`, etc.). SARIF output looks better with:

```suggestion
    let assertUnCaughtMdCond := {condition:= mkStmtExprMd (.StaticCall "Error..isNoError" [maybeExceptVarExpr]), summary:= "assert no uncaught exception"}
```

tautschnig (commented):
Four concerns that remained after reading everything, plus proof/test suggestions.

**1. Python exception hierarchy is not modeled.** CPython's `except ArithmeticError:` catches `ZeroDivisionError`, `OverflowError`, `FloatingPointError`; `except LookupError:` catches `IndexError`, `KeyError`; `except Exception:` catches almost everything. Our `errTyToGuard` at `PythonToLaurel.lean:1582-1590` does a *flat* string match — so `except ArithmeticError:` falls through the `_ =>` arm and maps to `Error..isUnknownError`, which will never fire. Valid Python code like
```python
try:
    1 / 0
except ArithmeticError as e:
    handle(e)
```
runs the handler in CPython and bypasses it in our translation. The current fixtures don't exercise any superclass-matching case, so the divergence is silent. Inline on `errTyToGuard` with a concrete sketch for the fix (subclass-closure map or transitive `isError` test), plus test suggestions.

**2. `KeyError` mapped to `IndexError`.** `DictStrAny_get` returns `exception(IndexError("Dict does not contain key"))` for a missing key (`PythonRuntimeLaurelPart.lean:450`). Python raises `KeyError`. Consequence: `try: d["missing"] except KeyError: pass` does not catch in our translation (KeyError falls through `errTyToGuard`'s `_ =>` arm to `Error..isUnknownError`), and `try: d["missing"] except IndexError: pass` catches when it shouldn't. No new `KeyError` constructor on the Error datatype, so even adding a handler arm wouldn't fix this without an `Error` extension. Worth filing and tracking; not blocking, but please at least extend the `Error` datatype with `KeyError` as part of this PR since it's a one-line addition alongside the `ZeroDivisionError` / `UnknownError` constructors already landing here.

**3. Dead code in the `.Return` arm** at `PythonToLaurel.lean:1761`: `let exceptionCheck := getExceptionCatch ctx e` is computed but never referenced — the generated return block uses `preamble` from line 1762, and `exceptionCheck` is discarded. Either use it (if intended — the `.Expr` arm at :1805 does use the analogous binding) or delete it. Inline.

**4. `_ =>` fallback in `errTyToGuard` silently swallows unknown exception names.** A user writing `except FileNotFoundError:` or `except KeyError:` gets no diagnostic; the generated code checks `Error..isUnknownError` which never fires because no runtime function raises it (the constructor exists but I didn't find a `raise UnknownError` site). If the intent is to reject unhandled exception types at translation time, `throw (.unsupportedConstruct "...")` is the right call. If the intent is to catch "any future unknown error", at least emit a `dbg_trace` warning so users know the handler is inert. Inline.

**Proof opportunities** on the newly-total runtime functions — the `requires` → exception-returning refactor should preserve happy-path provability, and these are the structural theorems that lock it in:

- `List_get_non_neg_in_bounds_no_exception`: `0 ≤ i < List_len(l) → ¬ Any..isexception(List_get_non_neg(l, i))`
- `List_get_in_bounds_no_exception`: `-List_len(l) ≤ i < List_len(l) → ¬ Any..isexception(List_get(l, i))`
- `DictStrAny_get_contains_no_exception`: `DictStrAny_contains(d, k) → ¬ Any..isexception(DictStrAny_get(d, k))`
- `PFloorDiv_nonzero_no_exception`: if the second arg is a non-zero bool/int Any, the result is not a `ZeroDivisionError` exception (type-error still possible).

These are the exact statements that the previous `requires` preconditions encoded — re-stating them as proven post-conditions (rather than callers' obligations) pins down that the refactor didn't widen the set of cases that raise.

**Test coverage gaps** (each small, each the concrete analogue of a concern above):

- `test_try_except_arithmetic_superclass.py` — `try: 1/0 except ArithmeticError as e: …` — currently broken per (1).
- `test_try_except_key_error.py` — `try: d["missing"] except KeyError: …` — currently broken per (2).
- `test_try_except_unknown_handler.py` — `try: … except NotARealException: …` — pins down the behaviour of the fallback per (4). Expected: either translation error or a clearly-inert handler.
- `test_try_except_as_e_used.py` — the current `as e` binding path assigns `e.coreError := maybe_except`, but no fixture exercises a handler that *reads* `e.coreError`. Add one where the handler body does `str(e)` or similar so the binding is regression-tested.
- `test_try_except_nested.py` — outer try, inner try inside the outer `try`'s body; `isInsideTryBlock` restoration at `:1886` (`{bodyCtx with isInsideTryBlock := ctx.isInsideTryBlock}`) is load-bearing for this case but not tested.
- `test_raise.py` — `raise ValueError("x")`. The PR description says `.Raise` is explicitly left as FIXME / dropped; a fixture that fails loudly when it's unimplemented would prevent silent regressions once it's added.

The 3rd case in `test_try_catch_multiple_exceptions.py` already expects `❓ unknown` on the "Assert No UnCaught Exception" — worth a comment in the fixture explaining that this is the *correct* negative case (uncaught `TypeError`) and not a solver-precision gap to be fixed later; otherwise someone tightening the solver will think it's a bug.

**Refactor callouts:**

- `errTyToGuard` (`:1582-1590`) is a flat switch over strings. With the concerns in (1)/(2)/(4), the natural shape is a table mapping Python exception names to (a) the `Error..isX` predicate, and (b) a list of Python subclass names that also route here. Extract to a `def pythonErrorToGuard : List (String × (List String × String)) := …` so the hierarchy is declarative.
- `Any_get_no_string!` (`PythonRuntimeLaurelPart.lean:522`) exists solely to keep `Any_sets!` from dispatching strings (strings are immutable in Python, so there's nothing to set). A one-line docstring explaining this is worth adding — the name hints at it but not at the reason.
- `getExceptionCatch` dispatches on `ctx.isInsideTryBlock`. Currently the flag is threaded through `TranslationContext`, used only here. If cleaner, consider making it a local parameter — but given the flag has to survive nested mutual recursion through `translateStmt` / `translateStmtList`, the context field is probably the right home. Leaving it as context is fine; one-line comment at the field definition explaining what flips it would help.

**Nit**: the user-visible summary `"Assert No UnCaught Exception"` at line 1615 has inconsistent capitalisation (CamelCase mid-word). `"Assert no uncaught exception"` reads cleaner in SARIF output alongside the other lowercase summaries.

thanhnguyen-aws:
@keyboardDrummer-bot Please add some more expected tests to cover newly added Exception types.

PR #754, opened 2026-04-03
Unify translateFunction and translateMethod

thanhnguyen-aws:
## Summary

Unifies `translateFunction` and `translateMethod` into a single `translateFunction` that handles both standalone functions and class methods. Previously these were separate ~70-line functions with nearly identical logic for parameter handling, body translation, and procedure construction.

### Removed `translateMethod`

The standalone `translateMethod` function (~65 lines) is deleted. Class methods are now translated by calling `translateFunction` with `ctx.currentClassName` set, which triggers the method-specific behavior.

### `translateFunction` handles both cases

- **`body` parameter changed to `Option`**: `body: Option (List (Python.stmt SourceRange))`. When `none` (e.g., for hierarchy classes whose bodies are opaque), no body translation occurs.
- **Self parameter injection**: When `ctx.currentClassName` is set, a `self` parameter (typed as the composite class) is prepended to the input list.
- **`__init__` postcondition suppression**: Return type postconditions are skipped for `__init__` methods (which return `None`, not the annotated return type).
- **Transparent vs Opaque body**: Procedures with no postconditions use `Body.Transparent`; those with postconditions use `Body.Opaque`. Previously all procedures were unconditionally `Body.Opaque`.

### `translateFunctionBody` refactored

- Now takes `kwargsName` and `inputs` (the full parameter list) instead of `inputTypes` (a string-typed list).
- Parameter renaming (`$in_` prefix) and local-copy generation moved into `translateFunctionBody`, so both functions and methods get the same treatment.
- `noneReturn` (default `LaurelResult := None`) prepended inside the body helper.

### Class translation simplified

`translateClass` now calls `translateFunction` for each method, passing `none` as the body for hierarchy classes (whose methods are opaque). The old pattern of calling `translateMethod` then overriding `body := .Opaque [] .none []` is replaced by `translateFunction ctx sr funcDecl (if inHierarchy then none else funcBody)`.

## Test changes

Updated expected outputs for class-related tests to reflect:
- Methods now get `$in_` parameter renaming (previously only standalone functions did)
- Methods now get `LaurelResult := None` initialization
- Procedures without postconditions are `Transparent` instead of `Opaque`


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1781,11 +1728,13 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
...
-    for stmt in body do
-      if let .FunctionDef .. := stmt then
-        let proc ← translateMethod ctx className stmt
+
+    for (funcDecl, sr,  funcBody) in classFunDeclsAndBody do
+      let (proc, _) ← translateFunction ctx sr funcDecl funcBody
         -- TODO stop replacing the body of instance proceduces with an empty one
-        instanceProcedures := instanceProcedures.push { proc with body := .Opaque [] .none [] }
+      instanceProcedures := instanceProcedures.push
+        { proc with body := .Opaque [] .none [], preconditions := [mkStmtExprMd (StmtExpr.LiteralBool true)] }
```
tautschnig: The old `translateMethod` set `determinism := .nondeterministic` for methods, but `translateFunction` sets `determinism := .deterministic none`. Since only `body` and `preconditions` are overridden here, methods now get `.deterministic none` instead of `.nondeterministic`.

Is this intentional? If not, add `determinism` to the override:

```suggestion
      instanceProcedures := instanceProcedures.push
        { proc with body := .Opaque [] .none [], preconditions := [mkStmtExprMd (StmtExpr.LiteralBool true)], determinism := .nondeterministic }
```

tautschnig: The body translation work done by `translateFunction` (translating the method body, generating parameter copies, etc.) is immediately discarded by the `body := .Opaque [] .none []` override. This is wasted computation.

Since the TODO comment says "stop replacing the body of instance procedures with an empty one," this is presumably temporary. But until that's resolved, you could avoid the wasted work by either:
- Passing a flag to skip body translation for methods, or
- Adding a comment noting the waste is intentional/temporary

Not a blocker — just noting the inefficiency.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1516,26 +1516,33 @@ def createBoolOrExpr (exprs: List StmtExprMd) : StmtExprMd :=
 
 def getUnionTypeConstraint (var: String) (md: MetaData) (tys: List String) (funcname: String): Option StmtExprMd :=
   let type_constraints := tys.filterMap (getSingleTypeConstraint var)
```
tautschnig, line 1855: The `$in_` prefix stripping is fragile — it assumes no user variable starts with `$in_`. Since `$` is not a valid Python identifier character, this is safe in practice, but a comment noting this invariant would help:

```lean
  -- Safe because `$` is not a valid Python identifier character,
  -- so user variables never start with "$in_".
  let originalVarName := if var.startsWith "$in_" then var.drop 4 else var
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1552,6 +1559,16 @@ def translateFunction (ctx : TranslationContext) (sourceRange: SourceRange) (fun
...
 
+    inputs := match ctx.currentClassName with
+    | some className =>
+    -- First parameter is self (typed as Composite to match call-site convention)
+      let selfParam : Parameter := {
+        name := "self"
+        type := mkHighTypeMd (.UserDefined (mkId className))
+      }
+      [selfParam] ++ inputs
+    | _ => inputs
```
tautschnig, line 1918: nit: The indentation of the comment and `let` block is off — the comment and body are indented one level less than the surrounding `match` arms. Consider:

```suggestion
    inputs := match ctx.currentClassName with
    | some className =>
      -- First parameter is self (typed as Composite to match call-site convention)
      let selfParam : Parameter := {
        name := "self"
        type := mkHighTypeMd (.UserDefined (mkId className))
      }
      [selfParam] ++ inputs
    | _ => inputs
```

tautschnig (commented):
### Behavioral changes to verify

This refactoring introduces several subtle behavioral changes that should be confirmed as intentional:

1. **Regular functions now get `$in_` parameter renaming** — previously only methods had this. This is arguably more correct (Python params are mutable everywhere), but changes the generated Laurel/Core output for all functions.

2. **Method `determinism` changed from `.nondeterministic` to `.deterministic none`** — the old `translateMethod` explicitly set `.nondeterministic`, but `translateFunction` uses `.deterministic none`. The `translateClass` override doesn't restore this.

3. **Method `variableTypes` no longer includes `self`** — the old code added all params (including `self`) to `variableTypes` for hoisting prevention. The new code derives `variableTypes` from `funcDecl.args` which excludes `self`.

tautschnig:

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1818,17 +1818,8 @@ def getReturnTypeEnsure (md: MetaData) (tys: List String) (funcname: String): Op
...
-  : Except TranslationError (StmtExprMd × TranslationContext) := do
-    let ctx := {ctx with variableTypes:= ("nullcall_ret", PyLauType.Any)::inputTypes}
-    let newDecls := collectDeclaredNamesAndTypes ctx body
-    let (varDecls, ctx) ←  createVarDeclStmtsAndCtx ctx newDecls
-    let (newctx, bodyStmts) ← translateStmtList ctx body
-    let bodyStmts := prependExceptHandlingHelper (varDecls ++ bodyStmts)
-    let bodyStmts := (mkStmtExprMd (.LocalVariable "nullcall_ret" AnyTy (some AnyNone))) :: bodyStmts
-    return (mkStmtExprMd (StmtExpr.Block bodyStmts none), newctx)
+def getInputTypePreconditions (funcDecl : PythonFunctionDecl): List StmtExprMd :=
+  funcDecl.args.filterMap (λ arg => getUnionTypeConstraint ("$in_" ++ arg.name) arg.md arg.tys funcDecl.name)
```
tautschnig: **Dead code:** `getInputTypePreconditions` is defined but never called. The actual precondition generation is inline in `translateFunction` (line 1882). This should be removed.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -2128,17 +2083,18 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
...
-        let proc ← translateMethod ctx className stmt
-        if inHierarchy then
+
+    for (funcDecl, sr,  funcBody) in classFunDeclsAndBody do
+      let (proc, _) ← translateFunction ctx sr funcDecl funcBody
+      if inHierarchy then
           instanceProcedures := instanceProcedures.push { proc with body := .Opaque [] .none [] }
-        else
+      else
           instanceProcedures := instanceProcedures.push proc
```
tautschnig: **Still outstanding (Thread 2):** For `inHierarchy` classes, `translateFunction` translates the full method body and then the result is immediately discarded (`body := .Opaque [] .none []`). This is wasteful — consider checking `inHierarchy` before calling `translateFunction` and skipping body translation entirely.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1858,13 +1865,24 @@ def translateFunction (ctx : TranslationContext) (sourceRange: SourceRange) (fun
...
 
+    inputs := match ctx.currentClassName with
+    | some className =>
+    -- First parameter is self (typed as Composite to match call-site convention)
+      let selfParam : Parameter := {
+        name := "self"
+        type := mkHighTypeMd (.UserDefined (mkId className))
+      }
+      [selfParam] ++ inputs
+    | _ => inputs
```
tautschnig, line 1918: **Still outstanding (Thread 4):** The indentation is inconsistent — the comment on line 1870 and the `let` block are indented one level less than the surrounding `match` arms.

```suggestion
    inputs := match ctx.currentClassName with
    | some className =>
      -- First parameter is self (typed as Composite to match call-site convention)
      let selfParam : Parameter := {
        name := "self"
        type := mkHighTypeMd (.UserDefined (mkId className))
      }
      [selfParam] ++ inputs
    | _ => inputs
```

tautschnig (dismissed):

keyboardDrummer (approved):
Test changes look like improvements and the refactoring increases code reuse: nice!

tautschnig (approved):

PR #758, opened 2026-04-06
Perform call elimination as a standalone phase

### LINKED ISSUE #756: Perform call elimination as a standalone phase
Recently, loop elimination was lifted as its own phase, which makes it easier to profile and prove soundness.
The final comments on https://github.com/strata-org/Strata/pull/753 indicate that procedure call elimination is not entirely done in its own phase.
For better soundness and performance profiling, procedure call elimination should be performed as a Core.Program to Core.Program transformation that is for example executed after loop elimination.

MikaelMayer:
Fixes #756

## Problem

Call elimination was not performed as a standalone phase in all verification paths. When verifying all procedures (`proceduresToVerify = none`), calls were handled inline during evaluation by `Command.evalCall`. Only targeted verification (`proceduresToVerify = some procs`) ran `callElimPipelinePhase` as a standalone `Program → Program` transformation.

This also meant the `callElimPipelinePhase` model validation — which conservatively demotes `sat` to `unknown` for obligations involving call elimination assumptions — was not applied in the common verification path. As a result, spurious counterexamples from the over-approximation of replacing call bodies with postconditions were reported as definite failures ("assertion does not hold") instead of unknowns ("assertion could not be proved").

## Solution

Unify both verification paths to use `corePipelinePhases`, so `callElimPipelinePhase` always runs as a standalone transform before type checking and evaluation. After call elimination inlines callee contracts (assert preconditions, havoc modified vars, assume postconditions), a new `updateModifiesClauses` pass updates each procedure's `modifies` clause to include globals now modified in the body, keeping the program well-formed for type checking.

The model validation now applies uniformly, correctly demoting spurious counterexamples to "unknown" for all verification paths.

`AbstractedPhase` now has a `getAssertDescription` field that lets each pipeline phase define how to produce diagnostic descriptions from obligation labels. Call elimination uses this to extract the original requires label from `callElimAssert_` labels and produce messages like `precondition 'counter_ge_zero' does not hold`. The Verifier queries all phases instead of hardcoding knowledge of any specific transform's label format.

## Follow-ups

- #792: Refactor label extraction in `getAssertDescription` to use structured metadata instead of string processing.

## Testing

All existing tests pass (updated expected outputs to reflect new call elimination labels, diagnostic locations, and precondition error messages). Re-enabled the `multipleRequiresCaller` test that was previously disabled because Core incorrectly reported error locations on procedure preconditions — now that call elimination reports errors at the call site, the test works correctly.

### DIFF
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -49,7 +49,8 @@ when needed for the validity check (line 64 for check-sat-assuming, line 77 for 
 def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
     (assumptionTerms : List Term) (obligationTerm : Term)
     (md : Imperative.MetaData Core.Expression)
-    (satisfiabilityCheck validityCheck : Bool) :
+    (satisfiabilityCheck validityCheck : Bool)
+    (label : String) :
     SolverM (List String × EncoderState) := do
   Solver.setLogic "ALL"
   prelude
@@ -104,11 +105,11 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
       let _ ← Solver.checkSat ids
 
   -- Emit the "message" metadata field at the very end (once per obligation).
-  match md.findElem Imperative.MetaData.message with
-  | some elem =>
-    let msg := toString (Std.format elem.value) |>.replace "\\" "\\\\" |>.replace "\"" "\\\""
-    Solver.setInfo "final-message" s!"\"{msg}\""
-  | none => pure ()
+  let rawMsg := match md.findElem Imperative.MetaData.message with
+    | some elem => toString (Std.format elem.value)
+    | none => label
+  let escaped := rawMsg.replace "\\" "\\\\" |>.replace "\"" "\\\""
+  Solver.setInfo "final-message" s!"\"{escaped}\""
 
   return (ids, estate)
 
@@ -122,6 +123,17 @@ open Std (ToFormat Format format)
 open Lambda Strata.SMT
 
 public section
+
+/-- Replace characters that are problematic on common filesystems
+    (parens, quotes, spaces, path separators, and Windows-invalid characters
+    such as `< > : | ? *`) with underscores or remove them.
+    Single-pass over the string. -/
+def sanitizeFilename (s : String) : String :=
+  String.ofList <| s.toList.filterMap fun c =>
+    if c == '"' || c == '\'' then none
+    else if c == '(' || c == ')' || c == ' ' || c == '/' || c == '\\'
+         || c == '<' || c == '>' || c == ':' || c == '|' || c == '?' || c == '*' then some '_'
+    else some c
 
 private def typedVarToSMTFn (ctx : SMT.Context) (id : Core.Expression.Ident)
   (ty : Core.Expression.Ty) := do
@@ -165,6 +177,7 @@ def dischargeObligation
   (obligationTerm : Term)
   (ctx : SMT.Context)
   (satisfiabilityCheck validityCheck : Bool)
+  (label : String)
   : IO (Except Format (SMT.Result × SMT.Result × EncoderState)) := do
   -- CVC5 requires --incremental for multiple (check-sat) commands
   let baseFlags := getSolverFlags options
@@ -177,7 +190,8 @@ def dischargeObligation
   Imperative.SMT.dischargeObligation
     (P := Core.Expression)
     (Strata.SMT.Encoder.encodeCore ctx (getSolverPrelude options.solver)
-      assumptionTerms obligationTerm md satisfiabilityCheck validityCheck)
+      assumptionTerms obligationTerm md satisfiabilityCheck validityCheck
+      (label := label))
     (typedVarToSMTFn ctx)
     vars
     options.solver
@@ -726,7 +740,7 @@ def getObligationResult (assumptionTerms : List Term) (obligationTerm : Term)
   let prog := f!"\n\n[DEBUG] Evaluated program:\n{Core.formatProgram p}"
   let counterVal ← counter.get
   counter.set (counterVal + 1)
-  let filename := tempDir / s!"{obligation.label}_{counterVal}.smt2"
+  let filename := tempDir / s!"{Core.SMT.sanitizeFilename obligation.label}_{counterVal}.smt2"
   let varsInObligation := ProofObligation.getVars obligation
   -- All variables in ProofObligation must have been typed.
   let typedVarsInObligation ← varsInObligation.mapM
@@ -741,7 +755,8 @@ def getObligationResult (assumptionTerms : List Term) (obligationTerm : Term)
             typedVarsInObligation
             obligation.metadata
             filename.toString
-          assumptionTerms obligationTerm ctx satisfiabilityCheck validityCheck)
+          assumptionTerms obligationTerm ctx satisfiabilityCheck validityCheck
+          (label := obligation.label))
   match ans with
   | .error e =>
     dbg_trace f!"\n\nObligation {obligation.label}: SMT Solver Invocation Error!\
--- a/StrataTest/Languages/Core/Examples/FreeRequireEnsure.lean
+++ b/StrataTest/Languages/Core/Examples/FreeRequireEnsure.lean
@@ -68,6 +68,8 @@ Model:
 
 
 [DEBUG] Evaluated program:
+program Core;
+
 var g : int;
 procedure Proc () returns ()
 spec {
--- a/StrataTest/Languages/Core/Tests/PolymorphicProcedureTest.lean
+++ b/StrataTest/Languages/Core/Tests/PolymorphicProcedureTest.lean
@@ -60,6 +60,8 @@ Model:
 
 
 [DEBUG] Evaluated program:
+program Core;
+
 datatype List (a : Type) {(
   (Nil())),
   (Cons(head : a, tail : (List a)))
--- a/.github/workflows/cbmc.yml
+++ b/.github/workflows/cbmc.yml
@@ -106,6 +106,10 @@ jobs:
         run: |
           pip install ./Tools/Python
           ./StrataTest/Languages/Python/run_py_cbmc_tests.sh
+      - name: Run property summary tests
+        shell: bash
+        run: |
+          ./StrataTest/Backends/CBMC/GOTO/test_property_summary_e2e.sh
       - name: Run Laurel CBMC pipeline tests
         shell: bash
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,7 +242,7 @@ jobs:
               sudo cp z3-4.13.4-x64-glibc-2.35/bin/z3 /usr/local/bin/
             fi
     - name: Run PySpec and dispatch tests
-      run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest StrataTest.Languages.Python.AnalyzeLaurelTest StrataTest.Languages.Python.Specs.IdentifyOverloadsTest StrataTest.Languages.Python.VerifyPythonTest
+      run: PYTHON=python PYTHON_TEST=1 lake build StrataTest.Languages.Python.SpecsTest StrataTest.Languages.Python.AnalyzeLaurelTest StrataTest.Languages.Python.Specs.IdentifyOverloadsTest StrataTest.Languages.Python.VerifyPythonTest StrataTest.Languages.Python.PropertySummaryTest
     - name: Run test script
       run: FAIL_FAST=1 ./scripts/run_cpython_tests.sh ${{ matrix.python_version }}
       working-directory: Tools/Python
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,8 +78,8 @@ can be useful to provide different feedback for an empty list.
 ### Prefer Extrinsic Proofs
 
 Given the `median` function above, we may want to know that it always
-produces an `Int` given a list of even size. We could have included a
-precondition that the list has even size, avoiding the need for error
+produces an `Int` given a list of odd size. We could have included a
+precondition that the list has odd size, avoiding the need for error
 handling, but that would have precluded the possibility of providing
 useful feedback in the case of working with unfiltered input.
 
@@ -179,7 +179,7 @@ encouraged.
 
 Every major function and theorem must be documented (NOTE: use `/--
 ... -/` and NOT `/- ... -/`). Use backticks to format code or math
-expressions in the docstring (e.g., `x + y`).
+expressions in the docstring (e.g., ``/-- ... `x + y` ... -/``).
 
 Describe implementation-level details in single- (`--`) or multi-line
 comments (`/- .. -/`) interspersed with the code.
--- a/Examples/DoubleTwice.core.st
+++ b/Examples/DoubleTwice.core.st
@@ -1,0 +1,18 @@
+program Core;
+
+procedure Double(n : int) returns (result : int)
+spec {
+  ensures [double_correct]: (result == n * 2);
+}
+{
+  result := n + n;
+};
+
+procedure TestProc(x : int) returns (output : int)
+spec {
+  ensures [testProc_result]: (output == x * 4);
+}
+{
+  call output := Double(x);
+  call output := Double(output);
+};
--- a/Examples/IrrelevantAxioms.core.st
+++ b/Examples/IrrelevantAxioms.core.st
@@ -1,0 +1,24 @@
+program Core;
+
+// Two independent uninterpreted functions, each with their own axioms.
+// removeIrrelevantAxioms --functions f should prune g's axioms (and vice versa).
+
+function f(x: int): int;
+axiom [f_positive]: forall x: int :: f(x) > 0;
+axiom [f_monotone]: forall x: int, y: int :: x < y ==> f(x) < f(y);
+
+function g(x: int): int;
+axiom [g_negative]: forall x: int :: g(x) < 0;
+axiom [g_idempotent]: forall x: int :: g(g(x)) == g(x);
+
+// h calls f, so h's axiom is also relevant when focusing on f.
+function h(x: int): int;
+axiom [h_def]: forall x: int :: h(x) == f(x) + 1;
+
+procedure TestF(x: int) returns (result: int)
+spec {
+  ensures [result_positive]: result > 0;
+}
+{
+  result := f(x);
+};
--- a/Examples/expected/DoubleTwice.callElim.core.expected
+++ b/Examples/expected/DoubleTwice.callElim.core.expected
@@ -1,0 +1,4 @@
+Successfully parsed.
+DoubleTwice.callElim.core.st(5, 2) [double_correct]: ✅ pass
+DoubleTwice.callElim.core.st(11, 2) [testProc_result]: ✅ pass
+All 2 goals passed.
--- a/Examples/expected/DoubleTwice.callElim.core.st
+++ b/Examples/expected/DoubleTwice.callElim.core.st
@@ -1,0 +1,21 @@
+program Core;
+
+procedure Double (n : int) returns (result : int)
+spec {
+  ensures [double_correct]: result == n * 2;
+  } {
+  result := n + n;
+  };
+procedure TestProc (x : int) returns (output : int)
+spec {
+  ensures [testProc_result]: output == x * 4;
+  } {
+  var tmp_arg_3 : int := x;
+  var tmp_output_4 : int := output;
+  havoc output;
+  assume [callElimAssume_double_correct_5]: output == tmp_arg_3 * 2;
+  var tmp_arg_0 : int := output;
+  var tmp_output_1 : int := output;
+  havoc output;
+  assume [callElimAssume_double_correct_2]: output == tmp_arg_0 * 2;
+  };
--- a/Examples/expected/DoubleTwice.inlineProcedures.core.expected
+++ b/Examples/expected/DoubleTwice.inlineProcedures.core.expected
@@ -1,0 +1,4 @@
+Successfully parsed.
+DoubleTwice.inlineProcedures.core.st(5, 2) [double_correct]: ✅ pass
+DoubleTwice.inlineProcedures.core.st(11, 2) [testProc_result]: ✅ pass
+All 2 goals passed.
--- a/Examples/expected/DoubleTwice.inlineProcedures.core.st
+++ b/Examples/expected/DoubleTwice.inlineProcedures.core.st
@@ -1,0 +1,25 @@
+program Core;
+
+procedure Double (n : int) returns (result : int)
+spec {
+  ensures [double_correct]: result == n * 2;
+  } {
+  result := n + n;
+  };
+procedure TestProc (x : int) returns (output : int)
+spec {
+  ensures [testProc_result]: output == x * 4;
+  } {
+  Double$inlined: {
+    var Double_n_3 : int := x;
+    var Double_result_4 : int;
+    Double_result_4 := Double_n_3 + Double_n_3;
+    output := Double_result_4;
+    }
+  Double$inlined: {
+    var Double_n_0 : int := output;
+    var Double_result_1 : int;
+    Double_result_1 := Double_n_0 + Double_n_0;
+    output := Double_result_1;
+    }
+  };
--- a/Examples/expected/DoubleTwice.inlineProcedures.filterProcedures.core.args
+++ b/Examples/expected/DoubleTwice.inlineProcedures.filterProcedures.core.args
@@ -1,0 +1,1 @@
+--pass inlineProcedures --procedures Double --pass filterProcedures --procedures TestProc
--- a/Examples/expected/DoubleTwice.inlineProcedures.filterProcedures.core.expected
+++ b/Examples/expected/DoubleTwice.inlineProcedures.filterProcedures.core.expected
@@ -1,0 +1,3 @@
+Successfully parsed.
+DoubleTwice.inlineProcedures.filterProcedures.core.st(5, 2) [testProc_result]: ✅ pass
+All 1 goals passed.
--- a/Examples/expected/DoubleTwice.inlineProcedures.filterProcedures.core.st
+++ b/Examples/expected/DoubleTwice.inlineProcedures.filterProcedures.core.st
@@ -1,0 +1,19 @@
+program Core;
+
+procedure TestProc (x : int) returns (output : int)
+spec {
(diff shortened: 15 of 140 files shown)


StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
```diff
@@ -15,9 +15,9 @@ namespace Strata.Laurel
 def program := r"
 procedure hasRequires(x: int) returns (r: int)
   requires x > 2
-//         ^^^^^ error: assertion does not hold
-// Core does not seem to report precondition errors correctly.
-// This should occur at the call site and with a different message
+// Call elimination reports precondition errors at the call site,
```
MikaelMayer, line 18: Now that call elimination reports precondition errors at the call site, the commented-out `multipleRequiresCaller` test (around line 53) might be unblocked. Its comment says it was disabled because "Core incorrectly report error locations on procedure preconditions" — which is exactly what this PR fixes. Worth checking if it can be re-enabled.

Strata/Transform/CallElim.lean
```diff
@@ -130,10 +130,34 @@ def callElimCmd (cmd: Command)
...
+    globals from the callee's `modifies` clause. Update each procedure's own
+    `modifies` clause so that it includes every global variable that is now
+    modified in the body. This keeps the program well-formed with respect to
+    `checkModificationRights`. -/
+private def updateModifiesClauses (p : Program) : Program :=
+  let globalNames : List Expression.Ident :=
+    p.decls.filterMap fun d => match d with | .var name _ _ _ => some name | _ => none
+  let decls := p.decls.map fun d => match d with
+    | .proc proc md =>
+      let bodyModified := (Imperative.Block.modifiedVars proc.body).eraseDups
```
MikaelMayer, line 143: [No action needed] `modifiedVars` + `eraseDups` followed by `filter` with `globalNames.contains` and `proc.spec.modifies.contains` gives O(n·m) list lookups. Fine for typical program sizes, but if this ever becomes a bottleneck, converting `globalNames` and `proc.spec.modifies` to sets would help.

Examples/expected/TypeError.core.expected
```diff
@@ -1,4 +1,4 @@
 Successfully parsed.
 TypeError.core.st(25, (2-18)) ❌ Type checking error.
-Impossible to unify (Map int (Map bool bool)) with (Map int int).
-First mismatch: (Map bool bool) with int.
+Impossible to unify (Map int int) with (Map int (Map bool bool)).
```
MikaelMayer, line 3: [No action needed] The unification error message swapped the order of the two types. This is a side effect of call elimination running before type checking — after the call is eliminated, the assignment direction changes. The new message is still correct, just flagging that this is an expected consequence rather than a regression.

StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
MikaelMayer: Uncommented the `multipleRequiresCaller` test — it now works correctly with call elimination reporting errors at the call site. The test builds and passes.

Strata/Transform/CallElim.lean
MikaelMayer: Acknowledged, no action taken. The O(n·m) complexity is acceptable for typical program sizes.

Examples/expected/TypeError.core.expected
MikaelMayer: Acknowledged, no action taken. The swapped type order in the unification error is expected given call elimination now runs before type checking.

StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
```diff
@@ -27,6 +27,7 @@ procedure hasRequires(x: int) returns (r: int)
 
 procedure caller() {
   var x: int := hasRequires(1);
+//^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
```
MikaelMayer: Is there any way we can see which precondition does not hold? Normally inlining preserves metadata, so we should be ablet to recover which expression does not hold and put it as a message here, which would be very useful to debug tests.

StrataTest/Languages/Laurel/Examples/Fundamentals/T8_Postconditions.lean
```diff
@@ -25,7 +25,7 @@ procedure callerOfOpaqueProcedure() {
   var x: int := opaqueBody(3);
   assert x > 0;
   assert x == 3
-//^^^^^^^^^^^^^ error: assertion does not hold
+//^^^^^^^^^^^^^ error: assertion could not be proved
```
MikaelMayer, line 28: Wow did this PR fix a soundness issue? Looks like yes, but what did actually solve that soundness issue before?

StrataTest/Languages/Core/Tests/PolymorphicProcedureTest.lean
```diff
@@ -132,14 +135,14 @@ true
...
-(Origin_MkCons_Ensures)MkCons_ensures_0: List..isCons($__r3)
+callElimAssume_MkCons_ensures_0_2: List..isCons($__r3)
 Obligation:
 List..isCons($__r3)
 
 Label: Test_ensures_0
 Property: assert
 Assumptions:
-(Origin_MkCons_Ensures)MkCons_ensures_0: List..isCons($__r3)
+callElimAssume_MkCons_ensures_0_2: List..isCons($__r3)
```
MikaelMayer, line 147: Why are these labels differents now? I don't understand.

StrataTest/Languages/Core/Tests/CallElimWithOldGlobals.lean
```diff
@@ -63,7 +63,7 @@ Result: ✅ pass
 
 Obligation: h_incorrect_ensures_1
 Property: assert
-Result: ❌ fail
+Result: ❓ unknown
```
MikaelMayer, line 66: Was the previous "fail" unsound, or is this a loss of precision?

StrataTest/Languages/Core/Examples/SelectiveVerification.lean
```diff
@@ -83,13 +83,13 @@ Obligation: result_correct
...
-Obligation: (Origin_Helper_Requires)n_positive
+Obligation: callElimAssert_n_positive_6
 Property: assert
 Result: ❌ fail
 
-Obligation: (Origin_Helper_Requires)n_positive
+Obligation: callElimAssert_n_positive_2
 Property: assert
-Result: ❌ fail
+Result: ❓ unknown
```
MikaelMayer, line 92: I guess so but is this loss of precision coherent?

StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
MikaelMayer: Yes! I added a `propertySummary` to the assert metadata in `createAsserts` (in `CoreTransform.lean`). When call elimination generates an assert for a precondition, it now attaches a summary like `"precondition 'requires_0'"`. The error message now reads e.g. "precondition 'requires' does not hold" instead of the generic "assertion does not hold". If the original requires clause already had a custom `propertySummary` (e.g. from an explicit error message in Laurel), that is preserved.

StrataTest/Languages/Laurel/Examples/Fundamentals/T8_Postconditions.lean
MikaelMayer: This is not a soundness fix — it's a loss of precision. The previous result ("assertion does not hold" = definite counterexample) was correct and sound: `assert x == 3` genuinely fails when we only know `x > 0` from the postcondition.

The change to "assertion could not be proved" (❓ unknown) happens because the call elimination phase introduces `callElimAssume` labels in the obligation's assumptions. The phase's `getValidation` sees these labels and returns `modelToValidate (fun _ => false)` (the TODO validator), which conservatively demotes the SMT solver's `sat` result to `unknown`. The assertion still fails — the solver just can't validate the model through the call elimination over-approximation.

StrataTest/Languages/Core/Tests/PolymorphicProcedureTest.lean
MikaelMayer: The labels changed because call elimination now runs as a standalone program-to-program pass instead of being interleaved with VC generation. Previously, the label was `(Origin_Extract_Requires)Extract_requires_0` — a composite label with origin tracking. Now, `createAsserts` in the call elimination transform generates fresh labels with the `callElimAssert_` prefix (e.g. `callElimAssert_Extract_requires_0_2`), where the suffix `_2` is a fresh counter to avoid collisions. The original requires label name is preserved in the middle. The `callElimAssume` labels follow the same pattern for postconditions. This naming convention is also used by the pipeline phase's `getValidation` to identify which obligations went through call elimination.

StrataTest/Languages/Core/Tests/CallElimWithOldGlobals.lean
MikaelMayer: This is a loss of precision, not a soundness fix. The previous "fail" was correct — `h_incorrect` sets `g := true`, then calls `f(i)` which returns `old g` (= `true` at the call site since `g` is not in `f`'s modifies), so `r == false` genuinely fails.

The change to "unknown" happens because call elimination introduces `callElimAssume` labels in the obligation's assumptions. The pipeline phase's `getValidation` detects these labels and applies `modelToValidate (fun _ => false)` — the TODO validator that always rejects models. This demotes the SMT solver's `sat` result to `unknown`, losing the definite counterexample. Once model validation is implemented for call elimination (the TODO), this should return to "fail".

StrataTest/Languages/Core/Examples/SelectiveVerification.lean
MikaelMayer: Yes, the loss of precision is coherent and follows the same pattern as the other cases. The second call `call output := Helper(output)` has `callElimAssert_n_positive_6` which still shows ❌ fail (the first call's output is used as input, and the solver can find a definite counterexample without relying on call elimination assumptions). The first call `call output := Helper(x)` has `callElimAssert_n_positive_2` which changed from ❌ fail to ❓ unknown — here `x >= 0` (from the requires) means `x` could be 0, so the precondition `n > 0` can fail. But the model validation through the call elimination phase demotes the `sat` to `unknown` via the TODO validator.

The key distinction: obligations whose models don't depend on `callElimAssume` assumptions keep their precision; those that do get demoted to unknown until model validation is implemented.

StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
MikaelMayer: Updated approach: instead of using `propertySummary` (which would have changed the pyAnalyzeLaurel output format), I detect the `callElimAssert_` label prefix in the Verifier's `toDiagnosticModel` and use "precondition" as the error description. The error message now reads "precondition does not hold" or "precondition could not be proved" for call elimination assert failures, while the full label (e.g. `callElimAssert_requires_4`) is preserved in the pyAnalyzeLaurel output.

StrataTest/Languages/Laurel/Examples/Fundamentals/T8_Postconditions.lean
MikaelMayer: I double down on what I say. If you run the program, x is always equal to 3 according to the program. It IS a spurious counter-example and it's misleading (and I say unsound) to label it as a counter-example. So what did actually solve that soundness issue ?

StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
MikaelMayer: "precondition is better" but since it's a test, can't we print which precondition does not hold in the comment?!

StrataTest/Languages/Laurel/Examples/Fundamentals/T8_Postconditions.lean
MikaelMayer: You're right — this PR does fix a soundness issue, and nothing was solving it before.

Previously, `callElimPipelinePhase` only ran when `proceduresToVerify = some _` (targeted verification). When verifying all procedures (`proceduresToVerify = none`, the common path for Laurel), call elimination didn't run as a standalone phase. Calls were handled inline during evaluation, but the `callElimPipelinePhase`'s model validation (which demotes `sat` to `unknown` via `modelToValidate (fun _ => false)`) never kicked in. So the solver's counterexample — which is spurious because it comes from the over-approximation of replacing the call body with just the postcondition — was reported as a definite failure.

Now that call elimination runs unconditionally as a standalone phase, the `callElimAssume` labels are present in the obligations, the model validation catches the spurious counterexample, and it's correctly reported as "could not be proved" (unknown) instead of "does not hold" (definite failure).

So to directly answer: nothing was handling this before. The soundness issue existed and this PR fixes it as a side effect of unifying the pipeline.

StrataTest/Languages/Laurel/Examples/Fundamentals/T6_Preconditions.lean
MikaelMayer: Done! The Verifier now extracts the original requires label from the `callElimAssert_` label and includes it in the error message. For example:
- Single requires: `precondition 'requires' does not hold`
- Multiple requires: `precondition 'requires_0' does not hold`
- Core named labels: `precondition 'counter_ge_zero' does not hold`

If the check already has a `propertySummary` (e.g. from an explicit Laurel error message), that takes priority. The change is localized to the Verifier's `toDiagnosticModel` so it doesn't affect the pyAnalyzeLaurel output format.

StrataTest/Languages/Laurel/Examples/Fundamentals/T10_ConstrainedTypes.lean
```diff
@@ -68,7 +68,7 @@ procedure argValid() returns (r: int) {
 // Argument to constrained-typed parameter — invalid (requires violation)
 procedure argInvalid() returns (r: int) {
   var x: int := takesNat(-1);
-//^^^^^^^^^^^^^^^^^^^^^^^^^^ error: precondition does not hold
+//^^^^^^^^^^^^^^^^^^^^^^^^^^ error: precondition 'requires' does not hold
```
MikaelMayer: 'requires' says nothing here. It should be "precondition 'nat' does not hold". And ideally it should be targeting the argument only. We shouldn't be modifying something if it's not improving the error reporting.
If easier, please revert the laurel test files so that we keep their old "//^^^^^^^^^^^^^^^^^^^^^^^^^^ error: precondition does not hold" without any change.

MikaelMayer: Fixed. When the extracted label is a generic `requires` or `requires_N` (which comes from Laurel's `translateChecks`), the Verifier now falls back to the plain `"precondition"` description instead of quoting the uninformative label. The error messages are now `precondition does not hold` for these cases, while Core programs with meaningful labels (like `counter_ge_zero`, `n_positive`) still show `precondition 'counter_ge_zero' does not hold`.

Strata/Languages/Core/Verifier.lean
```diff
@@ -1001,7 +999,22 @@ def toDiagnosticModel (vcr : Core.VCResult) : Option DiagnosticModel :=
...
+            -- Drop the trailing _N counter by finding the last underscore
+            let parts := stripped.splitOn "_"
+            let originalLabel := if parts.length > 1 then
+              "_".intercalate (parts.dropLast)
+            else stripped
+            -- Only quote the label if it's meaningful (not a generic "requires" / "requires_N")
+            if originalLabel == "requires" || originalLabel.startsWith "requires_" then
+              "precondition"
+            else
+              s!"precondition '{originalLabel}'"
```
aqjune-aws: Can we have this property extraction step as a part of `callElimPipelinePhase : PipelinePhase` in Strata/Transform/CallElim.lean as a new extra field (even if probably CallElim will be the only pass that will have meaningful function)? Then from the list of core pipelines we can fold the eventual output of this.

joscoh (dismissed):

StrataTest/Languages/Core/Examples/FreeRequireEnsure.lean
```diff
@@ -84,9 +81,12 @@ spec {
   assert [g_lt_10]: true;
   };
 procedure ProcCaller () returns (x : int)
-{
```
aqjune-aws, line 89: Should this original `ProcCaller` must have had `modifies g` as well (but type checker, specifically `checkModificationRights`, was wrongly passing it)? I was wondering why `updateModifiesClauses` was necessary after call elimination.

StrataTest/Languages/Core/Examples/OldExpressions.lean
```diff
@@ -80,40 +80,36 @@ true
 Label: T2_g2_eq_g
 Property: assert
 Assumptions:
-(Origin_T1_Ensures)T1_g_unchanged: true == true
```
aqjune-aws, line 83: It seems the `_unchanged` assertions disappeared - do you have idea why?

Strata/Languages/Core/Verifier.lean
MikaelMayer: 🤖 Done. Added a `getAssertDescription : String → Option String` field to `AbstractedPhase` (defaults to `fun _ => none`). The label extraction logic is now defined in `callElimPipelinePhase.phase.getAssertDescription` in `CallElim.lean`, and the Verifier queries all phases via `phases.findSome? (·.getAssertDescription label)` instead of hardcoding knowledge of `callElimAssertPrefix`. The Laurel and Python diagnostic paths now pass `coreAbstractedPhases` to `toDiagnosticModel`/`toDiagnostic`.

StrataTest/Languages/Core/Examples/FreeRequireEnsure.lean
MikaelMayer: 🤖 No, the original `ProcCaller` correctly had no `modifies g` — it does not directly modify `g` in its source body. `checkModificationRights` was not wrong: it checks the syntactic body, and before call elimination the body only contains `call := Proc()`, which is not a direct modification of `g`.

After call elimination replaces the call with `havoc g; assume ...`, the body now directly modifies `g`. That is why `updateModifiesClauses` exists: it scans the post-transformation body for globals that are now modified (via `havoc`) and adds them to the procedure's `modifies` clause to keep the program well-formed. This is a necessary fixup step of the call elimination transform, not a sign that the type checker was wrong.

StrataTest/Languages/Core/Examples/OldExpressions.lean
MikaelMayer: 🤖 Yes — the `_unchanged` assertions disappear because call elimination now correctly resolves `old g` for globals not in the callee's `modifies` clause.

`T1` declares `modifies g2` but not `modifies g`. The postcondition `T1_g_unchanged: g == old g` references `old g`, but since `g` is not modified by `T1`, call elimination substitutes `old g` with the current value of `g` at the call site (via `unmodifiedOldSubst`). This makes the postcondition `g == g`, which simplifies to `true`. The resulting `assume callElimAssume_T1_g_unchanged_N: true` is a no-op and gets optimized away from the VC assumptions.

Previously, when call elimination was interleaved with VC generation, the `old g` substitution was handled differently and the trivially-true assumption was still shown. The new behavior is more precise.

MikaelMayer: How was the old g substitutions handled differently? Please give a concrete example.

MikaelMayer: 🤖 Concrete example using this test case:

**Program state at the call site** (`T2` body):
- `g = true` (set by `g := true`)
- `g2 = false` (set by `g2 := false`)

**T1 declares** `modifies g2` (but NOT `modifies g`).

**Postcondition** `T1_g_unchanged: g == old g`.

---

**Old behavior** (`Command.evalCall` in `StatementEval.lean`):

1. `getCurrentGlobals` captures pre-call values: `g → true`, `g2 → false`
2. `old_g_subst` maps `old g → true` and `old g2 → false` (for ALL globals)
3. `globals_post_subst` maps `g → true` (unmodified, keeps current value) and `g2 → $__g211` (modified, fresh var)
4. Substituting `g == old g` with `postcond_subst_map` (which is `globals_post_subst ++ old_g_subst`): `g` → `true`, `old g` → `true` → result: `true == true`
5. This `true == true` is kept as an assumption in the VC (it's not simplified away during evaluation)

**New behavior** (standalone `callElim` in `CallElim.lean`):

1. `createOldVarsSubst` creates fresh snapshot vars only for globals IN `modifies`: `old g2 → $fresh_g2`
2. `unmodifiedOldSubst` maps `old g → g` directly (since `g ∉ modifies`, `old g` equals `g` at the call site)
3. Substituting `g == old g`: `old g` → `g` → result: `g == g`
4. This simplifies to `true`, and `assume callElimAssume_T1_g_unchanged: true` is a no-op that gets dropped from the VC

Both are semantically correct — the new version just simplifies earlier in the pipeline.

MikaelMayer: 🤖 Concrete example of how `old g` substitution differs:

**Postcondition:** `T1_g_unchanged: g == old g`
**T1 declares:** `modifies g2` (but NOT `modifies g`)

**On `main` (old path):** Call elimination did not run when `procs = none` (full verification). Instead, the verifier's inline VC generation handled calls. It substituted `old g` with the pre-call value of `g` (which was `true`), producing the assumption `T1_g_unchanged: true == true`. This trivially-true assumption was kept in the path conditions.

**On this branch (standalone call elimination):** Call elimination always runs. The `unmodifiedOldSubst` in `callElimCmd` detects that `g` is NOT in `T1`'s `modifies` clause, so it substitutes `old g` directly with the current fvar for `g`:

```
-- unmodifiedOldSubst maps: old_g → g
-- postcondition g == old g  →  g == g (after substitution)
```

The resulting `assume callElimAssume_T1_g_unchanged_3: g == g` is then partially evaluated. In `CmdEval.lean`, the assume evaluator does:

```lean
| .assume label e md =>
  ...
  match EC.denoteBool e with
  | some true => (c', σ)  -- satisfied, NOT added to path conditions
```

Since `g == g` evaluates to `true`, the assume is satisfied and never enters the path conditions. That's why `T1_g_unchanged` disappears from the VC assumptions entirely, rather than appearing as `true == true`.

StrataTest/Languages/Core/Examples/FreeRequireEnsure.lean
aqjune-aws: Hmm OK. so `modifies` is for syntactic check.... It is sad that one cannot rely on the `modifies` clause to figure out which global variables are not changed after the call, because if there is a nested procedure call it will not be captured. Leaning more to eventually removing global vars....

Strata/Transform/CallElim.lean
```diff
@@ -148,6 +172,18 @@ def callElimPipelinePhase : PipelinePhase where
     if obligationHasLabelPrefix obligation CallElim.callElimAssumePrefix then
       .modelToValidate (fun _ => /- TODO -/ false)
     else .modelPreserving
+  phase.getAssertDescription label :=
```
atomb, line 175: I'd really like a more structured way to extract this information. Sometimes string processing like this is unavoidable, but I think we should be able to refactor this to remove the need for it.

I'm happy to have that as a separate PR, but let's create an issue for it.

atomb (approved):
I think this is enough of an improvement over the previous state that it's worth merging, but I'd like us to prioritize refactoring soon to remove the need for matching on particular assertion label syntax.

StrataTest/Languages/Core/Examples/OldExpressions.lean
aqjune-aws: I see. What about removing `Command.evalCall` in Strata/Languages/Core/StatementEval.lean entirely, in this patch? :) We can leave a error message to `Core.Env` stating that partial evaluator does not support calls; it must have been handled by CallElim (or ProcedureInlining or any call-eliminating pass).

MikaelMayer: 🤖 Correction to my previous reply — the `unmodifiedOldSubst` logic was already present on `main`, so the substitution itself did not change.

The actual reason is simpler: on `main`, `callElimPipelinePhase` was only included when `procs` was `some _` (targeted verification):

```lean
-- main
let callElimPhase := match procs with
  | some _ => [callElimPipelinePhase]
  | none => []
```

The `OldExpressions.lean` test calls `verify oldExprPgm` without specifying `proceduresToVerify`, so `procs` is `none` and call elimination never ran. The postconditions were inlined by the old VC generation path, which used `(Origin_T1_Ensures)` prefix labels and produced `T1_g_unchanged: true == true` — a trivially-true but still-present assumption.

This PR changes call elimination to always run:

```lean
-- this PR
filterPhases ++ [callElimPipelinePhase] ++ precondPhase ++ ...
```

Now call elimination runs, `unmodifiedOldSubst` substitutes `old g` → `g` (since `g` is not in `T1`'s `modifies` clause), the postcondition `g == old g` becomes `g == g`, the partial evaluator evaluates it to `true`, and `CmdEval.lean` skips adding trivially-true assumes to path conditions (line 87-89: `| some true => (c', σ)`). So the assumption simply does not appear in the VCs.

Strata/Transform/CallElim.lean
MikaelMayer: Let's create an issue for it

StrataTest/Languages/Core/Examples/OldExpressions.lean
MikaelMayer: Please answer @aqjune-aws 's comment above too

MikaelMayer: 🤖 That's a good idea for a follow-up. Removing `Command.evalCall` and making the partial evaluator reject calls (requiring them to be eliminated by CallElim or ProcedureInlining first) would simplify the evaluator and make the pipeline contract clearer. However, it's a non-trivial change — `Command.evalCall` is referenced by the `EvalCallBodyRefinesContract` theorem in `StatementSemanticsProps.lean`, so the proof would need to be updated or restructured as well. I'd prefer to keep this out of the current PR and do it as a separate follow-up.

Strata/Transform/CallElim.lean
MikaelMayer: 🤖 Created issue #792 to track this: https://github.com/strata-org/Strata/issues/792

PR #761, opened 2026-04-06
Harden Python-to-Laurel translation: type assertions, bounds checks, and len() rejection

tautschnig:
This PR adds several checks to the Python-to-Laurel translation that catch common runtime error patterns earlier, at verification time:

1. **Type assertions for annotated assignments** — when a variable has a concrete type annotation (`int`, `str`, `bool`, `float`) and is assigned a value, emit an assertion that the value matches the declared type. This catches `None`-for-typed-variable bugs (e.g., `x: int = None`).

2. **Type assertions for dict-unpacking at call sites** — when `**dict` is expanded against a function signature, emit type assertions for each parameter with a concrete type, catching bugs like `{"DBSubnetGroupName": None}` unpacked to a `str` parameter.

3. **Bounds check for negative list indexing** — `xs[-n]` now emits `assert(len(xs) >= n)` and converts to `xs[len(xs) - n]`. Handles both `ConNeg` and `UnaryOp(USub, ConPos)` AST forms (the latter is what Python's parser actually produces for subscript indices). Skipped for dict types to avoid spurious coercion errors on `d[-1]`.

4. **Reject `len()` on Composite types without `__len__`** — calling `len()` on a Composite class (e.g., a DynamoDB Table) now emits a user error diagnostic.

5. **Shared type mapping infrastructure** — extracted `typeTester?` as a shared function for type-name-to-tester-predicate mapping (used by both kwargs assertions and annotated-assignment assertions), and `PythonIdent.toPyLauType?` as the canonical mapping from Python builtin identifiers to PyLauType string names (shared between `PySpecPipeline` and `PythonToLaurel`).


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1180,7 +1225,27 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
     match value.val with
-    | some value => translateAssign ctx target annotation value md
+    | some value =>
+      let (ctx, stmts) ← translateAssign ctx target annotation value md
+      -- Emit type assertion for concrete type annotations (int, str, bool, float).
+      -- This catches bugs like `x: int = None` where None is not a valid int.
+      let typeAssert := match target with
+        | .Name _ n _ =>
+          let annStr := pyExprToString annotation
+          let tester? := match annStr with
```
ssomayyajula: Can we refactor this and `tester?` above into a top-level function so they don't drift apart?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -589,7 +589,23 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
     let index ←  match slice with
       | .Slice _ start stop step => translateSlice ctx start.val stop.val step.val
       | _ => translateExpr ctx slice
-    return mkStmtExprMd (.StaticCall "Any_get" [dictOrList, index])
+    -- Emit bounds check for negative integer indices (e.g., xs[-1])
+    -- and convert to positive index: xs[-n] becomes xs[len(xs) - n]
+    let (boundsAssert, index) := match slice with
+      | .Constant _ (.ConNeg _ n) _ =>
```
ssomayyajula: Nit: is it possible to intercept calls to `__neg__` and not just constants or do you think that's out of scope fo rnow?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -589,7 +589,23 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+        -- xs[-n] requires len(xs) >= n; access becomes xs[len(xs) - n]
+        let listExpr := mkStmtExprMd (.StaticCall "Any..as_ListAny!" [dictOrList])
```
ssomayyajula: In case `dictOrList` is actually a dict, is applying the coercion `Any..as_ListAny!` problematic because it's technically ill-typed? It would be a nonsensical subscript anyway I guess...

ssomayyajula (changes requested):

Strata/Languages/Python/PythonToLaurel.lean
ssomayyajula: Resolving b/c of #777

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: The coercion is only applied inside a bounds-check guard for negative integer indices, which is inherently a list operation. For dicts, negative-integer subscripts are valid dict keys and wouldn't hit the .ConNeg branch.

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Done.

ssomayyajula (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -941,7 +974,23 @@ partial def translateCall (ctx : TranslationContext)
...
+    for arg in remainingParams do
+      let argType := match arg.tys with | [ty] => ty | _ => "Any"
+      if let some testerName := typeTester? argType then
+        let dictExpr := mkStmtExprMd (.StaticCall "Any..as_Dict!" [trans_dict])
+        let keyPresent := mkStmtExprMd (.StaticCall "DictStrAny_contains"
+          [dictExpr, mkStmtExprMd (.LiteralString arg.name)])
+        let val := DictStrAny_get_param trans_dict arg.name true
+        let isCorrectType := mkStmtExprMd (.StaticCall testerName [val])
+        let cond := mkStmtExprMd (.PrimitiveOp .Implies [keyPresent, isCorrectType])
+        typeAsserts := typeAsserts ++ [mkStmtExprMd (.Assert cond)]
```
MikaelMayer: `typeAsserts := typeAsserts ++ [...]` in a loop is O(n²) for lists. Consider prepending and reversing at the end:
```lean
typeAsserts := mkStmtExprMd (.Assert cond) :: typeAsserts
...
let typeAsserts := typeAsserts.reverse
```
Not a performance issue in practice (few parameters), but it's the idiomatic pattern for building lists in this codebase.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -589,7 +599,23 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
     let index ←  match slice with
       | .Slice _ start stop step => translateSlice ctx start.val stop.val step.val
       | _ => translateExpr ctx slice
-    return mkStmtExprMd (.StaticCall "Any_get" [dictOrList, index])
+    -- Emit bounds check for negative integer indices (e.g., xs[-1])
+    -- and convert to positive index: xs[-n] becomes xs[len(xs) - n]
+    let (boundsAssert, index) := match slice with
+      | .Constant _ (.ConNeg _ n) _ =>
+        -- xs[-n] requires len(xs) >= n; access becomes xs[len(xs) - n]
+        let listExpr := mkStmtExprMd (.StaticCall "Any..as_ListAny!" [dictOrList])
```
MikaelMayer: The bounds check unconditionally calls `Any..as_ListAny!` on `dictOrList`. If the subscript target is a dict with a negative integer key (unusual but syntactically valid Python), this would apply a list coercion to a dict. Consider guarding with a type check on `val`, or adding a comment explaining why this is safe (e.g., negative integer subscripts on dicts are not valid Python).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -941,7 +974,23 @@ partial def translateCall (ctx : TranslationContext)
...
     let kwargsArg := if funcDecl.kwargsName.isSome then [trans_dict] else []
-    emitCall (allArgs ++ kwargsArg)
+    -- Emit type assertions: if a key is present in the dict, its value
+    -- must match the declared parameter type. This catches {"key": None}
+    -- where the parameter type is str/int/bool/float.
+    let mut typeAsserts : List StmtExprMd := []
+    for arg in remainingParams do
+      let argType := match arg.tys with | [ty] => ty | _ => "Any"
+      if let some testerName := typeTester? argType then
+        let dictExpr := mkStmtExprMd (.StaticCall "Any..as_Dict!" [trans_dict])
```
MikaelMayer: `Any..as_Dict!` is called on `trans_dict` inside the loop for every parameter with a concrete type. Since `trans_dict` doesn't change across iterations, consider hoisting this outside the loop to avoid redundant coercion expressions in the generated Laurel.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1188,7 +1237,21 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
   -- Annotated assignment: x: int = expr or x: ClassName = ClassName(args) or self.field: int = expr
   | .AnnAssign _ target annotation value _ => do
     match value.val with
-    | some value => translateAssign ctx target annotation value md
+    | some value =>
+      let (ctx, stmts) ← translateAssign ctx target annotation value md
+      -- Emit type assertion for concrete type annotations (int, str, bool, float).
+      -- This catches bugs like `x: int = None` where None is not a valid int.
```
olivier-aws, line 1383: I'm surprised by this. In Python,
```
x: int = None

if (x is None):
	print(1)
else:
	print(0)
```
is valid and prints 1. Would we raise an assertion error in this case?

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [8128381](https://github.com/strata-org/Strata/pull/761/commits/81283810e594a82b8c25d15f4a493a4b89c48e0a).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [8128381](https://github.com/strata-org/Strata/pull/761/commits/81283810e594a82b8c25d15f4a493a4b89c48e0a).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [8128381](https://github.com/strata-org/Strata/pull/761/commits/81283810e594a82b8c25d15f4a493a4b89c48e0a).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Yes, this is intentional behavior: the assertion catches what mypy and other static type checkers would flag as a type error. `x: int = None` is a common bug pattern where the developer likely meant `x: Optional[int] = None`. The PR's purpose is specifically to add these hardening checks.

If we wanted to support `x: int = None` without a diagnostic, we'd need to treat all Python type annotations as `Optional[T]` by default.

tautschnig:
> Haven't looked at the PR but for the length check, just double checking you are not actually testing it only for syntactically negated argument but for every argument?

- This PR only handles the syntactic case (negative literals) as a hardening check
- PR #777 addresses the general case (any negative index at runtime) and will supersede this for the general case
- The two are complementary: this PR's static check catches the common xs[-1] pattern at translation time, while #777
handles the dynamic case.

ssomayyajula (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
atomb: Kiro gave me a more detailed finding on this, and I'm not sure #777 resolves it.

🤖 says:

Negative indexing bounds check only handles `ConNeg` literals, not `UnaryOp(USub, ConPos)`.

   `./Strata/Languages/Python/PythonToLaurel.lean:626-639`

   ```lean
   let (boundsAssert, index) := match slice with
     | .Constant _ (.ConNeg _ n) _ =>
       ...
     | _ => (none, index)
   ```

   Python's AST can represent `-1` as either `Constant(ConNeg 1)` or `UnaryOp(USub, Constant(1))` depending on the parser version and context. The `Specs.lean` file already handles both forms in `extractIntBound`. If the Python parser ever produces the `UnaryOp` form for subscript indices, the bounds check would be silently skipped. Consider matching `UnaryOp(USub, ConPos n)` as well, or adding a comment explaining why it's not needed (e.g., if the Python-to-Ion parser normalizes to `ConNeg`).

tautschnig: This is spot-on, and I was fooled by believing we had test coverage in this PR, but actually Test 5 was passing because `Any_get`'s own precondition (`index >= 0`) fails for negative indices, not because of the bounds check we added.

Python's `ast.parse` always represents `-1` as `UnaryOp(USub, Constant(1))`, never as `Constant(-1)`. The Strata Python parser preserves this structure, producing `UnaryOp(USub, ConPos n)` in the Strata AST. The previous code only matched `.ConNeg`, which is produced when a Constant node has a negative integer value directly - a form that Python's parser never generates for subscript indices.

Fixed in the latest update, which now also has a genuine test.

Strata/Languages/Python/PythonToLaurel.lean
olivier-aws: Thanks, fine with this design choice then.

StrataTest/Languages/Python/expected_laurel/test_list.expected
```diff
@@ -1,33 +1,2 @@
...
-test_list.py(25, 7): ✅ pass - Check PAdd exception
-test_list.py(25, 7): ✅ pass - Check PAdd exception
-test_list.py(25, 7): ✅ pass - Check PAdd exception
-test_list.py(25, 0): ✅ pass - assert_assert(298)_calls_Any_get_0
-test_list.py(25, 0): ✅ pass - assert_assert(298)_calls_Any_get_1
-test_list.py(25, 0): ✅ pass - assert_assert(298)_calls_Any_to_bool_2
-test_list.py(25, 0): ✅ pass - assert(298)
-DETAIL: 31 passed, 0 failed, 0 inconclusive
-RESULT: Analysis success
+DETAIL: Laurel to Core translation failed: [asserts are not YET supported in functions or contracts, asserts are not YET supported in functions or contracts, asserts are not YET supported in functions or contracts]
```
MikaelMayer: This is a regression: on `main` this test had 31 passing checks and `Analysis success`. Now it's an internal error because the bounds-check assertions emitted for `numbers[-2]`, `numbers[-5]`, and `numbers[-1]` land in module-level code where `Assert` nodes aren't supported by the Laurel-to-Core translation.

The negative-indexing bounds check in `translateExpr` needs to be skipped (or use a different mechanism) when the code is at module level rather than inside a function body.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1239,7 +1300,20 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+      let typeAssert := match target with
+        | .Name _ n _ =>
+          let annStr := pyExprToString annotation
+          match typeTester? annStr with
+          | some testerName =>
+            let varExpr := mkStmtExprMd (StmtExpr.Identifier n.val)
+            let cond := mkStmtExprMd (StmtExpr.StaticCall testerName [varExpr])
+            [mkStmtExprMdWithLoc (StmtExpr.Assert cond) md]
```
MikaelMayer: The type assertion for annotated assignments introduces new inconclusive results in several existing tests. For example, `name: str = acc.get_owner()` in `test_class_methods.py` now emits `assert(Any..isfrom_str(name))` which the verifier can't prove, because method return values on composite types aren't tracked precisely enough.

Consider gating this assertion so it's only emitted when the RHS is a literal or a simple expression whose type is already known to potentially mismatch (e.g., `None`), rather than unconditionally for every annotated assignment with a concrete type. That would avoid adding noise to tests where the assignment is clearly type-correct.

StrataTest/Languages/Python/tests/cbmc_expected.txt
```diff
@@ -33,4 +33,4 @@ test_if_elif.py.ion             SKIP
 test_variable_reassign.py.ion   SKIP
 test_datetime_now_tz.py.ion     SKIP
 test_timedelta_expr.py.ion      SKIP
-test_composite_return.py.ion    PASS
+test_composite_return.py.ion    SKIP
```
MikaelMayer: `test_composite_return.py.ion` changed from PASS to SKIP. This looks like a regression — the PR description doesn't mention this change. Is this intentional? If the new type assertions cause the CBMC pipeline to fail for this test, that should be investigated rather than skipped.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -606,7 +616,34 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+    -- Skip for dicts, where negative integer keys are valid dict lookups.
+    -- Note: Python's AST represents `-1` as UnaryOp(USub, Constant(1)),
+    -- not Constant(-1), so we must match both forms.
+    let valType := (inferExprType ctx val).toOption.getD PyLauType.Any
+    let isDictType := valType == PyLauType.DictStrAny
+    let negLitVal? := match slice with
+      | .Constant _ (.ConNeg _ n) _ => some n.val
+      | .UnaryOp _ (.USub _) (.Constant _ (.ConPos _ n) _) => some n.val
+      | _ => none
+    let (boundsAssert, index) := match negLitVal? with
```
MikaelMayer: The bounds check only fires for literal negative indices (`xs[-1]`, `xs[-2]`). A variable negative index like `xs[-n]` won't be caught, and also `xs[n]` where n is negative. Could you support all cases at once? That way we don't need to come back to this.

ssomayyajula (dismissed):

StrataTest/Languages/Python/DictNoneTest.lean
```diff
@@ -0,0 +1,116 @@
...
+#guard_msgs (drop info) in
+#eval withPython (warnOnSkip := false) fun pythonCmd => do
+  let program :=
+"def main() -> None:
+    x: int = None
+    y: int = 10
+    assert y == 10
+"
+  let diags ← processPythonFile pythonCmd (stringInputContext "test.py" program)
+  unless diags.any (fun d => isAssertionFailure d.message) do
```
MikaelMayer, line 59: Nit: I would rather have the assertion failure as part of the guard instead of that. It makes the test easier to understand. Especially to know what would the error message be, because it could as well be something else.

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: That's been covered by #777.

StrataTest/Languages/Python/DictNoneTest.lean
tautschnig: Fixed via [2bfc498](https://github.com/strata-org/Strata/pull/761/commits/2bfc498a0bbecfcc8335d1a7086c03a2800ccfef).

StrataTest/Languages/Python/expected_laurel/test_list.expected
tautschnig: Thanks, fixed in [f8cce9b](https://github.com/strata-org/Strata/pull/761/commits/f8cce9b06111154c5c35a22511f90636df496dbf).

StrataTest/Languages/Python/tests/cbmc_expected.txt
tautschnig: [f8cce9b](https://github.com/strata-org/Strata/pull/761/commits/f8cce9b06111154c5c35a22511f90636df496dbf) also addresses this as it had the same root cause.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1239,7 +1297,20 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+        | .Name _ n _ =>
+          let annStr := pyExprToString annotation
+          match typeTester? annStr with
+          | some testerName =>
+            let varExpr := mkStmtExprMd (StmtExpr.Identifier n.val)
+            let cond := mkStmtExprMd (StmtExpr.StaticCall testerName [varExpr])
+            [mkStmtExprMdWithLoc (StmtExpr.Assert cond) md]
+          | none => []
+        | _ => []
+      return withExceptionChecks ctx (ctx, stmts ++ typeAssert)
```
atomb, line 1399: Type assertion for `AnnAssign` reads the variable *after* assignment. The assertion `assert(Any..isfrom_int(x))` is emitted after `translateAssign` completes, reading the variable's current value. This is correct for simple cases (`x: int = None`), but if `translateAssign` produces multiple statements (e.g., for composite types with `__init__` calls), the assertion checks the final state, which may not correspond to the annotated type in all cases. Worth verifying this is intentional for composite-typed annotated assignments (though `typeTester?` only fires for `int/str/bool/float`, so composites are currently excluded).

atomb (commented):
One 🤖 comment.

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Now enforced via [c238633](https://github.com/strata-org/Strata/pull/761/commits/c238633e6426033eb316f9d39ba52cfc2ac2254c).

tautschnig:

tautschnig:

tautschnig:

tautschnig:

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -41,6 +41,17 @@ def reCompile := mk "re" "compile"
...
+/-- Map a PythonIdent to its PyLauType string name, if it's a recognized
+    builtin type. Used by both PySpecPipeline (for type extraction) and
+    PythonToLaurel (for type assertions). -/
+def PythonIdent.toPyLauType? (id : PythonIdent) : Option String :=
+  if id == PythonIdent.builtinsInt then some "int"
+  else if id == PythonIdent.builtinsStr then some "str"
+  else if id == PythonIdent.builtinsBool then some "bool"
+  else if id == PythonIdent.builtinsFloat then some "float"
+  else if id == PythonIdent.noneType then some "None"
+  else none
```
joehendrix: I think this should live in Specs/ToLaurel.lean.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1297,7 +1353,27 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+      let (ctx, stmts) ← translateAssign ctx target annotation value md
+      -- Emit type assertion for concrete type annotations (int, str, bool, float).
+      -- This catches bugs like `x: int = None` where None is not a valid int.
+      -- We assert on the variable after assignment. This is only safe when
+      -- translateAssign produces a simple assignment (1-2 statements); skip
+      -- the assertion for complex translations (e.g., composite __init__ calls)
+      -- where the variable state may not correspond to the annotated type.
+      let typeAssert := match target with
+        | .Name _ n _ =>
+          if stmts.length > 2 then []  -- complex translation, skip assertion
```
joehendrix: The `stmts.length > 2` threshold works today, but it's a fragile coupling to `translateAssign`'s output shape. Currently:
- 1 stmt: assign to existing variable
- 2 stmts: new variable (LocalVariable + Assign), or composite (New + `__init__`)
- 3+ stmts: tuple unpacking or other complex cases

If `translateAssign` ever grows a new 2-statement path where the variable doesn't yet hold the annotated type, this assertion would fire incorrectly. Consider either:
1. Documenting the expected statement counts here (so the invariant is visible), or
2. Having `translateAssign` return a signal (e.g., a flag) indicating whether a type assertion is safe, instead of inferring it from the statement count.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -298,6 +298,16 @@ def boolToAny (b: Bool) := mkStmtExprMd (.StaticCall "from_bool" [mkStmtExprMd (
 def AnyNone := mkStmtExprMd (.StaticCall "from_None" [])
 def Any_to_bool (b: StmtExprMd) := mkStmtExprMd (.StaticCall "Any_to_bool" [b])
 
+/-- Return the Laurel type-tester predicate name for a Python type annotation, if known.
+    E.g., `"int"` → `"Any..isfrom_int"`. Used to emit runtime type assertions. -/
+def typeTester? (typeName : String) : Option String :=
```
joehendrix, line 315: `typeTester?` and `PythonIdent.toPyLauType?` (in `Specs/Decls.lean:47`) both hardcode the same four builtin types (`int`, `str`, `bool`, `float`). If a fifth type is added, both must be updated in lockstep but there's no cross-reference connecting them.

Consider either:
1. Adding a comment to each function referencing the other (e.g., "Keep in sync with `PythonIdent.toPyLauType?` in Specs/Decls.lean"), or
2. Building `typeTester?` from `toPyLauType?`'s output so there's a single source of truth for which types are recognized, e.g.:
```lean
def typeTester? (typeName : String) : Option String :=
  -- Canonical list lives in PythonIdent.toPyLauType?
  match typeName with
  | "int"   => some "Any..isfrom_int"
  | "str"   => some "Any..isfrom_str"
  | "bool"  => some "Any..isfrom_bool"
  | "float" => some "Any..isfrom_float"
  | _       => none
```
The predicate name pattern `"Any..isfrom_" ++ typeName` could even be generated from the type string to avoid the parallel match arms.

Strata/Languages/Python/Specs/Decls.lean
tautschnig: Fixed in be3d8912fc.

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Addressed in be3d8912fc.

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Addressed in be3d8912fc.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -304,6 +304,18 @@ def boolToAny (b: Bool) := mkStmtExprMd (.StaticCall "from_bool" [mkStmtExprMd (
 def AnyNone := mkStmtExprMd (.StaticCall "from_None" [])
 def Any_to_bool (b: StmtExprMd) := mkStmtExprMd (.StaticCall "Any_to_bool" [b])
 
+/-- The set of PyLauType names that have runtime type-tester predicates
+    (`Any..isfrom_<type>`). Keep in sync with `PythonIdent.toPyLauType?`
+    in Specs/ToLaurel.lean. -/
+private def pyLauTypesWithTesters : List String := ["int", "str", "bool", "float"]
```
joehendrix, line 310: Maybe should be Std.HashSet String since this is just a membership test, but the number of types is small, so non-blocking.

joehendrix (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: I believe this would add an import and initialization overhead, so I'd rather leave it as-is.

joehendrix (approved):

ssomayyajula (approved):

PR #764, opened 2026-04-06
Refactor verify options and sanitize SMT-LIB filenames

shigoel:
## Summary

- Extract shared `verifyOptionsFlags` and `parseVerifyOptions` in StrataMain.lean, used by both `pyAnalyzeLaurel` and `verify` commands. This exposes all `VerifyOptions` fields (`--solver`, `--solver-timeout`, `--stop-on-first-error`) consistently across commands, eliminating duplication.
- Sanitize VC obligation labels for filesystem-safe `.smt2` filenames: replace parens, quotes, spaces, and slashes with underscores.
- Document `VerifyOptions` sync requirement with StrataMain flags.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,8 +78,8 @@ can be useful to provide different feedback for an empty list.
 ### Prefer Extrinsic Proofs
 
 Given the `median` function above, we may want to know that it always
-produces an `Int` given a list of even size. We could have included a
-precondition that the list has even size, avoiding the need for error
+produces an `Int` given a list of odd size. We could have included a
+precondition that the list has odd size, avoiding the need for error
 handling, but that would have precluded the possibility of providing
 useful feedback in the case of working with unfiltered input.
 
@@ -179,7 +179,7 @@ encouraged.
 
 Every major function and theorem must be documented (NOTE: use `/--
 ... -/` and NOT `/- ... -/`). Use backticks to format code or math
-expressions in the docstring (e.g., `x + y`).
+expressions in the docstring (e.g., ``/-- ... `x + y` ... -/``).
 
 Describe implementation-level details in single- (`--`) or multi-line
 comments (`/- .. -/`) interspersed with the code.


StrataMain.lean
```diff
@@ -147,15 +147,76 @@ def parseCheckLevel (pflags : ParsedFlags) : IO CheckLevel :=
...
+      | .none => exitFailure s!"Invalid solver timeout: '{s}'"
+  let noSolve := pflags.getBool "no-solve"
+  let removeIrrelevantAxioms ← match pflags.getString "remove-irrelevant-axioms" with
+    | .none => pure base.removeIrrelevantAxioms
+    | .some "off" => pure .Off
+    | .some "aggressive" => pure .Aggressive
+    | .some "precise" => pure .Precise
+    | .some s => exitFailure s!"Invalid remove-irrelevant-axioms mode: '{s}'. Must be 'off', 'aggressive', or 'precise'."
+  pure { base with
+    verbose := if pflags.getBool "verbose" then .normal else .quiet,
```
MikaelMayer: The `verbose` field always becomes `.quiet` when `--verbose` is absent, ignoring `base.verbose`. Same issue on line 209 with `checkMode`/`checkLevel`: `parseCheckMode`/`parseCheckLevel` return hardcoded defaults (`.deductive`/`.minimal`) instead of falling back to `base.checkMode`/`base.checkLevel`.

This contradicts the docstring: "Fields not present in the flags keep their base values."

For current callers this is harmless (the bases happen to match the hardcoded defaults), but it will silently break if a future caller passes a `base` with non-default `verbose`, `checkMode`, or `checkLevel`.

Consider:
```lean
verbose := if pflags.getBool "verbose" then .normal else base.verbose,
```
and threading `base.checkMode`/`base.checkLevel` as defaults into `parseCheckMode`/`parseCheckLevel`.

StrataMain.lean
```diff
@@ -147,15 +147,76 @@ def parseCheckLevel (pflags : ParsedFlags) : IO CheckLevel :=
...
-    takesArg := .arg "mode" }
-
-def checkLevelFlag : Flag :=
+    takesArg := .arg "mode" },
   { name := "check-level",
     help := s!"Check level: {CheckLevel.options}. Default: 'minimal'.",
-    takesArg := .arg "level" }
+    takesArg := .arg "level" },
+  { name := "verbose", help := "Enable verbose output." },
+  { name := "quiet", help := "Suppress warnings on stderr." },
```
MikaelMayer, line 165: The `--quiet` flag is declared here but `parseVerifyOptions` never reads it. For the `verify` command, passing `--quiet` is silently accepted and ignored. In `pyAnalyzeLaurel`, `quiet` is read separately (line 579) for the Laurel pipeline, but it doesn't affect `VerifyOptions`.

Either remove `quiet` from `verifyOptionsFlags` (since it's pyAnalyzeLaurel-specific) or wire it into `parseVerifyOptions` (e.g. mapping it to `VerboseMode.quiet`).

Strata/Languages/Core/Verifier.lean
```diff
@@ -123,6 +123,13 @@ open Lambda Strata.SMT
 
 public section
 
+/-- Replace characters that are problematic on common filesystems
+    (parens, quotes, spaces, path separators) with underscores or remove them. -/
+def sanitizeFilename (s : String) : String :=
```
MikaelMayer, line 131: Nit: backslash (`\`) is not handled. If this only targets Linux/macOS that's fine, but worth a brief note in the docstring if so.

StrataMain.lean
```diff
@@ -629,29 +681,19 @@ def pyAnalyzeLaurelCommand : Command where
       else pure coreProgram
 
     -- Verify using Core verifier
-    let checkMode ← parseCheckMode pflags
-    let checkLevel ← parseCheckLevel pflags
-    let noSolve := pflags.getBool "no-solve"
-    if noSolve && (pflags.getString "vc-directory").isNone && keepDir.isNone then
+    let pyAnalyzeBase : VerifyOptions :=
```
MikaelMayer, line 713: 👍 No action needed. The `pyAnalyzeBase` pattern — defining command-specific defaults and passing them to `parseVerifyOptions` — is a clean way to let each command customize its baseline while sharing the flag-parsing logic.

StrataMain.lean
shigoel: Thanks; fixed in https://github.com/strata-org/Strata/pull/764/commits/efdaa7da3557db35e614b2582986ec6c28af4dcb.

StrataMain.lean
shigoel: Thanks, fixed in https://github.com/strata-org/Strata/pull/764/commits/efdaa7da3557db35e614b2582986ec6c28af4dcb.

Strata/Languages/Core/Verifier.lean
shigoel: Thanks, fixed in https://github.com/strata-org/Strata/pull/764/commits/efdaa7da3557db35e614b2582986ec6c28af4dcb

StrataMain.lean
```diff
@@ -1086,45 +1132,23 @@ structure CommandGroup where
...
     let proceduresToVerify := pflags.getString "procedures" |>.map (·.splitToList (· == ','))
-    let opts : VerifyOptions :=
-      { VerifyOptions.default with
-        verbose := if verbose then .normal else .quiet,
-        checkOnly, typeCheckOnly, parseOnly, stopOnFirstError, uniqueBoundNames,
-        outputSarif, checkMode, checkLevel, solverTimeout,
-        vcDirectory := pflags.getString "vc-directory",
-        solver := pflags.getString "solver" |>.getD Core.defaultSolver }
+    let outputSarif := pflags.getBool "sarif" || pflags.getString "output-format" == some "sarif"
+    let opts ← parseVerifyOptions pflags
```
tautschnig: Behavioral change: the old `verifyCommand` defaulted to `verbose := .quiet` (i.e., `if verbose then .normal else .quiet`). The new code calls `parseVerifyOptions pflags` with the default `base = VerifyOptions.default`, where `verbose = .normal`. So `strata verify` without `--verbose` or `--quiet` now produces `.normal` output instead of `.quiet`.

If this is intentional (making verify more verbose by default), that's fine — but it should be called out in the PR description. If not, the fix is to pass a base with `.quiet`:

```lean
    let opts ← parseVerifyOptions pflags { VerifyOptions.default with verbose := .quiet }
```

StrataMain.lean
```diff
@@ -1086,45 +1132,23 @@ structure CommandGroup where
...
+    let opts := { opts with
+      checkOnly := pflags.getBool "check",
+      typeCheckOnly := pflags.getBool "type-check",
+      parseOnly := pflags.getBool "parse-only",
+      outputSarif := outputSarif || opts.outputSarif }
```
tautschnig: nit: `outputSarif` is computed redundantly. `parseVerifyOptions` already reads `pflags.getBool "sarif"` and sets `opts.outputSarif`. The local `outputSarif` variable adds the `--output-format sarif` check, which is verify-specific. The OR on line 1151 means `--sarif` is checked twice.

Consider simplifying to:
```lean
    let opts ← parseVerifyOptions pflags
    let opts := { opts with
      checkOnly := pflags.getBool "check",
      typeCheckOnly := pflags.getBool "type-check",
      parseOnly := pflags.getBool "parse-only",
      outputSarif := opts.outputSarif || pflags.getString "output-format" == some "sarif" }
```
This makes it clear that `parseVerifyOptions` handles `--sarif` and the local code only adds the `--output-format` path.

StrataMain.lean
```diff
@@ -133,29 +133,94 @@ def buildDialectFileMap (pflags : ParsedFlags) : IO Strata.DialectFileMap := do
...
+              else if pflags.getBool "quiet" then .quiet
+              else base.verbose,
+    solver := pflags.getString "solver" |>.getD base.solver,
+    solverTimeout,
+    checkMode, checkLevel,
+    stopOnFirstError := pflags.getBool "stop-on-first-error" || base.stopOnFirstError,
+    uniqueBoundNames := pflags.getBool "unique-bound-names" || base.uniqueBoundNames,
+    useArrayTheory := pflags.getBool "use-array-theory" || base.useArrayTheory,
+    removeIrrelevantAxioms,
+    outputSarif := pflags.getBool "sarif" || base.outputSarif,
```
tautschnig, line 225: Boolean flags like `stopOnFirstError`, `uniqueBoundNames`, `useArrayTheory`, `profile`, `skipSolver`, `outputSarif` use `|| base.X` — meaning once set to `true` in the base, they can never be turned off by the CLI. This is fine for the current callers (bases are mostly `false`), but the docstring says "Fields not present in the flags keep their base values," which could be read as implying the flag *overrides* the base.

Since boolean CLI flags can only express "on" (present) or "absent" (not present), there's no way to express "off" — so this OR-with-base pattern is the only reasonable approach. But it might be worth noting in the docstring that boolean flags can only *enable*, not *disable*, relative to the base.

Strata/Languages/Core/Verifier.lean
```diff
@@ -123,6 +123,14 @@ open Lambda Strata.SMT
...
+  s |>.replace "(" "_" |>.replace ")" "_"
+    |>.replace "\"" "" |>.replace "'" ""
+    |>.replace " " "_" |>.replace "/" "_"
+    |>.replace "\\" "_"
+
```
tautschnig: The function works via chained `String.replace` calls, which means it does multiple passes over the string. For the short labels used in practice this is fine, but a single-pass `String.map` (or fold) replacing each char would be cleaner and more extensible:

```lean
def sanitizeFilename (s : String) : String :=
  s.toList.filterMap (fun c =>
    if c == '"' || c == '\'' then none
    else if c == '(' || c == ')' || c == ' ' || c == '/' || c == '\\' then some '_'
    else some c) |> String.mk
```

Not a blocker — just a suggestion for maintainability if more characters need handling later.

tautschnig: No unit test for `sanitizeFilename`. A simple `#guard_msgs` or `#eval` test in `StrataTest/` would help prevent regressions, e.g.:

```lean
#guard Core.SMT.sanitizeFilename "foo(bar) baz/qux\\quux" == "foo_bar__baz_qux_quux"
#guard Core.SMT.sanitizeFilename "a\"b'c" == "abc"
```

StrataMain.lean
shigoel: Fix with commit ce1dc27ad.

StrataMain.lean
shigoel: Fix with commit a9bef2be6.

StrataMain.lean
shigoel: Fix with commit f76d56b7d.

Strata/Languages/Core/Verifier.lean
shigoel: Fix with commit edef484b8.

shigoel: Fix with commit 824dbcf15.

Strata/Languages/Core/Verifier.lean
```diff
@@ -124,6 +124,15 @@ open Lambda Strata.SMT
 
 public section
 
+/-- Replace characters that are problematic on common filesystems
+    (parens, quotes, spaces, path separators) with underscores or remove them.
+    Single-pass over the string. -/
+def sanitizeFilename (s : String) : String :=
```
fabiomadge, line 131: `sanitizeFilename` doesn't handle `< > : | ? *`, which are invalid on Windows (NTFS). Strata itself generates labels containing these characters (e.g. `<dead_branch: ...>` in `StatementEval.lean`). JVerify also runs on Windows in CI and works around this where it can by avoiding such characters in its own labels, but handling it in Strata is the right fix:

```lean
else if c == '(' || c == ')' || c == ' ' || c == '/' || c == '\\'
     || c == '<' || c == '>' || c == ':' || c == '|' || c == '?' || c == '*' then some '_'
```

Collisions aren't a concern since the `_{counterVal}` suffix ensures uniqueness.

StrataMain.lean
```diff
@@ -1195,45 +1243,22 @@ def transformCommand : Command where
 def verifyCommand : Command where
```
fabiomadge, line 1241: `verify` now accepts `--no-solve` via `verifyOptionsFlags`, but unlike `pyAnalyzeLaurel` (line 710) there is no check that `--vc-directory` is also set. Without it, SMT files are written to a temp dir that is deleted immediately — the user gets all-unknown results with no files preserved. Consider adding a similar guard here, or at least in `parseVerifyOptions` itself so all callers benefit.

Strata/Languages/Core/Verifier.lean
shigoel: Thanks! I always forget the NTFS case. Addressed in https://github.com/strata-org/Strata/pull/764/commits/ca4d516d36a008270a1e2cb922f54e599dbab084.

StrataMain.lean
shigoel: Indeed! I did that refactor to `pyAnalyzeLaurel` and forgot to address this in this PR. Thanks! Fixed in https://github.com/strata-org/Strata/pull/764/commits/ca4d516d36a008270a1e2cb922f54e599dbab084.

StrataMain.lean
```diff
@@ -134,29 +134,100 @@ def buildDialectFileMap (pflags : ParsedFlags) : IO Strata.DialectFileMap := do
...
+    stopOnFirstError := pflags.getBool "stop-on-first-error" || base.stopOnFirstError,
+    uniqueBoundNames := pflags.getBool "unique-bound-names" || base.uniqueBoundNames,
+    useArrayTheory := pflags.getBool "use-array-theory" || base.useArrayTheory,
+    removeIrrelevantAxioms,
+    outputSarif := pflags.getBool "sarif" || base.outputSarif,
+    profile := pflags.getBool "profile" || base.profile,
+    skipSolver,
+    alwaysGenerateSMT := noSolve || base.alwaysGenerateSMT,
+    vcDirectory
+  }
```
tautschnig, line 231: nit: The `--no-solve` error message says `"--no-solve requires --vc-directory to specify where SMT files are stored."` For `pyAnalyzeLaurel`, `--keep-all-files` is also a valid alternative (it sets `vcDirectory` in the base before calling `parseVerifyOptions`). The message is technically correct (the user would only see it if neither flag is provided), but could be slightly confusing for `pyAnalyzeLaurel` users who don't know about `--vc-directory`.

Consider either:
- Keeping as-is (the message is accurate for `verify` and the `pyAnalyzeLaurel` base handles `--keep-all-files` transparently)
- Or making the message slightly more generic: `"--no-solve requires a VC output directory (--vc-directory or equivalent)."`

Not a blocker.

shigoel: I'll keep it as-is, thanks.

PR #765, opened 2026-04-06
Fix augmented assign Python->Laurel translation

thanhnguyen-aws:
*Description of changes:* This PR fixes augmented assign Python->Laurel translation for cases where the left-hand-side is a subscription. Previously, the translation only can handle augmented assign statement when left-hand-side is a variable.

How: Convert the augmented assign statement into its equivalent assign expression in Python AST first, then call the translator for the normal assign statement.

### DIFF
--- a/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
+++ b/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
@@ -542,10 +542,10 @@ def translateProcedure (proc : Procedure) : TranslateM Core.Procedure := do
   -- Translate preconditions
   let preconditions ← translateChecks proc.preconditions "requires"
 
-  -- Translate postconditions for Opaque bodies
+  -- Translate postconditions for Opaque and Abstract bodies
   let postconditions : ListMap Core.CoreLabel Core.Procedure.Check ←
     match proc.body with
-    | .Opaque postconds _ _ =>
+    | .Opaque postconds _ _ | .Abstract postconds =>
         translateChecks postconds "postcondition"
     | _ => pure []
   let modifies : List Core.Expression.Ident := []
@@ -553,7 +553,12 @@ def translateProcedure (proc : Procedure) : TranslateM Core.Procedure := do
     match proc.body with
     | .Transparent bodyExpr => translateStmt proc.outputs bodyExpr
     | .Opaque _postconds (some impl) _ => translateStmt proc.outputs impl
-    | _ => pure [Core.Statement.assume "no_body" (.const () (.boolConst false)) mdWithUnknownLoc]
+    | _ =>
+      -- Bodiless procedure: assume postconditions so that verification of the
+      -- procedure itself passes trivially, and inlining only introduces the
+      -- postconditions as assumptions (not the unsound `assume false`).
+      pure (postconditions.map fun (label, check) =>
+        Core.Statement.assume label check.expr mdWithUnknownLoc)
   -- Wrap body in a labeled block so early returns (exit) work correctly.
   let body : List Core.Statement := [.block "$body" bodyStmts mdWithUnknownLoc]
   let spec : Core.Procedure.Spec := { modifies, preconditions, postconditions }
@@ -563,7 +568,7 @@ def translateInvokeOnAxiom (proc : Procedure) (trigger : StmtExprMd)
     : TranslateM (Option Core.Decl) := do
   let model := (← get).model
   let postconds := match proc.body with
-    | .Opaque postconds _ _ => postconds
+    | .Opaque postconds _ _ | .Abstract postconds => postconds
     | _ => []
   if postconds.isEmpty then return none
   -- All input param names become bound variables.
--- a/StrataTest/Languages/Laurel/Examples/Fundamentals/T8c_BodilessInlining.lean
+++ b/StrataTest/Languages/Laurel/Examples/Fundamentals/T8c_BodilessInlining.lean
@@ -1,0 +1,56 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.SimpleAPI
+
+/-! # Bodiless Procedure Inlining Test
+
+Verifies that inlining a bodiless Laurel procedure does not introduce
+`assume false`. Previously, bodiless procedures had `assume false` as their
+body, so inlining would make everything after the call trivially provable.
+Now the body assumes the postconditions instead, so `assert false` after
+the inlined call is correctly rejected. -/
+
+namespace Strata.Laurel.BodilessInliningTest
+
+private def laurelSource := "
+procedure bodilessProcedure() returns (r: int)
+  ensures r > 0
+;
+
+procedure caller() {
+  var x: int := bodilessProcedure();
+  assert x > 0;
+  assert false
+};
+"
+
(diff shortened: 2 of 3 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1379,18 +1379,11 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-      | .Add _      => mkStmtExprMd (StmtExpr.StaticCall "PAdd"      [targetExpr, valueExpr])
-      | .Sub _      => mkStmtExprMd (StmtExpr.StaticCall "PSub"      [targetExpr, valueExpr])
-      | .Mult _     => mkStmtExprMd (StmtExpr.StaticCall "PMul"      [targetExpr, valueExpr])
-      | .FloorDiv _ => mkStmtExprMd (StmtExpr.StaticCall "PFloorDiv" [targetExpr, valueExpr])
-      | .Mod _      => mkStmtExprMd (StmtExpr.StaticCall "PMod"      [targetExpr, valueExpr])
-      | _           => mkStmtExprMd .Hole
-    let assignStmt := mkStmtExprMdWithLoc (StmtExpr.Assign [targetExpr] rhs) md
-    return (ctx, [assignStmt])
+  | .AugAssign sr target op value => do
+    let rhs : Python.expr SourceRange := .BinOp default target op value
```
tautschnig: Using `default` for the `BinOp` source range means error messages from translating the synthetic `BinOp` (e.g., unsupported operator) will point to a zero/empty source location instead of the original `AugAssign` location. Consider using `sr` (the `AugAssign`'s own source range) instead:

```suggestion
    let rhs : Python.expr SourceRange := .BinOp sr target op value
```

This way, if the `BinOp` translation fails (e.g., for an unsupported operator like `BitAnd`), the error message will point to the correct source location.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1379,18 +1379,11 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+    let pyNormalAssign : Python.stmt SourceRange :=
+      .Assign sr {val:= #[target], ann:= target.ann} rhs {val:= none, ann:= default}
```
tautschnig: Similarly, `{val:= none, ann:= default}` uses `default` for the type-annotation's source range. This is fine since the field is `none` (no type annotation), but for consistency you could use `sr`:

```suggestion
      .Assign sr {val:= #[target], ann:= target.ann} rhs {val:= none, ann:= sr}
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1379,18 +1379,11 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+    translateStmt ctx pyNormalAssign
```
tautschnig: In Python semantics, `l[f()] += 1` evaluates `f()` once, while the desugared `l[f()] = l[f()] + 1` would evaluate it twice.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1409,18 +1409,27 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+        | target :: slices =>
+            let tempVars := (List.range slices.length).map λ n => s!"augAssignTempVar_{sr.start}_{n}"
+            let tempVarExprs: List (Python.expr SourceRange) := tempVars.map
+                (fun var => .Name default {val:= var, ann:= default} default)
+            let target: Python.expr SourceRange := tempVarExprs.foldl (λ s t =>
+              .Subscript default s t default) target
+            (target, tempVars, slices)
+        | _ =>  (target, [], [])
+      | _ => (target, [], [])
+    let slices ← slices.mapM (translateExpr ctx)
```
tautschnig, line 1481: The logic here is dense — it destructures the subscript, creates temp vars, and reconstructs the target in a single `match` expression returning a triple. A brief comment explaining the strategy would help future readers:

```lean
    -- For subscript targets like l[i][j], we extract the index expressions,
    -- bind them to temp variables (to avoid double-evaluation), and
    -- reconstruct the target using the temp var names.
```

Also, the inner `match getSubscriptList target with | target :: slices =>` shadows the outer `target` binding, which works but can be confusing. Consider using a different name like `base :: slices`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1409,18 +1409,27 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-    return (ctx, [assignStmt])
+  | .AugAssign sr target op value => do
+    let (target, tempVars, slices) := match target with
+      | .Subscript _ _ _ _ =>
+        match getSubscriptList target with
+        | target :: slices =>
+            let tempVars := (List.range slices.length).map λ n => s!"augAssignTempVar_{sr.start}_{n}"
+            let tempVarExprs: List (Python.expr SourceRange) := tempVars.map
+                (fun var => .Name default {val:= var, ann:= default} default)
+            let target: Python.expr SourceRange := tempVarExprs.foldl (λ s t =>
```
tautschnig, line 1476: nit: The reconstructed subscript nodes use `default` for all source ranges and `expr_context`. While this works for translation, it means any error originating from these synthetic nodes will have no source location. Consider using `sr` consistently:

```suggestion
            let tempVarExprs: List (Python.expr SourceRange) := tempVars.map
                (fun var => .Name sr {val:= var, ann:= sr} (.Load sr))
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1409,18 +1409,27 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-      | .Sub _      => mkStmtExprMd (StmtExpr.StaticCall "PSub"      [targetExpr, valueExpr])
-      | .Mult _     => mkStmtExprMd (StmtExpr.StaticCall "PMul"      [targetExpr, valueExpr])
-      | .FloorDiv _ => mkStmtExprMd (StmtExpr.StaticCall "PFloorDiv" [targetExpr, valueExpr])
-      | .Mod _      => mkStmtExprMd (StmtExpr.StaticCall "PMod"      [targetExpr, valueExpr])
-      | _           => mkStmtExprMd .Hole
-    let assignStmt := mkStmtExprMdWithLoc (StmtExpr.Assign [targetExpr] rhs) md
-    return (ctx, [assignStmt])
+  | .AugAssign sr target op value => do
+    let (target, tempVars, slices) := match target with
+      | .Subscript _ _ _ _ =>
```
tautschnig, line 1470: The temp var name `augAssignTempVar_{sr.start}_{n}` uses the byte offset of the `AugAssign` statement for uniqueness. This is fine for single-file analysis, but could theoretically collide if two files have `AugAssign` at the same byte offset and are analyzed together. Since the `$` prefix convention is used elsewhere for internal names, consider `$augAssign_{sr.start}_{n}` to make it clearly non-user-visible and avoid any potential clash with user variables.

thanhnguyen-aws: I think those tempVars are local variables, so it is ok as long as the collision does not occur within the function body. I added `$` before the names.

PR #767, opened 2026-04-07
Add strata transform command and fix Core round-trip bugs

shigoel:
### New `transform` CLI Command
- Add `strata transform` CLI command that reads a `.core.st` file, applies one or more named transforms left to right, and prints the result. Available passes: `inlineProcedures`, `loopElim`, `callElim`, `filterProcedures`, `removeIrrelevantAxioms`.
- Changes to`Examples/run_examples.sh` to include transform tests.                                                          

 ### CLI Infrastructure                                                                                                                               
 - Refactor `ParsedFlags` from two HashMaps to a single ordered Array, preserving command-line flag order needed for per-pass argument binding.

### Core AST -> CST Fixes:
- Make `Core.formatProgram` emit `program Core;` header so its output is re-parseable **This is the cause of this huge diff; many `#guard_msgs` changed.**
- Fix block scoping bug in CSTtoAST parser: `block_statement` was leaking variable bindings to the parent scope, causing wrong de Bruijn index resolution after named blocks

### Change in `inlineProcedures` transform:
- Fix undeclared fvar in `inlineProcedures`: output parameters were initialized from temp fvars that were never declared; replaced with nondet init and removed redundant havoc

### DIFF
--- a/Strata/SimpleAPI.lean
+++ b/Strata/SimpleAPI.lean
@@ -247,15 +247,49 @@ When `none`, all calls are inlined.
 structure Core.InlineTransformOptions where
   doInline : Option (String → Core.Transform.CachedAnalyses → Bool) := none
 
+/-- A single named transform pass with its arguments. -/
+inductive Core.TransformPass where
+  | inlineProcedures (opts : Core.InlineTransformOptions := {})
+  | loopElim
+  | callElim
+  | filterProcedures (procs : List String)
+  | removeIrrelevantAxioms (funcs : List String)
+
+/-- Apply a single pass inside a running `CoreTransformM` context. -/
+private def Core.applyPass (program : Core.Program) (pass : Core.TransformPass)
+    : Core.Transform.CoreTransformM Core.Program := do
+  match pass with
+  | .inlineProcedures opts =>
+    let pred := opts.doInline.getD (fun _ _ => true)
+    let (_, prog) ← Core.Transform.runProgram (coreInlineCallCmd (doInline := pred)) program
+    return prog
+  | .loopElim =>
+    pure (Core.loopElim program)
+  | .callElim =>
+    let (_, prog) ← Core.Transform.runProgram coreCallElimCmd program
+    return prog
+  | .filterProcedures procs =>
+    Core.FilterProcedures.run program procs
+  | .removeIrrelevantAxioms funcs =>
+    Core.IrrelevantAxioms.run program funcs
+
+/-- Run a chain of transform passes on a Core program.  All passes share a
+    single `CoreTransformState`, so fresh variable counters accumulate across
+    passes and cached analyses (e.g., call graphs) can be reused. -/
+def Core.runTransforms (p : Core.Program) (passes : List Core.TransformPass)
+    : Except String Core.Program :=
+  Core.Transform.run p (fun prog => do
+    let mut program := prog
+    for pass in passes do
+      program ← Core.applyPass program pass
+    return program)
+
 /--
 Transform a Core program to inline some or all procedure calls.
 -/
 def Core.inlineProcedures (p : Core.Program) (opts : Core.InlineTransformOptions)
     : Except String Core.Program :=
-  let pred := opts.doInline.getD (fun _ _ => true)
-  Core.Transform.run p (fun prog => do
-    let (_, prog) ← Core.Transform.runProgram (coreInlineCallCmd (doInline := pred)) prog
-    return prog)
+  Core.runTransforms p [.inlineProcedures opts]
 
 /--
 Transform a Core program to replace each loop with assertions and assumptions about
@@ -269,9 +303,7 @@ Transform a Core program to replace each procedure call with assertions and
 assumptions about its contract.
 -/
 def Core.callElimUsingContract (p : Core.Program) : Except String Core.Program :=
-  Core.Transform.run p (fun prog => do
-    let (_, prog) ← Core.Transform.runProgram coreCallElimCmd prog
-    return prog)
+  Core.runTransforms p [.callElim]
 
 /--
 Transform a Core program to keep only the named procedures and their
@@ -279,7 +311,7 @@ transitive callees, removing everything else.
 -/
 def Core.filterProcedures (p : Core.Program) (targetProcs : List String)
     : Except String Core.Program :=
-  Core.Transform.run p (fun prog => Core.FilterProcedures.run prog targetProcs)
+  Core.runTransforms p [.filterProcedures targetProcs]
 
 /--
 Transform a Core program to remove axiom declarations that are irrelevant
@@ -287,7 +319,7 @@ to the named functions (based on call graph analysis).
 -/
 def Core.removeIrrelevantAxioms (p : Core.Program) (functions : List String)
     : Except String Core.Program :=
-  Core.Transform.run p (fun prog => Core.IrrelevantAxioms.run prog functions)
+  Core.runTransforms p [.removeIrrelevantAxioms functions]
 
 /-! ### Analysis of Core programs -/
 
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -1163,38 +1163,34 @@ def transformCommand : Command where
     | .error msg =>
       exitFailure msg
     | .ok initProgram =>
-      -- Apply passes left to right
-      let mut program := initProgram
+      -- Validate and convert pass configs to TransformPass values
+      let mut passes : List Strata.Core.TransformPass := []
       for pc in passConfigs do
         match pc.name with
         | "inlineProcedures" =>
           let opts : Strata.Core.InlineTransformOptions :=
             if pc.procedures.isEmpty then {}
             else { doInline := some (fun name _ => name ∈ pc.procedures) }
-          match Strata.Core.inlineProcedures program opts with
-          | .ok p => program := p
-          | .error e => exitFailure s!"inlineProcedures failed: {e}"
+          passes := passes ++ [.inlineProcedures opts]
         | "loopElim" =>
-          program := Strata.Core.loopElimUsingContract program
+          passes := passes ++ [.loopElim]
         | "callElim" =>
-          match Strata.Core.callElimUsingContract program with
-          | .ok p => program := p
-          | .error e => exitFailure s!"callElim failed: {e}"
+          passes := passes ++ [.callElim]
         | "filterProcedures" =>
           if pc.procedures.isEmpty then
             exitFailure "filterProcedures requires --procedures"
-          match Strata.Core.filterProcedures program pc.procedures with
-          | .ok p => program := p
-          | .error e => exitFailure s!"filterProcedures failed: {e}"
+          passes := passes ++ [.filterProcedures pc.procedures]
         | "removeIrrelevantAxioms" =>
           if pc.functions.isEmpty then
             exitFailure "removeIrrelevantAxioms requires --functions"
-          match Strata.Core.removeIrrelevantAxioms program pc.functions with
-          | .ok p => program := p
-          | .error e => exitFailure s!"removeIrrelevantAxioms failed: {e}"
+          passes := passes ++ [.removeIrrelevantAxioms pc.functions]
         | other =>
           exitFailure s!"Unknown pass '{other}'. Valid passes: {validPasses}."
-      IO.print (Core.formatProgram program)
+      -- Run all passes in a single CoreTransformM chain so fresh variable
+      -- counters accumulate and cached analyses are reused across passes.
+      match Strata.Core.runTransforms initProgram passes with
+      | .ok program => IO.print (Core.formatProgram program)
+      | .error e => exitFailure s!"Transform failed: {e}"
 
 def verifyCommand : Command where
   name := "verify"


Examples/expected/DoubleTwice.callElim.core.st
```diff
@@ -0,0 +1,21 @@
...
+spec {
+  ensures [double_correct]: result == n * 2;
+  } {
+  result := n + n;
+  };
+procedure TestProc (x : int) returns (output : int)
+spec {
+  ensures [testProc_result]: output == x * 4;
+  } {
+  var tmp_arg_3 : int := x;
```
shigoel, line 13: To the reviewers: Here's a Strata generated variable with a `tmp_arg` prefix. We're running the risk of name clashes here.

Examples/expected/DoubleTwice.inlineProcedures.core.st
```diff
@@ -0,0 +1,25 @@
...
+spec {
+  ensures [double_correct]: result == n * 2;
+  } {
+  result := n + n;
+  };
+procedure TestProc (x : int) returns (output : int)
+spec {
+  ensures [testProc_result]: output == x * 4;
+  } {
+  Double$inlined: {
```
shigoel, line 13: To the reviewers: Note that duplicate labels at the same scope level are intentionally allowed. However, Strata Core's type checker (StatementType.lean) will reject label shadowing (i.e., if an outer and inner block have the same label) to make `exit` semantics cleaner.

Examples/expected/LoopSimple.loopElim.core.st
```diff
@@ -0,0 +1,50 @@
...
+        loop_havoc_0: {
+          havoc sum;
+          havoc i;
+          }
+        arbitrary_iter_assumes_0: {
+          assume [assume_guard_0]: i < n;
+          assume [assume_invariant_0_0]: i <= n;
+          assume [assume_invariant_0_1]: i * (i - 1) div 2 == sum;
+          }
+        var $__loop_measure_0 : int;
```
shigoel, line 29: To the reviewers: Internal variable name exposed here.

shigoel:
Adding some comments to this PR because being able to inspect the transformed Core programs has raised some potentially interesting issues. Their fix can be out of scope of this PR, though I note that I did fix dangling fvars in inlineProcedures.

StrataMain.lean
```diff
@@ -1083,6 +1099,71 @@ structure CommandGroup where
...
+      exitFailure msg
+    | .ok initProgram =>
+    -- Apply passes left to right
+    let mut program := initProgram
+    for pass in passes do
+      match pass with
+      | "inlineProcedures" =>
+        let opts : Strata.Core.InlineTransformOptions :=
+          if procedures.isEmpty then {}
+          else { doInline := some (fun name _ => name ∉ procedures) }
```
tautschnig: The semantics of `--procedures` with `inlineProcedures` is inverted relative to `filterProcedures`: here it means "inline everything *except* these procedures" (i.e., keep them as call targets), while `filterProcedures` means "keep *only* these procedures." This is correct behavior but could be confusing for users.

Consider either:
- Documenting this in the `--procedures` help text (e.g., "For inlineProcedures: procedures to *preserve* (not inline). For filterProcedures: procedures to *keep*.")
- Or using a separate flag like `--keep` for `inlineProcedures`

StrataMain.lean
```diff
@@ -1083,6 +1099,71 @@ structure CommandGroup where
...
+        | .ok p => program := p
+        | .error e => exitFailure s!"filterProcedures failed: {e}"
+      | "removeIrrelevantAxioms" =>
+        if functions.isEmpty then
+          exitFailure "removeIrrelevantAxioms requires --functions"
+        match Strata.Core.removeIrrelevantAxioms program functions with
+        | .ok p => program := p
+        | .error e => exitFailure s!"removeIrrelevantAxioms failed: {e}"
+      | other =>
+        exitFailure s!"Unknown pass '{other}'. Valid passes: inlineProcedures, loopElim, callElim, filterProcedures, removeIrrelevantAxioms."
```
tautschnig: nit: The valid-passes list is duplicated in three places (the `--pass` help text on line 1108, the empty-passes error on line 1121, and this unknown-pass error). If a new pass is added, all three must be updated in sync. Consider extracting a shared constant.

Examples/run_examples.sh
```diff
@@ -28,4 +45,80 @@ for test_file in *.st; do
...
+    # Read extra flags (e.g. --procedures) from .args sidecar file if present
+    extra_flags=""
+    args_file="expected/${transform_base}.core.args"
+    if [ -f "$args_file" ]; then
+        extra_flags=$(cat "$args_file")
+    fi
+
+    # 1. Check transform output matches the .core.st file
+    tmp_transform=$(mktemp)
+    (cd .. && lake exe strata transform "Examples/${source_file}" $pass_flags $extra_flags) > "$tmp_transform"
```
tautschnig: The exit code of `strata transform` is not checked. If the command fails (e.g., parse error), the output file will be empty or contain error text, and the `diff` will report a mismatch — but the error message will say "Transform output does not match expected" rather than "Transform command failed." Consider checking the exit code explicitly, or at minimum redirecting stderr into the temp file so it's visible in the diff output.

Strata/SimpleAPI.lean
```diff
@@ -271,6 +273,26 @@ def Core.callElimUsingContract (p : Core.Program) : Except String Core.Program :
...
+    return filtered)
+
+/--
+Transform a Core program to remove axiom declarations that are irrelevant
+to the named functions (based on call graph analysis).
+-/
+def Core.removeIrrelevantAxioms (p : Core.Program) (functions : List String)
+    : Except String Core.Program :=
+  Core.Transform.run p (fun prog => do
+    let pruned ← Core.IrrelevantAxioms.run prog functions
```
tautschnig: nit: Both `filterProcedures` and `removeIrrelevantAxioms` follow the same pattern: `Core.Transform.run p (fun prog => do let x ← SomeModule.run prog args; return x)`. The `let x ← ...; return x` can be simplified to just the monadic expression:

```lean
def Core.filterProcedures (p : Core.Program) (targetProcs : List String)
    : Except String Core.Program :=
  Core.Transform.run p (fun prog => Core.FilterProcedures.run prog targetProcs)
```

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -1230,10 +1230,11 @@ def Core.formatProgram (ast : Core.Program)
   }
   let formatted := Std.Format.joinSep (cmds.map fun cmd =>
     (mformat (ArgF.op cmd.toAst) ctx state).format) ""
+  let header : Std.Format := "program Core;\n\n"
```
tautschnig, line 1233: The `program Core;` header is hardcoded. This is fine for now since `formatProgram` is Core-specific, but if other dialects ever need similar formatting, this would need to be parameterized. A brief comment noting this is Core-specific would help future readers.

shigoel: Ack. I'll read this to mean "no action item", which including adding a comment. The function lives in the Core module, so I think adding comments pertaining to other dialects here won't be too helpful.

Strata/Transform/ProcedureInlining.lean
```diff
@@ -234,13 +234,12 @@ def inlineCallCmd
         -- `init outN` is not necessary because calls are only allowed to use
         -- already declared variables (per Core.typeCheck)
 
-        -- Create a fresh var statement for each LHS
+        -- Declare each renamed output parameter with a nondet init.
```
MikaelMayer, line 237: The pseudo-code comment above (lines 229-230) still says `init out1 : ty := <placeholder>` but the code now uses `nondet` initialization. Consider updating to `init out1 : ty := nondet` to match the implementation.

StrataMain.lean
```diff
@@ -1083,6 +1099,71 @@ structure CommandGroup where
...
+      takesArg := .arg "procs" },
+    { name := "functions",
+      help := "Comma-separated function names (used by removeIrrelevantAxioms).",
+      takesArg := .arg "funcs" }]
+  help := "Apply one or more transforms to a Core program and print the result."
+  callback := fun v pflags => do
+    let file := v[0]
+    let passes := pflags.getRepeated "pass"
+    if passes.isEmpty then
+      exitFailure "No --pass specified. Valid passes: inlineProcedures, loopElim, callElim, filterProcedures, removeIrrelevantAxioms."
```
MikaelMayer: The valid-passes list is repeated verbatim in three places (flag help string, empty-passes error, unknown-pass error). If a new pass is added, all three must be updated in sync. Consider extracting a `validPasses` constant (e.g. a `List String`) and deriving the help/error strings from it.

StrataMain.lean
```diff
@@ -1083,6 +1099,71 @@ structure CommandGroup where
...
+  name := "transform"
+  args := [ "file" ]
+  flags := [
+    { name := "pass",
+      help := "Transform pass to apply (repeatable, applied left to right). \
+               Valid passes: inlineProcedures, loopElim, callElim, filterProcedures, removeIrrelevantAxioms.",
+      takesArg := .repeat "name" },
+    { name := "procedures",
+      help := "Comma-separated procedure names (used by filterProcedures and inlineProcedures).",
+      takesArg := .arg "procs" },
```
MikaelMayer: The `--procedures` help says it is "used by filterProcedures and inlineProcedures" but doesn't clarify the semantics. For `inlineProcedures`, the named procedures are *excluded* from inlining (kept as-is); for `filterProcedures`, they are the procedures to *retain*. Both mean "keep these procedures", which is consistent, but a reader might expect `--procedures` to mean "inline these" for `inlineProcedures`. A brief clarification like "procedures to preserve (not inline / not filter out)" would help.

StrataMain.lean
```diff
@@ -1083,6 +1099,71 @@ structure CommandGroup where
...
+      | .ok r => pure r
+      | .error msgs =>
+        for e in msgs do println! s!"Error: {← e.toString}"
+        exitFailure s!"{msgs.size} parse error(s)"
+    match Strata.genericToCore pgm with
+    | .error msg =>
+      exitFailure msg
+    | .ok initProgram =>
+    -- Apply passes left to right
+    let mut program := initProgram
```
MikaelMayer: The `match ... | .ok initProgram =>` arm continues with the rest of the function body at the same indentation level as the `| .error` arm. This works but is a bit unusual — `let mut program := initProgram` and everything after it is inside the `.ok` branch. An early-return pattern (like the parse-error case above) would make the control flow clearer.

StrataMain.lean
shigoel: Thanks, addressed in https://github.com/strata-org/Strata/pull/767/commits/56645eaebcd31dde95d7029c5076c216287ed3b4. 
Indeed, `--procedures` is interpreted slightly differently for these two passes, but well, it's documented:
For filterProcedures: procedures to keep. 
For inlineProcedures: procedures to inline.

StrataMain.lean
shigoel: Addressed in https://github.com/strata-org/Strata/pull/767/commits/56645eaebcd31dde95d7029c5076c216287ed3b4, thanks.

StrataTest/Languages/Core/Tests/NestedVarScopingTest.lean
```diff
@@ -54,6 +56,52 @@ info: ok: procedure test (cond : bool, x : int, y : int) returns ()
...
+procedure test() returns () {
+  var x : int;
+  x := 1;
+  blk: {
+    var y : int;
+    y := 2;
+    x := y;
+  }
+  assert [use_x]: x == 2;
+};
```
aqjune-aws, line 80: In which file of this pull request was this bug fixed?

Examples/run_examples.sh
shigoel: Nice catch, thanks. Addressed in https://github.com/strata-org/Strata/pull/767/commits/ae2661c83ca4019cd5f66854e21363978be9dc2f

Strata/SimpleAPI.lean
shigoel: Thanks, addressed in https://github.com/strata-org/Strata/pull/767/commits/ae2661c83ca4019cd5f66854e21363978be9dc2f

Strata/Transform/ProcedureInlining.lean
shigoel: Thanks, addressed in https://github.com/strata-org/Strata/pull/767/commits/318eb5d47c00abe0a21d520c51454bff3c6f2499

StrataMain.lean
shigoel: Done in https://github.com/strata-org/Strata/commit/56645eaebcd31dde95d7029c5076c216287ed3b4.

StrataMain.lean
shigoel: Done in https://github.com/strata-org/Strata/pull/767/commits/56645eaebcd31dde95d7029c5076c216287ed3b4

StrataMain.lean
shigoel: Done in https://github.com/strata-org/Strata/pull/767/commits/56645eaebcd31dde95d7029c5076c216287ed3b4

StrataTest/Languages/Boole/stack_array_based.lean
shigoel, line 1: Slipped in  a `#guard_msg` here to clean up `lake test` output; not directly related to this PR.

StrataMain.lean
```diff
@@ -1083,6 +1100,102 @@ structure CommandGroup where
...
+      | .error msgs =>
+        for e in msgs do println! s!"Error: {← e.toString}"
+        exitFailure s!"{msgs.size} parse error(s)"
+    match Strata.genericToCore pgm with
+    | .error msg =>
+      exitFailure msg
+    | .ok initProgram =>
+      -- Apply passes left to right
+      let mut program := initProgram
+      for pc in passConfigs do
```
aqjune-aws, line 1168: I wonder whether we can chain the status of `CoreTransformState` by e.g., simply keeping this `do` chain on `CoreTransformM` so that
(1) the fresh variable generator doesn't start over: if we give `loopElim` followed by `callElim`, the fresh variable counter after the first `loopElim` is kept at the second `callElim`
(2) also, the `CachedAnalysis` built at the first pass could be reused at the second pass, because building CallGraph is expensive.

StrataMain.lean
```diff
@@ -1083,6 +1100,102 @@ structure CommandGroup where
   commands : List Command
   commonFlags : List Flag := []
 
+private def validPasses :=
+  "inlineProcedures, loopElim, callElim, filterProcedures, removeIrrelevantAxioms"
```
aqjune-aws, line 1104: Let's have a list of pass names and their mapping functions (something like `Program -> PassConfig -> Program`?)

StrataTest/Languages/Core/Tests/NestedVarScopingTest.lean
shigoel: It was a CST->AST parsing bug! Strata/Languages/Core/DDMTransform/Translate.lean

shigoel: Discovered during round-tripping tests I did for this PR.

aqjune-aws: This is really cool.

StrataMain.lean
shigoel: Thanks, nice catch! https://github.com/strata-org/Strata/pull/767/commits/c744e78ac3bc45326699a578b906d1e44c5f0bc2 should address this.

StrataMain.lean
shigoel: Perhaps `Core.TransformPass` added in SimpleAPI.lean addresses just enough of this concern.

StrataTest/Languages/Core/Tests/TestASTtoCST.lean
```diff
@@ -168,7 +172,9 @@ function boolId(x : bool): bool;
 #end
 
 /--
-info: datatype IntList {(
+info: program Core;
+
+datatype IntList {(
```
MikaelMayer, line 177: Can't wait to see these parentheses go away.

aqjune-aws (approved):

StrataTest/Languages/Core/Tests/TestASTtoCST.lean
shigoel: Me too, me too. Next PR.

PR #770, opened 2026-04-07
Use DOCGEN_SRC=file instead of SSH URL rewriting for doc generation

tautschnig:
Replace the workaround from PR #732 that temporarily mutated the git remote URL (SSH→HTTPS) with the doc-gen4-recommended DOCGEN_SRC variable. When the origin remote uses an SSH URL (git@github.com:...), the script sets DOCGEN_SRC="file" so doc-gen4 uses local file references instead of failing to parse the remote. HTTPS clones and explicit DOCGEN_SRC settings are unaffected.

See https://github.com/leanprover/doc-gen4/issues/376

### DIFF
--- a/docs/verso/generate.sh
+++ b/docs/verso/generate.sh
@@ -1,12 +1,12 @@
 set -ex
 
-# Normalize SSH git remote to HTTPS so doc-gen4 can parse it (see #427).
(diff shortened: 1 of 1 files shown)


atomb (approved):
This looks like an improvement to me.

keyboardDrummer (approved):

PR #774, opened 2026-04-07
Dispatch-based method resolution for self.field.method() in Python classes

tautschnig:
Adds support for resolving method calls on class fields that were initialized via dispatch calls. When `__init__` assigns `self.field = dispatch_call(service_name)`, the resolved composite type is now tracked in a `dispatchFieldTypes` map, enabling `self.field.method()` to resolve against the correct service type.

Changes:

1. **Detect dispatch-initialized fields** (`a988ee9e3`):
   - Add `dispatchFieldTypes` to `TranslationContext` (separate from `classFieldHighType` to avoid triggering composite field resolution for PySpec types with deeply nested TypedDict preconditions).
   - Scan `__init__` bodies for `self.field = dispatch_call(...)` patterns and record the resolved composite type.
   - In `refineFunctionCallExpr`, when `self.field.method()` is called and the field type is `Any`, look up `dispatchFieldTypes` to resolve the method against the correct service type.

2. **Enable dispatch detection in class translation context** (`fe909a978`):
   - Pass `overloadTable` into `initCtx` so `resolveDispatch` can find service type mappings during class translation.
   - Skip factory dispatch (Step 1) inside method bodies to prevent `self.field = dispatch_call()` from creating `New` expressions that pollute `classFieldHighType`.
   - Return `Hole` for composite-typed `self.field` access instead of throwing "Coercion to Any not supported" — method dispatch is handled by `refineFunctionCallExpr` via `dispatchFieldTypes`.


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1013,7 +1052,14 @@ partial def translateAssign  (ctx : TranslationContext)
...
+  -- to avoid composite coercion issues; dispatch-initialized fields are tracked
+  -- separately via dispatchFieldTypes for method resolution.
+  let isSelfFieldAssign := match lhs with
+    | .Attribute _ (.Name _ name _) _ _ => name.val == "self" && ctx.currentClassName.isSome
+    | _ => false
+  let rhsCtx := if isSelfFieldAssign then {ctx with suppressDispatch := true} else ctx
+  let rhs_trans ←  translateExpr rhsCtx rhs
   if let .Hole := rhs_trans.val then
   {
     match lhs with
```
atomb, line 1208: `suppressDispatch` propagates into the entire RHS expression tree, not just the top-level call. If the RHS is a nested expression like `factory.create(other_factory.client('s3'))`, the inner `other_factory.client('s3')` call would also have dispatch suppressed.

The suppression is meant to prevent the top-level call from being turned into `New className`, but it bleeds into nested calls. Would it be cleaner to suppress only at the `translateCall` level (e.g., a parameter on `translateCall`) rather than threading it through the entire context?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -615,10 +624,11 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
-            s!"Coercion from user-defined class '{name.text}' to Any is not yet supported"
-            name.text)
+        | some (.UserDefined _) =>
+          -- Composite-typed field (e.g. service client): return Hole.
+          -- Method dispatch on self.field.method() is handled by
+          -- refineFunctionCallExpr via dispatchFieldTypes.
+          return mkStmtExprMd .Hole
         | _ =>
           return fieldExpr
       else
```
atomb, line 708: Returning `Hole` for all `UserDefined` field accesses is a broader change than just dispatch support. Previously this threw an error, which was conservative but safe.

Now any `self.field` access where the field has a `UserDefined` type silently becomes `Hole`, even if the access isn't followed by a method call (e.g., `x = self.field` would assign `Hole` to `x`). This reduces bug-finding ability for standalone composite field reads. Worth considering whether the `Hole` return should be limited to cases where dispatch is actually available for that field.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1826,7 +1882,29 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
...
+                  matchDispatchField ctx target value
+                else pure none
+              | .AnnAssign _ target _ value _ =>
+                if let some value := value.val then
+                  matchDispatchField ctx target value
+                else pure none
+              | _ => pure none
+            if let some (fieldName, cn) := dispatchField? then
+              dispatchFields := dispatchFields.insert fieldName cn
+    let ctx := {ctx with
```
atomb, line 2255: The dispatch detection only inspects top-level statements inside `__init__`. If the dispatch call is inside a conditional or try/except:

```python
def __init__(self):
    try:
        self.client = boto3.client('s3')
    except:
        self.client = None
```

it won't be detected. Probably fine for now, but might be worth a brief comment noting this as a known limitation.

StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -206,4 +206,55 @@ def create_service() -> Any:
...
+
+def main() -> None:
+    w: Wrapper = Wrapper()
+    w.run()
+"
+  let diags ← processPythonFile pythonCmd (stringInputContext "test.py" program)
+  -- Translation should succeed; diagnostics (if any) should not include coercion errors
+  for d in diags do
+    if d.message.contains "Coercion to Any not supported" then
+      throw (IO.userError s!"Unexpected coercion error: {d.message}")
```
atomb, line 287: This test checks that no coercion errors appear, but doesn't assert a specific diagnostic count. If the translation produces unexpected non-coercion diagnostics, the test would still pass silently. Consider adding a bound on `diags.size` or logging the diagnostics for visibility.

atomb (commented):
🤖 says:

Nice work on the dispatch detection mechanism. The separation between detection (`matchDispatchField` in `translateClass`) and consumption (`refineFunctionCallExpr`) is clean, and the `compositeTypeReverse` precomputed reverse map is a good call for O(1) lookups. The `overloadTable` addition to `initCtx` is a necessary companion fix.

I left a few inline comments — the most significant ones are about `suppressDispatch` scope and the broader semantic change from error to `Hole` for `UserDefined` field accesses.

tautschnig:

StrataTest/Languages/Python/VerifyPythonTest.lean
tautschnig: Addressed in [c2086a8](https://github.com/strata-org/Strata/pull/774/commits/c2086a8aa9e8045e7dd91b279181caece68592cd).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [90d211f](https://github.com/strata-org/Strata/pull/774/commits/90d211fdce57264b1adaf05f065dcace6564e36c).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [90d211f](https://github.com/strata-org/Strata/pull/774/commits/90d211fdce57264b1adaf05f065dcace6564e36c).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [90d211f](https://github.com/strata-org/Strata/pull/774/commits/90d211fdce57264b1adaf05f065dcace6564e36c).

tautschnig:

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -2189,6 +2302,7 @@ def PreludeInfo.merge (a b : PreludeInfo) : PreludeInfo where
 /-- Translate Python module to Laurel Program using pre-extracted prelude info. -/
 def pythonToLaurel' (info : PreludeInfo)
     (body : Array (stmt SourceRange))
+    (prev_ctx: Option TranslationContext := none)
```
MikaelMayer: nit: `prev_ctx` is never passed as `some` by any caller (`PySpecPipeline` passes `none` explicitly, tests use the default). If this is scaffolding for a future multi-file translation feature, a brief doc comment on the parameter explaining the intended use case would help. Otherwise, consider removing it to avoid the dead `some` branch at line 2400 which silently drops locally computed `compositeTypeReverse`, `importedSymbols`, `classFieldHighType`, etc.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -2275,7 +2390,15 @@ def pythonToLaurel' (info : PreludeInfo)
...
 
-  let mut ctx : TranslationContext :=
+  -- Precompute reverse map: laurelName → pythonName for composite types
+  let compositeTypeReverse := importedSymbols.fold (init := ({}:Std.HashMap String String)) fun m k v =>
+    match v with
+    | .compositeType laurelName => if laurelName != k then m.insert laurelName k else m
+    | _ => m
+
+  let mut ctx : TranslationContext := match prev_ctx with
+  | some prev_ctx => {prev_ctx with functionSignatures:= prev_ctx.functionSignatures ++ allClassFuncDecls}
```
MikaelMayer: When `prev_ctx` is `some`, only `functionSignatures` is updated. The locally computed `compositeTypeReverse`, `importedSymbols`, `overloadTable`, `classFieldHighType`, `exhaustiveClasses`, and `classesInHierarchy` are all discarded. If a caller ever passes `some`, these fields in `prev_ctx` would be stale (missing classes/composites from the current file). This branch should either merge the locally computed values into `prev_ctx`, or be removed if `prev_ctx` is not yet needed.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -794,7 +810,32 @@ partial def refineFunctionCallExpr (ctx : TranslationContext) (func: Python.expr
...
+              else none
+            | _ => none
+          match dispatchTy with
+          | some ty =>
+            let resolvedTy :=
+              match ctx.importedSymbols[ty]? with
+              | some (ImportedSymbol.compositeType laurelName) => laurelName
+              | _ => ty
+            let unprefixedTy := ctx.compositeTypeReverse[ty]?
+            let candidates := [
```
MikaelMayer, line 879: nit: The candidate resolution logic (try `ty`, then `resolvedTy`, then `unprefixedTy`) is doing three different name-mangling strategies. A brief inline comment explaining when each candidate would match (e.g., "raw dispatch name", "laurel-resolved name", "python-visible name") would make this easier to follow.

tautschnig:

ssomayyajula (commented):
Questions:

1. Is this PR trying to get around the fact that class attributes (which are assigned to method calls that dispatch) are getting weak annotations, like `Any`? That is, if a class attribute is annotated with a more specific type, does ordinary method resolution work as it is for `self.field.method()`?
2. My sense is that `scanDispatchFields` will record the last-encountered assignment to the same attribute. So if the input program conditionally assigns to an object's attribute, then its type would resolve to the `else` branch as opposed to the union of both branches (or just calling it `Any`). Is that the desired behavior?

tautschnig:
To @ssomayyajula 's questions:

>     1. Is this PR trying to get around the fact that class attributes (which are assigned to method calls that dispatch) are getting weak annotations, like `Any`? That is, if a class attribute is annotated with a more specific type, does ordinary method resolution work as it is for `self.field.method()`?
> 

The dispatch lookup in `refineFunctionCallExpr` only fires when `callerTy == PyLauType.Any` (line 852). If a field has a concrete type annotation (e.g., `client: Storage`), `inferExprType` returns that type, and normal method resolution is used. Dispatch is purely a fallback for untyped/`Any` fields.

>     2. My sense is that `scanDispatchFields` will record the last-encountered assignment to the same attribute. So if the input program conditionally assigns to an object's attribute, then its type would resolve to the `else` branch as opposed to the union of both branches (or just calling it `Any`). Is that the desired behavior?

Fixed in [13addaf](https://github.com/strata-org/Strata/pull/774/commits/13addaf17809036d00ea16ddc3c34f681d57d4fd).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [329c4e6](https://github.com/strata-org/Strata/pull/774/commits/329c4e60136724ae9526b18878f4cbc637a0b73d).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Documented in [329c4e6](https://github.com/strata-org/Strata/pull/774/commits/329c4e60136724ae9526b18878f4cbc637a0b73d)[329c4e6](https://github.com/strata-org/Strata/pull/774/commits/329c4e60136724ae9526b18878f4cbc637a0b73d).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [329c4e6](https://github.com/strata-org/Strata/pull/774/commits/329c4e60136724ae9526b18878f4cbc637a0b73d)[329c4e6](https://github.com/strata-org/Strata/pull/774/commits/329c4e60136724ae9526b18878f4cbc637a0b73d).

tautschnig:

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -2435,9 +2563,10 @@ def pythonToLaurel' (info : PreludeInfo)
         (init := info.importedSymbols) fun m name =>
           m.insert name (ImportedSymbol.compositeType name)
       let initCtx : TranslationContext := {
```
MikaelMayer, line 2498: `initCtx` is missing `compositeTypeReverse`, which defaults to `{}`. This means during class method translation, `refineFunctionCallExpr`'s third name-mangling strategy (`unprefixedTy` at line 878) will never produce a candidate. If a dispatch-resolved method can only be found via the Python-visible short name (e.g., `S3Client@method` from `from mod import S3Client`), resolution will silently fall back to the default `manglePythonMethod resolvedTy callname`, which may not match.

Since `compositeTypeReverse` is precomputed from `importedSymbols`, and `initCtx` uses `localSymbols` (which may differ from the final `importedSymbols`), you'd need to compute it from `localSymbols` here — or compute it once and update incrementally as `compositeTypeNames` grows.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -2116,6 +2174,52 @@ def extractFieldsFromInit (ctx : TranslationContext) (initBody : Array (Python.s
...
+        let b ← scanDispatchFields ctx orelse.val.toList
+        pure (a ++ b)
+      | .Try _ body handlers orelse finalbody =>
+        let a ← scanDispatchFields ctx body.val.toList
+        let b ← handlers.val.toList.flatMapM fun h =>
+          match h with | .ExceptHandler _ _ _ hbody => scanDispatchFields ctx hbody.val.toList
+        let c ← scanDispatchFields ctx orelse.val.toList
+        let d ← scanDispatchFields ctx finalbody.val.toList
+        pure (a ++ b ++ c ++ d)
+      | .For _ _ _ body _ _ | .While _ _ body _ =>
```
MikaelMayer: Possible soundness issue: `For` and `While` both have an `orelse` clause (runs when the loop completes without `break`). The pattern `| .For _ _ _ body _ _ | .While _ _ body _` ignores the `orelse` field, so dispatch calls in `for/else` or `while/else` blocks would be missed. This means we might state incorrectly that some calls are unreachable when they would be if the "else" block was handled.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -2191,7 +2295,31 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
...
+          let dispatchResult ← scanDispatchFields ctx initBody.val.toList
+          for (fieldName, cn) in dispatchResult do
+            match dispatchFields[fieldName]? with
+            | some existing =>
+              if existing != cn then
+                -- Same field assigned different dispatch types in different
+                -- branches — mark as conflicting and remove from dispatch map.
+                conflictingFields := conflictingFields.insert fieldName
+            | none => pure ()
+            dispatchFields := dispatchFields.insert fieldName cn
```
MikaelMayer, line 2251: nit: `dispatchFields := dispatchFields.insert fieldName cn` runs unconditionally after the match, including for fields already marked as conflicting. The cleanup loop removes them afterward so it's not a bug, but the intent would be clearer if the insert only happened in the `none` branch (or if the `some existing` branch used `continue` when `existing != cn`).

MikaelMayer (changes requested):

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Addressed in [8b4d7e0](https://github.com/strata-org/Strata/pull/774/commits/8b4d7e0b40459af7302bb65971874825e488b362).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Fixed in [8b4d7e0](https://github.com/strata-org/Strata/pull/774/commits/8b4d7e0b40459af7302bb65971874825e488b362).

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Done in [8b4d7e0](https://github.com/strata-org/Strata/pull/774/commits/8b4d7e0b40459af7302bb65971874825e488b362).

StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -236,6 +236,89 @@ def create_service() -> Any:
...
+
+class Wrapper:
+    svc: Svc
+    def __init__(self) -> None:
+        try:
+            self.svc = Svc()
+        except:
+            pass
+    def run(self) -> None:
+        self.svc.do_thing(val=\"hello\")
```
MikaelMayer, line 312: To prove that the method is now resolved, what about having an end-to-end test for bug finding that will inline everything and able to detect a bug that only a true method resolution would find?

tautschnig: Test added in [28d1536](https://github.com/strata-org/Strata/pull/774/commits/28d15365e762547c79a8a1a8c1b419b9974094f9).

ssomayyajula (approved):

MikaelMayer (approved):

PR #776, opened 2026-04-07
Always put a message in final-message

### LINKED ISSUE #775: Always put a message in final-message
When we emit an SMT file, we sometimes emit a

`set-info :final-message`

when there is a message associated to the assertion we are proving.
We should actually always put a message here, even if there is no user-defined message. We can use the assertion's label.

MikaelMayer:
When emitting SMT files, the `set-info :final-message` was only emitted when the obligation metadata contained a user-defined message. This meant some SMT files had no final-message at all.

Now `final-message` is always emitted. When a user-defined message exists, it is used (unchanged behavior). Otherwise, the assertion's label is used as the fallback message. The escaping logic is applied once to whichever raw message is selected, avoiding duplication.

The `label` parameter in `encodeCore` and `dischargeObligation` is now required (non-optional) to prevent callers from accidentally omitting it.

Existing tests pass. Two new tests cover both paths: one where metadata has no message (label used as fallback), and one where metadata has a message (message used directly). Both print the full SMT output unquoted for visibility.

Fixes #775


Strata/Languages/Core/Verifier.lean
```diff
@@ -108,7 +107,9 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
   | some elem =>
     let msg := toString (Std.format elem.value) |>.replace "\\" "\\\\" |>.replace "\"" "\\\""
     Solver.setInfo "final-message" s!"\"{msg}\""
-  | none => pure ()
+  | none =>
+    let escapedLabel := label.replace "\\" "\\\\" |>.replace "\"" "\\\""
+    Solver.setInfo "final-message" s!"\"{escapedLabel}\""
```
MikaelMayer: I'd like to see a test to for that.

Strata/Languages/Core/Verifier.lean
```diff
@@ -165,6 +166,7 @@ def dischargeObligation
   (obligationTerm : Term)
   (ctx : SMT.Context)
   (satisfiabilityCheck validityCheck : Bool)
+  (label : String := "")
```
MikaelMayer: I don't like to see an optional parameter here, it could encourage anyone to forget about the label.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Added a test in `SMTEncoderTests.lean` that calls `encodeCore` with empty metadata and a label, then verifies the SMT output contains `(set-info :final-message "my-obligation")`.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Made `label` a required (non-optional) parameter in both `encodeCore` and `dischargeObligation`. Updated all call sites to pass the label explicitly.

StrataTest/Languages/Core/Tests/SMTEncoderTests.lean
```diff
@@ -310,11 +310,33 @@ end ArrayTheory
...
+    Strata.SMT.SolverM.run solver
+      (Strata.SMT.Encoder.encodeCore ctx (pure ()) [] obligationTerm md
+        (satisfiabilityCheck := false) (validityCheck := true) (label := "my-obligation"))
+  let contents ← b.get
+  let smt :=
+    if h : contents.data.IsValidUTF8
+    then String.fromUTF8 contents.data h
+    else ""
+  let parts : List String := smt.splitOn "(set-info :final-message \"my-obligation\")"
+  return (parts.length > 1 : Bool)
```
MikaelMayer: I would prefer a test that prints the SMT goal so that we can see this set-info in context and the guard_msgs displays it entirely. Otherwise it's buried.

MikaelMayer: Updated the test to print the full SMT output instead of checking a boolean. The `guard_msgs` now displays the entire SMT goal, making the `set-info :final-message` visible in context:

```
(set-logic ALL)
; Validity
(assert false)
(check-sat)
(set-info :final-message "my-obligation")
```

StrataTest/Languages/Core/Tests/SMTEncoderTests.lean
```diff
@@ -310,11 +310,32 @@ end ArrayTheory
...
       (Strata.SMT.Encoder.encodeCore ctx (pure ()) [] obligationTerm md
-        (satisfiabilityCheck := false) (validityCheck := true))
+        (satisfiabilityCheck := false) (validityCheck := true) (label := "test"))
   -- ids should contain "c" but not "f"
   let hasF := ids.any (· == "f")
   return (ids, !hasF)
 
+/-! ## Test that final-message falls back to label when metadata has no message -/
+
+/-- info: "(set-logic ALL)\n; Validity\n(assert false)\n(check-sat)\n(set-info :final-message \"my-obligation\")\n" -/
```
MikaelMayer: Excellent. Any way to ensure we actually print the string instead of quoting it?
Also, we are not testing what this PR claims it fixes. This PR claims that if there is no metadata with a message, we still emit a set-info :final-message. However, this test is always providing a label message. I want a test in which the metadata does not contain a message, and that test should print the assertion's name instead

MikaelMayer: Two changes:

1. Switched from `return smt` to `IO.print smt` so the output is printed as actual text (no quoting/escaping).

2. Renamed the label to `"assert_bounds_check"` to make it clear it's an assertion name, not a user message.

3. Added a second contrasting test where metadata **does** contain a message (`"Division by zero is impossible"`). This shows both paths:
   - No message in metadata → label (`assert_bounds_check`) is used as final-message
   - Message in metadata → that message (`Division by zero is impossible`) is used instead

Both tests build and pass.

Strata/Languages/Core/Verifier.lean
```diff
@@ -106,7 +107,9 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
   | some elem =>
     let msg := toString (Std.format elem.value) |>.replace "\\" "\\\\" |>.replace "\"" "\\\""
     Solver.setInfo "final-message" s!"\"{msg}\""
-  | none => pure ()
+  | none =>
+    let escapedLabel := label.replace "\\" "\\\\" |>.replace "\"" "\\\""
+    Solver.setInfo "final-message" s!"\"{escapedLabel}\""
```
MikaelMayer: The escaping logic (`.replace "\\" "\\\\" |>.replace "\"" "\\\""`) and the `Solver.setInfo` call are duplicated across both branches. Consider extracting the raw message from the match and doing the escaping + emit once:

```lean
let rawMsg := match md.findElem Imperative.MetaData.message with
  | some elem => toString (Std.format elem.value)
  | none => label
let escaped := rawMsg.replace "\\" "\\\\" |>.replace "\"" "\\\""
Solver.setInfo "final-message" s!"\"{escaped}\""
```

This removes the duplication and makes it easier to update the escaping logic in one place if it ever needs to change.

MikaelMayer (commented):
Clean PR — the change is well-scoped and the tests cover both paths. One suggestion to reduce duplication in the escaping logic.

Strata/Languages/Core/Verifier.lean
MikaelMayer: Extracted the raw message from the match and deduplicated the escaping + `Solver.setInfo` call, exactly as suggested.

Strata/Languages/Core/Verifier.lean
```diff
@@ -104,11 +103,11 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
...
 
   -- Emit the "message" metadata field at the very end (once per obligation).
-  match md.findElem Imperative.MetaData.message with
-  | some elem =>
-    let msg := toString (Std.format elem.value) |>.replace "\\" "\\\\" |>.replace "\"" "\\\""
-    Solver.setInfo "final-message" s!"\"{msg}\""
-  | none => pure ()
+  let rawMsg := match md.findElem Imperative.MetaData.message with
+    | some elem => toString (Std.format elem.value)
+    | none => label
```
shigoel, line 110: May make sense to escape potentially problematic characters in labels too.

ssomayyajula: +1

ssomayyajula (approved):
Preemptively approving on resolution of Shilpi's comment

Strata/Languages/Core/Verifier.lean
MikaelMayer: The label is already escaped — when the `none` branch fires, `rawMsg` is set to `label`, and the very next line applies the escaping to `rawMsg` regardless of which branch produced it:

```lean
let rawMsg := match md.findElem Imperative.MetaData.message with
    | some elem => toString (Std.format elem.value)
    | none => label
let escaped := rawMsg.replace "\\" "\\\\" |>.replace "\"" "\\\""
Solver.setInfo "final-message" s!"\"{escaped}\""
```

So both the metadata message and the label fallback go through the same escaping. No change needed.

shigoel: Thanks, I missed that

PR #780, opened 2026-04-07
Partial support of 'is' and 'is not' operators in Python

olivier-aws:
Add support for Python's 'is' and 'is not' comparison operators in the PythonToLaurel and PythonToCore translators, restricted to None comparisons only. This is by far the biggest usage of 'is' and 'is not' in Python.

Map Is to PEq and IsNot to PNEq when the RHS comparator is a Constant with ConNone. In the Any value model where None is the singleton from_None(), identity and equality coincide. All other uses of is/is not now report 'unsupported construct' because Python's identity check is unsound for non-None operands (e.g., True is not 1 but True == 1).

Changes:
- Add Is/IsNot handling guarded by None check in PythonToLaurel.lean
- Add positive and negative test cases for is None / is not None
- Add tests for unsupported non-None is/is not comparisons
- Document known limitation of Is→PEq mapping soundness

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Examples/DoubleTwice.core.st
+++ b/Examples/DoubleTwice.core.st
@@ -1,0 +1,18 @@
+program Core;
+
+procedure Double(n : int) returns (result : int)
+spec {
+  ensures [double_correct]: (result == n * 2);
+}
+{
+  result := n + n;
+};
+
+procedure TestProc(x : int) returns (output : int)
+spec {
+  ensures [testProc_result]: (output == x * 4);
+}
+{
+  call output := Double(x);
+  call output := Double(output);
+};
--- a/Examples/IrrelevantAxioms.core.st
+++ b/Examples/IrrelevantAxioms.core.st
@@ -1,0 +1,24 @@
+program Core;
+
+// Two independent uninterpreted functions, each with their own axioms.
+// removeIrrelevantAxioms --functions f should prune g's axioms (and vice versa).
+
+function f(x: int): int;
+axiom [f_positive]: forall x: int :: f(x) > 0;
+axiom [f_monotone]: forall x: int, y: int :: x < y ==> f(x) < f(y);
+
+function g(x: int): int;
+axiom [g_negative]: forall x: int :: g(x) < 0;
+axiom [g_idempotent]: forall x: int :: g(g(x)) == g(x);
+
+// h calls f, so h's axiom is also relevant when focusing on f.
+function h(x: int): int;
+axiom [h_def]: forall x: int :: h(x) == f(x) + 1;
+
+procedure TestF(x: int) returns (result: int)
+spec {
+  ensures [result_positive]: result > 0;
+}
+{
+  result := f(x);
+};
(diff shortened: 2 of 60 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -510,7 +510,16 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
       | .GtE _ => .ok "PGe"
       | .In _ => .ok "PIn"
       | .NotIn _ => .ok "PNotIn"
-      | _ => throw (.unsupportedConstruct s!"Comparison operator not yet supported: {repr ops.val[0]!}" (toString (repr e)))
+      -- `is`/`is not` are only sound when the RHS is None, because Python's
+      -- `is` checks object identity, not equality (e.g., True == 1 but
+      -- True is not 1). In the Any value model, None is a singleton so
+      -- identity and equality coincide for None comparisons.
+      | .Is _ => match comparators.val[0]! with
```
ssomayyajula: Two questions:

1. In case the LHS and RHS are of composite type, is it possible to emit reference (integer) (in)equality checks?
2. In case the LHS is an Any value and RHS is _not_ literally `None` nor of composite type, can it still make sense to use `P(N)Eq`?

shigoel: Can we avoid using `!` (in `comparators.val[0]!`)? That can cause panics.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -510,7 +510,16 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
...
+          | .Constant _ (.ConNone _) _ => .ok "PEq"
+          | _ => throw (.unsupportedConstruct "`is` is only supported with None" (toString (repr e)))
+      | .IsNot _ => match comparators.val[0]! with
```
shigoel: Same comment about `!`here.

Strata/Languages/Python/PythonToLaurel.lean
olivier-aws: Regarding @ssomayyajula questions, I believe it would be sound. However, I ran into multiple type checking errors when trying to do so, so I would prefer to keep this initial support simple and extend it later with more complete cases.

olivier-aws: > Can we avoid using ! (in comparators.val[0]!)? That can cause panics.

The accesses are guarded by 
```
if ops.val.size != 1 || comparators.val.size != 1 then
      throw (.unsupportedConstruct "Chained comparisons not yet supported" (toString (repr e)))
``` 
a few lines above, so it cannot panic IIUC

shigoel: Yes, that makes sense, but we typically prefer to provide a proof that this is safe in Lean. E.g.,
```
 if h : ops.val.size != 1 || comparators.val.size != 1 then                                                                                                                                                          
    throw (...)                                               
else 
   have hc : 0 < comparators.val.size := by omega  
   let comparator := comparators.val[0]'hc
....
```
But happy to move on if we revisit this pattern in Python->Laurel.

ssomayyajula: @olivier-aws I'm fine to approve pending Shilpi's requested changes then

shigoel: And I'm pre-approving, letting @olivier-aws decide whether to fix the `!` issue in this PR or later. (Happy to re-approve if it's in this PR)

PR #781, opened 2026-04-07
feat: Laurel.Program to Laurel concrete syntax (#779)

### LINKED ISSUE #779: Laurel.Program to Laurel concrete syntax is missing
Currently if we use `--keep-all-files` we dump a formatted representation of a Laurel.Program, but this representation is not in valid Laurel syntax, so we can't just modify something and re-run the pipeline.
Similar to the Core AST to DDM that Strata.Core has, we should have a Laurel.Program to Laurel's DDM concrete syntax so that `--keep-all-files` and emit a correct Laurel program.
This might also help in establishing tests for some Laurel to Laurel transformations, to investigate.
But at minimum a Laurel AST to Laurel DDM concrete syntax tree.

We might be able to test it using existing laurel tests.

MikaelMayer:
Fixes #779

Adds a Laurel AST to DDM concrete syntax tree translator (`programToStrata`), enabling serialization of Laurel programs through the generic Strata AST. Removes `LaurelFormat.lean` in favor of using the DDM grammar-based formatter for all formatting.

**What it does:**
- `programToStrata`: converts `Laurel.Program` to `Strata.Program` (DDM operations matching `LaurelGrammar.st`)
- All formatting (`Program`, `Procedure`, `StmtExpr`, `HighType`, `CompositeType`, etc.) goes through the abstract→concrete converter + DDM grammar formatter — single source of truth
- `instance : Std.ToFormat Program` and all sub-type `ToFormat` instances use the DDM-based path
- Implements `laurelToGeneric` in `SimpleAPI` (was `noncomputable opaque`)
- `--keep-all-files` emits `.laurel` and `.lowered.laurel` files using the DDM-based formatter
- Fixes `LaurelGrammar.st` formatting strings to produce readable, parseable output
- Single output parameters with non-`result` names are preserved through the DDM roundtrip

**Limitations:**
- `staticFields` and `constants` are not emitted because the Laurel grammar has no top-level commands for them
- `formatConstant` still does its own formatting since there's no grammar rule for constants, but delegates to DDM-based sub-type formatters

**How it was tested:**
- Roundtrip tests: parse Laurel source → `programToStrata` → `genericToLaurel` → DDM format → verify output matches original
- Every test re-parses its formatted output and verifies convergence
- All existing tests continue to pass


Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,491 @@
...
+    | .Abstract => laurelOp "identifier" #[ident "abstract"]
+    | .All => laurelOp "identifier" #[ident "all"]
+    | .PureFieldUpdate target field value =>
+      -- Not directly in grammar; emit as assignment to field
+      laurelOp "assign" #[
+        laurelOp "fieldAccess" #[stmtExprToArg target, ident field.text],
+        stmtExprToArg value
+      ]
+
+private def stmtExprToArgWithSummary (s : StmtExprMd) : Arg :=
```
MikaelMayer: `stmtExprToArgWithSummary` is defined but never used anywhere. Should be removed.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,491 @@
...
+partial def fmtHighType (t : HighTypeMd) : Format :=
+  match t.val with
+  | .TVoid => "void"
+  | .TBool => "bool"
+  | .TInt => "int"
+  | .TFloat64 => "float64"
+  | .TReal => "real"
+  | .TString => "string"
+  | .THeap => "Heap"
+  | .TTypedField vt => "Field[" ++ fmtHighType vt ++ "]"
```
MikaelMayer: The docstring on `fmtHighType` says "Format a HighType as valid Laurel DDM syntax", but `TTypedField` and `TSet` produce `Field[...]` and `Set[...]` which are not valid Laurel grammar syntax (the grammar only has `compositeType(name: Ident)`). This means `formatLaurelDDM` output won't round-trip for programs containing these types.

`highTypeValToArg` handles this correctly by emitting just the name. `fmtHighType` should do the same for consistency, or the docstring should note the limitation.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,491 @@
...
+  | .TBool => laurelOp "boolType"
+  | .TFloat64 => laurelOp "float64Type"
+  | .TReal => laurelOp "realType"
+  | .TString => laurelOp "stringType"
+  | .TMap k v => laurelOp "mapType" #[highTypeToArg k, highTypeToArg v]
+  | .UserDefined name => laurelOp "compositeType" #[ident name.text]
+  | .TCore s => laurelOp "coreType" #[ident s]
+  | .TVoid => laurelOp "compositeType" #[ident "void"]
+  | .THeap => laurelOp "compositeType" #[ident "Heap"]
+  | .TTypedField _vt => laurelOp "compositeType" #[ident "Field"]
```
MikaelMayer, line 52: `TTypedField` and `TSet` discard their type parameters (the grammar can't represent them). A brief comment explaining this — similar to the one on `Applied` at line 51 — would help readers understand this is intentional.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,491 @@
...
+      | none => laurelOp "block" #[semicolonSep stmtArgs]
+      | some l => laurelOp "labelledBlock" #[semicolonSep stmtArgs, ident l]
+    | .LocalVariable name ty init =>
+      let typeOpt := optionArg (some (laurelOp "typeAnnotation" #[highTypeToArg ty]))
+      let initOpt := optionArg (init.map fun e => laurelOp "initializer" #[stmtExprToArg e])
+      laurelOp "varDecl" #[ident name.text, typeOpt, initOpt]
+    | .Assign [target] value =>
+      laurelOp "assign" #[stmtExprToArg target, stmtExprToArg value]
+    | .Assign targets value =>
+      -- Multi-target assign: emit as assign of first target (best effort)
```
MikaelMayer: The second `Assign` branch has a comment saying "Multi-target assign: emit as assign of first target (best effort)" but the `[]` sub-case silently drops the assignment target entirely and just emits the value. This seems like it could produce surprising output for an empty target list.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,491 @@
...
+    The resulting program can be formatted using `Strata.Program.format` to
+    produce Laurel source text. -/
+def programToStrata (prog : Laurel.Program) : Strata.Program :=
+  let typeOps := prog.types.map typeDefinitionToOp |>.toArray
+  let procOps := prog.staticProcedures.map procedureCommandOp |>.toArray
+  Strata.Program.create Laurel_map "Laurel" (typeOps ++ procOps)
+
+open Std (Format format)
+open Std.Format
+
```
MikaelMayer, line 323: No action needed: the dual-path design (DDM tree via `programToStrata` + direct text via `formatLaurelDDM`) is well-motivated — the DDM path enables serialization/tooling integration while the text path gives readable output. The structural overlap with `LaurelFormat.lean` is acceptable since that formatter is for debug display and doesn't produce parseable syntax.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Removed `stmtExprToArgWithSummary` — it was indeed unused.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Updated the docstring on `fmtHighType` to clarify that `Field[T]` and `Set[T]` are kept for readability in the text path, even though they are not valid grammar syntax.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Added a comment above the `TTypedField` and `TSet` cases explaining that type parameters are discarded because the grammar cannot represent them.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Fixed the empty-targets case to emit `_ := value` (a placeholder assignment) instead of silently dropping the assignment structure.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Acknowledged, no changes needed.

StrataTest/Languages/Laurel/AbstractToConcreteTreeTranslatorTest.lean
```diff
@@ -0,0 +1,224 @@
...
+  | .ok program => pure program
+
+private def laurelToText (prog : Program) : String :=
+  (formatLaurelDDM prog).pretty
+
+private def roundtrip (input : String) : IO String := do
+  let program ← parseLaurel input
+  pure (laurelToText program)
+
+private def roundtripParse (input : String) : IO Unit := do
```
MikaelMayer: If we parse and print the second program, we should obtain the same (it should converge) can you please verify that?

MikaelMayer: 🤖 Updated `roundtripParse` to verify convergence: it now parses the emitted text a second time, emits it again, and checks that both outputs are identical. All 7 roundtrip tests pass with this stricter check.

StrataTest/Languages/Laurel/AbstractToConcreteTreeTranslatorTest.lean
```diff
@@ -0,0 +1,227 @@
+/-
```
MikaelMayer, line 1: Not part of the test suite but can you please check that --keep-all-files pretty print Laurel correctly now? On python files.

MikaelMayer: 🤖 Updated `--keep-all-files` to use `formatLaurelDDM` for both `.laurel` and `.lowered.laurel` output files, so they now emit valid Laurel concrete syntax instead of the debug format. The Lean compilation passes. I could not run the binary end-to-end on this machine due to a GLIBC version mismatch in the C linker, but the code path is straightforward: `formatLaurelDDM` replaces `f!"{combinedLaurel}"` in both write sites.

MikaelMayer: Use lake build StrataMain and then lake env to run it instead of running the executable.

MikaelMayer: 🤖 Verified end-to-end using `lake build StrataMain` + `lake env lean --run` with a simple Python file (`def add(x: int, y: int) -> int: assert x >= 0; return x + y`).

Both `.laurel` and `.lowered.laurel` outputs produce valid Laurel concrete syntax via `formatLaurelDDM`. The output includes properly formatted datatypes, composites, functions (with `external`, body expressions), procedures (with `requires`/`ensures`/`summary` contracts, imperative bodies), and all Laurel constructs from the Python pipeline.

The `--keep-all-files` code path correctly uses `formatLaurelDDM` for both files.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,305 @@
...
+    name := { dialect := "Laurel", name := "procedureCommand" }
+    args := #[.op (procedureToOp proc)] }
+
+/-- Convert a Laurel.Program to a Strata.Program (DDM concrete syntax tree).
+    The resulting program can be formatted using `Strata.Program.format` to
+    produce Laurel source text. -/
+def programToStrata (prog : Laurel.Program) : Strata.Program :=
+  let typeOps := prog.types.map typeDefinitionToOp |>.toArray
+  let procOps := prog.staticProcedures.map procedureCommandOp |>.toArray
+  Strata.Program.create Laurel_map "Laurel" (typeOps ++ procOps)
```
atomb, line 310: 🤖 says:

`programToStrata` silently drops `prog.staticFields` and `prog.constants`:
   ```lean
   def programToStrata (prog : Laurel.Program) : Strata.Program :=
     let typeOps := prog.types.map typeDefinitionToOp |>.toArray
     let procOps := prog.staticProcedures.map procedureCommandOp |>.toArray
     Strata.Program.create Laurel_map "Laurel" (typeOps ++ procOps)
   ```
   While the `ConcreteToAbstractTreeTranslator` always sets `staticFields := []` and `constants := []` (so roundtrip tests pass), programs constructed programmatically or from other frontends (e.g., the Python pipeline) may have non-empty fields/constants. If those programs are ever serialized via `programToStrata`, the data will be silently lost. Consider either:
   - Adding a comment documenting this limitation, or
   - Emitting the fields/constants (if the grammar supports them), or
   - Raising an error/warning if they are non-empty.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,305 @@
...
+      let typeOpt := optionArg (some (laurelOp "typeAnnotation" #[highTypeToArg ty]))
+      let initOpt := optionArg (init.map fun e => laurelOp "initializer" #[stmtExprToArg e])
+      laurelOp "varDecl" #[ident name.text, typeOpt, initOpt]
+    | .Assign [target] value =>
+      laurelOp "assign" #[stmtExprToArg target, stmtExprToArg value]
+    | .Assign targets value =>
+      -- Multi-target assign: emit as assign of first target (best effort)
+      match targets with
+      | [] => laurelOp "assign" #[laurelOp "identifier" #[ident "_"], stmtExprToArg value]
+      | t :: _ => laurelOp "assign" #[stmtExprToArg t, stmtExprToArg value]
```
atomb: 🤖  says:

The `Assign` pattern matching has overlapping arms:
   ```lean
   | .Assign [target] value =>
     laurelOp "assign" #[stmtExprToArg target, stmtExprToArg value]
   | .Assign targets value =>
     -- Multi-target assign: emit as assign of first target (best effort)
     match targets with
     | [] => laurelOp "assign" #[laurelOp "identifier" #[ident "_"], stmtExprToArg value]
     | t :: _ => laurelOp "assign" #[stmtExprToArg t, stmtExprToArg value]
   ```
   The second arm silently drops all targets after the first in multi-target assignments. The "best effort" comment acknowledges this, but it means a roundtrip through this translator can change program semantics for multi-output procedure calls. This is worth documenting more prominently (e.g., in the `programToStrata` docstring) so callers know the translation is lossy for this case.

StrataTest/Languages/Laurel/AbstractToConcreteTreeTranslatorTest.lean
atomb, line 1: 🤖  says:

`./StrataTest/Languages/Laurel/AbstractToConcreteTreeTranslatorTest.lean` — No test coverage for:
   - `ConstrainedType` roundtripping
   - `InstanceCall` (method calls on objects)
   - `Old`, `Fresh`, `Assigned` expressions
   - Procedures with `modifies` clauses
   - Nondeterministic holes (`Hole false`)
   - `LiteralDecimal`, `LiteralString`
   - `Return none` (void return)
   
   These are all handled in the translator code but untested. The most important gaps are `ConstrainedType` (it's a whole `TypeDefinition` variant) and `modifies` clauses (they affect verification semantics).

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Added a docstring note to `programToStrata` explaining that `staticFields` and `constants` are intentionally not emitted because the Laurel grammar has no top-level commands for them. The concrete-to-abstract parser also always sets `staticFields := []` and does not handle constants.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Merged the two `Assign` arms into a single pattern that handles all cases: uses the first target if available, or a `_` placeholder for empty target lists.

StrataTest/Languages/Laurel/AbstractToConcreteTreeTranslatorTest.lean
MikaelMayer: 🤖 Added roundtrip tests for `ConstrainedType`, procedures with `modifies` clauses, and nondeterministic holes (`<??>`).

`InstanceCall`, `Old`, `Fresh`, and `Assigned` are AST-only constructs not representable in the Laurel grammar — they are emitted as equivalent grammar constructs (e.g. `old(x)` becomes `call(identifier("old"), x)`) and parse back as `StaticCall`. This is by design, so text-based roundtrip tests are not applicable for them.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,306 @@
...
+private def modifiesClauseToArg (modifies : List StmtExprMd) : Arg :=
+  let refs := modifies.map stmtExprToArg |>.toArray
+  laurelOp "modifiesClause" #[commaSep refs]
+
+private def procedureToOp (proc : Procedure) : Strata.Operation :=
+  let opName := if proc.isFunctional then "function" else "procedure"
+  let params := proc.inputs.map parameterToArg |>.toArray
+  let returnTypeArg : Arg :=
+    match proc.outputs with
+    | [single] => optionArg (some (laurelOp "returnType" #[highTypeToArg single.type]))
```
MikaelMayer: When a single output parameter has a name other than `"result"` (e.g. `$result` from `eliminateHoles`), this always emits `returnType`, which only carries the type. After roundtripping through the grammar, the name becomes `"result"`, losing the original name.

The fix would mirror the logic in `formatProcedure`:
```lean
| [single] =>
  if single.name == "result"
  then optionArg (some (laurelOp "returnType" #[highTypeToArg single.type]))
  else optionArg none
```
and correspondingly in `returnParamsArg`:
```lean
| [single] =>
  if single.name == "result"
  then optionArg none
  else optionArg (some (laurelOp "returnParameters" #[commaSep #[parameterToArg single]]))
```
This would preserve non-`result` names through the DDM roundtrip.

Strata/Languages/Laurel/LaurelFormat.lean
```diff
@@ -170,24 +170,34 @@ def formatStmtExprWithMsg (s : StmtExprMd) : Format :=
...
-  | .Abstract posts => "abstract" ++ Format.join (posts.map (fun p =>
-      " ensures " ++ formatStmtExpr p ++
-      match p.md.getPropertySummary with
-      | none => Format.nil
-      | some msg => " propertySummary \"" ++ msg ++ "\""))
+      | some e => formatStmtExpr e
+  | .Abstract _ => Format.nil
   | .External => "external"
 
+def formatEnsures : Body → Format
```
MikaelMayer: The `Opaque` and `Abstract` arms have identical bodies. Consider extracting the shared logic:
```lean
private def formatPostconditions (posts : List StmtExprMd) : Format :=
  Format.join (posts.map (fun p =>
    Format.line ++ "  ensures " ++ formatStmtExpr p ++
    match p.md.getPropertySummary with
    | none => Format.nil
    | some msg => " summary \"" ++ msg ++ "\""))
```
Then `formatEnsures` becomes:
```lean
def formatEnsures : Body → Format
  | .Opaque postconds _ _ => formatPostconditions postconds
  | .Abstract posts => formatPostconditions posts
  | _ => Format.nil
```

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -0,0 +1,306 @@
...
+
+private def commaSep (args : Array Arg) : Arg := .seq sr .comma args
+
+private def semicolonSep (args : Array Arg) : Arg := .seq sr .semicolon args
+
+private def seqArg (args : Array Arg) : Arg := .seq sr .none args
+
+mutual
+
+partial def highTypeToArg (t : HighTypeMd) : Arg := highTypeValToArg t.val
```
MikaelMayer, line 38: Nit: `highTypeToArg`, `highTypeValToArg`, and `stmtExprToArg` are public (via the `public section`) but only used within this file. If Lean's `mutual`/`partial` restrictions prevent marking them `private`, a comment noting they're internal-only would help future readers.

MikaelMayer (commented):
Solid PR — the roundtrip architecture (parse → `programToStrata` → `parseProgram` → `formatProgram` → re-parse → verify convergence) is well-designed and gives strong confidence in correctness. The test coverage is thorough. A few items below.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Fixed: when a single output parameter has a name other than `"result"`, we now emit `returnParameters` (which preserves the name) instead of `returnType` (which loses it). This mirrors the logic already in `formatProcedure`.

Strata/Languages/Laurel/LaurelFormat.lean
MikaelMayer: 🤖 Extracted the shared postconditions logic into a `formatPostconditions` helper. The `Opaque` and `Abstract` arms now both delegate to it.

Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
MikaelMayer: 🤖 Added comments above the `mutual` block and `stmtExprToArg` noting they are internal-only but cannot be marked `private` due to `mutual`/`partial` restrictions.

PR #784, opened 2026-04-07
Preserve original parameter types in PySpec function signatures

tautschnig:
specArgToFuncDeclArg was hardcoding "Any" for all parameter types, discarding the original Smithy type information (str, int, bool, float). This prevented dict-unpacking type assertions from knowing the expected types, so passing None for a typed parameter was not detected.

Fix: extract the concrete type name from the SpecType atoms and use it in the PythonFunctionDecl args. This enables the dict-unpacking type assertion (implies(DictStrAny_contains, isfrom_<type>)) to fire when a dict literal contains None for a typed parameter.

The test constructs PySpec signatures with typed parameters, serializes them to Ion, reads them back via buildPySpecLaurel, and checks that the resulting PythonFunctionDecl entries have the correct tys values.


Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -49,7 +49,16 @@ private def specDefaultToExpr : Python.Specs.SpecDefault → Python.expr SourceR
...
 private def specArgToFuncDeclArg (arg : Python.Specs.Arg): Python.PyArgInfo :=
-  {name:= arg.name, md:= default, tys:= ["Any"], default:= arg.default.map specDefaultToExpr}
+  -- Extract the simple type name from the SpecType for type assertions.
+  let typeName := arg.type.atoms.findSome? fun a => match a with
+    | .ident nm _ =>
+      if nm == ⟨"builtins", "str"⟩ then some "string"
+      else if nm == ⟨"builtins", "int"⟩ then some "integer"
+      else if nm == ⟨"builtins", "bool"⟩ then some "boolean"
+      else if nm == ⟨"builtins", "float"⟩ then some "real"
+      else none
```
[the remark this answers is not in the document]
ssomayyajula: I'd something like this (a strongly typed mapping from PySpec types to to Python dialect types), but is this valid Lean code?

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -49,7 +49,18 @@ private def specDefaultToExpr : Python.Specs.SpecDefault → Python.expr SourceR
...
+  -- constraint generation (e.g., `Any..isfrom_str`).
+  let typeName := arg.type.atoms.findSome? fun a => match a with
+    | .ident nm _ =>
+      if nm == Python.PythonIdent.builtinsStr then some "str"
+      else if nm == Python.PythonIdent.builtinsInt then some "int"
+      else if nm == Python.PythonIdent.builtinsBool then some "bool"
+      else if nm == Python.PythonIdent.builtinsFloat then some "float"
+      else none
+    | _ => none
+  {name := arg.name, md := default, tys := [typeName.getD "Any"], default := arg.default.map specDefaultToExpr}
```
shigoel, line 68: `findSome?` on multi-atom `SpecType`s (union types) picks only the first recognized builtin, producing an overly restrictive type constraint. For `Optional[str]` parameters — which `ToLaurel.lean:168-176` shows are a known pattern — the atoms contain both `NoneType` and `builtins.str`. `findSome?` skips `NoneType` (not a recognized builtin) and returns `"str"`, yielding `tys = ["str"]`. Downstream, `getSingleTypeConstraint` generates `requires Any..isfrom_str(param)`, which rejects valid `None` arguments. This is a soundness regression vs. the previous `"Any"` default, which generated no constraint. The fix should either fall back to `"Any"` when `atoms.size > 1` (conservative), or map all recognized atoms into a multi-element `tys` list so `getUnionTypeConstraint` can generate a disjunction.

ssomayyajula: +1, there should also be a test involving an Optional/Union to hit this case

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -49,7 +49,18 @@ private def specDefaultToExpr : Python.Specs.SpecDefault → Python.expr SourceR
...
```
ssomayyajula: Just confirming: it's OK that this defaults to an `Any` when `nm` is a class type, right?

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -49,7 +49,18 @@ private def specDefaultToExpr : Python.Specs.SpecDefault → Python.expr SourceR
 
 /-- Convert a pyspec Arg to a PythonFunctionDecl arg tuple. -/
 private def specArgToFuncDeclArg (arg : Python.Specs.Arg): Python.PyArgInfo :=
-  {name:= arg.name, md:= default, tys:= ["Any"], default:= arg.default.map specDefaultToExpr}
+  -- Extract the simple type name from the SpecType for type assertions.
+  -- These tags must match the runtime/prelude names used by downstream
+  -- constraint generation (e.g., `Any..isfrom_str`).
+  let typeName := arg.type.atoms.findSome? fun a => match a with
```
ssomayyajula: Concerned that this and `typeTester` from #761 duplicates some logic...just flagging this as a need for PythonToLaurel and PySpecPipeline to share types so that this code can be shared.

Strata/Languages/Python/PySpecPipeline.lean
tautschnig: Thank you! Fixed in [c0db41b](https://github.com/strata-org/Strata/pull/784/commits/c0db41b324c08b1910b3569c8b4ca92bb66f5117).

Strata/Languages/Python/PySpecPipeline.lean
tautschnig: Yes — class types (e.g., `MyService`) don't have corresponding `Any..isfrom_<ClassName>` predicates in the runtime prelude, so the `filterMap` correctly returns `none` for them, and the fallback produces `tys = ["Any"]`. Since `getSingleTypeConstraint` skips `"Any"` (the `ty ≠ "Any"` guard), no constraint is generated, which is the safe behavior. Generating a constraint for an unknown type would be unsound.

Strata/Languages/Python/PySpecPipeline.lean
tautschnig: Good catch — both encode the same `PythonIdent/string → PyLauType` knowledge. Since #761 isn't merged yet and the two PRs use the mapping in slightly different directions (`PythonIdent → String` here vs `String → String` in #761), I'd suggest we land both as-is and then extract a shared canonical mapping as a follow-up. Something like a `PyLauType.ofPythonIdent? : PythonIdent → Option String` that both call sites can use.

ssomayyajula: Sounds good to me!

Issue #789, opened 2026-04-08
Correctly abort execution in LaurelToCoreTranslator for types that can not be translated to Core


keyboardDrummer:
The TODO on this line should be resolved:

https://github.com/strata-org/Strata/blob/567bffff3455b2ced513edcddad62b4d20beb594/Strata/Languages/Laurel/LaurelToCoreTranslator.lean#L68

translateType should use the `TranslateM` monad that other function in the file are using, and set `coreProgramHasSuperfluousErrors` to indicate that the compilation should be aborted.

PR #793, opened 2026-04-08
fix(python): Enable instance method call resolution and body analysis 

aqjune-aws:
*Issue #, if available:*

#762

*Description of changes:*

Emit StaticCall for user-defined class method calls instead of holes, and preserve translated method bodies instead of replacing them with opaque stubs. This enables end-to-end verification of class methods.

Changes to PythonToLaurel.lean:
- Add manglePythonMethod helper for ClassName@method name mangling
- In emitCall, emit StaticCall with self prepended for user-defined class methods (previously always emitted Hole)
- Conservatively emit Hole for classes in inheritance hierarchies (classesInHierarchy field on TranslationContext) since static resolution may be unsound with dynamic dispatch
- In translateClass, preserve method bodies for non-hierarchy classes instead of stripping to Opaque
- Add documentation comments for translateClass return type and instanceProcedures field

Without this patch, some method calls were lowered to hole expressions, making verification vacuously pass. With this patch, method bodies are analyzed and call-site preconditions are checked.

Tests:
- test_class_method_call_from_main: verifies that an assert inside a
  class method is reachable when called from __main__
- test_class_inheritance_no_dispatch: verifies that classes with
  inheritance have method calls lowered as holes, and assert False
  inside a child method is NOT reachable (conservative guard works)
- VerifyPythonTest: updated to use inheritance so the body is stripped
  to opaque, preserving the original zero-diagnostics expectation
- Updated 8 existing expected files for additional checks now generated
  from method body analysis


StrataTest/Languages/Python/AnalyzeLaurelTest.lean
```diff
@@ -229,22 +229,10 @@ private meta def runTestCase (pythonCmd : System.FilePath) (tmpDir : System.File
...
-      | (some core, []) =>
-        match Core.typeCheck Core.VerifyOptions.quiet core (moreFns := Strata.Python.ReFactory) with
-        | .error errors => return some s!"test_class_any_as_composite.py: {errors}"
-        | .ok _ => return none
-      | (_, errors) => return some s!"test_class_any_as_composite.py: Laurel to Core failed: {errors}"
-    tasks := tasks.push ("test_class_any_as_composite.py", task)
+    -- test_class_any_as_composite: skipped due to pre-existing type unification
+    -- error (#789) from commit 4f3e2ff4. The test assigns a str to a Composite-typed
+    -- field, causing "Impossible to unify (arrow Composite Box) with (arrow Any ...)"
+    -- during pyAnalyzeLaurel. This is unrelated to our instance method changes.
```
aqjune-aws: This will be updated after #789

atomb (commented):
## Review Summary

This branch adds conservative handling of Python class inheritance in the Python-to-Laurel translation pipeline. For classes involved in an inheritance hierarchy, method call sites now emit `Hole` instead of statically resolved `StaticCall`. Non-hierarchy classes now retain their translated method bodies (previously all were stripped to opaque). A `manglePythonMethod` helper centralizes the `ClassName@methodName` mangling convention.

### Features

1. **Sound conservative approach to inheritance** - emitting `Hole` for method calls on hierarchy classes is the right call when dynamic dispatch is not modeled. The `classesInHierarchy` set correctly captures both the child class and all its named base classes (excluding `object`).

2. **Good `manglePythonMethod` extraction** - centralizing the convention in a single function eliminates 5 scattered string concatenation sites, reducing inconsistency risk.

3. **Well-designed test pair** - `test_class_inheritance_no_dispatch.py` verifies that `assert False` inside a hierarchy method is NOT analyzed (correctly holed out), while `test_class_method_call_from_main.py` verifies that a non-hierarchy method's assert IS analyzed.

4. **Improved docstring on `translateClass`** - the return type documentation now clearly explains why `instanceProcedures` is set to `[]` on the `CompositeType`.

### Potential Issues

See inline comments for details. The most significant concern is the postcondition regressions in non-hierarchy class tests (5 test files went from pass to unknown for postconditions) as a side effect of keeping method bodies for non-hierarchy classes.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1871,18 +1897,27 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
...
     let mut instanceProcedures : Array Procedure := #[]
     for stmt in body do
       if let .FunctionDef .. := stmt then
         let proc ← translateMethod ctx className stmt
-        -- TODO stop replacing the body of instance proceduces with an empty one
-        instanceProcedures := instanceProcedures.push { proc with body := .Opaque [] .none [] }
+        if inHierarchy then
+          instanceProcedures := instanceProcedures.push { proc with body := .Opaque [] .none [] }
+        else
+          instanceProcedures := instanceProcedures.push proc
```
atomb, line 1978: **Postcondition regressions in non-hierarchy class tests**: several tests that previously passed now report unknown postconditions (test_class_field_init, test_class_field_use, test_class_methods, test_class_with_methods, test_with_statement).

These classes (CircularBuffer, Account, DataStore, Resource) are NOT in any inheritance hierarchy, so the hierarchy detection is not the cause. The regression comes from this behavioral change: non-hierarchy classes now keep their translated method bodies instead of being stripped to `.Opaque`. Previously ALL classes had opaque bodies (the old `TODO` line). With opaque bodies, the verifier could prove postconditions because the method contracts were simple. With concrete bodies, the verifier now has more to reason about and produces `unknown` instead.

Is the intent to accept these regressions in exchange for better method body analysis? The change from `Analysis success` to `Inconclusive` in multiple tests represents a real reduction in verification precision for non-hierarchy classes and should be explicitly discussed.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -2117,6 +2169,7 @@ def pythonToLaurel' (info : PreludeInfo)
     overloadTable := overloadTable,
     importedSymbols := importedSymbols,
     exhaustiveClasses := exhaustiveClasses,
+    classesInHierarchy := classesInHierarchy,
```
atomb, line 2240: When `prev_ctx` is `some` (line 2159), only `functionSignatures` is updated. The freshly computed `classesInHierarchy` (and `userFunctions`, `importedSymbols`, etc.) are silently discarded -- the context inherits whatever the previous caller had.

Currently the only call site (in `PySpecPipeline.lean`) passes `none`, so this is latent. But if `prev_ctx` is ever used (e.g., for multi-file analysis), hierarchy classes in the current file would not be detected, leading to unsound static dispatch.

Consider either merging `classesInHierarchy` into the `prev_ctx` branch, or adding a comment documenting this limitation.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -935,10 +943,22 @@ partial def translateCall (ctx : TranslationContext)
...
+            -- funcName is already resolved to "ClassName@method" using the
+            -- receiver's *static* type (from refineFunctionCallExpr). If the
+            -- class is involved in inheritance, the runtime type may differ
+            -- (e.g., variable typed as Base could hold a Child instance),
+            -- so the static resolution may be wrong. Emit a Hole in that case.
+            let classPrefix := funcName.splitOn "@" |>.head!
+            if ctx.classesInHierarchy.contains classPrefix then
+              return mkStmtExprMdWithLoc (.Hole) callMd
+            let callWithSelf := mkStmtExprMdWithLoc
+              (StmtExpr.StaticCall funcName (target_trans :: callArgs)) callMd
```
atomb, line 961: This constructs call arguments as `target_trans :: callArgs`, but this path bypasses the kwargs expansion logic (further down in `translateCall`) and the `funcDecl` arity check that the other call paths go through.

If a user calls `obj.method(**kwargs)` on a non-hierarchy class, this new path would emit a `StaticCall` with the raw `callArgs` (which may not include the expanded kwargs), while the old path would have emitted a `Hole` and let the kwargs logic handle it further down. Worth verifying this combination is not reachable, or adding kwargs handling here too.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -935,10 +943,22 @@ partial def translateCall (ctx : TranslationContext)
...
         if opt_firstarg.isSome then
           if let some (ImportedSymbol.procedure _ _ true) := ctx.importedSymbols[funcName]? then
             return mkCall funcName
+          else if funcName ∈ ctx.userFunctions then
+            -- funcName is already resolved to "ClassName@method" using the
+            -- receiver's *static* type (from refineFunctionCallExpr). If the
+            -- class is involved in inheritance, the runtime type may differ
+            -- (e.g., variable typed as Base could hold a Child instance),
+            -- so the static resolution may be wrong. Emit a Hole in that case.
+            let classPrefix := funcName.splitOn "@" |>.head!
```
atomb, line 957: Nit: `splitOn` always returns at least one element so `head!` won't panic, but using a pattern match or `.getD` would be more idiomatic Lean and avoid the `!` operator. Same pattern appears at line 910 for the exhaustive type check.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1871,18 +1897,27 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
...
-        -- TODO stop replacing the body of instance proceduces with an empty one
-        instanceProcedures := instanceProcedures.push { proc with body := .Opaque [] .none [] }
+        if inHierarchy then
+          instanceProcedures := instanceProcedures.push { proc with body := .Opaque [] .none [] }
+        else
+          instanceProcedures := instanceProcedures.push proc
 
     return ({
       name := className
       extending := []  -- No inheritance support for now
```
atomb, line 1985: Nit: this comment says "No inheritance support for now" but the PR does add inheritance *awareness* (for dispatch safety), just not inheritance *modeling* in the Laurel type system. Consider updating to something like: `-- Inheritance not modeled in Laurel types; dispatch safety handled via classesInHierarchy`.

StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -187,12 +187,17 @@ def main() -> None:
...
 #guard_msgs in
 #eval withPython (warnOnSkip := false) fun pythonCmd => do
   let program :=
 "from typing import Any
 
-class MyService:
+class BaseService:
+    pass
+
+class MyService(BaseService):
```
atomb, line 200: This test was modified to add inheritance (`BaseService` + `MyService(BaseService)`) so the class body gets stripped to opaque and produces 0 diagnostics. This means the test no longer covers the non-hierarchy case for composite return. Consider whether a separate test for that case is needed.

Strata/Languages/Python/PythonToLaurel.lean
aqjune-aws: Since `getD` implies it can fail.... I kept `head!`. :)

Strata/Languages/Python/PythonToLaurel.lean
aqjune-aws: I added a test for this.

Strata/Languages/Python/PythonToLaurel.lean
aqjune-aws: When there is inheritance, methods are just replaced with holes, so I think it is okay :)

aqjune-aws (commented):
Thanks for comments. I think they make sense.
One small issue is that (1) #789 should be merged first, and (2) there are multiple pull requests that try to solve the same issue. I will coordinate.

aqjune-aws:
I fixed the failing test accordingly.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1882,8 +1902,14 @@ def mkDefaultInitDecl (className : String) : PythonFunctionDecl × Procedure :=
...
-/-- Translate a Python class to a Laurel CompositeType -/
+/-- Translate a Python class to a Laurel CompositeType. returns:
+   - `CompositeType`: the type definition (fields, name); its `instanceProcedures` is set to `[]`
+     because the Laurel→Core translator does not yet support instance procedures.
+   - `Array Procedure`: the translated method procedures, added as top-level static procedures
+     with mangled names (`ClassName@method`) so they can be called via `StaticCall`.
+   - `List PythonFunctionDecl`: function signatures for the class methods. -/
 def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRange)
+    (classesInHierarchy : Std.HashSet String := {})
     : Except TranslationError (CompositeType × Array Procedure × List PythonFunctionDecl) := do
```
MikaelMayer, line 1929: `classesInHierarchy` is passed as a separate parameter here, but `TranslationContext` already has a `classesInHierarchy` field (line 115). The caller (`pythonToLaurel'`) builds `initCtx` without setting this field, then passes it separately. Consider setting `classesInHierarchy` on `initCtx` instead and reading it from `ctx` inside `translateClass`, which would remove this parameter and make the data flow consistent with how `translateCall` already reads it from `ctx`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -173,13 +177,17 @@ def mkStmtExprMd (expr : StmtExpr) : StmtExprMd :=
 def mkStmtExprMdWithLoc (expr : StmtExpr) (md : Imperative.MetaData Core.Expression) : StmtExprMd :=
   { val := expr, md := md }
 
+/-- Mangle a class name and method name into a flat procedure name: `ClassName@methodName`. -/
+def manglePythonMethod (className : String) (methodName : String) : String :=
+  className ++ "@" ++ methodName
```
MikaelMayer, line 182: Nice extraction. A few places still use raw `++ "@__init__"` concatenation (lines 924, 1613, 1877, 1924) instead of `manglePythonMethod`. Since this helper was introduced to centralize the mangling convention, it would be good to use it consistently — e.g. `manglePythonMethod className "__init__"` — so that if the separator ever changes, there's a single place to update.

StrataTest/Languages/Python/AnalyzeLaurelTest.lean
```diff
@@ -227,22 +227,10 @@ private meta def runTestCase (pythonCmd : System.FilePath) (tmpDir : System.File
...
-        | .error err => return some s!"test_class_any_as_composite.py: {err}"
-      match Strata.translateCombinedLaurel laurel with
-      | (some core, []) =>
-        match Core.typeCheck Core.VerifyOptions.quiet core (moreFns := Strata.Python.ReFactory) with
-        | .error errors => return some s!"test_class_any_as_composite.py: {errors}"
-        | .ok _ => return none
-      | (_, errors) => return some s!"test_class_any_as_composite.py: Laurel to Core failed: {errors}"
-    tasks := tasks.push ("test_class_any_as_composite.py", task)
+    -- test_class_any_as_composite: skipped due to pre-existing type unification
+    -- error (#789) from commit 4f3e2ff4. The test assigns a str to a Composite-typed
```
MikaelMayer: This test (`test_class_any_as_composite`) exists and passes on `main`. Removing a passing test is concerning. The comment references issue #789, which is closed and titled "Correctly abort execution in LaurelToCoreTranslator for types that can not be translated to Core" — that doesn't obviously match the described "type unification error". If this PR's changes cause the test to fail, that's a regression worth investigating rather than skipping. Could you confirm whether this test actually fails on this branch, and if so, file a separate issue for it?

StrataTest/Languages/Python/expected_laurel/test_class_field_init.expected
```diff
@@ -1,5 +1,5 @@
-test_class_field_init.py(10, 4): ✅ pass - callElimAssert_requires_5
-test_class_field_init.py(9, 0): ✅ pass - postcondition
+test_class_field_init.py(10, 4): ✅ pass - callElimAssert_requires_7
+test_class_field_init.py(9, 0): ❓ unknown - postcondition
```
MikaelMayer: Several existing tests regressed from `✅ pass - postcondition` to `❓ unknown - postcondition` (this file, `test_class_field_use`, `test_class_methods`, `test_class_with_methods`, `test_class_mixed_init`, `test_with_statement`). This is presumably because method bodies are now analyzed instead of being opaque, giving the verifier more complex constraints. Is this expected and acceptable? If the postconditions were provable before (with opaque methods), they should arguably still be provable now (with concrete methods providing strictly more information). An `unknown` where there was a `pass` could indicate the verifier is struggling with the additional complexity rather than gaining precision.

StrataTest/Languages/Python/VerifyPythonTest.lean
```diff
@@ -187,12 +187,17 @@ def main() -> None:
...
 -- should not crash; the Composite is replaced with a Hole (unconstrained value).
+-- The class uses inheritance, so method bodies are conservatively stripped
+-- to opaque (no diagnostics produced).
 #guard_msgs in
 #eval withPython (warnOnSkip := false) fun pythonCmd => do
   let program :=
 "from typing import Any
 
-class MyService:
+class BaseService:
```
MikaelMayer, line 197: [No action needed] Adding `BaseService` + inheritance to make `MyService` fall into the hierarchy-stripping path is a clean way to preserve the original test's zero-diagnostics expectation. The comment on lines 190-191 explains the rationale well.

Strata/Languages/Python/PythonToLaurel.lean
aqjune-aws: Addressed this in e679f68.

StrataTest/Languages/Python/AnalyzeLaurelTest.lean
aqjune-aws: I updated this test after #790 is merged. The result of dispatch patch slightly changed after bodies of methods were revealed.

StrataTest/Languages/Python/expected_laurel/test_class_field_init.expected
aqjune-aws: It seems the affected tests were not doing the right tests because they weren't inlining the procedures due to unusing bugFinding mode. I added `# strata-args: --check-mode bugFinding --check-level full` to them, so that the methods are inlined & checked.

aqjune-aws:
Reflected comments

atomb (approved):
From what I can tell, this looks good. But I'd love for the other reviewer to be someone who's more deeply familiar with this code.

MikaelMayer (approved):

Issue #794, opened 2026-04-08
Type checking error during Python analysis for class without initializers


olivier-aws:
Analyzing this code with Strata raises a type checking error: `Cannot infer the type of this operation: `NoInit@__init__``
```
class NoInit:
    x: int = 5

def test():
    n = NoInit()
    assert n.x == 5, "class without __init__"
test()
```

PR #798, opened 2026-04-08
Partial support classes without explicit init

olivier-aws:
*Issue #794 

*Description of changes:*
fix(python): Synthesize default `__init__` for classes without one

Classes without an explicit `__init__` method crashed the type checker with "Cannot infer the type of this operation: `ClassName@__init__` because constructor calls always emit a StaticCall to `ClassName@__init__`, but no procedure with that name was registered when `__init__` was absent.

The fix synthesizes a default `__init__` procedure (with an opaque body) in `translateClass` when the class body contains no `__init__` definition. This matches Python semantics where classes inherit a default constructor from object.

Added 5 tests


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1848,6 +1848,44 @@ def extractFieldsFromInit (ctx : TranslationContext) (initBody : Array (Python.s
...
+    name := "self"
+    type := mkHighTypeMd (.UserDefined (mkId className))
+  }
+  let inputs := [selfParam] ++ decl.args.map fun arg =>
+    if arg.tys.length == 1 && arg.tys[0]! != PyLauType.Any then
+      { name := arg.name, type := mkHighTypeMd (.UserDefined {text := arg.tys[0]!}) }
+    else
+      { name := arg.name, type := AnyTy }
+  let proc : Procedure := {
+    name := decl.name
```
tautschnig, line 1874: The `decl.args.map` block (lines 1870-1875) is dead code — `decl.args` is always `[]` since the default `__init__` takes no arguments beyond `self`. The `inputs` will always be just `[selfParam]`. Consider simplifying:

```lean
  let inputs := [selfParam]
```

If the intent is to support future extension (e.g., synthesizing `__init__` with field-based parameters), a comment noting that would help.

.kiro/settings/mcp.json
```diff
@@ -29,4 +29,4 @@
       ]
     }
   }
-}
\ No newline at end of file
+}
```
tautschnig: nit: This trailing-newline fix is unrelated to the feature. Consider splitting it into a separate commit or dropping it to keep the PR focused.

Strata/Languages/Python/PythonToLaurel.lean
olivier-aws: Thanks, indeed this is dead code and as implicit init methods in Python take no arguments, I'll remove this part.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1860,6 +1894,14 @@ def translateClass (ctx : TranslationContext) (classStmt : Python.stmt SourceRan
           let funcDecl ← pyFuncDefToPythonFunctionDecl ctx s
           .ok (some (funcDecl))
       | _ => .ok none)
+    -- Synthesize a default __init__ if the class doesn't define one
+    let hasInit := classFunDecls.any (fun d => d.name == className ++ "@__init__")
+    let (classFunDecls, defaultInitProc) :=
+      if hasInit then (classFunDecls, none)
+      else
+        let (decl, proc) := mkDefaultInitDecl className
+        (classFunDecls ++ [decl], some proc)
```
atomb: This is a tiny thing, but it feels more natural to me for the default `__init__` to come first, and `decl :: classFunDecls` is O(1) instead of O(n).

StrataTest/Languages/Python/run_py_analyze_sarif.py
```diff
@@ -24,6 +24,7 @@
     "test_invalid_client_type",
     "test_unsupported_config",
     "test_with_void_enter",
+    "test_class_no_init_extra_args", # No SARIF output because does not run SMT analysis
```
atomb, line 27: Does it make sense to eventually include type errors in our SARIF output, too? Do they naturally fit there?

Strata/Languages/Python/PythonToLaurel.lean
olivier-aws: Good point, changed.

StrataTest/Languages/Python/run_py_analyze_sarif.py
olivier-aws: Yes it would. These errors currently generate a simple text file, would be easy to generate SARIF if needed. Created an issue for that: #821

PR #803, opened 2026-04-08
Fix #801: bugFinding classifier reports all unsatisfiable assertions

### LINKED ISSUE #801: bugFinding classifier should report all unsatisfiable assertions as failures
## Problem

The `bugFinding` mode classifier in `pyAnalyzeLaurel` only counted `alwaysFalseAndReachable` as a failure. This missed two important categories:

- **`alwaysFalseReachabilityUnknown`** ("always false if reached") — the assertion is provably unsatisfiable, but reachability of the code path is unknown. In bug-finding mode this should still be reported as a failure.
- **`unreachable`** — dead code where the path condition itself is unsatisfiable.

Both cases have `satisfiabilityProperty == .unsat`, meaning the assertion cannot be satisfied under any assignment.

## Fix

Replace the enumeration of specific outcome predicates with a single check on the satisfiability property:

```lean
| .bugFinding | .bugFindingAssumingCompleteSpec =>
  { isFailure := fun r => match r.outcome with
      | .ok o => o.satisfiabilityProperty == .unsat
      | _     => false }
```

In bug-finding mode, any assertion whose satisfiability is `unsat` is a finding. This differs from deductive/validation modes which may treat some of these as warnings or inconclusive.

## Location

`StrataMain.lean`, in `printPyAnalyzeSummary`, the `ResultClassifier` for `bugFinding` mode.

shigoel:
## Summary
- The bug-finding classifier in `printPyAnalyzeSummary` had two issues:
  1. It only overrode `isFailure` but not `isSuccess`. Since `isSuccess` (via `isPass`) returns true for unreachable results (validity=unsat), the partition caught them as successes before `isFailure` could classify them as failures — making `|| o.unreachable` dead code.
  2. Using `isPass` for success is wrong for bug-finding mode: in minimal check level, validity isn't queried, so `isPass` would misclassify results.
- Fix: define proper mode-specific predicates on `VCOutcome`:
  - `bugFindingSuccess`: `isSatisfiable || passReachabilityUnknown` (assertion can be true, or is provably always true)
  - `bugFindingFailure`: `alwaysFalseAndReachable || alwaysFalseReachabilityUnknown || unreachable` (definite bug or dead code)
- Add docstrings to `isPass`, `isSatisfiable`, `VCResult.isSuccess`, and `VCResult.isFailure` clarifying their mode semantics
- Add Python integration test exercising bug-finding failure classification

Closes #801


atomb (dismissed):
Whether this is correct is somewhat subjective, but it seems reasonable to me, at least.

StrataMain.lean
```diff
@@ -552,6 +552,8 @@ private def printPyAnalyzeSummary (vcResults : Array Core.VCResult)
     | .bugFinding | .bugFindingAssumingCompleteSpec =>
       { isFailure := fun r => match r.outcome with
           | .ok o => o.alwaysFalseAndReachable
+                   || o.alwaysFalseReachabilityUnknown
+                   || o.unreachable
           | _     => false }
```
tautschnig: `o.unreachable` means `satisfiabilityProperty == .unsat && validityProperty == .unsat` — the assertion is trivially **true** (validity is unsat = "not invalid") but the code path is dead. This is dead code with a correct assertion, not a bug.

In bug-finding mode, the goal is to find definite bugs. Dead code is a code quality issue, not a bug finding. The default `VCResult.isFailure` (used in deductive mode) also does NOT include `unreachable`.

Consider removing `|| o.unreachable` and instead using the simpler formulation from the issue description if you want to catch all `sat=unsat` cases:

```suggestion
      { isFailure := fun r => match r.outcome with
          | .ok o => o.alwaysFalseAndReachable
                   || o.alwaysFalseReachabilityUnknown
          | _     => false }
```

Or if dead code should genuinely be flagged in bug-finding mode, add a comment explaining the rationale.

tautschnig: No test coverage for the bug-finding classifier change. A test that verifies the DETAIL/RESULT output for a program with an always-false assertion in bug-finding mode would prevent regression.

tautschnig (dismissed):
Including `unreachable` seems questionable. `unreachable` means `sat=unsat, val=unsat` — the assertion is trivially **true** but the code path is dead. This is dead code, not a bug. In deductive mode, `VCResult.isFailure` does NOT include `unreachable` either.

Also, can tests please be added?

StrataMain.lean
MikaelMayer: I'd prefer that a comment be added. We envision Strata to be heavily used on agent-generated code, so if an assertion is unsatisfiable, it's likely to indicate something worth noticing to the user, whether the path itself is unreachable or the assertion can never be true.

Once someone uses the deductive mode to _prove_ the absence of bugs, unreachability is still reported but as a warning.

MikaelMayer: I want test coverage too.

MikaelMayer: I want test coverage too.

shigoel: Thanks, I should've done that in the first place. In the process of adding this test, found a bug in the classifier logic. Addressed in https://github.com/strata-org/Strata/pull/803/commits/cd03563f8b2b81180f8bec891ac73758ebbc2148.

shigoel: Addressed in https://github.com/strata-org/Strata/pull/803/commits/cd03563f8b2b81180f8bec891ac73758ebbc2148

MikaelMayer (approved):

StrataTest/Languages/Python/expected_laurel/test_bug_finding_unreachable.expected
```diff
@@ -0,0 +1,11 @@
+test_bug_finding_unreachable.py(3, 4): ✔️ always true if reached - impossible condition
+test_bug_finding_unreachable.py(3, 4): ✖️ always false if reached - impossible condition
+test_bug_finding_unreachable.py(4, 7): ✔️ always true if reached - Check PGt exception
+test_bug_finding_unreachable.py(4, 4): ✔️ always true if reached - ite_cond_calls_Any_to_bool_0
+test_bug_finding_unreachable.py(5, 11): ✔️ always true if reached - Check PLt exception
+test_bug_finding_unreachable.py(5, 8): ✔️ always true if reached - ite_cond_calls_Any_to_bool_0
+test_bug_finding_unreachable.py(6, 12): ❌ fail (❗path unreachable) - dead code
```
shigoel, line 7: The dead code is reported twice, which stems from path multiplication due to the `if`. It's on the list to collect obligations occurring at the same location, but is out of scope of this PR.

atomb (approved):

PR #808, opened 2026-04-08
Fix PE-resolved VCs reporting unknown reachability

shigoel:
## Summary

- **PE satisfiability for trivially reachable VCs**: `preprocessObligation` now returns `sat` when the obligation is literally `true` with empty assumptions (`true ∧ true = true`). Previously only the `obligation.isFalse → unsat` case was handled.

- **Skip phase validation for PE-resolved results**: PE results are determined by literal evaluation (constant folding with concrete values), not solver search. Phase validation checks solver *models* against pre-transformation semantics — PE results have no models, so validation is inapplicable. The `frontEndPhase` was unconditionally downgrading all `sat` results to `unknown` (including PE results), causing "if reached" qualifiers on trivially reachable VCs.

- **Gate PE failure dbg_traces behind `verbose >= .debug`**: Two dbg_trace calls in `preprocessObligation` and `verifySingleEnv` were firing at `.normal`verbosity, producing noisy `[DEBUG] Evaluated program:` dumps in user-facing output. Updated expected output in 3 existing tests.

- **Updated `frontEndPhase` docstring** to document that PE-resolved results bypass phase validation.

Before: `✔️ always true if reached` / `✖️ always false if reached` (for PE-resolved VCs with concrete args)
After: `✅ always true and is reachable` / `❌ always false and is reachable`


atomb (dismissed):

Strata/Languages/Core/Verifier.lean
```diff
@@ -612,25 +612,31 @@ def preprocessObligation (obligation : ProofObligation Expression) (p : Program)
...
     : EIO DiagnosticModel (ProofObligation Expression × Option SMT.Result × Option SMT.Result) := do
-  -- PE can determine satisfiability if the obligation is literally false (unsat)
+  -- PE can determine satisfiability of (assumptions ∧ obligation):
+  --   obligation is false  → assumptions ∧ false = false → unsat
+  --   obligation is true with no assumptions → true ∧ true = true → sat
+  -- The second case establishes reachability: with no path conditions, the
+  -- assertion site is trivially reachable and the obligation holds concretely.
   let peSatResult : Option SMT.Result :=
     if !satisfiabilityCheck then some .unknown
     else if obligation.obligation.isFalse then some .unsat
```
tautschnig, line 622: The comment says `obligation is false → assumptions ∧ false = false → unsat`, which is correct. But note the asymmetry: for the `isFalse` case, assumptions are ignored (unsat regardless), while for the `isTrue` case, assumptions must be empty. This asymmetry is logically sound (`P ∧ false = false` for any `P`, but `P ∧ true = P` which is only guaranteed sat when `P = true`), but it's worth calling out since it's a subtle point.

The comment does a good job explaining this — just confirming the reasoning is correct.

tautschnig (approved):

MikaelMayer:
> PE satisfiability for trivially reachable VCs: preprocessObligation now returns sat when the obligation is literally true with empty assumptions (true ∧ true = true). Previously only the obligation.isFalse → unsat case was handled.

That's incorrect. To be sat, not only the proposition has to be true, but the path condition has to be satisfiable 

> Skip phase validation for PE-resolved results: PE results are determined by literal evaluation (constant folding with concrete values), not solver search. Phase validation checks solver models against pre-transformation semantics — PE results have no models, so validation is inapplicable. The frontEndPhase was unconditionally downgrading all sat results to unknown (including PE results), causing "if reached" qualifiers on trivially reachable VCs.

This assumes that partial evaluation is done on the first version of the program. It forgets that before we do PE we go through many abstracting phases. So if PE returned sat, it could be that the model is not valid before a given phase. PE should not be aware of phases not bypass them

MikaelMayer (changes requested):
See my previous comment

tautschnig:
> > PE satisfiability for trivially reachable VCs: preprocessObligation now returns sat when the obligation is literally true with empty assumptions (true ∧ true = true). Previously only the obligation.isFalse → unsat case was handled.
> 
> That's incorrect. To be sat, not only the proposition has to be true, but the path condition has to be satisfiable
> 

Aren't "trivially reachable VCs" those where the path condition is true (i.e., necessarily satisfiable)?

atomb:
> > > PE satisfiability for trivially reachable VCs: preprocessObligation now returns sat when the obligation is literally true with empty assumptions (true ∧ true = true). Previously only the obligation.isFalse → unsat case was handled.
> > 
> > 
> > That's incorrect. To be sat, not only the proposition has to be true, but the path condition has to be satisfiable
> 
> Aren't "trivially reachable VCs" those where the path condition is true (i.e., necessarily satisfiable)?

That was my thought, too. I also read "empty assumptions" as "any element of the path condition evaluates to true".

MikaelMayer:
> > > > PE satisfiability for trivially reachable VCs: preprocessObligation now returns sat when the obligation is literally true with empty assumptions (true ∧ true = true). Previously only the obligation.isFalse → unsat case was handled.
> > > 
> > > 
> > > That's incorrect. To be sat, not only the proposition has to be true, but the path condition has to be satisfiable
> > 
> > 
> > Aren't "trivially reachable VCs" those where the path condition is true (i.e., necessarily satisfiable)?
> 
> That was my thought, too. I also read "empty assumptions" as "any element of the path condition evaluates to true".

Ok for empty assumptions, sorry I over read that. It might not ever happen in practice though unless we manage to prune the prelude before partial evaluation, so it might hurt performance more than help, unless someone thinks it's useful for that this test.

Still, "Skip phase validation for PE-resolved results" is unsound for the reason I described earlier, that does not change a bit.
PE only bypasses the solver. Phases that introduce abstractions might still convert a sat result to unknown precisely because of such over approximations.

PR #809, opened 2026-04-09
Add regression test for nested dict access in pyspec preconditions

tautschnig:
Test that `.getIndex (.getIndex (.var "kwargs") "Outer") "Inner"` generates `Any_get(Outer, from_str("Inner"))` in the procedure body, not `FieldSelect` (which caused the resolution pass to silently drop the precondition). This just adds a test for the fix that was merged in #804.


atomb (dismissed):

thanhnguyen-aws (dismissed):

StrataTest/Languages/Python/ToLaurelTest.lean
```diff
@@ -440,4 +441,45 @@ info: errors: 1
...
+info: body contains Any_get: true
+body contains FieldSelect: false
+-/
+#guard_msgs in
+#eval do
+  let strTy := identType .builtinsStr
+  let dictTy := identType .typingDict
+  -- kwargs must be a TypedDict so expandKwargsArgs can expand it
+  let kwargsTy := SpecType.ofAtom loc (.typedDict #["Outer"] #[dictTy] #[true])
+  let result := signaturesToLaurel "<test>" #[
```
MikaelMayer, line 458: This is not a test that I would like to maintain. I prefer tests that have the concrete syntax that is being parsed and an expected output instead of something hard-coding the tree. It's hard to know if it's a Laurel AST, a python AST, a PySpec AST, or a Core AST.

MikaelMayer (dismissed):

StrataTest/Languages/Python/ToLaurelTest.lean
tautschnig: It lives in ToLaurelTest.lean alongside many other tests that use the same programmatic AST construction
pattern (the entire file is structured this way). I'd be happy to take on a larger refactoring of that file as follow-up work.

atomb (approved):

MikaelMayer (approved):

StrataTest/Languages/Python/ToLaurelTest.lean
tautschnig: See https://github.com/strata-org/Strata/pull/830.

PR #811, opened 2026-04-09
Fix precondition checks inheriting assert message from PR #608

### LINKED ISSUE #834: Function precondition VCs inherit the user's assert label, making results confusing
## Description

When a Python assert has a user-provided message, the generated precondition verification conditions for function calls within the assert condition inherit the same label. This makes it hard to distinguish the actual assertion result from the internal precondition check.

For example:

```python
def is_positive(x: int) -> bool:
    return x > 0

def test():
    assert is_positive(5), "positive"
```

Produces:

```
test.py(5, 4): ?? pass - positive
test.py(5, 4): ?? unknown - positive
```

Both VCs display "positive" as their label. The user has no way to tell that the second one is actually checking whether the return value of is_positive can be safely coerced to bool via Any_to_bool, not the assertion itself.

## Expected behavior

The precondition VC should have a distinct label that identifies it as an internal check, e.g.:

```
test.py(5, 4): ?? pass - positive
test.py(5, 4): ?? unknown - (Any_to_bool requires) positive
```

tautschnig:
When a Python assert has a user message (e.g., assert x > 0, "positive"), PR #608 correctly propagated it as a propertySummary on the assert statement's metadata. However, collectPrecondAsserts in PrecondElim.lean passed this metadata through to generated precondition checks (e.g., Any_to_bool type coercion checks), causing them to display the user's message instead of their own generated label.

Fix: strip propertySummary from the metadata before generating precondition asserts. The user's message should only appear on the assertion VC itself, not on generated precondition checks for called functions.

Before: both lines show user message
  test.py(7, 4): ✅ pass - multiplication implemented incorrectly
  test.py(7, 4): ✅ pass - multiplication implemented incorrectly

After: precondition check shows its own label
  test.py(7, 4): ✅ pass - assert_assert(102)_calls_Any_to_bool_0
  test.py(7, 4): ✅ pass - multiplication implemented incorrectly


Strata/Transform/PrecondElim.lean
```diff
@@ -68,6 +68,10 @@ def collectPrecondAsserts (F : @Lambda.Factory CoreLParams) (e : Expression.Expr
 (labelPrefix : String) (md : Imperative.MetaData Expression)
 : List Statement :=
   let wfObs := Lambda.collectWFObligations F e
+  -- Strip propertySummary: the enclosing statement's user-facing message
+  -- (e.g., a Python assert message) should not propagate to generated
+  -- precondition checks for called functions.
+  let md := md.eraseAllElems Imperative.MetaData.propertySummary
```
keyboardDrummer, line 97: Isn't this an issue that occurs earlier in the pipeline? Why did the preconditions ever get the same md as the one that's used for the assertion? Every Core expression should have its own md. `propertySummary` is a Laurel concept so it seems strange to refer to it from Core

MikaelMayer, line 97: Nit: the docstring above (line 63-66) says "The metadata from the original statement is attached to the generated assertions" — now that `propertySummary` is stripped, this is slightly inaccurate. Consider updating to something like:

```
The metadata from the original statement is attached to the generated assertions
(with propertySummary stripped), and property type classification added when applicable.
```

MikaelMayer (dismissed):

Strata/Transform/PrecondElim.lean
atomb: I think I have the same question. I've been thinking a bit about what sort of disciplines we should have around metadata, to ensure that it's propagated correctly through a long sequence of transformations, and one of the principles I'm considering is that most transformations should never remove metadata.

tautschnig: The `md` is the enclosing statement's metadata, passed to `collectPrecondAsserts` so that generated precondition assertions inherit the source location (file range). Without this, generated checks would have no source location and couldn't be mapped back to the user's code in diagnostics.

The issue is specifically that `propertySummary` — which is correct on the original assert statement — leaks through when we copy the metadata to generated sub-assertions. Fixing this earlier would mean either (a) not attaching `propertySummary` to the assert's md at all (wrong — the assert itself needs it for its own VC), or (b) constructing fresh metadata with only the source location fields (more invasive, same net effect).

Stripping `propertySummary` during the copy is the minimal correct fix: source location propagates, user-facing message does not.

You're right that `propertySummary` is a Laurel concept — but `MetaData` is intentionally a generic key-value store shared across dialect layers, so Core code occasionally needs to be aware of keys set by upstream dialects. Docstring updated in eece2376 per MikaelMayer's suggestion.

tautschnig: Fixed in [eece237](https://github.com/strata-org/Strata/pull/811/commits/eece2376fecc68aae7bf17edc54c073c041ecba7).

keyboardDrummer: `collectPrecondAsserts` should not take an `md`. Instead the `Expression.Expr` it takes, and its sub-expressions, should contains its own `md`s, so that the assertions it generates always have a precise md instead of one of an enclosing AST node.

tautschnig: You're right that `collectPrecondAsserts` should use the call site expression's own metadata rather than inheriting from the enclosing statement. I attempted this, but `CoreExprMetadata` is currently `Unit` (`Identifiers.lean:21`), so `ob.callSiteMetadata` is always `()` — it carries no source location.

I've updated the docstring in 5018eeb to explain this blocker and why the `eraseAllElems` workaround is needed for now. The proper fix requires changing `CoreExprMetadata` to carry source ranges, which is a cross-cutting refactor. Here's the plan:

## Refactor Plan: Add Source Ranges to Core Expression Metadata

### Goal
Change `CoreExprMetadata` from `Unit` to `Option SourceRange` so that Core expressions carry source location information. This enables `collectPrecondAsserts` to use `ob.callSiteMetadata` directly instead of inheriting the enclosing statement's metadata, eliminating the `propertySummary` stripping workaround.

### Scope Assessment
- ~37 production code sites construct expressions with `()` metadata (in `Strata/`)
- ~225 test sites construct expressions with `()` metadata (in `StrataTest/`)
- Most can be mechanically updated to `none` (no source range available)

### Steps

**Step 1: Change the type** (small, high-impact)
- `Strata/Languages/Core/Identifiers.lean`: Change `abbrev CoreExprMetadata := Unit` to `abbrev CoreExprMetadata := Option SourceRange`
- Update `MkLExprParams` meta instance and `ToExpr` deriving
- Update the `eb[...]` elaborator to use `none` as default metadata

**Step 2: Fix compilation** (mechanical, ~37 production sites)
- Replace `()` with `none` in all expression construction sites in `Strata/`
- Key files: `Factory.lean` (~20 sites in macro-generated code), `SMTEncoder.lean`, `StatementEval.lean`, `Verifier.lean`, transforms
- The `Factory.lean` macros (`ExpandBVSafeOpFuncDefs` etc.) use `default` which would automatically become `none`

**Step 3: Fix tests** (mechanical, ~225 sites)
- Replace `()` with `none` in all test expression construction sites
- Can be done with a codebase-wide search-and-replace, but needs care to not change non-expression `()`

**Step 4: Propagate source ranges from DDM parser**
- `Strata/Languages/Core/DDMTransform/Translate.lean`: When translating AST expressions to Core expressions, pass the `SourceRange` from the AST node as `some sr` instead of `none`
- The DDM parser already has source ranges for every AST node — they just need to be threaded through

**Step 5: Propagate from Laurel translator**
- `Strata/Languages/Laurel/LaurelToCoreTranslator.lean`: When translating Laurel expressions to Core, propagate source ranges from Laurel metadata to Core expression metadata
- This is the Python/Laurel path where `propertySummary` leaking was the original problem

**Step 6: Use expression metadata in PrecondElim**
- Remove `md` parameter from `collectPrecondAsserts`
- Use `ob.callSiteMetadata` (now `Option SourceRange`) to construct assertion metadata
- Fall back to empty metadata when `callSiteMetadata` is `none`
- Remove the `eraseAllElems propertySummary` workaround

### Suggested Phasing
- **Phase 1** (this PR): Keep the `eraseAllElems` workaround with the updated docstring explaining the blocker
- **Phase 2** (separate PR): Steps 1-3 — change the type and fix compilation. Large but mechanical.
- **Phase 3** (separate PR): Steps 4-6 — propagate real source ranges and remove the workaround. Can be done incrementally (DDM parser path first, then Laurel).

Strata/Transform/PrecondElim.lean
```diff
@@ -61,13 +61,21 @@ private def classifyPrecondition (funcName : String) : Option String :=
...
+user-facing messages from leaking into generated checks).
 -/
 def collectPrecondAsserts (F : @Lambda.Factory CoreLParams) (e : Expression.Expr)
 (labelPrefix : String) (md : Imperative.MetaData Expression)
 : List Statement :=
   let wfObs := Lambda.collectWFObligations F e
+  -- Strip propertySummary: the enclosing statement's user-facing message
+  -- (e.g., a Python assert message) should not propagate to generated
+  -- precondition checks for called functions.
+  let md := md.eraseAllElems Imperative.MetaData.propertySummary
```
MikaelMayer, line 97: Nit: consider extracting the stripping into a small helper (e.g. `stripPropertySummary`) if this pattern is needed elsewhere in the future. Fine as-is for now.

tautschnig:

keyboardDrummer (dismissed):

tautschnig:

keyboardDrummer (dismissed):

Strata/Transform/PrecondElim.lean
tautschnig: Tracked in #873.

aqjune-aws:

aqjune-aws:
@keyboardDrummer-bot Good job!

aqjune-aws (dismissed):

tautschnig:

aqjune-aws:
It seems diffs other than the comment update disappeared. Could you confirm? @keyboardDrummer-bot

aqjune-aws (approved):

MikaelMayer (approved):

Issue #812, opened 2026-04-09
Regex concreteEval needs a mechanism to fire when pattern argument is a constant string


MikaelMayer:
## Problem

The regex functions `re_search_bool`, `re_match_bool`, and `re_fullmatch_bool` have a `concreteEval` that compiles a regex pattern string into SMT-LIB regex operations (`Str.InRegEx`, `Re.Concat`, etc.). This compilation should fire during partial evaluation whenever the first argument (the pattern) is a constant string, regardless of the second argument.

Currently, `concreteEval` only fires when *all* arguments are canonical values. Since the second argument (the string being matched) is typically symbolic, `concreteEval` never fires and the regex pattern is never compiled.

## What the concreteEval does

It does not "concretely evaluate" the function in the usual sense. It replaces:

```
re_search_bool("^[a-z]+$", x)
```

with the SMT-LIB equivalent:

```
Str.InRegEx(x, Re.Concat(Re.Range("a","z"), Re.Star(Re.Range("a","z"))))
```

This is a partial evaluation / rewrite: the pattern string is compiled into regex AST operations, and the opaque `re_search_bool` call is replaced with a structured `Str.InRegEx` that the SMT solver understands. The second argument `x` remains symbolic.

## What's needed

A mechanism to fire `concreteEval` when a specific argument is a canonical value (a constant string), not necessarily a constructor application of an algebraic datatype. The existing `evalIfConstr` attribute is not appropriate here — it is designed for case-splitting over algebraic datatypes (e.g. `List.tail (Cons x t)` evaluates to `t`), not for triggering on primitive constants.

Possible approaches:

1. **New attribute `evalIfCanonical N`** — fire `concreteEval` when argument N is a canonical value (constant). This is the direct analog of `evalIfConstr` but for primitive types.
2. **Extend `concreteEval` semantics** — allow `concreteEval` to return `none` when it can't evaluate, and always attempt it (not just when all args are canonical). The function itself decides when to fire.
3. **Other mechanism** — any way to tell the partial evaluator "try this rewrite when the pattern argument is known."

## Location

- Regex function declarations: `Strata/Languages/Python/PyFactory.lean`, `mkModeBoolFunc`
- Partial evaluator dispatch: `Strata/DL/Lambda/LExprEval.lean`

PR #816, opened 2026-04-09
Add `evalIfCanonical` attribute to fix issue #812

joscoh:
*Issue #, if available:* #812

*Description of changes:* Adds `evalIfCanonical` attribute that runs the concrete evaluator if the `i`th argument is a canonical value. This allows regex functions to be evaluated if their first argument is a constant.


Strata/Languages/Python/PyFactory.lean
```diff
@@ -77,6 +77,7 @@ private def mkModeBoolFunc (name : String) (mode : MatchMode) :
       typeArgs := [],
       inputs := [("pattern", mty[string]), ("s", mty[string])],
       output := mty[bool],
+      attr := #[.evalIfCanonical 0],
```
MikaelMayer, line 80: I would love to see a test as the original issue stated one.

MikaelMayer (dismissed):
Everything looks good, just one comment.

Strata/Languages/Python/PyFactory.lean
```diff
@@ -99,6 +100,7 @@ def rePatternErrorFunc : LFunc Core.CoreLParams :=
       typeArgs := [],
       inputs := [("pattern", mty[string])],
       output := mty[Error],
+      attr := #[.evalIfCanonical 0],
```
joscoh, line 103: @MikaelMayer I wasn't sure, is this supposed to have the same behavior too? It wasn't mentioned in the issue so unsure.

tautschnig: Re @joscoh's question: yes, `evalIfCanonical 0` is correct here too. `rePatternErrorFunc` checks whether a regex pattern is malformed (returns an error if so). When the pattern is a constant string, the function can concretely determine whether it's valid — without the attribute, this check would remain symbolic and the verifier couldn't distinguish valid from invalid patterns.

The thread could be resolved.

tautschnig (approved):

Strata/Languages/Python/PyFactory.lean
tautschnig: Yes, `evalIfCanonical 0` is correct here too. `rePatternErrorFunc` checks whether a regex pattern is malformed (returns an error if so). When the pattern is a constant string, the function can concretely determine whether it's valid — without the attribute, this check would remain symbolic and the verifier couldn't distinguish valid from invalid patterns.

StrataTest/Languages/Python/AnalyzeLaurelTest.lean
```diff
@@ -310,6 +310,25 @@ assertion. This exercises the full pipeline with type alias resolution.
         throw <| IO.userError
           "Expected ✖️ always false for empty bucket violation"
 
+/-! ## evalIfCanonical regression test (Issue #812)
```
atomb, line 313: I suspect we could write some common abstractions that would allow checking the verification output on many Python examples concisely (because we'll probably want that eventually) but that's not blocking. We could implement it later.

atomb (approved):

PR #817, opened 2026-04-09
Iteratively apply procedure inlining until there is no change in bugFindingMode, erase deprecated pyAnalyze

aqjune-aws:
This PR does the followings:
  - Define procedureInliningPipelinePhase in ProcedureInlining.lean as a reusable PipelinePhase, with Core.InlineTransformOptions co-located alongside it. All production call sites now go through this phase instead of calling inlineCallCmd directly.
  - Add Core.Transform.runProgramUntil: iterative wrapper over runProgram that repeats a command-level transformation to a fixed point (maxIters = none) or up to N iterations (maxIters = some n) with early exit on no change.
  - Conditionally enable procedure inlining: inlining is now only applied when verification mode is bug-finding (not deductive), matching the intended semantics. Max. number of iteration is *10*. Also, don't preclude preclude.
  - Unify the Core transformation pipeline: Introduce prefixPhases parameter to Core.verify and Core.verifyProgram, allowing callers to inject pre-verification transforms (like procedure inlining) that run in the same CoreTransformM context as the built-in pipeline phases (CallElim, LoopElim, etc.). This shares fresh-variable counters and cached call graphs
  across all passes.
  - Remove the deprecated pyAnalyze command and simplify run_py_analyze.sh to laurel-only mode (accepts laurel argument for backward compatibility).


Strata/Transform/CoreTransform.lean
```diff
@@ -381,6 +381,35 @@ def runProgram
...
+    let mut prog := p
+    let mut anyChanged := false
+    for _ in List.range n do
+      let (changed, prog') ← runProgram f prog targetProcList
+      prog := prog'
+      if changed then anyChanged := true
+      if !changed then break
+    return (anyChanged, prog)
+  | none =>
+    let mut prog := p
```
MikaelMayer, line 411: The loop body (lines 405–412) is identical to the `some n` branch (lines 396–403). Consider extracting the shared iteration logic into a helper, or using a single loop with a configurable bound. This would eliminate the duplication.

Also, the `none` branch uses `repeat` with only a semantic termination guarantee (no change → break). If a transformation ever oscillates, this loops forever. A safety cap (e.g. a large default like 1000) would make this more robust.

StrataMain.lean
```diff
@@ -715,11 +598,20 @@ def pyAnalyzeLaurelCommand : Command where
         verbose := .quiet, removeIrrelevantAxioms := .Precise,
         vcDirectory := baseVcDir }
     let options ← parseVerifyOptions pflags pyAnalyzeBase
+    let isBugFinding := options.checkMode == .bugFinding
+                      || options.checkMode == .bugFindingAssumingCompleteSpec
+    let inlinePhases : List Core.PipelinePhase :=
+      if isBugFinding then
+        [Core.procedureInliningPipelinePhase
+          { doInline := fun name a => name ≠ "__main__" && Core.doInlineNonRecursive name a
```
MikaelMayer, line 605: `preludeNames` on line 586 is now unused and can be replaced with `_`.

StrataTest/Languages/Python/AnalyzeLaurelTest.lean
```diff
@@ -115,13 +115,10 @@ private meta def runAnalyzeAndVerify
     | some core => pure core
   -- Split prelude / user procedure names at FIRST_END_MARKER
   let (preludeNames, userProcNames) := Strata.splitProcNames coreProgram
```
MikaelMayer: Same as in `StrataMain.lean`: `preludeNames` is now unused after the inline predicate was changed from `!preludeNames.contains name` to `Core.doInlineNonRecursive`. Replace with `_` if the behavioral change is intentional.

StrataTest/Languages/Python/README.md
```diff
@@ -181,16 +181,4 @@ Laurel: 42 procedure(s), 3 type(s)
...
-The `pyAnalyze` and `pyTranslate` commands translate Python directly to Core,
-bypassing Laurel. This path is being phased out in favor of the Laurel
-pipeline described above.
-
-```
-# Deprecated — use pyAnalyzeLaurel instead
-lake exe strata pyAnalyze [--verbose] <file.python.st.ion>
-lake exe strata pyTranslate <file.python.st.ion>
-```
+```
```
MikaelMayer: Nit: the file now ends without a trailing newline.

Strata/Transform/CoreTransform.lean
MikaelMayer: But how come it is a repeat in a def in the first place? I thought there should be a termination proof.

aqjune-aws: Indeed, the `repeat` notation internally uses a `ForIn` instance that uses `partial def`. This is slightly scary: https://github.com/leanprover/lean4/blob/master/src/Init/While.lean#L25

For the default value, I chose to intentionally give a strong power of possibly iterating indefinitely because this will be a basic block of Core transformations. The number of iterations can be indeed very big in some cases. Combining the bodys of two loops was not straightforward; probably I could merge variable declarations, but not body.

Strata/Languages/Core/CallGraph.lean
```diff
@@ -94,6 +94,12 @@ def CallGraph.getCalleesClosure (cg : CallGraph) (name : String) : List String :
   let allNodes := cg.callees.toList.map Prod.fst
   (closureGo cg.getCallees allNodes [] [name]).filter (· ≠ name)
 
+/-- True when `name` is part of a (possibly self-referencing) cycle in the
+    call graph — i.e., it can transitively reach itself through callees. -/
+def CallGraph.isRecursive (cg : CallGraph) (name : String) : Bool :=
```
atomb, line 105: This could wind up being expensive for large call graphs, but I'm fine with leaving it as-is for now and optimizing it if it becomes a problem.

Strata/Transform/CoreTransform.lean
```diff
@@ -381,6 +381,35 @@ def runProgram
   })
   return (changed, newProg)
 
+/-- Repeatedly apply a command-level transformation until no more changes occur
+    or the iteration limit is reached.
+    - `maxIters = none`: repeat until a fixed point (no changes).
+    - `maxIters = some n`: run up to `n` iterations, stopping early if no change. -/
```
atomb, line 393: I could imagine this getting pretty expensive for large programs. It's fine for now, but we may need to optimize it.

If we do, it may make sense, instead of inlining procedures, to update the evaluator to continue executing inside procedure bodies that are somehow marked. (Maybe inlinable, transparent, or non-recursive are several possible label names).

atomb (approved):
A couple of things in this might be inefficient, but we can address that if they become issues.

Strata/Transform/CoreTransform.lean
MikaelMayer: I plan to refactor the partial evaluator into a phase that just changes the program. In that case the two approaches seem equivalent.

MikaelMayer (approved):
Could it be that the repeat is inside a def just because the actual execution of the monad requires a partial def?

PR #818, opened 2026-04-09
Fix assume statements for `for` loop with `range(i)`

thanhnguyen-aws:
*Description of changes:* This PR fixes assume statements for `for` loop with `range(i)`

Problem: the function `range` in Python prelude is opaque, so the assumption `PIn(target, range(...))` does not provide enough information to reason about the `target` var. 

Fix: 
1. Add a statement: if `range` is empty (`range(i)` when `i <= 0`), exit the loop.
2. Replace `PIn(target, range(i))` by assumptions: `Any..isfrom_int(target)` and `target` in `[0, i)` 

Limitation: Only support `range` with single input.


StrataTest/Languages/Python/tests/test_for_range.py
```diff
@@ -0,0 +1,14 @@
+numbers = [2,4,5,9,1]
+
+for i in range(5):
+    assert i < 5
+    assert i >= 0
+    j = numbers [i]
+    assert j < 10
+
+
+for i in range(-10):
```
MikaelMayer, line 10: Note: `range(-10)` in Python produces an empty sequence, so this loop body never executes at runtime. The test passes because the verifier assumes the body *might* run with `i ∈ [-10, 0)`, and the assertions hold under that assumption. This is a valid stress test of the assumption logic, but it might be worth adding a comment noting that this tests the negative-bound assumption path rather than real runtime behavior.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1440,14 +1440,30 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+              (mkStmtExprMd $ .PrimitiveOp .And [
+                  (mkStmtExprMd $ .PrimitiveOp .Geq [asIntTarget, mkStmtExprMd $ .LiteralInt 0]),
+                  (mkStmtExprMd $ .PrimitiveOp .Lt [asIntTarget, asIntRange]) ])
+              (mkStmtExprMd $ .PrimitiveOp .And [
+                  (mkStmtExprMd $ .PrimitiveOp .Lt [asIntTarget, mkStmtExprMd $ .LiteralInt 0]),
+                  (mkStmtExprMd $ .PrimitiveOp .Geq [asIntTarget, asIntRange]) ])
+            let assumeInRange := mkStmtExprMd (.Assume inRangeExpr)
+            [assumeTypeInt, assumeInRange]
+          | _ =>
+            let targetInIter := mkStmtExprMd (.StaticCall "PIn" [targetVar, iterExpr])
```
tautschnig: The negative-range branch assumes `i < 0 && i >= n`, but `range(n)` when `n < 0` produces an **empty** sequence in Python — no iterations occur. The correct assumption for the negative case should be `false` (the loop body is unreachable):

```lean
            let inRangeExpr := mkStmtExprMd $ .IfThenElse
              (mkStmtExprMd $ .PrimitiveOp .Geq [asIntRange, mkStmtExprMd $ .LiteralInt 0])
              (mkStmtExprMd $ .PrimitiveOp .And [
                  (mkStmtExprMd $ .PrimitiveOp .Geq [asIntTarget, mkStmtExprMd $ .LiteralInt 0]),
                  (mkStmtExprMd $ .PrimitiveOp .Lt [asIntTarget, asIntRange]) ])
              (mkStmtExprMd $ .LiteralBool false)  -- range(n) is empty when n < 0
```

Alternatively, since `range(n)` with `n < 0` is equivalent to `range(0)`, you could simplify to just `i >= 0 && i < n` unconditionally — when `n <= 0`, this is unsatisfiable, which correctly models the empty range.

StrataTest/Languages/Python/tests/test_for_range.py
```diff
@@ -0,0 +1,14 @@
...
+    assert i >= -10
+    assert i < 0
+    j = numbers[i+5]
+    assert j < 10
```
tautschnig: This test passes but for the wrong reason: `range(-10)` produces an empty sequence in Python, so the loop body never executes and the assertions are vacuously true. The test doesn't actually validate the negative-range assumptions.

Consider replacing with a test that demonstrates the empty-range behavior:
```python
# range with negative arg is empty — loop body should not affect result
x = 42
for i in range(-10):
    x = 0  # should never execute
assert x == 42, "range(-10) should be empty"
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1440,14 +1440,30 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-    | _ => []
-    let bodyStmts := assumeInStmts ++ bodyStmts
+    let assumeStmts : List StmtExprMd := match target with
+      | .Name _ n _ =>
+        let targetVar := mkStmtExprMd (StmtExpr.Identifier n.val)
+        match iterExpr.val with
+          | .StaticCall "range" [intExpr] =>
+            let assumeTypeInt := mkStmtExprMd (.Assume $ mkStmtExprMd (.StaticCall "Any..isfrom_int" [targetVar]))
+            let asIntRange := mkStmtExprMd $ .StaticCall "Any..as_int!" [intExpr]
+            let asIntTarget := mkStmtExprMd $ .StaticCall "Any..as_int!" [targetVar]
```
MikaelMayer, line 1459: Range supports booleans as well since booleans can be converted to ingegers.
Also this makes me feel uneasy about hard-coding everything here. Range is a function that can take 1 2 or 3 arguments. It might be nicer if we actually model that function accordingly and define axioms about PIn(x, range(from_int(a), None, None)) instead so that this code does not change and we can just maintain the prelude axioms until we have a proper executable interpretation of ranges.

thanhnguyen-aws: I added a check for that. The `range` function in Prelude now has 3 inputs, but the last 2 are only allowed to be `None`

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1440,14 +1440,33 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-      [assumeInStmt]
-    | _ => []
-    let bodyStmts := assumeInStmts ++ bodyStmts
+    let assumeStmts : List StmtExprMd ← do match target with
+      | .Name _ n _ =>
+        let targetVar := mkStmtExprMd (StmtExpr.Identifier n.val)
+        let isNone (s: StmtExprMd) := match s.val with
+          | .StaticCall "from_None" _ => true | _ => false
+        match iterExpr.val with
+          | .StaticCall "range" (startExpr::stopExpr::stepExpr::_) =>
```
tautschnig, line 1450: nit: The `isNone` helper checks for `StaticCall "from_None"` by pattern-matching on the `StmtExpr` value. This is fragile — if the `None` representation changes (e.g., to a constant), this check would silently stop working. Consider using a named constant for the `from_None` string.

tautschnig (dismissed):

MikaelMayer (approved):

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1440,14 +1441,33 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+      | .Name _ n _ =>
+        let targetVar := mkStmtExprMd (StmtExpr.Identifier n.val)
+        let isAnyNone (s: StmtExprMd) := match s.val with
+          | .StaticCall constructor _ => constructor == AnyConstructor.None | _ => false
+        match iterExpr.val with
+          | .StaticCall "range" (startExpr::stopExpr::stepExpr::_) =>
+            if ¬ (isAnyNone stopExpr && isAnyNone stepExpr) then
+              throw (.unsupportedConstruct "Unsupport range function with more than 1 input" (toString (repr iter)))
+            let asIntStart := mkStmtExprMd $ .StaticCall "Any..as_int!" [startExpr]
+            let emptyRangeExit := mkStmtExprMd $ .IfThenElse
```
atomb, line 1454: I'd love to see some better abstractions to make code like this that generates lots of Laurel code more readable and less error-prone, though I suspect a more wide-scale refactoring in that direction, in a separate PR, is perhaps the right approach.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1440,14 +1441,33 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+              (mkStmtExprMd $ .PrimitiveOp .Leq [asIntStart, mkStmtExprMd $ .LiteralInt 0])
+              (mkStmtExprMd $ .Exit breakLabel)
+              none
+            let assumeTypeInt := mkStmtExprMd (.Assume $ mkStmtExprMd (.StaticCall "Any..isfrom_int" [targetVar]))
```
atomb, line 1458: I'd prefer for this to use a named constant rather than an inline string, to ensure that multiple references to this name are consistent, and to make it easier to update the code if the name changes.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1440,14 +1441,33 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-    let bodyStmts := assumeInStmts ++ bodyStmts
+    let assumeStmts : List StmtExprMd ← do match target with
+      | .Name _ n _ =>
+        let targetVar := mkStmtExprMd (StmtExpr.Identifier n.val)
+        let isAnyNone (s: StmtExprMd) := match s.val with
+          | .StaticCall constructor _ => constructor == AnyConstructor.None | _ => false
+        match iterExpr.val with
+          | .StaticCall "range" (startExpr::stopExpr::stepExpr::_) =>
+            if ¬ (isAnyNone stopExpr && isAnyNone stepExpr) then
+              throw (.unsupportedConstruct "Unsupport range function with more than 1 input" (toString (repr iter)))
```
atomb, line 1452:
```suggestion
              throw (.unsupportedConstruct "Unsupported range function with more than 1 input" (toString (repr iter)))
```

atomb (approved):
This seems correct, though I noted a few stylistic things that could be improved.

PR #819, opened 2026-04-09
Fix #762: Instance method calls and bodies in Python-to-Laurel translation

### LINKED ISSUE #762: Instance method calls and bodies not handled in Python-to-Laurel translation

MikaelMayer:
Fixes #762

Instance method calls were translated as holes (`<?>`) and method bodies were stripped to opaque in the Python-to-Laurel translation, preventing analysis of class-based Python code.

This PR:
- Resolves instance method calls (`obj.method(args)`) to `ClassName@method(obj, args)` when the method belongs to a user-defined class
- Preserves translated method bodies instead of stripping them, except for methods that assign to `self.field` (where field assignment verification is not yet supported)
- Defaults `LaurelResult` to `None` at the start of method bodies so methods without explicit return are well-defined
- Skips `self.field` annotations in variable hoisting since they are field assignments, not local variable declarations
- Inlines user-defined procedures (up to 3 rounds for transitive call chains) when dispatch modules are present

Bug-finding failure classification and prelude-name filtering are preserved from main.

All existing tests pass.


StrataMain.lean
```diff
@@ -551,7 +551,7 @@ private def printPyAnalyzeSummary (vcResults : Array Core.VCResult)
     match checkMode with
     | .bugFinding | .bugFindingAssumingCompleteSpec =>
       { isFailure := fun r => match r.outcome with
-          | .ok o => o.alwaysFalseAndReachable
+          | .ok o => o.satisfiabilityProperty == .unsat
```
MikaelMayer: This changes the semantics of bug-finding failure classification. The previous `alwaysFalseAndReachable` required both `satisfiabilityProperty == .unsat` AND `validityProperty == .sat _` (i.e. the property is always false *and* the path is reachable). The new check `satisfiabilityProperty == .unsat` alone also classifies as failures:
- `(unsat, unsat)` — unreachable / dead code
- `(unsat, unknown)` — always false if reached, reachability unknown

Is this broadening intentional? If so, a brief comment explaining the rationale would help future readers. If not, `o.alwaysFalseAndReachable` (or the equivalent `o.isRefutedAndReachable`) would preserve the original behavior.

StrataMain.lean
```diff
@@ -693,18 +693,20 @@ def pyAnalyzeLaurelCommand : Command where
...
-            IO.println "\n==== Core Program (after inlining) ===="
-            IO.print inlined
-          if let some dir := keepDir then
-            let path := s!"{dir}/{baseName}.inlined.core"
-            IO.FS.writeFile path (toString inlined)
-          pure inlined
+      if pyspecFiles.size > 0 || dispatchModules.size > 0 then
+        let pred := ⟨.some (fun name _ => name ≠ "__main__")⟩
+        let mut prog := coreProgram
+        for _ in [:3] do
```
MikaelMayer: Two changes here worth noting:
1. The `preludeNames` filter was removed — the old predicate was `name ≠ "__main__" && !preludeNames.contains name`, so prelude procedures are now eligible for inlining. Is that intended?
2. The magic number `3` for the inlining loop iterations could use a brief comment explaining why 3 rounds are needed (e.g. to handle transitive call chains of depth ≤ 3).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1154,10 +1180,20 @@ partial def collectDeclaredNamesAndTypes (ctx : TranslationContext) (stmts : Lis
       let names := (lhs.val.toList.filter (λ e => match e with |.Name _ _ _ => true | _=> false)).map pyExprToString
       names.map (λ n => (n, ty))
     | .AnnAssign _ lhs annoTy value _ =>
```
MikaelMayer, line 1186: The type-inference logic for `AnnAssign` is duplicated between the `Attribute` branch and the catch-all `_` branch (lines 1188–1191 and 1193–1196). Consider extracting it:
```
let ty := match value.val with
  | some value => (inferClassTypeFromLaurelExpr ctx value).getD $ pyExprToString annoTy
  | _ => pyExprToString annoTy
```
into a local `let` before the `match lhs`, and then the two branches just differ in whether they return `[]` or `[(pyExprToString lhs, ty)]`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -935,10 +935,27 @@ partial def translateCall (ctx : TranslationContext)
...
+              pure <| mkStmtExprMd (StmtExpr.FieldSelect objExpr fieldAttr.val)
+            | _ => translateExpr ctx val
+          | _ => translateExpr ctx val
         if opt_firstarg.isSome then
           if let some (ImportedSymbol.procedure _ _ true) := ctx.importedSymbols[funcName]? then
             return mkCall funcName
+          else if funcName ∈ ctx.userFunctions then
+            let parts := funcName.splitOn "@"
+            let className := parts.head!
+            let methName := (parts.drop 1).head!
```
MikaelMayer: `parts.head!` and `(parts.drop 1).head!` will panic if `funcName` doesn't contain `@`. Since `funcName ∈ ctx.userFunctions` and user functions are built as `className ++ "@" ++ methodName`, this is safe in practice, but a `match parts` with a fallback (or at least a comment noting the invariant) would be more defensive.

StrataTest/Languages/Python/TestExamples.lean
```diff
@@ -16,6 +16,33 @@ open Lean.Parser (InputContext)
 
 namespace Strata.Python
 
+/-- Run the Python → Ion → Laurel pipeline and return the Laurel program.
+    The caller can inspect the Laurel IR directly or continue to Core/SMT. -/
+def processPythonToLaurel (pythonCmd : System.FilePath) (input : InputContext)
```
MikaelMayer, line 48: The Python→Ion→Laurel pipeline (temp dir setup, py_to_strata invocation, `pyAnalyzeLaurel` call) is now duplicated between `processPythonToLaurel` and `processPythonFile`. Consider having `processPythonFile` call `processPythonToLaurel` for the first half, then continue with the Core/verify steps. That would eliminate ~15 lines of duplication.

MikaelMayer (commented):
Solid PR that addresses a real gap in the Python-to-Laurel translation. The approach of resolving instance method calls to `ClassName@method(self, args)` and selectively keeping method bodies transparent is well-designed. The `hasSelfFieldAssignment` guard is a pragmatic way to avoid unsupported field assignment verification while still enabling analysis of pure methods.

A few items worth discussing below — one semantic concern in `StrataMain.lean` and some minor code quality observations.

StrataMain.lean
MikaelMayer: 🤖 Reverted to `o.alwaysFalseAndReachable` — the broadening was unintentional. The original semantics (requiring both `satisfiabilityProperty == .unsat` AND `validityProperty == .sat _`) is correct for bug-finding failure classification.

StrataMain.lean
MikaelMayer: 🤖 Restored the `preludeNames` filter — removing it was unintentional. Also added a comment explaining the 3 inlining rounds (transitive call chains of depth ≤ 3).

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Extracted the type-inference `let ty` before the `match lhs`, so the two branches now only differ in returning `[]` vs `[(pyExprToString lhs, ty)]`.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Added a comment documenting the invariant: `userFunctions` entries are always `"ClassName@methodName"`, so the split is safe.

StrataTest/Languages/Python/TestExamples.lean
MikaelMayer: 🤖 Extracted the shared Python→Ion→Laurel pipeline into a private `runPythonToLaurelPipeline` helper. Both `processPythonToLaurel` and `processPythonFile` now delegate to it, eliminating the duplication.

MikaelMayer:
Duplicate of #815.

PR #823, opened 2026-04-09
Add synthetic Python tests covering many Python constructs

olivier-aws:
This PR adds 326 Python tests that cover many Python constructs. Each test is simple and follows the same structure:
```python
def some_function():
   # do some computation
   assert(...)       # This assert must be proven valid

some_function()
```
The tests that can be analyzed with Strata and for which the assert is proven valid (99 tests) are in the `tests/` directory, with an `expected` file. The tests that do not yet pass with Strata (either failing with internal errors or the assert cannot be proven) are in the `pending/` directory, and are not launched by the CI test framework.


tautschnig (changes requested):
This is great, but I'd like to implement better CI integration so that we also check the remaining pending changes. This can done in a much better way with https://github.com/strata-org/Strata/pull/829 merged first, so putting in this comment as a temporary blocker - the actual changes are implemented in this branch already, but we'd have weird git history.

tautschnig (approved):

atomb (approved):
I haven't reviewed all the individual tests, but I like the overall approach. As simple as these are, they'll be valuable.

PR #825, opened 2026-04-10
Fix extra parentheses in constructor and binding formatting

ChengZ3:
*Description of changes:*
- Remove extra parentheses around datatype constructors in formatted output.
- Remove extra parentheses around binding types.

Before:
```
datatype List (a : Type) {(
  (Nil())),
  (Cons(head : a, tail : (List a)))
};
```
After:
```
datatype List (a : Type) {
  Nil(),
  Cons(head : a, tail : List a)
};
```

Problem: 
In `SyntaxDefAtom.formatArgs`, the formatter parenthesizes an argument when `prec > 0 ∧ innerPrec ≤ prec`, where `prec` is the precedence threshold on the ident position and `innerPrec` is the precedence of the argument filling that position. The grammar rules `constructorListAtom`, `constructorListPush`, `mkBinding`, and `casesBinding` had no explicit precedence on their ident positions, so they inherited high defaults:

- In the constructor list rules, ident positions got `prec = maxPrec` (1024). The `c : Constructor` idents in `constructorListAtom` and `constructorListPush` wrap each constructor in parens because `constructor_mk` outputs prec 50 and `50 ≤ 1024`. The `cl : ConstructorList` ident in `constructorListPush` adds another layer of parens because `constructorListAtom` outputs prec `maxPrec` and `maxPrec ≤ maxPrec`.
- Binding type idents got `prec = 40`. Type applications like `List a` are formatted with `appPrec` (20). `20 ≤ 40`, which adds unnecessary parens around bindings.

Fix: 
- Remove `@[prec(50)]` from `constructor_mk`.
- Add `:0` precedence (never parenthesize) to `constructorListAtom` and `constructorListPush` idents.
- Add `:0` to binding type positions `mkBinding`, `casesBinding`.


shigoel (approved):
Thank you so much, @ChengZ3! Much appreciated.

tautschnig (approved):

PR #826, opened 2026-04-10
Add per-node source ranges to SpecExpr using Option SourceRange

tautschnig:
Addressing the follow-up from #810, specifically https://github.com/strata-org/Strata/pull/810#discussion_r3058940210: Add per-node source ranges to SpecExpr so each expression node can carry its own source location from the Python AST, and use those locations when converting to DDM and Laurel metadata. Uses `Option SourceRange` (with default `none`) rather than a custom type.

Changes:
- Decls.lean: Replace `loc : SourceRange` with `loc : Option SourceRange := none` in all 17 SpecExpr constructors. `SpecType.loc` remains `SourceRange` (unaffected).
- DDM.lean: `toDDM` extracts via `loc.getD default`; `fromDDM` wraps via `some loc`.
- Specs.lean: `assumeCondition` takes `Option SourceRange`. Construction sites pass `some e.ann`.
- ToLaurel.lean: `nodeMd` pattern-matches on `none`/`some r`, falling back to parent metadata for `none` or empty ranges. `reportError` calls use `loc.getD default`.


atomb (dismissed):
Thanks for doing this!

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -307,6 +307,31 @@ def count (ad : ArgDecls) := ad.args.size + ad.kwonly.size
...
 end ArgDecls
 
+/-- A source location that either carries a real range or an explanation of
+    why no range is available. This makes it impossible to create an empty
+    location without documenting the reason. -/
+inductive SourceLoc where
+  /-- A concrete source range from the Python AST. -/
+  | range (r : SourceRange)
+  /-- No source range available; `reason` explains why. -/
+  | missing (reason : String)
```
keyboardDrummer: Is the `missing` constructor used in this PR ? If you're adding it then it would be good to use it to indicate what valid use-cases are.

I don't see valid use-cases because AFAIK the reason is always the same: that compilation generated an AST node that is guaranteed to be correct so no diagnostics will ever be emitted for it, so the empty source location will never be emitted.

We should prove at some point that Strata can never emit diagnostics with empty source locations

Secondly, the name `SourceLoc` is confusing to me. It sounds like a single position in the source, but really it's an optional source range.

keyboardDrummer (changes requested):

Strata/Languages/Python/Specs/Decls.lean
tautschnig: Thank you, rearchitected in c47491584a44a to use `Option SourceRange` instead of the newly introduced type.

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -316,39 +317,39 @@ and `placeholder`; interior nodes represent operations like `len`, `getIndex`,
...
     The original Python expression is preserved in `Assertion.message`. -/
-| placeholder (loc : SourceRange)
-| var (name : String) (loc : SourceRange)
-| getIndex (subject : SpecExpr) (field : String) (loc : SourceRange)
-| isInstanceOf (subject : SpecExpr) (typeName : String) (loc : SourceRange)
-| len (subject : SpecExpr) (loc : SourceRange)
-| intLit (value : Int) (loc : SourceRange)
-| intGe (subject : SpecExpr) (bound : SpecExpr) (loc : SourceRange)
-| intLe (subject : SpecExpr) (bound : SpecExpr) (loc : SourceRange)
+| placeholder (loc : Option SourceRange := none)
```
joehendrix, line 320: Could we go back to using `SourceRange` with `SourceRange.none` rather than `Option SourceRange`?   What is `None` being used for here?  If I understand correctly, this will just adds an extra allocation per `SpecExpr` expression.

joehendrix (changes requested):
I think the PR description is misleading.  SpexExpr already has per node source range information.

This is changing SpecExpr from using `SourceRange` rather than `Option SourceRange`.  I can speculate on the motivation, but I'd prefer the intention of the change to be spelled out so I can assess whether this is the right solution.

tautschnig:
> This is changing SpecExpr from using `SourceRange` rather than `Option SourceRange`. I can speculate on the motivation, but I'd prefer the intention of the change to be spelled out so I can assess whether this is the right solution.

My apologies, that information got lost in an earlier edit of the PR description (and is now restored): this is trying to address the follow-up from #810, specifically https://github.com/strata-org/Strata/pull/810#discussion_r3058940210.

PR #829, opened 2026-04-10
Share Lean build cache with Python CI jobs

tautschnig:
The Python matrix job (3.11-3.14) previously rebuilt the Lean strata executable from scratch in each of the 4 jobs. Since Lean build artifacts are Python-version-independent, this was redundant.

The Python job now depends on build_and_test_lean and restores the Lean build cache (saved with the SHA-keyed cache key) instead of rebuilding. lean-action with build: false installs the toolchain (needed for lake env) without compiling anything.

This eliminates 4 redundant Lean builds from the CI pipeline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,6 +196,7 @@ jobs:
           fail: true
 
   build_python:
+    needs: build_and_test_lean
     name: Build and test Python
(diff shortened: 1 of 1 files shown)


atomb (approved):
Thanks for doing this!

aqjune-aws (approved):

PR #835, opened 2026-04-10
Laurel: Lift procedure calls in asserts

ssomayyajula:
*Description of changes:*

Always transform asserts in LiftImperativeExpressions to hoist imperative procedure calls out of expressions. Previously, asserts containing imperative calls were skipped, leaving procedure calls inside pure expressions that Core rejects with 'calls to procedures are not supported in functions or contracts'.

Added test which produces an internal assert of the form `!Any..isexception(PSub(base, timedelta_func(...)))`.


Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -357,15 +357,11 @@ def transformStmt (stmt : StmtExprMd) : LiftM (List StmtExprMd) := do
...
-      -- Do not transform assert conditions with assignments — they must be rejected.
-      -- But nondeterministic holes need to be lifted.
-      if containsNondetHole cond && !containsAssignmentOrImperativeCall (← get).model cond then
-        let seqCond ← transformExpr cond
-        let prepends ← takePrepends
-        modify fun s => { s with subst := [] }
-        return prepends ++ [⟨.Assert seqCond, md⟩]
-      else
-        return [stmt]
+      -- Transform assert conditions to lift imperative calls and nondeterministic holes
```
keyboardDrummer: Lifting assignments and lifting procedure calls should not follow the same rules. Please prevent lifting assignments in assertions.

olivier-aws: @keyboardDrummer  Can I ask why? Do you think assignment should not be lifted from assertions because we should forbid to have them in the source language?

ssomayyajula: Made the change, but I also want to know the theory behind this

keyboardDrummer: We want assertions to be erasable at runtime so they must not have side-effects. If Laurel would allow assignments in assertions that would create a soundness issue, since the front-ends, when they optimize the sources for runtime performance, will remove the assertions and the assignments they contain.

You could also argue that the code that erases the assertion should do the check, but then that code could would need to be more complicated and still I think allowing contracts (like assertions) to have side-effects is asking for trouble.

keyboardDrummer (dismissed):

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -158,6 +158,22 @@ private def computeType (expr : StmtExprMd) : LiftM HighTypeMd := do
   return computeExprType s.model expr
 
 /-- Check if an expression contains any assignments or imperative calls (recursively). -/
```
tautschnig: Stale docstring: this function only checks for assignments, not imperative calls. The docstring was copied from `containsAssignmentOrImperativeCall`.

```suggestion
/-- Check if an expression contains any assignments (recursively). -/
```

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -158,6 +158,22 @@ private def computeType (expr : StmtExprMd) : LiftM HighTypeMd := do
...
+  | .StaticCall _ args => args.attach.any (fun x => containsAssignment x.val)
+  | .PrimitiveOp _ args => args.attach.any (fun x => containsAssignment x.val)
+  | .Block stmts _ => stmts.attach.any (fun x => containsAssignment x.val)
+  | .IfThenElse cond th el =>
+      containsAssignment cond || containsAssignment th ||
+      match el with | some e => containsAssignment e | none => false
+  | _ => false
+  termination_by expr
+  decreasing_by
+    all_goals ((try cases x); simp_all; try term_by_mem)
```
tautschnig, line 175: Refactoring suggestion (non-blocking): `containsAssignment` duplicates the recursive traversal structure of `containsAssignmentOrImperativeCall`, differing only in the `StaticCall` case. Consider refactoring `containsAssignmentOrImperativeCall` to build on `containsAssignment`:

```lean
def containsAssignmentOrImperativeCall (model: SemanticModel) (expr : StmtExprMd) : Bool :=
  containsAssignment expr || containsImperativeCall model expr
```

where `containsImperativeCall` handles only the non-functional `StaticCall` check. This would eliminate the duplicated traversal and make the relationship between the two predicates explicit.

Alternatively, a single parameterized traversal `containsImpure (checkCalls : Bool) (model : SemanticModel)` could serve both use sites.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -358,8 +374,8 @@ def transformStmt (stmt : StmtExprMd) : LiftM (List StmtExprMd) := do
   match val with
   | .Assert cond =>
       -- Do not transform assert conditions with assignments — they must be rejected.
-      -- But nondeterministic holes need to be lifted.
-      if containsNondetHole cond.condition && !containsAssignmentOrImperativeCall (← get).model cond.condition then
+      -- But nondeterministic holes and imperative calls need to be lifted.
+      if !containsAssignment cond.condition then
```
tautschnig, line 377: Two observations on the `Assert` case:

**1. Asymmetry with `Assume` (line 387):** The `Assert` and `Assume` cases now have divergent guard logic:
- `Assert`: `!containsAssignment` — lifts everything except assignments
- `Assume`: `containsNondetHole && !containsAssignmentOrImperativeCall` — only lifts nondet holes

Is this intentional? If an `assume` contains an imperative call (but no assignment), it won't be lifted and will presumably hit the same "calls to procedures are not supported" error downstream. If the asymmetry is deliberate, a comment explaining why would help future readers.

**2. Missing Laurel-level unit test:** The Python end-to-end test covers the motivating scenario, but there's no Laurel-level unit test for the new lifting behavior. The existing `LiftExpressionAssignmentsTest.lean` shows the pattern — a test like the following would directly exercise the new logic and guard against regressions:

```lean
def assertWithImperativeCall : String := r"
procedure impure(): int {
  var x: int := 0;
  x := x + 1;
  x
};
procedure test() {
  assert impure() == 1
};
"
```

This would verify that `assert impure() == 1` is transformed into something like `var $tmp := impure(); assert $tmp == 1` at the Laurel level, independent of the Python frontend.

Also consider a negative test confirming that `assert (x := 2) == 2` is still *not* lifted (passes through unchanged so the downstream error is preserved).

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressionsError.lean
```diff
@@ -41,7 +41,6 @@ procedure impureContractIsNotLegal1(x: int)
 //              ^^^^^^^^ error: calls to procedures are not supported in functions or contracts
 {
   assert impure() == 1
```
tautschnig, line 46: This is the key semantic change: `assert impure() == 1` in a procedure body is no longer an error because the imperative call gets lifted before Core translation. The change is correct given the new lifting logic.

Note that `requires x == impure()` (line 40-41) still correctly errors — contracts are not processed by `transformStmt`, so the lifting doesn't apply there. Worth adding a brief comment in the test to explain why the body assert is now accepted while the contract is not:

```suggestion
  assert impure() == 1  -- No error: impure() is lifted by LiftImperativeExpressions
```

keyboardDrummer (dismissed):

Strata/Languages/Laurel/LaurelTypes.lean
```diff
@@ -46,7 +46,7 @@ def computeExprType (model : SemanticModel) (expr : StmtExprMd) : HighTypeMd :=
     | .datatypeConstructor t _ => ⟨ .UserDefined t, source ⟩
     | .parameter p => p.type
     | .staticProcedure proc => match proc.outputs with
-      | [singleOutput] => singleOutput.type
+      | firstOutput :: _ => firstOutput.type
```
keyboardDrummer: This is incorrect code and I would prefer not to add it since I think it will hide issues down the line. I think it might be better to merge https://github.com/strata-org/Strata/pull/1076 first, which changes things around the "could not infer type" error, and might fix the build issue this PR was having without requiring this change.

keyboardDrummer (changes requested):

Strata/Languages/Laurel/LaurelTypes.lean
keyboardDrummer: Let me clarify the previous comment. When doing a call to a multiple output procedure, the resulting value is all the outputs, not the just the first one, so saying the type of the call is the type of the first output parameter, is incorrect.

The correct implementation of this line is to introduce a multi-value type that can hold multiple types. This is done in this PR that's up for review: https://github.com/strata-org/Strata/pull/1077/changes#diff-c94b4c5c1a51062a774ba1548faa0e7a793152b74dd4f900423fb9b56bc9b683R167

tautschnig: Yes, I figured we'd need #1077. So we should try to get #1077 merged first?

tautschnig: Addressed in the merge [371571d](https://github.com/strata-org/Strata/pull/835/commits/371571d2f5e3323c30a5df8a661181acbc2ac363), which now has #1077.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -421,8 +429,21 @@ def translateStmt (stmt : StmtExprMd)
       let translateCallTargets (calleeName : String) (args : List StmtExprMd) : TranslateM (List Core.Statement) := do
         let coreArgs ← args.mapM (fun a => translateExpr a)
         let (inits, lhs) ← initTargetsNondet
-        let outArgs : List (Core.CallArg Core.Expression) := lhs.map .outArg
-        return inits ++ [Core.Statement.call calleeName (coreArgs.map .inArg ++ outArgs) md]
+        -- Pad LHS with unused variables for extra outputs (e.g., $heap)
```
keyboardDrummer: @keyboardDrummer-bot can you investigate why this is necessary? Why is there a call with LHS targets that don't match? I would have expected that HeapParameterization, if it turns a dangling call into an assignment to heap, already adds any missing LHS targets.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -101,6 +101,14 @@ def translateType (ty : HighTypeMd) : TranslateM LMonoTy := do
       return .tcons "Composite" []
   | .TCore s => return .tcons s []
   | .TReal => return LMonoTy.real
+  | .MultiValuedExpr _ =>
```
keyboardDrummer: For what case is this necessary? I would've thought an `invalidCoreType` would be fine here.

keyboardDrummer:
@keyboardDrummer-bot Please update LiftImperativeExpressions so lifted calls get the correct number of targets, and then remove the code changes in LaurelToCoreTranslator.

keyboardDrummer:
@keyboardDrummer-bot show me the Laurel that's produced for the test `test_procedure_in_assert`

tautschnig:
> Note that the internal assert is checking `!Any..isexception(PSub(base, $c_1))` — it's using `$c_1` (the `Error` output of `timedelta_func`) as the second argument to `PSub`, rather than `$c_0` (the `delta: Any` output). This looks like a bug — the exception check should be `PSub(base, timedelta_func(...))` where the result used is the primary return value, not the error output.

@keyboardDrummer-bot This might have been addressed in [001e735](https://github.com/strata-org/Strata/pull/835/commits/001e735ae6f4b53316a820493d8ba85461cb76f9), can you please confirm?

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -274,15 +277,31 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
+      -- Fallback: if no outputs found, use a single target with the computed type
+      if targets.isEmpty then
+        let v ← freshCondVar
+        let ty ← computeType expr
+        declarations := [⟨.Var (.Declare ⟨v, ty⟩), source⟩]
+        targets := [⟨.Local v, source⟩]
+      let liftedCall := declarations ++ [⟨.Assign targets seqCall, source⟩]
       modify fun s => { s with prependedStmts := s.prependedStmts ++ liftedCall}
-      return bare (.Var (.Local callResultVar))
+      -- The first output is the procedure's primary return value
```
keyboardDrummer: Neither using the first or the last output is correct. I think this is a workaround for an issue that occurs elsewhere. It seems to be caused on the Python side and Laurel is incorrectly not reporting an error on that.

keyboardDrummer:
@keyboardDrummer-bot the initial Laurel code:

```
procedure main()
  returns (LaurelResult: (Core Any))
  opaque
  ensures Any..isfrom_int(LaurelResult)( summary "(main ensures) Return type constraint")
  modifies *
{ var nullcall_ret: (Core Any) := from_None();
  LaurelResult := from_None();
  var maybe_except: (Core Error) := NoError();
  var base: (Core Any) := <?>;
  var delta: (Core Any) := <?>;
  var result: (Core Any) := <?>;
  base := from_int(100);
  assert Any..isfrom_int(base);
  assert !Any..isexception(PSub(base, timedelta_func(from_int(7), from_None()))) summary "Check PSub exception";
  delta := PSub(base, timedelta_func(from_int(7), from_None()));
  result := from_int(1);
  assert Any..isfrom_int(result);
  assert Any_to_bool(PEq(result, from_int(1))) summary "should pass";
  LaurelResult := result;
  exit $body };
```

looks incorrect because `timedelta_func` returns multiple arguments yet is called without being assigned to a multi-assign. Can you determine why Resolution.lean does not emit a diagnostic?

keyboardDrummer (approved):
Looks good! Thanks

tautschnig (approved):

PR #836, opened 2026-04-10
Fix fvar index formatting in Boole programs by threading GlobalContext through formatProgram

kondylidou:
Description of changes:

Add `Boole.formatProgram (prog : Boole.Program) (gctx : GlobalContext) (dialects : DialectMap) : Std.Format` to fix `fvar` indices printing as `fvar!N` when formatting a Boole program.

When `prog.toAst` is used to reconstruct a `Strata.Program`, the single `Boole.prog` container op has no binding specs, so `Program.globalContext` ends up empty and `fvar` indices can't be resolved. The fix mirrors `Core.formatProgram`: accept the `GlobalContext` from the originating `Strata.Program` explicitly rather than recomputing it.

Includes a regression test in `StrataTest/Languages/Boole/format_program.lean`.


atomb (dismissed):

Strata/Languages/Boole/Verify.lean
```diff
@@ -717,6 +717,21 @@ def toCoreDecls (cmd : BooleDDM.Command SourceRange) : TranslateM (List Core.Dec
...
+This mirrors `Core.formatProgram`: both functions accept an external context rather than
+recomputing one from the program structure, because the container operation (`Boole.prog`)
+carries no binding specs and therefore produces an empty `GlobalContext` when processed alone.
+-/
+def formatProgram (prog : Boole.Program) (gctx : GlobalContext) (dialects : DialectMap) : Std.Format :=
+  let ctx := FormatContext.ofDialects dialects gctx {}
+  let state : FormatState := {
+    openDialects := dialects.toList.foldl (init := {}) fun a d => a.insert d.name
+  }
+  (mformat (ArgF.op prog.toAst) ctx state).format
```
tautschnig, line 759: Nit (non-blocking): This function currently has no production call-site — it's only exercised by the test. Consider wiring it into the `Boole.verify` pipeline's debug output (similar to how `Core.formatProgram` is used in `Core.Verifier.lean` for `[DEBUG] Evaluated program:` messages), or adding a `ToString`/`ToFormat` instance for `Boole.Program` (analogous to `Core.lean:91-95`). That way the fix benefits users who hit the original `fvar!N` issue in practice, not just in the test.

StrataTest/Languages/Boole/format_program.lean
```diff
@@ -0,0 +1,56 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.MetaVerifier
```
tautschnig: Nit (non-blocking): `Strata.MetaVerifier` transitively pulls in the SMT solver interface and the full Core verifier. Since this test only needs `Boole.formatProgram`, `Boole.getProgram`, and the `#strata` elaborator, a more targeted import like `import Strata.Languages.Boole.Verify` (plus `import Strata.DDM.Elab` if `#strata` isn't available from that path) would keep the dependency footprint smaller and the test faster to compile in isolation.

StrataTest/Languages/Boole/format_program.lean
```diff
@@ -0,0 +1,56 @@
...
+  let formatted := (Boole.formatProgram prog vec_program.globalContext vec_program.dialects).pretty
+  !formatted.contains "fvar!"
+
+-- The names that appear in types and function bodies must match the declared names.
+#guard
+  let prog := (Boole.getProgram vec_program).toOption.get!
+  let formatted := (Boole.formatProgram prog vec_program.globalContext vec_program.dialects).pretty
+  formatted.contains "Vec" &&
+  formatted.contains "Vec..Vec_len" &&
+  formatted.contains "Vec..Vec_data"
```
tautschnig, line 54: The two `#guard` blocks duplicate the `getProgram` + `formatProgram` + `.pretty` pipeline. Consider extracting a local `let` binding at the top (or a single `#guard` that checks both properties) to reduce repetition:

```suggestion
#guard
  let prog := (Boole.getProgram vec_program).toOption.get!
  let formatted := (Boole.formatProgram prog vec_program.globalContext vec_program.dialects).pretty
  -- No fvar!N tokens: every free variable reference resolves via the GlobalContext.
  !formatted.contains "fvar!" &&
  -- Declared names appear in the formatted output.
  formatted.contains "Vec" &&
  formatted.contains "Vec..Vec_len" &&
  formatted.contains "Vec..Vec_data"
```

On the other hand, keeping them separate gives slightly better error localization if one fails. Up to you — just flagging the duplication.

Strata/Languages/Boole/Verify.lean
```diff
@@ -761,6 +776,8 @@ def verify
   | .error e =>
     throw <| IO.Error.userError (toString (e.format (some ictx.fileMap)))
   | .ok prog =>
+    if options.verbose >= .normal then
+      dbg_trace f!"\n\n[DEBUG] Boole program:\n{Boole.formatProgram prog env.globalContext env.dialects}"
```
tautschnig, line 806: This `dbg_trace` fires at `verbose >= .normal`, which is the default verbosity. This is consistent with how `Core.verify` handles its `[DEBUG] Evaluated program:` output, so the pattern is fine.

However, the `stack_array_based.lean` test was the only test updated to include the new debug output. I checked all other Boole tests — they all use `options := .quiet` (some on continuation lines), so they won't be affected. Just confirming this is safe.

tautschnig (approved):

shigoel (dismissed):

tautschnig (dismissed):

kondylidou:
@shigoel  fixed the conflicts here too! thanks for the fast reviewing:)

tautschnig (approved):

kondylidou:
@tautschnig thank you too:)

joscoh (approved):

PR #839, opened 2026-04-10
Add support for list negative and out of bound slice

thanhnguyen-aws:
*Description of changes:* This PR adds support for list negative and out of bound slice.

Semantic: For `list[start:stop]` , when `start` or `stop` is negative, Python firstly add `len(list)` to them, then clamp them to the range `[0, len(list)]`. If `start>=stop`, it returns an empty list. 

There should be no precondition for `start` and `stop`.

### DIFF
--- a/Strata/Languages/Python/PythonRuntimeLaurelPart.lean
+++ b/Strata/Languages/Python/PythonRuntimeLaurelPart.lean
@@ -382,10 +382,30 @@ procedure List_drop_len(l : ListAny, i: int)
   invokeOn List_len(List_drop(l,i))
   ensures i >= 0 && i <= List_len(l) ==> List_len(List_drop(l,i)) == List_len(l) - i;
 
-function List_slice (l : ListAny, start : int, stop: int) : ListAny
-  requires start >= 0 && start < List_len(l) && stop >= 0 && stop <= List_len(l) && start <= stop
+function int_max (i1: int, i2: int) : int
 {
-  List_take (List_drop (l, start), stop - start)
+  if i1 >= i2 then i1 else i2
+};
+
+function int_min (i1: int, i2: int) : int
+{
+  if i1 <= i2 then i1 else i2
+};
+
+function List_slice_non_neg (l : ListAny, start : int, stop: int) : ListAny
+  requires start >= 0 && stop >= 0
+{
+  if (start >= List_len(l)) || (start >= stop) then ListAny_nil()
+  else List_take (List_drop (l, start), int_min(stop, List_len(l))  - start)
+};
+
+
+function List_slice (l : ListAny, start : int, stop: int) : ListAny
+{
+  List_slice_non_neg (l,
+    if start >= 0 then start else int_max (List_len(l) + start, 0),
+    if stop >= 0 then stop else int_max (List_len(l) + stop, 0)
+  )
 };
 
 function List_set_non_neg (l : ListAny, i : int, v: Any) : ListAny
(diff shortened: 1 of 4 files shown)


Strata/Languages/Laurel/Grammar/AbstractToConcreteTreeTranslator.lean
```diff
@@ -43,6 +43,7 @@ partial def highTypeValToArg : HighType → Arg
   | .TFloat64 => laurelOp "float64Type"
   | .TReal => laurelOp "realType"
   | .TString => laurelOp "stringType"
+  | .TBv n => laurelOp "bvType" #[.num sr n]
```
MikaelMayer: There is no mention of this bvType in the PR description. Is it the right PR, or is the PR description outdated?

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -382,10 +382,30 @@ procedure List_drop_len(l : ListAny, i: int)
...
+
+function int_min (i1: int, i2: int) : int
+{
+  if i1 <= i2 then i1 else i2
+};
+
+function List_slice_non_neg (l : ListAny, start : int, stop: int) : ListAny
+  requires start >= 0 && stop >= 0
+{
+  if (start >= List_len(l)) || (start >= stop) then ListAny_nil()
```
MikaelMayer, line 398: or `stop <= 0`
But more generally, this is an optimization, I'm wondering if we shouldn't just use the last else case rather than having these cases. It's perhaps better for execution, but for provability, it might make things wrong if we don't pay attention. It might be worth keeping these special cases for lemmas instead @shigoel  @atomb  that prove and dispatch these special cases. No action needed, just perhaps add a comment that these special cases might better suit a lemma rather than an implementation.

StrataTest/Languages/Python/tests/test_for_range.py
```diff
@@ -0,0 +1,13 @@
+numbers = [2,4,5,9,1]
+
+for i in range(5):
+    assert i < 5
+    assert i >= 0
+    j = numbers [i]
+    assert j < 10
+
+x = 15
+for i in range(-10):
```
MikaelMayer: Weird, I think this has been merged in another PR.

StrataTest/Languages/Python/tests/test_list_slice.py
```diff
@@ -7,3 +7,9 @@
 assert list_of_list_nums [2:][1][2] == 7
 
 assert numbers [:2] == [10, 20]
+
+assert numbers [-4:-1] == [20, 30, 40]
+
+assert numbers [-100:-1] == [10, 20, 30, 40]
```
MikaelMayer, line 13: Is there an expect file for this test?

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
thanhnguyen-aws: When `stop <= 0`, Python will add `len(l)` to `stop` first, if it is still negative then it returns an empty list. It is handled in the main function `List_slice`

StrataTest/Languages/Python/tests/test_list_slice.py
thanhnguyen-aws: Yes, I just modified both .py and .expected file.

aqjune-aws (approved):

MikaelMayer (approved):

PR #840, opened 2026-04-10
Let --keep-all-files emit all intermediate representations after each pass of Laurel and Core

aqjune-aws:
This patch makes `--keep-all-files` emit all intermediate representations after each pass of Laurel and Core. :)


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -699,51 +699,81 @@ abbrev TranslateResultWithLaurel := (Option Core.Program) × (List DiagnosticMod
 
 /--
 Translate Laurel Program to Core Program, also returning the lowered Laurel program.
+
+When `keepAllFilesPrefix` is provided, the program state after each named
+Laurel-to-Laurel pass is written to `{prefix}.{n}.{passName}.laurel.st`
+(numbered from 1).
 -/
-def translateWithLaurel (options: LaurelTranslateOptions) (program : Program): TranslateResultWithLaurel :=
+def translateWithLaurel (options: LaurelTranslateOptions) (program : Program)
```
joscoh, line 707: Does this mean that there is no pure Laurel -> Core pipeline anymore? Putting everything in the IO monad would make future reasoning about this pipeline much more difficult (not sure if Lean allows reasoning about the IO monad at all).

aqjune-aws: I think reasoning on the returned value of IO (`TranslateResultWithLaurel` part) is possible, albeit my experience in IO monad is limited to this example: https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Proving.20a.20simple.20equality.20in.20a.20code.20using.20IO.20monad/near/533285847 Probably the exact part to verify can be delineated out after proper experiments.

Strata/Languages/Core/Verifier.lean
```diff
@@ -935,29 +935,47 @@ def verifySingleEnv (pE : Program × Env) (options : VerifyOptions)
...
-      return current
-    match Transform.run program passes with
-    | .ok prog => .ok prog
-    | .error e => .error (DiagnosticModel.fromFormat f!"❌ Transform Error. {e}")
+        state := newState
+        step := step + 1
+        if let some pfx := keepAllFilesPrefix then
+          let path := s!"{pfx}.{step}.{pp.phase.name}.core.st"
+          IO.toEIO (fun e => DiagnosticModel.fromFormat f!"{e}")
+            (IO.FS.writeFile path (toString current))
```
MikaelMayer: The Laurel side appends a trailing newline (`(formatProgram p).pretty ++ "\n"`) but the Core side uses `toString current` which goes through `Core.formatProgram` → `.pretty` without a trailing newline. Might be worth adding `++ "\n"` here for consistency, so all emitted `.st` files end with a newline.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -699,51 +699,81 @@ abbrev TranslateResultWithLaurel := (Option Core.Program) × (List DiagnosticMod
...
+  let emit (name : String) (p : Program) : IO Unit :=
+    match keepAllFilesPrefix with
+    | some pfx => do
+      let n ← stepRef.modifyGet (fun n => (n + 1, n + 1))
+      IO.FS.writeFile s!"{pfx}.{n}.{name}.laurel.st"
+        ((formatProgram p).pretty ++ "\n")
+    | none => pure ()
+
+  -- Step 0: the input program before any passes
+  if let some pfx := keepAllFilesPrefix then
```
MikaelMayer: Step 0 duplicates the formatting and write logic that `emit` already encapsulates. One option: initialize `stepRef` to `0` instead of having `emit` start at 1, then call `emit "Initial" program` here. That would remove the need to check `keepAllFilesPrefix` a second time and keep the write format in one place.

StrataTest/Languages/Python/run_keep_all_files_test.sh
```diff
@@ -0,0 +1,58 @@
...
+TMPDIR=$(mktemp -d)
+trap 'rm -rf "$TMPDIR"' EXIT
+
+ION_FILE="$TMPDIR/test_arithmetic.python.st.ion"
+KEEP_DIR="$TMPDIR/keep-test-output"
+
+python3 -m strata.gen py_to_strata --dialect "$DIALECT" \
+    "$SCRIPT_DIR/tests/test_arithmetic.py" "$ION_FILE"
+"$STRATA" pyAnalyzeLaurel --keep-all-files "$KEEP_DIR" "$ION_FILE" > /dev/null
+
```
MikaelMayer: If the `strata` binary hasn't been built yet (e.g. a fresh CI checkout where the build step failed silently), this will produce a confusing "No such file" error. A quick guard like `[[ -x "$STRATA" ]] || { echo 'strata binary not found'; exit 1; }` would make failures easier to diagnose.

aqjune-aws:
Addressed comments :)

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
MikaelMayer: I would prefer to have translations be pure and the part that runs all of this can make use of IO at in between phases. It will also ensure profiling is not measuring debugging I/O calls.

aqjune-aws: I think the translation function becoming pure can make the measuring time harder because evaluation of pure function can happen anytime in theory. For example:
```
start_time()
let res := translateWithLaurel prog
end_time()
```
A part of `translateWithLaurel prog` can be moved before/after the timing functions if `translateWithLaurel` is inlined. But if `translateWithLaurel` is IO monad, the sequence of order will be preserved and hoisting will not happen.

In fact, Claude suggested collecting all intermediate programs during pipeline and then emitting them at last (as suggested in the comments), but I concerned this might increase memory usage of Strata by N times where N is number of program transformations. :/

aqjune-aws: Precisely speaking, even with the usage of `IO` monad, code-motion optimizations can still happen - since there is no data dependency between the output of `start_time()` and `translateWithLaurel`, there is no strict guarantee that compiler won't move code out of the time-measuring zone. It is about whether compiler will optimize more aggressively or not, but IO monad will look opaque from the compiler perspective.

I think the second issue (memory usage) can be more significant, if we collect intermediate programs.

aqjune-aws: Mikael and I discussed separately and concluded that since it is not touching the signature of individual passes like `InferHoleTypes` it is fine.

StrataMain.lean
```diff
@@ -558,17 +558,12 @@ def pyAnalyzeLaurelCommand : Command where
       IO.println "\n==== Laurel Program ===="
       IO.println f!"{combinedLaurel}"
 
-    if let some dir := keepDir then
-      let path := s!"{dir}/{baseName}.laurel"
-      IO.FS.writeFile path ((Laurel.formatProgram combinedLaurel).pretty ++ "\n")
+    let keepPrefix := keepDir.map (s!"{·}/{baseName}")
 
     let (coreProgramOption, laurelTranslateErrors, loweredLaurel) ←
```
atomb: `loweredLaurel` is bound but unused:
   ```lean
   let (coreProgramOption, laurelTranslateErrors, loweredLaurel) ←
   ```
   The old code used `loweredLaurel` to write the `.lowered.laurel` file, but that write was removed (the emit-per-pass approach replaces it). This should be `_loweredLaurel` or `_` to avoid a compiler warning and signal intent.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
aqjune-aws: I promised to write a short proof that validates my hypothesis that reasoning the return value of `IO` monad is possible. :) This is the one written by AI: https://gist.github.com/aqjune-aws/fbbd2a017f2141198219735dcfb68804

atomb: What would you think of introducing a `MonadFileLog` class, something like:

```
class [Monad m] MonadFileLog m where
  fileLog : String -> String -> m ()
```

And you could then write code that did something like:

```
do ...
...
fileLog "Resolve" ((formatProgram p).pretty ++ "\n")
...
```

And you could introduce a helper function for the formatting to make it prettier if needed.

Then you could implement two instances of `MonadFileLog`. One would be an instance for `IO` that would initialize a counter and a temporary directory at the start and then do what `emit` does. The other could be:

```
instance : MonadFileLog Identity where
  fileLog _ _ = return ()
```

Instantiating the function with the latter instance would give you a pure function.

aqjune-aws: It's an interesting idea - then we can switch this `translateWithLaurel` function to either a pure one or IO one by providing a proper instance we want to use.
Then would be the case that, for verification, we will want to target the pure version of `translateWithLaurel`? Ideally, if we can verify the IO version of `translateWithLaurel` then it will be the best. But it is true that there are lots of top-level functions that use IO monad, and there is a possibility that verifying these can be quite challenging for lots of reasons, and using a technique that can turn the functions into pure one easily and verifying the alternative can be an option.

StrataTest/Languages/Python/run_keep_all_files_test.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
keyboardDrummer: I think this test is fragile. When we change the passes it will need to be updated, and we'll only find out when CI fails because it's not part of `lake test`. I would prefer it if you either resolve both issues or leave out the test entirely. Since this PR is adding debugging features, I'm ok with the test missing. You could also modify this test so it checks that some files were generating, not checking how many or whether they have the correct names.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -699,51 +699,79 @@ abbrev TranslateResultWithLaurel := (Option Core.Program) × (List DiagnosticMod
...
-  -- dbg_trace (toString (Std.Format.pretty (Std.ToFormat.format program)))
-  -- dbg_trace "================================="
   let result := resolve program
   let resolutionErrors: List DiagnosticModel := if options.emitResolutionErrors then result.errors.toList else []
   let (program, model) := (result.program, result.model)
+  emit "Resolve" program
   let diamondErrors := validateDiamondFieldAccesses model program
 
   let (program, nonCompositeDiags) := filterNonCompositeModifies model program
+  emit "FilterNonCompositeModifies" program
```
keyboardDrummer, line 735: This is a bit repetitive (and the existing code was as well). It would be better to register a list of named passes and iterate through those, but we can do that in a separate PR.

StrataTest/Languages/Python/run_keep_all_files_test.sh
aqjune-aws: Hi Remy, I think this kind of test sometimes appears and can be quite helpful in the sense that it explicitly shows what kind of passes we have. In LLVM, we have phase-ordering tests that indeed have a list of transformations for each backend assembly language: https://github.com/llvm/llvm-project/blob/main/llvm/test/CodeGen/RISCV/O3-pipeline.ll Indeed it is fragile, but on the other hand it will help reviewers understand that the pipeline does/doesn't update.
Think about the case when an external contributor wants to do some significant change to Laurel, and you want to assure that the pipeline of Laurel doesn't change :)

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
aqjune-aws: Yes. Since each pass doesn't have the same Lean types (some only accepts program, some accepts extra parameters) we will want to have an array whose each element has 'unified' type. I have a plan to extend the profiler to print running time of each Laurel pass, and the refactoring can be a part of it.

StrataTest/Languages/Python/run_keep_all_files_test.sh
keyboardDrummer: > Think about the case when an external contributor wants to do some significant change to Laurel, and you want to assure that the pipeline of Laurel doesn't change :)

I don't want to assure that. The phases are an implementation detail. I don't think it should be part of the tests.

keyboardDrummer (approved):

MikaelMayer (approved):

PR #848, opened 2026-04-11
Add `--entry-point` to pyAnalyzeLaurel, filter out procedures that are called by other procedures

aqjune-aws:
Strata's pyAnalyzeLaurel with bugFinding mode was checking executions of procedures that are already invoked by the main function, but this might not be the exact procedure targets we want to verify. For example:
```
def f(x):
  assert(x != 1);

f(1); # So far, bugFinding was generating two alarms for the assert:
      # 1. one for the specific x=1 case, reachable from the top-level scope of this Python file
      # 2. another one for the generic case where `x` could be anything inside f(x) <- unnecessary in bugFinding
```

For more precise control, this pull request adds the `--entry-point <main/roots/all>` option.

* If it is `main`, it only targets the main Python function as the target procedure to verify, and drops all other procedures after inlining from the main. Hence, the `assert` in the example above is checked only once.
* If it is `all`, all procedures are checked. Therefore, the `assert` above is checked twice - once from main (which will fail because it is always false), once from `f` (which will succeed because when `x` is 0 the assert passes).
* If it is `roots`, all procedures that (1) have no caller from the call graph, OR (2) the 'root' procedure from the strongly-connected component from the callgraph (if mutually recursive) will be targetted. In the above example, `assert` will be checked only once because `f` is called by the main function.

The `roots` mode is the default.
The definition of a root procedure from call graph is as follows:
* If there is a user-defined procedure in the SCC, then it will become the root.
* If there are multiple user-defined procedure in the SCC, the one with smallest lexicographical order will be chosen.
* If there is no user-defined procedure in the SCC, the SCC will simply be ignored.

To implement this behavior concretely,

* Procedure inlining is updated to only target the procedures of interest, and
* The partial evaluator (VCGen) only traverses on the target procedures.

This also updates some `PrecondElim` pass related codes to use the `Factory` that is registered at `CoreTransformState`.


Strata/Transform/ProcedureInlining.lean
```diff
@@ -197,15 +197,15 @@ use the specification. This will have to change if Strata also wants to support
 the reachability query.
 -/
 def inlineCallCmd
-    (doInline:String -> CachedAnalyses -> Bool := λ _ _ => true)
+    (doInline:Option String -> String -> CachedAnalyses -> Bool := λ _ _ _ => true)
```
tautschnig: Nit: The parameter name in the signature is positional — consider naming it for readability:

```suggestion
    (doInline: Option String -> String -> CachedAnalyses -> Bool := λ _caller _callee _analyses => true)
```

This makes it clearer at the definition site what each argument represents, matching the doc comment on `InlineTransformOptions.doInline`.

Strata/Transform/ProcedureInlining.lean
```diff
@@ -284,16 +284,21 @@ end ProcedureInlining
...
 Options to control the behavior of inlining procedure calls in a Core program.
 -/
 structure InlineTransformOptions where
-  doInline : String → Transform.CachedAnalyses → Bool := doInlineNonRecursive
+  -- 'doInline caller callee cachedAnalysis' returns true if the call command
+  -- from caller to callee should be inlined. The caller can be none if the
+  -- command is an orphaned one (rare, but can happen if inlineCallCmd is run
+  -- directly on Command).
+  doInline : Option String → String → Transform.CachedAnalyses → Bool :=
+      fun _ callee analyses => doInlineNonRecursive callee analyses
```
tautschnig, line 301: The doc comment on `InlineTransformOptions.doInline` is good. Consider also updating the doc comment on `inlineCallCmd` (around line 193-197) to mention the new `caller` parameter, since that's the public-facing function that external transforms call directly.

StrataMain.lean
```diff
@@ -600,16 +594,31 @@ def pyAnalyzeLaurelCommand : Command where
...
+        userProcNames
+    -- For bug-finding mode, inline procedures that are reachable from the
+    -- top-level scope of Python ("__main__" procedure in Core)
     let inlinePhases : List Core.PipelinePhase :=
       if isBugFinding then
         [Core.procedureInliningPipelinePhase
-          { doInline := fun name a => name ≠ "__main__" && Core.doInlineNonRecursive name a
+          { doInline := fun caller callee a => caller == some "__main__"
+              && Core.doInlineNonRecursive callee a
             maxIters := some 10 }]
```
tautschnig: The logic is correct for transitive inlining: since `currentProcedureName` stays `"__main__"` throughout all iterations (we only process `__main__`), nested calls discovered after inlining will also satisfy `caller == some "__main__"` in subsequent iterations.

One edge case to consider: if a user function calls itself recursively, `doInlineNonRecursive` will block it. But if function `f` calls `g` which calls `f` (mutual recursion), `doInlineNonRecursive` checks `!cg.isRecursive callee` — does `isRecursive` detect mutual recursion (cycles), or only direct self-recursion? If it only detects direct recursion, the 10-iteration cap is the only safety net against infinite expansion. Worth a quick sanity check.

StrataTest/Languages/Python/tests/test_bug_finding_unreachable.py
```diff
@@ -4,3 +4,6 @@ def test_bug_finding_unreachable(x: int) -> None:
     if x > 0:
         if x < 0:
             assert False, "dead code"
+
+if __name__ == "__main__":
+    test_bug_finding_unreachable(0)
```
tautschnig: The argument `0` makes the test deterministic (the solver knows `x=0`), which is good for a regression test. However, the original test was designed to show that `x > 0 && x < 0` is unreachable *for any* `x`. With `x=0`, the unreachability is trivially established by the concrete value rather than by logical reasoning about the condition.

Is this intentional? If the goal is to test that bug-finding mode detects unreachable code via logical reasoning (not just concrete execution), consider using a non-trivial argument like `test_bug_finding_unreachable(42)` where `x > 0` is true but `x < 0` is still unreachable.

Strata/Transform/ProcedureInlining.lean
aqjune-aws: I think it is fine because `inlineCallCmd` is an internal function. :) The external users will have to use `procedureInliningPipelinePhase`.

StrataMain.lean
aqjune-aws: `isRecursive callee` indeed checks whether the function `callee` is included in any cycle of call graph. The mutual recursive function case should be handled.

StrataTest/Languages/Python/tests/test_bug_finding_unreachable.py
aqjune-aws: This was a good point - I updated the test :)

keyboardDrummer:
My feeling is that the three modes you're introducing are putting the end-user for a difficult choice and not providing them with the best UX. I think the only option we need is `all`, IF we group diagnostics by the assertion from which they originate. That way, if an assertion is reached in multiple ways, it still only leads to a single diagnostic.

For, example, in:
```python
def f(x):
  assert(x != 1);

f(1);
```
In bug finding mode, when using `all`, the assert above is checked twice - once from main (which will fail because it is always false), once from f (which will succeed because when x is 0 the assert passes). The diagnostics should thus show a single failure, originating from the main call.

The following PR attempts to implement that grouping concept although I'm not sure how to apply it to get the same test improvements as this PR gets: https://github.com/strata-org/Strata/pull/833

StrataMain.lean
```diff
@@ -484,6 +484,26 @@ private def deriveBaseName (file : String) : String :=
...
+/-- Determine which procedures to verify based on the entry-point mode.
+    - `"main"`: only `__main__`
+    - `"roots"`: user procedures with no callers among user procedures
+    - `"all"`: all user procedures -/
+private def determineProceduresToVerify
+    (coreProgram : Core.Program) (userSourcePath : String)
+    (entryPoint : String) : Except String (List String) := do
+  let (_preludeNames, userProcNames) :=
+    Strata.splitProcNames coreProgram [userSourcePath]
+  match entryPoint with
```
shigoel: We could benefit from having `entryPoint` defined as an inductive type.

aqjune-aws:
> My feeling is that the three modes you're introducing are putting the end-user for a difficult choice and not providing them with the best UX. I think the only option we need is all, IF we group diagnostics by the assertion from which they originate. That way, if an assertion is reached in multiple ways, it still only leads to a single diagnostic.

This pull request is mainly for performance, because if the user wants to only consider reachable procedures from `main`, procedure inlining of procedures never invoked by main and running partial evaluation on non-main is redundant. :) If the user wants faster, they can give this particular setup. This can bring more aggressive optimization than grouping asserts.
On the other hand, grouping asserts will still be necessary because (1) when the option is all, indeed this optimization won't be useful at all (2) even when the option is `roots`/`all`, we will want to see related/duplicated asserts folded.

StrataMain.lean
```diff
@@ -499,7 +519,10 @@ def pyAnalyzeLaurelCommand : Command where
               takesArg := .repeat "module" },
             { name := "keep-all-files",
               help := "Store intermediate Laurel and Core programs in <dir>.",
-              takesArg := .arg "dir" }]
+              takesArg := .arg "dir" },
+            { name := "entry-point",
```
shigoel, line 505: How does this interact with the `--procedures` option that comes with `verifyOptionsFlags`?

StrataMain.lean
```diff
@@ -499,7 +519,10 @@ def pyAnalyzeLaurelCommand : Command where
...
+              help := "Which procedures to verify: main (main fn only), roots (user procs with no user callers, default), or all (all user procs). Only valid in bugFinding mode.",
```
shigoel, line 506: Instead of saying "Only valid in bugFinding mode" and silently ignoring this flag in other modes, we should raise an error that says "flag unsupported in this mode".

StrataMain.lean
aqjune-aws: Since `transformCommand` is the one that receives the `--procedures` option, it will be safely orthogonal to the new `--entry-point` option of `pyAnalyzeLaurelCommand`. In `pyAnalyzeLaurelCommand`, we didn't have an option for specifying which procedures to verify.

shigoel: Yes, I too realized that just now. Thanks :)

StrataMain.lean
aqjune-aws: Great idea - I added `EntryPoint` and defined `EntryPoint.ofString?`/`EntryPoint.options`.

StrataMain.lean
aqjune-aws: Added a sanity check for this.

shigoel (dismissed):

StrataMain.lean
```diff
@@ -484,25 +484,45 @@ private def deriveBaseName (file : String) : String :=
   | some sfx => (name.dropEnd sfx.length).toString
   | none     => name
 
+/-- Which procedures to verify: all user procedures, only roots (no callers),
+    or only the main function. -/
+inductive EntryPoint where
+  | main   -- Only `__main__`
+  | roots  -- User procedures with no callers among user procedures
```
shigoel: Nit: no blocker at all, but if you use `/-- .. -/` above each constructor, that's rendered as documentation string in the API documentation. E.g., see https://strata-org.github.io/Strata/api/Strata/DL/Lambda/LExpr.html#Lambda.LConst

aqjune-aws: Looks like a necessary stuff; I pushed a commit.

StrataTest/Languages/Python/AnalyzeLaurelTest.lean
```diff
@@ -114,19 +114,21 @@ private meta def runAnalyzeAndVerify
     | none => return .error "Laurel to Core translation failed"
     | some core => pure core
   -- Split prelude / user procedure names at FIRST_END_MARKER
-  let (_preludeNames, userProcNames) := Strata.splitProcNames coreProgram
+  let (_preludeNames, _userProcNames) := Strata.splitProcNames coreProgram
   -- Inline all non-main, non-prelude procedures as a prefix phase
   let inlinePhases : List Core.PipelinePhase :=
     [_root_.Core.procedureInliningPipelinePhase
-      { doInline := fun name a => name ≠ "__main__" && _root_.Core.doInlineNonRecursive name a }]
+      { doInline := fun caller callee a => caller == some "__main__"
```
MikaelMayer: The test hardcodes `caller == some "__main__"` (equivalent to `--entry-point main`), but the CLI default is `roots`. Consider adding a test that exercises the `roots` path as well, to ensure `determineProceduresToVerify` with `.roots` produces the expected procedure list.

StrataMain.lean
```diff
@@ -484,6 +484,49 @@ private def deriveBaseName (file : String) : String :=
   | some sfx => (name.dropEnd sfx.length).toString
   | none     => name
 
+/-- Which procedures to verify: all user procedures, only roots (no callers),
+    or only the main function. -/
+inductive EntryPoint where
+  /-- Only `__main__` -/
+  | main
+  /-- User procedures with no callers among user procedures -/
```
MikaelMayer: What happens if there are two mutually recursive methods not reachable by main? I think they should be counted as roots, but the comment would say otherwise since they do have mutual callers.

StrataTest/Languages/Python/AnalyzeLaurelTest.lean
aqjune-aws: Refactored the tester function to accept an extra argument deciding whether it is the root mode or the main mode, and extended a test that invokes this helper function so it tests twice (main vs. root).

StrataMain.lean
aqjune-aws: This was a really valid point. I fixed the algorithm as follows.
For each SCC, it picks one 'representative' procedure to start bugFinding. The representative procedure must be a user procedure; if there are multiple user procedures in the SCC, the one with smallest lexicographical order is chosen. If all nodes of SCCs are from prelude, the SCC is simply dropped.

aqjune-aws:
Addressed comments.

shigoel (dismissed):

joehendrix:
**Suggestion: Extract `EntryPoint` and procedure-selection logic out of `StrataMain.lean`**

`StrataMain.lean` is the CLI entry point, but the new `EntryPoint` inductive, `EntryPoint.ofString?`, `EntryPoint.options`, and `determineProceduresToVerify` are domain logic about verification strategy — they don't depend on CLI infrastructure at all.

I'd suggest moving them to a new module like `Strata/Languages/Core/EntryPoint.lean`, alongside `CallGraph.lean` and `Options.lean` where similar verification-configuration types live.

To make the function language-agnostic (it currently depends on `splitProcNames` from the Python pipeline), `determineProceduresToVerify` could take the user procedure list directly instead of computing it internally:

```lean
def determineProceduresToVerify
    (coreProgram : Core.Program)
    (userProcNames : List String)
    (entryPoint : EntryPoint) : List String :=
  match entryPoint with
  | .main => ["__main__"]
  | .all => userProcNames
  | .roots =>
    let cg := coreProgram.toProcedureCG
    let userSet := Std.HashSet.ofList userProcNames
    userProcNames.filter fun name =>
      (cg.getCallers name).all fun caller => !userSet.contains caller
```

The caller in `pyAnalyzeLaurelCommand` would then do the `splitProcNames` call and pass the result in. This also avoids the unnecessary `splitProcNames` traversal in `.main` mode (where the result is discarded), and makes `EntryPoint` reusable if other language frontends need similar entry-point filtering.

Not a blocker — just a suggestion for keeping `StrataMain.lean` focused on CLI plumbing.

aqjune-aws:
> Suggestion: Extract EntryPoint and procedure-selection logic out of StrataMain.lean

It was a good idea. I factored out the logic to a separate file.

shigoel (approved):

MikaelMayer (approved):

joehendrix (approved):

keyboardDrummer:
> This pull request is mainly for performance, because if the user wants to only consider reachable procedures from `main`, procedure inlining of procedures never invoked by main and running partial evaluation on non-main is redundant. :) 

Sounds like the question for the user is really, is this program an executable (so only the paths from main matter), or is this a library (so all public entry points matter) ?

IMO the user should be able to decide per method whether it is verified modularly (so the method is verified on its own, so that when calling it, assertions in the method can be ignored), or through its call-sites. I think having this feature prevents the need for an all versus a root mode.

aqjune-aws:
> IMO the user should be able to decide per method whether it is verified modularly (so the method is verified on its own, so that when calling it, assertions in the method can be ignored), or through its call-sites. I think having this feature prevents the need for an all versus a root mode.

For a benchmark that has a large set of independent programs, it would be quite amount of work from the user perspective as well. But ideally fineer-grained options will help.

PR #859, opened 2026-04-13
Add arithmetic SpecExpr operators (intAdd, intSub, intMul, intEq)

stefanzetzsche:
### Summary

Adds four new `SpecExpr` variants for integer arithmetic and equality, enabling pyspec preconditions and postconditions that reference arithmetic expressions like `result == balance - amount` or `x >= y + z`.

### Changes

- `Strata/Languages/Python/Specs/Decls.lean` — New variants: `intAdd`, `intSub`, `intMul`, `intEq`, each with `(loc : SourceRange)`
- `Strata/Languages/Python/Specs/DDM.lean` — DDM ops (`intAddExpr`, `intSubExpr`, `intMulExpr`, `intEqExpr`) with `toDDM`/`fromDDM` round-trip; `SpecExprDecl.fromDDM` changed from `private` to `protected` to enable round-trip testing
- `Strata/Languages/Python/PythonLaurelTypedExpr.lean` — Typed helpers (`intAdd`, `intSub`, `intMul`, `intEq`) alongside existing `intGeq`/`intLeq`
- `Strata/Languages/Python/Specs/ToLaurel.lean` — All six integer binary ops (`intGe`, `intLe`, `intAdd`, `intSub`, `intMul`, `intEq`) collapsed into a single match arm via or-patterns + `intBinOp` helper
- `StrataTest/Languages/Python/Specs/DeclsTest.lean` — `toDDM`/`fromDDM` round-trip tests for all four new variants
- `StrataTest/Languages/Python/ToLaurelTest.lean` — `specExprToLaurel` integration tests verifying exact Laurel AST shape (e.g. `from_int(Any..as_int!(x) + Any..as_int!(y))`)

### Motivation

icontract specs commonly use arithmetic in contracts:

```python
@icontract.ensure(lambda balance, amount, result: result == balance - amount)
@icontract.require(lambda balance: balance >= first + second)
```

Without these operators, such expressions fall back to `.placeholder` and can't be verified.

### Part of a stack

This is PR 1 of 3 in the icontract integration stack:

1. **This PR** — Arithmetic SpecExpr operators
2. https://github.com/strata-org/Strata/pull/863
3. https://github.com/strata-org/Strata/pull/864


joehendrix (dismissed):

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -389,6 +389,41 @@ def specExprToLaurel (e : SpecExpr) (md : Imperative.MetaData Core.Expression)
...
+      let prod := mkStmt (.PrimitiveOp .Mul [lInt, rInt]) md
+      some (mkStmt (.StaticCall (mkId "from_int") [prod]) md)
+  | .intEq left right loc => do
+    let md ← nodeMd loc
+    let l? ← specExprToLaurel left md; let r? ← specExprToLaurel right md
+    return do
+      let l ← l?; let r ← r?
+      let lInt := mkStmt (.StaticCall (mkId "Any..as_int!") [l]) md
+      let rInt := mkStmt (.StaticCall (mkId "Any..as_int!") [r]) md
+      some (mkStmt (.PrimitiveOp .Eq [lInt, rInt]) md)
```
tautschnig: The four new cases are correct but nearly identical to each other and to the existing `intGe`/`intLe` cases. The only variation is the `PrimitiveOp` variant and whether the result is wrapped in `from_int`. There are now 6 cases (and counting, once float arithmetic lands) that share the same `as_int!` / operate / optionally-wrap pattern.

Consider extracting a helper, e.g.:
```lean
private def intBinOp (op : PrimitiveOp) (wrap : Bool)
    (left right : SpecExpr) (loc : SourceRange)
    (md : Imperative.MetaData Core.Expression) : SpecToLaurelM (Option StmtExprMd) := do
  let l? ← specExprToLaurel left md; let r? ← specExprToLaurel right md
  return do
    let l ← l?; let r ← r?
    let lInt := mkStmt (.StaticCall (mkId "Any..as_int!") [l]) md
    let rInt := mkStmt (.StaticCall (mkId "Any..as_int!") [r]) md
    let result := mkStmt (.PrimitiveOp op [lInt, rInt]) md
    some (if wrap then mkStmt (.StaticCall (mkId "from_int") [result]) md else result)
```
Then each case becomes a one-liner. This would also make it trivial to add `intDiv`, `intMod`, `intNe`, etc. later.

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -324,6 +324,14 @@ inductive SpecExpr where
...
 | intGe (subject : SpecExpr) (bound : SpecExpr) (loc : SourceRange)
 | intLe (subject : SpecExpr) (bound : SpecExpr) (loc : SourceRange)
+/-- Integer addition: `intAdd a b` represents `a + b`. -/
+| intAdd (left : SpecExpr) (right : SpecExpr) (loc : SourceRange)
+/-- Integer subtraction: `intSub a b` represents `a - b`. -/
+| intSub (left : SpecExpr) (right : SpecExpr) (loc : SourceRange)
+/-- Integer multiplication: `intMul a b` represents `a * b`. -/
+| intMul (left : SpecExpr) (right : SpecExpr) (loc : SourceRange)
+/-- Integer equality: `intEq a b` represents `a == b`. -/
+| intEq (left : SpecExpr) (right : SpecExpr) (loc : SourceRange)
```
tautschnig, line 318: No tests for the new variants. Since this is PR 1 of 3 and the constructors aren't created by any parser yet (only by `fromDDM`), end-to-end tests will come with PRs #863/#864. But there are two things worth testing now:

1. **`toDDM`/`fromDDM` round-trip** for the new variants. A `#guard` test like:
```lean
#guard (SpecExpr.intAdd (.var "x" loc) (.intLit 1 loc) loc).toDDM.fromDDM ==
       SpecExpr.intAdd (.var "x" loc) (.intLit 1 loc) loc
```
would catch serialization bugs. (This requires `BEq` on `SpecExpr`, which may not exist yet — if not, a string round-trip via `toString` would also work.)

2. **`specExprToLaurel` unit test** for at least one arithmetic op, verifying the generated Laurel AST has the expected shape (`from_int(as_int(l) + as_int(r))`).

These are cheap to write and would prevent regressions if the `toDDM`/`fromDDM` or translation logic is refactored.

shigoel:
@stefanzetzsche Could you please resolve merge conflicts?

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -380,6 +380,23 @@ private def lookupIdentifier (name : String) (loc : SourceRange) (md : Md)
...
+    (l r : TypedStmtExpr Laurel.tyAny) (md : Md)
+    : SomeTypedStmtExpr :=
+  let li := .anyAsInt l md; let ri := .anyAsInt r md
+  match e with
+  | .intGe  .. => .mkSome <| .intGeq li ri
+  | .intLe  .. => .mkSome <| .intLeq li ri
+  | .intAdd .. => .mkSome <| .fromInt (.intAdd li ri)
+  | .intSub .. => .mkSome <| .fromInt (.intSub li ri)
+  | .intMul .. => .mkSome <| .fromInt (.intMul li ri)
+  | _          => .mkSome <| .intEq li ri  -- intEq and fallback
```
joehendrix, line 398: The catch-all `| _ =>` here maps any unmatched `SpecExpr` variant to `intEq`. Currently safe because the single call site (line 453) restricts input to exactly six variants via the or-pattern, but this defeats Lean's exhaustiveness checking: if a future integer operator (e.g., `intDiv`) is added to the or-pattern without a corresponding arm here, it would silently produce equality semantics — a subtle miscompilation.

```suggestion
  | .intEq  .. => .mkSome <| .intEq li ri
```

With an explicit match, Lean will raise a compile-time error if `intBinOp` is ever called with an unhandled variant.

joehendrix (commented):
Two suggestions from reviewing the PR — one on defensive matching in `intBinOp`, one on test assertions. Details in the inline comments.

StrataTest/Languages/Python/ToLaurelTest.lean
```diff
@@ -803,4 +803,64 @@ private def hasTypeError (result : TranslationResult) : Bool :=
...
 
+/-! ## Arithmetic SpecExpr integration tests -/
+
+-- intEq: x == 0
+/--
+info: { assert !Any..isfrom_None(x); assert Any..as_int!(x) == Any..as_int!(from_int(0)) }
+-/
+#guard_msgs in
+#eval do
+  let (body, _) := translatePrecond
```
joehendrix, line 814: `translatePrecond` returns `(bodyString, errorCount)` but the error count is discarded in all four new arithmetic tests (lines 814, 826, 841, 858). If the translation silently produced errors while still generating body text, the `#guard_msgs` check would still pass.

The existing precondition tests in this file assert zero errors (e.g., line 739: `assert! errs == 0`). These should follow the same pattern:

```suggestion
  let (body, errs) := translatePrecond
```

Then add `assert! errs == 0` before `IO.println body` in each of the four tests.

PR #865, opened 2026-04-13
Denotational Semantics for Lambda

joscoh:
*Issue #, if available:*

*Description of changes:* This PR adds a denotational semantics for Lambda and proofs of consistency between this semantics and the existing small-step operational semantics. The structure of the added files is described in `Strata/DL/Lambda/Denote/README.md`. Some tests using the semantics are in `StrataTest/DL/Lambda/Denote/Tautologies.lean`.

Beyond the denotational semantics itself, this PR includes several changes to existing Lambda infrastructure:

### Factory (`Factory.lean`)
- Added a `nameMapConsistent` field ensuring that each key in the name map matches the name of the element it points to. This is needed to prove that factory lookups return the expected function.
- Added several utility theorems (`push_mem_iff`, `mem_iff_mem_names`, `getElem?_some_implies_mem`, `mem_name_eq_getElem`, etc.) used throughout the denotation proofs.

### Type substitution (`LTyUnify.lean`, `LTy.lean`)
- Added `LMonoTy.subst_unfold`, which gives a clean unfolding lemma for `LMonoTy.subst` that eliminates the `hasEmptyScopes` guard. This is essential for reasoning about type substitution in the denotation proofs.
- Added `LMonoTy.mkArrow'_injective` and supporting lemmas (`subst_bitvec`, `subst_nil`, `hasEmptyScopes_false_of_find`, etc.).

### TypeFactory derived functions (`TypeFactory.lean`, `TypeFactoryWF.lean`)
- Fixed type annotations in `testerFuncBody` and `testerFunc`: the eliminator op and free variable references now carry their types (`some elimType`, `some (dataDefault d)`) instead of `none`. This is required for the well-typedness assumptions used by the denotational semantics.
- Added `genRecTyDirect` to compute concrete case argument types for tester functions.
- Added `TypeFactoryWF` structure capturing well-formedness of a `TypeFactory` (currently: unique datatype names across mutual blocks).

### Other changes
- `Semantics.lean`: Simplified `callOfLFunc_getLFuncCall` and `computeTypeSubst_eraseMetadata_congr` proofs.
- `LExprWF.lean`: Added `liftBVars_eq_of_lcAt` and `@[expose]` on `substK`/`subst`.
- `Strata/DL/Util/`: Added utility lemmas for `Map`, `ListMap`, `Maps`, and `List` used by the denotation proofs.

This PR introduces several `sorry`s in a test file because `HashMaps` do not reduce in the Lean kernel. I will try to fix these in a future PR.


Strata/DL/Util/Map.lean
```diff
@@ -226,12 +451,53 @@ theorem Map.insert_values [DecidableEq α] (m : Map α β) :
...
+    · rename_i heq
+      cases h
+      exact ⟨0, by simp [heq], by simp, fun j hj => by omega⟩
+    · rename_i hne
+      obtain ⟨i, hk, hv, hfirst⟩ := ih h
+      refine ⟨i + 1, by simpa using hk, by simpa using hv, fun j hj => ?_⟩
+      cases j with
+      | zero => simp; exact hne
+      | succ n => simpa using hfirst n (by omega)
+/-- If `find?` succeeds on `m` at key `a`, and `cs` has the same length as `m`,
```
atomb, line 476: Tiny nit: it'd be nice to have a newline before this line.

Strata/DL/Lambda/Denote/README.md
```diff
@@ -0,0 +1,86 @@
...
+The second set of assumptions are given in `Assumptions.lean` and we will
+eventually prove that they are satisfied of well-typed terms.
+The first set is much trickier to handle. Implicitly, each concrete function and
+datatype definition gives rise to a set of axioms concerning its behavior, and
+these assumptions are intended to hold if these axioms are consistent. If the
+assumptions do not hold (for instance, if the typing rules are too permissive),
+then validity becomes trivial. Proving that consistent interpretations exist
+is possible but very difficult with recursive functions.
+TODO: Show that consistent interpretations exist for a simple but nontrivial
+example.
```
atomb, line 72: I don't think we need to do this in this PR, but it'd be nice to have an issue for it so we don't forget.

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1170 @@
...
+string, bitvec, arrow) are mapped to their Lean counterparts; all others
+are delegated to `tcInterp`. -/
+@[expose] public def SortDenote (tcInterp : TyConstrInterp) : LSort → Type
+  | .tcons "bool" []      => Bool
+  | .tcons "int" []       => Int
+  | .tcons "real" []      => Rat
+  | .tcons "string" []    => String
+  | .bitvec n             => BitVec n
+  | .tcons "arrow" [a, b] => SortDenote tcInterp a → SortDenote tcInterp b
+  | .tcons name args      => tcInterp name args
```
atomb, line 158: This may be obvious elsewhere in the code, but what would the denotation of constructor applications with no interpretation look like, given that `tcInterp` is a total function?

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1170 @@
...
+
+/-- Interpret a well-typed annotated `LExpr` into a Lean value.
+
+`opInterp` and `fvarVal` take sorts (ground types) rather than monomorphic
+types, making them independent of the type variable valuation `ρ`. This
+cleanly separates interpretations (fixed for a theory) from valuations
+(vary per context), enabling change-of-valuation theorems.
+
+Noncomputable due to `Classical.propDecidable` in the `eq` and `quant`
+cases; used only for reasoning, not computation. -/
```
atomb, line 256: It's not urgent (and so doesn't need to be in this PR), but I wonder whether it's worth defining a predicate that would classify computable Lambda expressions, and having a computable variation of this function that requires its input be a computable Lambda expression. Given how short this function is, it probably wouldn't be too hard to write that function and show that the two agree on any input that they both accept.

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1170 @@
...
+    {T : LExprParams}
+    (tcInterp : TyConstrInterp)
+    (opInterp : OpInterp tcInterp)
+    (fvarVal : FreeVarVal T tcInterp)
+    (vt : TyVarVal)
+    {Δ : List LMonoTy}
+    (bvarVal : BVarVal tcInterp vt Δ)
+    (e : LExpr T.mono) (τ : LMonoTy)
+    (h : HasTypeA Δ e τ)
+    : TyDenote tcInterp vt τ :=
```
atomb, line 267: Would it be possible to write an abbreviation for `denote tcInterp opInterp fvarVal vt` in a `let` expression here that would both type check and termination check? It's a minor thing, but it would make recursive calls so much nicer.

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1170 @@
...
+    if cond
+    then denote tcInterp opInterp fvarVal vt bvarVal t τ h_t
+    else denote tcInterp opInterp fvarVal vt bvarVal e τ h_e
+  | .eq _ e1 e2 =>
+    let ⟨ty', h_bool, h_1, h_2⟩ := HasTypeA.eq_inv h
+    let v1 := denote tcInterp opInterp fvarVal vt bvarVal e1 ty' h_1
+    let v2 := denote tcInterp opInterp fvarVal vt bvarVal e2 ty' h_2
+    h_bool ▸ (Classical.propDecidable (v1 = v2) |>.decide)
+  | .quant _ .all _ (some qty) tr body =>
+    let ⟨_τ_tr, h_bool, _h_tr, h_body⟩ := HasTypeA.quant_inv h
```
atomb, line 300: Is it possible to share more between the `.all` and `.exist` cases, but matching on that argument later? Or do the types not work out well?

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1170 @@
...
+    (tcInterp : TyConstrInterp)
+    (opInterp : OpInterp tcInterp)
+    (fvarVal : FreeVarVal T tcInterp)
+    (vt : TyVarVal)
+    {Δ : List LMonoTy}
+    (bvarVal : BVarVal tcInterp vt Δ)
+    (e : LExpr T.mono) (τ : LMonoTy)
+    (h : LExpr.HasTypeA Δ e τ)
+    : Denotes tcInterp opInterp fvarVal vt bvarVal e τ h
+        (LExpr.denote tcInterp opInterp fvarVal vt bvarVal e τ h) := by
```
atomb, line 414: It's nice that this theorem and the following one aren't that complex to prove. They get slightly long because there are a bunch of cases, but none of the cases are very complex.

Strata/DL/Lambda/Denote/LExprAnnotated.lean
```diff
@@ -0,0 +1,534 @@
...
+    guard (cty == .bool)
+    guard (tty == ety)
+    some tty
+  | .eq _ e1 e2 => do
+    let ty1 ← typeCheck ctx e1
+    let ty2 ← typeCheck ctx e2
+    guard (ty1 == ty2)
+    some .bool
+
+/-- Declarative typing rules for annotated expressions. -/
```
aqjune-aws: Let's add a comment explaining the meaning of the first argument of `HasTypeA` - they are types of bound variables, ordered by De Bruijn indices.

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1170 @@
...
+@[expose] public def TyConstrInterp := String → List LSort → Type
+
+/-- Every type produced by a `TyConstrInterp` is inhabited. -/
+class TyConstrInterp.AllInhabited (tcInterp : TyConstrInterp) : Type where
+  inhabited : ∀ name args, Inhabited (tcInterp name args)
+
+@[expose] public def isArrow (n: String) (args: List LSort) : Option (LSort × LSort) :=
+  match n, args with
+  | "arrow", [a, b] => some (a, b)
+  | _ , _ => none
```
aqjune-aws: It seems this function isn't used? I temporarily commented out this function and proofs built.

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1170 @@
...
+
+/-! ### Bound Variable Valuation -/
+
+/-- Bound variable valuation: an `HList` of semantic values indexed by the
+typing context. -/
+abbrev BVarVal (tcInterp : TyConstrInterp)
+    (ρ : TyVarVal) (Δ : List LMonoTy) :=
+  HList (TyDenote tcInterp ρ) Δ
+
+/-! ### Intepreting Free Variables and Operators -/
```
aqjune-aws: Intepreting -> Interpreting

Strata/DL/Lambda/Denote/README.md
```diff
@@ -0,0 +1,86 @@
+# Lambda's Denotational Semantics
```
aqjune-aws, line 1: This file is great :)

Strata/DL/Lambda/Denote/LExprAnnotated.lean
```diff
@@ -0,0 +1,534 @@
...
+    let tty ← typeCheck ctx t
+    let ety ← typeCheck ctx e
+    guard (cty == .bool)
+    guard (tty == ety)
+    some tty
+  | .eq _ e1 e2 => do
+    let ty1 ← typeCheck ctx e1
+    let ty2 ← typeCheck ctx e2
+    guard (ty1 == ty2)
+    some .bool
```
aqjune-aws, line 63: This function and `HasTypeA` relies on the fact that subexpressions are fully annotated with inferred types from `LExpr.resolve`. To make this assumption useful, I think we will need to plan next steps (can be left as Github issues for future):
- The type annotation pass: we have a pass that fills in the optional types with the results from `LExpr.resolve`. This will be needed for the end-to-end Lean denotation generation (one concern is - can we really use `LExpr.resolve` as it is...? If modification is necessary probably we should)
- Will we need to have a theorem stating the relation between `HasTypeA e ty` and `HasType e ty` (if the former holds then the latter works)? Or, if we have to modify `LExpr.resolve` anyway, the former `HasTypeA` can be simply chosen to restate the spec of `HasType`.

aqjune-aws (commented):
I left a few comments - I will leave a few more in the afternoon.

Strata/DL/Lambda/Denote/LExprAnnotated.lean
joscoh: 1. I think we should be able to use `LExpr.resolve` as-is (not 100% sure), but agreed that this is something we need to prove. I have a task for this already. I can create Github issues as well.
2. It should be the case (and not too hard to prove) that `HasTypeA` implies `HasType`. Since as per (1) we will want to prove that `LExpr.resolve` results in a term satisfying `HasTypeA`, I think that the existing soundness proof could be refactored into these two components. This is basically the experiment I did at https://github.com/strata-org/Strata/tree/josh/h-m/Strata/Experiment/HM.

Separately, I still think that `HasType` is useful. It is strictly weaker than `HasTypeA`, which means that we get a stronger progress theorem (and an incomparable preservation theorem) using this typing relation.

Strata/DL/Lambda/Denote/Assumptions.lean
```diff
@@ -0,0 +1,195 @@
...
+import all Strata.DL.Lambda.FactoryWF
+
+/-!
+## Typing Assumptions
+
+Collects all well-formedness and typing assumptions used by the denotational
+semantics proofs. Semantic (denotation-level) assumptions live in
+`LExprSemanticsConsistent`.
+
+### Expression-level predicates (TODO: prove satisfied after `LExpr.resolve`)
```
atomb, line 20: If we don't already have GitHub issues for this and the TODOs below, could you create some?

Strata/DL/Lambda/Denote/CallOfLFuncDenote.lean
```diff
@@ -0,0 +1,769 @@
...
+        (h_args : List.Forall₂ (LExpr.HasTypeA []) args argTys)
+        (hty_op: ty_op = LMonoTy.mkArrow' τ argTys),
+        callee = .op m name (some ty_op) ∧
+        let h_eq : LMonoTy.substTyVars vt ty_op =
+              LSort.mkArrow (LMonoTy.substTyVars vt τ) (argTys.map (LMonoTy.substTyVars vt)) :=
+            hty_op ▸ substTyVars_mkArrow' vt τ argTys
+        LExpr.denote tcInterp opInterp fvarVal vt .nil e τ h =
+          SortDenote.applyArgs tcInterp
+            (h_eq ▸ opInterp name.name (LMonoTy.substTyVars vt ty_op))
+            (denoteArgs tcInterp opInterp fvarVal vt .nil args argTys h_args) := by
```
atomb, line 742: I do wish the statement of this theorem could be simpler, but it's not immediately clear to me how to accomplish that.

Strata/DL/Lambda/Denote/LExprAnnotated.lean
```diff
@@ -0,0 +1,534 @@
...
+
+/-- Declarative typing rules for annotated expressions. -/
+inductive LExpr.HasTypeA {T : LExprParams} : List LMonoTy → LExpr T.mono → LMonoTy → Prop where
```
atomb, line 71: I love this definition.

Strata/DL/Lambda/Denote/LExprDenote.lean
joscoh: Thanks, I had moved it to `LTy` I believe but had forgotten to delete the redundant version

StrataTest/DL/Lambda/Denote/Tautologies.lean
atomb, line 1: I would love to explore how feasible it would be for the examples in this file to eventually be provable as straightforwardly as something like [this](https://github.com/strata-org/Strata/blob/main/StrataTest/Languages/Boole/find_max.lean#L55-L60).

atomb (dismissed):
This is really nice. A few of my comments suggest very minor changes, but I'll leave it up to you whether to make them (and whether to do it now or later, if so).

StrataTest/DL/Lambda/Denote/Tautologies.lean
joscoh: Agreed, I think that there are 2 primary tasks to be done:
1. Figure out a way to simplify `Factory`s. The SMT operators (e.g. `and`, `or`, etc) have semantics defined inline; the `Factory`-based approach of Lambda involves a lot more boilerplate. And we can't even prove once-and-for-all lemmas like `denote (P /\ Q) = denote P /\ denote Q`, since the `Factory` is not fixed but rather needs to define `/\` in the correct way.
2. Write some tactics to automatically unfold and simplify the denotations.

In the long run, I would like something better: using Lean metaprogramming, to construct Lean expressions whose proofs imply the validity of the underlying Lambda goal. But this will involve creating e.g. datatypes and function definitions on the fly, so it could get tricky, especially with polymorphism.

For now, I will add a task and Github issue for this. I think these questions should be postponed until a future PR.

atomb: Yeah, I agree. I didn't make it clear in my initial comment, but it's definitely an aspirational, long-term goal, not something to address in this PR.

Strata/DL/Lambda/Denote/LExprAnnotated.lean
aqjune-aws: > we will want to prove that LExpr.resolve results in a term satisfying HasTypeA, I think that the existing soundness proof could be refactored into these two components.

Makes sense! And since we have the exact spec of `LExpr.resolve` now, editing `LExpr.resolve` can be quite aggressive than what was possible in the past. It is fine as long as it still satisfies the spec.

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1165 @@
...
+where
+  map : List LMonoTy → List LSort
+  | [] => []
+  | x :: xs => substTyVars ρ x :: map xs
+
+/-- Interpretation of type constructors: maps a constructor name and its
+sort arguments to a Lean `Type`. -/
+@[expose] public def TyConstrInterp := String → List LSort → Type
+
+/-- Every type produced by a `TyConstrInterp` is inhabited. -/
```
aqjune-aws: Let's add a comment stating that `TyConstrInterp.AllInhabited` holds for inductive datatypes defined in Core because they are supposed to have at least one element.

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1165 @@
...
+    . rename_i n _; exact ⟨(0 : BitVec n)⟩
+    · exact ⟨fun _ => (ih _ (by simp_all)).default⟩
+    · exact h _ _
+  | hbitvec n => exact ⟨(0 : BitVec n)⟩
+
+instance SortDenote.instInhabited [TyConstrInterp.AllInhabited tcInterp]
+    (s : LSort) : Inhabited (SortDenote tcInterp s) :=
+  SortDenote.inhabited tcInterp TyConstrInterp.AllInhabited.inhabited s
+
+/-- Type-variable valuation: maps each type variable to a sort. -/
```
aqjune-aws: How are the TyIdentifiers that are not in the list of substituting variables defined? (e.g., if we want to say `ty_x` is not going to be substituted by any type in a TyVarVal...) Are they returning an identity type?

Strata/DL/Lambda/Denote/LExprDenote.lean
aqjune-aws: Or there can even be a case where a type of `.tcons name` has not been defined anywhere in a Core program - in this case, returned values of `tcInterp` will be arbitrary, making us unable to prove anything. If this understanding is right, let's mention this explicitly at the comment above definition of `TyConstrInterp`.

Strata/DL/Util/HList.lean
```diff
@@ -0,0 +1,103 @@
...
+Type-indexed lists used to represent bound-variable valuations in the
+denotational semantics.
+-/
+
+namespace Lambda
+
+public section
+
+/-- A heterogeneous list indexed by a list of elements of type `α`. -/
+inductive HList {α : Type} (f : α → Type) : List α → Type where
```
aqjune-aws, line 18: Maybe we can move this to Strata/DL/Util. I thought heterogeneous list was supposed to have `Type 1`, but it seems this is possible without defining in that way.
```
-- The one without 'type signature'
inductive HList where
  | nil  : HList
  | cons : (ty : Type) → a → HList → HList
```

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -0,0 +1,1165 @@
...
+/-! ### ADT Interpretation Consistency
+
+These properties express that the semantic interpretation (`opInterp`) of
+datatype constructors satisfies the standard algebraic datatype axioms:
+disjointness and injectivity. Note that this is NOT complete: there are two
+other properties ADT interpretations should have:
+1. an inversion principle asserting that we can produce the constructor and
+arguments that give rise to a given ADT interpretation instance
+2. a generalized induction principle (this justifies recursive functions
+over these types)
```
aqjune-aws, line 1059: I would like this comment to have explicit explanation stating that the denotational semantics on ADT won't directly generate the corresponding ADT in Lean, but will use a set of axioms. Probably we can add a reference to this style? :)

Strata/DL/Lambda/Denote/LExprDenote.lean
joscoh: I'm not sure I quite understand the question, so please feel free to ask follow ups.

We can always assume `tcInterp` is total; if we only care about its behavior on a subset of values we can set it to (e.g.) `unit` for everything else. The only conditions we impose on the interpretation are those implied by the `ConstrInterpConsistent` - we have to match the constructors in the `Factory` to injective and disjoint interpretations (when combined with `Factory.ConstrWellFormed` in `Assumptions.lean`, this gives a condition on the interpretations of ADTs on the given `TypeFactory`). So there are 3 kinds of types we may have:
1. Defined type constructors (from ADTs) - these are in the `TypeFactory` and have conditions on `tcInterp`
2. Uninterpreted type constructors - these are given arbitrary interpretations, which matches their semantics (we must be able to prove things about any possible interpretation)
3. Type constructors that do not appear in a Core program - by typing, these will never occur in a proof goal/term we care about. So we are free to do anything, such as give an arbitrary interpretation.

Strata/DL/Lambda/Denote/LExprDenote.lean
joscoh: Unfortunately, I don't think this will work, since Lean (annoyingly) does not unfold `let`-bindings for termination checking. We could use a mutually recursive `with` clause, but then a bunch of proofs need to be adjusted. Let me know what you prefer.

Strata/DL/Lambda/Denote/CallOfLFuncDenote.lean
joscoh: Yeah unfortunately the statement itself is complex enough: evaluating the denotation of a function applied to arguments is equivalent to interpreting the body after substituting the arguments for the inputs. And then we need some type casts to get the types to work out. I agree it would be nice to simplify a bit but I would need to think about it more.

Strata/DL/Lambda/Denote/Assumptions.lean
joscoh: https://github.com/strata-org/Strata/issues/988 created

Strata/DL/Lambda/Denote/README.md
joscoh: Created https://github.com/strata-org/Strata/issues/990

StrataTest/DL/Lambda/Denote/Tautologies.lean
joscoh: Created https://github.com/strata-org/Strata/issues/991

Strata/DL/Lambda/Denote/LExprDenote.lean
joscoh: For any type variable in a particular type, we have to substitute according to the `TyVarVal`. For any other type variable, we are free to assign it any sort (e.g. `int`) because this value will never be used. Added a comment clarifying this. 

As a side note, it is possible to prove that the denotation depends only on the value of `TyVarVal` on the type variables that appear in the term's type (and similar for `TyConstrInterp`). But the dependent types make stating and proving this theorem very difficult, so I won't prove it unless we need it.

Strata/DL/Lambda/Denote/LExprDenote.lean
atomb: That answers my question.

Strata/DL/Lambda/Denote/LExprDenote.lean
atomb: Nah, it's probably not worth it. Just thought I'd check!

Strata/DL/Lambda/Denote/CallOfLFuncDenote.lean
atomb: Yeah, it's not critical.

Strata/DL/Util/HList.lean
joscoh: Moved to `Strata/DL/Util`. It is important for our purposes to have the `\alpha` and `f` explicit, since when casting, we want to reason about the equality of the underlying lists of type `List \alpha` and not something of type `List Type` or `HList` directly.

Strata/DL/Lambda/Denote/LExprDenote.lean
joscoh: This turns out to be a bit tricky because of the inversion on the typing proof to produce the inner types in the `.app` and `.abs` case. We would need to make these functions transparent and computable as well, or else use the typechecker to find these types (or have `LExpr.resolve` store more information). This can be done, but I think it is best left for a future PR. Added an issue https://github.com/strata-org/Strata/issues/992.

Strata/DL/Lambda/Denote/LExprDenote.lean
joscoh: Improved this, and this led me to refactoring the duplicated quantifier proofs with a separate congruence lemma, making a bunch of proofs shorter. Thanks!

Strata/DL/Lambda/Denote/LExprDenote.lean
```diff
@@ -281,17 +296,12 @@ noncomputable def LExpr.denote
...
     h_bool ▸ (Classical.propDecidable
-      (∀ x : TyDenote tcInterp vt qty,
-        (denote tcInterp opInterp fvarVal vt (.cons x bvarVal) body .bool h_body : Bool) = true)
-      |>.decide)
-  | .quant _ .exist _ (some qty) tr body =>
-    let ⟨_τ_tr, h_bool, _h_tr, h_body⟩ := HasTypeA.quant_inv h
-    h_bool ▸ (Classical.propDecidable
-      (∃ x : TyDenote tcInterp vt qty,
-        (denote tcInterp opInterp fvarVal vt (.cons x bvarVal) body .bool h_body : Bool) = true)
+      (match k with | .all => ∀ x, pred x | .exist => ∃ x, pred x)
```
atomb, line 304: Nice!

atomb (approved):
Nice reduction in proof size from the quantifier congruence lemmas!

Strata/DL/Lambda/Denote/LExprDenoteProps.lean
```diff
@@ -0,0 +1,735 @@
...
+    (vt : TyVarVal)
+    {Δ : List LMonoTy}
+    (bvarVal : BVarVal tcInterp vt Δ)
+    {e₁ : LExpr T.mono} {τ : LMonoTy} (f : T.Metadata → NewMetadata)
+    (h₁ : LExpr.HasTypeA Δ e₁ τ):
+    let T' : LExprParams := ⟨NewMetadata, T.IDMeta⟩
+    let opInterp' : OpInterp tcInterp := opInterp
+    let fvarVal' : FreeVarVal T' tcInterp := fvarVal
+    LExpr.denote tcInterp opInterp fvarVal vt bvarVal e₁ τ h₁ =
+    LExpr.denote tcInterp opInterp' fvarVal' vt bvarVal (LExpr.replaceMetadata e₁ f) τ (replaceMetadata_HasTypeA f h₁) := by
```
aqjune-aws, line 454: This is good. :)

Strata/DL/Lambda/Denote/LExprDenoteEq.lean
```diff
@@ -0,0 +1,573 @@
...
+
+theorem eqModuloMeta_true_implies_denote_eq
+    {Δ : List LMonoTy}
+    {e₁ e₂ : LExpr T.mono} {τ : LMonoTy}
+    (bvarVal : BVarVal tcInterp vt Δ)
+    (h₁ : LExpr.HasTypeA Δ e₁ τ)
+    (h₂ : LExpr.HasTypeA Δ e₂ τ)
+    (heql : LExpr.eqModuloMeta e₁ e₂ = true)
+    : LExpr.denote tcInterp opInterp fvarVal vt bvarVal e₁ τ h₁ =
+      LExpr.denote tcInterp opInterp fvarVal vt bvarVal e₂ τ h₂ := by
```
aqjune-aws, line 47: While investigating the performance of symbolic simulation, there was a suggestion from AI that `LExpr.eqModuloMeta` could be a bottleneck because its `e1/e2.eraseMetadata` unconditionally traverses the tree and this is suboptimal. I thought this was not our case because `perf stat` didn't show that function as a hot code path in the report, but if all other performance bottlenecks are resolved then `eqModuloMeta` will need to be optimized and the proof might need some updates :) (just FYI)

aqjune-aws (approved):

PR #867, opened 2026-04-13
Teach pySpecs to recognize @icontract.require and @icontract.ensure d…

stefanzetzsche:
### Summary

Extends `pySpecs` to extract preconditions and postconditions from [icontract](https://github.com/Parquery/icontract) decorators, in addition to the existing `assert`-based extraction.

### Changes

- `Strata/Languages/Python/Specs.lean` — The decorator loop in `pySpecFunctionArgs` now recognizes `@icontract.require(lambda ...: cond)` and `@icontract.ensure(lambda ...: cond)`, extracting the lambda body as a precondition or postcondition via `transAssertExpr`. Unknown icontract decorators produce a warning. Precondition messages use `SpecExpr.toMessage` for human-readable output (e.g. `len(x) >= 1`) instead of raw AST dumps.
- `Strata/Languages/Python/Specs/Decls.lean` — Adds `SpecExpr.toMessage`, a precedence-aware pretty-printer for `SpecExpr` that generates human-readable assertion messages with correct parenthesization.

### Motivation

icontract is a widely-used Python design-by-contract library. Supporting it in `pySpecs` means users can write specs as:

```python
@icontract.require(lambda balance: balance >= 0)
@icontract.require(lambda amount: amount >= 0)
@icontract.ensure(lambda result: result >= 0)
def withdraw(balance: int, amount: int) -> int:
    safe = cap(amount, balance)
    return balance - safe
```

Both icontract-decorated and assert-based formats now produce identical `.pyspec.st.ion` output through the same `pySpecs` command.

### DIFF
--- a/Strata/Languages/Python/Specs.lean
+++ b/Strata/Languages/Python/Specs.lean
@@ -1218,14 +1218,34 @@ def pySpecFunctionArgs (fnLoc : SourceRange)
                        (decorators : Array (expr SourceRange))
                        (returns : Option (expr SourceRange)) : PySpecM FunctionDecl := do
   let mut overload : Bool := false
+  -- Collect icontract preconditions and postconditions from decorators
+  let mut icontractRequires : Array (expr SourceRange) := #[]
+  let mut icontractEnsures : Array (expr SourceRange) := #[]
   for pyd in decorators do
-    let (success, d) ← runChecked <| pySpecValue pyd
-    if success then
-      match d with
-      | .typingOverload =>
-        overload := true
-      | _ =>
-        specError pyd.ann s!"Decorator {repr d} not supported."
+    -- Check for @icontract.require(lambda ...: cond) or @icontract.ensure(lambda ...: cond)
+    match pyd with
+    | .Call _ (.Attribute _ (.Name _ ⟨_, "icontract"⟩ (.Load _)) ⟨_, attr⟩ (.Load _)) args _ =>
+      if args.val.size ≥ 1 then
+        match args.val[0]! with
+        | .Lambda _ _lamArgs lamBody =>
+          if attr == "require" then
+            icontractRequires := icontractRequires.push lamBody
+          else if attr == "ensure" then
+            icontractEnsures := icontractEnsures.push lamBody
+          else
+            specWarning pyd.ann s!"Unknown icontract decorator: icontract.{attr}"
+        | _ =>
+          specWarning pyd.ann s!"icontract.{attr} expects a lambda argument"
+      else
+        specWarning pyd.ann s!"icontract.{attr} requires at least one argument"
+    | _ =>
+      let (success, d) ← runChecked <| pySpecValue pyd
+      if success then
+        match d with
+        | .typingOverload =>
+          overload := true
+        | _ =>
+          specError pyd.ann s!"Decorator {repr d} not supported."
 
   let .mk_arguments _ ⟨_, posonly⟩ ⟨_, posArgs⟩ ⟨_, vararg⟩ ⟨_, kwonly⟩ ⟨_, kw_defaults⟩ ⟨_, kwarg⟩ ⟨_, defaults⟩ := arguments
   assert! posonly.size = 0
@@ -1288,7 +1308,22 @@ def pySpecFunctionArgs (fnLoc : SourceRange)
         match returns with
         | none => pure <| .ident fnLoc .typingAny
         | some tp => pySpecType tp
-  let as ← collectAssertions argDecls returnType <|
+  let as ← collectAssertions argDecls returnType <| do
+    -- Process icontract @require decorators as preconditions
+    for reqExpr in icontractRequires do
+      let formula ← transAssertExpr reqExpr
+      let msg := SpecExpr.toMessage formula
+      let message := #[MessagePart.str msg]
+      modify fun s => { s with
+        assertions := s.assertions.push { message, formula }
+      }
+    -- Process icontract @ensure decorators as postconditions
+    for ensExpr in icontractEnsures do
+      let formula ← transAssertExpr ensExpr
+      modify fun s => { s with
+        postconditions := s.postconditions.push formula
+      }
+    -- Process body assertions (assert statements)
     if overload then
       -- Overload stubs should have `...` as their only body statement.
       unless body.size = 1 &&
--- a/Strata/Languages/Python/Specs/Decls.lean
+++ b/Strata/Languages/Python/Specs/Decls.lean
@@ -335,6 +335,29 @@ inductive SpecExpr where
 | forallDict (dict : SpecExpr) (keyVar : String) (valVar : String) (body : SpecExpr) (loc : SourceRange)
 deriving Inhabited
 
+/-- Generate a human-readable message from a SpecExpr.
+    Handles operator precedence to insert parentheses where needed.
+    Used for assertion failure messages and postcondition labels. -/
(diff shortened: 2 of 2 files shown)


Strata/Languages/Python/Specs.lean
```diff
@@ -1218,14 +1218,34 @@ def pySpecFunctionArgs (fnLoc : SourceRange)
...
-      | .typingOverload =>
-        overload := true
-      | _ =>
-        specError pyd.ann s!"Decorator {repr d} not supported."
+    -- Check for @icontract.require(lambda ...: cond) or @icontract.ensure(lambda ...: cond)
+    match pyd with
+    | .Call _ (.Attribute _ (.Name _ ⟨_, "icontract"⟩ (.Load _)) ⟨_, attr⟩ (.Load _)) args _ =>
+      if args.val.size ≥ 1 then
+        match args.val[0]! with
+        | .Lambda _ _lamArgs lamBody =>
```
joehendrix, line 1230: The lambda args (`_lamArgs`) are discarded here. The code extracts only the lambda body and passes it to `transAssertExpr`, which preserves variable names symbolically (`.Name "x"` → `.var "x"`). This happens to produce correct output when the lambda param names match the function params, but nothing enforces that.

Two things to address:

**(a) Validate lambda params against function params.** `extractSubject` (which `transAssertExpr` uses) accepts any `.Name` node and converts it to `.var` without any scope checking. So `lambda xx: yy >= 0` on `def f(x: int)` silently produces `.intGe (.var "yy") (.intLit 0)` — a well-formed `SpecExpr` with a meaningless variable and no warning. The lambda param names should be checked to be a subset of the function's argument names (plus `result` for `@ensure`), with a `specWarning` on mismatch.

**(b) Support `result` in `@ensure`.** icontract `@ensure` lambdas receive a `result` parameter for the return value:
```python
@icontract.ensure(lambda result: result >= 0)
@icontract.ensure(lambda x, result: result > x)
```
Since `_lamArgs` is discarded, `result` passes through as `.var "result"` with no special meaning. Please add support for binding `result` so postconditions can reference the return value.

Strata/Languages/Python/Specs.lean
```diff
@@ -1218,14 +1218,34 @@ def pySpecFunctionArgs (fnLoc : SourceRange)
...
-    let (success, d) ← runChecked <| pySpecValue pyd
-    if success then
-      match d with
-      | .typingOverload =>
-        overload := true
-      | _ =>
-        specError pyd.ann s!"Decorator {repr d} not supported."
+    -- Check for @icontract.require(lambda ...: cond) or @icontract.ensure(lambda ...: cond)
+    match pyd with
+    | .Call _ (.Attribute _ (.Name _ ⟨_, "icontract"⟩ (.Load _)) ⟨_, attr⟩ (.Load _)) args _ =>
```
joehendrix, line 1227: The keyword arguments to the `.Call` are matched with `_`. icontract supports several keyword arguments that affect behavior:

```python
# Currently supported — bare positional lambda:
@icontract.require(lambda x: x >= 0)
def f(x: int) -> int: ...

# NOT supported — condition as keyword arg (silently skipped):
@icontract.require(condition=lambda x: x >= 0)
def f(x: int) -> int: ...

# NOT supported — enabled=False still extracted as a contract:
@icontract.require(lambda x: x >= 0, enabled=False)
def f(x: int) -> int: ...

# Ignored — description keyword discarded, toMessage generates one instead:
@icontract.require(lambda x: x >= 0, description="x must be non-negative")
def f(x: int) -> int: ...
```

Worth either handling `condition=` as a fallback for the lambda lookup and checking `enabled=`, or emitting a warning when keyword arguments are present so users know they're being ignored.

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -335,6 +335,29 @@ inductive SpecExpr where
...
+    | .getIndex s f _ => s!"{go 0 s}[{f}]"
+    | .len e _ => s!"len({go 0 e})"
+    | .not e _ => s!"not ({go 0 e})"
+    | .intGe a b _ => paren prec 10 s!"{go 11 a} >= {go 11 b}"
+    | .intLe a b _ => paren prec 10 s!"{go 11 a} <= {go 11 b}"
+    | .implies c b _ => paren prec 5 s!"{go 6 c} ==> {go 5 b}"
+    | .containsKey _ k _ => s!"\"{k}\" in <container>"
+    | .enumMember s vals _ => s!"{go 0 s} in {vals}"
+    | .regexMatch s p _ => s!"re.search(\"{p}\", {go 0 s})"
+    | _ => "<expr>"
```
joehendrix, line 357: A few `SpecExpr` variants that `transAssertExpr` can produce are missing here and fall through to `"<expr>"`:
- `.floatGe` / `.floatLe` — these mirror `.intGe` / `.intLe` and would be trivial to add
- `.isInstanceOf` — common in icontract preconditions
- `.forallList` / `.forallDict`

At minimum the float cases seem like they should match the int handling:
```suggestion
    | .intGe a b _ => paren prec 10 s!"{go 11 a} >= {go 11 b}"
    | .intLe a b _ => paren prec 10 s!"{go 11 a} <= {go 11 b}"
    | .floatGe a b _ => paren prec 10 s!"{go 11 a} >= {go 11 b}"
    | .floatLe a b _ => paren prec 10 s!"{go 11 a} <= {go 11 b}"
```

joehendrix (changes requested):
Nice addition — the approach of collecting icontract decorators separately and then processing them inside `collectAssertions` is clean, and reusing `transAssertExpr` for the lambda bodies is a good call.

A few things I'd like to see addressed before merging:

1. **Lambda parameter validation** — The lambda args (`_lamArgs`) are currently discarded. Since `transAssertExpr` preserves variable names symbolically via `extractSubject`, the output happens to be correct when lambda param names match the function params — but there's no validation that they actually do, and no downstream safety net (`extractSubject` accepts any `.Name` node without scope checking). A typo like `lambda xx: yy >= 0` on `def f(x: int)` silently produces a spec with a meaningless variable `yy` and no warning. Please validate that the lambda parameter names are a subset of the function's parameter names (plus `result` for `@ensure`).

2. **`@ensure` and `result`** — icontract `@ensure` lambdas receive a `result` parameter for the return value (e.g., `lambda result: result >= 0`). Since `_lamArgs` is discarded, `result` has no special meaning. Please add support for binding `result` in postconditions.

3. **Keyword arguments** — The `.Call` keywords are matched with `_`, so keyword-based icontract patterns are silently ignored. Currently only the bare positional-lambda form works:
   ```python
   # Supported:
   @icontract.require(lambda x: x >= 0)

   # NOT supported (silently skipped — lambda is in a keyword arg):
   @icontract.require(condition=lambda x: x >= 0)

   # NOT supported (enabled=False is ignored, contract still extracted):
   @icontract.require(lambda x: x >= 0, enabled=False)

   # NOT supported (description is ignored — but toMessage generates one):
   @icontract.require(lambda x: x >= 0, description="x must be non-negative")
   ```
   Worth either handling `condition=` and `enabled=`, or emitting a warning when keyword arguments are present so users know they're being ignored.

4. **Tests** — This PR needs at least basic test coverage: a Python file with `@icontract.require` / `@icontract.ensure` decorators fed through the translator, plus `#guard` tests for `SpecExpr.toMessage`. The existing `SpecsTest.lean` + `main.py` pattern makes this straightforward to add.

I've left inline comments with more detail.

PR #869, opened 2026-04-13
Concrete Core interpreter

robin-aws:
*Description of changes:*

Adds support for concrete execution of Core programs, via adding a general concept of "running" the various Strata semantic building blocks.

In several cases there are already various `eval` functions on pieces like Lambda expressions, Imperative statements, Procedures and Programs. These functions implement partial or symbolic execution, however. `run` functions should be compatible with the same semantics, but produce one concrete trace through the element rather than multiple/all possible traces. I've generally put the `run` functions for X in the same `XEval.lean` file, however.

With that in hand, also adds a `pyInterpret` top-level command based on converting from Python to Laurel to Core and then interpreting the result. This gives us a simple form of testing to help ensure the translation maintains correct semantics by doing differential testing against a "real" Python interpreter (and similarly for other languages in the future).

The `run_py_interpret.sh` script runs the `Python/tests` programs and expects by default they complete without error (i.e. all assertions dynamically pass). The `expected_interpret` directory contains `.expected` regular expression files for known failures, and `.skip` files for tests that should be skipped entirely for now (in particular a few overflow the Lean stack but only on macos, while they would pass in the linux-based CI)

There are a few known limitations, which I would like us to incrementally improve in follow-up PRs. I'd argue none of them prevent this initial version from being useful to merge and get into CI, to protect against regressions in executability in general.

1. The `findStuckRedex` utility is there because it will be useful for debugging, but I need to add more type parameterization to be able to format the expressions, and this can be improved in the next PR.
2. For elements of nondeterminism in statements and commands, the current `run` functions pick a hard-coded option arbitrarily. I think we should expose an oracle parameter to make these choices, and make it state we thread through computation so it can record and replay choices. But this is also additional scope for a follow-up.
3. About 20% of the tests are known failures, mostly because of stuck expressions due to built-in functions that we didn't provide a `body` or `concreteEval` for yet. Again it's still valuable to lock down the first 80% in the meantime.


Strata/Languages/Core/Env.lean
```diff
@@ -129,6 +129,15 @@ A substitution map from variable identifiers to expressions.
 -/
 @[expose] abbrev SubstMap := Map Expression.Ident Expression.Expr
 
+/-- Controls whether the evaluator operates in verification mode (symbolic,
+    contract-based) or concrete mode (executing bodies, iterating loops). -/
+inductive EvalMode where
```
robin-aws: This approach seems OKAY to me at least. :) In the future if we achieve a denotational interpreter, these two options would just be two different abstract domains in the spectrum of possibilities.

Strata/Languages/Python/PyInterpret.lean
```diff
@@ -0,0 +1,49 @@
...
+      (moreFns := Strata.Python.ReFactory) with
+    | .ok prog => pure prog
+    | .error e => throw s!"Core type checking failed: {e.message}"
+  let E ← match Core.initConcreteEnv core fuel with
+    | .ok e => pure e
+    | .error e => throw s!"init failed: {e.message}"
+  return Core.interpProcedure E "__main__"
+
+/-- Run the Python → Core (direct) → Interpret pipeline on a Python Ion file. -/
+def pyInterpretDirect
```
robin-aws: Is this worth having? If we plan to fully deprecate the direct translation then I would remove this.

Strata/Languages/Python/PyInterpret.lean
robin-aws, line 1: Not convinced this is worth it's own file, especially if we only keep the one definition.

robin-aws:
Another note to self: on #631 I opined that we shouldn't use both "interpreter" as well as "evaluator", and that "interpreter" was a slight better choice. But at the Core level there's already a lot of use of "eval", so I think I'm probably better off not using the term "interpreter".

MikaelMayer:
As a quick feedback, each phase has a model validator that makes "sat" become "unknown" if e.g. a procedure call was eliminated. Could you please look into validating the model using your concrete core interpreter?
Ideally we would be able with your interpreter to return whether the example provided by SMT is sat on this one, currently it should return "unknown"

```
procedure callee(i: int) returns (j: int)
{
   // j not initialized 
}
procedure caller(k: int) {
   var k;
   k := callee(2);
   assert k == 2; // currently returns "unknown" because of callee, but could return "fail" if we validate the counter-example
}
```

StrataTest/Languages/Python/run_py_interpret.sh
```diff
@@ -0,0 +1,140 @@
+#!/bin/bash
```
keyboardDrummer, line 1: Can this be a Lean test that's part of `lake test` ? Any standalone testing script that has a separate ci.yaml has a chance of being skipped during local development, increasing the amount of issues that are only found in CI and slowing down development.

robin-aws: Oh agreed that would definitely be better. Would it be alright to do that refactoring in a follow up PR? There are already similar non-Lean scripts for testing the SMT and CBMC backends that should also be refactored, and I would want to take those on too.

robin-aws: #914

aqjune-aws:
I looked at Strata/Languages/Core/Interpreter.lean of https://github.com/strata-org/Strata/pull/869/changes/b27463e87a0475a0dee4e7aa778cf4792a77fe8c and I liked it (attached https://gist.github.com/aqjune-aws/cf7ff34551d961e15c3e64614784c418). It is already invoking Lambda's partial evaluator internally and does not keep its own expression evaluator, which seems to be a fair design choice.

The main question would be whether we should merge it with the existing partial evaluator of Imperative/Core (or possibly be a symbolic simulator in the future) or keep a separate Interpreter.lean because it similarly does evaluation as well.

I think having a separate Interpreter.lean makes sense. The main reason is that we will want the Core interpreter to be highly optimized & its execution is as fast as possible. If the symbolic simulator was reused to implement the interpreter, any performance related improvements will meet additional burden due to generalizing it to symbolic simulation.

The second minor reason is that there will be parts of interpreter that will not be used by symbolic simulation. For example, visiting the body of a procedure call: symbolic simulation relies on the fact that calls are already properly inlined if it wants to visit the body, carefully controlled by the inlining pass. However, users of concrete interpreter will expect that interpreter visits the concrete body of a procedure. Therefore, this interpreter implementation will not be useful in simulator.

robin-aws:
Thanks @aqjune-aws, I think the point about performance is pretty important and enough reason to separate the interpreter from the partial evaluation. I've found that the pass rate on the separate version is lower, but I've realized that's because the version here is incorrectly deferring proof obligations from `assert` statements etc. and never checking them at all. I had to argue with Kiro a fair bit to explain why that wasn't sound. :)

I'm working on restoring the independent interpreter and addressing some of the easiest evaluation gaps before I ask for another review.

Strata/Languages/Core/Interpreter.lean
```diff
@@ -0,0 +1,375 @@
...
+/-! ## Helpers -/
+
+/-- Default value based on an optional monotype from the store. -/
+private def defaultValueOfMonoTy (ty : Option LMonoTy) : Expression.Expr :=
+  match ty with
+  | some (.tcons "int" _) => .intConst () 0
+  | some (.tcons "bool" _) => .boolConst () false
+  | some (.tcons "string" _) => .const () (.strConst "")
+  | some (.tcons "real" _) => .const () (.realConst 0)
+  | _ => .intConst () 0
```
aqjune-aws: What do you think about updating this function to throw "Unknown type: {ty}" error when `ty` doesn't match any of the above? You can use the `Except` return type like this:

```
private def defaultValueOfMonoTy (ty : Option LMonoTy) : Except String Expression.Expr :=
  match ty with
  | some LMonoTy.int => return .intConst () 0
  | some LMonoTy.bool => return .boolConst () false
  | some LMonoTy.string => return .const () (.strConst "")
  | some LMonoTy.real => return .const () (.realConst 0)
  | _ => throw s!"Unknown type: {ty}"
```
Then it is possible to guarantee that any successful invocation of `defaultValueOfMonoTy` yields exactly a value of the `ty` type (currently this doesn't hold because the placeholder value is `.intConst () 0`).

Strata/Languages/Core/Interpreter.lean
```diff
@@ -0,0 +1,375 @@
...
+
+/-- Produce a default value for a type. Used for havoc and uninitialized variables. -/
+private def defaultValue (ty : Expression.Ty) : Expression.Expr :=
+  if h : ty.isMonoType then
+    defaultValueOfMonoTy (ty.toMonoType h)
+  else .intConst () 0
```
aqjune-aws: This can also use `Except` to throw when the type isn't MonoType:
```
private def defaultValue (ty : Expression.Ty) : Except String Expression.Expr :=
  if h : ty.isMonoType then
    defaultValueOfMonoTy (ty.toMonoType h)
  else throw s!"Not a MonoType: {ty}"
```

Strata/Languages/Core/Interpreter.lean
```diff
@@ -0,0 +1,375 @@
...
+    why execution stopped. -/
+inductive StepResult where
+  /-- Normal completion (although E may have .error set). -/
+  | normal (E : Env)
+  /-- An `exit` statement propagating upward. -/
+  -- TODO: would it be better to add this state to Env instead, similar to .error?
+  -- That would make this datatype unnecessary, and it would likely
+  -- help define the semantics for .exit as well.
+  | exiting (label : Option String) (E : Env)
+  deriving Inhabited
```
aqjune-aws: Do you want to use the `CoreConfig` data type in Strata/Languages/Core/StatementSemantics.lean (which is instantiation of `Config` in Strata/DL/Imperative/StmtSemantics.lean)? It keeps the stream of "next statements to come", and also has `.terminal` status which states that there is no more statement to execute.
A benefit of using this data type is that
(1) this is exactly the program state data structure that is used in the small-step semantics of Imperative. So proving the consistency between symbolic simulation and the new interpreter becomes much easier.
(2) It has a few helper functions like `Config.exitsCoveredByBlocks` that can be useful and also already has proven good properties.

Strata/Languages/Core/Interpreter.lean
```diff
@@ -0,0 +1,375 @@
...
+function so that we can later prove it consistent with the small-step
+`Lambda.Step` relation from `Strata.DL.Lambda.Semantics`.
+
+Currently delegates to `LExpr.eval` with the fuel and state from `Env`.
+If the result contains a stuck redex (a fully-applied function that should
+have reduced but didn't), returns an error.
+-/
+def interpExpr (E : Env) (e : Expression.Expr) : Except Env Expression.Expr :=
+  let v := e.eval E.exprEnv.config.fuel E.exprEnv
+  match findStuckRedex E.factory v with
```
aqjune-aws: I think we can use this function to check whether the evaluator couldn't fully reduce `e` to a value:
```
-- Strata/DL/Lambda/LExprEval.lean
def isCanonicalValue (F : @Factory T.base) (e : LExpr T) : Bool :=
```
Then `interpExpr` can be refactored like this:
```
def interpExpr (E : Env) (e : Expression.Expr) : Except Env Expression.Expr :=
  let v := e.eval E.exprEnv.config.fuel E.exprEnv
  if LExpr.isCanonicalValue E.factory v then
    .ok v
  else
    match findStuckRedex E.factory v with
    | some stuckExpr => .error (stuck E s!"expression contains stuck redex: {format stuckExpr}")
    | none => .error (stuck E s!"partial evaluator stuck, but couldn't find the redex")
```

Strata/Languages/Core/Interpreter.lean
```diff
@@ -0,0 +1,375 @@
...
+  | 0 => .normal { E with error := some EvalError.OutOfFuel }
+  | fuel + 1 =>
+  match c with
+  | .cmd (.init x ty e _md) =>
+    match e with
+    | .det expr =>
+      match interpExpr E expr with
+      | .ok v => .normal (CmdEval.update E x ty v)
+      | .error sr => .normal sr
+    | .nondet => .normal (CmdEval.update E x ty (defaultValue ty))
```
aqjune-aws: You will want to consider the case when `defaultValue ty` failed when `ty` is not a type familiar to us.

Strata/Languages/Core/Interpreter.lean
```diff
@@ -0,0 +1,375 @@
...
+          let formalBindings : List (CoreIdent × (Option LMonoTy × Expression.Expr)) :=
+            proc.header.inputs.keys.zip proc.header.inputs.values |>.zip argVals
+            |>.map fun ((name, ty), val) => (name, (some ty, val))
+          let outputBindings : List (CoreIdent × (Option LMonoTy × Expression.Expr)) :=
+            proc.header.outputs.keys.zip proc.header.outputs.values
+            |>.map fun (name, ty) => (name, (some ty, LExpr.fvar () name none))
+          let callEnv := { E with
+            exprEnv := { E.exprEnv with
+              state := E.exprEnv.state.push (formalBindings ++ outputBindings) } }
+          match interpBlock fuel callEnv proc.body with
```
aqjune-aws: I guess this is a big-step semantics since it continues execution at the body of the block. What about making it clear at the comment at `interpCmd` that if the command is `call`, its body is executed until the fuel is exhausted?

StrataTest/Languages/Core/Tests/ProgramEvalTests.lean
```diff
@@ -436,6 +437,206 @@ Proof Obligation:
...
+program Core;
+procedure Test() returns (y : int)
+{
+  y := 42;
+};
+#end
+
+/-- info: y = (some 42) -/
+#guard_msgs in
+#eval! runProc simplePgm "Test"
```
aqjune-aws: This is cool. :)

StrataTest/Languages/Core/Tests/ProcedureEvalTests.lean
```diff
@@ -436,6 +437,206 @@ Proof Obligation:
...
+      | none => IO.println "success (no proc)"
+      | some p =>
+        let outputs := p.header.outputs.keys.map fun name =>
+          let val := (E.exprEnv.state.find? name).map (·.snd)
+          s!"{name} = {val.map (fun v => toString (format v))}"
+        IO.println (String.intercalate ", " outputs)
+    | .assertionFailure label _ _ => IO.println s!"assertion failure: {label}"
+    | .error msg => IO.println s!"error: {msg}"
+    | .fuelExhausted => IO.println "fuel exhausted"
+    | .stuck msg => IO.println s!"stuck: {msg}"
```
aqjune-aws: I couldn't find the `fuelExhausted` constructor. Where is it defined?

Strata/Languages/Core/Interpreter.lean
```diff
@@ -0,0 +1,375 @@
...
+      | .ok E' => E'
+      | .error _ => E
+    | .recFuncBlock fs _md =>
+      fs.foldl (fun E f =>
+        match E.addFactoryFunc f with
+        | .ok E' => E'
+        | .error _ => E) E
+    | .ax a _md =>
+      { E with pathConditions := E.pathConditions.addInNewest [(toString a.name, a.e)] }
+    | _ => E
```
aqjune-aws: It seems there is no `.proc`. Is this going to be processed by `interpProcedure`?

At the moment Core's execution is declaration order sensitive in the sense that the functions/axioms that were defined before a procedure are not visible to the procedure, to be conservative (as why Lean can't see the definitions located after).

aqjune-aws (commented):
I left a few comments to the pull request.

I think that it would be ideal if we can get a 'capture' of a program state when the fuel was exhausted because it will allow us to do debugger-like interface (give fuel = 1, do one-step execution, get the next program state, etc). Is there a plan for the interpreter to have that feature?

Strata/Languages/Core/Interpreter.lean
robin-aws: Completely agree - I'll make that change after I convert the bodies of the `interp*` functions to use `do` blocks so it's simpler to account for the new failure case.

Strata/Languages/Core/Interpreter.lean
robin-aws: YES, I found that data type just recently and had the same thought. And that would also give us the "captured state" on exhausting fuel you asked for elsewhere. It may end up a bit slower than the more direct evaluation approach here, but I don't think that's a significant enough downside - performance will be more important for the core interpreter, but not more important than proving it is consistent with the small step semantics, and using the same data structures will make that much easier.

Strata/Languages/Core/Interpreter.lean
robin-aws: I actually had `isCanonicalValue` in `interpExpr` originally and did not notice that Kiro removed it :)

StrataTest/Languages/Core/Tests/ProcedureEvalTests.lean
robin-aws: Ah I hadn't gotten to fixing this file after recent refactorings, will fix.

Strata/DL/Imperative/CmdEval.lean
```diff
@@ -113,6 +113,87 @@ def Cmds.eval [BEq P.Ident] [EvalContext P S] (σ : S) (cs : Cmds P) : Cmds P ×
...
+        | .det expr => EC.updateError σ (.AssignVarNotExists x expr)
+        | .nondet => EC.updateError σ (.HavocVarNotExists x)
+      | some (_xv, xty) =>
+        match e with
+        | .det expr =>
+          let (expr, σ) := EC.preprocess σ c expr
+          let expr := EC.eval σ expr
+          EC.update σ x xty expr
+        | .nondet =>
+          --
```
MikaelMayer: Nit: orphan `--` comment. Either add a description (e.g. what `nondet` means for a set) or remove it.

Strata/DL/Lambda/LExprEval.lean
```diff
@@ -316,6 +316,65 @@ end
...
+def run (σ : LState TBase) (e : (LExpr TBase.mono)) : Except String (LExpr TBase.mono) :=
+  let v := e.eval σ.config.fuel σ
+  if LExpr.isCanonicalValue σ.config.factory v then
+    .ok v
+  else
+    match findStuckRedex σ.config.factory v with
+    | some _ => .error "expression contains stuck redex"
+    | none => .ok v
+
+-- TODO: foldlM?
```
MikaelMayer: This is `List.mapM` (or `es.mapM (run σ)`), not `foldlM`. Either replace the body with `es.mapM (run σ)` or update the TODO.

Strata/Languages/Core/ProgramEval.lean
```diff
@@ -64,6 +72,38 @@ def eval (E : Env) : Except Strata.DiagnosticModel (List Env × Statistics) :=
...
+  let E <- E.addDatatypes datatypes
+  return prog.decls.foldl (fun E decl =>
+    match E.error with
+    | some _ => E
+    | none =>
+    match decl with
+    | .func f _md =>
+      match E.addFactoryFunc f with
+      | .ok E' => E'
+      | .error _ => E
```
MikaelMayer: Silently swallowing `addFactoryFunc` errors means a program with a malformed function declaration will run with that function missing, producing confusing "stuck redex" or "procedure not found" errors downstream. Consider propagating the error (this function already returns `Except`), or at minimum setting `E.error` so execution stops early with a clear message.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -749,6 +752,81 @@ def evalOne (E : Env) (old_var_subst : SubstMap) (ss : Statements) : Env :=
...
+                isConstr := decl.isConstr
+                inputs := decl.inputs.map (fun (id, ty) => (id, Lambda.LTy.toMonoTypeUnsafe ty))
+                output := Lambda.LTy.toMonoTypeUnsafe decl.output
+                body := decl.body
+                attr := decl.attr
+                concreteEval := decl.concreteEval
+                axioms := decl.axioms
+              } with
+              | .ok E' => E'
+              | .error _ => E
```
MikaelMayer: Same concern as in `Program.run`: if `addFactoryFunc` fails for an inline function declaration, the error is silently dropped and execution continues without the function. This will surface later as a confusing stuck-expression error. Consider `Env.stuck E s!"addFactoryFunc failed: {err}"` in the error branch.

StrataTest/Languages/Python/run_py_interpret.sh
```diff
@@ -0,0 +1,142 @@
...
+fuel=""
+direct=""
+verbose=0
+update=0
+
+while [ $# -gt 0 ]; do
+    case "$1" in
+        --filter) filter="$2"; shift ;;
+        --fuel) fuel="--fuel $2"; shift ;;
+        --direct) direct="--direct" ;;
```
MikaelMayer: `--direct` is accepted here and forwarded to `strata pyInterpret`, but `pyInterpretCommand` in `StrataMain.lean` does not define a `--direct` flag. Passing it will either be silently ignored or cause a CLI error. Remove it from the script until the command supports it, or add the flag to `pyInterpretCommand`.

Strata/DL/Imperative/CmdEval.lean
```diff
@@ -113,6 +113,87 @@ def Cmds.eval [BEq P.Ident] [EvalContext P S] (σ : S) (cs : Cmds P) : Cmds P ×
...
+        σ
+      | some false =>
+        EC.updateError σ (.Misc f!"assume ({label}) condition is false")
+      | none =>
+        EC.updateError σ (.Misc f!"assume ({label}) condition did not reduce to bool")
+
+    | .cover _ _ _ =>
+      -- In the future we can record when a cover is true
+      -- and assert it was hit at least once later on
+      EC.updateError σ (.Misc s!"cover is not yet supported")
```
MikaelMayer, line 192: Nit: `cover` raises an error unconditionally, which means any program containing a `cover` statement will fail during concrete execution even if the cover is never semantically relevant. A no-op (just returning `σ`) would be more consistent with the comment's intent of recording coverage later.

Strata/Languages/Core/Interpreter.lean
robin-aws: This is still a valid point in the refactored version as well AFAICT. But I believe the current implementation of extracting all the function definitions ahead of time and making them available to all procedures is strictly more permissive and doesn't allow any incorrect behavior, and I'm not giving axioms any runtime behavior at all.

Strata/DL/Imperative/CmdEval.lean
shigoel: We can also add logging capabilities in the evaluation context -- right now, we record just errors and warnings.

Strata/DL/Imperative/CmdEval.lean
```diff
@@ -113,6 +113,87 @@ def Cmds.eval [BEq P.Ident] [EvalContext P S] (σ : S) (cs : Cmds P) : Cmds P ×
     let (crest, σ) := Cmds.eval σ crest
     (c :: crest, σ)
 
+--------------------------------------------------------------------
+
+def stuck {P S} [EC : EvalContext P S] (σ : S) (message : String) : S :=
+  EC.updateError σ (.Misc message)
+
+/--
+Concrete execution for an Imperative Command.
```
shigoel, line 122: `run` and `eval` share nearly identical logic for `init` and `set`, and there's a possibility that they can share more if we support more commands. Could the common parts be factored out? If there's a reason to keep them separate (e.g. to allow them to diverge later, or to avoid complicating the types), a comment explaining the relationship would help so that we can keep them in sync: e.g., eval produces rewritten commands + proof obligations for symbolic execution, while run produces only a new state for concrete single-trace execution.

Strata/DL/Imperative/StmtEval.lean
```diff
@@ -0,0 +1,115 @@
...
+  /-- Pop a variable scope (for blocks). -/
+  popScope : S → S := id
+
+def runStep [BEq P.Expr] [HasBool P]
+    (ops : RunOps P CmdT S)
+    (c : RunConfig P CmdT S) : RunConfig P CmdT S :=
+  match c with
+  | .terminal _ => c
+  | .exiting _ _ => c
+
```
shigoel, line 53: The generic `runStep` returns `c` unchanged when `evalExpr` returns `none` (here and in the `loop` case on line 74). `runStmt` will then call `runStep` repeatedly on the same frozen config, draining all fuel without progress. The caller ultimately gets "failed to terminate" instead of a message pointing at the stuck expression.

The Core instantiation always wraps in `some`, so this is not triggered today, but a future instantiation that returns `none` for irreducible expressions would hit it. Worth either (a) documenting this as a known limitation in the docstring, or (b) returning `.terminal` with an error state when `evalExpr` yields `none`.

StrataMain.lean
```diff
@@ -1307,6 +1309,47 @@ def verifyCommand : Command where
         println! f!"Finished with {provedGoalCount} goals passed, {failedGoalCount} failed."
         IO.Process.exit ExitCode.failuresFound
 
+def pyInterpretCommand : Command where
+  name := "pyInterpret"
+  args := [ "file" ]
+  flags := [{ name := "fuel", help := "Maximum execution steps.", takesArg := .arg "n" },
+            { name := "verbose", help := "Show the generated Core program." }]
```
shigoel: `String.toNat!` panics on non-numeric input (e.g. `--fuel abc`). Consider using `s.toNat?` with a user-facing error or a default fallback.

Strata/Languages/Core/ProgramEval.lean
```diff
@@ -64,6 +72,38 @@ def eval (E : Env) : Except Strata.DiagnosticModel (List Env × Statistics) :=
 
 --------------------------------------------------------------------
```
shigoel, line 76: When `addFactoryFunc` returns `.error`, the function is silently skipped (same on line 101 for `recFuncBlock`). Since this is setting up the environment for concrete execution, a silently missing function will surface later as a confusing "stuck redex" or "procedure not found" error far from the root cause. Consider propagating the error or at least logging it.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -749,6 +752,81 @@ def evalOne (E : Env) (old_var_subst : SubstMap) (ss : Statements) : Env :=
...
+          match configAfter with
+          | .terminal callEnv' =>
+            match callEnv'.error with
+            | some _ => { E with error := callEnv'.error }
+            | none =>
+              let outputVals := proc.header.outputs.keys.map fun name =>
+                (callEnv'.exprEnv.state.findD name (none, LExpr.fvar () name none)).snd
+              lhs.zip outputVals |>.foldl (fun env (name, val) =>
+                env.insertInContext (name, none) val) E
+          | _ => Env.stuck E "failed to terminate"
```
shigoel: This is one of the most complex functions in the PR. An important invariant to preserve: **caller isolation** — caller-local variables must not be visible to the callee, and callee-local variables must not leak back into the caller. The globals-only + formals/outputs construction on lines 781–784 looks correct for this, but it would be valuable to state this invariant as a prose comment on the function so future modifications know what to protect.

shigoel (commented):
Overall this is a well-structured PR — the layered design (Lambda.run → Cmd.run → runStep/runStmt → Command.run/runCall) is clean and the test coverage is solid. A few items below.

Strata/DL/Imperative/CmdEval.lean
```diff
@@ -113,6 +113,87 @@ def Cmds.eval [BEq P.Ident] [EvalContext P S] (σ : S) (cs : Cmds P) : Cmds P ×
     let (crest, σ) := Cmds.eval σ crest
     (c :: crest, σ)
 
+--------------------------------------------------------------------
+
+def stuck {P S} [EC : EvalContext P S] (σ : S) (message : String) : S :=
```
shigoel, line 118: This is a confusing name because I see it's being used to set errors for non-stuck runs -- I don't think of `Env.stuck E s!"procedure '{procName}' not found"`
as a stuck run though (LMK if you disagree).

shigoel: A follow-up is: one could use the pre-existing `updateError` here, right?

Strata/DL/Lambda/LExprEval.lean
```diff
@@ -316,6 +316,65 @@ end
 instance : Traceable EvalProvenance Unit where
   combine _ := ()
 
+
+/-! ## Expression Evaluation -/
+
+/-- Walk a post-eval expression looking for a stuck redex: a fully-applied
+non-constructor factory function whose arguments are all canonical values.
+Such a call *should* have reduced during `eval` but didn't (e.g. missing body
```
shigoel: FWIW, I think there will always be functions for which we have neither body nor concrete eval.

Strata/Languages/Core/ProgramEval.lean
```diff
@@ -64,6 +72,38 @@ def eval (E : Env) : Except Strata.DiagnosticModel (List Env × Statistics) :=
...
+-/
+def run (prog : Program) : Except DiagnosticModel Env := do
+  let factory ← Core.Factory.addFactory Lambda.Factory.default
+  let datatypes := prog.decls.filterMap fun decl =>
+    match decl with
+    | .type (.data d) _ => some d
+    | _ => none
+  let σ ← Lambda.LState.init.addFactory factory
+  let E: Env := { Env.init with exprEnv := σ, program := prog }
+  let E <- E.addDatatypes datatypes
```
shigoel: Nit: Oof, `<-` alert :) I've gotten used to the pretty `←`

StrataTest/Languages/Core/Tests/ProgramEvalTests.lean
```diff
@@ -411,6 +412,205 @@ Proof Obligation:
...
+    y := x;
+  } else {
+    y := 0 - x;
+  }
+};
+#end
+
+/-- info: y = (some 7) -/
+#guard_msgs in
+#eval! runProc itePgm "Test" [.intConst () 7]
```
shigoel: Why do we need `eval!` instead of `eval`?

StrataMain.lean
```diff
@@ -1307,6 +1309,47 @@ def verifyCommand : Command where
...
+  help := "Interpret a Python Ion program concretely (Python → Laurel → Core → execute)."
+  callback := fun v pflags => do
+    let filePath := v[0]
+    let verbose := pflags.getBool "verbose"
+    let fuel := match pflags.getString "fuel" with
+      | some s => s.toNat!
+      | none => 10000
+
+    let (core, _diags) ←
+      match <- pyTranslateLaurel filePath #[] #[] (specDir := ".") |>.toBaseIO with
```
shigoel: Curious why seeing the resulting Laurel won't be helpful too.

StrataTest/Languages/Core/Tests/ProgramEvalTests.lean
shigoel: +1

Strata/DL/Lambda/LExprEval.lean
robin-aws: Aha thanks, I knew there had to be a nice monadic utility but hadn't found it yet :)

robin-aws: I removed this entirely as it's only used in one place and `mapM` is just as clear IMO.

Strata/Languages/Core/ProgramEval.lean
robin-aws: We should augment our `lake lint` script to enforce coding style points like this!

Strata/Languages/Core/ProgramEval.lean
robin-aws: Oof good catch, I'm annoyed Kiro thought this was fine. :P I introduced a `Decl.run` function and that addressed this and made `Program.run` much cleaner.

Strata/Languages/Core/ProgramEval.lean
robin-aws: Agreed, see reply below

Strata/Languages/Core/Interpreter.lean
robin-aws: Refactored to introduce a `Decl.run` and just process them in order, which is much cleaner and likely more correct.

Strata/Languages/Core/StatementEval.lean
robin-aws: See above

StrataTest/Languages/Python/run_py_interpret.sh
robin-aws: Yup that's left over from a prior version that did support this, but then it became clearer that we've deprecated that path so I partially took it out. Removed.

Strata/DL/Imperative/CmdEval.lean
robin-aws: Great point! I'm not personally confident in what the runtime semantics of `cover` should be yet though, and if they show up in any of the Python test cases we also want to execute with `python` then something's gone wrong, so I prefer to leave this as unsupported for the very first version if that's alright.

Strata/DL/Imperative/CmdEval.lean
robin-aws: It's expected to diverge further, so I added a comment (and I may add a more top-level one if I find a good place for it, since this is true of all the `run` functions)

Strata/DL/Imperative/StmtEval.lean
robin-aws: Good catch! `RunOps` should have a hook for adding an error to the `S` type. I've added that. It still doesn't include the actual stuck expression but I still want to add that throughout in a follow up PR.

Strata/DL/Imperative/CmdEval.lean
shigoel: My first thought is that runtime semantics of cover should just be skip, but yeah, I'm fine with punting for now.

StrataTest/Languages/Core/Tests/ProgramEvalTests.lean
robin-aws: Looks like we don't, Kiro just seems to assume you do by default. :) Fixed!

Strata/DL/Imperative/CmdEval.lean
aqjune-aws: Actually, I would vote for removing the .cover command - it isn't being seriously tested and people might depend on a problematic feature. @shigoel @MikaelMayer

Strata/DL/Imperative/StmtEval.lean
```diff
@@ -0,0 +1,115 @@
...
+/-! ## Generic small-step evaluator
+
+`RunConfig` mirrors `Config` but is parameterized by a generic state type `S`
+instead of `Env P`, so that concrete interpreters (e.g. Core) can thread their
+own rich environment through the stepper.
+-/
+
+/-- Configuration for the generic stepper, parameterized by state `S`. -/
+inductive RunConfig (P : PureExpr) (CmdT : Type) (S : Type) where
```
aqjune-aws, line 22: It seems this can conceptually be a pair type `Config P CmdT × List S` (where `List S` is a stack of `S`) :) Then we can totally reuse Config!

Strata/DL/Lambda/LExprEval.lean
aqjune-aws: Yes.... They will make `LExpr.eval` stuck. For the testing purpose, it would be good to gradually add the body or concrete eval - executable Regex or String operations might be helpful...

Strata/Languages/Core/ProgramEval.lean
aqjune-aws: The lines above will have been replaced packaged into `buildEnv`. Now one liner :)

StrataMain.lean
robin-aws: Agreed and I don't actually think "verbose" is the right flag. I replaced it with `--keep-all-files` since `pyAnalyzeLaurel` already has the same flag, and made it keep both Laurel and Core.

Strata/DL/Lambda/LExprEval.lean
robin-aws: I imagine so yes. But we'll also want to try to make sure that the executable parts of programs translated from front-end languages don't depend on them, as much as possible.

shigoel: > executable Regex or String operations might be helpful

FWIW, executable regex ops -- nontrivial :) Doable, yes.

Strata/Languages/Core/ProgramEval.lean
robin-aws: Nice! From https://github.com/strata-org/Strata/pull/753 you mean? I'll try to remember to replace it once that's merged.

Strata/DL/Imperative/CmdEval.lean
robin-aws: Ah I didn't spot `updateError`. `stuck` is only slightly more convenient because it adds the `.Misc` constructor for you. But I agree "stuck" is the wrong word here. I just removed this and used `E.updateError E (.Misc msg)` in a few places.

Independently we may want to add more `EvalError` cases to cover things like missing procedure definitions, but again would love to leave that to the next PR. :)

Strata/Languages/Core/StatementEval.lean
robin-aws: Good call, I added a comment. I also tried to chop up the function into smaller pieces since that would help the complexity, but it's tricky because of the `mutual` block. Might attempt it again in the next PR :)

Strata/DL/Lambda/LExprEval.lean
robin-aws: Oh very true. But I mean more if we translated say the implementation of Python `re.compile` itself, that would still produce executable Core.

Strata/DL/Imperative/StmtEval.lean
robin-aws: Totally! I'll see if I can replace it entirely in the next PR, or if it's still valuable to have this extra abstraction, if that's alright.

Strata/DL/Imperative/CmdEval.lean
shigoel: Agreed, yes. Let's punt.

Strata/DL/Imperative/StmtEval.lean
```diff
@@ -0,0 +1,117 @@
...
+  pushScope : S → S := id
+  /-- Pop a variable scope (for blocks). -/
+  popScope : S → S := id
+  /-- Produce a new state with an error -/
+  addError : S → String -> S
+  /-- Check if the state has an error (to short-circuit execution). -/
+  hasError : S → Bool := fun _ => false
+
+def runStep [BEq P.Expr] [HasBool P]
+    (ops : RunOps P CmdT S)
```
shigoel, line 48: Consider stating a simulation invariant: for deterministic cases, `runStep ops c = c'` should correspond to a `StepStmt` step on the matching `Config`. The structural correspondence with `StepStmt` in `StmtSemantics.lean` is clear (same case structure), but formalizing it would anchor correctness and catch drift if either side is modified independently.

Also, the hard-coded nondeterminism choices (ITE → then-branch, loop → exit) are acknowledged in the PR description but not in the code. A brief comment on each `.nondet` case documenting the choice (e.g. "Concrete execution picks the then-branch; see RunOps oracle TODO") would help readers who don't have the PR description at hand.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -749,6 +752,86 @@ def evalOne (E : Env) (old_var_subst : SubstMap) (ss : Statements) : Env :=
...
+
+Importantly, this creates a separate Env to execute the body of the procedure with,
+which initially only contains globals and the input/output variables.
+The resulting Env is the original passed in Env with the output variables copied back into it.
+-/
+def Command.runCall (lhs : List Expression.Ident) (procName : String) (args : List Expression.Expr)
+    (fuel : Nat) (E : Env) : Env :=
+    match fuel with
+    | 0 => { E with error := some .OutOfFuel }
+    | fuel' + 1 =>
```
shigoel, line 770: `List.zip` silently truncates when its arguments have different lengths. If `argVals.length != proc.header.inputs.keys.length` (e.g. from a programmatic caller like the `runProc` test helper passing the wrong number of args), some formals will silently get no binding, leading to confusing stuck-expression failures downstream.

Consider adding an explicit arity check before the zip:
```lean
if argVals.length != proc.header.inputs.keys.length then
  CmdEval.updateError E (.Misc s!"procedure '{procName}': expected {proc.header.inputs.keys.length} arguments, got {argVals.length}")
else ...
```

The same concern applies to the `lhs.zip outputVals` below (line ~470), though both are likely safe for type-checked programs.

shigoel (commented):
Good PR — clean architecture with RunConfig/RunOps abstracting the stepper, solid test coverage. Two suggestions below.

shigoel (dismissed):

Strata/DL/Imperative/StmtEval.lean
shigoel: Can do in a follow-up, if you like.

Strata/Languages/Core/StatementEval.lean
shigoel: preference for this to be done in this PR, but okay if you prefer a follow-up (since you're planning one anyway!).

Strata/DL/Imperative/CmdEval.lean
```diff
@@ -113,6 +113,90 @@ def Cmds.eval [BEq P.Ident] [EvalContext P S] (σ : S) (cs : Cmds P) : Cmds P ×
...
+    | .init x ty e _ =>
+      match EC.lookup σ x with
+      | none =>
+        match e with
+        | .det expr =>
+          let (expr, σ) := EC.preprocess σ c expr
+          let expr := EC.eval σ expr
+          EC.update σ x ty expr
+        | .nondet =>
+          -- Unconstrained initialization - generate a fresh value
```
MikaelMayer, line 142: I don't expect from an interpreter that unconstrained initialization generates a free variable. I would have imagined "a default unreadable value" although I was even imagining a concrete default value.
I'm ok with that if reading a free variable leads to an error though since the interpreter is not supposed to be a partial evaluator. would you agree with that?

Strata/Languages/Core/StatementEval.lean
```diff
@@ -749,6 +752,86 @@ def evalOne (E : Env) (old_var_subst : SubstMap) (ss : Statements) : Env :=
...
 
+---------------------------------------------------------------------
+
+mutual
+
+/--
+Interpret a single procedure call.
+
+Importantly, this creates a separate Env to execute the body of the procedure with,
+which initially only contains globals and the input/output variables.
```
MikaelMayer: We don't have globals anymore

StrataTest/Languages/Core/Tests/ProgramEvalTests.lean
```diff
@@ -409,6 +410,205 @@ Proof Obligation:
...
+};
+#end
+
+/-- info: y = (some 7) -/
+#guard_msgs in
+#eval runProc itePgm "Test" [.intConst () 7]
+
+/-- info: y = (some 3) -/
+#guard_msgs in
+#eval runProc itePgm "Test" [.intConst () (Int.negSucc 2)]
```
MikaelMayer:
```suggestion
#eval runProc itePgm "Test" [.intConst () (-3)]
```

StrataTest/Languages/Core/Tests/ProgramEvalTests.lean
```diff
@@ -409,6 +410,205 @@ Proof Obligation:
...
+  {
+    sum := sum + i;
+    i := i + 1;
+  }
+};
+#end
+
+/-- info: sum = (some 10) -/
+#guard_msgs in
+#eval runProc loopPgm "Test" [.intConst () 5]
```
MikaelMayer:
```suggestion
  while (i <= n)
  {
    sum := sum + i;
    i := i + 1;
  }
};
#end

/-- info: sum = (some 15) -/
#guard_msgs in
#eval runProc loopPgm "Test" [.intConst () 5]
```

It's a bit more standard to say we sum from 1 to n where n is included.

StrataTest/Languages/Python/tests/test_class_with_methods.py
```diff
@@ -1,32 +0,0 @@
-# strata-args: --check-mode bugFinding --check-level full
-import test_helper
```
MikaelMayer, line 2: This file was deleted with no justification in the PR description

Strata/Languages/Core/StatementEval.lean
robin-aws: Oh that's disappointing, I completely assumed `List.zip` had a precondition that they were the same lengths. :P I added the checks you suggested.

Side note, I'm very tempted to make `Env` a monad in the next PR to tidy up chained error conditions now. :)

Strata/DL/Imperative/StmtEval.lean
robin-aws: Agreed on both points, will add in the follow up - proving the consistency with `StmtSemantics.lean` was the very next thing I want us to tackle. :)

aqjune-aws:
Hi, there was a small error found from how `exit` was handled in `loop` and if-then-else. https://github.com/strata-org/Strata/pull/1052 Could you check whether your interpreter is free from these problems?

Strata/DL/Imperative/CmdEval.lean
robin-aws: Yes this was initially an LLM artifact, but then I realized it was sound anyway. :) I added a comment to clear it up and also describe why it's better than just immediately getting stuck.

StrataTest/Languages/Core/Tests/ProgramEvalTests.lean
robin-aws: Ha, more standard in math, less standard in computer science! :)

Strata/Languages/Core/StatementEval.lean
robin-aws: @shigoel Props for catching this, adding the checks revealed that the top-level call to `__main__` in `StrataMain.lean` was not necessarily providing some in or out parameters (e.g. Laurel's `$heap`)! Caused quite a few tests to no longer pass, but I was never aiming to block this initial PR on them all passing anyway. :)

robin-aws:
> Hi, there was a small error found from how exit was handled in loop and if-then-else. https://github.com/strata-org/Strata/pull/1052 Could you check whether your interpreter is free from these problems?

Thanks for the heads-up @aqjune-aws - I updated the interpreter accordingly. It caused a couple of tests to not work because simply pushing a new scope for each block is not completely right, but I'd rather leave fixing that for future PRs when we can actually compare execution behavior against `python` as a baseline.

robin-aws:
Actually, I decided to revert the change for now, since it also broke one of the nice test cases in `ProcedureEvalTests.lean`. It will definitely get fixed as part of proving the consistency between the interpreter and the semantics anyway, which is the next planned step.

shigoel (dismissed):

StrataTest/Languages/Python/expected_interpret/test_class_field_init.expected
```diff
@@ -0,0 +1 @@
+\[ERROR\] procedure '__main__': expected 1 arguments, got 0
```
MikaelMayer, line 1: What is the argument `__main__` expects?

robin-aws: `$heap`! It's an input and output parameter of everything translated from Laurel.

robin-aws: Also note https://github.com/strata-org/Strata/pull/869#discussion_r3149988219

MikaelMayer (dismissed):

shigoel (approved):

aqjune-aws (approved):

PR #870, opened 2026-04-14
Collect Statistics for the Core pipeline

aqjune-aws:
This patch collects statistics for the Core pipeline, which includes:

* Transformations
* Proof obligation generation
* Discharging proof obligations

This patch defines a new `Statistics` data structure, which is simply a HashMap from String to Int.
For transformations, the `CoreTransformState` keeps the `Statistics` data structure and transforms can read/write to it because they live in `CoreTransformM` monad.
For proof obligation generator/discharger, they are simply modeled to return a partial `Statistics` that only record the counters of interest, and then it is the `Statistics.merge` function that merges the results with others.  

The statistics is printed when the existing `--profile` flag is set. An example output is as follows:

```
[statistics] Evaluator.axioms: 4                                                                                                                                                                                                                    
[statistics] Evaluator.factoryOps: 195                                                                                                                                                                                                              
[statistics] Evaluator.functions: 58                                                                                                                                                                                                                
[statistics] Evaluator.procedures: 6                                                                                                                                                                                                                
[statistics] Evaluator.processIteBranches_diverged: 4                                                                                                                                                                                               
[statistics] Evaluator.recursiveFunctions: 11                                                                                                                                                                                                       
[statistics] Evaluator.simulatedStmts: 885                                                                                                                                                                                                          
[statistics] Evaluator.smtProofObligation_numAssumptions: 2189                                                                                                                                                                                      
[statistics] Evaluator.typeDecls: 9                                                                                                                                                                                                                 
[statistics] Evaluator.verificationEnvironments: 1                                                                                                                                                                                                  
[statistics] Evaluator.verify_numObligations: 561                                                                                                                                                                                                   
[statistics] FilterProcedures.erasedProcedures: 22                                                                                                                                                                                                  
[statistics] FilterProcedures.visitedProcedures: 34                                                                                                                                                                                                 
[statistics] PrecondElim.callSiteAssertsEmitted: 339
[statistics] PrecondElim.numFuncsRemovedAfterPrecondStripped: 10
[statistics] PrecondElim.wfProcedureBodyStmtsEmitted: 19
[statistics] PrecondElim.wfProceduresGenerated: 9
[statistics] ProcedureInlining.inlinedCalls: 24
[statistics] ProcedureInlining.visitedCalls: 24
```


keyboardDrummer:
What's the use-case of these statistics?

aqjune-aws:
Discussed internally

Strata/Languages/Core/Statistics.lean
```diff
@@ -0,0 +1,60 @@
...
+  | procedures
+  /-- Number of top-level function declarations in the input program. -/
+  | functions
+  /-- Number of functions inside recursive function in the input program. -/
+  | recursiveFunctions
+  /-- Number of (Program × Env) pairs produced by Program.eval. -/
+  | verificationEnvironments
+  /-- ITE where both branches yielded a single result with no exit label,
+      allowing the evaluator to merge them into one path. -/
+  | processIteBranches_onlyOneBranchChosen
```
shigoel: The name of this constructor doesn't match the comment (and the implementation too). Perhaps `processIteBranches_Merged`?

aqjune-aws: Thanks, fixed it.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -526,22 +532,28 @@ def evalAuxGo (steps : Nat) (old_var_subst : SubstMap) (Ewn : EnvWithNext) (ss :
...
 
-          | .exit l md => [{ Ewn with stk := Ewn.stk.appendToTop [.exit l md], exitLabel := .some l}]
+          | .exit l md => ([{ Ewn with stk := Ewn.stk.appendToTop [.exit l md], exitLabel := .some l}], noStats)
 
-      List.flatMap (fun (ewn : EnvWithNext) => go' ewn rest ewn.exitLabel) EAndNexts
+      let stmtStats := stmtStats.increment s!"{Evaluator.Stats.simulatedStmts}"
+      let (continuations, contStats) := EAndNexts.foldl
+        (fun (acc, statsAcc) ewn =>
+          let (results, s) := go' ewn rest ewn.exitLabel
+          (acc ++ results, statsAcc.merge s))
```
tautschnig, line 551: The statistics accumulation pattern in the hot path creates a small `Statistics` per statement step, then merges via `foldl`. Since each step's stats has 1-2 entries, the per-merge cost is O(1), making the total overhead O(total_steps) HashMap operations. This is acceptable for profiling, but worth noting that the `--profile` flag now has a non-zero cost even when statistics aren't printed (the counters are always accumulated).

Can the accumulation please be gated behind the `profile` flag?

Strata/Transform/LoopElim.lean
```diff
@@ -104,10 +121,11 @@ def Stmt.removeLoopsM
...
-  (s : Stmt P C) : StateM Nat (Stmt P C) :=
+  (s : Stmt P C) : StateM LoopElimState (Stmt P C) :=
   match s with
   | .loop guard measure invariants bss md => do
-    let loop_num ← StateT.modifyGet (fun x => (x, x + 1))
+    let loop_num ← modifyGet fun s =>
+      (s.loopCounter, { s with loopCounter := s.loopCounter + 1 })
     -- Havoc only loop-carried variables. Variables declared inside the loop
     -- body are block-local and should not be treated as pre-existing state by
     -- the passive loop encoding.
```
tautschnig, line 130: The `LoopElimState` structure replaces the bare `Nat` counter with a record containing both `loopCounter` and `statistics`. This is a clean refactoring — the state monad type changes from `StateM Nat` to `StateM LoopElimState`, and the loop counter access is updated accordingly.

Note: `loopElim` (line 228) uses `acc ++ [other]` which is O(n) per iteration. This is a pre-existing pattern (the old code used `List.map`), but since you're already refactoring this function to use `foldl`, consider accumulating with cons + reverse for O(1) per iteration.

Strata/Util/Statistics.lean
```diff
@@ -0,0 +1,79 @@
...
+    match ctorName with
+    | .str _ shortName => s!"    | .{shortName} => \"{pfxStr}.{shortName}\""
+    | _ => dbg_trace "unexpected ctor name: {ctorName}"; "<unknown>"
+  let armsStr := "\n".intercalate arms
+  let code := s!"instance : ToString {ty.getId} where\n  toString\n{armsStr}"
+  let env ← getEnv
+  let stx ← IO.ofExcept (runParserCategory env `command code)
+  elabCommand stx
+
+end
```
tautschnig, line 79: The `derive_prefixed_toString` macro is a nice piece of metaprogramming — it generates `ToString` instances from inductive constructors with a prefix, avoiding manual boilerplate. The approach of parsing generated code as a string and elaborating it is pragmatic.

One concern: the macro uses ``runParserCategory env `command code`` which parses the generated string as Lean syntax. If a constructor name contains characters that are special in Lean syntax (unlikely for stat keys, but possible), this could fail silently. A more robust approach would be to construct the syntax tree directly using `Syntax.mkApp` etc., but arguably that's significantly more complex. Up to you to decide.

Strata/Languages/Core/Core.lean
```diff
@@ -81,12 +82,30 @@ def typeCheckAndPartialEval (options : VerifyOptions) (program : Program)
...
+    | .ax _ _            => s.increment s!"{Evaluator.Stats.axioms}"
+    | .distinct _ _ _    => s.increment s!"{Evaluator.Stats.distincts}"
+    | .proc _ _          => s.increment s!"{Evaluator.Stats.procedures}"
+    | .func _ _          => s.increment s!"{Evaluator.Stats.functions}"
+    | .recFuncBlock fs _ => s.increment s!"{Evaluator.Stats.recursiveFunctions}" fs.length)
+    ({} : Statistics)
+
+  let stats := stats.increment s!"{Evaluator.Stats.factoryOps}" factory.toArray.size
+
+  let (pEs, evalStats) := Program.eval E
```
tautschnig, line 100: The declaration statistics collection uses `s!"{Evaluator.Stats.globalVars}"` etc. to convert enum constructors to string keys. This goes through the `ToString` instance generated by `derive_prefixed_toString`, so the keys are always consistent with the enum definition.

However, the `Evaluator.Stats` enum and the `derive_prefixed_toString` call are in `Statistics.lean`, while the `increment` calls are here. If someone adds a new `Stats` constructor but forgets to add the corresponding `increment` call, there's no compile-time check. Would it be possible to find a more robust design?

Strata/Transform/CallElim.lean
```diff
@@ -31,11 +38,14 @@ def callElimCmd (cmd: Command)
...
       | .call lhs procName args md =>
+        incrementStat s!"{Stats.visitedCalls}"
 
         let some p := (← get).currentProgram | throw "program not available"
 
         let some proc := Program.Procedure.find? p procName | throw s!"Procedure {procName} not found in program"
 
+        incrementStat s!"{Stats.eliminatedCalls}"
+
         -- For each global in modifies, generate a fresh variable to hold its pre-call value,
```
MikaelMayer, line 46: Nit: `eliminatedCalls` is incremented unconditionally after the procedure is found — if the lookup fails, the function throws. So `visitedCalls` and `eliminatedCalls` will always be equal. Consider dropping one of them, or renaming to clarify the intent if a future code path might visit a call without eliminating it.

Strata/Transform/LoopElim.lean
```diff
@@ -163,6 +181,21 @@ def Stmt.removeLoopsM
...
+    --   entry_invariants, entry_invariant_assumes, guard_assumes, inv_assumes,
+    --   maintain_invariants, exit_guard, invariant_assumes,
+    --   plus from measure: assume_m_old, assert_lb, assert_decrease
+    let measureAssertAssumes := if measure.isSome then 3 else 0
+    let numAssertAssumes := entry_invariants.length + entry_invariant_assumes.length +
+      guard_assumes.length + inv_assumes.length + maintain_invariants.length +
+      exit_guard.length + invariant_assumes.length + measureAssertAssumes
+    let newStats := (← get).statistics
+      |>.increment s!"{LoopElim.Stats.visitedLoops}"
+      |>.increment s!"{LoopElim.Stats.erasedLoops}"
```
MikaelMayer, line 193: Nit: `visitedLoops` and `erasedLoops` are always incremented together by 1 for every loop, so they will always be equal. If the intent is to distinguish loops that were visited but not erased (e.g., loops without invariants), the branching logic isn't here yet. Consider dropping one or adding a comment about the planned distinction.

Strata/Transform/LoopElim.lean
```diff
@@ -195,19 +228,25 @@ def Stmt.removeLoops
...
+def loopElim (p : Program) : Program × Statistics :=
+  let (decls, stats) := p.decls.foldl (fun (acc, stats) d =>
+    match d with
     | .proc proc md =>
-      .proc { proc with body := (StateT.run (Block.removeLoopsM proc.body) 0).fst } md
-    | other => other }
+      let (body, st) := StateT.run (Block.removeLoopsM proc.body) {}
+      (acc ++ [.proc { proc with body := body } md], stats.merge st.statistics)
+    | other => (acc ++ [other], stats)) ([], {})
+  ({ decls := decls }, stats)
```
MikaelMayer: The refactoring from `.map` to `.foldl` with `acc ++ [x]` introduces an O(n²) list-append pattern. Consider building in reverse and calling `.reverse` at the end:
```lean
| .proc proc md =>
  let (body, st) := StateT.run (Block.removeLoopsM proc.body) {}
  ((.proc { proc with body := body } md) :: acc, stats.merge st.statistics)
| other => (other :: acc, stats))
```
and then `({ decls := decls.reverse }, stats)` at the end.

Strata/Languages/Core/ProgramEval.lean
```diff
@@ -32,27 +32,31 @@ def initStmtToGlobalVarDecl (s : Statement) : Decl :=
...
     match decl with
 
     | .var name ty init md =>
-      let ssEs := Statement.eval declsE.env [] [(.init name ty init md)]
-      ssEs.flatMap (fun (ss, E) =>
+      let (ssEs, varStats) := Statement.eval declsE.env [] [(.init name ty init md)]
+      let stats := stats.merge varStats
+      ssEs.foldl (fun (acc, statsAcc) (ss, E) =>
                       let xdecls := ss.map initStmtToGlobalVarDecl
                       let declsE := { declsE with xdecls := declsE.xdecls ++ xdecls,
```
MikaelMayer, line 52: Same O(n²) concern: `acc ++ results` inside `foldl` rebuilds the accumulator on every iteration. Consider building in reverse and reversing at the end.

aqjune-aws: The `++` is between decls. This will disappear after #921.

Strata/Languages/Core/Core.lean
aqjune-aws: > If someone adds a new Stats constructor but forgets to add the corresponding increment call, there's no compile-time check. Would it be possible to find a more robust design?

This is true. I think this is a trade-off between making functions monadic vs. keeping their input type as they are & adding `Statistics` as an extra return value. I chose the latter because it seemed less invasive.

Strata/Util/Statistics.lean
aqjune-aws: Yeah, I think it is less likely for statistics keys.

Strata/Transform/LoopElim.lean
aqjune-aws: `acc ++ [other]`: updated the code to reduce time complexity.

Strata/Languages/Core/StatementEval.lean
aqjune-aws: It is a trade-off between chaining the `profile` flag down to StatementEval.lean vs. having it. I think the merging cost will be negligible.

aqjune-aws:
Addressed comments

MikaelMayer (approved):

Strata/Util/Statistics.lean
```diff
@@ -0,0 +1,79 @@
...
+  toString
+    | .ctor₁ => "Prefix.ctor₁"
+    | .ctor₂ => "Prefix.ctor₂"
+```
+-/
+
+open Lean Elab Command Parser in
+/-- Generate a `ToString` instance for an enum-like inductive type where each
+    constructor `.foo` is rendered as `"Prefix.foo"`. -/
+elab "derive_prefixed_toString " ty:ident pfx:str : command => do
```
shigoel, line 64: Nit: I typically expect commands to begin with `#`, so it took me a bit to understand this one was a command...

shigoel (approved):

Strata/Languages/Core/StatementEval.lean
shigoel: Indeed, threading `VerifyOptions` all the way through here does seem like an overkill to address this specific point re. `profile`. But yes, we should keep an open mind that we may need to do that in the future, esp. if we have new options (like configurable path merging).

Strata/Util/Statistics.lean
aqjune-aws: I will address this in the following ver. 2 patch for Laurel..!

Strata/Languages/Core/StatementEval.lean
aqjune-aws: Indeed.. there should be an option for different strategy of evaluation.

PR #877, opened 2026-04-14
 Improve source locations and assertion messages for call-site errors

stefanzetzsche:
> **Depends on:** https://github.com/strata-org/Strata/pull/867

Improves the quality of error diagnostics when `pyAnalyzeLaurel` reports assertion failures and user errors at call sites.

### 1. Error location now points at the method name, not the receiver

For `obj.method(...)`, errors previously pointed at `obj` instead of `method`. The `Call` expression handler and `Expr` statement handler both overwrote the `callRange` metadata with the outer node's range. Fixed by preserving the metadata returned by `translateCall`.

### 2. Human-readable assertion failure messages

Precondition violations previously displayed raw Lean AST dumps. `specExprToLaurel` creates fresh metadata from node source locations, discarding the `propertySummary` set on the input metadata. Fixed by explicitly propagating `propertySummary` onto the returned expression, and using `specExprToMessage` to generate descriptive messages from the formula (e.g. `len(x) >= 1`).

### 3. Concise error messages

Since the error location now points at the method name, messages no longer redundantly include it. Messages are also capitalized for consistency.

## Files changed

- `PythonToLaurel.lean` — Preserve `callRange` metadata in `.Call` and `.Expr` handlers; shorten and capitalize user error messages
- `Specs.lean` — icontract message text adjustment
- `Specs/ToLaurel.lean` — Propagate `propertySummary` through `specExprToLaurel`; use `specExprToMessage` for human-readable precondition labels; build `laurelPreconds` even when no postconditions exist

## Before / After

```
-- Before:
Assertion failed at line 9, col 8: Strata.Python.expr.Compare { start := ...
'method' called with unknown keyword arguments: [arg]

-- After:
Assertion failed at line 9, col 18: len(x) >= 1
Unknown keyword arguments: [arg]
```

### DIFF
--- a/Strata/Languages/Python/Specs.lean
+++ b/Strata/Languages/Python/Specs.lean
@@ -1212,14 +1212,34 @@ def pySpecFunctionArgs (fnLoc : SourceRange)
                        (decorators : Array (expr SourceRange))
                        (returns : Option (expr SourceRange)) : PySpecM FunctionDecl := do
   let mut overload : Bool := false
+  -- Collect icontract preconditions and postconditions from decorators
+  let mut icontractRequires : Array (expr SourceRange) := #[]
+  let mut icontractEnsures : Array (expr SourceRange) := #[]
   for pyd in decorators do
-    let (success, d) ← runChecked <| pySpecValue pyd
-    if success then
-      match d with
-      | .typingOverload =>
-        overload := true
-      | _ =>
-        specError pyd.ann s!"Decorator {repr d} not supported."
+    -- Check for @icontract.require(lambda ...: cond) or @icontract.ensure(lambda ...: cond)
+    match pyd with
+    | .Call _ (.Attribute _ (.Name _ ⟨_, "icontract"⟩ (.Load _)) ⟨_, attr⟩ (.Load _)) args _ =>
+      if args.val.size ≥ 1 then
+        match args.val[0]! with
+        | .Lambda _ _lamArgs lamBody =>
+          if attr == "require" then
+            icontractRequires := icontractRequires.push lamBody
+          else if attr == "ensure" then
+            icontractEnsures := icontractEnsures.push lamBody
+          else
+            specWarning pyd.ann s!"Unknown icontract decorator: icontract.{attr}"
+        | _ =>
+          specWarning pyd.ann s!"icontract.{attr} expects a lambda argument"
+      else
+        specWarning pyd.ann s!"icontract.{attr} requires at least one argument"
+    | _ =>
+      let (success, d) ← runChecked <| pySpecValue pyd
+      if success then
+        match d with
+        | .typingOverload =>
+          overload := true
+        | _ =>
+          specError pyd.ann s!"Decorator {repr d} not supported."
 
   let .mk_arguments _ posonly ⟨_, posArgs⟩ vararg kwonly kw_defaults kwarg defaults := arguments
   assert! posonly.val.size = 0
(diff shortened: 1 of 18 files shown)


Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -467,6 +507,20 @@ def SpecAssertMsg.render : SpecAssertMsg → String
...
+  | .intLit v _ => toString v
+  | .intGe a b _ => s!"{specExprToMessage a} >= {specExprToMessage b}"
+  | .intLe a b _ => s!"{specExprToMessage a} <= {specExprToMessage b}"
+  | .intEq a b _ => s!"{specExprToMessage a} == {specExprToMessage b}"
+  | .intAdd a b _ => s!"{specExprToMessage a} + {specExprToMessage b}"
+  | .intSub a b _ => s!"{specExprToMessage a} - {specExprToMessage b}"
+  | .intMul a b _ => s!"{specExprToMessage a} * {specExprToMessage b}"
+  | .not e _ => s!"not ({specExprToMessage e})"
+  | .len e _ => s!"len({specExprToMessage e})"
+  | _ => "<expr>"
```
joehendrix, line 522: I'd recommend taking precedence and ensuring you are adding parenthesis as needed. 

You may also want to consider making this a ToString instance.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -543,8 +601,64 @@ def funcDeclToLaurel (procName : String) (func : FunctionDecl)
     [{ name := "result", type := match retType.val with
       | .TVoid => tyAny
       | _ => retType }]
-  if func.postconditions.size > 0 then
-    reportError func.loc "Postconditions not yet supported"
+  if func.postconditions.size > 0 then do
+    -- When postconditions exist, use TCore "Any" for all parameters and outputs
```
joehendrix, line 605: I don't think we should be using TCore anymore.  Is this actually Laurel UserDefined?

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -543,8 +601,64 @@ def funcDeclToLaurel (procName : String) (func : FunctionDecl)
...
+      postconds ++ [resultIntExpr]
+    else postconds
+    -- Build Laurel-level precondition expressions for the Procedure.preconditions
+    -- field. CallElim uses these to assert preconditions at call sites and
+    -- assume them in the body, enabling transitivity for internal calls.
+    let laurelPreconds ← func.preconditions.toList.filterMapM fun assertion => do
+      let rawMsg := formatAssertionMessage assertion.message
+      -- Prefer a human-readable rendering of the formula (e.g. "len(Key) >= 1")
+      -- over the generic "icontract precondition" message.
+      let exprMsg := specExprToMessage assertion.formula
```
joehendrix, line 629: It seems like this logic around generating a better message should be moved to where we convert from icontract to PySpecs.

Strata/Languages/Python/Specs.lean
```diff
@@ -1212,14 +1212,34 @@ def pySpecFunctionArgs (fnLoc : SourceRange)
                        (decorators : Array (expr SourceRange))
                        (returns : Option (expr SourceRange)) : PySpecM FunctionDecl := do
   let mut overload : Bool := false
+  -- Collect icontract preconditions and postconditions from decorators
+  let mut icontractRequires : Array (expr SourceRange × String) := #[]
+  let mut icontractEnsures : Array (expr SourceRange) := #[]
```
joehendrix, line 1217: Should the icontract code be seperate from the current botomoog parsing?  With icontracts we probably don't need to parse bodies.

Issue #882, opened 2026-04-14
Python: Core type checker Composite/Any unification failure (13 tests)


olivier-aws:
## Description

When user-defined class instances are used in contexts that expect `Any` (or vice versa), the Core type checker fails with:

```
❌ Type checking error.
Impossible to unify Any with Composite.
```

This affects 13 tests covering object coercion, objects in containers, object reassignment, and class construction patterns.

## Root cause

The Python-to-Laurel-to-Core pipeline translates user-defined classes as `Composite` types, but many operations (list insertion, dict storage, `Any`-typed variables) expect `Any`. The Core type checker has no coercion path between `Composite` and `Any`.

## Affected tests (13)

**Object coercion to Any:** test_coerce_object_to_any, test_any_field_reassign

**Objects in containers:** test_object_in_list, test_object_in_dict, test_object_field_is_object

**Object reassignment:** test_object_reassign, test_recursive_data

**Class construction:** test_constructor_no_args, test_empty_class, test_missing_init

**Other:** test_with_statement, test_var_shadow_func, test_method_return_int

## Reproduction

```bash
cd StrataTest/Languages/Python
(cd ../../../Tools/Python && python3 -m strata.gen py_to_strata \
  --dialect "dialects/Python.dialect.st.ion" \
  "../../StrataTest/Languages/Python/tests/pending/test_coerce_object_to_any.py" \
  "../../StrataTest/Languages/Python/tests/pending/test_coerce_object_to_any.python.st.ion")
(cd ../../.. && ./.lake/build/bin/strata pyAnalyzeLaurel \
  "StrataTest/Languages/Python/tests/pending/test_coerce_object_to_any.python.st.ion")
```

olivier-aws:
Agents working on it

PR #919, opened 2026-04-14
fix(python): Add reassigned function params to modifies clause (#882)

olivier-aws:
## Summary

When a function parameter is reassigned inside the body (e.g. `x = x + 10`), the generated Core procedure's modifies clause did not include that parameter, causing the type checker to reject the procedure.

### Root cause

Python parameters are pass-by-value for immutable types — reassigning `x` inside `f` doesn't affect the caller's variable. However, the Laurel-to-Core translation used the same variable name for both the input parameter and the local mutable copy, so the Core type checker saw a modification to a variable not in the modifies clause.

### Fix

Parameters are now renamed with a `$in_` prefix in the generated Core procedure. The function body gets a local copy initialized from the prefixed input, so reassignment is properly tracked in the modifies clause.

### Reproduction

```python
def f(x: int) -> int:
    x = x + 10  # reassigns parameter
    return x
```

Before: `[f]: This procedure modifies variables it is not allowed to! Variables actually modified: [LaurelResult, x]`
After: passes type checking (Inconclusive — 0 failures)

### Files changed

- `Strata/Languages/Python/PythonToLaurel.lean` — parameter renaming logic
- `StrataTest/Languages/Python/` — new and updated test files and expected outputs

Partially addresses #882


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1743,7 +1743,8 @@ def translateFunction (ctx : TranslationContext) (sourceRange: SourceRange) (fun
     | some kwargs => inputs:= inputs ++ [{ name := kwargs, type := mkCoreType PyLauType.DictStrAny}]
     | _ => pure ()
 
-    let typeConstraintPreconditions := getInputTypePreconditions funcDecl
+    let typeConstraintPreconditions := funcDecl.args.filterMap
+      (fun arg => getUnionTypeConstraint ("$in_" ++ arg.name) arg.md arg.tys funcDecl.name)
```
tautschnig: The `$in_` prefix leaks into user-facing diagnostic messages. The type constraint preconditions now say "Type constraint of $in_x" instead of "Type constraint of x" (visible in all the expected output changes).

The constraint *expression* must reference `$in_x` (that's the actual parameter name), but the diagnostic *summary* should use the original name for better UX. Consider passing the original name for the summary while using the prefixed name for the expression:

```suggestion
    let typeConstraintPreconditions := funcDecl.args.filterMap
      (fun arg => getUnionTypeConstraint ("$in_" ++ arg.name) arg.md arg.tys funcDecl.name
        |>.map fun c => { c with summary := c.summary.map (·.replace "$in_" "") })
```

Or better, refactor `getUnionTypeConstraint` to accept separate `varExprName` and `varDisplayName` parameters. This way the expression references `$in_x` but the summary says "Type constraint of x".

tautschnig: Nit: `getInputTypePreconditions` (line 1714-1715) is now dead code — it's defined but no longer called. Consider removing it to avoid confusion.

olivier-aws: Fixed — added an optional `displayName` parameter to `getUnionTypeConstraint` (defaults to `var`). The call site now passes `arg.name` as the display name, so diagnostics show "Type constraint of x" while the expression still references `$in_x`. See cdeb77d7.

olivier-aws: Done — removed `getInputTypePreconditions` in the same commit (cdeb77d7).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1758,14 +1757,26 @@ def translateFunction (ctx : TranslationContext) (sourceRange: SourceRange) (fun
...
+    -- declaration  var <name> := $in_<name>  so the body works with a
+    -- freely-modifiable local copy.
+    let renamedInputs := inputs.map fun p =>
+      { p with name := mkId ("$in_" ++ p.name.text) }
+    let paramCopies := inputs.map fun p =>
+      let origName := p.name.text
+      let renamedName := "$in_" ++ origName
+      mkStmtExprMd (StmtExpr.LocalVariable origName p.type
+        (some (mkStmtExprMd (StmtExpr.Identifier renamedName))))
     let bodyBlock : StmtExprMd := match bodyBlock.val with
```
tautschnig: The parameter-renaming + local-copy logic is now duplicated almost verbatim between `translateFunction` (here) and `translateMethod` (lines 1905–1915). The only difference is the `self` exclusion in `translateMethod`.

Consider extracting a helper, e.g.:

```lean
def renameInputParams (inputs : List Parameter) (exclude : String → Bool := fun _ => false)
    : List Parameter × List StmtExprMd :=
  let toRename := inputs.filter (fun p => !exclude p.name.text)
  let renamed := inputs.map fun p =>
    if exclude p.name.text then p
    else { p with name := mkId ("$in_" ++ p.name.text) }
  let copies := toRename.map fun p =>
    let orig := p.name.text
    mkStmtExprMd (StmtExpr.LocalVariable orig p.type
      (some (mkStmtExprMd (StmtExpr.Identifier ("$in_" ++ orig)))))
  (renamed, copies)
```

This would reduce the duplication and make it easier to keep both paths in sync.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1758,14 +1757,26 @@ def translateFunction (ctx : TranslationContext) (sourceRange: SourceRange) (fun
       match arg.tys with | [ty] => (arg.name, ty) | _ => (arg.name, PyLauType.Any))
     let (bodyBlock, newCtx) ←  translateFunctionBody ctx inputTypes body
     let noneReturn := mkStmtExprMd (.Assign [mkStmtExprMd (.Identifier PyLauFuncReturnVar)] AnyNone)
+    -- In Python, parameters are mutable local variables.  In Core, input
+    -- parameters cannot be modified.  To bridge this gap we rename each
+    -- input parameter to "$in_<name>" and prepend a local variable
+    -- declaration  var <name> := $in_<name>  so the body works with a
+    -- freely-modifiable local copy.
+    let renamedInputs := inputs.map fun p =>
+      { p with name := mkId ("$in_" ++ p.name.text) }
```
tautschnig: Nit: The string `"$in_"` appears 5 times across the file. A named constant (e.g., `def paramInputPrefix := "$in_"`) would make the convention self-documenting and reduce the risk of typos if the prefix ever changes.

StrataTest/Languages/Python/tests/test_var_shadow_func.py
```diff
@@ -0,0 +1,11 @@
+def f(x: int) -> int:
```
tautschnig, line 1: This file is an exact copy of `tests/pending/test_var_shadow_func.py`, which still exists. The pending copy should be removed now that the test has been promoted — otherwise `--check-pending` in CI will flag it as a pending test that now passes.

StrataTest/Languages/Python/tests/test_param_reassign.py
```diff
@@ -0,0 +1,38 @@
...
+    r3: int = no_param_reassign(2, 3)
+    assert r3 == 5, "no reassign"
+
+    r4: int = partial_param_reassign(10, 5)
+    assert r4 == 15, "partial reassign"
+
+    r5: int = param_reassign_twice(3)
+    assert r5 == 8, "reassign twice"
+
+test_param_reassign()
```
tautschnig, line 38: Good coverage of the core scenarios (single param, multi param, no reassign, partial reassign, double reassign). Two suggestions:

1. Consider adding a test with `**kwargs` to exercise the `DictStrAny_get_param` path, which also uses `arg.name` from the function signature.
2. A cross-module test (calling a function defined in a separately-translated module with keyword arguments) would exercise the `PreludeInfo.ofLaurelProgram` path and catch the bug described in the review summary.

tautschnig (changes requested):
A few items to address:

**Bug — `PreludeInfo.ofLaurelProgram` leaks `$in_` into cross-module signatures (line 2158)**

`PreludeInfo.ofLaurelProgram` (line 2131) builds `PythonFunctionDecl` from already-translated Laurel procedures by reading `param.name.text`. After this PR, those names carry the `$in_` prefix. This means when module A imports module B, the function signatures for B's functions will have `args[i].name == "$in_x"` instead of `"x"`. This breaks:
- Keyword-argument matching in `combinePositionalAndKeywordArgs` (~line 882)
- `DictStrAny_get_param` expansion (~line 997)
- Error messages (e.g., `missing required argument '$in_x'`)

Suggested fix: strip the prefix in `ofLaurelProgram`:
```lean
let rawName := if param.name.text.startsWith "$in_" then param.name.text.drop 4 else param.name.text
{name:= rawName, md:= default, tys:= [getHighTypeName param.type.val], default:= some noneexpr}
```
Or better, define a constant `def paramInputPrefix := "$in_"` and use it everywhere.

olivier-aws:
@keyboardDrummer-bot Were the comments by the reviewer correctly handled?

olivier-aws:
@keyboardDrummer-bot Address the still open comments

StrataTest/Languages/Python/tests/test_param_reassign.py
tautschnig: @keyboardDrummer-bot Can you please add these tests, I don't think we should treat this as optional.

tautschnig:
@keyboardDrummer-bot Can you please add the tests proposed in https://github.com/strata-org/Strata/pull/919#discussion_r3088210201?

tautschnig (approved):

shigoel (approved):
Clean fix with thorough test coverage. The `renameInputParams` extraction, `paramInputPrefix` constant, `displayName` parameter for diagnostics, and prefix-stripping in `PreludeInfo.ofLaurelProgram` are all well done. Prior review feedback fully addressed.

PR #921, opened 2026-04-15
Update Core's evaluator do symbolic execution only without partial evaluation

aqjune-aws:
Strata's Core evaluator is truly a partially evaluator in the sense that it can simplify the given program for every possible execution path, and returns portfolio of simplified programs. It means that there are N diverging control flows in a program, it can return N different programs.
However, this can cause certain benchmarks' memory usage of Strata increase quickly to maintain a set of variants of programs after partial evaluation. I think that without explicitly keeping the partially evaluated programs Strata will still simplify symbolic expressions pretty well and the ultimate verification conditions will virtually look the same. It is because partial evaluator not only returns the partially evaluated programs, but also the symbolic environment after executing the program. The symbolic environment will already contain path conditions and proof obligations - these are independent from the partially evaluated program. The benchmark seems unregressed after this change.

This patch renames `Core.typeCheckAndPartialEval` to `Core.typeCheckAndEval`, and its underlying partial evaluators to purely do symbolic simulations only. :)

---

CLARIFICATION: The Lambda's evaluator will still remain as a partial evaluator, as it wasn't designed to diverge into multpiple different evaluation results for now

---


Strata/Languages/Core/ProgramEval.lean
```diff
@@ -19,84 +19,57 @@ open Lambda.LTy Lambda.LExpr Statement Procedure Program
...
     | .type _ _ =>
-      go rest { declsE with xdecls := declsE.xdecls ++ [decl] }
+      go rest declsE
 
     | .ax a _ =>
       -- All axioms go into the top-level path condition before anything is executed.
       -- There should be exactly one entry in the path condition stack at this point.
-      if declsE.env.pathConditions.length != 1 then
+      if declsE.pathConditions.length != 1 then
         panic! "Internal error: path condition stack misaligned when adding axiom"
```
shigoel: Probably out of scope of this PR (because it pre-existed it), but saw this while reviewing: it'd be nice for this panic to be replaced by an error, at the cost of changing the signature.

StrataTest/Languages/Core/Examples/BitVecParse.lean
```diff
@@ -49,8 +49,8 @@ program Core;
 
 procedure bitVecParseTest () returns ()
 {
-  assert [bitvec32_test]: true;
-  assert [bitvec64_test]: false;
+  assert [bitvec32_test]: bv{32}(4294945741) == bv{32}(4294945741);
```
shigoel: This looks like a regression in the capabilities of the evaluator, but it stems from not retaining the evaluated terms anymore, which means we end up displaying the original program. 

While not retaininig the evaluated programs anymore is the right thing to do, we should still have a mode that can record or display the simplifications the evaluator made to the program (perhaps only in `.debug` mode of verbosity or during some profile mode). It's fine for simpler examples like these, but for larger ones (e.g., StrataTest/Languages/Core/Examples/Seq.lean), having that visibility can be useful in debugging (even during development).

Also, I don't think there's any benefit to displaying the original program (not the evaluated one) in the `.normal` verbose mode. That does mean that all the `#guard_msgs` with the evaluated program in it will need to be modified.

MikaelMayer:
I would love to see how this PR could be influenced by or influence this other PR I'm working on:
https://github.com/strata-org/Strata/pull/753
Do you have any insights?

Strata/Languages/Core/StatementEval.lean
```diff
@@ -388,14 +388,12 @@ def StmtsStack.appendToTop (stk : StmtsStack) (ss : Statements) : StmtsStack :=
   (top ++ ss) :: stk
```
shigoel: `StmtsStack` and its four helper functions (`push`, `pop`, `top`, `appendToTop`) are now dead code after the `stk` field was removed from `EnvWithNext`. They should be removed to avoid confusion.

Strata/Languages/Core/ProcedureEval.lean
```diff
@@ -46,19 +46,18 @@ The differences across paths are:
 The `fallback` pair is returned when `results` is empty (which should not occur
 in practice, since `Statement.eval` always produces at least one result).
```
shigoel, line 47: Stale docstring: "The `fallback` pair is returned..." but `fallback` is now a single `Env`, not a pair. Should say "The `fallback` environment is returned...".

Strata/Languages/Core/StatementEval.lean
```diff
@@ -490,8 +479,7 @@ def evalAuxGo (steps : Nat) (old_var_subst : SubstMap) (Ewn : EnvWithNext) (ss :
                 if cond'.isTrue then (then_ss, else_ss) else (else_ss, then_ss)
               let deadDeferred := collectDeadBranchDeferred ss_dead c Ewn.env.pathConditions
               let Ewns :=
-                (go' Ewn ss_live .none).map fun (ewn : EnvWithNext) =>
-                  { ewn with stk := orig_stk.appendToTop [.ite (.det cond') ewn.stk.top [] md] }
+                (go' Ewn ss_live .none).map fun (ewn : EnvWithNext) => ewn
               -- Prepend dead-branch obligations to the first result only.
```
shigoel, line 470: This `.map fun (ewn : EnvWithNext) => ewn` is an identity map left behind after the `stk` update was removed. It can be simplified to just `go' Ewn ss_live .none`.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -609,17 +590,17 @@ statements.
...
+def eval (E : Env) (old_var_subst : SubstMap) (ss : Statements) : List Env :=
   (evalAux E old_var_subst ss .none).map exitToError
 
 /--
 Partial evaluator for statements yielding one environment and transformed
 statements.
 -/
-def evalOne (E : Env) (old_var_subst : SubstMap) (ss : Statements) : Statements × Env :=
+def evalOne (E : Env) (old_var_subst : SubstMap) (ss : Statements) : Env :=
   match eval E old_var_subst ss with
```
shigoel: The docstrings for both `eval` and `evalOne` still say "yielding a list of environments and transformed statements" / "yielding one environment and transformed statements" but neither function returns transformed statements anymore — they return `List Env` and `Env` respectively.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -490,8 +479,7 @@ def evalAuxGo (steps : Nat) (old_var_subst : SubstMap) (Ewn : EnvWithNext) (ss :
                 if cond'.isTrue then (then_ss, else_ss) else (else_ss, then_ss)
               let deadDeferred := collectDeadBranchDeferred ss_dead c Ewn.env.pathConditions
               let Ewns :=
-                (go' Ewn ss_live .none).map fun (ewn : EnvWithNext) =>
-                  { ewn with stk := orig_stk.appendToTop [.ite (.det cond') ewn.stk.top [] md] }
+                (go' Ewn ss_live .none).map fun (ewn : EnvWithNext) => ewn
```
MikaelMayer: Nit: This is an identity map (`.map fun ewn => ewn`). It can be simplified to just `go' Ewn ss_live .none`.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -609,17 +590,17 @@ statements.
 The argument `old_var_subst` below is a substitution map from global variables
 to their pre-state value in the enclosing procedure of `ss`.
 -/
-def eval (E : Env) (old_var_subst : SubstMap) (ss : Statements) : List (Statements × Env) :=
+def eval (E : Env) (old_var_subst : SubstMap) (ss : Statements) : List Env :=
```
MikaelMayer: Nit: The docstring still says "yielding a list of environments and transformed statements" but the function no longer returns statements. Same applies to `evalOne` below.

StrataTest/Languages/Core/Examples/BitVecParse.lean
aqjune-aws: > While not retaininig the evaluated programs anymore is the right thing to do, we should still have a mode that can record or display the simplifications the evaluator made to the program (perhaps only in .debug mode of verbosity or during some profile mode). It's fine for simpler examples like these, but for larger ones (e.g., StrataTest/Languages/Core/Examples/Seq.lean), having that visibility can be useful in debugging (even during development).

This is a legit point - and I think this can be achieved through implementing a transformation that simplifies programs. For example, we can write a transform that traverses expressions in a program and calls `LExpr.eval`. There are two benefits in the alternative approach:

(1) This allows other interesting optimizations to chime in *after* partial evaluation. Let's assume that we want to apply arithmetic optimizations like `x < x + a ==> 0 < a`. Currently, these optimizations have to bake in the partial evaluator - and it will make the debugging hard as well because partial evaluator will not have 'intermediately' optimized programs. For example, it will turn `x < x + (1 + 2)` to suddenly `true` without showing the intermediate state. However, if we have two different simplification transformations - one that simply invokes partial evaluator, and one that does these arithmetic optimizations, it is just about composing how to invoke the transformations. :) Call ArithSimplify after PE, PE after ArithSimplify, or call them iteratively until they reach fixpoint.

(2) It fits well with the existing profiling infrastructure of Strata. Strata can emit IR after each transformation can measure running time of each pass. If simplification is transformation, we can safely have e.g. an expensive CVC5-based optimization, and always measure its impact through existing `--profile` option.

Strata/Languages/Core/ProgramEval.lean
aqjune-aws: Let me address this in this PR..!

MikaelMayer (commented):
Clean simplification that removes the partial-evaluation output (transformed programs) from the evaluator, keeping only the symbolic environments. The core idea is sound: the proof obligations and path conditions already live in the `Env`, so carrying the rewritten program through the pipeline was unnecessary overhead.

A few items to address:

1. **Bug — missed rename in `Strata/MetaVerifier.lean`**: `genVCs` (line 60) still calls `Core.typeCheckAndPartialEval`, which was renamed to `Core.typeCheckAndEval`. This will fail to compile.

2. **Dead code — `StmtsStack` in `StatementEval.lean`**: The `StmtsStack` type alias and its four methods (`push`, `pop`, `top`, `appendToTop`, lines 371–389) are no longer referenced after removing the `stk` field from `EnvWithNext`. They should be removed.

3. See inline comments for two smaller nits.

MikaelMayer:
@keyboardDrummer-bot can you please look at this comment
https://github.com/strata-org/Strata/pull/921#issuecomment-4252496339
And do an analysis of the two PRs to see how they overlap, complement each other, or may influence each other?

aqjune-aws:
> I would love to see how this PR could be influenced by or influence this other PR I'm working on: #753 Do you have any insights?

I think in high level it goes well with #753 (even if it might have some merge conflicts), because the direction is towards simplifying the PE engine and factoring out program-transforming logics to program transformations. Every transformation that is newly defined in #753 will still be valid ones.

aqjune-aws:
Addressed comments

StrataTest/Languages/Core/Examples/BitVecParse.lean
aqjune-aws: > This looks like a regression in the capabilities of the evaluator, but it stems from not retaining the evaluated terms anymore, which means we end up displaying the original program.

I think I understood better this sentence - it will still emit `true` and `false` as single SMT queries at proof obligations, because Lambda's evaluator will remain as a partial evaluator.

shigoel: Indeed, and my comment was that it's confusing to NOT see that in the program that' printed out, but see that in the SMT queries or the VCs themselves. Which is why I think we should suppress printing the program in the first place (last paragraph of the original comment). Sorry, should've phrased this better...

shigoel (dismissed):

Strata/DL/Imperative/CmdEval.lean
```diff
@@ -103,7 +103,7 @@ def Cmd.eval [BEq P.Ident] [EC : EvalContext P S] (σ : S) (c : Cmd P) : Cmd P 
       (c', EC.deferObligation σ (ProofObligation.mk label .cover assumptions e md))
 
 /--
-Partial evaluator for Imperative's Commands.
+Symbolic simluation for Imperative's Commands.
```
tautschnig: Typo: "simluation" → "simulation". Same typo appears in two other places:
- `Strata/Languages/Core/StatementEval.lean` line 593: "simluator" → "simulator"
- `Strata/Languages/Core/Verifier.lean` line 626: "simluation" → "simulation"

```suggestion
Symbolic simulation for Imperative's Commands.
```

Strata/Languages/Core/ProgramEval.lean
```diff
@@ -19,88 +19,63 @@ open Lambda.LTy Lambda.LExpr Statement Procedure Program
...
-                      let declsE := { declsE with xdecls := declsE.xdecls ++ xdecls,
-                                                  env := E }
-                      let (results, s) := go rest declsE statsAcc
-                      (acc ++ results, s))
-        ([], stats)
+      ssEs.foldl (fun acc E => do
+                      let (accEs, statsAcc) ← acc
+                      let (results, s) ← go rest E statsAcc
+                      .ok (accEs ++ results, s))
+        (.ok ([], stats))
```
tautschnig, line 40: The `.var` case calls `Statement.eval` which can return multiple environments (from branching). The `foldl` accumulates results from all branches via `Except`. This is correct, but note that the old code had the same branching behavior — the difference is just that we no longer accumulate `xdecls`.

One subtle point: `Statement.eval` returns `List Env` where each `Env` still carries the original `program` field (set at the top of `typeCheckAndEval`). Since the program is never modified during evaluation anymore, all `Env`s in the result share the same `program`. This is the key invariant that makes the simplification sound — worth a brief comment.

Strata/Languages/Core/ProgramEval.lean
```diff
@@ -19,88 +19,63 @@ open Lambda.LTy Lambda.LExpr Statement Procedure Program
...
-  env : Env
-  xdecls : Decls := []
-
-def initStmtToGlobalVarDecl (s : Statement) : Decl :=
-  match s with
-  | .init x ty e md => (.var x ty e md)
-  | _ => panic s!"Expected a variable initialization; found {format s} instead."
-
-def eval (E : Env) : List (Program × Env) × Statistics :=
+def eval (E : Env) : Except String (List Env × Statistics) :=
```
tautschnig: Good change: replacing `panic!` with `Except String` (addressing shigoel's feedback). The error propagation via `do`-notation in the `.var` case is clean.

However, the `.func` and `.recFuncBlock` error cases still use `.ok ([{ declsE with error := ... }], stats)` — they set an error in the `Env` and return `Ok` rather than returning `.error`. This is pre-existing behavior, but now that the function returns `Except`, it would be more consistent to use `.error` for these cases too.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -612,32 +572,30 @@ end
...
 -/
-def eval (E : Env) (old_var_subst : SubstMap) (ss : Statements) : List (Statements × Env) × Statistics :=
+def eval (E : Env) (old_var_subst : SubstMap) (ss : Statements) : List Env × Statistics :=
   let (ewns, stats) := evalAux E old_var_subst ss .none
   (ewns.map exitToError, stats)
 
 /--
-Partial evaluator for statements yielding one environment and transformed
-statements.
+A symbolic simluator for statements yielding one environment.
```
tautschnig: Typo: "simluator" → "simulator".

```suggestion
A symbolic simulator for statements yielding one environment.
```

Strata/Languages/Core/ProgramEval.lean
aqjune-aws: This is a good point - I made this explicit at `typeCheckAndEval`.

Strata/Languages/Core/ProgramEval.lean
aqjune-aws: This was a very valid point. Addressed in my commit, thanks :)

Strata/Languages/Core/Core.lean
```diff
@@ -96,14 +96,16 @@ def typeCheckAndPartialEval (options : VerifyOptions) (program : Program)
     ({} : Statistics)
 
   let stats := stats.increment s!"{Evaluator.Stats.factoryOps}" factory.toArray.size
-
-  let (pEs, evalStats) := Program.eval E
+  let (pEs, evalStats) ← Program.eval E
+  -- Note: all .program fields in pEs will have identical values, because
+  -- Program.eval is supposed to unchange the program. The Program field is
```
tautschnig: Nit: "unchange" isn't standard English — consider "does not modify" or "leaves unchanged".

```suggestion
  -- Note: all .program fields in pEs will have identical values, because
  -- Program.eval does not modify the program. The Program field is
  -- kept for convenience.
```

tautschnig (approved):

shigoel (approved):

Issue #911, opened 2026-04-14
Python: imprecise verification for boolean short-circuit and bitwise operations (5 tests)


olivier-aws:
## Description

The analysis produces inconclusive VCs for:

1. **Boolean short-circuit evaluation** — `and`/`or` with side effects (division by zero guarded by short-circuit).
2. **Bitwise operations** — `&`, `|`, `^` are parsed but the solver cannot reason about them.

## Affected tests (5)

**Boolean short-circuit (2):**
- `test_bool_shortcircuit_and` — `(x != 0) and (10 // x > 0)`
- `test_bool_shortcircuit_or` — `(x == 0) or (10 // x > 0)`

**Bitwise operations (3):**
- `test_bitand` — `z = x & y; assert z == 0b1000`
- `test_bitor` — `z = x | y; assert z == 0b1110`
- `test_bitxor` — `z = x ^ y; assert z == 0b0110`

PR #922, opened 2026-04-15
Restore short-circuit ite encoding for AndThen/OrElse/Implies

fabiomadge:
PR #638 changed the Laurel→Core translation of `AndThen`/`OrElse`/`Implies` from `ite` (if-then-else) encoding to eager `boolAndOp`/`boolOrOp`/`boolImpliesOp`. This broke short-circuit well-definedness: expressions like `y != 0 && x / y > 0` now evaluate the division even when `y == 0`.

The `ite` encoding is the correct semantics — it prevents evaluation of the second operand when the first operand short-circuits. The `DesugarShortCircuit` pass only handles imperative operands, not well-definedness concerns like division by zero or precondition violations on pure function calls.

This reverts the three translator lines and removes the error annotations added to `T15_ShortCircuit.lean` to accept the broken behavior.

### DIFF
--- a/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
+++ b/Strata/Languages/Laurel/LaurelToCoreTranslator.lean
@@ -196,9 +196,9 @@ def translateExpr (expr : StmtExprMd)
     | .Neq => return .app () boolNotOp (.eq () re1 re2)
     | .And => return binOp boolAndOp
     | .Or => return binOp boolOrOp
-    | .AndThen => return binOp boolAndOp
-    | .OrElse => return binOp boolOrOp
-    | .Implies => return binOp boolImpliesOp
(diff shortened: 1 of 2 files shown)


tautschnig:
Does this fix, or at least partly fix, #911?

fabiomadge:
@tautschnig I would think so

keyboardDrummer:
Can we wait for the PR from https://github.com/strata-org/Strata/issues/924 ? I will ask the bot to see if that also allows making the same test changes as this PR does. That way, we can still use Core's `&&`, `||` and `==>` operators so we don't negatively affect debuggability of our pipeline.

fabiomadge:
I think we agreed that the plan in #924 doesn't work in general. At least for the postconditions we're looking to add eventually.

joscoh (approved):

keyboardDrummer (approved):

PR #923, opened 2026-04-15
Replace string-based operator dispatch with structured CoreOp types

### LINKED ISSUE #785: Replace string-based operator dispatch with structured types
## Problem

Operator identity throughout the codebase is represented as plain strings (e.g., `"Bv32.Add"`, `"Int.SafeDiv"`). Every consumer that needs to branch on operator kind does so via string matching — pattern matches on literal names, `startsWith` prefix checks, or membership in string lists. This appears in the SMT encoder, GOTO backend, ASTtoCST, PrecondElim, and others.

This causes several recurring issues:

1. **No exhaustiveness checking.** When a new operator is added, every string-matching site must be updated manually. The compiler cannot flag missing cases. (Example: operators generated by factory macros for certain bitvector sizes may be missing from the SMT encoder's match block without any compile error.)

2. **No typo detection.** A misspelled string like `"Bv32.SLtt"` silently falls through to the default/error case.

3. **Redundant size enumeration.** Logic that is uniform across bitvector sizes (e.g., `Add` has the same SMT encoding for all sizes) must be repeated per size as separate string cases, or use brittle string parsing (`(name.splitOn ".").head!.drop 2 |>.toNat!`).

4. **Classification by convention.** Determining properties of an operator (is it signed? is it a predicate?) requires re-deriving the answer from the string name at each use site, rather than querying a structured definition.

## Proposed approach

Introduce structured types for operator identity. For bitvector ops:

```lean
inductive BvOpKind where
  | Add | Sub | Mul | Neg | SDiv | SMod
  | Not | And | Or | Xor | Shl | SShr | Lshr
  | Eq | Lt | Le | SLt | SLe | SGt | SGe
  | Concat | Extract (hi lo : Nat)
  | ...

structure BvOp where
  size : BvSize
  kind : BvOpKind
```

For numeric ops that share a common interface across `Int` and `Real`:

```lean
inductive NumericType where
  | int | real

inductive NumericOpKind where
  | Add | Sub | Mul | Neg
  | Lt | Le | Gt | Ge
  | SafeDiv | SafeMod
  | ...

structure NumericOp where
  ty : NumericType
  kind : NumericOpKind
```

Then a top-level sum ties them together:

```lean
inductive CoreOp where
  | bv (op : BvOp)
  | numeric (op : NumericOp)
  | bool (kind : BoolOpKind)    -- And, Or, Not, Implies, ...
  | ...
```

With derived predicates:

```lean
def BvOpKind.isSigned : BvOpKind → Bool
def BvOpKind.isPredicate : BvOpKind → Bool
def NumericOpKind.hasPrecondition : NumericOpKind → Bool
def CoreOp.arity : CoreOp → Nat
```

Consumers would pattern match on the structured type (with compiler-enforced exhaustiveness) or query predicates, rather than parsing strings. The factory macros would produce structured values, and a single `toString`/`ofString` boundary would exist at the DDM parser/printer layer.

## Scope

This is a cross-cutting refactor touching the `Lambda.Factory` infrastructure, all backends (SMT, GOTO), the DDM parser/printer (ASTtoCST, CSTtoAST), and transforms (PrecondElim). It should be done incrementally — e.g., start with `BvOp` while keeping a string conversion at module boundaries, then migrate consumers one at a time.

tautschnig:
Introduces a structured `CoreOp` type hierarchy that replaces string-based operator identity throughout the codebase. Consumers now pattern match on `CoreOp` variants instead of string literals, giving us compiler-enforced exhaustiveness checking and eliminating typo risks.

It will likely make sense to review commit-by-commit.

## What changed

**New type definitions** (`Strata/Languages/Core/CoreOp.lean`):
- `CoreOp` inductive sum type with variants: `bv`, `numeric`, `bool`, `str`, `re`, `map`, `seq`, `bvExtract`, `trigger`, `other`
- Sub-types: `BvOp` (size × kind), `NumericOp` (int/real × kind), `BoolOpKind`, `StrOpKind`, `ReOpKind`, `MapOpKind`, `SeqOpKind`, `TriggerOpKind`
- Derived predicates: `isSigned`, `isPredicate`, `hasPrecondition`, `isUnary`
- Bidirectional `toString`/`ofString` with `CoreOp.other` escape hatch for user-defined functions
- Round-trip tests for all known operators

**Consumer-side migrations** (string match → CoreOp pattern match):
- `SMTEncoder.lean`: 115 per-size BV string cases collapse to 25 size-generic patterns
- `LambdaToCProverGOTO.lean`: eliminates `startsWith`/`drop`/`toNat!` string parsing; `isSignedBvOp` and `parseBvExtractLo` replaced with one-line CoreOp matches
- `ASTtoCST.lean`: all 5 handler functions migrated; `bvBinaryOpMap` keys changed from `String` to `BvOpKind`
- `PrecondElim.lean`: `startsWith` prefix checks replaced with CoreOp pattern match

**Producer-side migrations** (string literals → `CoreOp.toString`):
- `Translate.lean`, `C_Simp/Translate.lean`, `C_Simp/Verify.lean`, `Python/PythonToCore.lean`

## Design decisions

- `Func.name` remains string-based (`Identifier`) for DDM parser/printer compatibility. `CoreOp` is used at consumption/production sites with a single `ofString`/`toString` boundary.
- `CoreOp.other name` provides a fallback for user-defined functions, ensuring no behavioral change for non-built-in operators.
- Lambda-level test files are left with string literals since they test the Lambda dialect and shouldn't depend on `Core.CoreOp`.

Fixes: #785


Strata/Languages/Core/CoreOp.lean
```diff
@@ -0,0 +1,343 @@
...
+    let hi ← hiStr.toNat?
+    let lo ← loStr.toNat?
+    return .bvExtract size hi lo
+  | _ => .none
+
+/-- Parse a string operator name into a structured `CoreOp`.
+    Returns `CoreOp.other name` for unrecognized names. -/
+def CoreOp.ofString (name : String) : CoreOp :=
+  -- Try BV ops: "Bv{size}.{kind}"
+  let bvPrefixes := ["Bv1.", "Bv2.", "Bv8.", "Bv16.", "Bv32.", "Bv64."]
```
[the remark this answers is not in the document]
tautschnig: Copilot is wrong here, but this still points to a fragile design. Will revise to avoid this going stale.

Strata/Languages/Core/CoreOp.lean
```diff
@@ -0,0 +1,339 @@
...
+  | .bvExtract size hi lo => s!"Bv{size}.Extract_{hi}_{lo}"
+  | .trigger kind => kind.toString
+  | .other name => name
+
+instance : ToString CoreOp := ⟨CoreOp.toString⟩
+
+/-- Try to parse a BV extract suffix like "Extract_7_0" given a known size. -/
+private def parseBvExtract? (size : Nat) (rest : String) : Option CoreOp := do
+  guard (rest.startsWith "Extract_")
+  let nums := (rest.drop 8).toString.splitOn "_"  -- drop "Extract_"
```
MikaelMayer, line 297: Nit: `String.drop` already returns a `String`, so `.toString` is a no-op. Same applies to the other `.toString` calls after `.drop` in `CoreOp.ofString` (lines 293, 300, 305, 310, 316).

Strata/Languages/Core/CoreOp.lean
```diff
@@ -0,0 +1,339 @@
...
+  else if name.startsWith "Sequence." then
+    match SeqOpKind.ofString? (name.drop 9).toString with
+    | some kind => .seq kind
+    | none => .other name
+  -- Try Map ops (no prefix — "const", "select", "update")
+  else match MapOpKind.ofString? name with
+  | some kind => .map kind
+  | none =>
+  -- Try Trigger ops
+  match name with
```
MikaelMayer: Nit: `TriggerOpKind` is the only sub-type without an `ofString?` method — its parsing is inlined here as string matching. Every other sub-type follows the `ofString?` pattern. Consider adding `TriggerOpKind.ofString?` for consistency, especially if more trigger ops are added later.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
```diff
@@ -398,67 +400,69 @@ def handleZeroaryOps {M} [Inhabited M] (name : String)
...
+  | .bool .Not => pure (.not default arg)
   -- Strings and regexes
-  | "Str.Length" => pure (.str_len default arg)
-  | "Str.ToRegEx" => pure (.str_toregex default arg)
-  | "Re.Star" => pure (.re_star default arg)
-  | "Re.Plus" => pure (.re_plus default arg)
-  | "Re.Comp" => pure (.re_comp default arg)
+  | .str .Length => pure (.str_len default arg)
+  | .str .ToRegEx => pure (.str_toregex default arg)
+  | .re .Star => pure (.re_star default arg)
```
MikaelMayer, line 413: Nit: The BV unary ops enumerate 5 specific sizes (1, 8, 16, 32, 64) × 2 ops = 10 cases. This is correct because the CST constructors (`.bvnot`, `.neg_expr`) require a concrete type argument, but it means adding a new BV width requires updating this function. A brief comment noting this constraint would help future maintainers.

shigoel (dismissed):
Looks very nice! Thank you!

Strata/Languages/Core/CoreOp.lean
```diff
@@ -0,0 +1,339 @@
...
+  | .UDiv => "UDiv" | .UMod => "UMod" | .SDiv => "SDiv" | .SMod => "SMod"
+  | .Not => "Not" | .And => "And" | .Or => "Or" | .Xor => "Xor"
+  | .Shl => "Shl" | .UShr => "UShr" | .SShr => "SShr"
+  | .ULt => "ULt" | .ULe => "ULe" | .UGt => "UGt" | .UGe => "UGe"
+  | .SLt => "SLt" | .SLe => "SLe" | .SGt => "SGt" | .SGe => "SGe"
+  | .Concat => "Concat"
+
+instance : ToString BvOpKind := ⟨BvOpKind.toString⟩
+
+def BvOpKind.ofString? : String → Option BvOpKind
```
shigoel, line 159: I wish there was an easy way to keep the `toString` and `ofString?` in sync. (well, there's proof) so that we know at compile time that they're breaking. Not a blocker, just a thought.

Strata/Languages/Core/CoreOp.lean
```diff
@@ -0,0 +1,339 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+/-! ## Structured Operator Types for Strata Core
+
+Replaces string-based operator identity with structured types, enabling
```
joscoh, line 10: Minor, but comments should be stateless

Strata/Languages/Core/CoreOp.lean
```diff
@@ -0,0 +1,339 @@
...
+  | _ => false
+
+def NumericOpKind.isUnary : NumericOpKind → Bool
+  | .Neg => true
+  | _ => false
+
+/-! ### Boolean Operations -/
+
+inductive BoolOpKind where
+  | And | Or | Not | Implies | Equiv
```
joscoh, line 87: Not directly related to this PR, but why do we need `Equiv` when equality is built in already?

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -423,9 +421,7 @@ partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Conte
           Factory.ite abGeZero q negQ
         | _ => Term.app Op.div args retTy
       .ok (divTApp, .int, ctx)
-    -- Truncating modulo: tmod(a,b) = a - b * tdiv(a,b)
-    -- tdiv(a,b) = let q = ediv(abs(a), abs(b)) in ite(a*b >= 0, q, -q)
```
joscoh, line 495: I don't think these comments should be deleted.

Strata/Languages/Python/PythonToCore.lean
```diff
@@ -140,7 +141,7 @@ def handleAdd (translation_ctx: TranslationContext) (lhs rhs: Core.Expression.Ex
     let r_ty := translation_ctx.variableTypes.find? (λ p => p.fst == r.name)
     match l_ty, r_ty with
     | some (_, .tcons "int" []), some (_, .tcons "int" []) =>
-      .app () (.app () (.op () "Int.Add" mty[int → (int → int)]) lhs) rhs
+      .app () (.app () (.op () (Core.CoreOp.toString (.numeric ⟨.int, .Add⟩)) mty[int → (int → int)]) lhs) rhs
```
joscoh: It would be better to have a function to handle this. Something like `LExpr.coreOp : CoreOp -> LExpr` so that we don't need these `toString` calls each time.

StrataTest/Languages/Core/Tests/CoreOpTests.lean
```diff
@@ -0,0 +1,120 @@
...
+
+/-! ## Tests for CoreOp structured operator types -/
+
+namespace Core
+
+/-! ### Round-trip tests: toString ∘ ofString = id for all known operators -/
+
+section RoundTrip
+
+private def checkRoundTrip (name : String) : Bool :=
```
joscoh, line 17: As Shilpi mentioned, this kind of thing could just be proved, but it is not blocking for this PR.

MikaelMayer (commented):
Well-designed refactoring. The `CoreOp` type hierarchy is clean, the `BvOp` structure with `size` as a `Nat` field is the right call for collapsing per-size cases, and the `other` escape hatch preserves backward compatibility. The SMT encoder going from 115 per-size BV cases to 25 size-generic patterns is a clear win. Round-trip tests are thorough.

A few items below, mostly nits.

Strata/Languages/Core/CoreOp.lean
tautschnig: Fixed in 00dd030.

Strata/Languages/Core/CoreOp.lean
tautschnig: 00dd030 documents why `.toString` is needed.

Strata/Languages/Core/CoreOp.lean
tautschnig: Fixed in 5a24d57.

Strata/Languages/Core/DDMTransform/ASTtoCST.lean
tautschnig: Fixed in fdccf38.

Strata/Languages/Core/SMTEncoder.lean
tautschnig: Fixed in 47cf9e6.

Strata/Languages/Python/PythonToCore.lean
tautschnig: Fixed in 31f9322.

Strata/Languages/Core/CoreOp.lean
tautschnig: My understanding, but this could be wrong, is that `Bool.Equiv` is the Boogie-style biconditional (`<==>`) -- it's a binary boolean operator that returns `bool`, whereas the built-in `LExpr.eq` is structural equality that works across all types. They have different SMT encodings too: `Equiv` maps to `Op.eq` restricted to Booleans, while `LExpr.eq` maps to a general SMT equality. This distinction predates this PR -- `CoreOp` just mirrors what was already in the factory (`boolEquivFunc` in `IntBoolFactory.lean`).

Strata/Languages/Core/CoreOp.lean
tautschnig: Proof added in 6c2bcc3e427b25b11d34e14e7398c6739b6eb02b.

StrataTest/Languages/Core/Tests/CoreOpTests.lean
tautschnig: Proof added in 6c2bcc3e427b25b11d34e14e7398c6739b6eb02b - would particularly appreciate a review thereof!

Strata/Languages/Core/CoreOp.lean
```diff
@@ -322,18 +329,86 @@ def CoreOp.ofString (name : String) : CoreOp :=
...
   | .other _ => none
   | op => some op
 
+/-! ### Round-trip proofs: ofString? ∘ toString = some
+
+These ensure that toString and ofString? stay in sync for every
+constructor of each sub-type. -/
+
+theorem BvOpKind.ofString_toString (k : BvOpKind) :
+    BvOpKind.ofString? k.toString = some k := by cases k <;> native_decide
```
joscoh: We should not use `native_decide` in proofs whenever possible. It adds the Lean compiler correctness as an axiom. `decide` is a safer but less aggressive version, not sure if it works here.

tautschnig: I got this partly done, worked in one case but not the other. `CoreOp`-level proofs still require `native_decide` because `CoreOp.ofString` involves string operations (`splitOn`, `startsWith`) that the kernel cannot reduce. We could of course just drop those proofs...

joscoh: I would rather drop the proofs and leave a TODO for now. I believe that better handling of strings is supposed to be coming soon to Lean. @shigoel your thoughts?

shigoel (dismissed):
Thank you!

Strata/Languages/Core/CoreOp.lean
shigoel: I'm also all for leaving a TODO, yes.

joscoh (dismissed):

Strata/Languages/Core/CoreOp.lean
tautschnig: Done in [2d8a80b](https://github.com/strata-org/Strata/pull/923/commits/2d8a80bc175025aa9882029ac3004972bfb2849d).

aqjune-aws (approved):

Strata/Languages/Core/CoreOp.lean
joehendrix: My usual approach here would be to create something like `bvOpKindNames : Std.HashMap BvOpKind String` that maps to pretty printed representation.

We can then create another mapping from String to BVOpKind by folding over `bvOpKindNames` and even a theorem proving every `forall k :BvOpKind, k in bvOpKindNames`.

joscoh (approved):

Strata/Languages/Core/CoreOp.lean
tautschnig: Since this PR is pretty intrusive while the proof itself is not I'll take care of this in follow-up work.

PR #931, opened 2026-04-15
Fix incoherent CallGraph build in ProcedureInlining

aqjune-aws:
CallGraph is stored in CachedAnalyses of the Core transformation monad state (CoreTransformState), and it is supposed to be correctly incrementarily updated during transformations (e.g., when call is eliminated, etc). However, ProcedureInlining was incorrectly updating this.

The reason was as follows: the CachedAnalyses is (correctly) eagerly updated after every call is inlined, whereas the program that ProcedureInlining was actually using was not being updated. ProcedureInlining is supposed to look at the intermediate program after inlinings are happened, because that is exactly the program that CachedAnalyses stores for.

In theory, this can make ProcedureInlining more aggressive: if f1 is calling f2 and f2 is calling f3 and f3 is calling f4,
* After this patch: during one round of ProcedureInlining, if f2 -> f3 in f2 happened to be inlined first, inlining f1->f2 will see the latest update and introduce f1->f4
* Before this patch: one round of ProcedureInlining of f1->f2 would have always introduced f1->f3 because ProcedureInlining always saw the "old" program.

However, in practice, this doesn't matter because we are using the 'roots' mode during bugFinding (and ProcedureInlining is only called from bugFinding mode) and calls whose callers are the roots are only of concerns.

*Issue #, if available:*

*Description of changes:*


Strata/Languages/Core/CallGraph.lean
```diff
@@ -243,6 +243,20 @@ def Program.toProcedureCG (prog : Program) : ProcedureCG :=
...
+/-- Compare two `CallGraph`s for equality by checking that every node has the
+    same callees and callers with the same counts. -/
+def CallGraph.beq (cg1 cg2 : CallGraph) : Bool :=
+  let keys1 := cg1.callees.toList.map Prod.fst |>.mergeSort (· < ·)
+  let keys2 := cg2.callees.toList.map Prod.fst |>.mergeSort (· < ·)
+  if keys1 != keys2 then false
+  else keys1.all fun k =>
+    let c1 := cg1.getCalleesWithCount k |>.toList.mergeSort (·.1 < ·.1)
+    let c2 := cg2.getCalleesWithCount k |>.toList.mergeSort (·.1 < ·.1)
+    c1 == c2
```
tautschnig: **Bug:** The doc comment says "same callees **and callers** with the same counts" but the implementation only compares `callees`. The `callers` map is not checked at all.

This weakens the test — the cached call graph could have incorrect `callers` and the test would still pass. Since `updateCallGraph` in `ProcedureInlining` updates both `callees` and `callers`, this should be verified.

```suggestion
def CallGraph.beq (cg1 cg2 : CallGraph) : Bool :=
  let eqMap (m1 m2 : Std.HashMap String (Std.HashMap String Nat)) : Bool :=
    let keys1 := m1.toList.map Prod.fst |>.mergeSort (· < ·)
    let keys2 := m2.toList.map Prod.fst |>.mergeSort (· < ·)
    keys1 == keys2 && keys1.all fun k =>
      let c1 := (m1[k]?.getD {}).toList.mergeSort (·.1 < ·.1)
      let c2 := (m2[k]?.getD {}).toList.mergeSort (·.1 < ·.1)
      c1 == c2
  eqMap cg1.callees cg2.callees && eqMap cg1.callers cg2.callers
```

Strata/Transform/CoreTransform.lean
```diff
@@ -386,14 +350,37 @@ def runProgram
...
   : CoreTransformM (Bool × Program) := do
   modify (fun σ => { σ with currentProgram := .some p })
-  let (changed, newDecls) ← runProcedures
-    (targetProcList := targetProcList) f p.decls
-  let newProg := { decls := newDecls }
+
+  let mut anyChanged := false
+  let mut newDecls := p.decls
+  for i in [:p.decls.length] do
+    match p.decls[i]? with
```
tautschnig, line 357: Subtle but important: this reads from `p.decls[i]?` (the *original* program), not `newDecls[i]?`. So each procedure is always transformed from its original body, even if earlier procedures in the same pass were updated.

This is correct for the current use case — `inlineCallCmd` reads *other* procedures' bodies from `currentProgram` in state (which *is* updated). But it means the transformation function `f` cannot rely on seeing its *own* procedure's intermediate state from earlier in the same `runProgram` call.

Worth adding a brief comment here, e.g.:
```
-- N.B. We read from the original p.decls, not newDecls, so each procedure
-- is transformed from its original body. The transformation function can
-- see updated *other* procedures via currentProgram in state.
```

Strata/Transform/CoreTransform.lean
```diff
@@ -386,14 +350,37 @@ def runProgram
...
+          })
+    | _ => pure ()
+
+  let newProg : Program := { decls := newDecls }
   modify (fun σ => { σ with
-    currentProgram := .some newProg,
     currentProcedureName := .none
   })
-  return (changed, newProg)
+  return (anyChanged, newProg)
```
tautschnig, line 383: Nit: the old code explicitly set `currentProgram := .some newProg` here. The new code omits this because `currentProgram` is already correct in state (set either at line 353 or in the loop at line 374-376). This is correct but subtle — a reader might wonder whether `currentProgram` is stale after the loop. Consider adding `currentProgram := .some newProg` back for clarity, or at least a comment explaining why it's not needed.

Strata/Transform/CoreTransform.lean
```diff
@@ -108,14 +108,14 @@ structure CoreTransformState where
...
-  -- that currentProgram of this CoreTransformState is updated to be the updated
-  -- program.
+  -- If the transformation is implemented with runProgram or runProgramUntil,
+  -- the currentProgram field will store the latest versions of finished
+  -- procedures, but the procedure that is being updated might have stale
+  -- statements.
+  -- When a transformation is finished, currentProgram must contain the
+  -- program that is fully updated (or it has to be .none).
+  -- Using runProgram/runProgramUntil enforces that currentProgram of this
+  -- CoreTransformState is updated to be the updated program.
```
tautschnig, line 118: The updated comment is much clearer than the old one. One small refinement: "the procedure that is being updated might have stale statements" could be more precise. It's not that the statements are stale — it's that `runStmtsRec` transforms the *original* body of each procedure (read from `p.decls`), not from `currentProgram`. The staleness is by design: each procedure starts from its original body, but `f` can see the latest versions of *other* procedures via `currentProgram`.

StrataTest/Transform/ProcedureInlining.lean
```diff
@@ -440,4 +440,36 @@ info: true, some { callees := Std.HashMap.ofList [("f", Std.HashMap.ofList [("f"
...
+
+/-- After fully inlining the 3-procedure chain, the cached call graph must
+    equal the call graph freshly computed from the output program. -/
+def testThreeChainCG := do
+  let p := translate TestThreeChain
+  let _ ← setCallGraph p
+  let (_, p') ← runProgramUntil (inlineCallCmd) p
+  let cachedCG := (← get).cachedAnalyses.callGraph
+  let freshCG := p'.toProcedureCG
+  return (cachedCG.map (CallGraph.beq · freshCG))
```
tautschnig: Good test — this directly targets the bug. A few suggestions:

1. Consider also testing the `callers` side of the call graph (once `beq` is fixed to check it).

2. A stronger test would verify the *specific* expected call graph shape after inlining, not just that cached == fresh. For example, after fully inlining the chain, all three procedures should have empty callee sets (no calls remain). This would catch bugs where both cached and fresh are wrong in the same way.

3. This is also a good candidate for a **proof** rather than a test. The key property is:

```
theorem callGraph_coherence : ∀ (f : Command → CoreTransformM ...) (p : Program),
  let (_, p', σ') := runProgram f p initialState
  σ'.cachedAnalyses.callGraph = some (p'.toProcedureCG)
```

This would require formalizing the call graph update invariant maintained by `updateCallGraph`. Even a weaker version — proving it for a specific `f` like `inlineCallCmd` — would be valuable. I realize this is a significant undertaking, but it's worth tracking as a future goal since call graph coherence is a correctness-critical invariant.

Strata/Languages/Core/CallGraph.lean
aqjune-aws: Thanks for the catch up. It seems `deriving BEq` was enough, so I chose it. :)

Strata/Transform/CoreTransform.lean
aqjune-aws: I think this is a false alarm because the body of this `for` only updates the procedure of interest, and `i`th declaration of the program was not affected by the previous iterations.

Strata/Transform/CoreTransform.lean
aqjune-aws: I think having the assignment this here could be confusing to the readers too - it was already assigned in the previous for or initialization phase!

Strata/Transform/CoreTransform.lean
aqjune-aws: It seems this is related to https://github.com/strata-org/Strata/pull/931#discussion_r3092032509 . :) The visiting `p.decls[i]` and `currentProgram.decls[i]` will have the exactly same declarations.

StrataTest/Transform/ProcedureInlining.lean
aqjune-aws: Thanks - I fixed item 1 and also item 2 by updating the debugging message.

For 3, I think this is slightly nuanced because `f` cannot be universally quantified - someone can write a wrong transformation `f` that will break this property, so it cannot be quantified. We can only prove the property for a specific transformation.
However, the point is legit and this can be a good target for PBT. :)

tautschnig (approved):

Strata/Languages/Core/CallGraph.lean
```diff
@@ -171,13 +191,15 @@ def buildCallGraph (items : List (String × List String)) : CallGraph :=
     acc.insert name (Std.HashMap.ofList calls.occurrences))
     Std.HashMap.emptyWithCapacity
 
+  let emptyCallerMap := items.foldl (fun acc (name, _) =>
+    acc.insert name ([] : List String)) Std.HashMap.emptyWithCapacity
   let callerMapNodedup :=
     items.foldl (fun acc ⟨caller,callees⟩ =>
       callees.foldl (fun acc' callee =>
```
shigoel, line 198: Nice fix — pre-populating the caller map ensures every procedure has an entry in both `callees` and `callers`. Consider stating this invariant explicitly (e.g. a comment or a `theorem buildCallGraph_keys_eq : (buildCallGraph items).callees.keys == (buildCallGraph items).callers.keys` stub), since the correctness of `computeRoots` and the cache-coherence property depend on it.

Strata/Transform/CoreTransform.lean
```diff
@@ -386,14 +350,37 @@ def runProgram
...
-  let (changed, newDecls) ← runProcedures
-    (targetProcList := targetProcList) f p.decls
-  let newProg := { decls := newDecls }
+
+  let mut anyChanged := false
+  let mut newDecls := p.decls
+  for i in [:p.decls.length] do
+    match p.decls[i]? with
+    | some (.proc proc md) =>
+      let isTargetProc := match targetProcList with
```
shigoel, line 359: Subtle: the loop reads procedure bodies from the original `p.decls` (via `p.decls[i]?`), not from the mutating `newDecls`. So each procedure is always transformed starting from its original body — the eager `currentProgram` update on line 376 only affects what `f` sees when it inspects *other* procedures (e.g., looking up a callee's body during inlining). This is the correct behavior, but it's subtle enough to warrant a brief comment here, since the old comment about staleness was removed and readers might assume the loop re-reads from `newDecls`.

Strata/Transform/CoreTransform.lean
```diff
@@ -386,14 +350,37 @@ def runProgram
...
+        modify (fun σ => { σ with
+          currentProcedureName := .some proc.header.name.1
+        })
+
+        let (changed, new_body) ← runStmtsRec f proc.body
+
+        if changed then
+          newDecls := newDecls.set i (Decl.proc { proc with body := new_body } md)
+          anyChanged := true
+          modify (fun σ => { σ with
```
shigoel, line 374: Soft note: `List.set` is O(i) and `p.decls[i]?` is also O(i), making this loop O(n²) in the number of declarations. At realistic program sizes this is fine, but worth keeping in mind if `Program.decls` ever moves to `Array`.

Strata/Transform/CoreTransform.lean
aqjune-aws: It seems this comment is identical to https://github.com/strata-org/Strata/pull/931#discussion_r3092032509 - I left a reply there.

shigoel (approved):

PR #932, opened 2026-04-15
Add type-safe Laurel expressions and structured warning kinds

joehendrix:
## Summary

The PySpec-to-Laurel translation built Laurel expressions as untyped
`StmtExpr` values, making it easy to silently produce ill-typed Laurel
(e.g., passing an int where a bool was expected). Warnings were plain
strings with no machine-readable category, so downstream tooling could
not filter or summarize them.

- Introduce `TypedStmtExpr tp` / `SomeTypedStmtExpr` so the Lean type
  checker catches type mismatches at translation time rather than
  surfacing them as confusing Laurel or Core errors later in the
  pipeline.
- Refactor `specExprToLaurel` to return `SomeTypedStmtExpr` instead of
  `Option StmtExprMd`. Callers use `runChecked` to detect whether errors
  were reported, replacing the old `some`/`none` pattern. When
  translation fails or produces a non-Bool type, a specific warning is
  emitted (the old code emitted a nondeterministic `.Assert .Hole`).
- Add `SpecExprContext.argTypes` so identifier references are resolved
  against the function's actual parameter types instead of being
  unconditionally assumed `Any`. Unknown identifiers now produce a clear
  `.typeError` warning.
- Add structured `WarningKind` (`phase` + `category`) to `SpecError` so
  warnings can be counted, filtered, and triaged programmatically. ~25
  specific warning kinds covering type translation, precondition issues,
  declaration problems, and overload dispatch.
- Surface `pyspecWarnings` through `PySpecLaurelResult` and centralize
  stderr printing in `pythonAndSpecToLaurel` (including `WarningKind`
  info in output). Remove unused `quiet` parameter from
  `buildPySpecLaurel` and `readDispatchOverloads`.
- Add `--warning-summary` CLI flag that writes a JSON file with per-kind
  warning counts, for CI and benchmark scripts. Report write errors to
  stderr instead of silently ignoring them.
- Fix comparison operation docstrings (`Lt`, `Leq`, `Gt`, `Geq`) to say
  `Int` and `Real` -- `Float64` is not supported in the Core translation.

## Test changes

- Add precondition integration tests that run `signaturesToLaurel` and
  check the formatted Laurel body for correct operations (`|` for or,
  `==>` for implies, `!` for not, `DictStrAny_contains` for dict
  containsKey). These would have caught the bugs fixed in an earlier version
  of this PR where `TypedStmtExpr` wrappers emitted wrong `PrimitiveOp`
  constructors or wrong return types.
- Add `typeError` warning coverage tests (unknown identifier, non-Bool
  precondition formula).
- Add `runTestWarningKinds` helper and tests for every `WarningKind`
  using the `phase.category: message` format.
- Update existing `#guard_msgs` expected output for the new structured
  warning format.
- Update `IdentifyOverloadsTest.lean` for the new `SpecError` constructor
  (adding `WarningKind` field).

### DIFF
--- a/Strata/Languages/Python/PythonLaurelTypedExpr.lean
+++ b/Strata/Languages/Python/PythonLaurelTypedExpr.lean
@@ -1,0 +1,168 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+public import Strata.Languages.Laurel.Laurel
+
+/-!
+# Typed Laurel Expression Builders for the Python Prelude
+
+Type-safe wrappers around the Python Laurel prelude's runtime functions
+(`from_int`, `Any..as_string!`, `DictStrAny_contains`, etc.).  These
+enforce correct `HighType` at construction time so that type mismatches
+are caught by the Lean type checker rather than surfacing as confusing
+errors later in the Laurel or Core pipeline.
+
+The wrapped functions correspond to declarations in
+`PythonRuntimeLaurelPart.lean` and `PythonLaurelCorePrelude.lean`.
+-/
+
+public section
+namespace Strata.Python.Laurel
+
+open Strata.Laurel (HighType HighTypeMd StmtExpr StmtExprMd mkId)
+
+abbrev Md := Imperative.MetaData Core.Expression
+
+abbrev tyAny : HighType := .UserDefined "Any"
+
+/--
+A Laurel `StmtExprMd` tagged with its `HighType`.
+
+The benefit is that composition of typed expressions is checked by the Lean type
+system, catching mismatches like passing an `Int` where a `Bool` is expected.
+
+The type parameter is not statically enforced; correctness depends on the helper
+functions in this namespace producing the right type tag.  After a translation
+error is reported, helpers may return a mistyped expression to allow continued
+translation and further error collection.
+-/
+structure TypedStmtExpr (tp : HighType) where
+  stmt : StmtExprMd
+
+namespace TypedStmtExpr
+
+def ofStmt {tp} (s : StmtExpr) (md : Md) (source : Option FileRange := none) : TypedStmtExpr tp :=
+  { stmt := { val := s, source := source, md := md } }
+
+def identifier (v : String) (tp : HighType) (md : Md)
+    (source : Option FileRange := none) : TypedStmtExpr tp :=
+  .ofStmt (.Identifier (mkId v)) md source
+
+def literalBool (v : Bool) (md : Md)
+    (source : Option FileRange := none) : TypedStmtExpr .TBool :=
+  .ofStmt (.LiteralBool v) md source
+
+def literalInt (v : Int) (md : Md)
+    (source : Option FileRange := none) : TypedStmtExpr .TInt :=
+  .ofStmt (.LiteralInt v) md source
+
+def literalString (v : String) (md : Md)
+    (source : Option FileRange := none) : TypedStmtExpr .TString :=
+  .ofStmt (.LiteralString v) md source
+
+def stringEq (x y : TypedStmtExpr .TString)
+    (md : Md := x.stmt.md) (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
+  .ofStmt (.PrimitiveOp .Eq [x.stmt, y.stmt]) md source
+
+def intGeq (x y : TypedStmtExpr .TInt)
+    (md : Md := x.stmt.md) (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
+  .ofStmt (.PrimitiveOp .Geq [x.stmt, y.stmt]) md source
+
+def intLeq (x y : TypedStmtExpr .TInt)
+    (md : Md := x.stmt.md) (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
+  .ofStmt (.PrimitiveOp .Leq [x.stmt, y.stmt]) md source
+
+def realGeq (x y : TypedStmtExpr .TReal)
+    (md : Md := x.stmt.md) (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
+  .ofStmt (.PrimitiveOp .Geq [x.stmt, y.stmt]) md source
+
+def realLeq (x y : TypedStmtExpr .TReal)
+    (md : Md := x.stmt.md) (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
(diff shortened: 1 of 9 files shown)


Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -483,23 +521,23 @@ def buildSpecBody (preconditions : Array Assertion)
...
-      stmts := assertStmt :: stmts
+    let (⟨condType, condExpr⟩, success) ← runChecked <| specExprToLaurel assertion.formula md ctx
+    if success then
+      if let .TBool := condType then
+        let assertStmt ← mkStmtWithLoc (.Assert condExpr.stmt) default msg
+        stmts := stmts.push assertStmt
+      else
+        reportError .typeError default
+          s!"Precondition expression is not Bool in '{ctx.procName}' (skipping): {msg}"
     idx := idx + 1
```
tautschnig, line 554: **Behavioral change:** The old code emitted `.Assert .Hole` (a nondeterministic assertion) for untranslatable preconditions. The new code silently drops them — when `success` is false or the type isn't `TBool`, no assertion is emitted.

This changes the verification semantics: the old `.Assert .Hole` would generate a nondeterministic proof obligation (which the solver might flag), while the new behavior produces no obligation at all. This means a failed precondition translation now results in a *weaker* verification (fewer obligations) rather than a *conservative* one.

Is this intentional? If so, it's worth a comment explaining the rationale. If not, consider emitting a `.Assert .Hole` as a fallback when translation fails, to preserve the conservative behavior.

Strata/Languages/Python/PythonLaurelTypedExpr.lean
```diff
@@ -0,0 +1,131 @@
...
+open Strata.Laurel (HighType StmtExpr StmtExprMd mkId)
+
+abbrev Md := Imperative.MetaData Core.Expression
+
+abbrev tyAny : HighType := .UserDefined "Any"
+
+/-- A Laurel `StmtExprMd` tagged with its `HighType`, so that type-correctness
+    of expression composition is enforced by the Lean type checker. -/
+structure TypedStmtExpr (tp : HighType) where
+  stmt : StmtExprMd
```
tautschnig: The phantom type parameter is a good idea for catching mismatches at construction time. Worth noting that the safety is "best effort" — `asAny` and `asBool` use `⟨e.stmt⟩` to coerce on type mismatch (after reporting an error), so ill-typed expressions can still propagate after errors. This is fine for the current error-reporting use case, but a reader might expect stronger guarantees from the name "type-safe."

Consider adding a brief doc note: "Type correctness is enforced at construction time; after an error is reported, the wrapper may contain a mistyped expression to allow continued translation."

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -320,133 +336,154 @@ private def mkStmtWithLoc (e : StmtExpr) (loc : SourceRange) (msg : String := ""
...
+  let ctx ← read
+  let (se, success) ← runChecked <| act ctx
+  if !success then
+    return ⟨se.2.stmt⟩
+  match se with
+  | ⟨.UserDefined "Any", e⟩ => pure e
+  | ⟨tp, e⟩ =>
+    let pn := (← read).procName
+    reportError .typeError loc s!"Expected Any-typed expression but got {repr tp} in '{pn}'"
+    pure ⟨e.stmt⟩
```
tautschnig, line 375: The `asAny` and `asBool` helpers duplicate the same pattern: `runChecked`, check success, match on type, report error, coerce. Consider extracting a generic `asTyped` helper:
```lean
private def asTyped (tp : HighType) (loc : SourceRange)
    (act : ToLaurelExprM SomeTypedStmtExpr) : ToLaurelExprM (TypedStmtExpr tp) := do
  let ctx ← read
  let (se, success) ← runChecked <| act ctx
  if !success then return ⟨se.2.stmt⟩
  if se.1 == tp then pure ⟨se.2.stmt⟩
  else do
    reportError .typeError loc s!"Expected {repr tp} but got {repr se.1} in '{(← read).procName}'"
    pure ⟨se.2.stmt⟩
```
Then `asAny loc act = asTyped tyAny loc act` and `asBool loc act = asTyped .TBool loc act`. Not blocking — the current code is clear.

Strata/Languages/Python/PythonLaurelTypedExpr.lean
```diff
@@ -0,0 +1,131 @@
...
+
+def reSearchBool (pattern s : TypedStmtExpr .TString) (md : Md) : TypedStmtExpr .TBool :=
+  .ofStmt (.StaticCall (mkId "re_search_bool") [pattern.stmt, s.stmt]) md
+
+end TypedStmtExpr
+
+/-- Existentially-typed Laurel expression — pairs a `HighType` with a
+    `TypedStmtExpr` of that type, used when the type is determined at runtime
+    (e.g. translating a `SpecExpr` whose type depends on identifier lookup). -/
+abbrev SomeTypedStmtExpr := Σ(tp : HighType), TypedStmtExpr tp
```
aqjune-aws, line 153: !

Strata/Languages/Python/Specs/ToLaurel.lean
joehendrix: Yes it's intentional.  In the PR description, there is a note

> When translation fails or produces a non-Bool type, a specific warning is
emitted (the old code emitted a nondeterministic .Assert .Hole).

Strata/Languages/Python/Specs/ToLaurel.lean
joehendrix: HighType doesn't have BEq iirc.  I could use a callback, but I think the current approach is readable and easily maintained.

Strata/Languages/Python/PythonLaurelTypedExpr.lean
joehendrix: I updated the comment to clarify.

Strata/Languages/Python/PythonLaurelTypedExpr.lean
joehendrix: I'm not sure if there's a question, but I decided to edit the docstring to make it a little more clear what this is for.

aqjune-aws: No, it wasn't a question, sorry :)

aqjune-aws (dismissed):
I ran a sample program with `--warning-summary (some.json)` and it worked well. I found that the warnings are printed to the screen still (which seemed to be consistent with the contents of the json file) - is this expected?

tautschnig (approved):

joehendrix:
Yes, I still expect warnings to be printed.

aqjune-aws (approved):

PR #933, opened 2026-04-15
Make the hashing function of SMT terms faster

aqjune-aws:
Previously, the hash functions of a few datastructures in SMT were first generating their `repr` string, then hashed it. Since this is slow, this pull request adds their simpler implementations.

According to some internal benchmarks, with `--no-solve`, the total amount of time for file writing was significantly reduced.

Previous numbers:
```
    Solver/file writing                      17039       370.4        4912    43.4%
  VC discharge                               24826       539.7        7212    63.2%
SMT verification                             32391       704.2        8197    82.5%
-------------------------------------------------------------------------------------
TOTAL                                        39276       853.8        8484   100.0%
```

After this optimization:
```
    Solver/file writing                       6461       140.5        1783    22.8%
  VC discharge                               14128       307.1        4363    49.8%
SMT verification                             21681       471.3        5236    76.4%
-------------------------------------------------------------------------------------
TOTAL                                        28378       616.9        5357   100.0%
```

Also checked that this did not change the number of verified/unverified programs.


Strata/DL/SMT/TermType.lean
```diff
@@ -89,8 +89,16 @@ instance : LT TermType where
 instance TermType.decLt (x y : TermType) : Decidable (x < y) :=
   if h : TermType.lt x y then isTrue h else isFalse h
 
+def hashTermType : TermType → UInt64
+  | .prim p => mixHash 1 (hash p)
+  | .option t => mixHash 2 (hashTermType t)
+  | .constr id args => mixHash 3 (mixHash (hash id) (hashTermTypeList args))
+where hashTermTypeList : List TermType → UInt64
+  | [] => 7
+  | t :: ts => mixHash (hashTermType t) (hashTermTypeList ts)
```
joehendrix: Maybe just use `args.foldl (init := mixHash 3 (hash id)) fun h a => ...` (or `args.attach.foldl` if you run into termination issues in place of hashTermTypeList)?

Strata/DL/SMT/TermType.lean
```diff
@@ -89,8 +89,16 @@ instance : LT TermType where
...
+
 instance : Hashable TermType where
-  hash := λ a => hash s!"{repr a}"
+  hash := hashTermType
```
joehendrix: It might also be worth seeing if deriving (Hashable TermType) works.

Strata/DL/SMT/Op.lean
```diff
@@ -61,7 +61,7 @@ instance UF.decLt (x y : UF) : Decidable (x < y) :=
   if h : UF.lt x y then isTrue h else isFalse h
 
 instance : Hashable UF where
-  hash := λ a => hash s!"{repr a}"
+  hash a := mixHash (hash a.id) (mixHash (hash a.args) (hash a.out))
```
joehendrix: Can you just derive Hashable for UF now?

aqjune-aws:
Thanks, the patch is much simpler than before.

joehendrix (approved):

Strata/DL/SMT/TermType.lean
```diff
@@ -50,7 +50,7 @@ inductive TermType where
   -- (TODO) It looks like `option` is a special instance of `constr`.
   | option (ty : TermType)
   | constr (id : String) (args : List TermType)
-deriving instance Repr, Inhabited for TermType
+deriving instance Repr, Inhabited, Hashable for TermType
```
tautschnig, line 53: The derived `Hashable` for `TermType` is now structurally generated, while `BEq`/`DecidableEq` is hand-written (via `TermType.beq`). Today these are consistent because `beq` is a structural comparison, but there's no compile-time guard against future drift.

Consider adding a `#guard` test (e.g., in `StrataTest/DL/SMT/`) that constructs a few representative `TermType` values and checks `a == b → hash a == hash b`. Even better would be a `theorem hash_eq_of_beq` proving this, but a test is a pragmatic first step.

This isn't blocking, but worth tracking — if someone later changes `TermType.beq` without updating the hash, `HashMap` lookups will silently break.

tautschnig (approved):

Strata/DL/SMT/TermType.lean
aqjune-aws: We have `DecidableEq TermType` instance in this file, so `TermType.beq` is guaranteed to be correct. :) So it will be fine.

PR #938, opened 2026-04-16
Add running time printer and statistics for Laurel passes

aqjune-aws:
This pull request

* Adds running time printer (using the `profileStep` utility function in Strata) and statistics for Laurel passes
* Prints them when `--profile` is given
* Also does some cosmetic updates like wrapping the Laurel passes in an Array, and adding `#` in front of `derive_prefixed_toString` which was promised in https://github.com/strata-org/Strata/pull/870#discussion_r3089889469 .

```
[statistics] EliminateHoles.holesEliminated: 14                 
[statistics] InferHoleTypes.holesAnnotated: 15                  
```


aqjune-aws:
This pull request was written by an internal request for collecting measures useful for observing how behaviors of passes change on benchmarks after Strata is updated. I relied on an agent to figure out which numbers to report.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -34,18 +36,82 @@ public section
...
+/-- The ordered sequence of Laurel-to-Laurel lowering passes. -/
+private def laurelPipeline : Array LaurelPass := #[
+  { name := "EliminateValueReturns"
+    run := fun p _m =>
+      let (p', diags) := eliminateValueReturnsTransform p
+      (p', diags.toList, {}) },
+  { name := "FilterNonCompositeModifies"
+    run := fun p m =>
+      let (p', diags, stats) := filterNonCompositeModifies m p
+      (p', diags, stats) },
```
tautschnig: **Bug?** The pass order has changed relative to `main`. On `main` the order is:

1. `filterNonCompositeModifies`
2. `eliminateValueReturnsTransform`

Here it's reversed: `EliminateValueReturns` runs first, then `FilterNonCompositeModifies`. Was this intentional?

Also, on `main` `eliminateValueReturnsTransform` had no `emit` call after it (the intermediate state was never written to disk). In the new loop every pass is emitted. This is arguably better, but it's a behavioral change for `--keep-all-files` users — worth mentioning in the PR description if intentional.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -34,18 +36,82 @@ public section
...
+  resolvesAfter : Nat := 0
+  /-- The pass action. -/
+  run : Program → SemanticModel → Program × List DiagnosticModel × Statistics
+
+/-- The ordered sequence of Laurel-to-Laurel lowering passes. -/
+private def laurelPipeline : Array LaurelPass := #[
+  { name := "EliminateValueReturns"
+    run := fun p _m =>
+      let (p', diags) := eliminateValueReturnsTransform p
+      (p', diags.toList, {}) },
```
tautschnig, line 61: `EliminateValueReturns` is the only pass that doesn't return `Statistics` from its underlying function — the pipeline wrapper supplies `{}`. For consistency, consider adding a `Statistics` return to `eliminateValueReturnsTransform` (even if it's just counting procedures processed), or at least leave a `-- TODO` here explaining why it's omitted.

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
```diff
@@ -109,11 +110,25 @@ def eliminateReturnsInExpression (proc : Procedure) : Procedure :=
...
-def eliminateReturnsInExpressionTransform (program : Program) : Program :=
-  { program with staticProcedures := program.staticProcedures.map eliminateReturnsInExpression }
+def eliminateReturnsInExpressionTransform (program : Program) : Program × Statistics :=
+  let procs := program.staticProcedures.map eliminateReturnsInExpression
+  let nFunctional := program.staticProcedures.countP (·.isFunctional)
+  let nSkipped := program.staticProcedures.length - nFunctional
+  let stats := ({} : Statistics)
+    |>.increment s!"{ElimRetStats.functionalProceduresProcessed}" nFunctional
+    |>.increment s!"{ElimRetStats.nonFunctionalProceduresSkipped}" nSkipped
+  ({ program with staticProcedures := procs }, stats)
```
tautschnig: The stats here count procedures from the *input* list (`program.staticProcedures`), but the output `procs` is produced by `map eliminateReturnsInExpression` which applies to every procedure (functional or not — the function itself checks `isFunctional` internally). So `functionalProceduresProcessed` is really "functional procedures in the input" rather than "procedures actually transformed". The name is slightly misleading — consider `functionalProceduresInInput` or adding a doc comment clarifying this.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -501,13 +502,21 @@ def transformProcedure (proc : Procedure) : LiftM Procedure := do
...
 Transform a program to lift all assignments that occur in an expression context.
 -/
-def liftExpressionAssignments (model: SemanticModel) (program : Program) : Program :=
+def liftExpressionAssignments (model: SemanticModel) (program : Program) : Program × Statistics :=
   let initState : LiftState := { model := model }
   let (seqProcedures, _) := (program.staticProcedures.mapM transformProcedure).run initState
-  { program with staticProcedures := seqProcedures }
+  let stats := ({} : Statistics)
+    |>.increment s!"{LiftStats.proceduresProcessed}" seqProcedures.length
+  ({ program with staticProcedures := seqProcedures }, stats)
```
tautschnig: Minor: `proceduresProcessed` counts `seqProcedures.length` (the output), which equals `program.staticProcedures.length` (the input) since `mapM` preserves length. This stat will always be non-zero for any non-empty program, so it's more of a "program size" metric than a "work done" metric. Consider whether something more informative (e.g., number of assignments actually lifted) would be more useful for benchmarking.

Strata/Languages/Laurel/DesugarShortCircuit.lean
```diff
@@ -25,30 +26,39 @@ public section
...
 
-/-- Local rewrite of a single short-circuit node. Recursion is handled by `mapStmtExpr`. -/
-private def desugarShortCircuitNode (model : SemanticModel) (expr : StmtExprMd) : StmtExprMd :=
+inductive DesugarStats where
+  /-- Number of short-circuit operators (AndThen/OrElse/Implies) rewritten to IfThenElse. -/
+  | operatorsDesugared
+
+#derive_prefixed_toString DesugarStats "DesugarShortCircuit"
+
+private abbrev DesugarM := StateM Statistics
```
tautschnig: Nit: the other passes that use `StateM` embed `statistics : Statistics` inside a custom state structure (e.g., `InferHoleState`, `THState`). Here `DesugarM` aliases `StateM Statistics` directly. Both approaches work, but the inconsistency is worth noting — if `DesugarShortCircuit` ever needs additional state, this will need refactoring. Fine for now since the pass is simple.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
aqjune-aws: To avoid conflicts with updates in other Laurel passes, I left stats in eliminateHoles and inferHoleTypes only.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -34,18 +36,77 @@ public section
...
+abbrev TranslateResultWithLaurel := (Option Core.Program) × (List DiagnosticModel) × Program × Statistics
+
+/-- A single Laurel-to-Laurel pass. Each pass receives the current program and
+    semantic model and returns the (possibly modified) program, accumulated
+    diagnostics, and statistics. -/
+structure LaurelPass where
+  /-- Human-readable name, used for profiling and file emission. -/
+  name : String
+  /-- How many times `resolve` should be run after the pass. -/
+  resolvesAfter : Nat := 0
```
keyboardDrummer: Can this be more than `1` ? Why not use a bool?

Strata/Languages/Laurel/InferHoleTypes.lean
```diff
@@ -44,9 +45,18 @@ private def calleeParamTypes (model : SemanticModel) (callee : Identifier) : Opt
   | .staticProcedure proc => some (proc.inputs.map (·.type))
   | _ => none
 
+inductive InferHoleTypesStats where
```
keyboardDrummer, line 48: Is this useful? It looks a bit like a showcase for collecting data.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
aqjune-aws: It seems there were two `resolve` invocations after the previous `modifiesClausesTransform` call. Or should one of these be removed?

```
  let (program, modifiesDiags) := modifiesClausesTransform model program
  let result := resolve program (some model)
  let result := resolve program (some model)
```

Strata/Languages/Laurel/InferHoleTypes.lean
aqjune-aws: I think this can be (somewhat?) useful because it is good to have an idea about how many annotated/unannotated holes there are in benchmarks.

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
aqjune-aws: This and the below diffs are erased in the latest commits because adding statistics to Laurel passes that are not directly related to holes seemed too pervasive for now.

Strata/Languages/Laurel/EliminateHoles.lean
```diff
@@ -75,10 +76,21 @@ After this pass the program contains only non-deterministic `Hole` nodes.
...
+
+#derive_prefixed_toString ElimHoleStats "EliminateHoles"
+
+def eliminateHoles (program : Program) : Program × Statistics :=
   let initState : ElimHoleState := {}
   let (procs, finalState) := (program.staticProcedures.mapM elimProcedure).run initState
-  { program with staticProcedures := finalState.generatedFunctions ++ procs }
+  let stats := ({} : Statistics)
+    |>.increment s!"{ElimHoleStats.holesEliminated}" finalState.counter
+    |>.increment s!"{ElimHoleStats.functionsGenerated}" finalState.generatedFunctions.length
```
shigoel: Nit: by construction each eliminated hole produces exactly one generated function, so `functionsGenerated == holesEliminated` should always hold. Consider stating that as an `assert` (or at minimum a comment) so a future change that breaks the 1-to-1 relationship is caught early.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -65,54 +126,33 @@ private def runLaurelPasses (options : LaurelTranslateOptions) (program : Progra
...
+  let mut program := program
+  let mut model := model
+  let mut allDiags : List DiagnosticModel := resolutionErrors ++ diamondErrors
+  let mut allStats : Statistics := {}
+
+  for pass in laurelPipeline do
+    let (program', diags, stats) ← profileStep options.profile s!"  {pass.name}" do
+      pure (pass.run program model)
+    program := program'
+    allDiags := allDiags ++ diags
```
shigoel, line 146: Soft note: `allDiags := allDiags ++ diags` is left-nested List append inside the loop (O(n) per iteration). With 11 passes and typically small diagnostic lists this is fine, but if diagnostic counts grow consider switching to `Array` accumulation with `Array.append`.

Strata/Languages/Laurel/EliminateHoles.lean
aqjune-aws: Good catch, thank you.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
aqjune-aws: Yes, `diags` will be pretty small. :) It would be fine.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
keyboardDrummer: One should be removed. The bool can be named something like `introducesUnresolvedAstNodes: Bool` or `requiresReresolving: Bool`

Strata/Languages/Laurel/InferHoleTypes.lean
keyboardDrummer: What what you do with that information?

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -7,6 +7,7 @@ module
 
 public import Strata.Languages.Laurel.MapStmtExpr
 public import Strata.Languages.Laurel.Resolution
+public import Strata.Util.Statistics
```
robin-aws: Unused import?

#966 :)

StrataMain.lean
```diff
@@ -579,10 +579,13 @@ def pyAnalyzeLaurelCommand : Command where
...
 
-    let (coreProgramOption, laurelTranslateErrors, _loweredLaurel) ←
+    let (coreProgramOption, laurelTranslateErrors, _loweredLaurel, laurelPassStats) ←
       profileStep profile "Laurel to Core translation" do
         Strata.translateCombinedLaurelWithLowered combinedLaurel
-          (keepAllFilesPrefix := keepPrefix)
+          (keepAllFilesPrefix := keepPrefix) (profile := profile)
+
+    if profile && !laurelPassStats.data.isEmpty then
+      let _ ← (IO.println laurelPassStats.format |>.toBaseIO)
```
robin-aws:
```suggestion
      IO.println laurelPassStats.format
```

robin-aws (changes requested):

Strata/Languages/Laurel/EliminateHoles.lean
```diff
@@ -75,10 +76,18 @@ After this pass the program contains only non-deterministic `Hole` nodes.
 
 Assumes `inferHoleTypes` has already annotated holes with types.
 -/
-def eliminateHoles (program : Program) : Program :=
+inductive ElimHoleStats where
```
MikaelMayer: Nit: The docstring above (lines 71-78) describes `eliminateHoles`, but it's now attached to `ElimHoleStats` instead. Consider moving the inductive + `#derive_prefixed_toString` above the docstring, or splitting the docstrings so each definition has its own.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
MikaelMayer: Nit: This `public import Strata.Util.Statistics` appears unused — `Statistics` isn't referenced anywhere in this file.

MikaelMayer (commented):
Clean refactoring of the Laurel pipeline into a declarative `LaurelPass` array — this is a nice improvement that reduces duplication and makes the pass ordering easy to read. The statistics plumbing and new tests look solid. A few minor items below.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
aqjune-aws: Thanks, I updated it to call `resolve` only once & made the flag a boolean type.

Strata/Languages/Laurel/InferHoleTypes.lean
aqjune-aws: I think the user can get an idea about whether type inference in Laurel is improving on a particular benchmark of interest. Since holes are truly unknown symbolic values, they cannot be easily simplified during partial evaluation in Core, so understanding their portfolio seems important.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
aqjune-aws: Thanks, I removed it. :)

StrataMain.lean
aqjune-aws: Updated it, thanks.

Strata/Languages/Laurel/EliminateHoles.lean
aqjune-aws: Good catch, fixed.

shigoel (approved):
All prior feedback from reviewers has been addressed. The LaurelPass array is a clean refactor, statistics plumbing is non-invasive, and tests are solid. LGTM.

robin-aws (approved):

Issue #952, opened 2026-04-16
Sharing cached .lake directory can lead to incorrect CI results


robin-aws:
(Split this off from #914 - addressing that would make this problem worse since it will introduce more Lean tests that depend on external Python files, but it's independent. There's some related discussion from [this comment](https://github.com/strata-org/Strata/issues/914#issuecomment-4255967371) on)

Lake's incremental build system only pays attention to the content of Lean files and their dependencies. That means building a test like [PropertySummaryTest.lean](https://github.com/strata-org/Strata/blob/main/StrataTest/Languages/Python/PropertySummaryTest.lean) for example will repeat a cached result from the incremental build trace, even if the external resources it depends on (python and cvc5) have changed since the last run. The fact that we are reusing cached .lake directories between builds, and across branches/commits, means it is very easy for CI to report stale results, so this is ultimately a soundness issue.

The short-term fix will be to disable the caching in the merge queue, as a last stop gap against stale builds. 

In the longer team, I'm thinking that because Lean explicitly tracks anything that might read external state (i.e. it has to be something like an #eval on some monad type), it's actually tractable for the building process to just record that the build MIGHT have depended on external state, and therefore that the module should always be rebuilt without using the cache. It seems like a fairly small change and would enable provably safe Lean build caching, so I'll see if can get a POC working and potentially cut a Lean RFC issue.

joehendrix:
I posted the same comment to #953, but is there a Lean4 issue related to this?

This seems like it potentially affects all software development projects using Lean with any external dependencies.  e.g., the Lean4 action should be safe by design.

robin-aws:
Just cut a Lean issue for this: https://github.com/leanprover/lean4/issues/13449

PR #953, opened 2026-04-16
Only use Lean caches on PRs

robin-aws:
*Issue #, if available:*
#952 (and now logged as a Lean bug: https://github.com/leanprover/lean4/issues/13449)

*Description of changes:*
Disables the main Lake cache restore on the merge queue and on pushes to main, to avoid the possibility of main being in a bad state because of incorrect CI results. The subsequent restores on child workflows are safe and save a lot of Lean build time, so they are still unconditionally enabled.

See the successful build on this commit for assurance that this should also build correctly when it hits the merge queue: https://github.com/strata-org/Strata/pull/953/checks?sha=40814092b217640935ad2aeb8cd6b638a0acc407


tautschnig (dismissed):

joscoh (dismissed):

tautschnig:
Ah, we will now actually need to explicitly pass on selected build artifacts that were previously handed on via the cache.

robin-aws:
> Ah, we will now actually need to explicitly pass on selected build artifacts that were previously handed on via the cache.

Aha as it actually turns out, it's not that artifacts like the ION jar or the `strata` exe are being passed on via the cache, it's that the tests that depend on them were not being run AT ALL in the later workflows, because the `.lake` cache indicated they were already built. In other words this pointed to even more examples of the incorrectly-cached Lean tests I'm worried about in #952!

.github/workflows/ci.yml
```diff
@@ -255,6 +262,9 @@ jobs:
       run: pip install .
       working-directory: Tools/Python
     - name: Restore Lean build cache
+      # Only use the caches on PRs because there is a risk of stale results:
+      # https://github.com/strata-org/Strata/issues/952
+      if: github.event_name == 'pull_request'
```
tautschnig: Don't we need to change something around here as that was where the merge queue previously failed?

robin-aws: Nope! The merge queue failed on the child workflows because I wasn't letting them have the cache. I've changed that part back and https://github.com/strata-org/Strata/pull/953/checks?sha=40814092b217640935ad2aeb8cd6b638a0acc407 shows it should build now even when the main workflow doesn't restore the cache.

joehendrix:
Is there a concrete build failure or other test case so I can understand the issue here?  Alternatively, is there discussion about this on the Lean Zulip?

I'd like to get more eyes on the problem being solved by disabling caches since the bug this fixes clearly potentilly affects all projects using Lean.

robin-aws:
There is a bit more detail on the problem in #952.

Completely agree this is a general Lean problem, and I was planning to cut a Lean4 issue. I'll link it here and engage with the Lean Zulip for discussion. That issue will include a simpler description of the problem as well.

MikaelMayer (dismissed):

joscoh (dismissed):

joscoh (approved):

MikaelMayer (approved):

PR #963, opened 2026-04-17
Fix Python->Laurel translation of `for` loop

thanhnguyen-aws:
*Issue #, if available:* 962

*Description of changes:* Previously, `for` loop does not havoc the variables which are modified inside the loop's body, which leads to incorrect verification results (in many expected tests). 

Solution: Transform the `for` loop into an equivalent `while` loop with a counter variable. 

Note that many VC results in some expected tests change from `PASS` to `unknown`.  Although those results should be `PASS`, we cannot verify them without loop invariants, so the correct results at the current stage should be `unknown`.

Minor additional support to make relevant expected tests consistent: translating Python Tuple into ListAny.


StrataTest/Languages/Python/run_py_analyze_sarif.py
```diff
@@ -58,6 +58,10 @@
     "test_subscription",
     "test_try_except",
     "test_try_except_scoping",
+    "test_tuple_creat",
```
shigoel: Typo: `"test_tuple_creat"` should be `"test_tuple_create"` (missing `e`). This means the test won't be matched by the skip/include logic.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -536,6 +536,12 @@ function Any_sets! (indices: ListAny, dictOrList: Any, val: Any): Any
     Any_sets!(ListAny..tail!(indices), Any_get!(dictOrList, ListAny..head!(indices)), val))
 };
 
+function Any_len (v: Any) : int;
+
+function Any_len_to_Any (v: Any) : Any {
```
shigoel, line 540: `Any_len` is completely abstract with no axioms. The while-loop guard is `counter < Any_len(v)`, but without a non-negativity axiom, the solver may assign negative values to `Any_len(v)`, causing sound-but-unnecessary `unknown` results.

Consider adding a positivity procedure similar to `List_len_pos`:

```
procedure Any_len_pos(v: Any)
  invokeOn Any_len(v)
  ensures Any_len(v) >= 0;
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1632,10 +1637,18 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-    let bodyStmts := assumeStmts ++ bodyStmts
+    let counterLtLen := match iterExpr.val with
+      | .StaticCall "range" (startExpr::_) =>
+          mkStmtExprMd $ .PrimitiveOp .Lt [counterVar,
+                          mkStmtExprMd $ .StaticCall "Any..as_int!" [startExpr]]
+      | _ =>
+          mkStmtExprMd $ .PrimitiveOp .Lt [counterVar,
+                          mkStmtExprMd $ .StaticCall "Any_len" [iterExpr]]
+    let bodyStmts := assumeStmts ++ bodyStmts ++ [counterIncrease]
     let innerBlock := mkStmtExprMd (StmtExpr.Block bodyStmts (some continueLabel))
```
shigoel, line 1689: Two observations:

1. **Looser pattern than `assumeStmts`**: This matches `.StaticCall "range" (startExpr::_)` while the `assumeStmts` block above (line 1615) matches `(startExpr::stopExpr::stepExpr::_)`. If `range` translation ever emits fewer or more arguments, these two blocks could diverge. Consider aligning the patterns.

2. **Undocumented tuple change**: The tuple translation change (`.Tuple .. => .Hole` → `.Tuple _ elems _ => translateList ...`) at line 715 is a distinct feature from the for-loop fix. Consider mentioning it in the PR description so reviewers know it's intentional.

thanhnguyen-aws: 1. The `if` statement  after the match :`(startExpr::stopExpr::stepExpr::_)` will throw an exception if `stopExpr` or `stepExpr` are not None. Therefore if the translator can reach to the later match, the two arguments are guaranteed to be None, we don't need them in the `match`. 
2. I added that change in the PR description.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1602,9 +1602,14 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     let sr := target.ann
```
shigoel, line 1645: Stale comment — this no longer describes the implementation. Now that the for-loop is translated to a while-loop with a counter, consider updating to describe that transformation. Also consider stating the key soundness property: the while-loop translation over-approximates all reachable states of the Python for-loop at loop exit.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1632,10 +1637,18 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
             let assumeInStmt := mkStmtExprMdWithLoc (.Assume (Any_to_bool targetInIter)) md
             pure [assumeInStmt]
       | _ => pure []
-    let bodyStmts := assumeStmts ++ bodyStmts
+    let counterLtLen := match iterExpr.val with
+      | .StaticCall "range" (startExpr::_) =>
+          mkStmtExprMd $ .PrimitiveOp .Lt [counterVar,
+                          mkStmtExprMd $ .StaticCall "Any..as_int!" [startExpr]]
+      | _ =>
+          mkStmtExprMd $ .PrimitiveOp .Lt [counterVar,
```
shigoel, line 1686: `emptyRangeExit` (generated by `assumeStmts` for range loops, line ~1619 on main) is now dead code. The while guard `counter < n` already prevents entry when `n <= 0`, so the `if n <= 0 then exit breakLabel` check inside the body can never fire. Consider removing it to avoid dead instructions in the generated VCs.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1632,10 +1637,18 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+                          mkStmtExprMd $ .StaticCall "Any_len" [iterExpr]]
+    let bodyStmts := assumeStmts ++ bodyStmts ++ [counterIncrease]
     let innerBlock := mkStmtExprMd (StmtExpr.Block bodyStmts (some continueLabel))
-    let loopBlock := mkStmtExprMdWithLoc (StmtExpr.Block [innerBlock] (some breakLabel)) md
-    return (finalCtx, (getExceptionAssertions ctx iterExpr) ++ targetDecls ++ [loopBlock])
+    let loopStmt := mkStmtExprMdWithLoc (StmtExpr.While counterLtLen [] none innerBlock) md
+    let loopBlock := mkStmtExprMdWithLoc (StmtExpr.Block [loopStmt] (some breakLabel)) md
```
shigoel, line 1691: Worth confirming that `assumeStmts` (target-in-range / target-in-iter assumptions) interact correctly with the verifier's while-loop havocking semantics. If the verifier re-havocs the target at the loop head on each iteration, these assumptions correctly constrain the fresh havoc'd value. If the verifier does *not* re-havoc the target (because it's not assigned inside the loop body), then the target retains its value from the previous iteration and the assumptions are vacuously repeated — still sound, but the target value could be stale for iterations 2+.

thanhnguyen-aws: I moved target havoc stmts to the top of loop body (line 1647 , 1651)

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I edited the comment. Could you please give me an example where the while-loop translation strictly over-approximates the for-loop?

shigoel: Concrete example of over-approximation:

```python
def main() -> None:
    total: int = 0
    for x in [1, 2, 3]:
        total += x
    assert total == 6  # provably true in Python
```

At loop exit the verifier knows `0 <= counter` and `counter == Any_len([1,2,3])`, but `x` and `total` have been havoc'd on each iteration (no loop invariant to preserve their relationship), so `total == 6` is `unknown` — sound but incomplete. A loop invariant like `total == sum([1,2,3][:counter])` would recover the result, but without one the while-loop encoding necessarily over-approximates the set of reachable states at loop exit.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -712,7 +712,7 @@ partial def translateExpr (ctx : TranslationContext) (e : Python.expr SourceRang
 
   -- Tuple literal: (1, 2)
   -- Abstract: return havoc'd tuple (sound abstraction)
-  | .Tuple .. => return mkStmtExprMd .Hole
+  | .Tuple _ elems _ => translateList ctx elems.val.toList
```
shigoel, line 712: Stale comment — now that tuples are translated via `translateList` (producing a concrete `ListAny`), the "return havoc'd tuple (sound abstraction)" description no longer applies. Consider updating to something like "Translate tuple literal as ListAny".

Not a blocker.

Strata/Languages/Python/PythonToLaurel.lean
shigoel: Perhaps this is alright for now, but needs to be documented in any case. I'll leave you to decide, @thanhnguyen-aws.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -536,6 +536,16 @@ function Any_sets! (indices: ListAny, dictOrList: Any, val: Any): Any
     Any_sets!(ListAny..tail!(indices), Any_get!(dictOrList, ListAny..head!(indices)), val))
 };
 
+function Any_len (v: Any) : int;
```
keyboardDrummer, line 538: Why do we have these uninterpreted functions? Can't we have something that reads like it could be executed?

keyboardDrummer (commented):
> Solution: Transform the for loop into an equivalent while loop with a counter variable.

Could you add a for loop to Laurel? Seems useful.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I replaced the havoc statements of `target` variables (`x` in your example) by assigning them to (`Any_iter_index (iter, counter)`). If we have correct model for `Any_len` and `Any_iter_index`, we don't loose information when converting `for-loop` to `while-loop` (we don't havoc anything at translation phase). I added some comments about incompleteness.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
thanhnguyen-aws: At current stage, even with those functions interpreted, we cannot reason about post-loop stage without loop-invariant. I will add the bodies (or axioms) for them when they are necessary to target some verification problems.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1600,11 +1602,17 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-    let holeRhs := expr.Constant sr (constant.ConEllipsis sr) ⟨sr, none⟩
-    let (bodyCtxNoLabels, targetDecls, _) ← translateAssign ctx target none holeRhs md
+    let counterName := s!"@for_loop_counter_{s.toAst.ann.start.byteIdx}"
+    let counterVar := freeVar counterName
+    let counterDecl := mkStmtExprMd $ .LocalVariable counterName (mkHighTypeMd $ .TInt) (mkStmtExprMd $ .LiteralInt 0)
+    let counterIncrease := mkStmtExprMd $ .Assign [counterVar] (mkStmtExprMd $ .PrimitiveOp .Add [counterVar, mkStmtExprMd $ .LiteralInt 1])
+    let indexRhs := expr.Call sr (.Name sr {val:= "Any_iter_index", ann:= sr} default)
+                        {val:= #[iter, .Name sr {val:= counterName, ann:= sr} default], ann:= sr} {val:= #[], ann:= sr}
+    let (bodyCtxNoLabels, targetDecls, _) ← translateAssign ctx target none indexRhs md
     let bodyCtx := { bodyCtxNoLabels with
```
shigoel, line 1654: The synthetic `expr.Call` node here round-trips through `translateAssign` → `translateExpr` to resolve `Any_iter_index`. If `translateExpr` doesn't recognize `Any_iter_index` (it's not in `importedSymbols` or `functionSignatures`), this will silently translate to `.Hole`, losing the index semantics entirely.

Consider either:
1. Directly constructing the Laurel `StmtExpr.StaticCall "Any_iter_index" [...]` instead of synthesizing a Python AST node, or
2. Adding a comment documenting the dependency on `translateExpr`'s handling of this function name.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1589,9 +1586,14 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-  -- For loop: for target in iter: body (target may be any assignment target)
-  -- Abstract: execute body once with havoc'd target, then havoc all modified variables
-  -- This is sound: if there are 0 iterations, we havoc; if >0, we execute once and havoc
+  -- For loop is translated into a while loop:
+  -- for x in iter : \n body
+  -- is translated into
+  -- @for_loop_counter_xxx = 0
+  -- while (@for_loop_counter_xxx < Any_len(iter)):
+  --  body
+  --  @for_loop_counter_xxx += 1
```
shigoel, line 1625: Nit: Consider stating the intended invariant for this translation, e.g.:

```
-- Invariant: for `for x in iter: body`, the emitted while loop visits
-- each index 0 ≤ i < len(iter) exactly once with x = iter[i], and all
-- variables modified in body are properly havocked by while semantics.
```

This would help future readers verify coverage of edge cases (empty iterators, break/continue, nested loops).

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1616,26 +1624,30 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
-              (mkStmtExprMd $ .Exit breakLabel)
-              none) md
             let assumeTypeInt := mkStmtExprMdWithLoc (.Assume $ mkStmtExprMd (.StaticCall "Any..isfrom_int" [targetVar])) md
             let asIntTarget := mkStmtExprMd $ .StaticCall "Any..as_int!" [targetVar]
             let inRangeExpr := mkStmtExprMd $ .PrimitiveOp .And [
                   (mkStmtExprMd $ .PrimitiveOp .Geq [asIntTarget, mkStmtExprMd $ .LiteralInt 0]),
                   (mkStmtExprMd $ .PrimitiveOp .Lt [asIntTarget, asIntStart]) ]
             let assumeInRange := mkStmtExprMdWithLoc (.Assume inRangeExpr) md
-            pure [emptyRangeExit, assumeTypeInt, assumeInRange]
+            pure [assumeTypeInt, assumeInRange]
```
shigoel, line 1675: Nit: In the `range` case, `counterLtLen` uses `startExpr`, but this is actually the *bound* (upper limit), not the start — `range(n)` starts at 0. The confusing name is inherited from the existing destructuring `(startExpr::stopExpr::stepExpr::_)` where `range(n)` maps `n` to the first position. A clarifying comment here (or renaming upstream) would help readability.

shigoel (commented):
Review from Claude Code — 2 potential correctness issues, 2 nits.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1600,11 +1605,18 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+    let counterIncrease := mkStmtExprMd $ .Assign [counterVar] (mkStmtExprMd $ .PrimitiveOp .Add [counterVar, mkStmtExprMd $ .LiteralInt 1])
+    let indexRhs := expr.Call sr (.Name sr {val:= "Any_iter_index", ann:= sr} default)
+                        {val:= #[iter, .Name sr {val:= counterName, ann:= sr} default], ann:= sr} {val:= #[], ann:= sr}
+    -- Any_iter_index is defined in PythonRuntimeLaurelPart, so indexRhs would be translated into .StaticCall "Any_iter_index" ..., hot .Hole
+    let (bodyCtxNoLabels, targetDecls, _) ← translateAssign ctx target none indexRhs md
     let bodyCtx := { bodyCtxNoLabels with
       loopBreakLabel := some breakLabel
-      loopContinueLabel := some continueLabel }
+      loopContinueLabel := some continueLabel
+      variableTypes := bodyCtxNoLabels.variableTypes ++ [(counterName, "int")]}
```
shigoel, line 1657: The comment here says `Any_iter_index` would be translated into `.StaticCall`, not `.Hole` — but this is incorrect.

`Any_iter_index` is declared as an **external** Laurel function (no body, just `function Any_iter_index(...) : Any;`). In `PreludeInfo.ofLaurelProgram`, the `functions` field is built by:
```lean
let funcNames := prog.staticProcedures.filterMap fun p =>
  if p.body.isExternal || !p.isFunctional then none else some p.name.text
```
This **excludes** external functions. So `Any_iter_index` never enters `merged.functions`, never gets registered in `importedSymbols` (PySpecPipeline.lean:281-282), and `hasModel ctx "Any_iter_index"` returns `false`. `translateCall` then returns `.Hole`.

Compare with `Any_len_to_Any`, which works because it has a body (`{ from_int(Any_len(v)) }`) and is therefore not external.

The tests still pass because `.Hole` is a sound over-approximation (havoc the target), but the index semantics (`x = iter[counter]`) are silently lost, defeating the purpose of the while-loop encoding.

**Fix options**:
1. Give `Any_iter_index` a non-external body (e.g., define it in terms of `Any_get!` or similar), or
2. Change `PreludeInfo.ofLaurelProgram` to include external functions in `functions`, or
3. Bypass `translateCall` entirely and directly construct `StmtExpr.StaticCall "Any_iter_index" [iterExpr, counterVar]` instead of synthesizing a Python AST node.

shigoel (commented):
Re-review after updates — the invariant and boundExpr rename look good. One remaining bug.

Strata/Languages/Python/PythonToLaurel.lean
shigoel: @thanhnguyen-aws You can defer this to a follow-up PR, if you like.

shigoel (dismissed):

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -536,6 +536,18 @@ function Any_sets! (indices: ListAny, dictOrList: Any, val: Any): Any
...
+  from_int(Any_len(v))
+};
+
+procedure Any_len_pos(v: Any)
+  invokeOn Any_len(v)
+  ensures Any_len(v) >= 0;
+
+function Any_iter_index(iter: Any, index: int) : Any;
+
 function PIn (v: Any, dictOrList: Any) : Any
```
tautschnig, line 550: `Any_iter_index` is uninterpreted (no body), which means it can return any value — including an exception. In Python, indexing an iterable can raise `IndexError` or `StopIteration`. Since `Any_iter_index` is not in `AnyMaybeExceptionList`, exception checks won't be generated at call sites.

This is a soundness gap: if the loop body depends on the target not being an exception, the analysis could miss a real error. Consider either adding `Any_iter_index` to `AnyMaybeExceptionList` or documenting why it's safe to omit.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1616,26 +1628,30 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
           | _ =>
             let targetInIter := mkStmtExprMd (.StaticCall "PIn" [targetVar, iterExpr])
             let assumeInStmt := mkStmtExprMdWithLoc (.Assume (Any_to_bool targetInIter)) md
             pure [assumeInStmt]
       | _ => pure []
-    let bodyStmts := assumeStmts ++ bodyStmts
+    let counterLtLen := match iterExpr.val with
+      | .StaticCall "range" (boundExpr::_) =>
+          mkStmtExprMd $ .PrimitiveOp .Lt [counterVar,
+                          mkStmtExprMd $ .StaticCall "Any..as_int!" [boundExpr]]
```
tautschnig, line 1684: The `counterLtLen` condition matches on `iterExpr.val` to distinguish `range(n)` from general iterables. For `range`, it uses `Any..as_int!(boundExpr)` directly; for others, it uses `Any_len(iterExpr)`.

This is correct but fragile — it depends on `range` being translated to a `.StaticCall "range"` at this point. If the range translation changes upstream, this match would silently fall through to the `Any_len` case, which would be incorrect for `range` (since `range` is not a collection with a `.len()` method). Do we have a test exercising this coupling so that it would fail if it ever changed?

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
keyboardDrummer: I'm not sure I follow. My comment is that we should design everything for executability straight from the start, and this does not look executable.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I added a check at line 1633-1637

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
thanhnguyen-aws: I added some doc at line 1627 - 1629

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: I checked and confirmed that `Any_iter_index` is translated to itself (not Hole) at the final `Core` program, and it is included in `funcNames `. I believe the reason is that the empty body without the `external` keyword is not External.

thanhnguyen-aws:
> > Solution: Transform the for loop into an equivalent while loop with a counter variable.
> 
> Could you add a for loop to Laurel? Seems useful.

That is a good idea but I will leave it for a future PR as it requires more research. The main purpose of this PR is just to fix the bug that the current translation of `for` loop does not havoc the modified variables.

tautschnig (approved):

shigoel (approved):
Looks good — the while-loop desugaring is a solid correctness fix. Two minor nits below (neither is a blocker).

PR #968, opened 2026-04-17
Replace SpecAtomType.pyClass with .ident, add BinaryIO, and align type mappings

joehendrix:
Replace the `SpecAtomType.pyClass` variant with `.ident` using qualified `PythonIdent`, add `typing.BinaryIO` support, and align PySpec→Laurel type mappings with PythonToLaurel so non-primitive builtins map to `Any` instead of lossy approximations.

### 1. Fix missing source locations in PySpec translation errors (`SourceRange.lean`, `Specs.lean`)

`SourceRange.format` crashed when called with a `none` range because it unconditionally called `fm.toPosition` on zero offsets. Now it guards on `isNone` and renders `"path:unknown"` instead. Several pattern matches in `Specs.lean` were extracting annotated values via `name.val`/`name.ann` but discarding the source location — these now destructure as `⟨_, name⟩` at the binding site and pass the parent `loc` to error reporters, ensuring translation errors always have usable locations.

### 2. Remove `SpecAtomType.pyClass` variant (`Decls.lean`, `DDM.lean`, `Specs.lean`, `ToLaurel.lean`)

`pyClass` stored an unqualified class name string, which was ambiguous across modules. All construction sites now use `.ident` with a fully qualified `PythonIdent` (module + name). DDM deserialization of legacy `typeClass`/`typeClassNoArgs` Ion payloads produces `.ident` with empty `pythonModule` for backward compatibility. All match arms and comparison logic for `.pyClass` are removed.

### 3. Add `typing.BinaryIO` support (`Decls.lean`, `Specs.lean`, `ToLaurel.lean`)

`BinaryIO` appears in boto3 stubs (e.g., `lambda_/client.py`'s `invoke` returns `StreamingBody` typed as `BinaryIO`). Without this, pySpecs throws "BinaryIO is not defined in module typing". Added `PythonIdent.typingBinaryIO`, registered it in `preludeSig`, and mapped it to `Any` in `knownIdentTypes`.

### 4. Align `knownIdentTypes` with PythonToLaurel (`ToLaurel.lean`)

Previously, PySpec→Laurel mapped types more aggressively than PythonToLaurel: `bytes`/`bytearray`→`TString`, `complex`→`TReal`, `dict`→`DictStrAny`, `Exception`→`Error`, `typing.List`→`ListStr`, `typing.Dict`→`DictStrAny`. PythonToLaurel maps all of these to `Any`. The mismatch meant pyspec-dispatched analysis could generate spurious preconditions (e.g., string-length bounds on `bytes` parameters). Now both paths agree: only `int`, `str`, `bool`, `float`, and `None` get concrete Laurel types; everything else is `Any`.

### DIFF
--- a/Strata/Languages/Python/Specs.lean
+++ b/Strata/Languages/Python/Specs.lean
@@ -410,6 +410,10 @@ def checkEq {α : Type} (loc : SourceRange) (name : String) (as : Array α) (n :
     pure none
 
 def valueAsType (loc : SourceRange) (v : SpecValue) : PySpecM SpecType := do
+  let loc := if loc.isNone then
+                @panic _ ⟨loc⟩ s!"Empty location"
+              else
+                loc
   match v with
   | .typeValue itp =>
     pure itp
@@ -423,6 +427,7 @@ def valueAsType (loc : SourceRange) (v : SpecValue) : PySpecM SpecType := do
     | some (.typeValue tp) =>
       return tp
     | _ =>
+      specError loc s!"Unknown type in valueAsType {repr val}."
       recordTypeRef loc val
       return .ofAtom loc (.pyClass val #[])
   | _ =>
@@ -525,7 +530,7 @@ def translateCall (loc : SourceRange) (func : SpecValue)
     specError loc s!"Unknown call {repr func}."
     return default
 
-def translateConstant (value : constant SourceRange) : PySpecM SpecValue := do
+def translateConstant (loc : SourceRange) (value : constant SourceRange) : PySpecM SpecValue := do
   match value with
   | .ConFalse .. =>
     return .boolConst false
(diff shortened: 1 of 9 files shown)


Strata/Languages/Python/Specs.lean
```diff
@@ -410,6 +411,10 @@ def checkEq {α : Type} (loc : SourceRange) (name : String) (as : Array α) (n :
     pure none
 
 def valueAsType (loc : SourceRange) (v : SpecValue) : PySpecM SpecType := do
+  let loc := if loc.isNone then
```
shigoel: This `@panic` will crash at runtime if a `none` location reaches `valueAsType`. If this is meant as a debug assertion to catch a "should never happen" case, consider using `assert! !loc.isNone` (which is already used elsewhere in this file, e.g., line 151 `assert! !loc.isNone`) — it gives the same crash behavior in debug but the intent is clearer. Alternatively, if there's a realistic path where `loc.isNone` could be true, a `dbg_trace` + fallback would avoid a production crash.

joehendrix: That was just debugging code to help diagnose when valueAsType was getting empty locations.  I removed it.

shigoel (approved):

MikaelMayer (approved):

StrataTest/Languages/Python/ToLaurelTest.lean
```diff
@@ -502,36 +502,26 @@ body contains FieldSelect: false
 
 /-! ## Warning kind tests -/
 
--- Type translation: unsupportedGenericClass
+-- bytes, bytearray, complex now map to Any (matching PythonToLaurel)
```
MikaelMayer, line 505: Nit: "now" references the previous state. A stateless version would be:
```
-- bytes, bytearray, complex map to Any (matching PythonToLaurel)
```

Strata/Languages/Python/Specs.lean
```diff
@@ -424,7 +425,9 @@ def valueAsType (loc : SourceRange) (v : SpecValue) : PySpecM SpecType := do
       return tp
     | _ =>
       recordTypeRef loc val
-      return .ofAtom loc (.pyClass val #[])
+      let mod := toString (← read).currentModule
```
MikaelMayer, line 428: Nit: This `let mod := toString (← read).currentModule` + `PythonIdent` construction pattern appears 4 times in this file (lines 428, 708, 1384, 1726). A small helper like:
```lean
def currentModuleIdent (name : String) : PySpecM PythonIdent := do
  return { pythonModule := toString (← read).currentModule, name }
```
would reduce the repetition.

PR #978, opened 2026-04-20
Fix Python->Laurel: Havoc callee or Heap after unmodeled InstanceCall

### LINKED ISSUE #977: Python->Laurel: Unmodeled InstanceCall does not havoc callee
Run pyAnalyzeLaurel on the following code:
```Python
xs = [1, 2]
xs.append(3)
assert xs[1] == 2, "appended element"
```

The result is 
```
test.py(3, 0): ✅ pass - appended element
```

Note that, `list.append` is unmodeled and translated into `.Hole`, so we should havoc `xs` after calling `list.append`.
Therefore, tt the current stage, the correct result should be 

```
test.py(3, 0): ❓ unknown - appended element
```

thanhnguyen-aws:
Fixes: #977

## Problem

When the Python→Laurel translator encounters an unmodeled instance call (e.g., `xs.some_method()`), it previously only havocked `maybe_except`. It did not account for the fact that the unmodeled call could mutate the receiver or heap-allocated fields, leading to unsound verification results where the verifier assumed variables retained their pre-call values.

For example, after `xs.some_unmodeled_call()`, the verifier would still prove `xs == [1, 2]` — incorrectly, since the call could have mutated `xs`.

## Fix

Introduces `mkHavocStmtsForUnmodeledCall`, a centralized helper that computes the correct set of havoc statements for any unmodeled call. The helper is used in both `translateAssign` (for `y = x.method()`) and `translateStmt` (for bare `x.method()` calls), replacing the previous inline `holeExceptHavoc` logic.

For an unmodeled call, the helper emits:
1. **Always:** havoc `maybe_except` (the call could throw)
2. **If the receiver is a non-composite variable** (e.g., a list): havoc the receiver variable (the method could mutate it)
3. **If the receiver is composite, or any argument is composite:** havoc `$heap` (the method could mutate heap-allocated fields)


shigoel:
@thanhnguyen-aws Can you please add some tests here?

thanhnguyen-aws:
> @thanhnguyen-aws Can you please add some tests here?

I added an expected test.

shigoel (commented):
## Review Summary

The fix is logically sound — when an unmodeled instance method call (e.g., `xs.some_unmodeled_call(3)`) is a black box, the receiver should be havocked since the method could mutate it. The implementation is correct for the standalone expression case.

### Soundness gap in `translateAssign`

The PR only adds callee havocking in the `.Expr _ value` case (standalone expression statements). But `y = xs.some_unmodeled_call(3)` goes through `translateAssign` where the RHS is an unmodeled Call → Hole. The Hole case (lines 1132–1160) havocs `maybe_except` but does **not** havoc the callee. This is the same soundness gap the PR intends to fix — it's just not covered in the assignment case.

### Test coverage

Only one test for the happy path. Consider adding:
- `y = xs.method()` (assignment form — exercises the `translateAssign` path)
- `a.b.method()` (chained attribute — verifies correct base extraction)
- Non-variable base like `get_list().method()` (should NOT havoc)

No performance concerns — the membership check pattern matches the rest of the file.

MikaelMayer (changes requested):
I haven't read the PR but it seems that if a method name is not recognized, we should just havoc the heap and that's it. The receiver is just a special case.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1492,6 +1503,20 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+
+    let calleeHavoc :=
+      if let .Call _ (.Attribute _ callee _ _) _ _ := value then
+        let (base, _) := getListAttributes callee
+        if let .Name _ n _ := base then
+          if n.val ∈ ctx.variableTypes.unzip.fst then
+            [mkStmtExprMdWithLoc (StmtExpr.Assign [freeVar n.val] (mkStmtExprMd (.Hole false none))) md]
+          else
+            []
+        else []
```
shigoel: This `calleeHavoc` block is identical to the one in `translateAssign` (~line 1142). Consider extracting a shared helper, e.g.:

```lean
private def mkCalleeHavoc (ctx : TranslationContext)
    (callExpr : Python.expr SourceRange)
    (md : Imperative.MetaData Core.Expression) : List StmtExprMd :=
  if let .Call _ (.Attribute _ callee _ _) _ _ := callExpr then
    let (base, _) := getListAttributes callee
    if let .Name _ n _ := base then
      if n.val ∈ ctx.variableTypes.unzip.fst then
        [mkStmtExprMdWithLoc (StmtExpr.Assign [freeVar n.val]
          (mkStmtExprMd (.Hole false none))) md]
      else []
    else []
  else []
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1492,6 +1503,20 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
+      else []
+
+    let havocStmts := [mkStmtExprMd $ .Block (holeExceptHavoc ++ calleeHavoc) none]
+
```
shigoel, line 1521: The `Block` wrapper around `holeExceptHavoc ++ calleeHavoc` is unnecessary — a label-less `Block` is semantically equivalent to inlining the statements. The original code just appended `++ holeExceptHavoc` directly. Consider:

```lean
let havocStmts := holeExceptHavoc ++ calleeHavoc
```

(Same applies to the `translateAssign` site at line 1152.)

StrataTest/Languages/Python/tests/test_havoc_callee_after_hole_call.py
```diff
@@ -0,0 +1,10 @@
+xs = [1, 2]
+xs.some_unmodeled_call_1(3)
+assert xs == [1, 2], "expected unknown because xs should be havocked"
+
+
+xs = [1,2]
+ys = xs.some_unmodeled_call_2()
+assert xs == [1, 2], "expected unknown because xs should be havocked"
```
shigoel, line 12: Good that the assignment case is covered now. Still missing two edge cases:
1. **Chained attribute**: `a.b.method()` — verifies `getListAttributes` extracts the correct base variable (`a`).
2. **Non-variable base**: `get_list().method()` — should NOT produce a havoc (the `Name` guard rejects it). This ensures we don't spuriously havoc.

shigoel (commented):
@thanhnguyen-aws Could you please address @MikaelMayer's comment re. the design here?

StrataTest/Languages/Python/tests/test_havoc_callee_after_hole_call.py
thanhnguyen-aws: I added more test cases to the file.

Strata/Languages/Python/PythonToLaurel.lean
thanhnguyen-aws: Without the `Block` wrapper, if the statements are in a `Try-Catch` block, the translation of  `Try` statement will emit unnecessary exception handler for those havoc statements (see line 1552 to 1558). There should be a better way to handle it by enhancing the translation of `Try` statement but it is beyond the scope of this PR.

thanhnguyen-aws:
> I haven't read the PR but it seems that if a method name is not recognized, we should just havoc the heap and that's it. The receiver is just a special case.

I added Heap havocking for cases when inputs of the unmodeled function call contains Class type. Note that `List` and `Dict` are currently wrapped by `Any` type and separated from the Heap, so we need to havoc them separately.

MikaelMayer (dismissed):
Super nice tests. You might consider updating the PR description and title.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1101,6 +1101,34 @@ def freeVar (name: String) := mkStmtExprMd (.Identifier name)
...
+    (md : Imperative.MetaData Core.Expression) : List StmtExprMd :=
+  if let .Call _ funccall args kwords := callExpr then
+    let holeExceptHavoc := [mkStmtExprMdWithLoc (StmtExpr.Assign [maybeExceptVar] (mkStmtExprMd (.Hole false none))) md]
+    let calleeHavoc :=
+      if let (.Attribute _ callee _ _) := funccall then
+        let (base, _) := getListAttributes callee
+        if let .Name _ n _ := base then
+          match ctx.variableTypes.find? (fun (v, _) => v == n.val) with
+          | some (varName, ty) =>
+            if isCompositeType ctx ty then []
```
shigoel: **Potential soundness gap**: When the callee base (`a` in `a.some_method()`) is a composite type, this returns `[]` for `calleeHavoc` — correct, since we don't want to havoc the reference itself. But the `involveHeap` check below only inspects explicit `args`/`kwords`, not the implicit `self`. So a call like `a.unmodeled_method()` with no composite-typed arguments would skip the `$heap` havoc entirely, even though the method could have mutated `a` on the heap.

Consider also setting `involveHeap := true` when `isCompositeType ctx ty` holds here, so that `$heap` is havocked for any instance call on a composite receiver.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1101,6 +1101,34 @@ def freeVar (name: String) := mkStmtExprMd (.Identifier name)
 def maybeExceptVar := freeVar "maybe_except"
 def nullcall_var := freeVar "nullcall_ret"
 
+private def mkHavocStmtsForUnmodeledCall (ctx : TranslationContext)
```
shigoel, line 1108: **Invariant suggestion**: This is a non-trivial function that determines havoc behavior for soundness. Consider stating its contract, e.g.:

```
-- Invariant: if `callExpr` is not `.Call`, returns `[]`.
-- Otherwise the returned block always havocs `maybe_except`;
-- additionally havocs callee (if non-composite instance call)
-- and `$heap` (if any argument — or the implicit receiver — is composite).
```

This would help future readers verify that all havoc paths are covered.

shigoel (commented):
Review from Claude Code — 1 soundness concern, 1 invariant suggestion, 1 minor style nit.

shigoel (approved):
LGTM — soundness gap for composite receivers is fixed, invariant is documented, tests cover the key cases. Two minor style nits (missing space in `let inputExprs:=`, `Prod.fst` vs destructuring) but nothing blocking.

shigoel:
@thanhnguyen-aws Please update the PR description.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1101,6 +1101,38 @@ def freeVar (name: String) := mkStmtExprMd (.Identifier name)
...
+        else ([], false)
+      else ([], false)
+    let inputExprs:= args.val.toList ++ kwords.val.toList.map (λ kw => match kw with
+            | keyword.mk_keyword _ _ expr => expr)
+    let involveHeap := calleeIsComposite || (inputExprs.any fun inputExpr =>
+        match inferExprType ctx inputExpr with
+        | .ok ty => isCompositeType ctx ty
+        | _ => false)
+    let heapHavoc := if !involveHeap then [] else
+        [mkStmtExprMdWithLoc (StmtExpr.Assign [freeVar "$heap"] (mkStmtExprMd (.Hole false none))) md]
```
tautschnig, line 1132: Minor: `$heap` is referenced here during Python-to-Laurel translation, but the `$heap` variable is introduced by the heap parameterization pass which runs later. This works because heap parameterization recognizes the assignment and transforms it appropriately, but it creates a temporal coupling between this code and the heap parameterization pass. A comment noting this dependency would help future readers.

tautschnig (approved):

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Tracked in #1016.

keyboardDrummer:
I think the right design here needs a Laurel change. When calling an unmodeled thing, we should call a Laurel procedure like this:
```
procedure Unmodelled() returns (maybe_except: ...)
  modifies *
  opaque;
```

The `modifies *` is currently missing in Laurel, which is why this PR had to refer to `$heap`. The follwoing in progress PR https://github.com/keyboardDrummer/Strata/pull/12 adds `wildcard *` to Laurel, but we can also create a separate PR for adding `modifies *` to Laurel, which might be the fastest way forward.

keyboardDrummer:
Asked a bot to create a separate PR for adding `modifies *`: https://github.com/strata-org/Strata/issues/1030

PR #984, opened 2026-04-20
Add hypothesmith-based fuzz testing for Python front-end

tautschnig:
Adds fuzz testing infrastructure for Strata's Python front-end, inspired by [CBMC's use of CSmith](https://github.com/diffblue/cbmc/blob/develop/.github/workflows/csmith.yaml) and the [hypothesmith](https://github.com/Zac-HD/hypothesmith) project for random Python program generation.

## Test modes

**Syntax mode** — Uses hypothesmith to generate syntactically valid Python from the grammar, then runs each program through the parse → Ion → round-trip pipeline. Tests that the parser doesn't crash or lose information on arbitrary valid Python.

**Semantic mode** — Uses a custom generator to produce typed Python programs with assertions whose expected values are computed by CPython at generation time (the CSmith checksum analogy). Each program is validated with CPython first, then verified with `pyAnalyzeLaurel`. A verification failure on a CPython-passing program indicates a semantic modelling bug in the Python → Laurel → Core pipeline.

## `--unrestricted` option

By default, only constructs known to work in Strata are generated. With `--unrestricted`, ~60 additional generators are included covering classes, generators, comprehensions, try/except, match, async, decorators, walrus operator, *args/**kwargs, and more. Programs hitting unsupported constructs are reported as SKIP, not failures.

## CI integration

A new workflow (`.github/workflows/python-fuzz.yml`) runs on every PR:

```
Pick random seed
        │
        ▼
Run 10 fuzz tests (both modes, --unrestricted)
        │
    ┌───┴───┐
  PASS    FAIL
    │       │
    ▼       ▼
  Done   Checkout merge base, rebuild, re-run same seed
            │
        ┌───┴───────┐
    Base FAILS    Base PASSES
        │              │
        ▼              ▼
  Create issue    CI FAILS
  (pre-existing   (regression
   bug, labeled    from this PR)
   'fuzz-bug')
```

- Pre-existing bugs create a GitHub issue but don't block the PR
- Regressions fail CI with reproduction instructions
- The seed is always reported for deterministic reproduction

## Local usage

```bash
# Set up (one-time)
cd Tools/Python
python3 -m venv .venv
.venv/bin/pip install -e . hypothesmith

# Run 10 tests with both modes
./scripts/hypothesmith.sh 10

# Reproduce a specific failure
./scripts/hypothesmith.sh 10 1745165432 both --unrestricted

# Only semantic mode, restricted to known-working constructs
./scripts/hypothesmith.sh 20 42 semantic
```

## Files

| File | Purpose |
|------|---------|
| `Tools/Python/scripts/hypothesmith.sh` | Main driver script (analogous to CBMC's `csmith.sh`) |
| `Tools/Python/scripts/gen_random_python.py` | Program generator with syntax and semantic modes |
| `Tools/Python/scripts/gen_unrestricted.py` | ~60 extended generators for `--unrestricted` mode |
| `.github/workflows/python-fuzz.yml` | CI workflow for PR-triggered fuzz testing |

## Testing

Tested locally with multiple seeds and program counts. All generated programs pass CPython validation (500+ tested). The restricted semantic mode produces 100% pass rate through `pyAnalyzeLaurel`. The unrestricted mode correctly identifies unsupported constructs as SKIP.


robin-aws:
High level feedback (without having looked at the details) - can you ensure these tests are run somehow as part of `lake test`, rather than having to add this as yet another custom GH workflow step that developers will not think of running?

I see at least two options:

1. Include a `*.lean` file that runs the shell script in an `#eval`
2. Avoid the shell script and implement the test harness in Lean directly.

Tools/Python/scripts/hypothesmith.sh
```diff
@@ -0,0 +1,357 @@
...
+      cat "$pyfile"
+      sem_failures=$((sem_failures + 1))
+      continue
+    fi
+
+    # Step 2: Parse Python → Ion.
+    if ! timeout "$PARSE_TIMEOUT" "$PYTHON" -m strata.gen py_to_strata \
+        --dialect "$DIALECT" "$pyfile" "$ion" 2>/dev/null; then
+      echo "  FAIL (parse): $base"
+      echo "--- Source code ---"
```
shigoel: **Parse failures in semantic mode should be SKIP, not FAIL**

This classifies parse failures as `FAIL (parse)`, but syntax mode (line 183) classifies the same situation as `SKIP (parse)`. With `--unrestricted`, the generator produces programs with classes, async, generators, etc. that the Strata parser may not support yet. These parse failures are false positives — they'll cause CI to report regressions that aren't real.

The documented result classification (line 49) also lists `SKIP (parse)` for "Strata's parser couldn't handle the syntax."

Suggestion: change to `SKIP (parse)` here to match syntax mode, or at minimum gate the FAIL on whether `--unrestricted` is set.

shigoel (commented):
One inline comment: semantic mode parse failure classification.

Tools/Python/scripts/hypothesmith.sh
tautschnig: Fixed in [0860954](https://github.com/strata-org/Strata/pull/984/commits/08609541a0ff8a51880d99db5ee8d9278846bf08).

tautschnig:
> High level feedback (without having looked at the details) - can you ensure these tests are run somehow as part of `lake test`, rather than having to add this as yet another custom GH workflow step that developers will not think of running?
> 
> I see at least two options:
> 
>     1. Include a `*.lean` file that runs the shell script in an `#eval`
> 
>     2. Avoid the shell script and implement the test harness in Lean directly.

Addressed in [7f9aaa5](https://github.com/strata-org/Strata/pull/984/commits/7f9aaa50869389b1ff9f0e383ad01ad88f8153f3) closer to option 1 - doing everything in Lean would mean re-implementing hypothesmith in Lean (for all that I can tell).

shigoel (dismissed):

Tools/Python/scripts/hypothesmith.sh
```diff
@@ -0,0 +1,355 @@
...
+      # constructs — expected when using --unrestricted.
+      echo "  SKIP (unsupported): $base"
+      continue
+    fi
+
+    # Step 4: Check verification results. If pyAnalyzeLaurel reports any
+    # failed assertions, but CPython ran the program successfully (Step 1),
+    # then Strata's semantic model disagrees with CPython. This is the
+    # most interesting kind of bug to find — analogous to CBMC finding a
+    # counterexample to the CSmith checksum assertion.
```
robin-aws, line 314: Just to double check, "failed assertion" here means "proven to not hold" as opposed to "failed to prove it holds"? It's not clear to me that the grep pattern is specific enough. 

It might be even better to also check these programs against the `pyInterpret` command from https://github.com/strata-org/Strata/pull/869 once it's available, because then we'll have concrete pass or fails to compare with CPython with no unknown middle ground.

Tools/Python/scripts/gen_random_python.py
robin-aws, line 1: Just a top level thought, making sure we agree on something implicit: an alternative would be to use the Python dialect to generate Python programs. But that would bias the generation towards parts of Python we handle/understand well I imagine. Therefore it's better to have this fuzzing implemented separately in Python.

.github/workflows/python-fuzz.yml
robin-aws, line 1: This is a lot of complexity in a GH actions workflow, and that has a lot of downsides since you can't run the logic locally (there are tools that let you execute actions locally in a container but I've never had much success with them).

I like the automatic cutting of issues. I feel that the sophistication around determining whether it's new or a regression is unnecessary though. If anything it seems unfair for someone to cut a PR and then for new random tests to appear and fail the CI, when it passed for them before cutting the PR. :) I'd vote for always cutting an issue and not blocking the PR. And given that, perhaps it would be much simpler to make this a scheduled workflow that runs once a day instead?

.github/workflows/python-fuzz.yml
```diff
@@ -0,0 +1,311 @@
...
+                `**Seed:** \`${seed}\``,
+                `**PR:** #${context.issue.number}`,
+                `**Merge base also fails:** yes`,
+                ``,
+                `### Reproduce`,
+                `\`\`\`bash`,
+                `cd Tools/Python`,
+                `./scripts/hypothesmith.sh 10 ${seed} both --unrestricted`,
+                `\`\`\``,
+                ``,
```
olivier-aws, line 293: Could the script also include the generated Python program? It may help to then have an agent run through all the created issues and group them by root cause, without having to re-generate the code each time.

I would not block merging on this

Tools/Python/scripts/gen_random_python.py
```diff
@@ -0,0 +1,383 @@
...
+    return (
+        f"    a: str = {a!r}\n"
+        f"    b: str = {b!r}\n"
+        f"    r: str = a + b\n"
+        f'    assert r == {result!r}, "str concat"\n'
+    )
+
+def _gen_if_else():
+    """If/else with a known branch outcome."""
+    cond_val = _rand_bool()
```
olivier-aws: Could we generate more complex if-then-else, where the condition is generated from `_gen_bool_logic`, extended with `none`?

olivier-aws (commented):
Two comments mostly because I'm unsure how the Python code generation works from the list of generators.

Tools/Python/scripts/gen_random_python.py
tautschnig: Fixed in [30ef165](https://github.com/strata-org/Strata/pull/984/commits/30ef165d4a61d9a915788e0eef61d96cb57fcef1).

.github/workflows/python-fuzz.yml
tautschnig: Fixed in [6c1b779](https://github.com/strata-org/Strata/pull/984/commits/6c1b779118713bb63a0cf042b4b63ff0cf27e238).

robin-aws (changes requested):

.github/workflows/python-fuzz.yml
tautschnig: Actually it's pretty easy to run locally, most of the complexity is just to avoid unduly blocking a PR. Running locally can by done via `./scripts/hypothesmith.sh`. Most of the remaining logic is to determine the cause of a failure: if the same test would have passed before the changes in the PR, then the PR broke something, and it is upon the PR author to fix this. If the test equally failed before the changes in the PR, an issue is filed as surely the PR should not be blocked.

Running once a day means we cannot tie regressions to root causes and we don't have a clear owner for who is to deal with any failing outcomes. Even worse, nobody might notice that the action failed in the first place.

Tools/Python/scripts/gen_random_python.py
tautschnig: Yes, we should avoid implicit bias.

olivier-aws:
LGTM. This PR will provide nice automation for bugs in our Python support.

olivier-aws (dismissed):

Tools/Python/scripts/hypothesmith.sh
tautschnig: Clarified comment (and added a `pyInterpret` TODO) in [31b3687](https://github.com/strata-org/Strata/pull/984/commits/31b3687a2966af7dbf4c245c3a993b3161eb616f).

.github/workflows/python-fuzz.yml
```diff
@@ -0,0 +1,325 @@
...
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v6
+        with:
+          # Full history is needed to find the merge base for failure analysis.
+          fetch-depth: 0
+
+      # Generate a seed from the current nanosecond timestamp. This is
+      # reported in the step name and all log output so that any failure
+      # can be reproduced deterministically.
```
robin-aws, line 113: Just an observation, not sure there's anything we can do about it: in many cases just pushing an empty commit to re-run the CI with a different seed could easily result in a green run. :)

.github/workflows/python-fuzz.yml
```diff
@@ -0,0 +1,325 @@
...
+#    that CPython runs correctly, that's a semantic modelling bug.
+#
+# The `--unrestricted` flag is used, which includes generators for Python
+# constructs beyond what Strata supports today (classes, generators,
+# comprehensions, try/except, match, async, decorators, etc.). Programs
+# that hit unsupported constructs are reported as SKIP, not failures.
+#
+# ## Failure analysis
+#
+# When fuzz tests fail on the PR, the workflow performs a bisection-style
```
robin-aws, line 44: "bisection-style" makes it sound much more expensive than it is, since it's just re-running on a single other commit. :)

.github/workflows/python-fuzz.yml
```diff
@@ -0,0 +1,325 @@
...
+#
+# - Tools/Python/scripts/hypothesmith.sh    — Main driver script
+# - Tools/Python/scripts/gen_random_python.py — Program generator
+# - Tools/Python/scripts/gen_unrestricted.py  — Extended generators
+# =============================================================================
+
+name: Python Fuzz
+
+on:
+  pull_request:
```
robin-aws, line 93: Good that it's not running on `merge_group`, since at that point we don't know which PR caused the regression if it failed. But perhaps add a comment that this is intentional and not an omission?

.github/workflows/python-fuzz.yml
```diff
@@ -0,0 +1,325 @@
...
+          fetch-depth: 0
+
+      # Generate a seed from the current nanosecond timestamp. This is
+      # reported in the step name and all log output so that any failure
+      # can be reproduced deterministically.
+      - name: Pick random seed
+        id: seed
+        run: echo "value=$(date +%s%N | cut -c1-10)" >> $GITHUB_OUTPUT
+
+      - name: Install cvc5
```
robin-aws, line 118: (non-blocking, somewhat orthogonal) Duplicating a lot of content between different workflows at this point, we should factor out some reusable workflows.

.github/workflows/python-fuzz.yml
```diff
@@ -0,0 +1,325 @@
...
+          ARCH=$(uname -m)
+          if [ "$ARCH" = "x86_64" ]; then ARCH_NAME="x86_64"
+          elif [ "$ARCH" = "aarch64" ] || [ "$ARCH" = "arm64" ]; then ARCH_NAME="arm64"
+          else echo "Unsupported architecture: $ARCH"; exit 1; fi
+          wget -q https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-Linux-${ARCH_NAME}-static.zip
+          unzip -q cvc5-Linux-${ARCH_NAME}-static.zip
+          chmod +x cvc5-Linux-${ARCH_NAME}-static/bin/cvc5
+          echo "$GITHUB_WORKSPACE/cvc5-Linux-${ARCH_NAME}-static/bin/" >> $GITHUB_PATH
+
+      # Try to restore a cached Lean build to avoid rebuilding from scratch.
```
robin-aws, line 130: Not sound in general, but okay since this is only a best-effort random testing attempt. Might be worth a comment though.

.github/workflows/python-fuzz.yml
robin-aws: Fair, but it's still a lot of other shell script content (and javascript) that could be moved into separate files. My philosophy is to minimize how much code is only in workflow files and therefore not locally testable.

> Running once a day means we cannot tie regressions to root causes and we don't have a clear owner for who is to deal with any failing outcomes. Even worse, nobody might notice that the action failed in the first place.

It will cut an issue if it fails though, and we have to ensure we don't ignore those entirely anyway. Agreed we would lose some visibility about which PR introduced a regression.

.github/workflows/python-fuzz.yml
```diff
@@ -0,0 +1,325 @@
...
+            });
+
+      # --- Regression: fail CI ---
+      # The merge base passes but the PR fails — this PR broke something.
+      - name: Fail if PR introduced a regression
+        if: steps.analyse.outputs.result == 'regression'
+        run: |
+          echo "::error::Fuzz test regression detected (seed=${{ steps.seed.outputs.value }}). The merge base passes but this PR fails."
+          echo ""
+          echo "Reproduce locally:"
```
robin-aws, line 305: Should we also provide instructions to explicitly add the failing tests to the actual test suite somewhere, once the PR addresses them?

Tools/Python/scripts/hypothesmith.sh
```diff
@@ -0,0 +1,362 @@
...
+    [ -f "$pyfile" ] || continue
+    count=$((count + 1))
+    local base
+    base=$(basename "$pyfile" .py)
+    local ion="$gen_dir/${base}.python.st.ion"
+    local ion2="$gen_dir/${base}.python.st.ion2"
+
+    # Step 1: Parse Python → Ion via the external Python tool.
+    # Parse failures are acceptable — hypothesmith may generate constructs
+    # that the Strata parser doesn't support yet.
```
robin-aws, line 179: Never failing on parse errors is a bit unfortunate, since we could potentially find cases we think are supported but don't work. Any way to be more precise?

Tools/Python/scripts/hypothesmith.sh
```diff
@@ -0,0 +1,362 @@
...
+    [ -f "$pyfile" ] || continue
+    count=$((count + 1))
+    local base
+    base=$(basename "$pyfile" .py)
+    local ion="$gen_dir/${base}.python.st.ion"
+
+    # Step 1: Sanity check — run with CPython to confirm assertions pass.
+    # This is the "reference execution" (analogous to CSmith's GCC run
+    # that computes the reference checksum). If CPython fails, the bug
+    # is in our generator, not in Strata.
```
robin-aws, line 241: Any way to cut an issue when this happens? We're definitely never going to notice otherwise.

Tools/Python/scripts/hypothesmith.sh
```diff
@@ -0,0 +1,362 @@
...
+    # then Strata's semantic model disagrees with CPython. This is the
+    # most interesting kind of bug to find — analogous to CBMC finding a
+    # counterexample to the CSmith checksum assertion.
+    #
+    # The grep matches "N failed" in the DETAIL line, which means the solver
+    # *proved* the assertion does not hold (found a counterexample). This is
+    # distinct from "inconclusive" (solver couldn't determine either way),
+    # which is not treated as a failure here.
+    # TODO: Once pyInterpret (#869) is available, use it for concrete
+    # pass/fail comparison with CPython (no unknown middle ground).
```
robin-aws, line 321: It is now! :) But this can be added later.

robin-aws (changes requested):
Did not review the actual generation logic super thoroughly, partially because it is only generating test input and there are no major consequences if it is wrong. It would be good to have a bit more meta-testing to have higher confidence it's at least mostly right.

.github/workflows/python-fuzz.yml
tautschnig: Yes, I acknowledge that, and it's down to developer discipline to (not) do that. This is fuzz testing, our real defense is proof. My goal is to establish infrastructure that helps us build a better system. Once we have it, we could have a farm of hosts run those tests all day long.

.github/workflows/python-fuzz.yml
tautschnig: Starting this in https://github.com/strata-org/Strata/pull/1137. Whichever of this vs #1137 comes second will be updated.

PR #985, opened 2026-04-20
Add Core-to-Core datatype partial evaluation transform with correctness proof

tautschnig:
Add a general Core-to-Core transform that simplifies datatype tester and selector applications on known constructor terms:
- tester(C(args)) → true/false depending on whether tester matches C
- selector_i(C(args)) → args[i] when selector matches C

Components:
- DatatypeInfo + collectDatatypeInfo: collect constructor, tester, and selector metadata from Core datatype declarations
- matchConstrApp: decompose an expression into constructor + arguments
- partialEvalDatatypesCore: recursive expression simplifier
- partialEvalDatatypesInProgram: apply across all procedure bodies, axioms, and specifications

Correctness proof (zero sorry):
- Define DtEquiv inductive relation capturing datatype axiom equivalence (reflexivity, symmetry, transitivity, congruence, tester/selector axioms)
- Prove partialEvalDatatypesCore_correct by structural induction on LExpr: all rewrite cases justified by DtEquiv axioms applied to simplified subexpressions, composed with congruence and transitivity

Wired into the GOTO pipeline after procedure inlining. Includes unit tests for collectDatatypeInfo, tester simplification (positive and negative), and selector projection.

### DIFF
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -7,6 +7,7 @@
 import Strata.Backends.CBMC.CollectSymbols
 import Strata.Backends.CBMC.GOTO.CoreToCProverGOTO
 import Strata.Transform.ProcedureInlining
+import Strata.Transform.DatatypePartialEval
 
 /-! ## Core-to-GOTO translation pipeline
 
@@ -510,6 +511,7 @@ public def inlineCoreToGotoFiles (program : Core.Program)
       let (_, prog') ← phase.transform prog; return prog') with
     | .ok r => pure r
     | .error msg => throw msg
+  let inlined := Strata.partialEvalDatatypesInProgram inlined
   let (tcPgm, Env) ← match typeCheckCore inlined factory with
     | .ok r => pure r
     | .error msg => throw msg
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -511,8 +511,8 @@ public def inlineCoreToGotoFiles (program : Core.Program)
       let (_, prog') ← phase.transform prog; return prog') with
     | .ok r => pure r
     | .error msg => throw msg
-  let inlined := Strata.partialEvalDatatypesInProgram inlined
-  let (tcPgm, Env) ← match typeCheckCore inlined factory with
+  let simplified := Strata.partialEvalDatatypesInProgram inlined
+  let (tcPgm, Env) ← match typeCheckCore simplified factory with
     | .ok r => pure r
     | .error msg => throw msg
   coreToGotoFiles tcPgm Env baseName sourceText entryPoints
--- a/Strata/DL/Lambda/TypeFactory.lean
+++ b/Strata/DL/Lambda/TypeFactory.lean
@@ -538,6 +538,37 @@ def unsafeDestructorFuncs {T} [BEq T.Identifier] [Inhabited T.IDMeta] [Inhabited
 
 ---------------------------------------------------------------------
 
+/-! ## Datatype Info
+
+Collected metadata about datatype constructors, testers, and selectors.
+Shared by the Lambda evaluator and Core-to-Core transforms.
+-/
+
+/-- Collected datatype metadata for partial evaluation and simplification. -/
+structure DatatypeInfo where
+  /-- Tester name → the constructor it tests for -/
+  testerToConstr : Std.HashMap String String := {}
+  /-- Selector name (e.g. "Any..as_int!") → (constructor name, field index) -/
+  selectorInfo : Std.HashMap String (String × Nat) := {}
+  /-- Set of all constructor names -/
+  constrNames : Std.HashSet String := {}
+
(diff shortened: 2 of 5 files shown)


joscoh:
I am a bit confused about this transformation's purpose. These simplifications are already part of the Lambda (and hence Core) partial evaluator.

tautschnig:
> I am a bit confused about this transformation's purpose. These simplifications are already part of the Lambda (and hence Core) partial evaluator.

My primary intended use for this transform is the GOTO pipeline: The Lambda evaluator handles these simplifications during symbolic evaluation via `inlineIfConstr` and `concreteEval`. The GOTO pipeline, however, translates Core to CProver GOTO without running the Lambda evaluator — after procedure inlining, tester/selector applications on known constructors remain unsimplified.

To address the duplication concern, I've factored out the shared piece in 6ca07fc: `DatatypeInfo` (the constructor/tester/selector name mappings) now lives in `Lambda.TypeFactory` with a `DatatypeInfo.ofDatatypes` builder that works on `List (LDatatype IDMeta)`. The standalone transform uses this shared structure, and `matchConstrApp` now delegates to the existing `getLFuncCall` from `Lambda.Factory` instead of reimplementing application decomposition. The Lambda evaluator could adopt `DatatypeInfo` in the future to replace its ad-hoc tester/selector dispatch.

joscoh:
Two general comments (I haven't looked at this in close detail yet). 
1. I would prefer to see something that pulls the datatype simplification out of the partial evaluator into a separate function so that it can be called both in the Lambda partial evaluator and in the Core -> GOTO transformation. Right now, this PR results in two separate implementations of the same thing, and this means we need to keep them in sync.
2. The proofs of correctness should not be given with respect to a new equivalence relation (and it is better to not include a proof than to include a `sorry` one). The datatype simplification is already proved correct wrt the Lambda denotational semantics in `LExprSemanticsConsistent.lean`. I would prefer to modify this proof and prove this transformation correct wrt an already existing semantics. I am happy to help with this once the datatype simplification has been fully abstracted.

aqjune-aws:
For the top-level statement of correctness, github.com/strata-org/Strata/blob/main/Strata/Transform/Specification.lean has a few styles and I think `Overapproximates` is probably the closest definition. https://github.com/strata-org/Strata/blob/main/Strata/Transform/DetToKleeneCorrect.lean#L637-L640 is an example that uses the definition. However, since this is supposed to be equivalent, I can add the notion in a separate pull request (or even directly push to this PR, if you want).

aqjune-aws:
@MikaelMayer After #754, would there be an opportunity to utilize this pass in the core pipeline (`corePipelinePhases`), under the assumption that this pass has more simplification rules like folding `x /\ ~x` to `false`? I thought it could be a very good preprocessing phase before `symbolicEvalPhase`, even if they both have partial evaluation, because folding like `x /\ ~x => false` is helpful to simplify asserts.

shigoel:
@tautschnig Thanks for this PR! IIUC, this PR handles syntactic constructor applications (`isfrom_int(from_int(x))`), where a constructor is literally present in the expression. This composes well with constant propagation (#1005) -- that PR can give expressions of the form above, and this PR can take over from there.

I like the `DatatypeInfo` structure and agree that simplifying datatype-related tester/selector applications is crucial to do for any datatype-aware transform (and it looks like it's one piece that'd unblock the GOTO pipeline!). And I strongly agree with @joscoh's comment about removing duplication and reusing existing semantics/specs. I'd like to talk about a related problem in Core's symbolic evaluation path, where this would be directly reusable. In any case, I'd like to see `DatatypeInfo` land, regardless of what happens with the rest of this PR.

Here's the related symbolic evaluation problem I want to talk about here in this PR (since that'd probably use `DatatypeInfo`): In the non-GOTO path, I've seen some exponential blowup caused by patterns like Python's `PAdd(v1: Any, v2: Any): Any`. Such functions have large type-dispatch branches (`PAdd` has 12 ITEs). The symbolic evaluator inlines these function bodies into expressions, so nested calls like `PAdd(PAdd...PAdd(a, b))..)` produce an exponential ITE tree. In my experiments, 6 typed-`int` parameters takes ~130s to evaluate and 8 hangs indefinitely.

IIUC, the GOTO pipeline does not have this problem because functions like `PAdd` stay as GOTO functions and CBMC handles the dispatch. Right? I'd like Core to handle the dispatch at least as well as CBMC does so that it's reusable across all analyses. 

To handle `PAdd`-like cases in Core, we need a capability in Core's evaluator to consult assumptions and prune dead branches, ideally during their construction so that they're never built and consequently, the size of expressions is small and cost of traversing them is low. Note that this is a more general solution than #985 (and #1005) -- the datatype constructor isn't syntactically present, but we're obtaining assumptions from places like `requires` clauses, path conditions, etc. as well.

I'd like to see this more general capabilities in Core's evaluator -- lemma-driven rewriter (very similar to CBMC!), where datatype simplification rules can be derived from `DatatypeInfo`, plus the user can also provide their own lemmas that'd be useful as equational rewrite rules (pending the addition of a `lemma` construct in Core). Of course, there are performance considerations to take into account, but I'd rather we keep moving towards a general solution.

So here are some possibilities for the short term:
1. Land `DataTypeInfo` separately as shared infrastructure (per @joscoh's suggestion) and let both Core's evaluator and GOTO path use it.
2. Land the current simplification for the GOTO pipeline, and if/when we have the general rewriting capabilities in Core, we can re-assess whether that subsumes this work.

PR #1006, opened 2026-04-21
Add Core-to-Core factory concrete evaluation transform

tautschnig:
Add evalConcreteFactoryApps: fires concreteEval for factory functions with all-concrete arguments. Compiles e.g. re_fullmatch_str("abc") into Core regex operations. Parameterized by Lambda.Factory.

Implemented in Strata/Transform/FactoryEval.lean (not GOTO-specific).

Includes:
- Unit tests for literal identity, variable identity, and ite recursion
- Proofs that leaves (const, op, bvar, fvar) are unchanged (zero sorry)
- Correctness documentation: semantics preservation follows from FuncWF.concreteEval_argmatch in the factory well-formedness conditions

### DIFF
--- a/Strata/Transform/FactoryEval.lean
+++ b/Strata/Transform/FactoryEval.lean
@@ -9,23 +9,59 @@ public import Strata.Languages.Core.Program
 public import Strata.Languages.Core.Statement
 public import Strata.Languages.Core.Expressions
 public import Strata.DL.Lambda.Factory
+import Strata.DL.Lambda.LExprEval
 
 namespace Strata
+
+open Lambda.LExpr (isCanonicalValue isConstrApp)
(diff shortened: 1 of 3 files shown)


joscoh:
I would prefer to factor out the `Factory` function evaluation in the concrete evaluator and have it defined once rather than twice. This would also mean we can use the existing proofs of soundness. The correctness argument in this PR is not a proof; it is more like unit tests.

tautschnig:
> I would prefer to factor out the `Factory` function evaluation in the concrete evaluator and have it defined once rather than twice. This would also mean we can use the existing proofs of soundness. The correctness argument in this PR is not a proof; it is more like unit tests.

I have partially addressed this, but still need to figure out how to make proofs work.

PR #1013, opened 2026-04-22
Add --path-cap N for bounded symbolic evaluation

shigoel:
## Summary

- Add `--path-cap N` CLI option to control path explosion in the symbolic evaluator. When the number of continuing paths exceeds N between statements, sibling paths from the same split are merged via `Env.merge`, trading path count for VC complexity.
- Refactor `evalAuxGo` from single-path to batch evaluation: all active paths evaluate the current statement together, results are merged down to the cap, then proceed to the next statement.

## Design

`--path-cap N` is a knob that trades fewer solver calls for more complex individual VCs. Without the cap, each path carries simple path conditions. With the cap, merged paths carry `if-then-else` expressions from `Env.merge` that the solver must reason about symbolically. **`Env.merge` is the function that determines the quality of the merged encoding, and _there is a lot of room to improve it in the future (e.g., incorporate term rewriting to simplify merged conditions)_.**


Strata/Languages/Core/Env.lean
```diff
@@ -139,6 +139,7 @@ structure Env where
   pathConditions : Imperative.PathConditions Expression
   warnings : List (Imperative.EvalWarning Expression)
   deferred : Imperative.ProofObligations Expression
+  pathCap : Option Nat := .none
```
tautschnig, line 142: **Design concern: `pathCap` in `Env`**

`pathCap` is a static configuration option — it never changes during evaluation. Storing it in every `Env` instance means:
1. It's duplicated across all paths.
2. `enforcePathCap` reads it from the first element of the list (`ewn :: _`), which is fragile — it assumes the list is non-empty and all paths agree.
3. It pollutes the `Env` structure with evaluation-strategy concerns.

Consider threading `pathCap` as a parameter to `evalAuxGo`/`processIteBranches`/`enforcePathCap` instead, or wrapping it in a reader-like config that's separate from `Env`. This would also avoid the `pathCap := _` pattern-match ignore in `ToFormat`.

Not a blocker, but worth considering for a follow-up.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+    | [] => (Ewns, noStats, nextSplitId)
+    | s :: rest =>
+      let (continuing, exiting) :=
+        good.partition (fun (ewn : EnvWithNext) => ewn.exitLabel.isNone)
+      let (results, stmtStats, nextSplitId) := continuing.foldl
+        (fun (acc, statsAcc, nId) (ewn : EnvWithNext) =>
+          let (res, stmtS, nId) := evalOneStmt old_var_subst ewn s nId
+            (fun e ss nId => evalAuxGo steps' old_var_subst [e] ss nId)
+            (fun e c c' t f nId => processIteBranches steps' old_var_subst e c c' t f nId)
+          (acc ++ res, statsAcc.merge stmtS, nId))
```
tautschnig: **Performance: quadratic list append in hot loop**

`acc ++ res` inside `foldl` is O(|acc|) per iteration. When there are `k` continuing paths each producing `m` results, this is O(k²·m). For large path counts (before the cap kicks in), this could be noticeable.

Consider collecting results in reverse and doing a single `List.reverse` at the end, or using `List.flatMap`/`List.bind`:
```suggestion
          (res.reverse ++ acc, statsAcc.merge stmtS, nId))
```
and then reversing the final `results`. Or even simpler, replace the `foldl` with a `flatMap`-style approach that maps each continuing path to its results and concatenates once.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+    (ts unmatched_t remaining_f paired : List EnvWithNext) : CondPairsResult :=
+  match ts with
+  | [] => ⟨paired.reverse, unmatched_t.reverse, remaining_f⟩
+  | e_t :: rest_ts =>
+    match e_t.splitConds.back? with
+    | some (splitId, cond_t, _) =>
+      match extractMatchingSplitId splitId remaining_f [] with
+      | some (e_f, remaining_f') =>
+        let merged : EnvWithNext := {
+          env := Env.merge cond_t e_t.env e_f.env,
```
tautschnig: **Correctness question: merge condition from true-side only**

The merge uses `cond_t` extracted from `e_t.splitConds.back?`. Both sides store the same `cond'` from `processIteBranches`, so this is correct. But the code doesn't verify that `e_f`'s condition matches `e_t`'s — it only matches on `splitId`.

This is fine given the current invariant (both sides of a split always get the same condition expression), but it's worth a brief comment here noting that `cond_t` is the shared branch condition, not a condition specific to the true side. The variable name `cond_t` is slightly misleading — consider `cond` or `split_cond`.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+  termination_by ewns.length
+  decreasing_by assumption
+
+/-- Apply the path cap between statements. Continuing paths (no active
+    exit) are merged down to the cap via `mergeCondPairs`. Exiting paths
+    are not merged — they skip remaining statements so they don't
+    contribute to exponential blowup.
+    `pathCap` is read from the first path's `Env`, where it is set once
+    at initialization and never modified during evaluation. -/
+private def enforcePathCap (ewns : List EnvWithNext) (stats : Statistics) :
```
tautschnig, line 563: **Fragile: reading `pathCap` from first element**

```lean
  | ewn :: _ =>
    match ewn.env.pathCap with
```

This assumes all paths share the same `pathCap`. If `pathCap` were ever modified per-path (e.g., by a future feature), this would silently use the wrong value. This is another argument for threading `pathCap` as a separate parameter rather than reading it from `Env`.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+      attr := decl.attr,
+      concreteEval := decl.concreteEval,
+      axioms := decl.axioms.map (captureFreevars Ewn.env paramNames)
+    }
+    match Ewn.env.addFactoryFunc func with
+    | .ok env' => ([{ Ewn with env := env' }], noStats, nextSplitId)
+    | .error e =>
+      ([{ Ewn with env := { Ewn.env with error := some (.Misc f!"{e}") } }], noStats, nextSplitId)
+  | .typeDecl _ _ => ([Ewn], noStats, nextSplitId)
+  | .exit l _ => ([{ Ewn with exitLabel := .some l}], noStats, nextSplitId)
```
tautschnig, line 651: **Refactoring suggestion: `evalOneStmt` is large and takes higher-order function parameters**

`evalOneStmt` takes two function parameters (`evalSub` and `processBranches`) to break the mutual recursion. This works but makes the code harder to follow — the reader has to trace back to the call site to understand what `evalSub` and `processBranches` actually do.

Consider either:
1. Making `evalOneStmt` part of the `mutual` block directly (if termination can be shown), or
2. Adding a brief doc comment to `evalOneStmt` explaining that `evalSub` and `processBranches` are the recursive calls from `evalAuxGo`/`processIteBranches`.

Option 2 is minimal and would help readability.

StrataTest/Languages/Core/Tests/VCGPathTests.lean
```diff
@@ -39,33 +84,263 @@ Property: assert
...
+    (options := { Core.VerifyOptions.quiet with pathCap := some 1 })
+  IO.println (statsLine stats numObs)
+
+---------------------------------------------------------------------
+-- exponentialItePgm
+---------------------------------------------------------------------
+
+-- Each ITE has both branches continuing (exit from the inner block), so
+-- paths double at each ITE: 2^4 = 16 without cap.
+def exponentialItePgm :=
```
tautschnig, line 213: **Missing: negative soundness test with path-cap**

All path-cap tests use programs that verify successfully. This is the most important gap in test coverage. Please add at least one test where:
1. A program has a genuine bug (e.g., an `ensures` that doesn't hold on some path).
2. Verification correctly reports `❌ fail` both *without* the cap and *with* `--path-cap 1`.

This would demonstrate that merging paths doesn't weaken the VCs to the point of missing real bugs. For example:
```
procedure buggy(c1 : bool) returns (r : int)
spec { ensures [post]: (r > 0); }
{
  done: {
    if (c1) { r := 1; exit done; }
    r := 0;  -- bug: 0 is not > 0
  }
};
```
Verify this fails both with and without `--path-cap 1`.

Even better: a test where the bug is only visible after merging multiple ITE branches (e.g., the merged ITE expression in the VC must still be checked by the solver).

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+
+mutual
+/-- Batch symbolic evaluator: evaluates a statement list for all input
+    paths simultaneously. Between each statement, enforces the path cap
+    by merging continuing (.none exit) paths. -/
+def evalAuxGo (steps : Nat) (old_var_subst : SubstMap)
+    (Ewns : List EnvWithNext) (ss : Statements) (nextSplitId : Nat) :
+    List EnvWithNext × Statistics × Nat :=
+  let (errors, good) := Ewns.partition (fun (ewn : EnvWithNext) => ewn.env.error.isSome)
+  if good.isEmpty then (Ewns, noStats, nextSplitId)
```
tautschnig, line 661: **Observation: error partition on every recursive call**

```lean
  let (errors, good) := Ewns.partition (fun (ewn : EnvWithNext) => ewn.env.error.isSome)
  if good.isEmpty then (Ewns, noStats, nextSplitId)
```

This partitions the full path list on every recursive call to `evalAuxGo`. In the old single-path code, error checking was a simple `match Ewn.env.error`. Now it's O(n) per statement. This is fine for correctness but worth noting as a potential optimization target if batch sizes grow large — errors could be filtered once at the top level and not re-checked.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+        simp only [List.length_cons] at ih ⊢; omega
+    · have ih := ih (e_t :: unmatched_t) remaining_f paired
+      simp only [List.length_cons] at ih ⊢; omega
+
+private theorem partition_length (l : List α) (p : α → Bool) :
+    (List.filter p l).length + (List.filter (not ∘ p) l).length = l.length := by
+  induction l with
+  | nil => simp
+  | cons h t ih => simp [List.filter]; split <;> simp_all <;> omega
+
```
tautschnig: Nit: Lean 4's `List` already has `List.length_filter_add` or similar. Check if `List.filter_length_add_filter_length_compl` (or equivalent in your toolchain) exists — it might let you drop this custom theorem.

If not available, this is fine as-is.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -378,18 +378,14 @@ structure EnvWithNext where
...
-| rest, .none => (rest, .none)
-| _, .some exitLabel => ([], .some exitLabel) -- Skip all remaining statements
+  /-- Stack of unmerged split conditions. Each entry is
+      `(splitId, cond, side)` where `splitId` uniquely identifies the
+      split origin (both paths from the same split get the same
+      `splitId`), `cond` is the branch condition expression, and
+      `side` distinguishes the branches (`true` = then, `false` = else
+      for ITE splits). Last element is the most recent (innermost)
+      unmerged split. -/
+  splitConds : Array (Nat × Expression.Expr × Bool) := #[]
```
tautschnig, line 394: **Documentation: `splitConds` invariant**

The doc comment explains the structure well. Consider also documenting the key invariant that makes `mergeCondPairs` work:
- Both sides of a split always have the same `splitId` and `cond` expression.
- The `side` field (`true`/`false`) distinguishes them.
- `splitConds` entries are only pushed in `processIteBranches` when `pathCap` is active.

This invariant is critical for correctness and is currently only mentioned in the `mergeCondPairs` comment.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+  induction pool generalizing acc with
+  | nil => simp [extractMatchingSplitId] at h
+  | cons e_t ts' ih =>
+    simp only [extractMatchingSplitId] at h
+    split at h
+    · rename_i sid _ _
+      split at h
+      · simp at h; obtain ⟨_, rfl⟩ := h
+        simp [List.length_reverse, List.length_append]; omega
+      · have := ih (e_t :: acc) h; simp [List.length] at this ⊢; omega
```
joscoh, line 445: Minor, but we should not use Lean auto-generated names.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+          exitLabel := .none,
+          splitConds := e_t.splitConds.pop } :: paired)
+        simp only [List.length_cons] at ih ⊢
+        omega
+      · have ih := ih (e_t :: unmatched_t) remaining_f paired
+        simp only [List.length_cons] at ih ⊢; omega
+    · have ih := ih (e_t :: unmatched_t) remaining_f paired
+      simp only [List.length_cons] at ih ⊢; omega
+
+private theorem partition_length (l : List α) (p : α → Bool) :
```
joscoh: This theorem (if it does not already exist in the standard library) should be in `Strata/DL/Util/List.lean`.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+-/
+private def mergeCondPairs (ewns : List EnvWithNext)
+    (target : Nat) : List EnvWithNext :=
+  if ewns.length <= target then ewns
+  else
+  let p := (fun e : EnvWithNext =>
+    match e.splitConds.back? with
+    | some (_, _, b) => b
+    | none => true)
+  let trues := ewns.filter p
```
joscoh: `List.partition` is more efficient than (but provably equal to) filtering twice

Strata/Languages/Core/StatementEval.lean
```diff
@@ -410,118 +406,291 @@ private def collectDeadBranchDeferred
...
+  -- that invariant were ever violated.
+  if h_nonempty : r.paired.isEmpty then
+    ewns
+  else
+    have h_part : trues.length + falses.length = ewns.length :=
+      partition_length ewns p
+    have h_fcpl : 2 * r.paired.length + r.unmatched_t.length + r.unmatched_f.length =
+        trues.length + falses.length := by
+      have := findCondPairs_length trues [] falses []
+      simp at this; exact this
```
joscoh: We should use explicit names here. Also, does `apply (findCondPairs_length trues [] falses [])` work? (I guess it depends on what `simp` is doing)

StrataTest/Languages/Core/Tests/VCGPathTests.lean
shigoel: Fixed by 45238c9cc.

Strata/Languages/Core/StatementEval.lean
shigoel: Fixed by 2e5525458.

Strata/Languages/Core/StatementEval.lean
shigoel: Fixed by 166f1d03e.

Strata/Languages/Core/StatementEval.lean
shigoel: Fixed by ad5dcab94.

Strata/Languages/Core/StatementEval.lean
shigoel: Fixed by 8cbcf3232.

Strata/Languages/Core/StatementEval.lean
shigoel: Checked: Lean Std (v4.29.1) has `length_eq_countP_add_countP` and `countP_eq_length_filter` but no direct equivalent for the filter+complement form. The predicate mismatch (`not ∘ p` vs `fun a => decide ¬p a = true`) requires glue code that's longer than the 3-line inductive proof. Kept the custom theorem but moved it to `Strata/DL/Util/List.lean` per joscoh's comment (8a7ee9c2c).

Strata/Languages/Core/StatementEval.lean
shigoel: Fixed by 24f360a10.

Strata/Languages/Core/StatementEval.lean
shigoel: Fixed by 8a7ee9c2c.

Strata/Languages/Core/StatementEval.lean
shigoel: Fixed by 68f1539e8.

Strata/Languages/Core/StatementEval.lean
shigoel: Fixed by 53833b2da. `apply` didn't work directly because `simp` is needed to simplify away the empty-accumulator zeros.

Strata/Languages/Core/Env.lean
shigoel: Agreed — `Env` already mixes static config with per-path state (e.g., `program` is set once at init and never changes during evaluation, same as `pathCap`). Splitting out a read-only config struct is a good cleanup but touches many call sites, so better as a follow-up PR.

Strata/Languages/Core/StatementEval.lean
shigoel: Same as above — this goes away once we split static config out of `Env`.

Strata/Languages/Core/StatementEval.lean
shigoel: Noted as a potential optimization target.

Strata/Languages/Core/StatementEval.lean
shigoel: Not minor, btw. Thanks for flagging!

Strata/Languages/Core/Env.lean
tautschnig: Could you please create an issue tracking that?

tautschnig (dismissed):

Strata/Languages/Core/Env.lean
shigoel: Filed #1018.

shigoel: Thanks for pushing on that!

joscoh (approved):

tautschnig (approved):

PR #1020, opened 2026-04-22
Minimize imports via lake shake and add import lint/stats

joehendrix:
## Summary

Run `lake shake` to minimize the import graph across all 293 Strata modules,
reducing average transitive Strata imports by 32% (56 → 38) and total transitive
imports (including Lean/Std) by 40% (1379 → 831). Add CI linting to ensure
`Strata.lean` transitively imports every module, and add an import statistics
script for ongoing measurement.

## Changes

- **Import lint driver (`Scripts/CheckImports.lean`)**: Verifies that all `.lean`
  files under `Strata/` are transitively imported by `Strata.lean`. Registered as
  `lintDriver` in `lakefile.toml` so `lake lint` runs it automatically. Supports
  `-- noimport: Strata.Foo` comments in `Strata.lean` to allowlist intentionally
  excluded modules.

- **Import statistics script (`Scripts/ImportStats.lean`)**: Computes per-module
  transitive import counts (both total and Strata-only), reporting average,
  median, min, max, top/bottom lists, and histograms. Run via `lake exe ImportStats`.

- **`lake shake Strata --fix`**: Applied shake's suggestions to remove unnecessary
  imports across 225 files (net -185 lines). This is the bulk of the change.

- **Aggregator modules use `-- shake: keep`**: Pure aggregator modules
  (`CBMC.lean`, `GOTO.lean`, `B3.lean`, `Dyn.lean`, `Python.lean`,
  `Imperative.lean`, `Lambda.lean`, `DDM.lean`, `LExprType.lean`,
  `B3/Verifier.lean`) intentionally re-export their submodules and are annotated
  with `-- shake: keep` so `lake shake --fix` won't strip them.

- **DDM elaboration imports use `-- shake: keep`**: Files using `#dialect`,
  `#strata`, `#strata_gen`, and related macros have elaboration-time import
  dependencies that shake cannot detect. These are annotated with
  `-- shake: keep` to prevent breakage.

- **Renamed aggregator files**: `Dyn/Dyn.lean` → `Dyn.lean`,
  `Python/Python.lean` → `Python.lean`, `Imperative/Imperative.lean` →
  `Imperative.lean` — the `Foo/Foo.lean` pattern is unnecessary now that these
  are proper `module` files.

- **Orphaned modules added to `Strata.lean`**: When shake stripped re-exports
  from real code files (e.g. `ProgramWF.lean` no longer re-exports `StatementWF`),
  the orphaned modules were added directly to `Strata.lean` to maintain full
  coverage.

## Results

| Metric | Before | After | Change |
|---|---|---|---|
| Avg transitive imports (Strata-only) | 56.2 | 38.2 | -32% |
| Median transitive imports (Strata-only) | 38 | 22 | -42% |
| Avg transitive imports (all) | 1379 | 831 | -40% |
| Total import edges (Strata-only) | 16,455 | 11,194 | -32% |
| Clean build time | 3m 02s | 2m 58s | -2% |


MikaelMayer (commented):
🤖🔍 Overall this is a solid mechanical import-minimization PR with clear metrics. A few concerns:

1. **Internal Std imports**: Several files now import `Std.Data.DTreeMap.Internal.Operations`, `Std.Tactic.BVDecide.Normalize.Prop`, `Std.Tactic.BVDecide.Normalize.BitVec`, `Std.Tactic.BVDecide.Normalize.Bool`, and `Std.Do.Triple.SpecLemmas`. These look like internal implementation modules that may not be part of Std's stable API. If they get reorganized upstream, this will break. Consider whether these are truly the minimal necessary imports or if there's a more stable public entry point.

2. **Emptied aggregator modules**: `Imperative.lean`, `SMT.lean`, and `LExprType.lean` are now empty `module` files with no imports. If they're truly dead, they should be deleted rather than left as empty shells. If they're kept for the `-- shake: keep` aggregator pattern, they need the annotation.

3. **MetaVerifier.lean workaround**: The comment "For some reason shake wants to meta import the following while lake itself only requires imports" suggests shake may be producing incorrect suggestions here. The pattern of duplicating imports as both `meta import` and `import -- shake: keep` is unusual and warrants investigation.

4. **Deletion candidates list**: The 23 `-- noimport:` modules in `Strata.lean` is a significant number of orphaned modules. Is there a plan to either delete them or re-integrate them?

Strata/DL/Lambda/LTy.lean
```diff
@@ -5,9 +5,10 @@
 -/
 module
 
-public import Strata.DL.Util.Map
 import Strata.Util.Tactics
-public meta import Lean.Elab.Term
+public meta import Lean.Elab.Term.TermElabM
+import Std.Data.DTreeMap.Internal.Operations
```
MikaelMayer: 🤖🔍 `Std.Data.DTreeMap.Internal.Operations` is an internal module (note `Internal` in the path). Importing internal modules creates fragile dependencies that can break on Std version bumps. What definition from this module is actually needed here? If it's just for a `DecidableEq` or `BEq` instance, there's likely a public re-export.

Strata/MetaVerifier.lean
```diff
@@ -5,17 +5,32 @@
...
+import Strata.DL.SMT.Denote
+import Strata.Languages.Boole.Verify
+import Strata.Languages.C_Simp.DDMTransform.Translate
+import Strata.Languages.C_Simp.Verify
+import Strata.Languages.Core.Core
+import Strata.Languages.Core.DDMTransform.Translate
+import Strata.Languages.Core.ProgramEval
+
+-- For some reason shake wants to meta import the following
+-- while lake itself only requires imports.
```
MikaelMayer, line 24: 🤖🔍 This pattern of duplicating each import as both `meta import` and `import -- shake: keep` is a maintenance hazard. If shake is producing incorrect suggestions, it may be better to annotate the entire block with a single `-- shake: keep` comment explaining why, rather than doubling every import line. The comment acknowledges this is unexpected behavior — is there a shake bug to file?

Strata/DL/Imperative/Imperative.lean
```diff
@@ -5,21 +5,8 @@
 -/
 module
```
MikaelMayer: 🤖🔍 This file is now an empty module (just the copyright header and `module`). If it's meant to be a `-- shake: keep` aggregator, it needs that annotation and its re-exports. If it's truly dead code, it should be deleted rather than left as an empty file. Same applies to `Strata/DL/SMT/SMT.lean` and `Strata/DL/Lambda/LExprType.lean`.

Strata/Util/Tactics.lean
```diff
@@ -5,7 +5,12 @@
 -/
 module
 
-public meta import Lean.Elab.Tactic
+public meta import Init.Grind.Cases
```
MikaelMayer, line 8: 🤖🔍 `Std.Do.Triple.SpecLemmas` appears to be an internal module from Std's do-notation verification infrastructure. Is this actually needed for tactics, or is shake pulling it in as a transitive dependency of something that was previously imported via a broader module? Same question for `Init.Grind.Cases` and `Init.Grind.Ext` — these are grind internals.

Strata/Languages/Boole/Grammar.lean
```diff
@@ -5,11 +5,13 @@
...
-public import Strata.Languages.Core.DDMTransform.Grammar
-meta import Strata.DDM.Integration.Lean
+public import Strata.Languages.Core.DDMTransform.Grammar -- shake: keep
+import Strata.DDM.Integration.Lean.HashCommands -- shake: keep
+import Strata.DDM.BuiltinDialects.Init
 
 ---------------------------------------------------------------------
 
+public section
 namespace Strata
```
MikaelMayer, line 16: 🤖🔍 Nit: The added `public section` / `end` wrapping around the existing `namespace Strata` block changes the visibility of all declarations in this file. Was this intentional? If the namespace content was already public, this is redundant; if it wasn't, this is a semantic change beyond import minimization.

Strata.lean
```diff
@@ -71,4 +71,28 @@ import Strata.SimpleAPI
 /- Pipeline -/
 import Strata.Pipeline.PyAnalyzeLaurel
 
--- noimport: Strata.Util.Random -- deletion candidate: nothing imports this module
+ -- deletion candidates: nothing imports these modules:
+
```
MikaelMayer, line 77: 🤖🔍 23 orphaned modules is a lot. Some of these (like the `Denote.*` family and `ProgramWF`, `StatementWF`) look like they contain proofs that are still valuable even if nothing currently imports them. Is there a tracking issue for deciding which to delete vs. re-integrate? Leaving them indefinitely as `-- noimport:` comments risks them silently bit-rotting.

Strata/Languages/Boole/Grammar.lean
joehendrix: Yes, it is intentional.  The dialect should be public.

Strata/MetaVerifier.lean
joehendrix: Yes, there's almost certainly a bug to file once we get this minimized.

MikaelMayer (commented):

- Internal Std imports (`Std.Data.DTreeMap.Internal.Operations`, `Std.Do.Triple.SpecLemmas`, `Std.Tactic.BVDecide.Normalize.Prop`) replaced with stable public entry points (`Std.Data.DTreeMap`, `Std.Do`, `Std.Tactic.BVDecide`)
- Empty aggregator shells deleted and replaced with proper `-- shake: keep-all` aggregator modules at the parent level
- Orphaned `-- noimport:` modules now properly imported in `Strata.lean`
- MetaVerifier pattern acknowledged as a shake bug to file (acceptable)
- `public section` in Boole/Grammar.lean confirmed intentional by author

MikaelMayer (dismissed):

MikaelMayer (commented):

New commit `046c7bf` adds `Strata.Languages.Laurel.Grammar` import and fixes `ReflectTests.lean` imports — both look correct.

Minor note: `Std.Tactic.BVDecide.Normalize.Prop` remains in MetaVerifier.lean (an internal module), but given the acknowledged shake bugs in that file, this is acceptable for now.

MikaelMayer (commented):
🤖🔍 All previous inline comments appear addressed:
- Internal Std imports in LTy.lean and Tactics.lean replaced with stable entry points ✅
- Empty aggregator modules (Imperative, SMT, LExprType) properly fixed or deleted ✅
- Orphaned modules now properly imported in Strata.lean ✅
- MetaVerifier shake bug acknowledged ✅
- public section in Boole/Grammar confirmed intentional ✅

One new concern: `DDMAxiomsExtraction.lean` had a `#guard_msgs in #eval examplePgm.commands` test removed (commit f381871). This test exists and passes on main. Was this removal intentional, or did the import change break the `.commands` output? If the output changed due to import minimization, the expected output should be updated rather than the test deleted.

Also minor: `Std.Tactic.BVDecide.Normalize.BitVec` remains in `Verifier.lean` (an internal module), while the same pattern was fixed to `Std.Tactic.BVDecide` in `LTy.lean`. Not blocking but worth noting for consistency.

shigoel (approved):
LGTM!

MikaelMayer (approved):

Strata/DL/Imperative/Imperative.lean
joehendrix: Fixed

PR #1021, opened 2026-04-22
Ensure Strata.lean imports all modules and add import stats

joehendrix:
## Summary

Add CI linting to ensure `Strata.lean` transitively imports every Strata module,
and add a statistics script for measuring the import graph.

## Changes

- **Import lint driver (`Scripts/CheckImports.lean`)**: Verifies that all `.lean`
  files under `Strata/` are transitively imported by `Strata.lean`. Registered as
  `lintDriver` in `lakefile.toml` so `lake lint` runs it automatically. Supports
  `-- noimport: Strata.Foo` comments in `Strata.lean` to allowlist intentionally
  excluded modules.

- **`Strata.lean` updated** to transitively import all 293 modules. Added
  aggregator modules (`CBMC.lean`, `GOTO.lean`, `B3.lean`, `DDM.lean`) that
  re-export submodules.

- **Import statistics script (`Scripts/ImportStats.lean`)**: Computes per-module
  transitive import counts (both total and Strata-only), reporting average,
  median, min, max, top/bottom lists, and histograms. Run via `lake exe ImportStats`.

### DIFF
--- a/Strata.lean
+++ b/Strata.lean
@@ -33,8 +33,37 @@ import Strata.Transform.DetToKleeneCorrect
 import Strata.Transform.ProcBodyVerifyCorrect
 import Strata.Transform.Specification
 
+/- Strata Languages — additional -/
+import Strata.Languages.B3
+import Strata.Languages.Boole.Boole
+import Strata.Languages.Boole.Verify
+import Strata.Languages.C_Simp.C_Simp
+import Strata.Languages.C_Simp.Verify
+import Strata.Languages.Core.EntryPoint
+import Strata.Languages.Core.VerifierProofs
+import Strata.Languages.Dyn.Dyn
+import Strata.Languages.Dyn.Verify
+import Strata.Languages.Python.Python
+
+/- DDM -/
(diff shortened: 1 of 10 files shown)


shigoel (approved):
Clean infrastructure PR. Import lint driver and aggregator modules look good, CI green. LGTM.

Strata.lean
```diff
@@ -33,8 +33,37 @@ import Strata.Transform.DetToKleeneCorrect
...
+/- Code Transforms — additional -/
+import Strata.Transform.StructuredToUnstructured
+
+/- Other -/
+import Strata.MetaVerifier
 
 /- Simple API -/
 import Strata.SimpleAPI
+
+-- noimport: Strata.Util.Random -- deletion candidate: nothing imports this module
```
MikaelMayer, line 69: Let's delete it?

MikaelMayer (approved):

Strata.lean
joehendrix: I can make it in a followup PR.

PR #1033, opened 2026-04-23
Improve verifier profiling accuracy; add PAdd perf test

shigoel:
## Summary

- **Fix eager `formatProgram`**: `getObligationResult` was computing `Core.formatProgram p` unconditionally for every obligation (only used in debug traces). Deferred behind a thunk.

- **Fix `toSMTTerms` timing**: The existing profiling bound `toSMTTerms` with a pure `let`, so its cost was deferred and misattributed to "Solver/file writing". Inlined the `match` so SMT encoding time is correctly reported.

- **Add per-obligation profiling**: When `--profile` is set, print per-obligation timing and `getVars`/`serialize/write/solve` breakdown inside `getObligationResult`.

- **Add `test_add6.py`**: Pending Python test with 6 `int` parameters whose nested `PAdd` calls trigger exponential expression blowup in VC processing.

## Profiling results on `test_add6.py`

To reproduce:
```
source StrataTest/Languages/Python/venv/bin/activate
python3 Tools/Python/strata/gen.py py_to_strata \
  --dialect "Tools/Python/dialects/Python.dialect.st.ion" \
  StrataTest/Languages/Python/tests/pending/test_add6.py \
  /tmp/test_add6.python.st.ion

lake exe strata pyAnalyzeLaurel --no-solve --vc-directory /tmp/add6_vcs \
  --profile /tmp/test_add6.python.st.ion
```

Before this PR, `--profile` reported SMT encoding as 0ms. After:

```
obligation[0] assert(77):      53,939ms
  getVars: 3,900ms, serialize/write/solve: 8,910ms
obligation[1] postcondition:   57,232ms
  getVars: 4,423ms, serialize/write/solve: 8,784ms
obligation[2..9]:              < 1ms each

Preprocess obligations: 18,545ms   (getOps tree walk for axiom pruning)
SMT encoding:           66,609ms   (toSMTTerms tree walk — was 0ms before)
Solver/file writing:    26,021ms   (serialize + write + solve)
```

Only 2 of 10 obligations are slow — those containing the deeply nested `PAdd` ITE expression. All four components (preprocess, SMT encoding, getVars, file writing) suffer from the same root cause: DAG-as-tree traversal of shared sub-expressions.

## Test plan

- [ ] `lake build strata:exe` succeeds
- [ ] Run `test_add6.py` through `pyAnalyzeLaurel --profile` and verify the new profiling output shows per-obligation timing and correct SMT encoding attribution

### DIFF
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -888,7 +888,7 @@ def getObligationResult (assumptionTerms : List Term) (obligationTerm : Term)
           (label := obligation.label))
   let tDone ← IO.monoNanosNow
   if profile then
-    let _ ← (IO.println s!"[profile]         getVars: {nsToMs (tDischarge - tGetVars)}ms, discharge: {nsToMs (tDone - tDischarge)}ms" |>.toBaseIO)
+    let _ ← (IO.println s!"[profile]         getVars: {nsToMs (tDischarge - tGetVars)}ms, serialize/write/solve: {nsToMs (tDone - tDischarge)}ms" |>.toBaseIO)
   match ans with
   | .error e =>
     dbg_trace f!"\n\nObligation {obligation.label}: SMT Solver Invocation Error!\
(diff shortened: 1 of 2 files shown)


shigoel:
Closing; stale PR.

PR #1035, opened 2026-04-23
Fix DictNoneTest Test 6: len() rejection is an error, not a diagnostic

tautschnig:
Test 6 (len() on a class without __len__) was always broken since its introduction in PR #761 (bb11e70e). The test expected processPythonFile to return diagnostics, but the len() rejection throws a TranslationError in PythonToLaurel.lean, which withPythonToLaurel converts to an IO.Error before the diagnostic stage is reached.

The test was never caught because:
1. It was authored in an environment without Python, so withPython skipped and #guard_msgs saw empty output (matching the empty docstring).
2. DictNoneTest was not in CI's explicit Python test list in ci.yml, so it was never run in CI either.

Fix: catch the IO.Error and verify it contains the expected message, and add DictNoneTest to CI's Python test list.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -304,7 +304,7 @@ jobs:
               sudo cp z3-4.13.4-x64-glibc-2.35/bin/z3 /usr/local/bin/
             fi
     - name: Run PySpec and dispatch tests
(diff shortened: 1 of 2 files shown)


joehendrix (dismissed):

keyboardDrummer (approved):

MikaelMayer (approved):
Clean, well-scoped fix. The PR correctly identifies that `processPythonFile` throws an `IO.Error` (via `TranslationError`) for `len()` on a class without `__len__`, rather than returning diagnostics. The `.toBaseIO` pattern to catch and inspect the error is consistent with how the rest of the codebase handles fallible IO. Adding DictNoneTest to CI closes the gap that allowed this to go undetected.

Verified on main: Test 6 was indeed broken (calls `processPythonFile` expecting diagnostics, but the call throws before producing any), and DictNoneTest was not in CI.

PR #1036, opened 2026-04-23
feat(core): Add Sequence.empty<T>() syntax for creating empty sequences

### LINKED ISSUE #1027: Make creating empty sequences and maps easier
Currently, the `Sequence` type in Strata Core lacks an `empty` function, so creating an empty sequence is awkward. You could do something like `var s : Sequence T; assume Sequence.length(s) == 0;`, which doesn't compute well. Using the same typing setup as the rest of the sequence operators doesn't work well for `empty`, so it's currently commented out. We'd probably need to use `op` instead of `fn` in the DDM definition and have a special case in the translator.

The `Map` type is even worse, because it doesn't even have a function to extract its size or domain. An `empty` function would need to be created in the same way as for `Sequence`, but the semantics of `Map.empty` are unclear with the current `Map` type, given that the current `Map` type is total. We'll need to either extend it with a notion of domain, or model it with a domain tracking value on the side.

atomb:
Enable parsing Sequence.empty from .st files by adding a DDM fn declaration with an explicit type parameter. Previously, the 0-ary polymorphic function could not be parsed because the DDM grammar had no value arguments to infer the type parameter from.

Users now write Sequence.empty<int>() instead of the workaround of declaring a variable and assuming its length is zero.

- Grammar.lean: Add fn seq_empty with explicit Type argument
- Translate.lean: Add translation case for seq_empty
- FormatCore.lean: Format Sequence.empty back to <T>() syntax
- Factory.lean: Update seqEmptyFunc comment (now parseable)
- Seq.lean: Add verification tests for the new syntax

Addresses but does not completely resolve #1027


aqjune-aws (dismissed):

MikaelMayer (dismissed):

tautschnig (changes requested):
See comment above.

Strata/Languages/Core/DDMTransform/FormatCore.lean
```diff
@@ -289,13 +289,21 @@ def lconstToExpr {M} [Inhabited M] (c : Lambda.LConst) :
...
+    (opTy : Option Lambda.LMonoTy := none)
     : ToCSTM M (CoreDDM.Expr M) :=
   open Core in
   match CoreOp.ofString name with
   | .re .All => pure (.re_all default)
   | .re .AllChar => pure (.re_allchar default)
   | .re .None => pure (.re_none default)
-  -- TODO: seq_empty is not yet parseable (see Grammar.lean); handle here when added.
+  | .seq .Empty => do
+    match opTy with
```
tautschnig, line 300: Minor refactor: `seqTy` in `Factory.lean` is annotated `@[expose, match_pattern]` specifically so callers can pattern-match without hard-coding the `"Sequence"` string and the arity. That keeps this site decoupled from the concrete representation of `seqTy`.

```suggestion
    | some (Core.seqTy elemTy) =>
```

(You'll want `open Core in` already in scope, which it is.)

Strata/Languages/Core/DDMTransform/FormatCore.lean
```diff
@@ -289,13 +289,21 @@ def lconstToExpr {M} [Inhabited M] (c : Lambda.LConst) :
...
+    | some (.tcons "Sequence" [elemTy]) =>
+      let ety ← lmonoTyToCoreType elemTy
+      pure (.seq_empty default ety)
+    | _ =>
+      let ety := CoreType.tvar default unknownTypeVar
```
tautschnig, line 306: The silent fallback to `unknownTypeVar` worries me a bit: by construction, by the time we're in `FormatCore` the expression has been type-checked, so hitting this branch means an internal invariant has been violated (`Sequence.empty` whose op-type annotation isn't a `Sequence _`, or is missing entirely). Every other comparable path in this file calls `ToCSTM.logError` on the unexpected-case branch; would you mind following the same convention here so we surface the miscompile instead of printing `$__unknown_type` to the user?

Something like:
```suggestion
    | _ => do
      ToCSTM.logError "handleZeroaryOps" "Sequence.empty missing or malformed op-type annotation" name
      let ety := CoreType.tvar default unknownTypeVar
      pure (.seq_empty default ety)
```

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -103,9 +103,8 @@ fn map_get (K : Type, V : Type, m : Map K V, k : K) : V => m "[" k "]";
 fn map_set (K : Type, V : Type, m : Map K V, k : K, v : V) : Map K V =>
   m "[" k ":=" v "]";
 
-// TODO: seq_empty is not yet supported in the grammar because the DDM parser
-// cannot currently handle 0-ary polymorphic functions (no arguments to infer
-// the type parameter from). The Factory definition exists for programmatic use.
+fn seq_empty (A : Type) : Sequence A =>
+  "Sequence.empty" "<" A ">" "(" ")";
```
tautschnig, line 109: Optional: worth a one-line comment here explaining *why* this fn is the odd one out (every other `seq_*` takes a value argument from which DDM can infer `A`, while `seq_empty` has no value args, so the type argument has to be explicit in the surface syntax). Otherwise a future reader is likely to try to delete `"<" A ">"` "for consistency".

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -848,6 +848,13 @@ partial def translateExpr (p : Program) (bindings : TransBindings) (arg : Arg) :
   | .fn _ q`Core.re_all, [] =>
     let fn ← translateFn .none q`Core.re_all
     return fn
+  -- Sequence.empty (0-ary polymorphic, takes only a type argument)
+  | .fn _ q`Core.seq_empty, [_atp] =>
+     let ety ← translateLMonoTy bindings _atp
+     let fn : LExpr Core.CoreLParams.mono :=
+       Core.coreOpExpr (.seq .Empty)
+         (.some (Core.seqTy ety))
```
tautschnig, line 856: Would you consider adding a small `#guard`-style test (in `StrataTest/Languages/Core/Tests/` — `CoreOpTests.lean` or a new `TranslateTests.lean`) that exercises the expected structural shape of this translation for a couple of element types? Something along the lines of:

```lean
/-- The translation of `Sequence.empty<T>()` produces a `.seq .Empty` operator
    annotated with `Sequence T`. -/
#guard
  match translateExpr (mkTestPgm "Sequence.empty<int>()") with
  | .ok (.op _ name (.some ty)) =>
    name.name == "Sequence.empty" && ty == Core.seqTy .int
  | _ => false
```

The `#eval verify` test in `Seq.lean` exercises this path end-to-end, but a direct structural test would catch regressions (missing `seqTy` wrapper, wrong kind tag, etc.) without relying on an SMT call.

For an actual theorem, the nice property to prove is the round-trip with `FormatCore`: for any `T` in the Core grammar, `format ∘ translate` on `Sequence.empty<T>()` is the identity (up to formatting whitespace). That would tie the three pieces of this PR together.

Strata/Languages/Core/Factory.lean
```diff
@@ -432,9 +432,7 @@ def seqLengthFunc : WFLFunc CoreLParams :=
     ])
 
 /- An empty `Sequence` constructor with type `∀a. Sequence a`.
-   NOTE: This is registered in the Factory for programmatic use, but is not yet
-   parseable from `.st` files because the DDM grammar cannot currently handle
-   0-ary polymorphic functions (no arguments to infer the type parameter from). -/
+   `Sequence.empty<A>()` returns an empty sequence of element type `A`. -/
```
tautschnig: Nit: doc now says the user writes `Sequence.empty<A>()`, but the Factory function itself is still `Sequence.empty : ∀a. Sequence a` (no value args). It's worth one extra sentence clarifying that the `<A>` is purely surface syntax produced by `Grammar.lean` / consumed by `Translate.lean`, so readers of `Factory.lean` alone don't go hunting for a `A` parameter on `seqEmptyFunc`.

StrataTest/Languages/Core/Examples/Seq.lean
```diff
@@ -305,3 +305,86 @@ Result: ✅ pass
...
+Result: ✅ pass
+
+Obligation: build_on_empty_length
+Property: assert
+Result: ✅ pass
+
+Obligation: build_on_empty_elem
+Property: assert
+Result: ✅ pass
+-/
```
tautschnig, line 386: Test coverage is narrow — only `Sequence.empty<int>()`. I verified manually that each of the following also works end-to-end with the PR as-is, so please extend the test to include them (otherwise we have no regression protection for these cases):

1. Non-`int` primitive element: `Sequence.empty<bool>()`
2. Nested sequences: `Sequence.empty<Sequence int>()`
3. Compound element type: `Sequence.empty<Map int bool>()`

Example follow-up test (feel free to consolidate into one procedure):

```suggestion
#eval verify seqEmptyPgm

----------------------------------------------------------------------

-- Exercise various element types for Sequence.empty<T>().
private def seqEmptyTypesPgm :=
#strata
program Core;

procedure SeqEmptyTypes()
{
  var sb : Sequence bool;
  var sss : Sequence (Sequence int);
  var smi : Sequence (Map int bool);

  sb := Sequence.empty<bool>();
  sss := Sequence.empty<Sequence int>();
  smi := Sequence.empty<Map int bool>();

  assert [bool_len]: Sequence.length(sb) == 0;
  assert [seq_seq_len]: Sequence.length(sss) == 0;
  assert [seq_map_len]: Sequence.length(smi) == 0;
};
#end

/-- info: true -/
#guard_msgs in
#eval TransM.run Inhabited.default (translateProgram seqEmptyTypesPgm) |>.snd |>.isEmpty
```

Also worth adding: a negative test showing that `Sequence.empty()` (no type argument) is rejected at parse time, so we can't silently regress to the old polymorphism-inference behavior if the grammar rule is ever weakened.

Related proof suggestion for `SeqModel.lean`: you already prove `length_empty : ([] : List α).length = 0`. It would be nice to extend the axiom-model block with a short lemma tying `Sequence.build(Sequence.empty, v)` to `[v]`, matching the `build_on_empty_*` obligations this PR's test introduces.

StrataTest/Languages/Core/Examples/Seq.lean
```diff
@@ -305,3 +305,86 @@ Result: ✅ pass
 #eval verify seqOpsPgm
 
 ---------------------------------------------------------------------
+
```
tautschnig, line 308: Note that the PR-description / commit-message wording is inconsistent with the actual scope: the commit message on `271fb5d68` says "Resolves #1027", but the PR body correctly says "Addresses but does not completely resolve #1027" (the Map part of the issue is not touched). Would you mind amending the commit message to match, so that squash-merging the PR doesn't auto-close #1027?

tautschnig (changes requested):
Nice, focused fix that lifts a real user-facing restriction on `Sequence.empty`. The happy path is well exercised via the `#eval verify` test (parse → translate → format → typecheck → VC → SMT).

Before approving I'd like to see:

- Tighter behavior in `handleZeroaryOps` for the `.seq .Empty` fallback (see inline).
- Slightly decoupled pattern match using the existing `seqTy` `match_pattern` helper.
- Broader test coverage: non-`int` element types (`bool`, `string`), nested sequences, and `Sequence.empty<Map _ _>()`. I verified manually that all three work with the current code, but the regressions would slip past CI with only the current `int` test.
- A few tiny doc / commit-message nits (commit says "Resolves #1027", PR body / issue scope say it only addresses half of it).
- Some proof / `#guard` coverage of the translation shape (see inline on `Translate.lean`).

Refactoring notes:

- `handleZeroaryOps` is now asymmetric — only one of four cases consumes `opTy`. If we ever gain another 0-ary polymorphic op, growing this match further will get ugly. Worth a small helper that normalizes `opTy` → element type where relevant.
- The silent `CoreType.tvar default unknownTypeVar` fallback mirrors the other unknown-type paths in this file, but elsewhere those paths call `ToCSTM.logError`. Sticking to that convention here would help catch well-typed-but-malformed inputs.

StrataTest/Languages/Core/Examples/Seq.lean
atomb: Doesn't squash-merging use the PR description as the commit message? I don't think it includes messages from the intermediate commits on the branch.

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -103,9 +103,10 @@ fn map_get (K : Type, V : Type, m : Map K V, k : K) : V => m "[" k "]";
 fn map_set (K : Type, V : Type, m : Map K V, k : K, v : V) : Map K V =>
   m "[" k ":=" v "]";
 
-// TODO: seq_empty is not yet supported in the grammar because the DDM parser
-// cannot currently handle 0-ary polymorphic functions (no arguments to infer
-// the type parameter from). The Factory definition exists for programmatic use.
+// Unlike other seq_* ops, seq_empty has no value arguments from which DDM can
+// infer the element type, so the type argument must be explicit in surface syntax.
+fn seq_empty (A : Type) : Sequence A =>
+  "Sequence.empty" "<" A ">" "(" ")";
```
MikaelMayer, line 109: We need also the roundtrip of pretty-printing it that way

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -103,9 +103,10 @@ fn map_get (K : Type, V : Type, m : Map K V, k : K) : V => m "[" k "]";
 fn map_set (K : Type, V : Type, m : Map K V, k : K, v : V) : Map K V =>
   m "[" k ":=" v "]";
 
-// TODO: seq_empty is not yet supported in the grammar because the DDM parser
-// cannot currently handle 0-ary polymorphic functions (no arguments to infer
-// the type parameter from). The Factory definition exists for programmatic use.
+// Unlike other seq_* ops, seq_empty has no value arguments from which DDM can
+// infer the element type, so the type argument must be explicit in surface syntax.
+fn seq_empty (A : Type) : Sequence A =>
```
MikaelMayer, line 108: This PR's functionality is also implemented in https://github.com/strata-org/Strata/pull/1100 but could be merged before.

MikaelMayer:
Everything unique in this PR was imported into https://github.com/strata-org/Strata/pull/1100 which was just merged

PR #1039, opened 2026-04-24
Fix #1038 quantifier bvar order bug

### LINKED ISSUE #1038: Bug in higher-order function+axiom typechecking
```
def axiomApplyBoundVar :=
#strata
program Core;

function apply(f : int -> int, x : int) : int;
axiom forall f : int -> int, x : int :: apply(f, x) == f(x);
#end

#guard_msgs in
#eval (Std.format ((Core.typeCheck .default (translate axiomApplyBoundVar).stripMetaData)))
```

Gives `error: Impossible to unify int with (arrow int $__ty2).`

joscoh:
*Issue #, if available:* #1038

*Description of changes:* Fixes bug where quantifier `bvars` were reversed leading to typechecker error. Includes regression tests.


shigoel (approved):

MikaelMayer (approved):
Nit: If the PR description says "Issue #, if available: #1038" it won't be detected as a fix of that issue. It should be "Fixes #1038"

PR #1041, opened 2026-04-24
Use original parameter names in symbolic evaluation output

MikaelMayer:
Use original parameter names in symbolic evaluation output

## Problem

The symbolic evaluator generated fresh variable names for procedure parameters (`$__x0`, `$__y1`, `$__r2`) instead of using the original names. This made the obligations program harder to read. Additionally, the parameter name `exp` (from a Python test) conflicted with the SMT-LIB `exp` function.

## Solution

Replace the `$__` prefix + gen counter naming scheme with a `usedNames` HashSet-based approach. Parameters now get their original names (e.g., `x`, `y`, `r`) with `@N` suffixes added only when disambiguation is needed. The SMT encoder's reserved keyword list was extended with nonlinear arithmetic symbols (`exp`, `sin`, `cos`, `tan`, `sqrt`, `pi`) to prevent solver conflicts.

Key changes:
- `EvalConfig` now tracks `usedNames : HashSet String` instead of `varPrefix`/`gen`
- `genSym` uses `findUnique` to prefer bare names, falling back to `@N` suffixes
- `ProcedureEval.eval` pre-marks parameter names as used so fvar names always differ from scope keys (preventing `old(x)` from resolving to the post-assignment value)
- SMT encoder sanitizes names starting with `@` or `.` via pipe-quoting
- Shared `Strata.Util.Name` module for disambiguation utilities

## Testing

All existing tests pass (updated ~40 test files with new parameter names). Formal verification proofs (`smtVCsCorrect`) continue to hold.

## Follow-ups

- [ ] Emit `var` declarations for parameters in the obligations program body so the program is self-contained


MikaelMayer:
🤖 Self-review completed. No code duplication, dead code, or anti-patterns found. All existing tests pass (4 pre-existing failures unrelated to this PR: z3/javac not available, VerifyPythonTest opaque body issue). The implementation correctly handles the `old(x)` mechanism by pre-marking parameter names as used so fvar names always differ from scope keys.

PR #1042, opened 2026-04-24
Add uncached test driver for external-IO tests

### LINKED ISSUE #1040: CI: VerifyPythonTest is never actually executed with Python
## Summary

`VerifyPythonTest` appears to pass on CI but is never actually executed with a Python interpreter. The tests silently skip in both CI jobs that touch them.

## Root cause

The `withPython` helper (in `StrataTest/Util/Python.lean`) silently returns `()` when Python is not found (with `warnOnSkip := false`). Every test in `VerifyPythonTest.lean` uses `#guard_msgs` with no expected output, so a skipped test produces empty output — which matches.

Two CI jobs are involved:

1. **`build_and_test_lean`** — compiles all of `StrataTest` including `VerifyPythonTest.lean`, but Python is not installed until *after* the Lean build completes (line 116 in `ci.yml`). So `withPython` skips every test, `#guard_msgs` sees empty output, and the `.olean` is cached as passing.

2. **`build_python`** — restores the `.lake` cache from step 1 and runs `lake build StrataTest.Languages.Python.VerifyPythonTest`. Since the `.olean` is already up to date, Lake does not recompile. The test never runs against Python.

## Impact

Regressions in the Python → Laurel → Core → SMT pipeline are not caught by CI. For example, a local build with Python 3.14 + `strata.gen` installed shows `Calculator@add` getting an Opaque body instead of Transparent (the test at line 243), but this failure is invisible on CI.

## Possible fixes

- **Move Python setup before the Lean build** in `build_and_test_lean`, so `VerifyPythonTest` compiles with Python available. This would require `strata.gen` to be installed before the Lean build step.
- **Invalidate the cached `.olean`** for Python test files in the `build_python` job (e.g., delete the relevant `.olean`/`.ilean` files before `lake build`), forcing recompilation with Python available.
- **Set `PYTHON_TEST=1`** during `build_and_test_lean` so that `withPython` throws instead of silently skipping when Python is missing — this would make the silent-skip problem visible but would require Python to be set up first.

## Relevant files

- `.github/workflows/ci.yml` — lines 88 (Lean build), 116 (Python setup), 263–307 (`build_python` job)
- `StrataTest/Util/Python.lean` — `withPython` (line 205), `findPython3` (line 173)
- `StrataTest/Languages/Python/VerifyPythonTest.lean`

joehendrix:
## Summary

Add an uncached test driver (`StrataTestMain`) that spawns `lean` per test file
and migrate tests depending on external tools to it, so Lake caching cannot mask
failures caused by tool environment changes.

## Usage

```bash
# Run all uncached tests
lake test

# Run only tests matching a prefix
lake test -- DDM.Integration.Java

# Combine include and exclude
lake test -- Languages.Python --exclude Languages.Python.VerifyPythonTest

# Exclude specific tests
lake test -- --exclude Languages.Python.VerifyPythonTest

# Show help
lake test -- --help
```

## Changes

- **Uncached test driver (`Scripts/StrataTestMain.lean`)**: Runs each `.lean` file
  under `StrataTestExtra/` as an independent `lean` process, bypassing Lake's
  elaboration cache. Supports module prefix filters and `--exclude` (which can be
  combined with includes). Tests run concurrently with deterministic result ordering.

- **Migrate external-IO tests to `StrataTestExtra/`**: Move 8 Python tests
  (DictNoneTest, PropertySummaryTest, AnalyzeLaurelTest, SpecsTest,
  PySpecsDirTest, RelativeImportTest, IdentifyOverloadsTest, VerifyPythonTest)
  and the Java integration test (DDM/Integration/Java/TestGen) out of the cached
  `StrataTest` library. These tests shell out to Python, `javac`, or `lake exe`
  and must re-run when the tool environment changes.

- **Simplify `withPython` (`StrataTest/Util/Python.lean`)**: Remove the
  `warnOnSkip` parameter and `PYTHON_TEST` environment variable logic. Now the test driver
  allows selectively excluding tests.

- **Replace hardcoded `/tmp` paths in Java TestGen**: Tests 12 and 15 used
  `/tmp/strata-java-test` and `/tmp/strata-javagen-preloaded-test`, which could
  collide in parallel runs and leaked on early-return errors. Replaced with
  `IO.FS.withTempDir` for automatic `try`/`finally` cleanup.

## CI changes

- **`build_and_test_lean`**: Exclude all Python tests (`--exclude Languages.Python`)
  since they now run in the dedicated `build_python` job. Uses `lake test` directly
  instead of lean-action's `test-args` input, which has a
  [bug](https://github.com/leanprover/lean-action) where `TEST_ARGS` is not
  passed as an environment variable to the test script.

- **`build_python`**: Replace the explicit `lake build` of individual test modules
  with `lake test -- Languages.Python --exclude Languages.Python.VerifyPythonTest`.
  This uses the new test driver to run all Python tests except the pre-existing
  `VerifyPythonTest` failure, and automatically picks up new Python test files
  without CI changes.

- **`check_pending_python`** and other steps: Update hardcoded paths from
  `StrataTest/DDM/Integration/Java/` to `StrataTestExtra/DDM/Integration/Java/`
  for the ion-java JAR download and testdata regeneration verification.

## Known limitations

- **No cross-imports within `StrataTestExtra/`**: Files under `StrataTestExtra/`
  can import from compiled `lean_lib`s (e.g., `Strata`, `StrataTest`) but cannot
  import from each other. This is intentional — since each file is spawned as an
  independent `lean` process, there are no intra-`StrataTestExtra` dependencies to
  manage, which keeps concurrent execution straightforward. Supporting cross-imports
  would require building an import dependency graph, which could be added in a
  future PR if needed.

This could be used to fix #1040 one that is building.


tautschnig:
I trust that the `VerifyPythonTest` exclusion (`--exclude Languages.Python.VerifyPythonTest`) is just temporary until we have fixed #1040 (and I am working on that fix).

tautschnig:
> I trust that the `VerifyPythonTest` exclusion (`--exclude Languages.Python.VerifyPythonTest`) is just temporary until we have fixed #1040 (and I am working on that fix).

See #1047.

joehendrix:
Yes, the `--exclude ..VerifyPythonTests` flag was added specifically since fixing that test seemed out of scope for this PR's work.  Once it is working, the `--exclude` flag can be dropped.

robin-aws (approved):
Overall I think this is a good idea. Given there's no way to coerce `lake build` to build a single test without the `.lake` cache (`--no-cache` only means don't download remote caches), and `lake test` doesn't even let you filter what to build at all (which I did not realize until trying it just now), this test runner seems valuable, and might even be useful to any Lean project!

I don't have confidence that you've identified and moved all of the tests with IO into `StrataTestExtra`, but as you said elsewhere I think I have the Lean static analysis tool to overapproximate that myself, and we can use that to find others if and when I get that merged. I might also merge `StrataTestExtra` back into `StrataTest` if we no longer have to be so careful about that division.

joehendrix:
Indeed, I haven't moved all tests with IO just since I thought a few may be controversial.

Here's the current tests with IO that are still in StrataTest:

- **`StrataTest/Util/Python.lean`** — `#eval miseWhere "Python@1.0"` spawns `mise` if installed; returns `none` in most environments but has a real external tool dependency at elaboration time. Kept because it's a utility module imported by all Python tests in `StrataTestExtra/` (cross-imports within `StrataTestExtra/` are not supported), and the test is a very basic sanity check of mise that is low risk for caching issues.

- **`StrataTest/Util/IO.lean`** — Temp file read/write round-trips. Local filesystem only, no external tool dependencies.

- **`StrataTest/Languages/Core/Examples/AlwaysRunSMT.lean`** — Writes `.smt2` files to a temp directory and verifies they exist. Local filesystem only; does not invoke external SMT solvers despite the name.

I think we could migrate any of these to StrataTestExtra in a future PR.

robin-aws:
> No cross-imports within StrataTestExtra/ ...

I just noticed there's a `lake lean <file>` command. Any chance using that instead of bare `lean` would address this?

joehendrix:
I'm not sure the differences between `lake lean` and `lean`.  Both use the lean toolchain version.  Presumedly `lake lean` may include additional context in the path.

My impression is that if we want modules in StrataTestExtra, then we should setup a build directory to cache oleans, and I'd prefer to only do that if needed in a future PR.

tautschnig (approved):

PR #1043, opened 2026-04-24
Add support for List append

thanhnguyen-aws:
## Summary

Adds support for Python's `list.append(element)` method in the Python-to-Laurel translation pipeline. Since Laurel is a functional language without mutable data structures, `xs.append(e)` is modeled as a functional update: `xs = Any_List_append(xs, e)`.

## Motivation

`list.append()` is one of the most commonly used list operations in Python. Without this, any Python code that builds or extends lists via `append` could not be translated and verified.

### Runtime (`PythonRuntimeLaurelPart.lean`)

- **`List_append`**: Recursively traverses a `ListAny` and appends an element at the end, returning a new list. Mirrors the structure of the existing `List_extend`.
- **`Any_List_append`**: Wrapper that unwraps an `Any`-typed list, calls `List_append`, and re-wraps the result. This is the entry point used by translated code.

#### Function name resolution
- Added `"ListAny@append" => "Any_List_append"` mapping in `reMapFunctionName` so that method-call mangling (`ListAny@append`) resolves to the new runtime function.
- In `refineFunctionCallExpr`, the resolved method name now goes through `reMapFunctionName` (previously returned as-is). A guard rejects `append` calls where the receiver is not a simple variable (e.g., `[1,2].append(3)` is rejected with an explicit error).

#### Type inference improvements
- `inferExprType` now returns `ListAny` for list literals and `DictStrAny` for dict literals, instead of falling through to `Any`.
- `translateType` short-circuits for known `Any`-family types via `isOfAnyType`, avoiding a lookup miss in `importedSymbols`.
- `collectDeclaredNamesAndTypes` uses `inferExprType` as a fallback when `inferClassTypeFromLaurelExpr` returns `none`, so variables initialized from list/dict literals get their concrete type recorded.
- `createVarDeclStmtsAndCtx` now preserves the inferred type in `variableTypes` instead of collapsing non-composite types to `Any`. This is necessary for the translator to later recognize a variable as a `ListAny` when it appears as the receiver of `.append()`.

#### Translation of `append` calls
- In `translateCall`, when the receiver is a `ListAny`-typed variable, the call is translated as an assignment back to the receiver: `xs = Any_List_append(xs, e)`. This models Python's in-place mutation as a functional update in Laurel.
- When `append` appears on the RHS of an assignment (e.g., `y = xs.append(4)`), `translateAssign` detects the resulting `.Assign` node from the RHS translation and emits the list mutation followed by assigning `None` to the LHS — matching Python's semantics where `list.append` returns `None`.
- `freeVar`, `maybeExceptVar`, and `nullcall_var` are moved earlier in the file so they are available in `translateCall`.
- Added a path in `translateCall` for functions found in `importedSymbols` as `ImportedSymbol.function` (not just `procedure`), so that `Any_List_append` (which returns a value) is emitted correctly when a first-arg receiver is present.

### Tests

| File | Description |
|------|-------------|
| `test_list_append.py` | Basic `xs.append(e)` followed by element access assertion (moved from `pending/`) |
| `test_mixed_list_ops.py` | Combines `append` with `__setitem__` and index access (moved from `pending/`) |
| `test_list_append_assign.py` | Verifies `y = xs.append(4)` mutates `xs` and assigns `None` to `y` (new) |

All three tests pass with expected output files added.

## Limitations

- Only `xs.append(e)` where `xs` is a simple variable name is supported.
- Calling `append` on an arbitrary expression (e.g., `get_list().append(e)` or `self.items.append(e)`) is explicitly rejected with an error.

## Testing

All new and existing tests pass. The three test cases cover:
1. Basic append and element verification
2. Append combined with other list mutations (`__setitem__`)
3. Assignment semantics (`y = xs.append(4)` → `y is None`)

---


thanhnguyen-aws:
@keyboardDrummer-bot  Please rewrite and add more details to the PR description.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1194,7 +1210,7 @@ partial def translateCall (ctx : TranslationContext)
       if let .Name _ n _ := arg then
         match ctx.variableTypes.find? (λ v => Prod.fst v == n.val) with
         | some (varName, ty) =>
-          if ty == PyLauType.Any then
+          if ¬ isCompositeType ctx ty then
```
tautschnig: **The havoc scope is too broad.** Previously (before `c5814f918`) this condition was `ty == PyLauType.Any`. Now it's `¬ isCompositeType ctx ty`, which includes:

- `Any`, `ListAny`, `DictStrAny` — correctly havoced (mutable, can be aliased).
- `Int`, `Str`, `Bool`, `Float`, `Bytes`, `Datetime` — **incorrectly havoced** (Python primitives are immutable; `f(x)` where `x: int` cannot mutate `x`).
- Composite types — excluded (this is arguably also wrong, since Python class instances are reference types with mutable attributes — but that's pre-existing behaviour, separate concern).

Concrete consequence: a test like
```python
def f(x): pass
x: int = 5
f(x)
assert x == 5
```
will no longer prove after this PR, even though it's trivially true in Python.

Suggestion — restrict to the *actually-mutable* Any-family types:

```suggestion
          if ty == PyLauType.Any ∨ ty == PyLauType.ListAny ∨ ty == PyLauType.DictStrAny then
```

Rationale for the comment at the match site: primitives are immutable in Python so there is nothing for a call to mutate; `Any`-family types can wrap a mutable reference and have to be havoced conservatively; composite types have known call semantics and are tracked elsewhere. If you intend to widen havoc coverage to composite types as a separate follow-up, that's fine — but the current change catches primitives, which weakens the verifier without cause.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1300,6 +1318,14 @@ partial def translateCall (ctx : TranslationContext)
...
       return mkStmtExprMd (.Block (typeAsserts.toList ++ [call]) none)
   else
+  match opt_firstarg with
+  | some (.Name _ n _) => match ctx.variableTypes.find? (λ v => v.fst == n.val) with
+      | some (_, "ListAny") =>
+          let trans_args ← args.mapM (translateExpr ctx)
+          let rhs ← emitCall ((freeVarExpr n.val)::trans_args)
+          return mkStmtExprMd $ .Assign [freeVarMd n.val] rhs
+      | _ => pure ()
+  | _ => pure ()
```
tautschnig, line 1341: **Aliasing divergence — please document inline.** This branch rewrites `xs.append(e)` to `xs = Any_List_append(xs, e)`. For the primary use case (no aliasing) this is correct. For aliased lists it silently diverges from CPython:

```python
xs = [1, 2, 3]
ys = xs          # same list object
ys.append(4)
assert xs == [1, 2, 3, 4]  # CPython: passes; our translation: fails (xs still [1,2,3])
```

At minimum, add an inline comment flagging the known limitation, with a reference to a tracking issue:

```suggestion
  -- For the primary case `xs.append(e)` where `xs` is a simple variable,
  -- model Python's in-place mutation as a functional update `xs = Any_List_append(xs, e)`.
  --
  -- KNOWN LIMITATION (Python aliasing): if `xs` is aliased (e.g. `ys = xs; ys.append(e)`),
  -- only the binding on the LHS (`ys` here) is updated; other bindings (`xs`) still
  -- reference the original list, which diverges from CPython's in-place-mutation semantics.
  -- Tracking: <issue-link-goes-here>
  match opt_firstarg with
  | some (.Name _ n _) => match ctx.variableTypes.find? (λ v => v.fst == n.val) with
      | some (_, "ListAny") =>
          let trans_args ← args.mapM (translateExpr ctx)
          let rhs ← emitCall ((freeVarExpr n.val)::trans_args)
          return mkStmtExprMd $ .Assign [freeVarMd n.val] rhs
      | _ => pure ()
  | _ => pure ()
```

(File or link an issue, then replace `<issue-link-goes-here>`.) The goal is that a future reader running into the divergence has one line to grep for instead of re-deriving the semantics gap.

Also worth filing a follow-up test (marked `xfail` or `pending/`) that exercises the aliasing case, so the day aliasing support lands there is an already-present fixture to flip.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1429,6 +1451,12 @@ partial def translateAssign  (ctx : TranslationContext)
               let varType := mkHighTypeMd (.UserDefined className)
               let newStmt := mkVarDeclInitWithLoc n.val varType rhs_trans source
               [newStmt]
+        | .Assign _ _ =>
+            -- Only occurs when the right hand side is a Python built-in List or Dict procedure
+            if rhsIsCall then
+              [rhs_trans, mkStmtExprMdWithLoc (StmtExpr.Assign [target] AnyNone) source]
+            else
+              [mkStmtExprMdWithLoc (StmtExpr.Assign [target] rhs_trans) source]
         | _ => [mkStmtExprMdWithLoc (StmtExpr.Assign [target] rhs_trans) source]
```
tautschnig, line 1473: **`translateAssign` is pattern-matching on the RHS being a *statement* node**. `translateCall` returns a value that is supposed to be a `StmtExprMd` usable as an expression, but the new `xs.append(e)` path returns a `.Assign` (a *statement*) and this branch catches and unpacks it, emitting two statements (`xs = Any_List_append(xs, e); target = None`).

This is fragile in two ways:

1. A future change to `translateCall` that emits a different shape for mutating methods (e.g. `DictStrAny.pop` returning the popped value) will break this pattern match silently. The `| _ =>` fallback at line 1460 quietly mishandles — it would try to embed a statement as an RHS.
2. The invariant "`translateCall` sometimes returns a statement" is not documented at `translateCall` itself. Grepping for `.Assign` in `translateAssign` doesn't tell you where the statement came from.

Two cleaner alternatives:

**(a)** Return a dedicated `Sum StmtExprMd Statement` (or `Either ⟨StmtExprMd⟩ ⟨List StmtExprMd⟩`) from `translateCall`, so the caller must explicitly handle both cases. Slightly more plumbing but makes the invariant type-level-enforced.

**(b)** Detect the `xs.append(...)` shape at the *statement* level in `translateStmt` (the `.Expr (.Call …)` case and `.Assign _ (.Call …)` case) and emit the two-statement sequence there directly, leaving `translateCall` pure for expression contexts.

Option (b) is closer to how `del` statements are handled elsewhere in the file and avoids introducing a new return-shape contract. Happy to leave the current approach as a first pass, but please add an inline comment at the `translateCall` site documenting that the `.Assign` return shape is intentional and where it is consumed:

```lean
-- NOTE: for `xs.append(e)` shape we return an `.Assign` statement wrapped in
-- StmtExprMd as a signal to `translateAssign` to emit the mutation + None-assign
-- sequence. See `translateAssign` arm at line 1454.
```

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1037,7 +1047,13 @@ partial def refineFunctionCallExpr (ctx : TranslationContext) (func: Python.expr
             match ctx.importedSymbols[callerTy]? with
             | some (ImportedSymbol.compositeType laurelName) => laurelName
             | _ => callerTy
-          return (manglePythonMethod resolvedTy callname, some v, false)
+          if resolvedTy == PyLauType.ListAny then
+            if let .Name _ _ _ := v then
+              pure ()
+            else
+              throw (.unsupportedConstruct "List procedure calleer not a variable is unsupported" (toString (repr v)))
```
tautschnig: Minor — typo and wording:

```suggestion
              throw (.unsupportedConstruct "Method call on a non-variable list receiver is not supported; only `name.method(...)` where `name` is a plain variable" (toString (repr v)))
```

The original had a typo (`calleer`) and used "List procedure", which doesn't match Python vocabulary — `append` is a *method*, and in Python terminology "procedure" usually refers to a void function. The replacement text also tells the user the exact supported shape so they can fix their code.

Also worth considering: this check fires only for `resolvedTy == PyLauType.ListAny`. When `DictStrAny`-method support lands (`d.pop`, `d.clear`, etc.), the same constraint applies and the check will need to generalise. A helper like `requireSimpleReceiverForMutatingMethod (ty : String) (v : expr) : Except …` called in both places would keep the constraint in one spot.

StrataTest/Languages/Python/tests/test_list_append_assign.py
```diff
@@ -0,0 +1,5 @@
+def test():
+    xs = [1, 2, 3]
+    y = xs.append(4)
+    assert xs == [1, 2, 3, 4]
+    assert y == None
```
tautschnig: Two small issues:

1. The file is missing the `test()` invocation at end-of-file. Every other fixture in this suite (`test_list_append.py`, `test_mixed_list_ops.py`, `test_del_*.py`, …) ends with `test()` or `main()`. The verifier analyses the function body regardless, so the `#guard_msgs` snapshot still passes, but the file diverges from what CPython would actually execute if you `python test_list_append_assign.py`.

2. There is trailing whitespace on the last line (per the diff: `    assert y == None  `). Minor lint, but will be flagged by most tooling.

Suggestion:

```suggestion
def test():
    xs = [1, 2, 3]
    y = xs.append(4)
    assert xs == [1, 2, 3, 4]
    assert y == None
test()
```

While here, consider also adding a second fixture for the *aliasing* case to flag the functional-update model's divergence from CPython — see the main review body:

```python
# test_list_append_aliasing.py — documents the functional-update model's
# divergence from CPython. Currently fails; see translateCall comment at line 1321.
def test():
    xs = [1, 2, 3]
    ys = xs
    ys.append(4)
    # CPython: passes. Our translation: xs is unchanged because the functional
    # update only rebinds `ys`.
    assert xs == [1, 2, 3, 4]
test()
```

(Mark this as `xfail` / keep it in a `pending/` directory if appropriate, so it's discoverable when aliasing support lands.)

PR #1044, opened 2026-04-24
feat(transform): Add optional SSA transformation for Strata Core

sagjoshi:
Introduce a Static Single Assignment (SSA) transformation that converts
procedure bodies so every variable is assigned exactly once via init.
Mutations (set, havoc) become fresh init declarations. At if-then-else
join points, conditional init expressions merge divergent variable
versions.

The transformation:
- Runs after callElim and loopElim (no calls or loops expected)
- Converts set to init with fresh variable names (x_0, x_1, ...)
- Converts havoc to nondet init
- Rewrites expressions to reference current SSA variable versions
- Emits conditional merge inits at if-then-else join points
- Uses modelPreserving validation (semantics-preserving)

Integration:
- Available as --pass ssa in the transform CLI command
- Available as TransformPass.ssa in the SimpleAPI
- Not part of the default verification pipeline (opt-in)

Files:
- Strata/Transform/SSA.lean: transformation implementation
- StrataTest/Transform/SSA.lean: 16 tests covering straight-line,
  if-then-else, havoc, nested branches, multiple variables, inout
  params, and pipeline chaining
- Strata/SimpleAPI.lean: TransformPass.ssa + applyPass integration
- StrataMain.lean: ssa added to valid transform passes

### DIFF
--- a/StrataTest/Transform/SSA.lean
+++ b/StrataTest/Transform/SSA.lean
@@ -1,0 +1,320 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+import Strata.DDM.Integration.Lean
+import Strata.Languages.Core.Core
+import Strata.Languages.Core.DDMTransform.Translate
+import Strata.Transform.SSA
+import Strata.SimpleAPI
+
+open Core
+open Core.SSA
+open Strata
+
+private def translate (p : Strata.Program) : Core.Program :=
+  let (program, _) := Core.getProgram p
+  program
+
+private def runSSA (p : Core.Program) : Core.Program :=
+  match ssaTransform p with
+  | .ok (res, _) => res
+  | .error e => panic! e
+
+/-! ## SSA Basic Tests -/
+section SSABasicTests
+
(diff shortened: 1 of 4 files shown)


aqjune-aws:
Is the conditional merge init equivalent to the phi(Φ) node in the SSA literature? I have two questions in a high level:
(1) How are the code locations to insert conditional merge init statements(= phi node) decided? A classic algorithm for SSA transformation will require a concept of dominance frontier and it will decide where to insert the phi node. Or, it is possible to eagerly insert the phi nodes whenever a variable is assigned & used in different basic blocks.
(2) Could you elaborate about motivation for introducing an SSA pass?

StrataTest/Transform/SSA.lean
```diff
@@ -0,0 +1,322 @@
...
+  } else {
+    x := 1;
+  }
+  y := x;
+};
+#end
+
+/-- info: true -/
+#guard_msgs in
+#eval! do
```
joehendrix: I'd recommend using `#eval`.  The `#eval!`  is probably from some run with sorry.

sagjoshi: Updated

PR #1047, opened 2026-04-24
Fix: accept Opaque-with-implementation in assertTransparent test

tautschnig:
PR #754 unified translateFunction and translateMethod, which applies return type constraints to class methods. This makes method bodies Opaque (with postconditions and implementation) instead of Transparent.

The Opaque body with an implementation is semantically correct — the verifier can still reason about the body AND enforce the return type constraint. The test's assertTransparent was overly strict, rejecting Opaque bodies even when an implementation was present.

Fix: accept both Transparent and Opaque-with-implementation in assertTransparent, since both provide the body for verification.


thanhnguyen-aws (dismissed):

shigoel (approved):
Looks good. The `assertHasBody` rename and expanded pattern match correctly capture the intent — a procedure has a body available for verification whether it's Transparent or Opaque-with-implementation. CI and `warnOnSkip` changes are coherent.

thanhnguyen-aws (approved):

PR #1048, opened 2026-04-24
PySpec pipeline: fix type mismatches, add type assertions, and clean up translation

joehendrix:
## Overview

Overhaul the PySpec-to-Laurel translation pipeline to fix type incompatibilities between pyspec and Python-source code paths, generate accurate type assertions in procedure bodies, and clean up dead/redundant code. These changes improve verification accuracy for pyspec-annotated libraries by ensuring type constraints are generated consistently with the Laurel runtime type system.

## Issues Fixed

**Cross-module type references resolve incorrectly.** `specArgToFuncDeclArg` in PySpecPipeline.lean derived Laurel type names using the function's module prefix rather than the type's own `PythonIdent.pythonModule`. A pyspec function in module `foo` referencing a type from module `bar.baz` would produce `foo_MyClass` instead of `bar_baz_MyClass`. Fixed by introducing `PythonIdent.toLaurelName` which uses the type's own module, and `specArgLaurelType` which mirrors `specTypeToLaurelType`'s logic.

**Redundant call-site type preconditions.** PySpec procedure bodies already generate `assert Any..isfrom_int(x)` etc. via `buildSpecBody`, but `specArgToFuncDeclArg` also populated `typeTesters`, causing PythonToLaurel to emit duplicate type constraints at each call site. Fixed by setting `typeTesters := #[]` for pyspec args and removing the now-unused `specArgTypeTesters` function.

**`typeTester?` only handled singleton types.** The old `typeTester?` used `SpecType.asIdent` which requires exactly one atom, silently producing no assertion for union types like `Union[int, str]`. Replaced with `typeAssertion?` which iterates all atoms and builds a disjunction (`isfrom_int(v) || isfrom_str(v)`).

**Missing type testers for common builtins.** `toTypeTester?` only covered `int`, `str`, `bool`, `float`, and `None`. Added mappings for `bytes` → `Any..isfrom_bytes`, `List`/`Sequence` → `Any..isfrom_ListAny`, `Dict`/`Mapping` → `Any..isfrom_DictStrAny`, and `Exception` → `Any..isexception`, matching the `Any` datatype constructors in the Laurel Python prelude.

**No assertions for literal types.** `Literal[42]` and `Literal["hello"]` produced no type assertions. `atomAssertion?` now generates `isfrom_int(v) && as_int!(v) == 42` for int literals and `isfrom_str(v) && as_string!(v) == "hello"` for string literals. TypedDict atoms are approximated as `isfrom_DictStrAny` with a warning.

**Unsupported types in unions silently dropped.** When a union contains atoms without testers (e.g., user-defined classes in `Union[None, MyClass]`), the old code silently skipped them. `typeAssertion?` now warns via `unsupportedUnion` for ident atoms without testers in unions and for TypedDict atoms. Singleton composite types (user-defined classes used alone) are not warned — they are expected to have no tester.

**Redundant `!isfrom_None` checks for typed parameters.** `buildSpecBody` had separate loops for type assertions and required-param checks. A required `int` parameter would get both `assert isfrom_int(x)` and `assert !isfrom_None(x)`, but the first already implies the second. Combined into a single loop: typed params get only the type assertion; untyped required params still get the `!isfrom_None` check.

**`PyArgInfo.tys` was a raw `List String` coupling pyspec and Python paths.** Replaced with structured `laurelType : HighTypeMd` and `typeTesters : Array String` fields. Similarly, `PythonFunctionDecl.ret` changed from `Option (List String × MetaData)` to `Option PyRetInfo`. All construction and consumption sites updated.

## Upstream Compatibility Fixes

Three targeted fixes address regressions exposed by the more complete pipeline translation, all caused by pre-existing issues in upstream data or earlier translation stages:

**Skip return-type `assume` for NoneType (ToLaurel.lean).** Generated Python stubs declare `-> None` for API methods that actually return response dicts. With transparent procedure bodies, `buildSpecBody` now generates `assume Any..isfrom_None(result)`, making all downstream field access on the return value unreachable — a false positive. The fix skips the return-type `assume` when the declared type is `NoneType`. This is a temporary workaround pending a stub-generator fix to emit correct return types. Resolves benchmark regressions.

**Return Hole for module attribute access (PythonToLaurel.lean).** `sys.argv`, `os.path`, and similar module-level attribute accesses generated `FieldSelect (Identifier "sys") "argv"`, but imported modules are not declared as Laurel variables. The fix returns `Hole` when the object expression is a known package (via `isPackage`), matching the existing handling for unmodeled package function calls. Resolves benchmark regression.

**Drop bare `raise` instead of emitting untyped Hole (PythonToLaurel.lean).** A bare `raise` inside an exception handler produced a Hole whose return type could not be inferred, leaving it as `Unknown` and failing the Laurel-to-Core translator. The fix emits no statements for `raise`, which is safe under the current exception model where propagation is tracked via `maybe_except`/`isError` checks. This is a placeholder pending proper exception modeling. Resolves benchmark regression.

## Additional Details

- `toTypeTester?` refactored from an if-else chain to a `Std.HashMap` lookup (`typeTestersMap`).
- `SpecType.mk` made private; callers use smart constructors (`ident`, `noneType`, `intLiteral`, `stringLiteral`, `typedDict`, `unionArray`). `sizeOf_atom_lt_of_mem` theorem added for termination proofs.
- Duplicate-element bug fixed in `unionAux` (equal elements were double-counted).
- Dead `pythonToLaurel` wrapper removed; `pythonToLaurel'` renamed to `pythonToLaurel`.
- Dead `PythonIdent.toPyLauType?` removed (replaced by `toTypeTester?` / `specArgLaurelType`).
- Dead `emptyType` warning kind and unreachable code branch removed.
- Unified `transExpr` replaces ~10 separate expression translators (`extractSubject`, `transAssertExpr`, `transCondition`, `transMessageExpr`, `inferExprType`, `extractIntBound`, `extractFloatBound`, `collectEnumValues`, `isIntType`, `isFloatType`, `lookupTypedDictField`, `extractElementType`, `extractDictKeyValueTypes`) with a single recursive function that returns `(SpecExpr × SpecType)`. This enables type-directed translation (e.g., choosing int vs float comparison) and eliminates duplicate pattern matching.
- `SpecExpr.len` renamed to `SpecExpr.stringLen` for clarity.
- `extractFunctionSignatures` uses `Array` instead of `List` accumulator.
- `runTest` in ToLaurelTest.lean now prints warnings before procedures so `#guard_msgs` can verify them.
- `SpecAssertionContext.kwargs` simplified from separate `kwargsParamName`/`kwargsType` fields to a single `Option (String × SpecType)`.
- `assumeCondition` simplified to always take a condition (no `Option` wrapper).
- `PySpecMClass.runNoWarn` added to check that an action produces no errors or warnings.

## Test Changes

- **ToLaurelTest.lean**: All `SpecAtomType` references replaced with `SpecType` smart constructors. Body structure tests updated to reflect combined type/required-param assertions (typed params no longer have `!isfrom_None`). Expected output updated for new warnings on `bytearray`, `complex`, user-defined types in unions, and TypedDict approximation.
- **PySpecArgTypeTest.lean**: Updated to verify `typeTesters` is empty (body assertions make call-site ones redundant). Added end-to-end test using `formatProcedure` to verify full procedure body through DDM round-trip, confirming body contains `assert Any..isfrom_int(x)`, `assert Any..isfrom_str(y)`, and `assume Any..isfrom_float(result)`.
- **DeclsTest.lean**: Migrated to use `unionArray` for dedup tests.
- **SpecsTest.lean**: Minor import path update.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the MIT license.


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -486,6 +502,12 @@ def ListAny_mk (es: List StmtExprMd) : StmtExprMd := match es with
   | [] => mkStmtExprMd (.StaticCall "ListAny_nil" [])
   | e::t => mkStmtExprMd (.StaticCall "ListAny_cons" [e, ListAny_mk t])
 
+def createBoolOrExpr (exprs: List StmtExprMd) : StmtExprMd :=
+  match exprs with
+  | [] => mkStmtExprMd (.LiteralBool true)
```
ssomayyajula: Shouldn't the unit of OR be false here?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -927,9 +953,10 @@ partial def getFunctionReturnType (ctx : TranslationContext) (func: Python.expr
   | .ok none =>
     let (fname, _) ← refineFunctionCallExpr ctx func
     match ctx.functionSignatures.find? (λ f => f.name == fname) with
-      | some funcDecl => match funcDecl.ret with
-        | some ([ty], _) => return ty
-        | _ => return PyLauType.Any
+      | some funcDecl =>
+        match funcDecl.ret with
+        | some retInfo => return highTypeToPyLauType retInfo.laurelType.val
```
ssomayyajula, line 948: Are there situations where `highTypeToPyLauType retInfo.laurelType.val` is not equal to the original raw type string contained by `retInfo` before this change?

Strata/Languages/Python/PythonToLaurel.lean
joehendrix: Good catch — fixed in 290806ce. The empty-list base case now returns `false` (the identity of OR). In practice this case wasn't reachable from current call sites since `createBoolOrExpr` is only called when `typeTesters` is non-empty, but it was still wrong.

Strata/Languages/Python/PythonToLaurel.lean
joehendrix: Yes — this was a real bug. `preludeSignatureToPythonFunctionDecl` was setting `laurelType := AnyTy` for all return types, so `highTypeToPyLauType` always returned `"Any"` instead of the actual type name (e.g. `"int"`, `"DictStrAny"`). The old code returned the raw type string directly from `getTypeName tp`. Fixed in 290806ce by storing `UserDefined (getTypeName tp)` so the round-trip is faithful.

ssomayyajula (approved):

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -249,19 +293,100 @@ private def removeAdjDups {α} [BEq α] (a : Array α) : Array α :=
...
+def stringLiteral (loc : SourceRange) (value : String) : SpecType :=
+  ofAtom loc (.stringLiteral value)
 
-def isAtom (tp : SpecType) (atp : SpecAtomType) : Bool := tp.asSingleton.any (· == atp)
+def typedDict (loc : SourceRange) (fields : Array String)
+    (fieldTypes : Array SpecType) (fieldRequired : Array Bool) : SpecType :=
+  ofAtom loc (.typedDict fields fieldTypes fieldRequired)
+
+def unionArray (loc : SourceRange) (elts : Array SpecType) : SpecType :=
+  { loc := loc, atoms := elts.foldl (init := #[]) (unionElts · ·.atoms) }
```
aqjune-aws, line 317: Claude said: replacing ofArray (qsort + dedup = O(n log n)) with iterative unionElts · ·.atoms is O(n²) worst-case when building a large union fold. Fine for typical hand-written Literal unions (size ≤ ~10), but it's 
  a quiet regression for auto-generated unions. Probably not worth fixing today, but worth noting in the PR description if there's any autogen in the stub pipeline.

aqjune-aws (approved):

Strata/Languages/Python/Specs/Decls.lean
joehendrix: Indeed.  ofArray wasn't really used except in a handful of test cases.  The next PR #1072 will address the actual underlying performance issue.

PR #1049, opened 2026-04-24
Add concrete syntax for lambdas and better error checks for higher-order constructs

joscoh:
*Issue #, if available:*

*Description of changes:*                                                                                                                                                                                                                             
  - **Lambda expressions in concrete syntax**: Added `fun x : T => body` syntax to the Core grammar (Grammar.lean) and CST→AST translation (Translate.lean), supporting single and multi-binding lambdas with correct de Bruijn index assignment.       
  - **Expression application**: Added `(expr)(arg)` syntax (`apply_expr`) to support applying arbitrary expressions (not just named operators), enabling direct lambda application and currying. Translates to standard `.app` in the AST.                 
  - **Informative SMT error messages**: The SMT encoder now produces actionable error messages when it encounters constructs it cannot encode — function-typed parameters, arrow-typed return values, lambda abstractions in function bodies, bare lambda expressions, and datatypes with function-typed constructors.                                                                                                                                                                                    
                                                                                                                                                                                                                                                           
  ## Tests                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                           
  - **LambdaHigherOrderTests.lean**: Comprehensive coverage of lambda/higher-order interactions with the rest of the language, including partial evaluation, recursive and non-recursive functions, polymorphism, and datatypes.                                                                                                                                                                                                                                                                                                         
  - **TestASTtoCST.lean**: Three new AST→CST round-trip tests for lambda formatting (applied lambda, curried lambda, higher-order lambda)  



Strata/DL/Lambda/LExpr.lean
```diff
@@ -340,6 +340,18 @@ def collectFvarNames {T : LExprParamsT} : LExpr T → List (Identifier T.base.ID
   | .eq _ e1 e2 => collectFvarNames e1 ++ collectFvarNames e2
   | _ => []
 
+/-- Checks if the expression contains a lambda abstraction anywhere. -/
```
MikaelMayer, line 343: If you are looking at lambdas but not at directly applied lambdas, you will miss let expressions

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -25,6 +25,7 @@ namespace Strata
 
 -- Sequence operations increase the grammar size enough to require a higher recursion limit.
 set_option maxRecDepth 10000
+set_option maxHeartbeats 400000
```
MikaelMayer, line 29: Nit: The comment on line 27 says "Sequence operations increase the grammar size enough to require a higher recursion limit" but now also covers `maxHeartbeats`. Consider updating the comment to mention both, or adding a separate comment for `maxHeartbeats` explaining what pushed it over the default.

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -687,6 +687,24 @@ def translateFn (ty? : Option LMonoTy) (q : QualifiedIdent) : TransM Core.Expres
 
 mutual
 
+partial
+def translateLambda
```
MikaelMayer: The first 6 lines of `translateLambda` (translateDeclList, computing `n`, `newBoundVars`, `boundVars'`, `xbindings`, and translating the body) are identical to the first 6 lines of `translateQuantifier`. Consider extracting a helper like `withScopedBindings` that takes `bindings`, `xsa`, `bodya`, and a continuation, to avoid this duplication.

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -587,6 +608,24 @@ partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Conte
...
+          .error f!"Cannot encode function '{func.name}' to SMT: \
+                    it returns a function type '{func.output}'. \
+                    Higher-order functions cannot be encoded to SMT. \
+                    Consider marking the function as `inline`."
+        else if func.body.any LExpr.hasAbs then
+          .error f!"Cannot encode function '{func.name}' to SMT: \
+                    its body contains a lambda expression. \
+                    Lambda abstractions cannot be encoded to SMT. \
+                    Consider marking the function as `inline`."
+        else
```
MikaelMayer, line 630: Nit: The `else` here followed by an unindented `let` on the next line reads oddly compared to the `else if` branches above which have indented bodies. Consider indenting the `let` block under `else` for consistency:
```
        else
          let (ctx, isNew) ←
```

MikaelMayer (commented):
Solid PR — the lambda concrete syntax, expression application, and SMT error messages are well-designed and thoroughly tested. The `unifyTypes` tvar-into-arrow propagation is well-motivated and clearly commented. The test coverage is excellent, with good variety across polymorphism, recursion, datatypes, and error paths.

A few items below worth addressing.

Strata/DL/Lambda/LExpr.lean
joscoh: The partial evaluator should beta-reduce these let-expressions already. If we change the behavior of the partial evaluator, then I agree that this check would have to change. I added a comment explaining this.

shigoel (commented):
Good PR — clean refactoring of `withScopedBindings`, thorough test coverage, and helpful SMT error messages.

shigoel (dismissed):

MikaelMayer (approved):

shigoel (approved):

PR #1052, opened 2026-04-24
Small-step Imperative semantics: loop/ite exit handling and invariant checking

aqjune-aws:
- Small-step `.exit` semantics: According to the previous small-step statement, `{ loop .. { exit ; } print "a"; } print "b" }` was exiting the outerblock as well and made this print "b", not "ab". This patch makes exit without destination block only escape the loop. For if { exit } else { exit }, this is consistently applied.

- Loop invariants are now checked by `step_loop_{,nondet_}{enter,exit}`: each invariant must evaluate to `tt`/`ff`, and any `ff` result ORs a fresh `hasInvFailure` flag into the env's cumulative `hasFailure`, mirroring how `step_cmd` threads assert failures. (refer to: Strata/DL/Imperative/StmtSemantics.lean)

- `Stmt.loop` invariants are now labeled pairs `List (String × P.Expr)` (distinct labels per invariant, like assert labels).  Downstream frontends, backends, and passive-encoding transforms are updated to carry the labels through. (refer to: Strata/DL/Imperative/Stmt.lean, Strata/DL/Imperative/StmtSemantics.lean)

- The Core syntax is extended to allow a label to invariants.
```
  while (i < n)
    invariant [test]: 0 <= i
  {
    i := (i + 1);
  }

```

*Issue #, if available:*

*Description of changes:*


Strata/Transform/StructuredToUnstructured.lean
```diff
@@ -123,10 +123,15 @@ match ss with
       pure ([initCmd, assumeCmd, lbCmd], ldec, [decBlock])
   -- Body jumps to bodyK (either directly to lentry, or through the decrease block)
   let (bl, bbs) ← stmtsToBlocks bodyK bss exitConts []
```
MikaelMayer: The loop body is translated with the same `exitConts` without prepending `(.none, kNext)`. With the new semantics, `step_loop_enter` wraps the body in `.block .none (...)`, so an unlabeled `exit` inside the loop body should terminate just the loop (jumping to `kNext`). But here, an unlabeled `exit` in `bss` will look up `(.none, ...)` in the inherited `exitConts` and jump to an enclosing block's continuation instead.

Should this be `stmtsToBlocks bodyK bss ((.none, kNext) :: exitConts) []`?

Strata/Transform/Specification.lean
```diff
@@ -567,11 +570,13 @@ theorem sound_id : Sound L L some := by
 
 /-! ## Overapproximate predicate
 
-`Overapproximates L₁ L₂ T` says that any terminal or exiting env reachable
-from `st` in `L₁` is also reachable from `T st` in `L₂`.
+`Overapproximates L₁ L₂ T` says that (1) any terminal or exiting env reachable
```
MikaelMayer: The section comment says `Overapproximates` includes "(2) if there is a state reachable from `st` in L₁ that fails an assertion, there also is a state reachable from `T st` in L₂ that fails an assertion." The actual definition below only has terminal/exiting env forwarding — property (2) is not present. Either the comment is stale or the definition is incomplete.

Strata/DL/Imperative/KleeneSemanticsProps.lean
```diff
@@ -0,0 +1,85 @@
...
+    StepKleeneStar P (EvalCmd P)
+      (.stmt (.cmd (.assume label expr md)) ρ₀) (.terminal ρ₀) := by
+  have raw : StepKleeneStar P (EvalCmd P)
+      (.stmt (.cmd (.assume label expr md)) ρ₀)
+      (.terminal { ρ₀ with store := ρ₀.store, hasFailure := ρ₀.hasFailure || false }) :=
+    .step _ _ _ (.step_cmd (EvalCmd.eval_assume hcond hwfb)) (.refl _)
+  rwa [assume_env_eq] at raw
+
+omit [HasVal P] [HasBoolVal P] in
+theorem kleene_assume_then
```
MikaelMayer: nit: `kleene_assume_then` is declared `public` but has no callers in the codebase. If it's intended for future use, a brief docstring explaining the intended use case would help; otherwise consider removing it to avoid dead code.

Strata/Transform/DetToKleeneCorrect.lean
```diff
@@ -329,51 +346,95 @@ private def loop_sim
...
+    match hstarT, hlen with
+    | .step _ _ _ (@StepStmt.step_loop_exit _ _ _ _ _ _ _ _ _ _ _ _
+        hasInvFailure _ _ hff_iff _) hrest, hlen =>
+      have h_no : hasInvFailure = false := by
+        cases hb : hasInvFailure with
+        | false => rfl
+        | true =>
+          rw [hb] at hff_iff
+          have ⟨_, hmem, _⟩ := hff_iff.mp rfl
+          exact ((List.mem_nil_iff _).mp hmem).elim
```
MikaelMayer: nit: The 6-line `h_no : hasInvFailure = false` proof (cases on `hasInvFailure`, derive `⟨_, hmem, _⟩` from `hff_iff`, then `List.mem_nil_iff`) is duplicated 4 times across `loop_sim` and `loop_sim_kleene`. A small helper like `private theorem empty_inv_no_failure (hff_iff : hasInvFailure = true ↔ ∃ le ∈ ([] : List _), ...) : hasInvFailure = false` would reduce the repetition.

MikaelMayer (commented):
Solid PR — the exit-handling fix (wrapping loop bodies in `.block .none`) is a clean design, and the invariant-checking additions are well-integrated into the existing `StepStmt` framework. The new `StepStmtTest.lean` tests are a welcome addition. A few items to address, one potentially significant (the `StructuredToUnstructured` exit-handling gap).

Strata/DL/Imperative/KleeneSemanticsProps.lean
```diff
@@ -0,0 +1,85 @@
...
+
+public section
+
+variable {P : PureExpr} [HasFvar P] [HasBool P] [HasNot P] [HasVal P] [HasBoolVal P]
+
+/-! ## Env helpers -/
+
+omit [HasFvar P] [HasVal P] [HasBool P] [HasNot P] in
+theorem assume_env_eq (ρ : Env P) :
+    ({ ρ with store := ρ.store, hasFailure := ρ.hasFailure || false } : Env P) = ρ := by
```
joscoh, line 27: Is this theorem just saying that `b || false = b`?

Strata/DL/Imperative/KleeneSemanticsProps.lean
```diff
@@ -0,0 +1,85 @@
...
+  cases ρ; simp [Bool.or_false]
+
+omit [HasFvar P] [HasNot P] in
+theorem eval_tt_is_tt
+    (δ : SemanticEval P) (σ : SemanticStore P)
+    (hwfv : WellFormedSemanticEvalVal δ) :
```
joscoh, line 33: Not directly related to this PR, but I believe that we do not have a `WellFormedSemanticEvalVal` for Lambda yet, right? At the very least, it would require an evaluator that gives `None` when it runs out of fuel or evaluates a term containing free vars/uninterpreted functions.

Strata/DL/Imperative/Stmt.lean
```diff
@@ -346,23 +352,23 @@ by an enclosing `block` — either within `s` itself or with a label in
 
 When `s.exitsCoveredByBlocks []`, execution of `s` can never produce `.exiting`. -/
 
-@[expose] def Stmt.exitsCoveredByBlocks : List String → Stmt P CmdT → Prop
+@[expose] def Stmt.exitsCoveredByBlocks : List (Option String) → Stmt P CmdT → Prop
```
joscoh, line 359: What is the idea behind making this take in a list of options? This function never adds/checks to see if None is in the list, right?

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -811,13 +839,21 @@ private theorem step_preserves_exitsCoveredByBlocks
   | step_ite_nondet_false => intro _ hwp; exact hwp.2
   | step_loop_enter _ _ =>
     intro labels hwp
+    -- Goal: (.block .none (.stmts (body ++ [.loop ...]) ρ')) covers labels
```
joscoh, line 842: Don't think we need this, but up to you.

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -1112,9 +1151,14 @@ private theorem noMatchingAssert_not_isAtAssert
...
   | .stmt (.cmd (.assume ..)) _
   | .stmt (.cmd (.cover ..)) _
-  | .stmt (.block ..) _ | .stmt (.ite ..) _ | .stmt (.loop ..) _
+  | .stmt (.block ..) _ | .stmt (.ite ..) _
   | .stmt (.exit ..) _ | .stmt (.funcDecl ..) _ | .stmt (.typeDecl ..) _ =>
     simp [isAtAssert]
+  | .stmt (.loop _ _ inv _ _) _ =>
+    simp [Config.noMatchingAssert, Stmt.noMatchingAssert] at hno
+    intro hat
+    -- hat : (label, expr) ∈ inv
```
joscoh: Dead code? (and same for below)

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -1306,8 +1356,41 @@ theorem assert_tail_getEvalStore
 
 /-! ## hasFailure preservation -/
 
+/-- Helper: when all asserts at a loop config pass (via `hv`), the
+    loop-step's `hasInvFailure` boolean is forced to `false`. -/
+private theorem loop_step_hasInvFailure_false
```
joscoh: Not directly related to this PR, but we should have better guidelines for what to put in `StmtSemantics` vs `StmtSemanticsProps` (and similar). This seems very similar to `loop_step_hasInvFailure_false_core` in `StmtSemanticsProps.lean`. Can these theorems be combined?

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -1320,18 +1403,34 @@ private theorem step_preserves_noFailure
...
   | step_block => simp [Config.getEnv]; exact hnf
   | step_ite_true _ _ => exact hnf
   | step_ite_false _ _ => exact hnf
   | step_ite_nondet_true => exact hnf
   | step_ite_nondet_false => exact hnf
-  | step_loop_enter _ _ => exact hnf
-  | step_loop_exit _ _ => exact hnf
-  | step_loop_nondet_enter => exact hnf
-  | step_loop_nondet_exit => exact hnf
+  | step_loop_enter _ _ hff_iff _ =>
```
joscoh: Is it possible to combine these into one case? i.e. does the following proof work:
```
| step_loop_enter _ _ hff_iff _  | step_loop_exit _ _ hff_iff _ | step_loop_nondet_enter _ hff_iff  | step_loop_nondet_exit _ hff_iff =>
   simp only [Config.getEnv]
   have := loop_step_hasInvFailure_false (P := P) (extendEval := extendEval) (hc_shape := rfl) hv hff_iff
   ...
```
Just wondering

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -1320,18 +1403,34 @@ private theorem step_preserves_noFailure
...
+    simp only [Config.getEnv]
+    have := loop_step_hasInvFailure_false (P := P) (extendEval := extendEval) (hc_shape := rfl) hv hff_iff
+    simp [Config.getEnv] at hnf
+    rw [hnf, Bool.false_or]; exact this
```
joscoh: Should not use Lean auto-generated names in proofs

Strata/Languages/Core/StatementSemanticsProps.lean
```diff
@@ -2392,20 +2427,36 @@ theorem core_step_preserves_noFailure
...
         rw [hff] at htt; exact absurd (Option.some.inj htt) HasBool.tt_is_not_ff.symm
       | _ => simp_all [Config.getEnv]
     | @call_sem _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ =>
         simp only [Config.getEnv, Bool.or_false]; exact hnf
   | step_block => simp [Config.getEnv]; exact hnf
   | step_ite_true _ _ => exact hnf
   | step_ite_false _ _ => exact hnf
-  | step_loop_enter _ _ => exact hnf
-  | step_loop_exit _ _ => exact hnf
+  | step_loop_enter _ _ hff_iff _ =>
```
joscoh: Similar comments as before (both for repeated cases and avoiding `this`)

Strata/Transform/DetToKleeneCorrect.lean
```diff
@@ -329,51 +346,95 @@ private def loop_sim
...
+        hasInvFailure hg _ hff_iff hwfb) hrest, hlen =>
+      have h_no : hasInvFailure = false := by
+        cases hb : hasInvFailure with
+        | false => rfl
+        | true =>
+          rw [hb] at hff_iff
+          have ⟨_, hmem, _⟩ := hff_iff.mp rfl
+          exact ((List.mem_nil_iff _).mp hmem).elim
+      subst h_no
+      -- post-state env = {ρ₀ with hasFailure := ρ₀.hasFailure || false}.
```
joscoh: Not sure this comment is necessary

Strata/Transform/DetToKleeneCorrect.lean
```diff
@@ -382,35 +443,80 @@ private def loop_sim_kleene
       StepKleeneStar P (EvalCmd P) (.stmt b ρ₀) (.terminal ρ'))
     (hcov : Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks (P := P) (CmdT := Cmd P) [] body)
     (hnofd_body : Block.noFuncDecl body = true)
-    (ρ₀ ρ' : Env P)
+    (ρ₀ ρ' : Env P) (n : Nat)
```
joscoh, line 447: Just curious, wondering why the induction on `hstarT` no longer suffices

StrataTest/Transform/ProcedureInlining.lean
```diff
@@ -87,12 +87,13 @@ private def alphaEquivExprsOpt (e1 e2: Option Expression.Expr) (map:IdMap)
...
 
-private def alphaEquivExprsList (l1 l2 : List Expression.Expr) (map : IdMap)
+private def alphaEquivExprsList (l1 l2 : List (String × Expression.Expr)) (map : IdMap)
     : Except Format Bool :=
   if l1.length != l2.length then
     .error "invariant lists have different lengths"
   else
-    return (l1.zip l2).all (fun (a, b) => alphaEquivExprs a b map)
+    return (l1.zip l2).all (fun ((lbl1, a), (lbl2, b)) =>
+      lbl1 == lbl2 && alphaEquivExprs a b map)
```
joscoh, line 96: Shouldn't alpha equivalence ignore labels here?

Strata/DL/Imperative/KleeneSemanticsProps.lean
aqjune-aws: Yes, but it seems to be useful for rewriting. :)

Strata/DL/Imperative/KleeneSemanticsProps.lean
aqjune-aws: Yes. I think this can be implemented by writing a wrapper of `LExpr.eval` that returns an optional value if the result of `LExpr.eval` is equal to the input for any reason. Making `LExpr.eval` return an optional value seems desirable, but since it is a partial evaluator, it can return anything and when it has to return None is slightly ambiguous...

Strata/DL/Imperative/KleeneSemanticsProps.lean
aqjune-aws: Removed

Strata/DL/Imperative/Stmt.lean
aqjune-aws: This is slightly tricky to simplify to `List String` because there is an `exit` statement without explicit destination label (`|  labels, .exit none _ => labels.length > 0`). In this case, existence of any `None` element in the List will matter. I updated the comment to describe this.

Strata/DL/Imperative/StmtSemantics.lean
aqjune-aws: Strata/Languages/Core/StatementSemanticsProps.lean is for the Core language whereas StmtSemantics.lean is for the generic Imperative. But Core is an  instantiation of Imperative, so I gave it a try to generalize `loop_step_hasInvFailure_false` and `step_preserves_noFailure` and made them take an arbitrary `Cmd` type - the result was slightly verbose because it had to carry some properties about the custom `Cmd` but it worked :)

Re. files for theorems it makes sense to discuss whether to put the theorems in a separate file or in a same file. Initially I thought that putting theorems into the file where definitions are was better because it helps keep some definitions private, but as the theorems grow files are growing big. Would it be a good idea if we have a systematic lint that checks theorems are in *Props.lean, for a list of interested files?

Strata/Languages/Core/StatementSemanticsProps.lean
aqjune-aws: This theorem is replaced by the more general version `Imperative.step_preserves_noFailure`.

Strata/Transform/DetToKleeneCorrect.lean
aqjune-aws: Fixed

Strata/Transform/DetToKleeneCorrect.lean
aqjune-aws: This might not be a satisfactory answer (TL;DR: the agent chose this), but if it helps - `loop_sim_kleene` proves that it is possible to simulate the Imperative `.loop` with `.nondet` guard with the Kleene language's `.loop`. The previous proof structure is somewhat straightforward - it does one iteration from Imperative's loop, then prove that it is possible to make the corresponding next iteration in Kleene. This explains why the proof was doing a match of `.step_loop_nondet_enter`.
After this patch, this 'alignment' became slightly ugly because now `.body` is introduced after `.step_loop_nondet_enter` in the Imperative side, but this `.body` doesn't exist on the Kleene side. The `.body` shouldn't matter because `hcov` states that there is no exit that makes the loop escaped, but still has to be proven. But with this overall complication the agent just falled back to the very generic style parameteric to the number of steps.

Strata/Transform/Specification.lean
aqjune-aws: Thanks, fixed

StrataTest/Transform/ProcedureInlining.lean
aqjune-aws: These compare loop invariants with their labels. I think it makes sense to check the labels even if this can be a bit strict test... Renamed the function name here.

Strata/DL/Imperative/KleeneSemanticsProps.lean
joscoh: I'm not sure this would work. `LExpr.eval e = e` in a few cases: when `e` is stuck, when `e` is a value, etc. Meanwhile, `LExpr.eval e` is likely different from `e` if the evaluator runs out of fuel, when it gets stuck along the way, and when it successfully reduces to a value, and it would be difficult to distinguish them. But that's a separate conversation :)

joscoh (dismissed):

Strata/DL/Imperative/StmtSemantics.lean
joscoh: Makes sense. Unfortunately there doesn't seem to be a good solution to the module/private issue. My rough feeling has been that for "general-purpose" lemmas, it could make sense to keep them in the same file as the definitions (e.g. the lemmas in `Factory.lean`), while for more specific results, they can move to a separate file. But that is pretty fuzzy and ambiguous, so I am not sure that's actually a good distinction. In my mind, having theorems in the same file as the definitions would be nice ideally but if there are too many (and the amount will only grow) then 1) the file gets very large and disorganized 2) the definitions get drowned out by a huge amount of "noise" in the theorems and proofs.

Strata/DL/Imperative/KleeneSemanticsProps.lean
aqjune-aws: Ah, I meant a loop that iterates when `LExpr.eval e != e`, and stops when `LExpr.eval e = e` holds. :) (sorry  for imprecise statement) But then the termination of this loop is a concern, and if `e` takes too many steps to reduce to a value or stuck then a seeming reasonable fuel will be insufficient... I think it will be great if there is a function that receives `e` and calculates the sufficient amount of fuel to make `LExpr.eval` either stuck or return a value. I feel this kind of topic will be discussed soon (slightly related to termination checking?)

Strata/DL/Imperative/StmtSemantics.lean
aqjune-aws: Yeah, makes sense. I think making the scope of refactoring tight & considering only definitions regarding semantics might be a good target. Made https://github.com/strata-org/Strata/issues/1067

Strata/DL/Imperative/KleeneSemanticsProps.lean
joscoh: Agree that we should think about this, but I think this will be hard; it is strictly harder than normalization. Termination checking is a necessary prerequisite but not enough; there is also reasoning about concreteEval executions, beta reduction, `substFvarsFromState`, etc.

Strata/Languages/C_Simp/DDMTransform/Translate.lean
```diff
@@ -275,14 +275,15 @@ def translateMeasure (bindings : TransBindings) (arg : Arg) : TransM (Option (LE
...
+def translateInvariant (bindings : TransBindings) (arg : Arg) :
+    TransM (List (String × LExpr CSimpLParams.mono)) := do
   translateOption (fun maybe_arg => do
                     match maybe_arg with
                     | none => return []
                     | some a =>
                       let e ← checkOpArg a q`C_Simp.invariant 1
                       assert! e.size == 1
-                      return [← translateExpr bindings e[0]!])
+                      return [("", ← translateExpr bindings e[0]!)])
```
MikaelMayer: Shouldn't the name come outside of the function rather than being hard-coded as an empty string here?

MikaelMayer (dismissed):

Strata/Languages/C_Simp/DDMTransform/Translate.lean
aqjune-aws: Resolved :)

StrataTest/Languages/C_Simp/Examples/LinearSearch.lean
```diff
@@ -66,7 +66,7 @@ info: function linearSearch {
   while
     (~Int.Lt idx (~Array.Len arr))
     (some (~Int.Sub (~Array.Len arr) idx))
-    [#true]
+    [[linearSearch_invariant_381_398]: #true]
```
joscoh, line 69: Is this going to re-introduce `guard_msgs` brittleness?

joscoh (approved):

StrataTest/Languages/C_Simp/Examples/LinearSearch.lean
aqjune-aws: Could be, but these numbers are the text file byte offsets and I expect(hope) C_Simp example files aren't going to be updated frequently.

MikaelMayer (approved):

PR #1053, opened 2026-04-25
Fix bool_*_interp reductions via concreteEval + native_decide

julesmt:
The four `bool_*_interp` proofs in `Tautologies.lean` were failing because Lean couldn’t reduce expressions involving `ceval`, which comes from a `HashMap` lookup inside `Factory`. Common tactics don’t look inside that structure, so they got stuck.
Instead of working directly with `ceval`, first prove the evaluation result using the concrete field `F["..."].concreteEval`, written with `Option.bind` so everything stays total. Then use `native_decide` to evaluate the lookup and boolean cases. After that, rewrite using `h_ceval_eq` to replace `F["..."].concreteEval` with `some ceval`. At this point, Lean can reduce `(some ceval).bind ...` normally, and `simpa` completes the proof.
This same approach works for `And`, `Or`, `Implies`, and `Not`.


StrataTest/DL/Lambda/Denote/Tautologies.lean
```diff
@@ -234,9 +231,15 @@ private theorem bool_and_interp (I : Interp F) :
...
-    -- Blocked: Factory HashMap lookup doesn't reduce. TODO.
-    sorry
+    have h_concrete :
+        ∀ (b1 b2 : Bool),
+          (F["Bool.And"]'bool_and_mem).concreteEval.bind
+            (fun f => f () [.const () (.boolConst b1), .const () (.boolConst b2)])
+          = some (.const () (.boolConst (b1 && b2))) := by
+      intro b1 b2; cases b1 <;> cases b2 <;> native_decide
+    have := h_concrete p q
+    rw [h_ceval_eq] at this
```
joscoh: We should not use auto-generated names; it is better to give an explicit name to the `have` block.

joscoh (commented):
Great, thank you for this! Next question is to determine if this can be done without `native_decide` :)

julesmt:
@joscoh Thanks, replaced `anonymous` this with `h_inst_concrete`

julesmt:
And I have no idea if it can be easily done without `native_decide` :)

StrataTest/DL/Lambda/Denote/Tautologies.lean
```diff
@@ -234,9 +231,15 @@ private theorem bool_and_interp (I : Interp F) :
   funext p q
   have h_eval : ceval () [.const () (.boolConst p), .const () (.boolConst q)]
       = some (.const () (.boolConst (p && q))) := by
-    -- ceval comes from F["Bool.And"].concreteEval. Use simp to reduce the HashMap lookup.
-    -- Blocked: Factory HashMap lookup doesn't reduce. TODO.
-    sorry
+    have h_concrete :
+        ∀ (b1 b2 : Bool),
+          (F["Bool.And"]'bool_and_mem).concreteEval.bind
+            (fun f => f () [.const () (.boolConst b1), .const () (.boolConst b2)])
```
joscoh: Not blocking, but `.boolConst () b1` should work directly instead of `.const () (.boolConst b1)`

joscoh (dismissed):

StrataTest/DL/Lambda/Denote/Tautologies.lean
julesmt: ok I will test that

julesmt: @joscoh done !

joscoh (approved):

aqjune-aws (approved):

aqjune-aws:
Thanks!

PR #1054, opened 2026-04-27
Add basic support for `del` key word in Python->Laurel

thanhnguyen-aws:
This PR adds `del` statement support to handle `list[index/slice]` and `dict[key]` when `list` and `dict` are variables, which also supports:

1. **Negative indices**: `del xs[-1]` now correctly converts negative indices to positive before removal, following the same pattern as `List_get`.

2. **Slice deletion**: `del xs[start:stop]` removes a range of elements, keeping elements before `start` and from `stop` onwards. Supports both positive and negative slice bounds.

### Changes

**Runtime (`PythonRuntimeLaurelPart.lean`)**:
- Split `List_remove` into `List_remove_non_neg` (requires `i >= 0`) and `List_remove` (handles negative-to-positive conversion)
- Added `List_remove_slice` for removing a range of elements
- Added `Any_remove_slice` with exception handling and slice unpacking

**Translator (`PythonToLaurel.lean`)**:
- `translateDel` now matches on `.Slice` to call `Any_remove_slice`, falling back to `Any_remove` for single-index deletion

**Tests**:
- `test_del_list_negative_index.py`: `del xs[-1]` on `[1, 2, 3, 4]`
- `test_del_list_slice.py`: `del xs[1:3]` on `[1, 2, 3, 4, 5]`

All existing del tests continue to pass.


thanhnguyen-aws:
@keyboardDrummer-bot Please add similar support for list's negative index and slice.

thanhnguyen-aws:
@keyboardDrummer-bot Please merge PR https://github.com/strata-org/Strata/pull/1055 to this PR. Don't create a new one.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1417,6 +1417,20 @@ def withExceptionChecks (ctx : TranslationContext)
   let exceptionCheck := rhs_exprs.flatMap $ getExceptionAssertions ctx
   (newctx, exceptionCheck ++ stmts)
 
+def translateDel (ctx : TranslationContext) (md: MetaData) (e:  Python.expr SourceRange) : Except TranslationError StmtExprMd := do
```
tautschnig: Nit: `translateDel` only handles `Subscript` on a `Name` — `del obj.attr.items[0]` or `del (a if cond else b)[key]` would fail. The error message is clear, so this is fine for a first implementation, but worth noting as a known limitation.

tautschnig (dismissed):

aqjune-aws (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Let's address this later.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -415,6 +414,26 @@ function List_slice (l : ListAny, start : int, stop: int) : ListAny
...
+  if i >= 0 then List_remove_non_neg(l, i)
+  else List_remove_non_neg(l, List_len(l) + i)
+};
+
+function List_remove_slice(l: ListAny, start: int, stop: int) : ListAny
+{
+  List_extend(
+    List_take(l, if start >= 0 then int_min(start, List_len(l)) else int_max(List_len(l) + start, 0)),
+    List_drop(l, if stop >= 0 then int_min(stop, List_len(l)) else int_max(List_len(l) + stop, 0)))
+};
```
tautschnig, line 437: **Bug: `start > stop` after normalization produces a corrupted list.**

I verified against CPython: `xs = [1,2,3,4,5]; del xs[3:1]; xs` → `[1, 2, 3, 4, 5]` (unchanged — reverse slice is a no-op). The current implementation for `l=[1,2,3,4,5]`, `start=3`, `stop=1`:
- clamped `start'` = 3, clamped `stop'` = 1
- `List_take(l, 3)` = `[1, 2, 3]`
- `List_drop(l, 1)` = `[2, 3, 4, 5]`
- `List_extend(...)` = `[1, 2, 3, 2, 3, 4, 5]` ← wrong (+2 elements)

`List_slice_non_neg` at line 406 already has the equivalent guard and is the shape to copy:

```suggestion
function List_remove_slice(l: ListAny, start: int, stop: int) : ListAny
{
  let start_c: int := if start >= 0 then int_min(start, List_len(l)) else int_max(List_len(l) + start, 0);
  let stop_c: int := if stop >= 0 then int_min(stop, List_len(l)) else int_max(List_len(l) + stop, 0);
  if start_c >= stop_c then l
  else List_extend(List_take(l, start_c), List_drop(l, stop_c))
};
```

(Apologies if Laurel's `let` surface syntax differs — if the above isn't parseable, the equivalent `(start_c, stop_c, …)` inline is fine. The point is the `if start_c >= stop_c then l else …` guard and binding the two clamped values once to avoid the repeated `if start >= 0 then …` sub-expressions.)

Test to add: `del xs[3:1]` on `[1,2,3,4,5]` — should pass `assert xs == [1,2,3,4,5]`. With the current code this assertion fails.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -543,6 +569,35 @@ function Any_sets! (indices: ListAny, dictOrList: Any, val: Any): Any
...
+  else if Any..isexception(index) then index
+  else if !(Any..isfrom_DictStrAny(dictOrList) && Any..isfrom_str(index)) && !(Any..isfrom_ListAny(dictOrList) && Any..isfrom_int(index)) then
+    exception (TypeError("Invalid subscription type"))
+  else if Any..isfrom_DictStrAny(dictOrList) && Any..isfrom_str(index) && DictStrAny_contains(Any..as_Dict!(dictOrList), Any..as_string!(index)) then
+    from_DictStrAny(DictStrAny_remove(Any..as_Dict!(dictOrList), Any..as_string!(index)))
+  else if Any..isfrom_ListAny(dictOrList) && Any..isfrom_int(index) && Any..as_int!(index) >= - List_len(Any..as_ListAny!(dictOrList)) && Any..as_int!(index) < List_len(Any..as_ListAny!(dictOrList)) then
+    from_ListAny(List_remove(Any..as_ListAny!(dictOrList), Any..as_int!(index)))
+  else
+    exception (IndexError("Invalid subscription"))
+};
```
tautschnig, line 589: **Wrong exception type for dict missing-key deletion.**

CPython: `del d["missing"]` raises `KeyError`, not `IndexError`. I verified: `KeyError: 'missing'`.

The current code's final `else` covers two semantically different failure modes with the same exception:
- List index out of range → `IndexError` ✓
- Dict key not present → `IndexError` ✗ (should be `KeyError`)

The third else-if at line 578 guards with `DictStrAny_contains(...)`, so the missing-key path falls through to the generic `IndexError` else. Split the arm:

```suggestion
function Any_remove (dictOrList: Any, index: Any): Any
{
  if Any..isexception(dictOrList) then dictOrList
  else if Any..isexception(index) then index
  else if !(Any..isfrom_DictStrAny(dictOrList) && Any..isfrom_str(index)) && !(Any..isfrom_ListAny(dictOrList) && Any..isfrom_int(index)) then
    exception (TypeError("Invalid subscript type"))
  else if Any..isfrom_DictStrAny(dictOrList) && Any..isfrom_str(index) then
    if DictStrAny_contains(Any..as_Dict!(dictOrList), Any..as_string!(index)) then
      from_DictStrAny(DictStrAny_remove(Any..as_Dict!(dictOrList), Any..as_string!(index)))
    else
      exception (KeyError(Any..as_string!(index)))
  else if Any..as_int!(index) >= - List_len(Any..as_ListAny!(dictOrList)) && Any..as_int!(index) < List_len(Any..as_ListAny!(dictOrList)) then
    from_ListAny(List_remove(Any..as_ListAny!(dictOrList), Any..as_int!(index)))
  else
    exception (IndexError("list assignment index out of range"))
};
```

The CPython error string for out-of-range `del` is `"list assignment index out of range"`, and the `KeyError` carries the missing key as its payload — matching those in the emitted exceptions makes the runtime observationally closer to CPython and aids downstream tooling that pattern-matches on error text.

Test to add:
- `del d["missing"]` (expect `KeyError`)
- `del xs[10]` on a 3-element list (expect `IndexError`)

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1626,6 +1626,20 @@ def withExceptionChecks (ctx : TranslationContext)
...
+    match slice with
+    | .Slice _ start stop step =>
+      let index ← translateSlice ctx start.val stop.val step.val
+      let rhs := mkStmtExprMd $ .StaticCall "Any_remove_slice" [freeVarExpr n.val, index]
+      return mkStmtExprMdWithLoc (.Assign [freeVarMd n.val] rhs) source
+    | _ =>
+      let slice ← translateExpr ctx slice
+      let rhs := mkStmtExprMd $ .StaticCall "Any_remove" [freeVarExpr n.val, slice]
+      return mkStmtExprMdWithLoc (.Assign [freeVarMd n.val] rhs) source
+  | _ => throw (.unsupportedConstruct "Only support del statement for list[index] and dict[key] where list and dict are variables, unsupported: " (toString (repr e)))
```
tautschnig, line 1641: **Exception propagation: two gaps.**

**(a) `translateDel` produces `.Assign` statements but doesn't wrap with `withExceptionChecks`.** Every other assignment-emitting case in `translateStmt` (see lines 1653, 1666, 1691) wraps its output so that `getExceptionAssertions` can insert `assert !Any..isexception(...)` before the assignment if the RHS can produce one. Without that wrap:

```python
try:
    del xs[10]
except IndexError:
    pass
assert xs == [1, 2, 3]  # CPython: passes; our translation: fails (xs is now the exception value)
```

The Delete case at line 2034 in `translateStmt` should route through `withExceptionChecks ctx (ctx, delStmts)` like the `.Assign` case at line 1653.

**(b) `Any_remove` and `Any_remove_slice` must be in `AnyMaybeExceptionList`** (`PythonRuntimeLaurelPart.lean:1153`) — they return exception values via the `isfrom_…` type-check and `DictStrAny_contains` failure arms, but without the list entry `getMaybeExceptionExprs` at `PythonToLaurel.lean:1565` won't recognise their call sites. (a) by itself is not enough; both are needed for exception checks to fire.

Suggested changes inline below; feel free to reject if I've misread the exception-propagation model:

```suggestion
def translateDel (ctx : TranslationContext) (source: Option FileRange) (e:  Python.expr SourceRange) : Except TranslationError StmtExprMd := do
  match e with
  | .Subscript _ (.Name _ n _) slice _ =>
    match slice with
    | .Slice _ start stop step =>
      let index ← translateSlice ctx start.val stop.val step.val
      let rhs := mkStmtExprMd $ .StaticCall "Any_remove_slice" [freeVarExpr n.val, index]
      return mkStmtExprMdWithLoc (.Assign [freeVarMd n.val] rhs) source
    | _ =>
      let slice ← translateExpr ctx slice
      let rhs := mkStmtExprMd $ .StaticCall "Any_remove" [freeVarExpr n.val, slice]
      return mkStmtExprMdWithLoc (.Assign [freeVarMd n.val] rhs) source
  | _ => throw (.unsupportedConstruct "Only support del statement for list[index] and dict[key] where list and dict are variables, unsupported: " (toString (repr e)))
```

…and in `translateStmt` at line 2034 (the `.Delete _ targets` arm):

```lean
  | .Delete _ targets =>
      let delStmts ← targets.val.toList.mapM $ translateDel ctx md
      return withExceptionChecks ctx (ctx, delStmts)
```

…and in `PythonRuntimeLaurelPart.lean` at line 1153:

```lean
public def AnyMaybeExceptionList := ["Any_get!", "Any_set!", "Any_sets!", "Any_remove", "Any_remove_slice", "PNeg", "PBitNot", ...]
```

(Or, if you intend to keep the per-call exception as lazy propagation via subsequent `Any_get!`, document that choice with a comment here — but note that the `try: del xs[bad]; except IndexError: pass` case will not match CPython behaviour either way; the lazy model changes *when* the exception is observed, not *whether*.)

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
tautschnig, line 437: **Proof coverage — four cheap theorems that would lock in Python semantics.**

All four new runtime functions have closed-form element/length specifications. These can live in a new `StrataTest/Languages/Python/PythonRuntimeLaurelPartProofs.lean` or, more in keeping with the rest of the tree, in the existing Python runtime test file as `#guard`-style snapshots. `rfl` should handle most of them once the surface syntax is lowered.

1. **`List_remove_non_neg_length`**:
   ```
   ∀ l i, 0 ≤ i < List_len(l) → List_len(List_remove_non_neg(l, i)) = List_len(l) - 1
   ```

2. **`List_remove_non_neg_get`** (the element-preservation property — this is the one that catches accidental off-by-one in `i+1`):
   ```
   ∀ l i j, 0 ≤ i < List_len(l) → 0 ≤ j < List_len(l) - 1 →
     List_get(List_remove_non_neg(l, i), j) =
       if j < i then List_get(l, j) else List_get(l, j + 1)
   ```

3. **`List_remove_slice_length`** (the one that would catch concern (1) above as a proof failure):
   ```
   ∀ l start stop,
     let start_c = clamp(start, List_len(l))
     let stop_c  = clamp(stop,  List_len(l))
     List_len(List_remove_slice(l, start, stop)) =
       if start_c >= stop_c then List_len(l)
       else List_len(l) - (stop_c - start_c)
   ```
   With the current implementation, for `l=[1,2,3,4,5], start=3, stop=1`, LHS = 7 and RHS = 5 — the proof doesn't go through, forcing a rewrite.

4. **`List_remove_roundtrip`** (sanity check tying `List_remove_non_neg` and `List_remove_slice` together):
   ```
   ∀ l i, 0 ≤ i < List_len(l) →
     List_remove_non_neg(l, i) = List_remove_slice(l, i, i + 1)
   ```

Even if only (3) lands, that's the regression-proof version of the concern above, and is the highest-value theorem in the set.

Similarly worth adding for `DictStrAny_remove`:

5. **`DictStrAny_remove_contains_false`**:
   ```
   ∀ d k, DictStrAny_contains(DictStrAny_remove(d, k), k) = false
   ```

6. **`DictStrAny_remove_other`** (other keys preserved):
   ```
   ∀ d k k' v, k ≠ k' →
     DictStrAny_get(DictStrAny_remove(d, k), k') = DictStrAny_get(d, k')
     (when d contains k')
   ```

StrataTest/Languages/Python/tests/test_del_list_slice.py
```diff
@@ -0,0 +1,7 @@
+def test():
+    xs = [1, 2, 3, 4, 5]
+    del xs[1:3]
+    assert xs[0] == 1, "first unchanged"
+    assert xs[1] == 4, "fourth shifted"
+    assert xs[2] == 5, "fifth shifted"
+test()
```
tautschnig: The single happy-path slice fixture covers `[1:3]` on a list of 5. Four more fixtures — each small — would cover the edge cases this PR's new slice-deletion touches and that Python has non-obvious behaviour on:

```python
# test_del_list_reverse_slice.py — reverse slice is a no-op in CPython
def test():
    xs = [1, 2, 3, 4, 5]
    del xs[3:1]
    assert xs[0] == 1, "first unchanged"
    assert xs[3] == 4, "fourth unchanged"  # would fail today (xs has 7 elements)
    assert xs[4] == 5, "fifth unchanged"
test()

# test_del_list_slice_to_end.py — omit stop
def test():
    xs = [1, 2, 3, 4, 5]
    del xs[2:]
    assert xs[0] == 1 and xs[1] == 2, "head preserved"
test()

# test_del_list_slice_from_start.py — omit start
def test():
    xs = [1, 2, 3, 4, 5]
    del xs[:2]
    assert xs[0] == 3 and xs[2] == 5, "tail preserved"
test()

# test_del_list_slice_all.py — del xs[:] empties the list
def test():
    xs = [1, 2, 3]
    del xs[:]
    assert len(xs) == 0, "emptied"
test()
```

Similarly for `tests/test_del_list_negative_index.py`: consider adding `del xs[-4]` on `[1,2,3]` (out of range on the negative side — should be `IndexError`) and `del xs[-1]` on a single-element list.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
thanhnguyen-aws: I added the proofs for `List`'s theorems in Laurel and add the `guard` test for them. The `Dict` theorems cannot be proved because it requires that the `Dict` is constructed by `Dict_insert`. The runtime is slow, so it is just a temporary approach. We need to have a Lean backend so that the theorems can be proved efficiently.

PR #1058, opened 2026-04-27
Minimize DDM imports via lake shake

joehendrix:
## Summary

Run `lake shake` on DDM modules to minimize their import graph. This is the
DDM-only subset of #1020, split out to reduce merge conflict surface.

## Changes

- **DDM import minimization**: Applied `lake shake --fix` across 30
  `Strata/DDM/` files, removing unnecessary imports and demoting `public
  import` to `import` where re-export was not needed by downstream consumers.

- **`HashCommands.lean` visibility fixes**: Reorganized `meta section` /
  `public section` boundaries and removed unnecessary `private` qualifiers
  to maintain correct elaboration-time visibility after import changes.

- **Downstream import fixes**: 13 non-DDM files that lost transitive access
  to DDM symbols received explicit imports (`Strata.DDM.Format`,
  `Strata.DDM.Util.Ion`, `Lean.Parser.Types`, `Lean.Parser.Extension`).
  Two unnecessary `public import Lean.Expr` / `public import Lean.ToExpr`
  were removed from `Translate.lean`.

### DIFF
--- a/Strata/DDM/Integration/Lean/Gen.lean
+++ b/Strata/DDM/Integration/Lean/Gen.lean
@@ -6,20 +6,20 @@
 module
 
 public meta import Lean.Elab.Command
-public meta import Strata.DDM.AST
--- `public import` provides object-level access for generated code at call sites.
--- `public meta import` provides meta-level access for #strata_gen elaboration.
-public import      Strata.DDM.BuiltinDialects.Init  -- Generated code uses Init types
-public import      Strata.DDM.HNF  -- Generated ofAst uses ExprF.hnf
-public meta import Strata.DDM.BuiltinDialects.Init
+import Lean.Parser.Command
+
+public import      Strata.DDM.BuiltinDialects.Init -- shake: keep (Generated code uses Init types)
+public meta import Strata.DDM.BuiltinDialects.Init -- shake: keep
 meta import        Strata.DDM.BuiltinDialects.StrataDDL
-public meta import Strata.DDM.Integration.Categories
+public import      Strata.DDM.HNF  -- shake: keep (Generated ofAst uses ExprF.hnf)
+public meta import Strata.DDM.Integration.Categories --shake: keep (shake bug)
 public meta import Strata.DDM.Integration.Lean.Env
-public meta import Strata.DDM.Integration.Lean.GenTrace  -- trace option
-public import      Strata.DDM.Integration.Lean.OfAstM  -- Generated ofAst combinators
-public meta import Strata.DDM.Integration.Lean.OfAstM
-public meta import Strata.DDM.Util.Graph.Tarjan
-meta import        Strata.Util.DecideProp
+public meta import Strata.DDM.Integration.Lean.GenTrace  -- shake: keep (trace option)
+public import      Strata.DDM.Integration.Lean.OfAstM -- shake: keep (Generated ofAst combinators)
+public meta import Strata.DDM.Integration.Lean.OfAstM -- shake: keep
+public meta import Strata.DDM.Util.Graph.Tarjan -- shake: keep (shake bug)
+import Strata.DDM.Util.Graph.Tarjan
+meta import Strata.Util.DecideProp
 
 /-!
 Implements the `#strata_gen` command, which reads a dialect definition and
--- a/Strata/DDM/Integration/Lean/Gen.lean
+++ b/Strata/DDM/Integration/Lean/Gen.lean
@@ -12,7 +12,7 @@ public import      Strata.DDM.BuiltinDialects.Init -- shake: keep (Generated cod
 public meta import Strata.DDM.BuiltinDialects.Init -- shake: keep
 meta import        Strata.DDM.BuiltinDialects.StrataDDL
 public import      Strata.DDM.HNF  -- shake: keep (Generated ofAst uses ExprF.hnf)
-public meta import Strata.DDM.Integration.Categories --shake: keep (shake bug)
+public meta import Strata.DDM.Integration.Categories -- shake: keep (shake bug)
 public meta import Strata.DDM.Integration.Lean.Env
 public meta import Strata.DDM.Integration.Lean.GenTrace  -- shake: keep (trace option)
 public import      Strata.DDM.Integration.Lean.OfAstM -- shake: keep (Generated ofAst combinators)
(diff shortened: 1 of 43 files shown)


Strata/DDM/Integration/Lean/Gen.lean
```diff
@@ -6,20 +6,20 @@
...
-public import      Strata.DDM.HNF  -- Generated ofAst uses ExprF.hnf
-public meta import Strata.DDM.BuiltinDialects.Init
+import Lean.Parser.Command
+
+public import      Strata.DDM.BuiltinDialects.Init -- shake: keep (Generated code uses Init types)
+public meta import Strata.DDM.BuiltinDialects.Init -- shake: keep
 meta import        Strata.DDM.BuiltinDialects.StrataDDL
-public meta import Strata.DDM.Integration.Categories
+public import      Strata.DDM.HNF  -- shake: keep (Generated ofAst uses ExprF.hnf)
+public meta import Strata.DDM.Integration.Categories --shake: keep (shake bug)
```
robin-aws: Are there any existing issues tracking the shake bug(s)?

Strata/DDM/Integration/Lean/Gen.lean
```diff
@@ -6,20 +6,20 @@
...
 public meta import Lean.Elab.Command
-public meta import Strata.DDM.AST
--- `public import` provides object-level access for generated code at call sites.
--- `public meta import` provides meta-level access for #strata_gen elaboration.
-public import      Strata.DDM.BuiltinDialects.Init  -- Generated code uses Init types
-public import      Strata.DDM.HNF  -- Generated ofAst uses ExprF.hnf
-public meta import Strata.DDM.BuiltinDialects.Init
+import Lean.Parser.Command
+
+public import      Strata.DDM.BuiltinDialects.Init -- shake: keep (Generated code uses Init types)
```
robin-aws, line 11: Is the fact that you're having to add these exceptions an indication that `lake shake` is not working correctly?

robin-aws (dismissed):

Strata/DDM/Integration/Lean/HashCommands.lean
```diff
@@ -49,21 +52,21 @@ private instance : HasInputContext CoreM where
     }
   getFileName := return (← read).fileName
 
-private def mkScopedName {m} [Monad m] [MonadError m] [MonadEnv m] [MonadResolveName m] (name : Name) : m Name := do
+def mkScopedName {m} [Monad m] [MonadError m] [MonadEnv m] [MonadResolveName m] (name : Name) : m Name := do
   let scope ← getCurrNamespace
```
MikaelMayer, line 56: nit: `mkScopedName`, `offsetPos`, `offsetSourceRange`, `offsetMessage`, and `addDefn` were all `private` before and are now namespace-visible (inside `meta section` but no longer `private`). This was needed because the old `public meta section` made them implicitly visible, and now the section restructuring requires them to be callable from the `public section` below. But since they're implementation details, could they stay `private` and still work? If `declareDialect` (in the `public section`) calls `mkScopedName` and `addDefn`, and both sections are inside the same `meta section`, `private` should still be accessible within the same file. Worth checking — if it works, it would keep the API surface minimal.

Strata/DDM/Integration/Lean/Gen.lean
MikaelMayer: nit: Inconsistent comment spacing — `--shake: keep` (no space after `--`) vs `-- shake: keep` elsewhere.

Strata/Util/IO.lean
```diff
@@ -9,7 +9,7 @@ import Strata.DDM.Elab
 import Strata.DDM.Ion
 import Strata.DDM.BuiltinDialects
 public import Strata.DDM.Elab.LoadedDialects
-import Strata.DDM.Util.Ion.Serialize
+import Strata.DDM.Util.Ion
```
MikaelMayer, line 12: This broadens the import from `Strata.DDM.Util.Ion.Serialize` to `Strata.DDM.Util.Ion` (which re-exports `AST`, `Deserialize`, `Serialize`, and `SymbolTable`). Was the narrower `Serialize` import insufficient after the other changes, or does `IO.lean` now need symbols from the other `Ion` submodules that it previously got transitively?

Strata/DDM/Integration/Lean/Gen.lean
joehendrix: Shake has a known limitation that it doesn't detect usage generated by custom elaborators.  So files that provide custom elaborators (like Gen.lean and HashCommands.lean will typically need manual annotations.

To be honest, I am not quite sure of the meta/normal distinction when it comes to public imports, so I am a bit surprised that we need both for Init and some other modules.

Strata/DDM/Integration/Lean/Gen.lean
joehendrix: No, I'll try a minimization step to understand why it didn't see this as needed.

MikaelMayer (commented):
Clean import minimization PR. The `lake shake` results look correct, and the visibility adjustments in `HashCommands.lean` and `TaggedRegions.lean` are well-reasoned. A few minor items below.

Strata/DDM/Integration/Lean/HashCommands.lean
joehendrix: They are still private, and their visibility has not changed.

I moved the start of the public section to after these definitions so their default visibility changed from public to private, and thus the private annotation is now redundant.

Strata/Util/IO.lean
joehendrix: This file depends on `isIonFile` which is in `Strata.DDM.Util.Ion`.  Previously this file had `Strata.DDM.Util.Ion` imported indirectly via one of the other imports (and the Serialize import wasn't redundant).

That's no longer true after `shake` reduced public imports, and so it must be explicitly imported.

MikaelMayer (dismissed):

Strata/DDM/Integration/Lean/Gen.lean
joehendrix:
```suggestion
public meta import Strata.DDM.Integration.Categories -- shake: keep (shake bug)
```

robin-aws (approved):

MikaelMayer (approved):

PR #1060, opened 2026-04-28
Add Strata internal benchmarks Github Action

aqjune-aws:
https://github.com/strata-org/Strata/pull/1060 adds a Codebuild setup connected to Github Action for Strata internal benchmark setup. 

Only authorized people can access to the S3 bucket and download it (hence external users outside Amazon cannot get any info about the internal benchmarks). The Github Action log only shows that the Codebuild is running or finished.

Once successful, the CI check results will look like that of https://github.com/aqjune-aws/Strata/pull/3 . This pull request 1060 to strata-org fails however for now, and according to Claude, GitHub Actions doesn't grant OpenID Connect tokens to workflows triggered by PRs from forks, and for this reason github/workflow/ci.yml of the main branch to be updated first. For this reason, this PR 1060 only enables the Codebuild check on the commits of the main branch. Once the check becomes successful on main, I will send another PR that enables the Codebuild for all PRs.


tautschnig (changes requested):
Comments out-of-band.

.github/workflows/ci.yml
```diff
@@ -321,3 +321,48 @@ jobs:
...
+  strata-benchmarks:
+    name: Run internal benchmarks of Strata
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
```
tautschnig:
```suggestion
        uses: aws-actions/configure-aws-credentials@v6
```
to use the latest release.

.github/workflows/ci.yml
```diff
@@ -321,3 +321,48 @@ jobs:
...
+          BUILD_ID=$(aws codebuild start-build \
+            --project-name strata-benchmarks \
+            --source-type-override GITHUB \
+            --source-location-override https://github.com/strata-org/Strata.git \
+            --query 'build.id' --output text \
+            --region us-east-2)
+          echo "Build started: $BUILD_ID"
+
+          LOG_KEY="logs/${BUILD_ID}.log"
+          echo "[View build log in S3](https://s3.console.aws.amazon.com/s3/object/strata-benchmarks-logs-${{secrets.AWS_BENCHMARK_ACCOUNT}}-us-east-2-an?prefix=${LOG_KEY})" >> $GITHUB_STEP_SUMMARY
```
tautschnig: Let's please rename the bucket for else there is no point in storing the account id as a secret -- it would be exposed via the public URL in the summary.

.github/workflows/ci.yml
```diff
@@ -321,3 +321,48 @@ jobs:
...
+            STATUS=$(aws codebuild batch-get-builds \
+              --ids "$BUILD_ID" \
+              --query 'builds[0].buildStatus' --output text \
+              --region us-east-2)
+            echo "Current status: [$STATUS]"
+            case "$STATUS" in
+              SUCCEEDED) break;;
+              FAILED|FAULT|TIMED_OUT|STOPPED) echo "Build failed: $STATUS" ; break ;;
+              IN_PROGRESS) sleep 30 ;;
+              *) echo "Unexpected status: $STATUS"; sleep 10 ; break ;;
```
tautschnig, line 363: Shouldn't we abort when we find an unexpected status?

.github/workflows/ci.yml
```diff
@@ -321,3 +321,48 @@ jobs:
...
+            esac
+          done
+
+          echo "View build log in S3: https://s3.console.aws.amazon.com/s3/object/strata-benchmarks-logs-${{secrets.AWS_BENCHMARK_ACCOUNT}}-us-east-2-an?prefix=${LOG_KEY}"
```
tautschnig: As above, this would expose the AWS account id.

.github/workflows/ci.yml
```diff
@@ -321,3 +321,48 @@ jobs:
...
+          [ "$STATUS" = "SUCCEEDED" ] && exit 0 || exit 1
```
tautschnig:
```suggestion
          test "$STATUS" = "SUCCEEDED"
```

.github/workflows/ci.yml
aqjune-aws: In the summary and echo below, these will appear as `***` and I was planning to let them know how to replace '***' with an account id. But it is a good idea to make this address agnostic to the account id, because it will be simpler. :) Let met fix this...

.github/workflows/ci.yml
aqjune-aws: Good idea, fixed

.github/workflows/ci.yml
aqjune-aws: Fixed

.github/workflows/ci.yml
aqjune-aws: The `test` check after this loop will guarantee non-zero exit. It helps print the S3 log url.

tautschnig: Well, but we might never exit this loop.

aqjune-aws: `break` will break the while look :) Something like this:
```
while true; do
  case "A" in
    A) echo 'h'; break;;
  esac
done
```
will print `h` once.

aqjune-aws:
The updated ci.yml is being tested in https://github.com/aqjune-aws/Strata/pull/4.

aqjune-aws:
The internal benchmarks result of CI run on my local repo seems to be consistent with the known numbers. Posted in a separate thread

.github/workflows/ci.yml
```diff
@@ -321,3 +321,49 @@ jobs:
     permissions:
       contents: read
     uses: ./.github/workflows/cbmc.yml
+
+  strata-benchmarks:
+    name: Run internal benchmarks of Strata
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
```
robin-aws, line 328: In the next PR, could we add logging of the event type as one of the first steps in the whole workflow? The GitHub UX makes it very difficult to determine it for debugging in my experience.

(This is orthogonal to this PR, but the more we have event-specific CI the more important this is)

robin-aws (approved):

aqjune-aws:
Superceded by https://github.com/strata-org/Strata/pull/1070

PR #1063, opened 2026-04-28
Use fresh type variable instead of Error as dummy type for failed type inference

tautschnig:
When Laurel→Core type translation fails, return a fresh type variable (ftvar) instead of the concrete Error type as a placeholder.

Error is a real type (Python exceptions), and using it as a dummy caused type unification failures ("Impossible to unify Error with X") when the type checker tried to unify it with the actual expected type. A fresh type variable is the correct "unknown type" placeholder — the unifier can assign it any concrete type.

Also handle Hole (nondet) in Assign statements by emitting havoc instead of trying to translate the hole as an expression.

Testing: Test that a fresh type variable (ftvar) unifies with concrete types during Core type checking, while the concrete Error type does not. This documents the invariant that throwTypeDiagnostic relies on: ftvar placeholders are transparent to the unifier, Error is not.


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -103,7 +108,16 @@ def translateType (ty : HighTypeMd) : TranslateM LMonoTy := do
...
+  | .Unknown => do
+    -- Unknown types get a fresh type variable. The Core type checker will
+    -- unify it with the expected type from context. We emit a diagnostic
+    -- but do NOT set coreProgramHasSuperfluousErrors — the program is still
+    -- valid because the type checker can resolve the ftvar.
+    emitDiagnostic ((astNodeToCoreMd ty).toDiagnostic "could not infer type")
+    let s ← get
+    let id := s.nextId
+    set { s with nextId := id + 1 }
+    return .ftvar s!"$__ty_unused_{id}"
```
MikaelMayer: This block (emit diagnostic, allocate fresh ID, return ftvar) is nearly identical to `throwTypeDiagnostic` above, differing only in whether `coreProgramHasSuperfluousErrors` is set. Consider extracting a shared helper, e.g.:

```lean
private def freshTypePlaceholder : TranslateM LMonoTy := do
  let id ← freshId
  return .ftvar s!"$__ty_unused_{id}"
```

Then `throwTypeDiagnostic` becomes `emitDiagnostic ..; modify ..; freshTypePlaceholder` and the `.Unknown` case becomes `emitDiagnostic ..; freshTypePlaceholder`. This also eliminates the magic string duplication across three call sites.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -103,7 +108,16 @@ def translateType (ty : HighTypeMd) : TranslateM LMonoTy := do
...
   | .TReal => return LMonoTy.real
-  | .Unknown => throwTypeDiagnostic ty "could not infer type"
+  | .Unknown => do
+    -- Unknown types get a fresh type variable. The Core type checker will
+    -- unify it with the expected type from context. We emit a diagnostic
+    -- but do NOT set coreProgramHasSuperfluousErrors — the program is still
+    -- valid because the type checker can resolve the ftvar.
+    emitDiagnostic ((astNodeToCoreMd ty).toDiagnostic "could not infer type")
+    let s ← get
+    let id := s.nextId
```
MikaelMayer: Not setting `coreProgramHasSuperfluousErrors` for `.Unknown` means the Core program is kept and processed even when type inference failed. If the ftvar ends up in a position where the unifier has no concrete type to assign (e.g., an expression used only for side effects), does the type checker produce a clear error, or does the unresolved ftvar silently propagate to SMT encoding? Worth documenting what happens in the "ftvar is never unified" case — either here or in the PR description.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -452,6 +469,13 @@ def translateStmt (stmt : StmtExprMd)
                 | .Identifier name => some (Core.Statement.havoc ⟨name.text, ()⟩ md)
                 | _ => none
               return (havocStmts)
+          | .Hole _ _ =>
+              -- Nondet hole: havoc all target variables
+              let havocStmts := targets.filterMap fun t =>
+                match t.val with
+                | .Identifier name => some (Core.Statement.havoc ⟨name.text, ()⟩ md)
+                | _ => none
+              return (havocStmts)
```
MikaelMayer, line 477: The `.InstanceCall` and `.Hole` branches are identical. Combine them:
```lean
          | .InstanceCall .. | .Hole .. =>
              let havocStmts := targets.filterMap fun t =>
                match t.val with
                | .Identifier name => some (Core.Statement.havoc ⟨name.text, ()⟩ md)
                | _ => none
              return havocStmts
```
Same applies to the single-target case above (lines 444–448).

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -103,7 +108,16 @@ def translateType (ty : HighTypeMd) : TranslateM LMonoTy := do
       return .tcons "Composite" []
   | .TCore s => return .tcons s []
   | .TReal => return LMonoTy.real
-  | .Unknown => throwTypeDiagnostic ty "could not infer type"
+  | .Unknown => do
```
shigoel, line 119: This change means that all Laurel `Unknown` types are delegated to Core's Hindley-Milner type inference. Is that the right semantics for all front-ends that flow through Laurel?

StrataTest/Languages/Laurel/ThrowTypeDiagnosticTest.lean
```diff
@@ -0,0 +1,75 @@
...
+2. A Core program using `tcons "Error"` as a placeholder FAILS type checking
+   (Error does not unify with int), documenting the bug that the ftvar fix solves.
+-/
+
+open Strata
+
+/-- Build a minimal Core program:
+    `procedure main() { var x : <placeholderTy> := 0; var y : int := x; }`
+    The placeholder type of `x` must unify with `int` when `x` is assigned to `y`. -/
+private def mkTestProgram (placeholderTy : Lambda.LMonoTy) : Core.Program :=
```
shigoel, line 32: Can this be done using the DDM syntax?

MikaelMayer (commented):
The core idea — using ftvar instead of a concrete Error type as a placeholder for failed type inference — is sound and well-motivated. The test clearly demonstrates the unification difference. A few items to address around code duplication and a design question about the `.Unknown` semantics.
+ test format in DDM syntax

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
tautschnig: Done in [f3d28ce](https://github.com/strata-org/Strata/pull/1063/commits/f3d28ce46bee0bc5e5b05caccd3984a63c1403a5).

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
tautschnig: Added an extensive comment in [f3d28ce](https://github.com/strata-org/Strata/pull/1063/commits/f3d28ce46bee0bc5e5b05caccd3984a63c1403a5).

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
tautschnig: Done in [f3d28ce](https://github.com/strata-org/Strata/pull/1063/commits/f3d28ce46bee0bc5e5b05caccd3984a63c1403a5).

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
tautschnig: I believe so (and added a comment in [f3d28ce](https://github.com/strata-org/Strata/pull/1063/commits/f3d28ce46bee0bc5e5b05caccd3984a63c1403a5)), but perhaps others know better?

StrataTest/Languages/Laurel/ThrowTypeDiagnosticTest.lean
tautschnig: DDM parsing would infer types, defeating the purpose of testing type-checker behavior on specific types. Added a comment in the source stating the same (in [f3d28ce](https://github.com/strata-org/Strata/pull/1063/commits/f3d28ce46bee0bc5e5b05caccd3984a63c1403a5)).

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -72,11 +72,22 @@ structure TranslateState where
...
+  return .ftvar s!"$__ty_unused_{id}"
+
+/-- Emit a diagnostic and return a fresh type variable as a placeholder.
+    Sets `coreProgramHasSuperfluousErrors` to signal that the Core program
+    may contain type errors that should prevent verification. -/
 private def throwTypeDiagnostic (ty : HighTypeMd) (msg : String) : TranslateM LMonoTy := do
   emitDiagnostic ((astNodeToCoreMd ty).toDiagnostic msg)
   modify fun s => { s with coreProgramHasSuperfluousErrors := true }
-  return .tcons "Error" []
+  freshTypePlaceholder
```
keyboardDrummer, line 92: What's the point of doing this, the line above has already assigned `coreProgramHasSuperfluousErrors := true` so the compilation pipeline stops and the Core program won't be evaluated?

keyboardDrummer (changes requested):

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: A `.Unknown` Laurel type means there's a type error. Compilation should stop before heading to Core.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -445,8 +468,8 @@ def translateStmt (stmt : StmtExprMd)
                 | _ => none
               let outArgs : List (Core.CallArg Core.Expression) := lhsIdents.map .outArg
               return [Core.Statement.call callee.text (coreArgs.map .inArg ++ outArgs) (astNodeToCoreMd value)]
-          | .InstanceCall .. =>
-              -- Instance method call: havoc all target variables
+          | .InstanceCall .. | .Hole _ _ =>
```
keyboardDrummer: A hole is an unknown value, not a potentially side-effecting computation, so this does not seem right

keyboardDrummer:
Can I ask what triggered this PR ?

tautschnig:
> Can I ask what triggered this PR ?

Trying to make more benchmarks work via the GOTO/CBMC pipeline, but arguably #1064 might be what we really need. The changes in this PR mainly make the root cause visible.

keyboardDrummer:
> Error is a real type (Python exceptions), and using it as a dummy caused type unification failures ("Impossible to unify Error with X") 

The Core pipeline shouldn't be run if Laurel emits an error type. We can also change `.tcons "Error" []` to `.tcons "ErrorOccurredInLaurelDoNotContinue" []`

> Also handle Hole (nondet) in Assign statements by emitting havoc instead of trying to translate the hole as an expression.

I don't think that implements the semantics of Hole correctly. The documentation for hole reads:
>   /-- A hole representing an unknown expression.
      - `deterministic`: if true, the hole represents a deterministic unknown
        (translated as an uninterpreted function); if false, a nondeterministic
        unknown (translated as a havoced variable). Nondeterministic holes are
        not allowed in functions.
      - `type`: inferred by the hole type inference pass; `none` means not yet inferred. -/
  | Hole (deterministic : Bool := true) (type : Option (AstNode HighType) := none)

Maybe `unknown expression` should be replaced with `unknown value` in the documentation, to make the hole semantics more clear.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
tautschnig: Fixed in [ca5daef](https://github.com/strata-org/Strata/pull/1063/commits/ca5daefc75f2e91a479b6050c816b0435672d669).

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
tautschnig: Added a comment - by continuing we at least can keep collecting diagnostics.

PR #1064, opened 2026-04-28
Fix field access resolution: do not treat field names as variables

tautschnig:
When resolveFieldRef cannot find a field name in any type scope, it previously fell back to resolveRef which looked up the field name as a regular variable. This caused spurious "Resolution failed: X is not defined" errors for dynamic field accesses like session.region_name where the field is not a variable in scope.

The fix: leave unresolved field names without error. Heap parameterization handles them later (replacing with readField or Hole for dynamic accesses).


Strata/Languages/Laurel/Resolution.lean
```diff
@@ -240,7 +240,7 @@ private def resolveFieldInTypeScope (typeName : String) (fieldName : Identifier)
     Falls back to the instance type name (for `self.field` in instance methods),
```
MikaelMayer: The docstring says "then to unqualified lookup if the target type cannot be determined" but that fallback was removed by this PR. Suggest updating to reflect the current behavior, e.g.:
```
/-- Resolve a field reference using the target's type to build a qualified lookup key.
    Falls back to the instance type name (for `self.field` in instance methods).
    Unresolved field names are returned as-is for later passes to handle. -/
```

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -250,7 +250,10 @@ def resolveFieldRef (target : StmtExprMd) (fieldName : Identifier)
   if let some instTypeName := (← get).instanceTypeName then
     if let some resolved ← resolveFieldInTypeScope instTypeName fieldName then
       return resolved
-  resolveRef fieldName md
+  -- Fallback: field names that can't be resolved in any type scope are left
+  -- unresolved without error. Heap parameterization will handle them (typically
+  -- replacing with a Hole for dynamic field accesses like s#region_name).
+  return fieldName
```
keyboardDrummer, line 259: Shouldn't this emit an error?

> Heap parameterization will handle them (typically
  -- replacing with a Hole for dynamic field accesses like s#region_name)

Can you say more about that?

MikaelMayer (commented):
Clean, well-scoped fix. The change correctly removes the fallback to `resolveRef` that was producing spurious errors for dynamic field accesses. The tests cover both the known-field and unknown-field cases. One stale docstring to update.

Strata/Languages/Laurel/Resolution.lean
tautschnig: Docstring updated in [329d47c](https://github.com/strata-org/Strata/pull/1064/commits/329d47c9014f88473de5154a696afc0d22b2c420).

Strata/Languages/Laurel/Resolution.lean
tautschnig: I have expanded that comment string as it seems later passes _do_ handle this - but I don't know whether this is intended behaviour! @thanhnguyen-aws @ssomayyajula perhaps know better what the intended behaviour is?

keyboardDrummer: Well, regardless of how Python is using field access, an unresolved field access in Laurel should be an error. Even the `if let some instTypeName := (← get).instanceTypeName then` on line slightly above is a hack that should be removed/replaced.

StrataTest/Languages/Laurel/DuplicateNameTests.lean
```diff
@@ -174,4 +174,31 @@ datatype Foo { A }
...
+};
+"
+
+#guard_msgs (error, drop all) in
+#eval testInputWithOffset "FieldAccessKnown" fieldAccessKnown 192 processResolution
+
+def fieldAccessUnknown := r"
+composite Session { var name: int }
+procedure test(s: Session) {
+  var x: int := s#region_name
```
MikaelMayer, line 197: I don't understand. There should be an error here. Perhaps not the error that states that region_name is not a variable, but there should be an error still.
Also, was it true that the following program was not emitting an error when it should?

```
def fieldAccessUnknown := r"
composite Session { var name: int }
procedure test(s: Session) {
  var region_name: int := 2;
  var x: int := s#region_name
```
? The PR description says that unresolved field access was treated as a variable before, so that mean the above test was passing?

PR #1072, opened 2026-04-28
Replace SpecType sorted-array with HashSet-backed representation

joehendrix:
## Summary

- Store int and string literals in `Std.HashSet` for O(1) amortized union instead of O(n) sorted-array merge
- Add named `SpecIdent` and `SpecTypedDict` structures for the idents and typed-dict array elements
- Retain `atoms` as a computed property for backward compatibility with iteration-based code
- Add field-based `toDDM` helpers (`SpecIdent.toDDM`, `SpecTypedDict.toDDM`) to avoid sorry in DDM serialization

Reduces pyspec prelude loading ("Resolve and build Laurel prelude") from **1369ms to 612ms (55%)** on one benchmark.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the MIT license.

### DIFF
--- a/Strata/Languages/Python/Specs/Decls.lean
+++ b/Strata/Languages/Python/Specs/Decls.lean
@@ -93,12 +93,31 @@ The `fieldRequired` array is parallel to `fields`/`fieldTypes`.
             (fieldRequired : Array Bool)
 deriving Inhabited, Repr
 
+structure SpecIdent where
+  name : PythonIdent
+  args : Array SpecType
+deriving Inhabited, Repr
+
+structure SpecTypedDict where
+  fields        : Array String
+  fieldTypes    : Array SpecType
+  fieldRequired : Array Bool
+deriving Inhabited, Repr
+
 /--
-A PySpec type is a union of atom types.
+A PySpec type is a union of atom types, stored with separate collections
+for each variant for efficient union operations on literal-heavy types.
 -/
 structure SpecType where
   private mk ::
-  atoms : Array SpecAtomType
+  /-- Named type identifiers, sorted by PythonIdent ordering. -/
+  idents     : Array SpecIdent
+  /-- Integer literal values. -/
+  intLits    : Std.HashSet Int
+  /-- String literal values. -/
+  stringLits : Std.HashSet String
+  /-- TypedDict types, sorted by field names. -/
+  typedDicts : Array SpecTypedDict
   /-- Source location of this type. May be `.none` for builtin types. -/
   loc : SourceRange
 deriving Inhabited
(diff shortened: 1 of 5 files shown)


Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -150,68 +169,139 @@ protected def SpecAtomType.compare (x y : SpecAtomType) : Ordering :=
     compare xfields yfields |>.then $
     compareHLex (fun ⟨xe, _⟩ ye => xe.compare ye) xfieldTypes.attach yfieldTypes |>.then $
     compare xisTotal yisTotal
+
+protected def SpecIdent.compare (x y : SpecIdent) : Ordering :=
+  compare x.name y.name |>.then $
+    compareHLex (fun ⟨xe, _⟩ ye => xe.compare ye) x.args.attach y.args
+termination_by sizeOf x
+decreasing_by cases x; decreasing_tactic
+
```
tautschnig, line 178: Nit: `SpecType.compare` calls `toArray.qsort` on both `intLits` and `stringLits` for every comparison, which allocates and sorts. This is fine since `compare` isn't on the hot path (union is), but worth a comment noting the trade-off.

tautschnig (approved):

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -150,68 +169,139 @@ protected def SpecAtomType.compare (x y : SpecAtomType) : Ordering :=
...
+/-- Total number of atoms across all fields. -/
+def size (tp : SpecType) : Nat :=
+  tp.idents.size + tp.intLits.size + tp.stringLits.size + tp.typedDicts.size
+
+/-- Reconstruct the flat array of atoms. Output order: idents (sorted),
+    int literals (sorted), string literals (sorted), typedDicts (sorted). -/
+def atoms (tp : SpecType) : Array SpecAtomType :=
+  let r := tp.idents.map fun si => .ident si.name si.args
+  let ints := tp.intLits.toArray.qsort (· < ·)
```
keyboardDrummer, line 218: The toArray calls looks a bit wasteful. Can we avoid them?

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -223,110 +313,77 @@ namespace SpecType
...
-  else
-    removeAdjDupsAux a 1 #[a[0]] (by simp +arith)
-
-/-- Construct a `SpecType` from an array of atoms by sorting and
-    removing duplicates to produce a canonical representation. -/
-private def ofArray (loc : SourceRange) (atoms : Array SpecAtomType) : SpecType :=
-  let elts := atoms.qsort (compare · · == .lt)
-  { loc := loc, atoms := removeAdjDups elts }
+  { idents     := sortedMerge x.idents y.idents
+    intLits    := x.intLits.union y.intLits
```
keyboardDrummer, line 337: How come this is amortized O(1) ? I thought that a HashSet union would be O(n)

keyboardDrummer (approved):

Strata/Languages/Python/Specs/Decls.lean
joehendrix: I think this is probably the fastest sort, but generally code that accesses this isn't super performance critical.

Strata/Languages/Python/Specs/Decls.lean
joehendrix: It isn't, per Lean, HashSet.union is `O(min(x.size, y.size))` as it merges smaller set into larger.

In parsing a DDM file, the second argument is typically going to be a singleton so it should be `O(1)` if either `x` or `y` is a singleton.  I thought about changing the DDM parsing code to explicitly add atomic type sets to a type set, but it would add more complexity and this is fast enough to shift the dominant cost to the file read.

Strata/Languages/Python/Specs/Decls.lean
joehendrix: I'll add a note in a future PR.

Issue #1050, opened 2026-04-24
Type checker fails when selector type variable name differs from function type variable


joscoh:
## Description

When a polymorphic function uses a different type variable name than the datatype it operates on, type checking fails with a misleading error.

## Minimal reproduction

```lean
def selectorTypeVarBug :=
#strata
program Core;

datatype Box (a : Type) { MkBox(val: a) };

inline function unbox<T>(b : Box T) : T
{
  Box..val(b)
}

procedure Test(out result : int)
spec {
  ensures result == 5;
}
{
  result := unbox(MkBox(5));
};
#end

#eval (Std.format ((Core.typeCheck .default (translate selectorTypeVarBug).stripMetaData)))
```

**Error:** `Expression has type a when T expected.`

## Expected behavior

The type checker should unify `a` with `T` since `b : Box T` implies `a = T`. The type variable name in the datatype definition should not matter.

## Workaround

Use the same type variable name as the datatype:

```
inline function unbox<a>(b : Box a) : a { Box..val(b) }
```

This is likely a bug in DDM type checking where selector return types use the original datatype type variable names without substitution.

Dragon-Hatcher:
See also #1022

Issue #1022, opened 2026-04-22
Inference for type parameters fails.


Dragon-Hatcher:
The following program which should work:
```
import Strata.MetaVerifier

open Strata

private def otp_program : Strata.Program :=
#strata
program Core;

function left<T, U>(t: T, u: U) : T {
  t
}

function right<V, W>(v: V, w: W) : W {
  w
}

function double(a: int) : int {
  right(a, a) + left(a, a)
}

#end

#eval Strata.Boole.verify "cvc5" otp_program (options := .quiet)
```
instead produces the error `Expression has type T when W expected.` at `left(a, a)`.

There is no error when `double` returns just `right` or just `left`. I am investigating the source of this issue.

PR #1074, opened 2026-04-29
Fix generic type inference

Dragon-Hatcher:
Fix generic type inference for user-defined polymorphic functions. See #1022 and #1050.                            
                                                                                               
  Calling multiple user-defined generic functions together in one expression would fail with a 
  spurious type error. For example:                                                            
  
```                                                                                             
function left<T, U>(t: T, u: U) : T { t }                       
function right<T, U>(t: T, u: U) : U { u }                                                   
function double(a: int) : int {                                                              
  right(a, a) + left(a, a)  -- error: Expression has type T when U expected                  
}                                                                                            
```
                                                                  
The issue is that type inference in the DDM layer treated the polymorphic types as literal strings. So the return type of `left` is `.tvar "T"` and the return type of `right` is `.tvar "U"`. As a result, swapping the names of the parameters on one of the functions would make the code compile.

To fix this, each time we instantiate a function call, we replace each `.tvar` with a `.uvar` with a unique index. When we infer a type for one of these `.uvar`s, we store it in the `ElabState`. An error is produced if identical `.uvar`s get solved to two different types.


joscoh:
Closing since https://github.com/strata-org/Strata/pull/1125 will bypass the issue for now

PR #27, opened 2025-07-31
Improve Strata DDM documentation

joehendrix:
This cleans up the Strata DDM documentation package to introduce example and basic concepts as well as improve readme.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #25, opened 2025-07-30
Add `bv1` type to Strata.Boogie

atomb:
Add a `bv1` type to Strata.Boogie to complement the other bit vector types.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1075, opened 2026-04-29
Boole language extensions for dalek-lite benchmarks

kondylidou:
## Summary

Extends the Boole language pipeline with new language features, curated benchmark
targets anchored to dalek-lite (Curve25519/Ed25519), and a testing infrastructure
upgrade across all Boole seeds.

## Language features

**`Sequence T` type and slicing ops**
- `toCoreMonoType` handles `.Sequence _ elem → .tcons "Sequence" [elem]`
- All 8 Core inherited ops wired up in `Verify.lean`
- Three Boole-specific wrappers: `Sequence.skip`, `Sequence.dropFirst`,
  `Sequence.subrange`
- Typed empty-sequence constants: `Sequence.empty_bv8/bv16/bv32/bv64/int` — each
  needs a distinct token since 0-ary polymorphic `Sequence.empty` has no arguments
  to infer the type from

**Bitvector loop variables** (`for i : bvN := init to limit`)
- `for_to_by` and `for_downto_by` dispatch guard/step/increment to
  `Bv{N}.ULe/Add/Sub` when the loop variable is a bitvector type

**`decreases` annotations**
- `for v := init to/downto limit` accepts an optional `decreases e` clause;
  forwarded to the Core while-loop measure field and actively verified by cvc5
- Functions and procedures accept an optional `decreases e` clause using Core's
  existing `Measure` category — no new grammar category introduced
- All three forms reuse Core's single `measure_mk` op; no duplicate constructs
- Function termination is verified by #1092; procedure-level `decreases` is
  silently dropped with a `dbg_trace` warning pending int-based termination support

**Lambda abstraction and application**
- `fun x : T => body` lowers to nested Core `.abs` nodes
- `(f)(x)` lowers to `.app () f x`

**Inline `let`-block postconditions**
- `let v := e in body` in spec/ensures positions lowers via `withBVarExprs`

**`choose` assignment**
- `w := choose z : T :: pred(z)` lowers to `havoc w; assume pred[z/w]`

**Bitvector comparisons**
- Unsigned (`<`, `<=`, `>`, `>=`) default to `Bv{N}.ULt/ULe/UGt/UGe` via
  `toBvCmpOp`
- Signed (`<s`, `<=s`, `>s`, `>=s`) lower to `Bv{N}.SLt/SLe/SGt/SGe`

## New seeds

- `embedded_postcondition.lean` — inline let-block in `ensures`
- `montgomery_loop_invariant.lean` — relational while-loop invariant; linear
  arithmetic case verifies via cvc5 and `smtVCsCorrect`
- `scalar_reduce.lean` — B2 `reduce()` axiom with abstract types
- `sha256_compact_indexed.lean` — SHA-256 compact port (indexed `Sequence`
  encoding); all 19 VCs pass; loop counters use `int` (faithful to Rust's `usize`
  semantics, avoids uninterpreted cast to `Sequence` index); remaining gaps:
  iterator protocol (#27), fixed-size array syntax (#25), slice types (#26)

Fully implemented seeds graduated from `FeatureRequests/` to the main Boole test
folder: `early_return.lean`, `choose_operator.lean`, `bitvector_ops.lean`,
`embedded_postcondition.lean`.

## Seed test infrastructure

Replaced `#guard_msgs (drop info) in` with explicit `/-- info: ... -/` +
`#guard_msgs in` across all Boole seeds. Added `(options := .quiet)` uniformly.

## Documentation

- New `docs/BooleBenchmarks.md`: five real-world benchmark targets from dalek-lite
- Updated `docs/BooleFeatureRequests.md`: all new seeds in inventory table,
  implemented features extended


tautschnig:
Would it be possible to have the non-Boole changes factored out into a separate PR? This PR is very large and it would be easy to fail to notice those changes. Also, we should have dedicated tests for those changes.

kondylidou:
> Would it be possible to have the non-Boole changes factored out into a separate PR? This PR is very large and it would be easy to fail to notice those changes. Also, we should have dedicated tests for those changes.

The only non-Boole related change is a little fix. It's one "word" fix that's why I didn't put it in it's own pr.

Edit: there might be ways to separate it, I do agree it's a large pr. Let me try to break it in two.

kondylidou:
@tautschnig was this:

Fix: smtVCsCorrect / gen_smt_vcs type mismatch for while-loop programs

DL/SMT/Translate.lean: integer literals (.prim (.int x)) were incorrectly typed as mkProp instead of mkInt in translateTerm, causing smtVCsCorrect and gen_smt_vcs to fail on any while-loop program with local var declarations. Fixed by returning (mkInt, toExpr x)

fixed already? Because after I merged main I don't see the fix anymore

kondylidou:
@tautschnig as the only change to Core is 2 lines I would prefer to keep the pr as is. Would that be ok with you? It's difficult to split but if you'd prefer a smaller pr then I can try to split it. Most changes are document changes. The language changes are small but a few so understandable that it seems like a lot. Everything is tested and has regression tests.

Strata/Languages/Boole/Verify.lean
```diff
@@ -508,6 +573,14 @@ def toCoreStmt (s : BooleDDM.Statement SourceRange) : TranslateM Core.Statement
...
       | .condNondet _ => pure .nondet
     return .ite cond thenb elseb (← toCoreMetaData m)
+  | .choose_assign m ⟨_, lhs⟩ _v pred =>
+    let lhsExpr : Core.Expression.Expr := .fvar () (mkIdent lhs) none
+    let predExpr ← withBVarExprs #[lhsExpr] (toCoreExpr pred)
+    let md ← toCoreMetaData m
+    let label ← defaultLabel m "choose" none
+    let havocStmt := Core.Statement.havoc (mkIdent lhs) md
+    let assumeStmt := Core.Statement.assume label predExpr md
+    return .block label [havocStmt, assumeStmt] md
```
tautschnig: **Soundness: the lowering is weaker than the intended encoding.** The seed file `StrataTest/Languages/Boole/choose_operator.lean:17-19` states the intended lowering is:

```
havoc z; assume (∃ z' . g(z')) ==> g(z);
```

with the existence guard. What this code emits is `havoc w; assume pred[z/w]` — no existence check. Consequence: if a user writes

```
w := choose z: int :: z != z;   -- unsatisfiable predicate
assert false;                    -- verifies trivially
```

the `havoc w; assume false` path makes `assert false` (and any subsequent obligation) pass with no counterexample, because `assume false` contradicts the path and the solver reports `unsat` on `P ∧ ¬false`. That's a **false-positive verification** — a real soundness issue.

The current test suite doesn't catch this because `choose_operator.lean:28` carries `requires ∃ z: int . good(z, x)` as an assumption the solver can see. A program without that `requires` produces unsound proofs.

Suggested fix — emit the assert-then-havoc-then-assume encoding:

```suggestion
  | .choose_assign m ⟨_, lhs⟩ _v pred =>
    let lhsExpr : Core.Expression.Expr := .fvar () (mkIdent lhs) none
    let predExpr ← withBVarExprs #[lhsExpr] (toCoreExpr pred)
    let md ← toCoreMetaData m
    let label ← defaultLabel m "choose" none
    -- Existence obligation: before havocing, prove that *some* witness exists
    -- for the predicate.  Without this guard, `assume pred(w)` after
    -- `havoc w` can silently `assume false` when `pred` is unsatisfiable,
    -- making the rest of the program verify trivially.  See seed comment at
    -- `choose_operator.lean:17-19`.
    let existsBody ← withBVarExprs #[Core.Expression.Expr.bvar () 0]
      (toCoreExpr pred)
    let .mono_bind_mk _ _ vty := _v
    let existsExpr : Core.Expression.Expr :=
      .quant () .exists "" (some (← toCoreMonoType vty)) (.noTrigger ()) existsBody
    let existenceAssert := Core.Statement.assert
      s!"{label}_exists" existsExpr md
    let havocStmt := Core.Statement.havoc (mkIdent lhs) md
    let assumeStmt := Core.Statement.assume label predExpr md
    return .block label [existenceAssert, havocStmt, assumeStmt] md
```

(Exact quant/trigger spelling to be matched to Core's surface — the idea is the existence obligation fires as an assert before the havoc.)

Test to add (should verify after the fix, and *fail* under the current lowering so the regression is visible):

```lean
private def chooseUnsatSeed : Strata.Program :=
#strata
program Boole;
procedure bad() returns (w: int) { w := choose z: int :: z != z; assert false; };
#end

/-- info: Obligation: … Property: assert Result: ❌ fail -/
#guard_msgs in
#eval Strata.Boole.verify "cvc5" chooseUnsatSeed (options := .quiet)
```

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -298,6 +298,8 @@ op ensures_spec (label : Option Label, free? : Option Free, b : bool) : SpecElt
   free?:0 "ensures " label b ";\n";
 op requires_spec (label : Option Label, free? : Option Free, b : bool) : SpecElt =>
   free?:0 "requires " label b ";\n";
+op decreases_spec (e : Expr) : SpecElt =>
+  "decreases " e ";\n";
```
tautschnig: **Grammar production without Core translator support.** `decreases_spec` is added to Core's `SpecElt` category here, but Core's `translateSpecElem` (Translate.lean:1543) errors with `translateSpecElem unimplemented for …` on anything other than `requires_spec` / `ensures_spec`, and `translateFnPreconds` (Translate.lean:1247) errors with `only requires allowed`. Only Boole's own `toCoreSpecElts` (Verify.lean:765) silently drops it.

Net result: a user writing pure Core (not via Boole) with `decreases e;` in a `spec { … }` block syntactically parses but hits an `unimplemented` translator error. That's worse than either (a) rejecting at the grammar level, or (b) silently dropping in Core's translator to match Boole.

Pick one:

- **If `decreases` is a Boole-level surface concept**, move this grammar production into Boole's Grammar.lean (extending the `SpecElt` category there), not Core's. Boole's `toCoreSpecElts` stays as-is; Core stays un-touched.
- **If `decreases` should eventually exist in Core too** (proper procedure-level measure, like `recfn_decl`'s PR #1092), add a no-op arm in `translateSpecElem` with a `dbg_trace s!"decreases_spec ignored — TODO procedure-level termination"` so Core accepts it now but flags the missing translator work.

Either option makes the grammar/translator pair self-consistent.

Strata/Languages/Boole/Grammar.lean
```diff
@@ -70,24 +80,59 @@ fn forall_unicodeT (d : DeclList, @[scope(d)] triggers : Triggers, @[scope(d)] b
...
+// the DDM parser cannot resolve without arguments, so each element type uses
+// a distinct token `Sequence.empty_<type>`.
+fn seq_empty_bv8  () : Sequence bv8  => "Sequence.empty_bv8";
+fn seq_empty_bv16 () : Sequence bv16 => "Sequence.empty_bv16";
+fn seq_empty_bv32 () : Sequence bv32 => "Sequence.empty_bv32";
+fn seq_empty_bv64 () : Sequence bv64 => "Sequence.empty_bv64";
+fn seq_empty_int  () : Sequence int  => "Sequence.empty_int";
+
+fn seq_skip (A : Type, s : Sequence A, n : int) : Sequence A =>
+  "Sequence.skip" "(" s ", " n ")";
```
tautschnig, line 131: **Typed empty-sequence constants are a workaround that doesn't scale.**

The comment is honest ("the DDM parser cannot resolve without arguments"), but:

- `Sequence String`, `Sequence (Sequence bv32)`, user-defined element types can't construct an empty literal today.
- Adding a new element type requires **two** edits (Grammar.lean here + Verify.lean:424-425), which future contributors will forget.
- This is a DDM-polymorphism workaround, not a Boole language feature — it belongs in the DDM backlog, not permanently in Boole's grammar.

Mitigations, roughly in order of preference:

1. **File a tracking issue for DDM 0-ary polymorphic resolution** (or for type-ascription-driven resolution, e.g. `Sequence.empty : Sequence String`). Replace the five tokens with a single `Sequence.empty` once that's supported.
2. **Narrow the 5-way enumeration to a single elaboration-time constructor**. If DDM's type inferer can already resolve `Sequence.empty` when the surrounding context gives it a concrete `Sequence T`, that's preferable.
3. **Document the add-new-type procedure** at the top of this block so the two-file edit isn't tribal knowledge.

At minimum (3) — a one-line comment naming the sibling edit site — is a no-cost improvement:

```suggestion
-- Typed empty-sequence constants. `Sequence.empty` is 0-ary polymorphic which
-- the DDM parser cannot resolve without arguments, so each element type uses
-- a distinct token `Sequence.empty_<type>`.
--
-- To add a new element type: extend this list AND update the
-- `.seq_empty_*` match in `Verify.lean`'s `toCoreExpr`.
-- TODO: remove these in favour of a single `Sequence.empty : Sequence T`
-- when DDM supports 0-ary polymorphic resolution (tracking: <issue-number>).
fn seq_empty_bv8  () : Sequence bv8  => "Sequence.empty_bv8";
fn seq_empty_bv16 () : Sequence bv16 => "Sequence.empty_bv16";
fn seq_empty_bv32 () : Sequence bv32 => "Sequence.empty_bv32";
fn seq_empty_bv64 () : Sequence bv64 => "Sequence.empty_bv64";
fn seq_empty_int  () : Sequence int  => "Sequence.empty_int";
```

Strata/Languages/Boole/Verify.lean
```diff
@@ -402,18 +455,30 @@ def toCoreInvariants (is : BooleDDM.Invariants SourceRange) :
...
 
+-- Return (leOp, addOp, subOp, one) appropriate for the loop variable type.
+private def forArithOps (ty : Lambda.LMonoTy) :
+    Core.Expression.Expr × Core.Expression.Expr × Core.Expression.Expr × Core.Expression.Expr :=
+  match ty with
+  | .bitvec n =>
+    (.op () ⟨s!"Bv{n}.ULe", ()⟩ none,
+     .op () ⟨s!"Bv{n}.Add", ()⟩ none,
+     .op () ⟨s!"Bv{n}.Sub", ()⟩ none,
+     .bitvecConst () n 1)
```
tautschnig, line 529: `forArithOps` picks `Bv{n}.ULe/Add/Sub` based purely on the loop variable's declared type. The emitted guard / step have no compatibility check against the init, limit, or step expressions, so a program like

```
for i : bv32 := bv{32}(0) to someBv16Var by bv{64}(1)
```

emits `Bv32.ULe(i, someBv16Var)` and `Bv32.Add(i, bv{64}(1))` — both ill-typed. The mismatch surfaces as a low-level type-mismatch error in the SMT encoder, far from the user's source.

Add a front-end compatibility check so the error fires here with source-range info:

```suggestion
-- Return (leOp, addOp, subOp, one) appropriate for the loop variable type.
-- NOTE: caller is responsible for ensuring init/limit/step all have the
-- same bitvector width as the loop variable; otherwise the emitted ops
-- (`Bv{n}.ULe`, `Bv{n}.Add`, `Bv{n}.Sub`) will fail the SMT encoder's
-- type check with a low-level diagnostic.  A front-end type-compat check
-- would surface the error at the source location of `for …`.
private def forArithOps (ty : Lambda.LMonoTy) :
    Core.Expression.Expr × Core.Expression.Expr × Core.Expression.Expr × Core.Expression.Expr :=
  …
```

An explicit check in `lowerFor` that the computed init/guard/step expression types match the loop-var type (via the existing `toCoreMonoType` + `LMonoTy` equality) would make the error user-visible. Deferable to a follow-up, but worth a TODO comment.

Strata/Languages/Boole/Verify.lean
```diff
@@ -678,6 +762,7 @@ private def toCoreSpecElts (_m : SourceRange) (pname : String) (elts : Array (Bo
   for e in elts.toList do
     match e with
     | .modifies_spec _ _ => pure ()
+    | .decreases_spec _ _ => pure ()
```
tautschnig: **Silent drop of procedure/function-level `decreases`.** `toCoreSpecElts` sees a `decreases_spec` and returns `pure ()`, losing the measure expression entirely. Combined with the per-procedure `_decr` silent drop at `:903` (`| .boole_procedure m nameAnn targsAnn ins outsAnn _decr specAnn bodyAnn =>`), this means:

- Writing `decreases n` on a procedure or function has zero effect on verification.
- A user who believes the annotation is enforced gets no feedback that it isn't.

The test `decreasesFunctionSeed` (FeatureRequests/decreases_metadata.lean) specifically demonstrates this — the comment says "silently dropped by the Boole lowering" and the procedure it tests is non-recursive, so the drop is benign. But a user who writes a *recursive* Boole function or procedure with `decreases` and hits it hasn't gotten termination checking.

Two options:

1. **Pass `decreases` through to Core.** The for-loop arms (`:680`, `:695`) already forward the measure to `Core.Statement.loop`'s measure field. Do the analogous thing for procedure/function decreases: Core's `recfn_decl` got a `decreases` clause in PR #1092 — consider whether `Function.measure` / `Procedure.measure` should be populated here.
2. **Emit a diagnostic.** If proper propagation is out of scope for this PR, at minimum emit a one-line `dbg_trace` / user-visible warning so the silent drop is visible:

```suggestion
    | .modifies_spec _ _ => pure ()
    | .decreases_spec m _ =>
      -- TODO: propagate to `Procedure.measure` / `Function.measure`; for
      -- now the annotation is lost, which means recursive definitions
      -- carrying `decreases` are not termination-checked at the Core
      -- level.  Emit a warning so the silent drop is visible.
      dbg_trace s!"Boole: procedure/function-level `decreases` at {repr m} is ignored by the current lowering (see FeatureRequests/decreases_metadata.lean)"
      pure ()
```

docs/BooleFeatureRequests.md
```diff
@@ -14,26 +14,53 @@ This document tracks the selected Boole feature-request seeds kept under
...
 - **Mutual recursion over datatypes** (#599)
-  - `rec function ... ;` blocks work end-to-end; two `Verify.lean` fixes: `lowerPureFuncDef` propagates `@[cases]` to `FuncAttr.inlineIfConstr`, and `toCoreDecls` injects preceding sibling op-exprs as De Bruijn bvars so cross-sibling calls resolve.
-  - Remaining gap: mutual recursion over `int` still needs function-level `decreases` (not yet implemented).
+  - `rec function ... ;` blocks work end-to-end for structural recursion over datatypes.
+  - Remaining gap: mutual recursion over `int` — see #17.
   - Benchmark: [`mutual_recursion.lean`](../StrataTest/Languages/Boole/FeatureRequests/mutual_recursion.lean) (`even`/`odd` over `MyNat`).
+- **`choose` syntax**
+  - `w := choose z : T :: pred(z)` desugars to `havoc w; assume pred[z/w]`.
+  - Benchmark: [`choose_operator.lean`](../StrataTest/Languages/Boole/choose_operator.lean).
+- **`decreases` annotation on functions, procedures, and `for` loops**
```
joscoh, line 49: FYI: decreases on functions will be added in https://github.com/strata-org/Strata/pull/1092. Procedures will come shortly (after int-valued recursive functions)

Strata/Languages/Boole/Grammar.lean
```diff
@@ -45,11 +45,21 @@ op boole_procedure (name : Ident,
...
+              decr s body ";\n";
+
+// choose assignment: `w := choose z : T :: pred(z);`
+// Lowers to: `havoc w; assume pred[z/w];`
+// `" :: "` is used (not `" . "`) for the same reason as `Sequence.skip` etc.:
+// the DDM init dialect parses `id.id` as a qualifiedIdentExplicit before
+// Expr-level trailing rules apply, so a bare dot between bound variable and
+// predicate would parse as a field access on the variable name.
+op choose_assign (lhs : Ident, v : MonoBind, @[scope(v)] pred : bool) : Statement =>
+  lhs " := choose " v " :: " pred ";";
```
joscoh, line 69: IIUC, this is HIlbert's epsilon operator, right? How do you ensure that `pred` is inhabited? E.g. what would `chose x :: false` give?

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -298,6 +298,8 @@ op ensures_spec (label : Option Label, free? : Option Free, b : bool) : SpecElt
   free?:0 "ensures " label b ";\n";
 op requires_spec (label : Option Label, free? : Option Free, b : bool) : SpecElt =>
   free?:0 "requires " label b ";\n";
+op decreases_spec (e : Expr) : SpecElt =>
```
joscoh: https://github.com/strata-org/Strata/pull/1092 uses the existing `Measure` category in `Core for decreases clauses. We should avoid duplicate constructs when possible.

joscoh (commented):
Looks good, just a few minor comments

tautschnig (commented):
Large, useful PR — the dalek-lite benchmark targets are a nice north-star, the per-seed VC snapshots are a solid regression harness, and the test-infrastructure cleanup (`/-- info: ... -/` + explicit `#guard_msgs in` + uniform `.quiet`) makes future churn less painful. `lake build` passes. No prior review threads.

Six concerns, roughly in priority order — the first is the only one I'd call a blocker.

**1. Soundness: `choose_assign` lowering is weaker than the intended encoding.** The seed file's own comment (`StrataTest/Languages/Boole/choose_operator.lean` lines 17–19) names the intent: `havoc z; assume (∃ z' . g(z')) ==> g(z)`. The actual lowering in `Verify.lean:577-583` is `havoc w; assume pred(w)` with no existence guard. When `pred(w)` is unsatisfiable, `assume pred(w)` becomes `assume false` and every subsequent obligation verifies trivially — false-positive verifications without a counterexample. The current `choose_operator.lean` test happens to supply `requires ∃ z: int . good(z, x)` in its spec, which the SMT solver carries as an assumption, so the gap is invisible to the suite. I'd consider this a blocker: implement the guard, or (at minimum) require callers to assert existence at the call site. Inline on `Verify.lean:576-583`.

**2. Core grammar / Core translator mismatch for `decreases_spec`.** `Strata/Languages/Core/DDMTransform/Grammar.lean:301-302` adds `decreases_spec` to the `SpecElt` category, which makes it syntactically valid in *any* Core program's `spec { … }` block and in function preconds. Only Boole's `toCoreSpecElts` (`Verify.lean:765`) silently drops it; Core's `translateSpecElem` (Translate.lean:1543) and `translateFnPreconds` (:1247) both throw `unimplemented`/`only requires allowed`. Adding a grammar production that Core's own translator rejects is surprising — either lift the production to Boole, or teach Core's translator to consume it (silently drop if appropriate). Inline.

**3. Typed empty-sequence constants are a 5-token workaround that doesn't scale.** `Grammar.lean:107-111` hand-rolls `Sequence.empty_bv8/16/32/64/int`. Three gaps:
  - `Sequence String`, `Sequence (Sequence T)`, user-defined element types → no empty literal.
  - Adding a new element type requires editing the grammar *and* `Verify.lean:424-425`.
  - Conceptually this is a DDM-polymorphism workaround, not a Boole language feature — the comment even says so ("the DDM parser cannot resolve without arguments").

Worth either (a) filing a tracking issue for 0-ary polymorphic resolution in DDM and leaving this as an explicit TODO, or (b) gating empty literals behind a type annotation (`Sequence.empty : Sequence T`) so the type system drives resolution. As-is it's silent coverage debt. Inline on the Grammar hunk.

**4. `forArithOps` assumes the loop variable and the init/limit/step have matching bitvector widths.** `Verify.lean:458-466` picks `Bv{n}.ULe/Add/Sub` based on the loop variable's static type but doesn't cross-check against the init/limit/step types. A user writing `for i : bv32 := 0bv64 to n : bv16` — or more plausibly, `for i : bv32 := 0 to n` where the int-to-bv32 coercion isn't inserted — will emit `Bv32.ULe(i, n)` where `n : bv16`, which errors inside the SMT encoder with a low-level type-mismatch message. Add a type-compatibility check in `lowerFor` so the error fires at the front-end instead. Inline.

**5. Test coverage is happy-path only.** Every new seed uses postconditions that pass. Nothing in the suite catches:
  - Unsatisfiable `choose` predicate (concern 1): `procedure bad() returns (w: int) { w := choose z: int :: z != z; assert false; };` — should fail verification. Today it silently verifies.
  - Lambda type mismatch: `(fun x : int => x)(true)` — should error at type-check or propagation.
  - Loop `decreases` with non-decreasing measure: should fail termination (if the decreases is enforced; the PR drops procedure/function-level decreases silently).
  - Sequence out-of-bounds access: `Sequence.select(Sequence.empty_int, 0)` — shape test for the precondition obligation added in PR #1100.

Each is a small fixture; concern 1's test is the one I'd want to see before merge.

**6. `let_in_expr` substitution doesn't defend against variable capture.** `Verify.lean:353-355` is `withBVarExprs #[value']` which inserts `value'` as a bvar substitution in `body`. If `value'` contains free variables that happen to share names with `body`'s binders, the substitution can capture them. For the current seeds (pure arithmetic, no name collision) this is fine, but the lowering should note the assumption.

**Proof-coverage suggestions** (three small theorems, each `rfl`-shaped once the constructors unfold):

```lean
/-- `choose` with existence guard preserves postconditions. Formalises the
    intended encoding from `choose_operator.lean:17-19`. -/
theorem choose_assign_sound (P : α → Prop) (w : α)
    (hex : ∃ z, P z) : P w ∨ True
    -- shape depends on the Boole semantics spec; the point is: the
    -- *intended* lowering (with the existence guard) is sound and
    -- the *current* lowering is not.

/-- `Sequence.subrange` desugaring preserves length when bounds are valid. -/
theorem seq_subrange_length (s : Sequence α) (lo hi : Int)
    (h_lo : 0 ≤ lo) (h_hi : lo ≤ hi) (h_len : hi ≤ Sequence.length s) :
    Sequence.length (Sequence.take (Sequence.drop s lo) (hi - lo)) = hi - lo

/-- `Sequence.skip` desugaring is observationally equal to `Sequence.drop`. -/
theorem seq_skip_desugar (s : Sequence α) (n : Int) :
    Sequence.skip s n = Sequence.drop s n := rfl
```

The first is the one that catches the soundness bug — if it fails to prove under the *current* lowering, that's the signal.

**Refactor:**

- The `for_to_by_statement` / `for_down_to_by_statement` arms in `Verify.lean:665-700` are near-identical. Extract once, parameterise on `addOp`/`subOp` + `initExpr`/`guardExpr` shape.
- `toBvCmpOp` silently defaults comparisons to unsigned (`<` → `ULt`). Worth a one-line doc comment at `:259-263` stating this is the *convention* (Boole users opting into signed comparison use `<s`).
- Grammar has a good comment block on `choose_assign`'s `::` vs `.` ambiguity; the parallel `Sequence.xxx`-as-one-token workaround at `Grammar.lean:94-102` is also nicely explained. Same ambiguity logic, same comment — consider extracting to a one-paragraph header comment on "DDM `id.id` parsing ambiguities and how this file works around them".

Strata/Languages/Core/DDMTransform/Grammar.lean
kondylidou: Thanks! Now procedure-level decreases is a Boole-level surface concept expressed via the existing Measure category, Core is untouched.

Strata/Languages/Boole/Verify.lean
kondylidou: thanks. added a trace

Strata/Languages/Boole/Verify.lean
kondylidou: Thanks for that! Fixed choose_assign soundness bug by emitting `assert ∃ z : T . pred(z)` before the havoc. Added the regression test (chooseUnsatSeed) with predicate `z != z` that now correctly fails the existence VC instead of verifying as a spurious pass.

Strata/Languages/Boole/Grammar.lean
kondylidou: look at the fix:) and thanks for pointing this out

Strata/Languages/Boole/Grammar.lean
kondylidou: 1. File a tracking issue for DDM 0-ary polymorphic resolution 
this was the intention. I just hardcoded some options for an important test to pass until there is Sequence.empty. But this is supposed to be a temporary patch.

kondylidou:
@tautschnig @joscoh can you please have a look? I addressed all your comments and added 2 TODO for follow up prs.

Strata/Languages/Boole/Verify.lean
```diff
@@ -303,7 +303,7 @@ private def oldifyExpr (inoutNames : List String) : Core.Expression.Expr → Cor
 
 mutual
 
-def toCoreQuant
+partial def toCoreQuant
```
joscoh, line 353: Wondering why these became `partial`?

StrataTest/Languages/Boole/FeatureRequests/sha256_compact_indexed.lean
```diff
@@ -152,34 +189,32 @@ spec {
...
   var tmp4 : (Sequence bv32);
-  var k : bv64;
+  var k : int;
   state_out := state;
-  k := bv{64}(0);
-  while (bv64_to_int_u(k) < Sequence.length(blocks))
-  decreases Sequence.length(blocks) - bv64_to_int_u(k)
-  invariant bv64_to_int_u(k) <= Sequence.length(blocks)
+  k := 0;
+  while (k < Sequence.length(blocks))
```
joscoh: Not a blocker, but wondering why this is not a `for` loop

joscoh (commented):
Could you please change the branch to `main2`? Thanks

joscoh:
Actually, never mind about `main2`, we will need a separate PR for the internal benchmarks anyway

joscoh (dismissed):

StrataTest/Languages/Boole/FeatureRequests/sha256_compact_indexed.lean
kondylidou: Cheng changed it to a for loop. I will update the benchmark

StrataTest/Languages/Boole/FeatureRequests/sha256_compact_indexed.lean
```diff
@@ -137,26 +113,30 @@ spec {
...
       call tmp22 := rotate_right(w2, bv{32}(17));
 
       call tmp23 := rotate_right(w2, bv{32}(19));
 
       assert 0 <= 10 && 10 < 32;
       s1 := tmp22 ^ tmp23 ^ (w2 >> bv{32}(10));
+      assert bv{64}(16) != bv{64}(0);
+      assert bv{64}(16) != bv{64}(0);
       new_w := Sequence.select(block_local, (i - 16) mod 16) + s0 + Sequence.select(block_local, (i - 7) mod 16) + s1;
+      assert bv{64}(16) != bv{64}(0);
```
joscoh: Wondering why this needs to be asserted so many times

joscoh (dismissed):

StrataTest/Languages/Boole/FeatureRequests/sha256_compact_indexed.lean
kondylidou: probably the translation did this? @ChengZ3 ?

ChengZ3: That's true. Verus synthesizes one divisor non-zero check for each mod operation. Line 138 has two mod ops so we have duplicated assertions. For the benchmark it is fine to remove them.

docs/BooleFeatureRequests.md
```diff
@@ -40,22 +40,22 @@ This document tracks the selected Boole feature-request seeds kept under
...
-  - `decreases` on functions: termination verification implemented (#1092).
-  - `decreases` on procedures: `decr : Option Measure` parameter on `boole_procedure`, reusing Core's existing `Measure` category (consistent with `command_function`); currently parsed and silently dropped — termination verification coming after int-based recursive function support.
-  - Remaining gap (→ #1167): `decreases <int expr>` closes the language gap; int-recursive functions are pure UFs in SMT — manual axioms still needed for functional properties.
+  - `decreases` on functions (structural): termination verification implemented (#1092).
+  - `decreases <int expr>` on `rec function`: implemented in `upstream/main2` (#1167), pending merge into `upstream/main`. Non-negativity and strict-decrease obligations generated at each call site. Int-recursive functions are pure UFs in SMT — no definitional axioms; manual axioms still needed for functional properties.
```
joscoh: Minor: I wouldn't mention `main2` and `main` here. For these purposes, `main2` should be considered the primary developement branch for the immediate future.

joscoh (commented):
Can you change the PR to target `main2` and remove the comments about things being upstreamed to `main`?

kondylidou:
> Can you change the PR to target `main2` and remove the comments about things being upstreamed to `main`?

done:)

joscoh (dismissed):

Strata/Languages/Boole/Verify.lean
shigoel: +1

shigoel (dismissed):
LGTM, modulo the `partial` functions -- won't block on that right now though

joscoh (approved):

shigoel (approved):

PR #1078, opened 2026-04-29
Add missing 'source-version' to the internal benchmark command

aqjune-aws:
This patch adds the missing 'source-version' to the internal benchmark command.

- The first commit should fail: the S3 log will show that internal benchmarks all fail because the commit contains wrong modification (and Codebuild must have picked it up).
- The second commit's S3 log should show that the internal benchmarks ran successfully.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -338,10 +338,12 @@ jobs:
       - name: Trigger CodeBuild and wait
         shell: bash
         run: |
+          aws --version
           BUILD_ID=$(aws codebuild start-build \
             --project-name strata-benchmarks \
             --source-type-override GITHUB \
             --source-location-override https://github.com/strata-org/Strata.git \
+            --source-version ${{ github.event.pull_request.head.sha || github.sha }} \
             --query 'build.id' --output text \
             --region us-east-2)
           echo "Build started: $BUILD_ID"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -347,6 +347,7 @@ jobs:
             --query 'build.id' --output text \
             --region us-east-2)
           echo "Build started: $BUILD_ID"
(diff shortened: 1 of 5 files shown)


robin-aws (changes requested):
Good call trying to show a failed build based on this fix first! But it looks like the build failed in the regular "Build and Test Lean" job and the CodeBuild check still passed, which seems to say the fix isn't working yet.

aqjune-aws:
It is a good catch. The failing step from the 'Build and test Lean' check is `Run pyAnalyze SARIF tests` which runs a sequence of `python run_py_analyze_sarif.py --laurel`s. This is currently not being run in the Codebuild setup, because pyAnalyze SARIF tests is publicly available and hence not chosen to be run in Codebuild. There are a few tests like this, including `Run pyInterpret golden-test files` (this is also run during 'Build and test Lean', but not during Codebuild).

aqjune-aws:
To summarize, the tests run on Codebuild are not strictly a superset of tests running on Github Action, so there could be a failure that is caught by Github Action whereas Codebuild could not.

robin-aws:
Fair, so the check may be fixed and it's just that we don't have a successful demonstration of that yet. Can you introduce a bigger temporary breakage that will make building the Lean itself fail? That way we should see both the GitHub action and the CodeBuild job fail.

aqjune-aws:
Having a look...

aqjune-aws:
The CI checks of commit '[rerun CI](https://github.com/strata-org/Strata/pull/1078/commits/71a4f9dd4757c4a880738bf1db0d2452fbc9fce4)' has both main Lean build CI and Codebuild CI failed. :) The Buildspec had to be slightly edited (`set -e`).

In the last commit, I updated the Github Action to print the Codebuild link.

robin-aws (approved):
Great, NOW I'm convinced! :)

tautschnig (approved):

Issue #993, opened 2026-04-21
[Proposal] Documented specification for `exit` and labeled-block semantics


seebees:
**Target repo:** `strata-org/Strata`
**Kind:** Documentation + proven-properties contribution
**Reference branch:** [`seebees/experimental-work`](https://github.com/seebees/Strata/tree/seebees/experimental-work) on the seebees fork

---

## Summary

The small-step semantics of `exit` and labeled blocks
in `Strata/DL/Imperative/StmtSemantics.lean` is implemented but not documented.
`Config.exiting`, `Config.block`, `step_exit`, and the `step_block_exit_*` family of transitions
together define the exit behavior —
but there is no written spec a contributor can point to,
and no reusable set of proven properties
that downstream features
(Laurel exceptions, break/continue, Python `try`/`except`)
can cite as lemmas.

This issue proposes contributing:

1. A design document under `docs/design/exit-semantics/`
   describing the behavior of `exit` over the current small-step relation.
2. Two Lean files with proven properties about that behavior:
   - `Strata/DL/Imperative/ExitProperties.lean` —
     ten theorems (E1–E10) over `StepStmt` and `StepStmtStar`.
   - `Strata/DL/Imperative/ExitWellFormedness.lean` —
     the `WFLabels` / `WFLabelsBlock` predicates and exit-resolution theorems.

All proofs are zero-sorry.
The reference implementation is in daily use
by the Laurel exception translation,
JVerify's Java→Laurel frontend,
and a downstream formal-verification project.

We are filing this issue first,
per CONTRIBUTING.md's guidance on small focused PRs,
so maintainers can shape the scope before PRs open.

## Motivation

`exit` is the control-flow primitive under several surface features:

- Laurel `throw` / `try` / `catch` / `finally`
- Early return (desugars to `exit $body`)
- Python `try` / `except` (already present in `PythonToLaurel.lean`)
- Potential future `break` / `continue` in loops

Every downstream proof that reasons about any of these
currently has to re-derive the core facts —
that an exit preserves the store,
that a matching labeled block consumes it,
that a non-matching label propagates it,
and so on.

With a named set of proven theorems and a written spec,
those proofs become one-liners citing
`matching_block_consumes`,
`nonmatching_exit_propagates`,
`exit_target_in_labels`,
etc.

### 1. `docs/design/exit-semantics/`

- `spec.md` —
  states the behavior in terms the code uses today:
[...]

PR #1080, opened 2026-04-30
docs: Add exit-semantics design documentation

seebees:
Adds a spec and decisions document for the small-step semantics of `exit` and labeled blocks in `Strata/DL/Imperative/StmtSemantics.lean`.

The spec uses RFC 2119 normative language only where an obligation maps to a specific Lean construct:
- Preservation of well-formedness by every step, discharged by `step_preserves_exitsCoveredByBlocks`.
- No-escape (exit does not leave a well-formed procedure body), discharged by `exitsCoveredByBlocks_noEscape` and its statement-list companion.

Per-transition behavior is documented descriptively, since each transition is directly realized by a constructor of `StepStmt` and does not carry a separate normative obligation.

The decisions document records six design choices and a future-work section, incorporating feedback from #993 and #869.

Tracks: https://github.com/strata-org/Strata/issues/993

### DIFF
--- a/Strata/DL/Imperative/ExitProperties.lean
+++ b/Strata/DL/Imperative/ExitProperties.lean
@@ -1,0 +1,120 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+public import Strata.DL.Imperative.StmtSemantics
+
+/-!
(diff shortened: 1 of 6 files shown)


atomb:
This is a very nice addition, but currently has a build failure. If you can resolve that and retarget it to `main2`, we will review it.

PR #1083, opened 2026-04-30
Support str.prefixof and str.suffixof in Core dialect

### LINKED ISSUE #1081: Support str.prefixof and str.suffixof in Core dialect
The SMT layer (`Strata/DL/SMT/Op.lean`) already defines `str_prefixof` and `str_suffixof`, and the SMT encoder recognizes them as reserved symbols. However, the Core dialect does not expose these operations to users.

Currently, `StrOpKind` in `CoreOp.lean` only includes:
```
Length | Concat | Substr | ToRegEx | InRegEx
```

To support `str.prefixof` and `str.suffixof` in Core, the following changes are needed:

1. **`CoreOp.lean`** — Add `PrefixOf` and `SuffixOf` to `StrOpKind` and its `names` list
2. **`Factory.lean`** — Add `strPrefixOfFunc` and `strSuffixOfFunc` definitions (binary: `string × string → bool`)
3. **`SMTEncoder.lean`** — Map `.str .PrefixOf` to `Op.str_prefixof` and `.str .SuffixOf` to `Op.str_suffixof`
4. **DDM grammar** — Add parsing support for the new operations in the `.core.st` surface syntax

Both operations have the SMT-LIB signature `(String, String) → Bool`:
- `str.prefixof(s, t)` returns true iff `s` is a prefix of `t`
- `str.suffixof(s, t)` returns true iff `s` is a suffix of `t`

MikaelMayer:
Fixes #1081

Adds `str.prefixof` and `str.suffixof` support to the Core dialect. These SMT-LIB string operations (`(String, String) → Bool`) were already defined in the SMT layer but not exposed in Core.

Both operations use `binaryOp` with concrete Lean implementations (`String.isPrefixOf` and `String.endsWith`), enabling the evaluator to simplify concrete inputs (e.g. `str.prefixof("ab", "abcd")` → `true`).

The SMT encoding maps to the existing `Op.str_prefixof` / `Op.str_suffixof`.

Tested: all existing tests pass, new tests added for CoreOp round-trips, AST→CST round-trips, SMT encoding output, and concrete evaluation with literal strings (both true and false cases resolve without the SMT solver).


MikaelMayer (commented):
Clean, well-structured PR. The new `str.prefixof` and `str.suffixof` operations follow the established patterns exactly — same factory helper (`binaryFuncUneval`), consistent ordering in all lists, correct SMT encoding, and matching DDM grammar/translate/format entries. The statistics bump from 284 to 286 is correct (two new factory ops). Editor syntax files are updated. No issues found.

Strata/Languages/Core/DDMTransform/FormatCore.lean
```diff
@@ -472,6 +472,8 @@ def handleBinaryOps {M} [Inhabited M] (name : String)
   -- String and Regex operations
   | .str .Concat => pure (.str_concat default arg1 arg2)
   | .str .InRegEx => pure (.str_inregex default arg1 arg2)
+  | .str .PrefixOf => pure (.str_prefixof default arg1 arg2)
+  | .str .SuffixOf => pure (.str_suffixof default arg1 arg2)
   | .re .Range => pure (.re_range default arg1 arg2)
```
shigoel, line 477: The CST→AST direction is handled here, but the reverse direction (`ASTtoCST.lean`) is missing the corresponding cases. `handleBinaryOps` in `ASTtoCST.lean` (~line 568) needs:
```lean
  | .str .PrefixOf => pure (.str_prefixof default arg1 arg2)
  | .str .SuffixOf => pure (.str_suffixof default arg1 arg2)
```
Without these, the AST→CST path falls through to `mkGenericCall` and produces incorrect output.

StrataTest/Languages/Core/Tests/StatisticsTest.lean
```diff
@@ -32,7 +32,7 @@ spec {
 #end
 
 /--
-info: [statistics] Evaluator.factoryOps: 284
+info: [statistics] Evaluator.factoryOps: 286
 [statistics] Evaluator.procedures: 1
```
shigoel, line 36: The existing test updates only verify the factory op count changed. Could you add a functional test that actually exercises `str.prefixof` / `str.suffixof`? For example:
- A Core program that calls `str.prefixof("ab", "abcd")` and checks the SMT encoding produces `(str.prefixof ...)`
- A DDM round-trip test that verifies AST→CST→AST preserves these operations

This would also catch the missing `ASTtoCST` case mentioned above.

shigoel (commented):
Nice clean addition — the pattern is followed consistently across most files. Two things to address:

1. **Bug: Missing `ASTtoCST.lean` cases.** `handleBinaryOps` in `ASTtoCST.lean` (around line 568) matches on binary `StrOpKind` cases but has no cases for `.str .PrefixOf` or `.str .SuffixOf`. These fall through to `handleBitvecBinaryOps` → `mkGenericCall`, producing a degraded free-variable reference instead of the proper `str_prefixof`/`str_suffixof` DDM constructors. The CST→AST direction (`FormatCore.lean`) is correctly updated, but the AST→CST direction is not.

2. **No functional tests.** The test updates only adjust expected factory listings and counts. Please add tests that exercise the new operations — e.g., an SMT encoding test that checks `str.prefixof`/`str.suffixof` appear in the generated SMT-LIB, or a DDM round-trip test.

Strata/Languages/Core/DDMTransform/FormatCore.lean
MikaelMayer: 🤖 Good catch on wanting to verify the round-trip! The `handleBinaryOps` function in `FormatCore.lean` (which `ASTtoCST.lean` imports) already has the `.str .PrefixOf` and `.str .SuffixOf` cases at lines 475–476, so the AST→CST direction was covered. To confirm this, I added an AST→CST round-trip test in `TestASTtoCST.lean` that parses a Core program using `str.prefixof`/`str.suffixof` and verifies the formatted output matches.

StrataTest/Languages/Core/Tests/StatisticsTest.lean
MikaelMayer: 🤖 Added two kinds of functional tests:

1. **CoreOpTests.lean**: Round-trip tests (`#guard checkRoundTrip "Str.PrefixOf"` and `"Str.SuffixOf"`) verifying `toString ∘ ofString = id`.
2. **TestASTtoCST.lean**: An AST→CST round-trip test with a Core program that uses `str.prefixof` and `str.suffixof` in requires/ensures/assert, verifying the full parse→format pipeline preserves these operations.

MikaelMayer:
🤖 Both items addressed:

1. **ASTtoCST cases**: The `handleBinaryOps` function in `FormatCore.lean` (imported by `ASTtoCST.lean`) already had `.str .PrefixOf` and `.str .SuffixOf` cases. Added an AST→CST round-trip test in `TestASTtoCST.lean` to confirm the full pipeline works.

2. **Functional tests**: Added `Str.PrefixOf`/`Str.SuffixOf` round-trip tests in `CoreOpTests.lean`, and a Core program test in `TestASTtoCST.lean` that uses `str.prefixof`/`str.suffixof` in requires, ensures, and assert statements.

Strata/Languages/Core/Factory.lean
```diff
@@ -325,6 +325,12 @@ def strToRegexFunc : WFLFunc CoreLParams :=
 def strInRegexFunc : WFLFunc CoreLParams :=
   binaryFuncUneval "Str.InRegEx" mty[string] mty[regex] mty[bool]
 
+def strPrefixOfFunc : WFLFunc CoreLParams :=
+  binaryFuncUneval "Str.PrefixOf" mty[string] mty[string] mty[bool]
```
shigoel: Both `strPrefixOfFunc` and `strSuffixOfFunc` use `binaryFuncUneval`, so they have no `concreteEval`. This means the evaluator can't simplify e.g. `str.prefixof("ab", "abcd")` to `true` on concrete inputs — everything is deferred to SMT.

Lean has `String.isPrefixOf` and `String.isSuffixOf`, so these could use `binaryOp` instead (like `Str.Concat` uses `binaryOp String.append` and `Str.Length` uses `unaryOp (fun s => Int.ofNat s.length)`). Worth considering whether concrete eval should be provided here.

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -554,6 +554,8 @@ partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Conte
     | .str .Substr   => .ok (.app Op.str_substr,    .string, ctx)
     | .str .ToRegEx  => .ok (.app Op.str_to_re,     .regex,  ctx)
     | .str .InRegEx  => .ok (.app Op.str_in_re,     .bool,   ctx)
+    | .str .PrefixOf => .ok (.app Op.str_prefixof,  .bool,   ctx)
+    | .str .SuffixOf => .ok (.app Op.str_suffixof,  .bool,   ctx)
     | .re .All       => .ok (.app Op.re_all,        .regex,  ctx)
```
shigoel, line 559: The SMT encoding looks correct, but there's no test that exercises it. Could you add a test that runs a Core program with `str.prefixof`/`str.suffixof` through the SMT encoder and checks that the generated SMT-LIB contains `(str.prefixof ...)` / `(str.suffixof ...)`? Ideally both a concrete-string case (e.g. `str.prefixof("ab", "abcd")` where the result is known) and a symbolic case.

shigoel (commented):
Two more items to consider on the new ops.

shigoel (commented):
Correction to my earlier comment on concrete eval.

Strata/Languages/Core/Factory.lean
shigoel: Correction to my earlier comment: `String.isPrefixOf` exists in Lean 4.29.1 and has the right argument order (`isPrefixOf s t` = "s is a prefix of t", matching `str.prefixof`), so `strPrefixOfFunc` could use `binaryOp String.isPrefixOf`.

However, `String.isSuffixOf` does **not** exist. The closest alternative is `String.endsWith`, but its argument order is reversed (`endsWith t s` = "t ends with s"), so `strSuffixOfFunc` would need `binaryOp (fun s t => t.endsWith s)` to match SMT-LIB's `(str.suffixof s t)` semantics.

MikaelMayer: 🤖 Changed `strPrefixOfFunc` to use `binaryOp "Str.PrefixOf" String.isPrefixOf` and `strSuffixOfFunc` to use `binaryOp "Str.SuffixOf" (fun s t => String.endsWith t s)`. Both now have concrete evaluation. `String.isSuffixOf` doesn't exist in this Lean version, so I used `String.endsWith` with flipped arguments instead.

MikaelMayer: 🤖 Addressed — switched both to `binaryOp` with concrete Lean functions as you suggested.

MikaelMayer:
🤖 Thanks for the correction on concrete eval — addressed by switching both `strPrefixOfFunc` and `strSuffixOfFunc` to `binaryOp` with concrete Lean implementations (`String.isPrefixOf` and `String.endsWith`).

Strata/Languages/Core/SMTEncoder.lean
MikaelMayer: 🤖 Added `#guard_msgs` tests in `SMTEncoderTests.lean` that verify `str.prefixof` and `str.suffixof` are correctly generated in the SMT output.

MikaelMayer: We are still missing the concrete string case that should be solved even if using the --no-solve option. That would help validating the implementation of these functions, especially the endsWith.

MikaelMayer: 🤖 Added concrete evaluation tests for `str.prefixof` and `str.suffixof` in `String.lean`. Four new assertions with literal strings test both true and false cases:
- `str.prefixof("abc", "abcdef")` → true
- `!str.prefixof("xyz", "abcdef")` → true
- `str.suffixof("def", "abcdef")` → true
- `!str.suffixof("xyz", "abcdef")` → true

All four VCs resolve to `Obligation: true` via concrete evaluation, confirming the `concreteEval` implementation works correctly without needing the SMT solver.

shigoel (approved):
All review comments have been addressed. Concrete eval via `binaryOp`, SMT encoder tests, DDM round-trip test, and CoreOp round-trip guards all look good. LGTM!

aqjune-aws (approved):
Thank you!

joscoh (approved):

PR #1084, opened 2026-04-30
Rewrite Core language reference and add transforms doc

shigoel:
## Summary

- Rewrites `LangDefDoc.lean` as a proper 5-section language reference: Introduction, Lambda (syntax + type system), Imperative (commands, statements, KleeneStmt, metadata), Strata Core (expressions, built-in types, built-in operators, type declarations, ADTs, functions, recursive functions, axioms, commands, procedures, programs), and Semantics (Lambda operational + denotational, command, statement, KleeneStmt).
- Adds `TransformsDoc.lean` documenting program transforms (`strata transform` CLI), analysis modes, and the SMT analysis pipeline.
- Adds missing docstrings to `CmdExt`, `CallArg`, `Decl`, `Program`, `LSort`, `StepKleene`, and `StepKleeneStar` in the source.
- Adds deprecation notes to `Datatypes.md`, `RecursiveFunctions.md`, and `VerificationModes.md` pointing to the canonical Verso docs.
- Cross-document links point to rendered HTML output rather than GitHub source files.


docs/verso/TransformsDoc.lean
```diff
@@ -0,0 +1,117 @@
...
+shortTitle := "Core Transforms and Analysis"
+%%%
+
+# Introduction
+
+This document describes the transforms and analyses available for Strata Core
+programs. Transforms are program-to-program rewrites that are independent of any
+specific analysis. Analyses consume a (possibly transformed) Core program and
+produce verification results. For the language definition itself, see the
+[Strata Core Language Definition](https://github.com/strata-org/Strata/tree/main/docs/verso/LangDefDoc.lean).
```
MikaelMayer: Nit: This links to the `.lean` source file on GitHub rather than the rendered HTML documentation. Consider linking to the rendered doc (e.g., the `langdef/html-single/` output) so readers land on the formatted page.

docs/verso/TransformsDoc.lean
```diff
@@ -0,0 +1,117 @@
...
+generated with the unevaluated LHS as a trigger pattern and the partially
+evaluated RHS as the body.
+
+## Axiom Encoding
+
+Axioms are emitted as universally quantified SMT assertions.
+
+# GOTO Analysis
+
+TODO: Document the GOTO analysis pipeline.
```
MikaelMayer: Nit: A bare `TODO` in a shipped doc page will be visible to readers. Consider either removing this section until content is ready, or adding a brief placeholder sentence (e.g., "Documentation forthcoming.") that reads better in the rendered output.

docs/verso/LangDefDoc.lean
```diff
@@ -157,153 +165,107 @@ Although {name LExpr}`LExpr` can be parameterized by an arbitrary type system,
...
+1. {name LContext}`LContext`: information that is typically constant during
+   expression type checking, but may be extended during statement type checking
+   (e.g., when a `funcDecl` statement adds a new function to the factory).
+   This includes information about built-in functions, using the
+   {name Lambda.Factory}`Factory` type, and built-in types, using the
+   {name TypeFactory}`TypeFactory` type. Built-in functions optionally include
+   concrete evaluation functions, which can be used in the semantics described
+   below.
 
 {docstring Lambda.LContext}
```
MikaelMayer: The `{docstring Lambda.LContext}` block between items `1.` and `2.` of this numbered list may break list continuity in the rendered Verso output, causing item 2 to restart numbering at 1. Worth checking the rendered HTML to confirm the list renders as intended.

shigoel:
@tautschnig We can document the GOTO transforms in the Transforms Doc., going forward.

docs/verso/LangDefDoc.lean
```diff
@@ -157,153 +165,107 @@ Although {name LExpr}`LExpr` can be parameterized by an arbitrary type system,
...
-## Structured Deterministic Statement Syntax
+## Structured Deterministic Statements
 
 Statements allow commands to be organized into standard control flow
 arrangements, including sequencing, alternation, and iteration. Sequencing
 statements occurs by grouping them into blocks. Loops can be annotated with
 optional invariants and decreasing measures, which can be used for deductive
-verification.
+verification. An `exit` statement transfers control out of the nearest
+enclosing block with a matching label. In addition, statements include
```
joscoh: As of https://github.com/strata-org/Strata/pull/1052, `exit`s also allow escaping one enclosing block without a label I believe.

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+Strata Core is the concrete verification language built by instantiating the
+generic `Lambda` and `Imperative` building blocks with specific types. This
+section describes the top-level constructs that make up a Strata Core program.
+
+## Expressions
+
+Strata Core expressions are `Lambda` expressions instantiated with a specific
+identifier type (`CoreIdent`) and a monomorphic type system (`LMonoTy`). A
+`CoreIdent` carries scope information, including whether it refers to a global
+variable, a local variable, or an `old` expression (denoting a pre-state value).
```
joscoh: There are no more global variables, right?

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+
+### Built-in Types
+
+Strata Core provides the following built-in types:
+
+- `bool` — booleans
+- `int` — unbounded integers
+- `real` — rationals (used to model reals)
```
joscoh, line 314: This is true currently, but we should discuss if this is the right decision long-term.

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+- *Boolean*: conjunction, disjunction, negation, implication, equivalence.
+- *Numeric (int and real)*: addition, subtraction, multiplication, negation,
+  division, modulus (Euclidean and truncating variants), comparisons
+  (`<`, `≤`, `>`, `≥`). Safe variants of division and modulus generate
+  precondition checks for division by zero.
+- *Bit vector*: arithmetic (add, sub, mul, div, mod — both signed and
+  unsigned), bitwise (and, or, xor, not, shifts), comparisons (signed and
+  unsigned), concatenation, and extraction of sub-ranges. Safe arithmetic
+  variants generate overflow precondition checks.
+- *String*: length, concatenation, substring extraction, conversion to regular
```
joscoh: Should this mention https://github.com/strata-org/Strata/pull/1083?

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+auxiliary functions:
+
+1. *Constructors*: functions that create values of the datatype (e.g.,
+   `None() : Option<T>`, `Cons(head: T, tail: List<T>) : List<T>`).
+2. *Tester functions*: for each constructor, a function that returns `true` if
+   a value was created with that constructor. The naming convention is
+   `<Datatype>..is<Constructor>` (e.g., `Option..isNone`, `List..isCons`).
+3. *Field accessors*: for each field, a function that extracts the field value.
+   The naming convention is `<Datatype>..<fieldName>` (e.g., `Option..val`,
+   `List..head`, `List..tail`). Field accessors are partial functions — calling
```
joscoh: More precisely, Strata generates safe and unsafe variants of field accessors. The default versions (e.g. `List..head`) add precondition checks requiring that the argument is an instance of the given constructor. The unsafe versions (e.g. `List..head!`) do not check this and have undefined behavior if called on the wrong constructor.

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+   them on the wrong constructor variant is undefined behavior.
+
+## Functions
+
+Strata Core functions are pure, named operations with typed input parameters
+and a return type. A function may have an optional body (a lambda expression);
```
joscoh: This is a bit confusing. Do you mean a lambda expression as an an expression in the Lambda dialect or that a function body is morally equivalent to a lambda (`fun _ => _`)?

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+if absent, it is uninterpreted. Functions may also have preconditions that
```
joscoh: Am I correct in assuming that `concreteEval` is therefore only used for built-ins and is not intended to be used by external clients of Strata? (I would be in favor of doing this explicitly and handling concrete evaluation separately from `func`, but that is a separate discussion)

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+A function with a body:
+
+```
+function Name<TypeArgs>(x₁ : T₁, ..., xₙ : Tₙ) : ReturnType
+{
+  body_expression
+};
+```
+
+A function may be prefixed with `inline` to request inlining.
```
joscoh: during partial evaluation

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+
+### Recursive Functions
+
+Both single and mutually recursive functions are supported. Recursive functions
+are declared with the `rec` keyword, and exactly one parameter must be annotated
```
joscoh: This will change soon, but maybe for now we should mention here explicitly that at the moment, only recursive functions over ADTs are supported.

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+{docstring Core.CmdExt}
+
+Each call argument specifies whether the corresponding parameter is passed by
+value (input), by mutable reference (input-output), or as an output-only
+variable.
+
+{docstring Core.CallArg}
+
+A Strata Core `Statement` is an `Imperative.Stmt` parameterized by Core's
+expression type and extended command type. Convenience abbreviations provide
```
joscoh: This is worded a bit awkwardly IMO. Maybe: Strata provides convenient abbreviations: Statement.init, etc

docs/verso/LangDefDoc.lean
```diff
@@ -351,21 +500,22 @@ name, type parameters, and input/output signatures.
...
-The ensures clauses cannot reason about the initial value of an out variable,
-but requires clauses can.
+1. *Input parameters* (`name : T`): Passed by value from the caller to the
+   callee. They are immutable within the procedure body.
+2. *Output parameters* (`out name : T`): Passed by value from the callee back
+   to the caller. They are mutable within the procedure body and their final
+   values are returned to the caller.
+3. *Input-output parameters* (`inout name : T`): Appear in both input and
+   output roles. The input value is the pre-state and the output value is the
+   post-state.
```
joscoh, line 528: It might be good to have a small example demonstrating the difference here, but not essential for this PR.

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
-(2) assert each non-free precondition with actuals substituted for formals,
-(3) assume each postcondition with actuals substituted for formals and `old v`
-bound to the pre-call value of `v`, and
-(4) update the caller's state.
+Note that `out` and `inout` keywords can only be attached when `eᵢ` is a variable.
+
+The semantics of a call are:
+
+1. Evaluate the argument expressions `e₁, ..., eₙ`.
+2. *Assert* each (non-free) precondition, substituting actuals for formals.
```
joscoh, line 573: Does the documentation mention what free preconditions are before this?

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
 
-## Body and verification
+### Body and verification
 
-If a procedure has a body, the preconditions are assumed, the body is
-symbolically executed, and the postconditions are asserted at the end. If the
-body is absent, the procedure is abstract and can only be reasoned about via its
+If a procedure has a body, the preconditions are assumed to hold on entry, the
+body is executed, and the postconditions must hold on exit. If the body is
+absent, the procedure is abstract and can only be reasoned about via its
```
joscoh: Is this true? I thought we did not allow empty procedure bodies, but my knowledge could be out of date.

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
+## Programs
+
+A Strata Core program is an ordered list of declarations. Each declaration is
+one of:
+
+- a type declaration (abstract type, type synonym, or algebraic datatype),
+- an axiom,
+- a `distinct` declaration (asserting that a list of expressions are pairwise distinct),
+- a procedure,
+- a function, or
```
joscoh: nit: a (non-recursive) function

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
+{docstring Lambda.LSort}
+
+{docstring Lambda.SortDenote}
+
+The functional denotation `LExpr.denote` interprets a well-typed annotated
+expression into a Lean value of the appropriate type. It is noncomputable
+(due to `Classical.propDecidable` in the `eq` and `quant` cases) and is used
+for reasoning, not computation.
+
+The relational predicate `Denotes` provides an equivalent specification that
```
joscoh: I don't think this is necessary in the documentation. The relational version is more of an implementation detail that lets us write some proofs a bit easier in my opinion.

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
+are mapped to their Lean counterparts, and all others are delegated to a
+user-provided type constructor interpretation
+({name Lambda.TyConstrInterp}`TyConstrInterp`).
+
+{docstring Lambda.LSort}
+
+{docstring Lambda.SortDenote}
+
+The functional denotation `LExpr.denote` interprets a well-typed annotated
+expression into a Lean value of the appropriate type. It is noncomputable
```
joscoh: An alternative:

The denotation function `LExpr.denote` interprets a well-typed annotated expression into a Lean value of the appropriate type. It is parameterized by interpretations for type constructors, operators, and free variables. Each Lambda construct is denoted into the corresponding Lean one; for example, an if-then-else becomes a Lean if-then-else, a `forall` quantifier becomes a Lean `forall`, and so on. Since Lambda allows unbounded quantification and equality over arbitrary types, this denotation can be used only for reasoning, not for computation. Validity of a Lambda expression means that the `LExpr.denote` evaluates to `true` under all possible interpretations.

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
+soundness and completeness of `Denotes` with respect to `denote`.
+
+### Consistency with Operational Semantics
+
+A key metatheoretic result is that the operational and denotational semantics
+agree. The theorem `Step.denote_preserved` states that a single evaluation step
+preserves the denotation of an expression. Its multi-step counterpart,
+`StepStar.denote_preserved`, extends this to arbitrary reduction sequences:
+if `e₁` reduces to `e₂` in zero or more steps, the denotations of `e₁` and
+`e₂` are equal. The theorem `eval_denote_sound` composes these results to show
```
joscoh: This is not exactly true. `eval_denote_sound` composes `StepStep.denote_preserved` and `eval_StepStar`, which shows that the partial evaluator is sound wrt `StepStar` (independent of the denotational semantics).

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
+The relational predicate `Denotes` provides an equivalent specification that
+avoids dependent-type casts, relating well-typed expressions directly to their
+semantic values. The `denote_Denotes` and `Denotes_denote` theorems establish
+soundness and completeness of `Denotes` with respect to `denote`.
+
+### Consistency with Operational Semantics
+
+A key metatheoretic result is that the operational and denotational semantics
+agree. The theorem `Step.denote_preserved` states that a single evaluation step
+preserves the denotation of an expression. Its multi-step counterpart,
```
joscoh: I would just say something like: `StepStar.denote_preserved` lifts this to the multi-step relation.

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
+
+## Command Semantics
+
+The semantics of commands are specified in terms of how they interact with a
+program state. An execution environment ({name Imperative.Env}`Env`) bundles
+three components:
+
+1. A *store* mapping variables to their current values.
+2. An expression *evaluator*.
+3. A cumulative *failure flag* that is OR-ed with per-command assertion failures.
```
joscoh: nit: maybe: "the disjunction of" rather than "OR-ed"?

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
+Given these state relations, the semantics of each command is specified in
+a standard way.
+
+{docstring Imperative.EvalCmd}
+
+## Structured Statement Semantics
+
+The semantics of the {name Stmt}`Stmt` type is defined in terms of
+*configurations*, represented by the {name Imperative.Config}`Config` type.
+A configuration tracks the current execution state: a single statement, a list
```
joscoh: I don't know that this list is needed when we have the docstrings for each case right below.

docs/verso/LangDefDoc.lean
```diff
@@ -393,32 +543,182 @@ both the pre-state (on entry) and the post-state (on exit) of a procedure
...
+
+The {name Imperative.StepKleene}`StepKleene` relation describes a single
+execution step for non-deterministic statements.
+
+{docstring Imperative.StepKleene}
+
+The {name Imperative.StepKleeneStar}`StepKleeneStar` relation is the reflexive,
+transitive closure.
+
+{docstring Imperative.StepKleeneStar}
```
joscoh, line 732: After this, should we mention the main definitions of program-wide semantics and soundness? E.g. `Triple` and `AllAssertsValid` in `Specification.lean`. These link the small-step semantics of each component with program-wide specifications

docs/verso/TransformsDoc.lean
```diff
@@ -0,0 +1,117 @@
...
+## Procedure Filtering (`filterProcedures`)
+
+Removes all procedures except those named in the `--procedures` flag (and any
+procedures they transitively depend on).
+
+## Irrelevant Axiom Removal (`removeIrrelevantAxioms`)
+
+Removes axioms that do not mention any of the functions named in the
+`--functions` flag, reducing the size of the analysis problem.
+
```
joscoh, line 69: Should we mention `PrecondElim.lean`, `DetToKleene.lean` `StructuredToUnstructured.lean`, etc? I guess I am asking if this list is supposed to be exhaustive or not.

docs/verso/TransformsDoc.lean
```diff
@@ -0,0 +1,117 @@
...
+encoded using the `declare-datatypes` command; the generated functions
+(constructors, testers, accessors) are mapped to the corresponding SMT
+functions (e.g., `Option..isNone` maps to `is-None`).
+
+## Function Encoding
+
+Functions with bodies are inlined by the partial evaluator where possible.
+Functions without bodies are declared as uninterpreted functions.
+
+Recursive functions are encoded as uninterpreted functions together with
```
joscoh: Maybe: Recursive functions are simplified by the partial evaluator but are encoded as uninterpreted functions with per-constructor axioms in the SMT encoding.

docs/verso/TransformsDoc.lean
```diff
@@ -0,0 +1,117 @@
...
+per-constructor axioms equipped with trigger patterns. For each constructor `C`
+of the ADT at the `@[cases]` parameter, a universally quantified axiom is
+generated with the unevaluated LHS as a trigger pattern and the partially
```
joscoh: The LHS/RHS explanation probably doesn't make sense in this context. I would say something like: For each constructor `C` of the ADT at the `@[cases]` parameter, the encoding generates an axiom representing the corresponding rewrite rule (e.g. `List.length Nil = 0` and `forall h t, List.length (Cons h t) = 1 + List.length t`).

docs/verso/LangDefDoc.lean
```diff
@@ -329,13 +292,199 @@ metadata elements.
...
+1. *Constructors*: functions that create values of the datatype (e.g.,
+   `None() : Option<T>`, `Cons(head: T, tail: List<T>) : List<T>`).
+2. *Tester functions*: for each constructor, a function that returns `true` if
+   a value was created with that constructor. The naming convention is
+   `<Datatype>..is<Constructor>` (e.g., `Option..isNone`, `List..isCons`).
+3. *Field accessors*: for each field, a function that extracts the field value.
+   The naming convention is `<Datatype>..<fieldName>` (e.g., `Option..val`,
+   `List..head`, `List..tail`). Field accessors are partial functions — calling
+   them on the wrong constructor variant is undefined behavior.
+
```
joscoh, line 409: We don't really mention typing much in the documentation, but I would say something about datatypes specifically, maybe:
Strata Core datatypes must satisfy several well-formedness properties: they must be inhabited (subject to a syntactic check that tries to determine a provably inhabited constructor), uniform (the type parameters cannot change through recursion), strictly positive (all recursive occurences of a type in a constructor do not appear to the left of any arrow), and non-nested (disallowing e.g. `datatype foo = | A (x: List foo)`).

MikaelMayer (commented):
Well-structured rewrite of the language reference. The reorganization into five clear sections (Introduction, Lambda, Imperative, Strata Core, Semantics) is a significant improvement. The new docstrings on source types are accurate and useful. The denotational semantics section and the Strata Core section (expressions, types, operators, ADTs, functions, axioms, procedures, programs) are thorough additions.

A few minor items below.

docs/verso/LangDefDoc.lean
shigoel: I'm a little worried about "undefined behavior", having battled that in previous settings. Perhaps out of scope of this PR, but do we have a plan to not have this at all?

joscoh: Maybe undefined behavior is not the right word, but semantically, the value of the function on the wrong constructors is an arbitrary value of the return type (guaranteed to exist by the fact that all Strata types are inhabited). Equivalently, they can be treated as uninterpreted functions in these cases. These unsafe destructors were requested by Laurel as a way to avoid lots of extra proof obligations in their encoding.

docs/verso/LangDefDoc.lean
```diff
@@ -635,25 +648,26 @@ user-provided type constructor interpretation
...
-preserves the denotation of an expression. Its multi-step counterpart,
-`StepStar.denote_preserved`, extends this to arbitrary reduction sequences:
-if `e₁` reduces to `e₂` in zero or more steps, the denotations of `e₁` and
-`e₂` are equal. The theorem `eval_denote_sound` composes these results to show
-that the evaluator is sound with respect to the denotational semantics.
+preserves the denotation of an expression. `StepStar.denote_preserved` lifts
+this to the multi-step relation. The theorem `eval_denote_sound` composes
+`StepStar.denote_preserved` and `eval_StepStar` (which shows that the partial
+evaluator is sound with respect to `StepStar`, independent of the denotational
+semantics) to show that the evaluator is sound with respect to the denotational
```
joscoh: Nit: this is phrased somewhat awkwardly, I would remove "independent of the denotational semantics". Although on second thought, maybe we don't mention anything about `eval_denote_sound` at all? This document doesn't say much about the partial evaluator, so it might be enough to just mention that we prove that the two versions of semantics are consistent. But I don't feel strongly ether way.

joscoh (dismissed):

docs/verso/TransformsDoc.lean
```diff
@@ -0,0 +1,121 @@
...
+shortTitle := "Core Transforms and Analysis"
+%%%
+
+# Introduction
+
+This document describes the transforms and analyses available for Strata Core
+programs. Transforms are program-to-program rewrites that are independent of any
+specific analysis. Analyses consume a (possibly transformed) Core program and
+produce verification results. For the language definition itself, see the
+[Strata Core Language Definition](https://github.com/strata-org/Strata/tree/main/docs/verso/LangDefDoc.lean).
```
MikaelMayer: 🤖 Updated the link to point to the rendered HTML documentation (`../langdef/html-single/`) instead of the `.lean` source file on GitHub.

MikaelMayer:
🤖 Removed the empty "GOTO Analysis" section entirely. It can be added back when content is ready.

docs/verso/LangDefDoc.lean
```diff
@@ -157,153 +165,107 @@ Although {name LExpr}`LExpr` can be parameterized by an arbitrary type system,
...
 
-The semantics of commands are specified in terms of how they interact with a
-program state, written `σ`. A state can be applied to a variable to obtain its
-current value. And an expression `e` can be evaluated using the evaluation
-function in a given state: `δ σ e` gives the result of evaluating `e` in state
-`σ`. This generic description allows the details of the program state
-representation to vary, as long as it supports these operations.
+{docstring Imperative.ExprOrNondet}
 
-Given a state `σ`, the {name InitState}`InitState` relation describes how a
```
MikaelMayer, line 257: 🤖 The `{docstring Lambda.LContext}` and `{docstring Lambda.TContext}` blocks are placed after item 2, not between items 1 and 2 — the numbered list (lines 190–199) completes before the docstrings appear (lines 202–204). List continuity should be preserved as-is.

joscoh (approved):

MikaelMayer (approved):

PR #1085, opened 2026-04-30
Fix self.field.method() calls bypassing dispatch resolution

### LINKED ISSUE #1082: Fix: self.field.method() calls bypass dispatch resolution inside class methods
### Summary

When a class method calls `self.field.method(...)` where `self.field` was initialized via dispatch (e.g. `self.store = servicelib.connect("storage")`), the call is not resolved through the pyspec dispatch. Instead, it's flattened to an underscore-separated name like `self_store_put_item` and treated as a package call, producing a Hole.

### Root cause

In `PythonToLaurel.lean`, `isPackage` checks if the root name of an expression is NOT in `ctx.variableTypes`. Inside method bodies, `self` is not added to `variableTypes`, so `isPackage` returns `true` for `self.field.method(...)`. This causes `refineFunctionCallExpr` to fall through to `pyExprToString` which flattens the dotted name with underscores, bypassing the dispatch resolution entirely.

### Fix

Add `self` to `variableTypes` with the class name in `translateFunctionBody` when `currentClassName` is set:

```lean
-- In translateFunctionBody, after createVarDeclStmtsAndCtx:
let ctx := match ctx.currentClassName with
  | some cn => {ctx with variableTypes := ("self", cn) :: ctx.variableTypes}
  | none => ctx
```

This 3-line change ensures `isPackage` returns `false` for `self.field` patterns, allowing `refineFunctionCallExpr` to resolve the field type through `classFieldHighType` and produce the correct `TypeName@method` name.

### Test

Add `test_self_field_dispatch_precondition.py` to `StrataTestExtra/Languages/Python/Specs/dispatch_test/`:

```python
import servicelib

class MyService:
    def __init__(self):
        self.store: Storage = servicelib.connect("storage")

    def save_empty_bucket(self) -> bool:
        self.store.put_item(Bucket="", Key="k", Data="v")
        return True

def trigger() -> bool:
    svc: MyService = MyService()
    return svc.save_empty_bucket()
```

Add an `#eval` test in `AnalyzeLaurelTest.lean` (following the `test_precondition_violation` pattern) that runs this file with `useRoots := true` and asserts that at least one VCResult has a `✖️` (always false) result for a `servicelib_Storage_` label.

**Without fix:** 2 passed, 0 failed (precondition violations not detected)
**With fix:** 4 passed, 2 failed (`Bucket must not be empty`, `Bucket must match ^[a-z0-9-]+$`)

MikaelMayer:
Fixes #1082

`self.field.method()` calls inside class methods were bypassing dispatch resolution. The type inference for `self` fell through to `Package`, so chained calls like `self.store.put_item(...)` were flattened into underscore-separated package calls instead of resolving through the field's type and dispatching to the correct spec.

The fix registers `self` with the current class name in `variableTypes` during method body translation, so `inferExprType` can resolve `self` → class → field type → correct method dispatch.

Tested with a new end-to-end test that verifies `self.store.put_item()` correctly dispatches through the Storage spec and detects a precondition violation. Existing tests pass.


MikaelMayer (commented):
Clean, minimal fix with a well-constructed end-to-end test. The 3-line change correctly registers `self` in `variableTypes` so that `inferExprType` resolves `self` to the class name instead of falling through to `Package`. The test follows the established pattern and exercises the exact bug path through the Storage spec's precondition.

tautschnig (approved):

joehendrix (approved):

PR #1090, opened 2026-04-30
Separate SMT error categories: encoding, timeout, crash

shigoel:
## Summary

Distinguishes three error types in `VCResult` instead of a single "Implementation Error" string:
- `VCError.encoding`: SMT encoding failure (Strata bug, exit code 3)
- `VCError.solverTimeout`: solver exceeded time limit (exit 0 — treated as inconclusive)
- `VCError.solverCrash`: solver process failed (exit code 3)

Timeout detection checks z3's `"timeout"` on stdout and cvc5's `"interrupted by timeout"` on stderr.

`pyAnalyzeLaurel` summary partitions timeouts separately from internal errors. `strata verify` command uses exit code 3 for internal errors (previously all non-success was exit 2). Solver invocation errors now return a `VCResult` instead of throwing `DiagnosticModel`, so remaining obligations continue processing.

### Exit code table

| Code | Meaning | When |
|------|---------|------|
| 0 | Success / Inconclusive | All assertions pass, some inconclusive, or solver timed out |
| 1 | User error | Bad input, unsupported flag |
| 2 | Failures found | Assertion violations |
| 3 | Internal error | SMT encoding failure or solver crash |
| 4 | Known limitation | Unsupported Python construct |


robin-aws:
Why do we need a separate exit code for this though? Especially given the definition of exit code 0 already allows for inconclusive results. Individual results will indicate individual timeouts in greater detail.

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -115,6 +115,25 @@ def runSolver (solver : String) (args : Array String) : IO IO.Process.Output :=
...
+  toString
+    | .timeout d => s!"Solver timeout: {d}"
+    | .crash d   => s!"{d}"
+
+/-- True when the word "timeout" appears as a whitespace-delimited token
+    in the solver's stdout or stderr. z3 emits "timeout" as a standalone
+    line on stdout; cvc5 prints "interrupted by timeout." on stderr.
+    Only called when verdict parsing has already failed. -/
+private def isTimeoutOutput (output : IO.Process.Output) : Bool :=
+  let hasWord (s : String) := s.splitOn " " |>.any fun tok =>
```
MikaelMayer: `splitOn " "` does not split on newlines. If stdout ever contains a preceding line (e.g. `"(error \"…\")\ntimeout\n"`), the `\n` ends up *inside* a token and `trimAscii` only strips the ends, so `"timeout"` is never matched.

Splitting on any whitespace would be more robust:
```lean
private def isTimeoutOutput (output : IO.Process.Output) : Bool :=
  let hasWord (s : String) := s.splitOn "\n" |>.any fun line =>
    line.splitOn " " |>.any fun tok =>
      tok.trimAscii.toString == "timeout" || tok.trimAscii.toString == "timeout."
  hasWord output.stdout || hasWord output.stderr
```
(Split on newlines first, then on spaces within each line.)

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -115,6 +115,25 @@ def runSolver (solver : String) (args : Array String) : IO IO.Process.Output :=
   --                         stdout: {repr output.stdout}"
   return output
 
+/-- Classifies the error when the solver fails to produce a valid verdict. -/
+inductive SolverError where
+  | timeout (detail : String)
+  | crash (detail : String)
+
+instance : ToString SolverError where
```
MikaelMayer, line 123: Nit: This `ToString SolverError` instance is never called — `getObligationResult` extracts the `detail` field via pattern matching. If it's kept for debugging, the `.crash` case should probably have a prefix like `.timeout` does (currently `.crash d` renders as just `d`, while `.timeout d` renders as `"Solver timeout: {d}"`).

Strata/Languages/Core/Verifier.lean
```diff
@@ -561,13 +561,26 @@ def LExprModel.format (model : LExprModel) : Format :=
...
 A collection of all information relevant to a verification condition's
 analysis.
 -/
 structure VCResult where
   obligation : Imperative.ProofObligation Expression
-  outcome : Except String VCOutcome := .error "not yet computed"
+  outcome : Except VCError VCOutcome := .error (.encoding "not yet computed")
   estate : EncoderState := EncoderState.init
   verbose : VerboseMode := .normal
   checkLevel : CheckLevel := .minimal
```
MikaelMayer, line 613: Using `.encoding "not yet computed"` as the default `outcome` means an uninitialized `VCResult` is indistinguishable from a real encoding error. Consider a dedicated `VCError` constructor (e.g. `.uninitialized`) or making the default `outcome` an `Option` / removing the default altogether so the caller must supply it.

shigoel:
> Why do we need a separate exit code for this though? Especially given the definition of exit code 0 already allows for inconclusive results. Individual results will indicate individual timeouts in greater detail.

That's a very good point! Will do.

MikaelMayer (commented):
Clean separation of error categories. The `SolverError` / `VCError` split, exit code table, and summary partitioning all look correct. Two items worth addressing below.

StrataMain.lean
```diff
@@ -1307,8 +1320,13 @@ def verifyCommand : Command where
       else
         let provedGoalCount := (vcResults.filter Core.VCResult.isSuccess).size
         let failedGoalCount := (vcResults.filter Core.VCResult.isNotSuccess).size
+        let hasImplError := vcResults.any (fun r => r.isImplementationError || r.hasSMTError)
```
aqjune-aws, line 1324: Could you add a short comment what `r.hasSMTError` does & why this is counted as hasImplError please?
Same for hasFailure below.

Strata/DL/Imperative/SMTUtils.lean
MikaelMayer: 🤖 This is already implemented — `isTimeoutOutput` splits on `"\n"` first, then on `" "` within each line, exactly as suggested.

Strata/DL/Imperative/SMTUtils.lean
MikaelMayer: 🤖 Already addressed in a previous commit — the `.crash` case now renders as `"Solver crash: {d}"`, consistent with the `.timeout` prefix.

Strata/Languages/Core/Verifier.lean
MikaelMayer: 🤖 Already addressed — the `outcome` field no longer has a default value, so callers must supply it explicitly. The `.encoding "not yet computed"` default has been removed.

MikaelMayer:
🤖 Agreed — solver timeouts already exit 0 (same as inconclusive), so no separate exit code is needed. The implementation already handles this correctly: timeouts fall through to exit 0 in both `printPyAnalyzeSummary` and the `verify` command. Will update the PR description to avoid any confusion.

StrataMain.lean
MikaelMayer: 🤖 Added comments explaining both:
- `hasImplError`: encoding failures, solver crashes, or per-check SMT errors (exit 3)
- `hasFailure`: assertion violations that are not timeouts or internal errors (exit 2)

tautschnig (approved):

aqjune-aws (approved):

PR #1091, opened 2026-04-30
Eliminate redundant exception checks in try/except bodies

tautschnig:
Only insert isError(maybe_except) checks after statements that actually modify maybe_except (procedure calls that may throw). Previously, every statement in a try body got an isError check, creating massive ite nesting in verification conditions.

For a chosen benchmark this reduces 62 down to 8 isError checks (87% reduction).


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1788,13 +1788,29 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
         handlerCtx := hCtx
         handlerStmts := handlerStmts ++ hStmts
 
-    -- Insert exception checks after each statement in try body
+    -- Insert exception checks only after statements that could modify
+    -- maybe_except (procedure calls that may throw). Consecutive checks
+    -- without intervening modifications are redundant and create unnecessary
+    -- ite branches in the verification conditions.
+    let modifiesMaybeExcept (stmt : StmtExprMd) : Bool :=
```
MikaelMayer: The target-matching predicate (`targets.any fun t => match t.val with | .Identifier n => n == "maybe_except" | _ => false`) is duplicated verbatim between the top-level `.Assign` case and the `.Block` inner `.Assign` case. Extracting it (e.g., `let assignsMaybeExcept targets := targets.any fun t => ...`) would reduce duplication and make the recursive version suggested in the other review comment easier to build on.

MikaelMayer (commented):
The optimization idea is sound: skipping `isError(maybe_except)` checks after statements that provably don't modify `maybe_except` is a valid way to reduce ite nesting. The existing review comments about the shallow traversal in `modifiesMaybeExcept` (missing writes inside `IfThenElse`, `While`, nested `try`, `With`, etc.) are the main correctness concern and I won't repeat them.

One additional observation: the existing test suite has no try-body that contains a modeled procedure call (one that actually sets `maybe_except`). Every current try-body test uses only simple assignments or `raise`. This means the optimization's core behavior — inserting `isError` checks selectively after throwing calls — is exercised by zero tests, not just the nested-construct edge case.

thanhnguyen-aws (approved):

thanhnguyen-aws (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
tautschnig: Addressed in [439a3cb](https://github.com/strata-org/Strata/pull/1091/commits/439a3cbae44bc53b251a7f08abfa2600bb3ca9e0).

thanhnguyen-aws (dismissed):

tautschnig:

MikaelMayer (approved):

aqjune-aws (approved):

PR #1092, opened 2026-05-01
Add termination checking for recursive functions

joscoh:
*Issue #, if available:*

*Description of changes:*                                                                                                                              
                                                                                                                                                
  - Adds a new `TerminationCheck` pipeline phase that generates `$$term` verification procedures for `recFuncBlock` declarations, asserting that
   an `adtRank` measure strictly decreases at each recursive call site                                                                          
  - Generates `D..adtRank : D → Int` uninterpreted functions and per-constructor axioms for ADTs used as decreasing measures, supporting mutual 
  datatypes and polymorphic datatypes instantiated at concrete types. These axioms are added as preconditions to the generated termination-checking procedures to avoid extra axioms in other contexts.       
- Adds `decreases` syntax for functions in Core, where the `decreases` is inferred from the `@[cases]` if not provided.
- Fixes two issues caught by termination checking: `List_get_non_neg`/`List_set_non_neg` in the Python runtime were calling the wrong sibling functions (the existing mechanism was probably fine but the termination argument is needlessly complex), and `nth` in tests was missing an `isNil` base case guard. 


Note that `PrecondElim` is not run on these generated procedures, since this will duplicate the precondition assertions from the original program.
For the moment, only functions are checked for termination. Procedure termination checking will be added in a future PR.
                                                                                                                                          
  ## Tests                                                                                                                                
                                                                                                                                                
  -`StrataTest.DL.Lambda.AdtRankAxiomsTests` — unit tests for `adtRank` axiom generation (IntList, MyNat, RoseTree/RoseList mutual,  
  no-recursive-field, polymorphic)                                                                                                              
  -`StrataTest.Languages.Core.Tests.TerminationCheckTests` — tests covering termination checking, including standard recursion, mutual recursion, error cases, preconditions, polymorphic datatypes, etc.                                      



Strata/DL/Util/Func.lean
```diff
@@ -80,6 +80,7 @@ structure Func (IdentT : Type) (ExprT : Type) (TyT : Type) (MetadataT : Type) wh
   concreteEval : Option (MetadataT → List ExprT → Option ExprT) := .none
   axioms   : List ExprT := []  -- For axiomatic definitions
   preconditions : List (FuncPrecondition ExprT MetadataT) := []
+  measure  : Option ExprT := .none
```
MikaelMayer: Nit: The other fields have inline comments (e.g. `axioms` has `-- For axiomatic definitions`). A brief comment like `-- Termination measure expression (from `decreases` clause)` would be consistent.

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,321 @@
...
+
+/-- Find the decreasing parameter index for a function: explicit `measure`
+    (from `decreases` clause), or fallback to `@[cases]` (`inlineIfConstr`). -/
+private def getDecreasesIdx (func : Function) : Option Nat :=
+  match func.measure with
+  | some (.fvar _ id _) => func.inputs.keys.findIdx? (· == id)
+  | some _ => none
+  | none => FuncAttr.findInlineIfConstr func.attr
+
+/-- Extract the datatype name from a monomorphic type. -/
```
MikaelMayer: Nit: `adtNameOf` silently returns `""` for non-ADT types, which would produce a bogus `"..adtRank"` function name if it ever reached `mkAdtRankLt` or `mkAdtRankDecls`. The error checking in Step 1 of `termCheck` prevents this in practice, but an `Option String` return (or a `panic!`) would make the invariant explicit and prevent silent corruption if the precondition is ever violated.

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,321 @@
...
+        for func in funcs do
+          if func.typeArgs.isEmpty then
+            match getDecreasesIdx func with
+            | none =>
+              throw (s!"recursive function '{func.name.name}' requires a 'decreases' clause or a '@[cases]' parameter for termination checking")
+            | some idx =>
+              match func.inputs.values[idx]? with
+              | some (.tcons n _) =>
+                if (tf.getType n).isNone then
+                  throw (s!"recursive function '{func.name.name}': decreasing parameter type '{n}' is not a known datatype")
```
MikaelMayer: `collectAdtRank` is called twice — once for `newAdtNames` and once for `allAdtNames` — and `allAdtNames` is a superset of `newAdtNames`. This means `mkAdtRankDecls` is called again for the datatypes already processed in the first call. Since `mkAdtRankDecls` does a linear scan of `tf.toList`, this is O(n²) in the number of datatypes per block. Not a problem at current scale, but if this ever matters, you could compute `allAdtRank` once and derive `newAdtRank` by filtering its `funcDecls`.

MikaelMayer (commented):
Solid PR. The architecture is clean — `AdtRankAxioms` for axiom generation, `TerminationCheck` for the pipeline phase, with good separation of concerns. The `wrapImplications` refactoring to share code between PrecondElim and TermCheck is well done. The `List_get_non_neg`/`List_set_non_neg` bug fixes are genuine catches. Test coverage is comprehensive (17 tests covering basic, mutual, polymorphic, error cases, preconditions, non-termination detection). Documentation is thorough and well-written.

Strata/Transform/TerminationCheck.lean
joscoh: I don't think it should `panic`. It is possible to change to an `Option` but it doesn't actually help all that much: we still need a default case for the call sites (`mkAdtRankLt` and `extractTermObligations`). Since the error checking catches this already I didn't think it was worth the additional complication.

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,318 @@
...
+            else []
+          | _ => []
+        argObs ++ callObs
+    | .eq _ e1 e2 => go e1 implications ++ go e2 implications
+    | .abs _ _ _ body => go body implications
+    | .quant _ _ _ _ trigger body =>
+      go trigger implications ++ go body implications
+    | _ => []
+  termination_by e.sizeOf
+  decreasing_by
```
tautschnig, line 146: `extractTermObligations` recurses into `ite` branches, `eq`, `abs`, `quant`, and function applications (via `getLFuncCall`). However, it does not recurse into the body of a let-binding.

In the locally nameless representation, a let-binding `let x = e1 in e2` is encoded as `.app _ (.abs _ _ _ e2) e1`. When `getLFuncCall` encounters this, it returns `(.abs _ _ _ e2, [e1])`. The `argObs` recurses into `e1`, but `e2` (the let body) is never visited — the `.abs` case at line 113 only fires for standalone `abs` nodes, not ones inside `.app`.

This means a recursive call inside a let body would be missed:
```
rec function f(@[cases] xs : IntList) : int {
  let y = g(xs) in f(IntList..tl(xs)) + y
}
```

In practice this doesn't matter today because Core functions don't use let bindings, but it's worth a comment documenting this limitation so future maintainers don't assume full coverage:

```suggestion
    | .app _ _ _ =>
      -- Note: let-bindings (.app _ (.abs _ _ _ body) arg) are decomposed by
      -- getLFuncCall as (abs, [arg]). The abs body is NOT recursed into.
      -- This is acceptable because Core recursive functions currently don't
      -- use let-bindings. If that changes, this case needs updating.
      match _h : getLFuncCall e with
```

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,318 @@
...
+      | .recFuncBlock funcs md => do
+        -- Step 1: error checking
+        -- Skip polymorphic functions: adtRank axioms are monomorphic, so we
+        -- cannot generate them for polymorphic datatypes yet. The user-facing
+        -- error is in Env.addFactoryFunc; when that restriction is lifted,
+        -- this filter must be updated to handle polymorphic adtRank generation.
+        let fileRange := Imperative.getFileRange md |>.getD FileRange.unknown
+        let throwErr (msg : String) : CoreTransformM Unit :=
+          throw (DiagnosticModel.withRange fileRange msg)
+        for func in funcs do
```
tautschnig, line 276: The error checking in Step 1 throws on the first error it finds. For a `recFuncBlock` with multiple functions, if the first function has a valid `decreases` but the second doesn't, only the second's error is reported. This is fine for single errors, but if multiple functions in the same block have issues, the user has to fix them one at a time.

Consider collecting all errors and throwing them together, or at minimum adding a comment noting the fail-fast behavior.

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,318 @@
...
+
+/-- Add a termination-check procedure as a leaf node in the cached call graph. -/
+private def addTermProcToCallGraph (name : String) : CoreTransformM Unit :=
+  modify fun σ => match σ.cachedAnalyses.callGraph with
+  | .some cg => { σ with cachedAnalyses := { σ.cachedAnalyses with
+      callGraph := .some (cg.addLeafNode name) } }
+  | .none => σ
+
+/-- Result of generating adtRank declarations for a mutual datatype block. -/
+private structure AdtRankDecls where
```
tautschnig: `mkTermCheckProc` embeds ALL adtRank axioms as preconditions of the generated `$$term` procedure, even axioms for datatypes not used by this particular function. For a mutual block with functions over different datatypes, each `$$term` proc gets axioms for all datatypes in the block.

This is sound (extra axioms don't hurt) but could make the SMT queries larger than necessary. For small programs this is fine, but for large mutual blocks with many datatypes, filtering axioms to only those relevant to the function's decreasing parameter type would reduce VC size.

Strata/DL/Lambda/AdtRankAxioms.lean
```diff
@@ -0,0 +1,109 @@
...
+          | [] => ltExpr
+          | ty :: rest =>
+            let inner := .quant m .all "" (.some ty) adtRankConstr ltExpr
+            rest.foldl (fun body ty =>
+              .quant m .all "" (.some ty) (LExpr.noTrigger m) body
+            ) inner
+        axiom_ :: acc
+    ).reverse
+
+/-- Generate the non-negativity axiom `∀ x. adtRank(x) >= 0` for a datatype.
```
tautschnig, line 80: The axiom generation is well-structured. A few observations:

1. **Proof opportunity**: The key soundness property is that `adtRank` is well-founded — i.e., there is no infinite descending chain. This follows from the fact that every constructor's recursive fields have strictly smaller rank. A formal theorem stating this would strengthen confidence:

```
-- Informal: For any ADT D with constructors C_i, if field f_j of C_i has type D' in the mutual block,
-- then the generated axiom `D'..adtRank(f_j) < D..adtRank(C_i(fields...))` is consistent
-- (i.e., there exists a model satisfying all axioms simultaneously).
```

This could be proved by exhibiting a concrete model: define `adtRank` as the structural depth of the ADT value.

2. **Non-negativity axiom**: `mkAdtRankNonNegAxiom` generates `∀ x. adtRank(x) >= 0`. Combined with the per-constructor axioms, this ensures `adtRank` is a well-founded measure on natural numbers. Good.

3. **Trigger placement**: The per-constructor axioms use `adtRank(C(fields...))` as the trigger, which means they only fire when the SMT solver encounters an `adtRank` application on a constructor term. This is the right choice — it avoids matching loops while ensuring the axioms fire when needed.

Strata/Transform/PrecondElim.lean
```diff
@@ -376,6 +377,10 @@ where
     | d :: rest =>
       match d with
       | .proc proc md => do
+        if TermCheck.isTermProc proc.header.name.name then
+          let (changed, rest') ← transformDecls rest
+          return (changed, d :: rest')
+        else
         let F ← getFactory
```
tautschnig, line 384: Good — `$$term` procedures are skipped by `PrecondElim` to avoid duplicating precondition assertions. The check uses `TermCheck.isTermProc` which tests for the `$$term` suffix.

However, this creates a coupling between `PrecondElim` and `TerminationCheck` — `PrecondElim` now imports `TerminationCheck`. If more pipeline phases need to skip `$$term` procedures in the future, consider adding a general `skipPrecondElim` flag to the procedure header (similar to `noFilter`) instead of hard-coding the name check.

StrataTest/Languages/Core/Tests/TerminationCheckTests.lean
```diff
@@ -0,0 +1,952 @@
...
+
+/-!
+# Termination Checking Tests
+
+Tests termination checking for recursive functions over algebraic datatypes.
+-/
+
+namespace Strata.TerminationCheckTest
+
+---------------------------------------------------------------------
```
tautschnig, line 17: Consider adding a test for a function where the recursive call is nested inside a non-recursive function call, e.g.:
```
rec function f(@[cases] xs : IntList) : int {
  if IntList..isNil(xs) then 0 else g(f(IntList..tl(xs)))
}
```
where `g` is a non-recursive function. This tests that `extractTermObligations` correctly recurses into arguments of non-recursive calls.

Strata/Transform/TerminationCheck.lean
tautschnig: How about:
```suggestion
/-- Extract the datatype name from a monomorphic type.
    Precondition: `ty` must be a `.tcons` (enforced by error checking in `termCheck`). -/
private def adtNameOf (ty : LMonoTy) : String :=
```

Strata/DL/Lambda/AdtRankAxioms.lean
```diff
@@ -0,0 +1,109 @@
...
+    Trigger: `adtRank(x)` — fires only when adtRank terms already exist. -/
+def mkAdtRankNonNegAxiom {T : LExprParams} [Inhabited T.Metadata] [Inhabited T.IDMeta]
+    (dt : LDatatype T.IDMeta) (m : T.Metadata) : LExpr T.mono :=
+  let dtTy := dataDefault dt
+  let rankTy := .arrow dtTy .int
+  let rankX := .app m (.op m (adtRankFuncName dt.name) (.some rankTy))
+    (.bvar m 0)
+  let geTy := .arrow .int (.arrow .int .bool)
+  let geExpr := .mkApp m (.op m "Int.Ge" (.some geTy)) [rankX, .intConst m 0]
```
aqjune-aws: Can we use `(@intGeFunc T).opExpr` which will hide hard-coded `"Int.Ge"` and `geTy`?
```
let geExpr := .mkApp m (@intGeFunc T).opExpr [rankX, .intConst m 0]
```

Strata/DL/Lambda/AdtRankAxioms.lean
```diff
@@ -0,0 +1,109 @@
...
+    (c.args.foldlIdx (init := []) fun acc i (_, fieldTy) =>
+      match block.find? (fun d => fieldTy == dataDefault d) with
+      | none => acc
+      | some fieldDt =>
+        let fieldBvar := .bvar m (numFields - 1 - i)
+        let fieldRankTy := .arrow (dataDefault fieldDt) .int
+        let adtRankField := .app m (.op m (adtRankFuncName fieldDt.name)
+          (.some fieldRankTy)) fieldBvar
+        let ltTy := .arrow .int (.arrow .int .bool)
+        let ltExpr := LExpr.mkApp m (.op m "Int.Lt" (.some ltTy))
```
aqjune-aws: Similarly to line 89, let's use
```
let ltExpr := .mkApp m (@intLtFunc T).opExpr [adtRankField, adtRankConstr]
```

docs/verso/LangDefDoc.lean
```diff
@@ -437,7 +437,8 @@ evaluation transform, when applied.
 Currently, only recursive functions over algebraic datatypes are supported (both
 single and mutually recursive). Recursive functions are declared with the `rec`
 keyword, and exactly one parameter must be annotated with `@[cases]` to indicate
-the algebraic datatype argument being recursed on.
+the algebraic datatype argument for per-constructor axiom generation.
```
aqjune-aws: Are the generated axioms describing the recursive function unfolded with respect to each constructor? Let's clarify this.

docs/verso/LangDefDoc.lean
```diff
@@ -446,6 +447,21 @@ rec function listLen (@[cases] xs : IntList) : int
...
+```
+rec function zipLen (@[cases] xs : IntList, ys : IntList) : int
+  decreases ys
+{
+  if IntList..isNil(xs) then 0
+  else if IntList..isNil(ys) then 0
+  else 1 + zipLen(IntList..tl(xs), IntList..tl(ys))
+};
+```
+
```
aqjune-aws, line 466: Can a recursive function have neither `@[cases]` nor `decreasing` (which will make termination checking impossible)? My understanding is that this is not true because exactly one argument must be annotated with `@[cases]` as mentioned in the previous paragraph. Let's state that as a result Strata doesn't accept a recursive function without termination hint.

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -382,8 +382,9 @@ op recfn_decl (name : Ident,
                @[scope(typeArgs)] b : Bindings,
                @[scope(typeArgs)] r : Type,
                @[scope(b)] preconds : SpacePrefixSepBy SpecElt,
+               @[scope(b)] decreases : Option Measure,
```
aqjune-aws, line 385: Does it accept arbitrary expression? If it is targeting a free variable in this patch for now, what about limiting it to an identifier and gradually extending to arbitrary expressions in the future?

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
```diff
@@ -353,7 +353,7 @@ function List_get_non_neg (l : ListAny, i : int) : Any
 {
   if ListAny..isListAny_nil(l) then from_None()
   else if  i == 0 then ListAny..head!(l)
-  else List_get(ListAny..tail!(l), i - 1)
+  else List_get_non_neg(ListAny..tail!(l), i - 1)
```
aqjune-aws, line 356: Do you have a behind story about this finding & update? :)

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,318 @@
...
+                  [wrapImplications implications ltExpr]
+                | none => []
+              | none => []
+            else []
+          | _ => []
+        argObs ++ callObs
+    | .eq _ e1 e2 => go e1 implications ++ go e2 implications
+    | .abs _ _ _ body => go body implications
+    | .quant _ _ _ _ trigger body =>
+      go trigger implications ++ go body implications
```
aqjune-aws: It seems `.abs` and `.quant` cases can be slightly complicated because visiting their bodies will have to consider bound variables. What about throwing when meeting this kind of case?

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,318 @@
...
+def termProcName (name : String) : String := s!"{name}{termSuffix}"
+
+def isTermProc (name : String) : Bool := name.endsWith termSuffix
+
+/-- Find the decreasing parameter index for a function: explicit `measure`
+    (from `decreases` clause), or fallback to `@[cases]` (`inlineIfConstr`). -/
+private def getDecreasesIdx (func : Function) : Option Nat :=
+  match func.measure with
+  | some (.fvar _ id _) => func.inputs.keys.findIdx? (· == id)
+  | some _ => none
```
atomb, line 54: If a user writes `decreases someComplexExpr` (not a bare variable), this silently falls through to `none`, which then triggers the error "requires a 'decreases' clause or a '@[cases]' parameter". The error message is misleading — the user *did* provide a `decreases` clause, it just wasn't a simple variable reference. A more specific error message like "decreases expression must be a parameter name" would be clearer.

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,318 @@
...
+    (func : Function)
+    (recFuncNames : List String)
+    (funcDecreasesMap : List (String × Nat × LMonoTy))
+    (adtRankAxioms : List (String × Expression.Expr))
+    (tf : @TypeFactory Unit)
+    (md : Imperative.MetaData Expression)
+    : Option (Decl × Nat) := do
+  let decrIdx ← getDecreasesIdx func
+  let inputTys := func.inputs.values
+  let decrTy ← inputTys[decrIdx]?
```
atomb: If `getDecreasesIdx` returns an index that's valid for `func.inputs.keys` but somehow not for `func.inputs.values` (shouldn't happen with a well-formed `ListMap`, but defensive), the function returns `none` and no termination check is generated. The `do` notation with `←` makes this silent.

Strata/DL/Lambda/AdtRankAxioms.lean
```diff
@@ -0,0 +1,109 @@
...
+          [adtRankField, adtRankConstr]
+        let axiom_ := match fieldTys with
+          | [] => ltExpr
+          | ty :: rest =>
+            let inner := .quant m .all "" (.some ty) adtRankConstr ltExpr
+            rest.foldl (fun body ty =>
+              .quant m .all "" (.some ty) (LExpr.noTrigger m) body
```
atomb: The bound variable name is `""` (empty string). While this works with de Bruijn indices, it makes debugging SMT output harder. The generated SMT uses `__q0`, `__q1` etc. which is fine, but if the pretty-printer ever uses these names, they'd be unhelpful.

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -382,8 +382,9 @@ op recfn_decl (name : Ident,
...
                @[scope(b)] c : r) : RecFnDecl =>
-  "function " name typeArgs b " : " r indent(2, preconds) "\n{\n  " indent(2, c) "\n}";
+  "function " name typeArgs b " : " r indent(2, preconds) "\n" indent(2, decreases) "{\n  " indent(2, c) "\n}";
```
atomb, line 387: When `decreases` is `none`, this produces an extra newline before `{`. This is cosmetic and may be intentional for readability, but differs from how `preconds` handles the empty case (no extra whitespace when empty).

atomb (commented):
I still want to do a more thorough review myself, but here are a few 🤖 notes so far.

Strata/Languages/Python/PythonRuntimeLaurelPart.lean
joscoh: Not really, only that the termination check failed :) I think that the behavior is equivalent for both (since the sign cannot flip until it reaches zero), but the behavior is much more convoluted when they are making mutually recursive calls. In any case, I believe that this fix is the intended behavior of the `nth` function.

Strata/Transform/TerminationCheck.lean
joscoh: A `ListMap` is a pair of elements, so there is no way to have such an index.

Strata/Transform/TerminationCheck.lean
joscoh: Added comment

Strata/Transform/TerminationCheck.lean
joscoh: Added handling for (some) let-bindings, as long as the bound variable is not the decreasing argument to the recursive call. Added a test for this.

Strata/Transform/TerminationCheck.lean
joscoh: I think it is better to throw the first error. If any error is hit, this definition is ill-formed, so subsequent definitions depend on a nonexistent definition, which could lead to spurious error reporting. I added a comment noting the fail-fast behavior.

Strata/Transform/TerminationCheck.lean
joscoh: Added axiom filtering

Strata/DL/Lambda/AdtRankAxioms.lean
joscoh: 1. Once we complete the denotational semantics for datatypes and add a generalized induction principle, I agree that this should be done. But this ultimately still means that we trust that the ADT axiomatization is correct (including the induction principle). It is possible but very, very difficult to give models for these axiomatizations in general (this was a significant part of my PhD :)).

Strata/Transform/PrecondElim.lean
joscoh: Yes, we could do that, but it would be a global change to `LFunc` which would only be needed for the specific case of the termination predicates. So I felt this was significantly simpler.

StrataTest/Languages/Core/Tests/TerminationCheckTests.lean
joscoh: Added, thanks

docs/verso/LangDefDoc.lean
joscoh: I added a sentence to the docs that explains that this case is an error.

Strata/Languages/Core/DDMTransform/Grammar.lean
joscoh: The next PR will involve other `decreases` expressions, so I thought it didn't make sense to introduce it and then immediately change it.

Strata/Transform/TerminationCheck.lean
joscoh: Added support for going inside bodies but throwing an exception if we ever have a termination obligation where the decreasing variable is bound. Added some test cases for this as well.

Strata/DL/Lambda/AdtRankAxioms.lean
joscoh: Do you have any ideas for reasonable names?

Strata/Languages/Core/DDMTransform/Grammar.lean
joscoh: Unfortunately this seems difficult to change without messing up `decreases` clauses for while loops, which are formatted differently. We could create separate `decreases` concrete syntax for each use case if needed, let me know if you think that would be better.

Strata/Languages/Core/DDMTransform/Grammar.lean
MikaelMayer: Can these expressions return other datatypes? I was thinking of the case where sometimes we have constructor inputs like Hello(x, y) and the recursion is Hello(x  y-1) which would not be provable to terminate, but if we transformed it into Succ(...Hello'(x)) then we can show that it decreases.

MikaelMayer (commented):
I still feel uncomfortable that the termination proof is emitted as a procedure because all functions can be included in the context of proving procedures, including the function for which the procedure is supposed to prove the termination of.
I know you mitigated this by checking that the termination proof should not depend on the function (haven't seen the code precisely for that though) but that means the soundness proof is going to be somewhat difficult since we will need the hypothesis that the proof does not depend on the function being declared as a function.

We also discussed how a possible long term solution would be to let people define recursive procedures instead and have a phase that can convert these into recursive functions, but we would still need to put a dependency between the function declaration and the proof.

MikaelMayer (dismissed):

Strata/DL/Lambda/AdtRankAxioms.lean
```diff
@@ -0,0 +1,108 @@
...
+          (.some fieldRankTy)) fieldBvar
+        let ltExpr := LExpr.mkApp m (@intLtFunc T).opExpr
+          [adtRankField, adtRankConstr]
+        let axiom_ := match fieldTys with
+          | [] => ltExpr
+          | ty :: rest =>
+            let inner := .quant m .all "" (.some ty) adtRankConstr ltExpr
+            rest.foldl (fun body ty =>
+              .quant m .all "" (.some ty) (LExpr.noTrigger m) body
+            ) inner
```
tautschnig: **Empty bvar names (atomb's follow-up question).** These quantifier binders all get `prettyName := ""`. De Bruijn makes the semantics fine, but it hurts SMT pretty-printing and cvc5/z3 error traces. Concrete suggestion — bind the quantified variable to the field's parameter name from the constructor signature:

```suggestion
        let axiom_ :=
          let fieldNames := (c.args.map (·.fst.name)).reverse
          match fieldTys, fieldNames with
          | [], _ => ltExpr
          | ty :: restTys, name :: restNames =>
            let inner := .quant m .all name (.some ty) adtRankConstr ltExpr
            (restTys.zip restNames).foldl (fun body (ty, name) =>
              .quant m .all name (.some ty) (LExpr.noTrigger m) body
            ) inner
          | _ :: _, [] =>
            -- unreachable: fieldNames and fieldTys come from the same `c.args`
            ltExpr
        axiom_ :: acc
```

The analogous change in `mkAdtRankNonNegAxiom` (line 89) would use `"x"` (matching the function's own input name at line 32 — already `"x"`) so the SMT output reads `∀ x : IntList. IntList..adtRank(x) >= 0` rather than `∀ __q0 : IntList. …`.

Minor — no soundness impact, purely an ergonomics win when someone has to read generated SMT during debugging.

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,357 @@
...
+    E.g., for `MyList` with `typeArgs = ["a"]` and concrete type `MyList int`,
+    produces `[("a", int)]`. -/
+private def mkTySubst (tf : @TypeFactory Unit) (concreteTy : LMonoTy) : Subst :=
+  match concreteTy with
+  | .tcons adtName concreteArgs =>
+    if concreteArgs.isEmpty then []
+    else match tf.getType adtName with
+      | some dt => [dt.typeArgs.zip concreteArgs]
+      | none => []
+  | _ => []
```
tautschnig: **Silent length mismatch.** `dt.typeArgs.zip concreteArgs` truncates to the shorter list. If a program ever reaches this with `concreteArgs.length ≠ dt.typeArgs.length` — shouldn't happen in a well-typed program, but parser / translator bugs could produce it — the resulting `Subst` is partial, `applySubst` leaves some type vars uninstantiated, and the generated `$$term` procedure uses an ill-formed precondition that will either fail type-check downstream or (worse) be silently ignored.

Two low-cost options:

```suggestion
private def mkTySubst (tf : @TypeFactory Unit) (concreteTy : LMonoTy) : Subst :=
  match concreteTy with
  | .tcons adtName concreteArgs =>
    if concreteArgs.isEmpty then []
    else match tf.getType adtName with
      | some dt =>
        if dt.typeArgs.length == concreteArgs.length then
          [dt.typeArgs.zip concreteArgs]
        else
          -- Length mismatch indicates an upstream bug (translator / type-checker);
          -- emit an empty substitution so applySubst leaves names intact and the
          -- ill-formed result fails visibly in later pipeline phases.
          []
      | none => []
  | _ => []
```

Or, if this is truly unreachable by construction (and the error checking in `termCheck` Step 1 guarantees it), a `panic!` with a message would make that invariant load-bearing and tested. Either is better than silent truncation.

Also a one-line docstring note: the `@[cases]`/`decreases` error check in `termCheck` Step 1 already rejects non-`.tcons` types, so the `_ => []` arm below is documentation-as-code for "unreachable"; worth saying so explicitly.

Strata/DL/Lambda/AdtRankAxioms.lean
```diff
@@ -0,0 +1,108 @@
...
+/-- Generate the non-negativity axiom `∀ x. adtRank(x) >= 0` for a datatype.
+    Trigger: `adtRank(x)` — fires only when adtRank terms already exist. -/
+def mkAdtRankNonNegAxiom {T : LExprParams} [Inhabited T.Metadata] [Inhabited T.IDMeta]
+    (dt : LDatatype T.IDMeta) (m : T.Metadata) : LExpr T.mono :=
+  let dtTy := dataDefault dt
+  let rankTy := .arrow dtTy .int
+  let rankX := .app m (.op m (adtRankFuncName dt.name) (.some rankTy))
+    (.bvar m 0)
+  let geExpr := .mkApp m (@intGeFunc T).opExpr [rankX, .intConst m 0]
+  .quant m .all "" (.some dtTy) rankX geExpr
```
tautschnig: **Proof-coverage suggestion (restatement of thread 6, concrete).** You acknowledged well-foundedness is a longer-term target. In the meantime, two cheaper theorems in `StrataTest/DL/Lambda/AdtRankAxiomsTests.lean` (or a new `AdtRankAxiomsProofs.lean`) would nail the load-bearing properties of the *current* axiom set — and would fail compilation if a future edit to this file regresses them:

1. **Shape theorem** — for the `IntList { Nil, Cons(hd:Int, tl:IntList) }` case, the generated non-neg axiom is literally `∀ x : IntList. IntList..adtRank(x) >= 0` and the Cons-decrease axiom is `∀ hd : Int, tl : IntList. IntList..adtRank(tl) < IntList..adtRank(Cons(hd, tl))`. If these are concrete-enough expressions, an `rfl`-shaped equality check catches any reshuffle of the fold order, trigger placement, or operator substitution.

   ```lean
   private def intListDt : LDatatype Unit := … -- the canonical IntList fixture
   private def intListBlock : MutualDatatype Unit := [intListDt]

   /-- The non-negativity axiom for IntList has exactly the expected shape. -/
   example :
       mkAdtRankNonNegAxiom intListDt () =
       -- literal expected expression goes here
       := by rfl

   /-- The Cons-decrease axiom for IntList has exactly the expected shape. -/
   example :
       (mkAdtRankPerConstrAxioms intListDt intListBlock ()).head? =
       some /- expected ∀ hd tl. … -/
       := by rfl
   ```

2. **Satisfiability by witness (harder but more meaningful)** — pick an interpretation (e.g., `adtRank = size`) and show by structural induction that it satisfies all generated axioms for a given datatype. This gives you the "axioms are consistent" property *without* going through general well-foundedness first. For mutual types like `Rose/RoseList`, the witness is the combined tree-size measure.

3. **Well-foundedness** (the full theorem, deferred): for any model of the axioms, there is no infinite descending chain `adtRank(xₙ₊₁) < adtRank(xₙ)`. Follows from the `>= 0` axiom + `<` chains being bounded below over `Int`. Straightforward once the denotational-semantics-for-datatypes work lands, per your response.

(1) alone would be worth adding in this PR — it's `rfl`-provable at near-zero maintenance cost and would catch a class of regressions that the current structural-count-only tests would miss.

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,357 @@
...
+  | adtRankAxiomsGenerated
+
+#derive_prefixed_toString Stats "TermCheck"
+
+/-- Suffix for generated termination-checking procedures. -/
+def termSuffix : String := "$$term"
+
+def termProcName (name : String) : String := s!"{name}{termSuffix}"
+
+def isTermProc (name : String) : Bool := name.endsWith termSuffix
```
tautschnig, line 47: **Refactor opportunity — generated-procedure-kind abstraction.** With this PR there are now two parallel families of generated sentinel procedures: `$$wf` (from `PrecondElim`) and `$$term` (from here). Each has a suffix constant, a `fooProcName` builder, a `fooSuffix` string, and an `isFooProc` predicate, and downstream code (`transformPipelinePhases:878`, `PrecondElim:380`) has to know the set of suffixes to handle generated procs specially. A third kind is easy to imagine (procedure-termination, which the PR description mentions is the next PR).

A small, single-file abstraction would collapse this:

```lean
/-- Internal markers on procedure names that identify auto-generated
    sentinel procedures produced by lowering passes. -/
inductive GeneratedProcKind where
  | wf      -- PrecondElim: contract/body well-formedness checks
  | term    -- TermCheck: termination obligations at recursive calls
  deriving DecidableEq

def GeneratedProcKind.suffix : GeneratedProcKind → String
  | .wf   => "$$wf"
  | .term => "$$term"

def GeneratedProcKind.decoratedName (k : GeneratedProcKind) (n : String) : String :=
  s!"{n}{k.suffix}"

def GeneratedProcKind.ofProcName? (name : String) : Option GeneratedProcKind := ...
```

Then `PrecondElim`'s `if TermCheck.isTermProc proc.header.name.name then …` becomes `match GeneratedProcKind.ofProcName? proc.header.name.name with | some .term => …`, and `transformPipelinePhases`'s per-kind `map _ProcName` becomes a single loop over `GeneratedProcKind` variants.

Not blocking this PR. Worth filing as a follow-up to do before the procedure-termination PR adds a third family.

Strata/Transform/TerminationCheck.lean
```diff
@@ -0,0 +1,357 @@
...
+        -- Step 1: error checking (fail-fast: an ill-formed function may
+        -- invalidate subsequent definitions in the mutual block)
+        -- Skip polymorphic functions: adtRank axioms are monomorphic, so we
+        -- cannot generate them for polymorphic datatypes yet. The user-facing
+        -- error is in Env.addFactoryFunc; when that restriction is lifted,
+        -- this filter must be updated to handle polymorphic adtRank generation.
+        let fileRange := Imperative.getFileRange md |>.getD FileRange.unknown
+        let throwErr (msg : String) : CoreTransformM Unit :=
+          throw (DiagnosticModel.withRange fileRange msg)
+        for func in funcs do
```
tautschnig, line 276: **Fail-fast error reporting — acknowledging your rationale.** Thread from last round. Your reply ("subsequent definitions depend on a nonexistent definition, which could lead to spurious error reports") is fair for the *semantic* consequences — a missing function in a mutual block does cascade.

One cheap UX improvement that preserves the fail-fast model: when we throw the first error, include a signal that the check *stopped* rather than *completed*, so the user knows other functions may also be wrong:

```suggestion
        for func in funcs do
          if func.typeArgs.isEmpty then
            match getDecreasesIdx func with
            | none =>
              match func.measure with
              | some _ =>
                throwErr s!"recursive function '{func.name.name}': decreases clause must be a parameter name. Non-structural recursion is not yet supported (other functions in this mutual block may also have termination issues; subsequent errors not reported)"
              | none =>
                throwErr s!"recursive function '{func.name.name}' requires a 'decreases' clause or a '@[cases]' parameter for termination checking (other functions in this mutual block may also have termination issues; subsequent errors not reported)"
            | some idx =>
              match func.inputs.values[idx]? with
              | some (.tcons n _) =>
                if (tf.getType n).isNone then
                  throwErr s!"recursive function '{func.name.name}': decreasing parameter type '{n}' is not a known datatype (other functions in this mutual block may also have termination issues; subsequent errors not reported)"
              | some _ =>
                throwErr s!"recursive function '{func.name.name}': decreasing parameter must have a datatype type (other functions in this mutual block may also have termination issues; subsequent errors not reported)"
              | none =>
                throwErr s!"recursive function '{func.name.name}': decreasing parameter index {idx} is out of range (other functions in this mutual block may also have termination issues; subsequent errors not reported)"
```

Alternatively, extract the "other functions in this mutual block …" phrase as a local `def` so the repetition doesn't pile up. The user-visible change is one extra clause per error that tells them to expect more work after the first fix.

If you think this clutters the messages, a one-line comment at the `for func in funcs do` explaining the design decision (fail-fast to avoid cascading from missing definitions) is the cheaper alternative and keeps the code as-is.

tautschnig (dismissed):

Strata/Languages/Core/DDMTransform/Grammar.lean
joscoh: As of the next PR, the `decreases` options will be: 1. structural recursion over a datatype (so the expression must be a free variable of ADT type) 2. an int-valued measure. This could change in the future if we want to allow e.g. lexicographic termination arguments. But no, I do not envision having some other datatype-producing function as the termination argument anytime soon.

Strata/Transform/TerminationCheck.lean
joscoh: I this could actually make things less clear for the user, since in practice I imagine many recursive functions are not mutually inductive, so we don't always want to show messages about mutual blocks.

Strata/Transform/TerminationCheck.lean
joscoh: I think this refactor makes sense if we add a third family as you mentioned.

Strata/DL/Lambda/AdtRankAxioms.lean
joscoh: 1. I don't quite understand this suggestion/comment. The given example is just a unit test for `intListDt`, which already exists in `AdtRankAxiomsTest.lean`. 
2. I added `AdtRankAxiomsModels.lean`, which gives the model for `IntList` and a RoseTree type. But doing this in general requires the semantics discussed earlier.

Strata/Transform/TerminationCheck.lean
joscoh: Added a length check and the unreachable comment.

Strata/DL/Lambda/AdtRankAxioms.lean
joscoh: Added nicer names

MikaelMayer (approved):

Strata/DL/Lambda/AdtRankAxioms.lean
atomb: Not really, and I'm not attached to changing this.

atomb (approved):

PR #1093, opened 2026-05-01
Structured pipeline diagnostics with PipelineM, phase timing, and --metrics JSONL

joehendrix:
## Motivation

The pipeline previously had no unified infrastructure for diagnostics or
timing. Warnings were handled ad-hoc (some printed to stderr, some silently
dropped), phase timing was manual and inconsistent, and the only structured
output (`--warning-summary`) wrote a single JSON blob at the end — useless
if the process was killed by a timeout. There was also no distinction between
fatal and non-fatal errors, so callers couldn't tell whether a result was
trustworthy.

This PR introduces `PipelineContext`/`PipelineM` as a shared backbone that
gives every pipeline phase consistent behavior: categorized diagnostics,
nested phase timing, fatal-error abort, and streaming metrics output that
survives timeouts.

## Summary

- Introduce `PipelineContext` and `PipelineM` for structured diagnostic
  accumulation and phase timing across the entire pipeline
- Replace ad-hoc `profileStep`/`startPhase` with `withPhase` combinator that
  handles nesting, timing, and profile output via `try`/`finally`
- Add `withRepeatedPhase` for iteration-level timing (preprocess, smtEncode,
  solver) that accumulates count+duration and reports as `[profile]` lines
- Replace `--warning-summary` (single JSON blob) with `--metrics <file>`
  (streaming JSONL, flushed per-record for timeout resilience)
- Add `--verbose` program dumps: prints intermediate Laurel and Core programs
  with `---- BEGIN/END ----` delimiters for programmatic parsing
- Refactor module resolution: fatal errors on invalid/missing modules,
  unified `resolveModules` for dispatch, auto-resolved, and explicit modules
- Fix `pyResolveOverloads` to print errors from pipeline context on failure
- Split `Messages.lean` into data types (`Messages.lean`) and runtime
  infrastructure (`Context.lean`)
- Move Python-specific `MessageKind` constants to
  `Languages/Python/Specs/MessageKind.lean`

## `PipelineContext` and `PipelineM`

`PipelineContext` is a structure carrying immutable config (`outputMode`,
`profilePipeline`) plus mutable state as individual `IO.Ref` fields:

- `messagesRef` — accumulated `PipelineMessage` diagnostics
- `toolErrorsRef` / `userCodeErrorsRef` — bucketed by impact for fast access
- `currentPhaseRef` — current position in the phase hierarchy (managed via
  `push`/`pop`)
- `repeatedDepthRef` — nesting depth of `withRepeatedPhase` scopes (when > 0,
  `withPhase` aggregates silently)
- `phaseStateRef` — per-phase scoped state bundling repeated-phase aggregation
  and message counts (saved/restored atomically on phase entry/exit)
- `metricsHandle` — optional file handle for streaming JSONL metrics

Because all state is in `IO.Ref`s, any monad with `BaseIO` access can use
pipeline capabilities by passing a `PipelineContext` value directly —
`withPhase` is polymorphic over the monad (`[MonadLiftT BaseIO m]`).

`PipelineM` is defined as:

```lean
abbrev PipelineM := ReaderT PipelineContext (EIO Unit)
```

The `EIO Unit` base ensures that fatal errors (configuration errors,
internal errors, user code issues) are impossible to silently ignore — they
throw `()`, aborting the pipeline. Non-fatal diagnostics (warnings, known
limitations) are added to `messagesRef` and execution continues. Callers
that want to attempt recovery (e.g., to gather more diagnostics) can catch
the exception, but should rethrow to avoid proceeding with invalid state.

The top-level runner catches the exception and still returns all accumulated
messages — so the caller always gets complete diagnostics regardless of how
the pipeline terminated.

## Phase tracking state machine

The phase system operates in two modes controlled by `repeatedDepthRef`:
- **Mode N** (normal, depth = 0): `withPhase` prints `[start]`/`[end]` in
  profile mode and emits JSONL timing metrics.
- **Mode R** (repeated, depth > 0): `withPhase` silently aggregates timing
  into `phaseStateRef.repeatedPhases` — no print, no individual metrics.

`withRepeatedPhase` increments `repeatedDepthRef` on entry and decrements
on exit. `withPhase` never changes the depth — it inherits the mode and
behaves accordingly.

Key invariants:
- `currentPhaseRef` always reflects the innermost active scope's full path
- `phaseStateRef` is scoped: saved on entry, restored on exit (no leakage)
- In mode R, all timing flows through `phaseStateRef.repeatedPhases` only

## `--metrics <file>` (replaces `--warning-summary`)

Writes streaming JSONL (one JSON object per line, flushed immediately after
each record). If the process is killed mid-pipeline, all records should remain.

Record types:
- `timing` — phase start/end in ms, with optional count for aggregated phases
- `diagnostic` — pipeline messages with phase, file, category, impact, location
- `outcome` — result/exit_code/total_ms, written last (absence = killed)

Example output:
```jsonl
{"type":"timing","phase":"pythonAndSpecToLaurel","start_ms":0,"end_ms":97}
{"type":"diagnostic","phase":"pythonAndSpecToLaurel.resolveAndBuildPrelude","file":"specs/builtins.ion","category":"unsupportedUnion","impact":"knownLimitation","message":"unsupported overload form for 'dict.update'"}
{"type":"timing","phase":"verification.vcDischarge.solver","start_ms":0,"end_ms":1102,"count":12}
{"type":"outcome","result":"verified","exit_code":0,"total_ms":1830}
```

## `--verbose` program dumps

When `outputMode == .verbose`, the pipeline prints intermediate programs to
stdout with delimiters for programmatic parsing:

```
---- BEGIN Laurel Program ----
<full Laurel program text>
---- END Laurel Program ----
---- BEGIN Core Program ----
<full Core program text>
---- END Core Program ----
```

## Module resolution refactoring

- `resolveModuleEntry` now takes a parsed `ModuleName` (not a string)
- Single `resolveModules` handles dispatch, auto-resolved, and explicit modules
- `invalidModuleName` changed from `internalWarning` to `configurationError` (fatal)
- Merged `missingDispatchModule` + `missingExplicitPySpec` into `missingPySpecModule`
- Eliminated duplicate message scenario where both invalid-name and missing-module
  errors fired for the same input

## `pyResolveOverloads` error handling

When `readDispatchOverloads` encounters a fatal error (e.g., nonexistent file),
the command now reads `pctx.messagesRef` and prints all accumulated messages
to stderr before exiting with a non-zero code — instead of silently yielding
empty results.

## Profile output format

`--profile` emits structured `[start]`/`[end]`/`[profile]` lines to stdout:

```
[start] pythonAndSpecToLaurel (time: 0ms)
  [start] readPythonIon (time: 0ms)
  [end] readPythonIon (time: 2ms)
  [start] resolveAndBuildPrelude (time: 2ms)
  [end] resolveAndBuildPrelude (time: 97ms)
  [warnings] resolveAndBuildPrelude: 2 unsupportedUnion
[end] pythonAndSpecToLaurel (time: 97ms)
[start] verification (time: 105ms)
  [start] vcDischarge (time: 320ms)
    [profile] preprocess (×12, total: 45ms, avg: 3ms)
    [profile] solver (×12, total: 1102ms, avg: 91ms)
  [end] vcDischarge (time: 1185ms)
[end] verification (time: 1185ms)
```

## Design: impact drives behavior

Each `MessageKind` has a `MessageImpact` that determines pipeline behavior:

| Impact | Fatal? | Meaning |
|--------|--------|---------|
| `configurationError` | yes | Invalid arguments or unreadable on-disk pyspecs |
| `internalError` | yes | Unexpected failure preventing output |
| `userCodeIssue` | yes | Issue in the user's source code |
| `internalWarning` | no | Unexpected condition that didn't prevent output |
| `knownLimitation` | no | Documented gap in modeling |

Note: `emitMessageAndAbort` is the function that aborts — callers choose
whether to abort. A caller may use `emitMessage` (continues) and check
`isFatal` at phase boundaries for custom abort strategies.

## Not addressed: `[statistics]` output

The `[statistics]` lines (transform counters like `Evaluator.simulatedStmts`,
`FilterProcedures.erasedProcedures`) are orthogonal to phase timing. They
currently bypass `PipelineContext` — printed via bare `IO.println` in
`Core.verify` and `StrataMainLib.lean` — so they don't respect phase nesting
or indentation. Integrating statistics into the phase system is future work.

## Key files changed

- `Strata/Pipeline/Messages.lean` — pure data types: `Phase`, `MessageImpact`, `MessageKind`, `PipelineMessage`
- `Strata/Pipeline/Context.lean` — runtime: `PipelineContext`, `PipelineM`, `withPhase`, `withRepeatedPhase`, `emitMetric`, `OutputMode`
- `Strata/Languages/Python/Specs/MessageKind.lean` — Python-specific `MessageKind` constants
- `Strata/Pipeline/PyAnalyzeLaurel.lean` — top-level pipeline orchestration, `--verbose` program dumps, `metricsHandle` threading
- `Strata/Languages/Core/Verifier.lean` — replaced manual timing with `withRepeatedPhase`
- `Strata/Languages/Laurel/LaurelCompilationPipeline.lean` — replaced `profileStep` with `withPhase`
- `Strata/Languages/Python/PySpecPipeline.lean` — fatal abort via throw, module resolution refactor, diagnostic metric emission
- `StrataMainLib.lean` — `--metrics` flag, outcome record emission, error printing in `pyResolveOverloads`
- `StrataTest/Pipeline/PhaseTimingTest.lean` — phase timing integration tests


Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -405,6 +405,8 @@ public instance : ToString PipelineError where
     | .knownLimitation msg => s!"Known limitation: {msg}"
     | .internal msg => msg
 
+def verboseWarnings : Bool := false
```
tautschnig: This is a hardcoded constant with no way for users to toggle it. Consider making it a parameter of `pythonAndSpecToLaurel` (or reusing the existing `verbose` flag from the CLI) instead of a dead `def`:

```suggestion
-- Individual pyspec warnings are suppressed by default; pass
-- verbose := true to see them.
```

Then thread it through from the caller.

tautschnig (commented):
`verboseWarnings : Bool := false` is a hardcoded constant — there's no way for users to enable verbose warnings without editing source code. This should either be:
1. A CLI flag (e.g., `--verbose-pyspec-warnings`) passed through to `pythonAndSpecToLaurel`, or
2. Controlled by the existing `--verbose` flag that's already available in the pipeline

The existing `quiet` parameter already suppresses the count line. Adding a `verbose` parameter (or reusing the one from the caller) would be cleaner than a dead constant.

Strata/Languages/Python/PySpecPipeline.lean
joehendrix: Yes, this was intentionally made a constant to avoid adding an extra flag that folks will almost never want to enable.

I could delete the code instead if that's preferred.

The other option is to set this aside a bit and come back with a more comprehensive overhaul of what gets printed.

tautschnig: I think we just need to figure out how to avoid bitrot here. Perhaps code removal would indeed be clearer?

joehendrix: I think I was being lazy and should just clean this up.  I'll revise with a more substantive change that will also align with the profiling goals.

joehendrix: This is ready for review again.

tautschnig:

StrataMain.lean
```diff
@@ -923,10 +941,7 @@ def pyResolveOverloadsCommand : Command where
     let pythonFile : System.FilePath := v[0]
     let dispatchPath := v[1]
     -- Read dispatch overload table
-    let overloads ←
-      match ← readDispatchOverloads #[dispatchPath] |>.toBaseIO with
-      | .ok (r, _) => pure r
-      | .error msg => exitFailure msg
+    let (overloads, _) ← readDispatchOverloads #[dispatchPath]
```
tautschnig: **Regression.** Before this PR `readDispatchOverloads` returned `EIO String (OverloadTable × _)` and a read failure propagated up as `exitFailure msg`. After this PR it returns `BaseIO (OverloadTable × PipelineState)` and on failure emits a fatal `.pySpecReadError` into `state.messages` and returns an empty `OverloadTable`. Tuple-destructuring `(overloads, _)` throws the state away, so a malformed or unreadable dispatch file now silently yields a zero-module stdout with no indication to the user that the dispatch table was empty.

Please either surface the state here, e.g.:

```suggestion
    let (overloads, dispatchState) ← readDispatchOverloads #[dispatchPath]
    if dispatchState.shouldAbort then
      for m in dispatchState.messages do
        IO.eprintln s!"{m.file}: {m.kind}: {m.message}"
      IO.Process.exit ExitCode.userError
```

or add a small helper (`readDispatchOverloadsOrExit` / `runPipelineOrExit`) that wraps the state check so other future callers can't forget.

Also worth a regression test that `pyResolveOverloads <py> <missing-dispatch.ion>` exits with a non-zero code rather than printing a blank result.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -133,95 +152,112 @@ private def mergeOverloads (old new : OverloadTable) : OverloadTable :=
...
   return { laurelProgram := combinedLaurel, overloads := allOverloads
            functionSignatures := funcSigs.toList,
            typeAliases := allTypeAliases
-           exhaustiveClasses := allExhaustiveClasses
-           pyspecWarnings := allWarnings }
+           exhaustiveClasses := allExhaustiveClasses }
+
+/-- Read PySpec Ion files and collect their Laurel declarations and overload
+    tables into a single combined result. -/
+public def buildPySpecLaurel (pyspecEntries : Array (String × String))
```
tautschnig: Exposing `(PySpecLaurelResult × PipelineState)` puts a load-bearing invariant on the caller: the `PySpecLaurelResult` is only meaningful when `state.shouldAbort = false`. `PySpecLaurelResult`'s `deriving Inhabited` plus the `return default` paths in `resolveAndBuildLaurelPrelude` (L305, L322) mean a caller that forgets to check `shouldAbort` gets an empty-but-typechecking result — which is exactly how the `pyResolveOverloadsCommand` regression manifests.

Options, in order of preference:

1. Return a sum type:
   ```lean
   inductive PipelineRun (α : Type) where
     | ok (value : α) (warnings : Array PipelineMessage)
     | aborted (errors warnings : Array PipelineMessage)
   ```
   so that callers can't accidentally read `value` when the run aborted.
2. Keep the current shape but add a `PipelineState.toExcept : PipelineState → α → Except (Array PipelineMessage) α` helper and convert the non-legacy `Except`-consuming callers via one helper.
3. Minimum: tighten the docstring on `PySpecLaurelResult` and on `buildPySpecLaurel` / `readDispatchOverloads` to spell out that the first component must not be inspected when `shouldAbort = true`.

Either (1) or (2) would have caught the `pyResolveOverloadsCommand` regression at compile time.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -133,95 +152,112 @@ private def mergeOverloads (old new : OverloadTable) : OverloadTable :=
...
 private def resolveModuleEntry (modName : String) (specDir : System.FilePath)
-    (quiet : Bool := false)
-    : EIO String (Option (String × String)) := do
+    : PipelineM (Option (String × String)) := do
   match Python.Specs.ModuleName.ofString modName with
   | .error _ =>
-    if !quiet then
-      let _ ← IO.eprintln
-        s!"warning: invalid module name '{modName}', skipping" |>.toBaseIO
+    emitMessage .invalidModuleName s!"invalid module name '{modName}', skipping" (file := specDir)
```
tautschnig: Pair this with the downstream fatal emit at L298 / L320: a bad `--dispatch foo.bar.baz?` where the module-name parse fails will produce both an `.invalidModuleName` (internalWarning, here) **and** a `.missingDispatchModule` (configurationError, below) referring to the same input. They share the `file := specDir` metadata, so the user sees two related messages for one cause.

One way to fix: have `resolveModuleEntry` return a finer result (e.g. `.ok entry | .invalid | .missing`) and only promote `.missing` to fatal in the caller, letting `.invalid` stay as the lone warning it already emitted. A test exercising `--dispatch 'not.a.valid.name!'` should show a single message and a non-zero exit.

Strata/Languages/Python/PipelineMessages.lean
```diff
@@ -0,0 +1,194 @@
...
+
+/-- A pipeline phase that produced a message. The `phaseNumber` determines
+    the display order so that warnings are printed in pipeline execution order. -/
+structure Phase where
+  private mk ::
+  phaseNumber : Nat
+  name : String
+  deriving BEq, DecidableEq, Hashable, Ord, Repr
+
+instance : LT Phase where
```
tautschnig: `Phase` derives `Ord` (lex over `phaseNumber` then `name`) and here hand-rolls `LT` with the same lex order, plus a `Decidable` instance. Since the two instances express the same order, one of them is dead weight and future drift (e.g. adding a field) will be silent.

Preferred: drop the hand-rolled `LT` + `Decidable` and rely on `ltOfOrd`:

```suggestion
instance : LT Phase := ltOfOrd
instance (a b : Phase) : Decidable (a < b) := inferInstance
```

(Or drop `Ord` from the `deriving` clause if you specifically don't want structure-lex comparison.)

Same comment applies to `MessageKind` below — that case is more serious, because the hand-rolled `LT` ignores `impact`, so for `a b : MessageKind` with equal phase/category but different impact, `compare a b = .eq` while `a ≠ b` (via `DecidableEq`). Today only `phase`+`category` distinguish `MessageKind`s in practice, but nothing in the type enforces that.

Strata/Languages/Python/PipelineMessages.lean
```diff
@@ -0,0 +1,194 @@
...
+      are invalid (e.g., missing module, unreadable file). -/
+  | configurationError
+  deriving BEq, DecidableEq, Hashable, Ord, Repr
+
+/-- Whether this impact level should abort the pipeline. Fatal impacts
+    indicate that the pipeline output would be incomplete or incorrect. -/
+def MessageImpact.isFatal : MessageImpact → Bool
+  | .internalError => true
+  | .configurationError => true
+  | .internalWarning => false
```
tautschnig: `MessageImpact.isFatal` is the semantic backbone of the abort flag — please lock it in with something testable. Two low-cost options:

1. `#guard` sanity checks inside this file (compile-time lockin, no proof):
   ```lean
   #guard MessageImpact.isFatal .internalError == true
   #guard MessageImpact.isFatal .configurationError == true
   #guard MessageImpact.isFatal .internalWarning == false
   #guard MessageImpact.isFatal .knownLimitation == false
   #guard MessageImpact.isFatal .userCodeIssue == false
   ```
2. A small theorem that the two fatal constructors are exactly the fatal impacts:
   ```lean
   theorem MessageImpact.isFatal_iff (i : MessageImpact) :
       i.isFatal = true ↔ i = .internalError ∨ i = .configurationError := by
     cases i <;> simp [isFatal]
   ```

The second is trivial to prove by case analysis and it rejects any future edit that silently reshuffles the match arms. Either is a much cheaper guarantee than discovering the drift by watching the test-suite exit codes change.

Bonus lock-ins worth considering for the `emitMessage` state transition (in `PySpecPipeline.lean`):

```lean
theorem emitMessage_shouldAbort {kind msg file loc s}
    (h : kind.impact.isFatal) :
    ((emitMessage kind msg file loc).run s |>.2).shouldAbort = true := by
  simp [emitMessage, h]

theorem addMessages_messages {msgs s} :
    ((addMessages msgs).run s |>.2).messages = s.messages ++ msgs := by
  simp [addMessages]
```

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -405,6 +449,44 @@ public instance : ToString PipelineError where
...
+      the failure are still available. -/
+  | failure (error : PipelineError) (warnings : Array Python.PipelineMessage)
+
+/-- Get warnings from a pipeline result regardless of success or failure. -/
+public def PythonToLaurelResult.warnings : PythonToLaurelResult → Array Python.PipelineMessage
+  | .success _ ws => ws
+  | .failure _ ws => ws
+
+/-- Generate a JSON warning summary from pipeline messages. -/
+public def Python.PipelineMessage.toSummaryJson
```
tautschnig: No tests cover `PipelineMessage.toSummaryJson`, which is now an externally-consumable JSON schema (`phase`, `category`, `impact`, `count`, `totalWarnings`). Please add at least one snapshot test in `StrataTest/Languages/Python/`, e.g.:

```lean
/--
info: {"pyspecWarningSummary":[{"phase":"pySpecToLaurel","category":"unsupportedUnion","impact":"knownLimitation","count":2}],"totalWarnings":2}
-/
#guard_msgs in
#eval (Strata.Python.PipelineMessage.toSummaryJson #[
  { file := "", loc := default, kind := .unsupportedUnion, message := "" },
  { file := "", loc := default, kind := .unsupportedUnion, message := "" }
]).compress
```

Without this, renaming a `MessageKind` constant, flipping an `impact`, or reordering the JSON keys can silently change the contract with any downstream consumer of `--warning-summary`.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -42,8 +42,28 @@ public structure PySpecLaurelResult where
...
-  pyspecWarnings : Array Python.Specs.SpecError := #[]
+  deriving Inhabited
+
+/-- Accumulated state for pipeline steps: messages and abort flag. -/
+public structure PipelineState where
+  messages : Array Python.PipelineMessage := #[]
+  shouldAbort : Bool := false
+
+/-- Monad for pipeline steps that accumulate messages. -/
+public abbrev PipelineM := StateT PipelineState BaseIO
```
tautschnig: Minor: consider moving `PipelineState`, `PipelineM`, `emitMessage`, `addMessages` into `PipelineMessages.lean`. The PR description says "`PipelineMessages` module" owns this vocabulary, but `PipelineState` and the monad live in `PySpecPipeline.lean` instead, which couples the abort-flag shape to the Python pyspec pipeline rather than making it reusable. If Laurel or the SMT encoder ever want similar accumulation, they'd have to depend on `PySpecPipeline` only for this type.

Strata/Languages/Python/PipelineMessages.lean
```diff
@@ -0,0 +1,194 @@
...
+  { phase := .moduleResolution, category := "missingAutoResolvedPySpec", impact := .knownLimitation }
+def missingDispatchModule : MessageKind :=
+  { phase := .moduleResolution, category := "missingDispatchModule", impact := .configurationError }
+def missingExplicitPySpec : MessageKind :=
+  { phase := .moduleResolution, category := "missingExplicitPySpec", impact := .configurationError }
+
+-- Overload resolution phase
+def overloadResolveWarning : MessageKind :=
+  { phase := .overloadResolution, category := "resolveWarning", impact := .internalWarning }
+
```
tautschnig: Nice that categories are all declared in one place — but all 20+ `MessageKind` constants share the same shape `{ phase := .X, category := "Y", impact := .Z }` and the `category` string literal is almost always a copy of the constant name. This is a lot of manual repetition that a macro (or a sum-type-based design with `impact`/`phase`/`category` derived from a constructor name) could replace. Not blocking, just a maintainability nudge — the current form makes it very easy to have typos in `category` strings that go undetected because there is no exhaustiveness check.

tautschnig (commented):
## Main concerns

1. **Real regression in `pyResolveOverloadsCommand` (see inline on `StrataMain.lean:943`).** The old `EIO String`-returning `readDispatchOverloads` surfaced dispatch-file read failures via `exitFailure`. The new `BaseIO (OverloadTable × PipelineState)` version emits a fatal `.pySpecReadError` into the state and returns an empty table — but the command tuple-destructures `(overloads, _)`, throwing the state away. A broken dispatch file now silently produces an empty overload table and a zero-module stdout.
2. **`shouldAbort` is a load-bearing invariant that the type system doesn't enforce.** `buildPySpecLaurel` and `readDispatchOverloads` return `(α × PipelineState)` with `α` constructed via `Inhabited.default` on fatal paths (`return default` in `resolveAndBuildLaurelPrelude`). Callers that forget to check `shouldAbort` silently proceed with empty/zero results — `pyResolveOverloadsCommand` is one case today. It would be worth either (a) returning a sum (`PipelineRun α := .ok α | .aborted (msgs : Array PipelineMessage)`), or (b) wrapping `BaseIO (α × PipelineState)` in a helper `runPipelineOrThrow : … → EIO String α` for callers that want legacy semantics, and documenting loudly in the structure docstring that a `PySpecLaurelResult` returned alongside `shouldAbort = true` is meaningless. See inline.
3. **Duplicate warnings for invalid module names.** `resolveModuleEntry` emits `.invalidModuleName` on parse failure and returns `none`; `resolveAndBuildLaurelPrelude` then emits `.missingDispatchModule` / `.missingExplicitPySpec` for every `none`. So an invalid name in `--dispatch` yields two messages (one internalWarning + one configurationError) where one is enough. See inline.
4. **`LT` / `Ord` instances disagree on `Phase` and `MessageKind`.** Both types derive `Ord` (lex over all fields) and hand-roll `LT` (lex over only some fields). On `MessageKind` the hand-rolled `LT` ignores the `impact` field entirely, so `compare a b = .eq` can coexist with `a ≠ b`. The `qsort` usage in `toSummaryJson` uses the hand-rolled `LT`, which is fine, but the two orderings should agree. Suggest dropping the derived `Ord` (or dropping the hand-rolled `LT` and using `ltOfOrd`).
5. **`MessageImpact.isFatal` has no proof/test coverage and the `(impact, isFatal)` contract lives in three places.** The mapping impact→fatal is encoded in: the `isFatal` function, the impact field of each `MessageKind` constant, and readers of `isFatal` in `emitMessage`. Adding a new impact level is a three-file change with no compiler help. At minimum, please add a `#guard`-level lock-in (see inline) and/or a small theorem bank (also inline).

## Testing / proof suggestions

- `emitMessage`/`addMessages` behavioral locks (small state-threading theorems — see `PipelineMessages.lean` inline).
- Snapshot / roundtrip for `PipelineMessage.toSummaryJson` so the external JSON schema can't drift silently (inline).
- Regression test for the `pyResolveOverloadsCommand` dispatch-file-read failure path (bullet #1).
- Regression test that a bad `--dispatch <modname>` produces **one** message and an exit with a single `configurationError`, not two overlapping warnings (bullet #3).
- Small theorems on `Phase` and `MessageKind` orderings (`∀ a b, a < b ↔ compare a b = .lt` to lock in LT/Ord agreement).

## Refactoring notes

- `mergeOverloads`: the old `Option`-based `paramName <|> n.paramName` semantics are now gone (the field is `String`), which is an improvement; the new code does emit `.overloadParamNameDisagreement` when two registrations conflict, which is great.
- `PipelineM`: `StateT PipelineState BaseIO` means any `BaseIO` effect can be mixed in. In practice only `readDDM` escapes into it. Consider narrowing (e.g. `ExceptT Abort (StateT … BaseIO)`) so the abort control flow is visible in types and `return default` isn't needed.
- Consider consolidating `MessageImpact` and `PipelineError` (in `PySpecPipeline.lean`): today the `PipelineError.userCode` branch exits with code 1 (hard fatal) while `MessageImpact.userCodeIssue` is non-fatal. The semantic gap is defensible but undocumented; a comment explaining why the two vocabularies exist would help.

StrataMain.lean
joehendrix: Thanks.  This is now fixed. `readDispatchOverloads` now runs in `PipelineM` where
fatal `.pySpecReadError` messages throw `()`. The `pyResolveOverloadsCommand` catches
the error and exits with a non-zero code.

joehendrix: I also just changed it to print the error messages and added a negative test to validate that occurs.

Strata/Languages/Python/PySpecPipeline.lean
joehendrix: Thanks for the suggestion.  I agree its error prone.  I've addressed this by the architectural change — `PipelineState` is gone.  Fatal errors record their error message in PipelineContext which has a mutable errors array, and also throw in `PipelineM`, so there is no valid-looking result to accidentally inspect after failure.

The `EIO Unit` exception serves the same role as the suggested sum type.  It doesn't signal a specific fatal error, but rather one needs to print the errors in the context.

Strata/Languages/Python/PySpecPipeline.lean
joehendrix: Fixed — refactored so `resolveModuleEntry` takes a parsed `ModuleName`, and all module resolution (dispatch, auto-resolved, explicit) goes through a single `resolveModules` that emits fatal errors on invalid or missing modules. The duplicate message is structurally impossible now.

Strata/Languages/Python/PipelineMessages.lean
joehendrix: Resolved — `Phase` is now just `{ path : Array String }` with no ordering instances. The hand-rolled `LT` and `phaseNumber` are gone.

Strata/Languages/Python/PipelineMessages.lean
joehendrix: I'm pretty skeptical this adds much.   This assumes we won't catch the `isFatal` code change, but will catch the theorem change.

That might be the case, but I'd prefer we just treat all code changes as potentially breaking things.

Strata/Languages/Python/PySpecPipeline.lean
joehendrix: Done. These now types live in `Strata.Pipeline.Messages`.

Strata/Languages/Python/PipelineMessages.lean
joehendrix: I agree this is a bit repetitive, but it's fairly easy to inspect.  I've changes it so the phase is no longer needed.  I'd consider adding a macro a bit more work than justified currently.  Maybe something for a future PR.

Strata/Languages/Python/PySpecPipeline.lean
joehendrix: I removed the `--warning-summary` in favor of a `--metrics` jsonl file that supports streaming.  This has a basic test.

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,477 @@
...
+
+/-- Whether this impact level represents an error (vs a warning). -/
+def MessageImpact.isError : MessageImpact → Bool
+  | .internalError => true
+  | .configurationError => true
+  | .internalWarning => false
+  | .knownLimitation => false
+  | .userCodeIssue => true
+
+instance : ToString MessageImpact where
```
tautschnig: **`isError` conflates two concerns.** The predicate is used at `Messages.lean:447` purely as a display tag (`[error]` vs `[warning]`), but its name suggests a behavioral property. The PR description's table has a `Fatal?` column — whether an impact level aborts the pipeline — which is a third, orthogonal concept that doesn't appear as a predicate anywhere.

Three notions tangled in `isError`:
1. **Display**: should the verbose-mode tag read `[error]` or `[warning]`? Used at line 447.
2. **Fatality**: does the impact level abort the pipeline when an `emit*` function is called? Per the PR description, yes for `configurationError`/`internalError`, no otherwise.
3. **User-visible**: is this something a user cares about (vs tool-internal)? Used nowhere currently, but a natural grouping.

`userCodeIssue` is a good stress test: it is (1) display-as-error, (2) non-fatal (per the table), (3) user-visible. The current `isError` returns `true`, aligning with (1), but the predicate name makes readers assume (2).

Suggestion: split into two named predicates, leaving behavioral semantics explicit:

```suggestion
/-- Display as an error tag (vs warning) in verbose output. -/
def MessageImpact.displayAsError : MessageImpact → Bool
  | .internalError | .configurationError | .userCodeIssue => true
  | .internalWarning | .knownLimitation => false

/-- Whether the pipeline should abort on this impact.
    Matches the "Fatal?" column in the PR description's behaviour table. -/
def MessageImpact.isFatal : MessageImpact → Bool
  | .internalError | .configurationError => true
  | .internalWarning | .knownLimitation | .userCodeIssue => false
```

And at `:447`, use `displayAsError`.

Related: with `isFatal` as a first-class predicate, `emitMessage` can derive the abort decision:

```lean
public def emitMessage (kind : Pipeline.MessageKind) (message : String) … : PipelineM Unit := do
  addMessage { file, loc, phase, kind, message }
  if kind.impact.isFatal then throw ()
```

and `emitFatalMessage` becomes redundant, or becomes a `panic!`-guarded wrapper that asserts `impact.isFatal` at the call site (compile-time lockin of the invariant).

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,477 @@
...
+  private skipTiming : Bool := false
+  private messagesRef : IO.Ref (Array PipelineMessage)
+  private toolErrorsRef : IO.Ref (Array PipelineMessage)
+  private userCodeErrorsRef : IO.Ref (Array PipelineMessage)
+  private timingRef : IO.Ref (Array PhaseTimingEntry)
+  private currentPhaseRef : IO.Ref Phase
+  private messageCountAtPhaseStartRef : IO.Ref Nat
+  private repeatedPhasesRef : IO.Ref (Std.HashMap String (Nat × Nat))
+  private metricsHandle : Option IO.FS.Handle := none
+
```
tautschnig: Two small things worth documenting on `PipelineContext`:

1. **Thread safety.** All eight `IO.Ref` fields are plain refs, not `IO.Mutex`-guarded. If a future parallel backend (parallel per-obligation SMT, parallel-phase translation) ever runs `PipelineM` actions concurrently against the same context, the races are silent data corruption. Add a docstring note:

   ```lean
   /-- Pipeline context carrying immutable config and mutable state as individual IORefs.
       …

       NOTE: this context is single-threaded. The mutable refs are not guarded
       by a mutex, so running multiple `PipelineM` actions in parallel against
       the same context races. Adopt `IO.Mutex` (or equivalent) before
       parallelising any phase. -/
   structure PipelineContext where
   ```

2. **`metricsHandle` ownership is ambiguous.** The field is `Option IO.FS.Handle`; nothing in this file opens, closes, or documents who is responsible. Today, `StrataMain.lean:640` opens and leaves open; the handle lives until GC. This works by accident — a future refactor that moves metric-file lifetime inside the pipeline runner wouldn't notice the ambiguity.

   Either (a) make the lifetime explicit in `PipelineContext.create` (takes a `System.FilePath` and opens inside; closes on pipeline end), or (b) leave the caller responsible and say so in the field docstring. See the separate inline on `PyAnalyzeLaurel.lean:41` for the caller-side half.

Strata/Pipeline/PyAnalyzeLaurel.lean
```diff
@@ -0,0 +1,129 @@
...
+  sourcePath : Option String := none
+  profile : Bool := false
+  keepAllFilesPrefix : Option String := none
+  verifyOptions : Core.VerifyOptions
+  entryPoint : Core.EntryPoint := Core.EntryPoint.roots
+  isBugFinding : Bool := true
+  outputMode : OutputMode := .default
+  skipVerification : Bool := false
+  skipTiming : Bool := false
+  metricsHandle : Option IO.FS.Handle := none
```
tautschnig, line 41: **`metricsHandle` has no explicit close/flush lifecycle in this file or its caller.** `StrataMain.lean:640` opens the handle and passes it via config; `PipelineContext` holds it; `emitMetric` (Messages.lean:302-304) flushes after each write; but nothing calls `h.close` after the pipeline terminates.

In practice Lean's GC finalises the handle, but there are two latent concerns:

1. **`h.flush` is `fflush()`-equivalent — it reaches the OS buffer, not durable disk.** The PR body claims "all records written up to that point remain on disk" even on SIGKILL; that's true relative to the process's own stdio buffers, not the OS's (which may still lose data if the host crashes). Worth a docstring note clarifying the exact guarantee.

2. **Outcome record is the only record written after the pipeline returns** — at `StrataMain.lean:651-657` (`emitOutcome`). If any step between pipeline-end and `emitOutcome` throws (e.g., reportUserCodeError's file writing), the outcome record is missing with no signal that the pipeline actually completed. Consider wrapping `emitOutcome` in `try … finally` so the record lands regardless.

Cleaner shape: have `runPyAnalyzePipeline` take a `FilePath` for metrics (not a pre-opened handle), open/close with `IO.FS.withFile` internally, and emit the outcome record in a `finally` block. The caller doesn't have to remember to close, and the outcome record is guaranteed to land on any pipeline exit path.

Documentation-only alternative if you prefer to keep the current handle-passing API:

```suggestion
  /-- Optional metrics file handle (caller opens, caller closes).
      The pipeline appends one JSONL record per `emitMetric` call and flushes
      after each write. Guarantees:
      - Records written before a crash are in the OS buffer (not necessarily
        durable — `fsync`-equivalent is not called).
      - The caller is responsible for emitting a final `outcome` record and
        closing the handle. -/
  metricsHandle : Option IO.FS.Handle := none
```

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,477 @@
...
+  addMessage { file, loc, phase, kind, message }
+
+/-- Emit a diagnostic message and abort the pipeline.
+    Polymorphic return type allows use in expression position. -/
+public def emitFatalMessage (kind : Pipeline.MessageKind) (message : String)
+    (file : System.FilePath) (loc : SourceRange := default) : Pipeline.PipelineM α := do
+  emitMessage kind message file loc
+  throw ()
+
+/-- All messages with a given impact. -/
```
tautschnig: **The abort decision lives here, not in `MessageImpact`.** `emitMessage` unconditionally accumulates + continues; `emitFatalMessage` unconditionally accumulates + throws. Neither inspects `kind.impact`. In practice callers across the tree pair the function with an impact-consistent kind (I checked the 12 emit-sites in `PySpecPipeline.lean` and `PyAnalyzeLaurel.lean`), but nothing enforces the pairing.

Two concrete ways to close the gap:

**(a) Derive from impact** (collapses to one emit function):

```suggestion
/-- Emit a diagnostic message. Aborts the pipeline if the kind's impact is
    fatal (`.internalError` or `.configurationError`), continues otherwise.
    Polymorphic in `α` so it can be used in expression position when fatal. -/
public def emitMessage (kind : Pipeline.MessageKind) (message : String)
    (file : System.FilePath := default) (loc : SourceRange := default) : Pipeline.PipelineM Unit := do
  let phase ← (← read) |>.currentPhaseRef.get
  addMessage { file, loc, phase, kind, message }
  if kind.impact.isFatal then throw ()
```

and drop `emitFatalMessage`.

**(b) Guard at call time** (keeps two entry points, locks the pairing):

```lean
public def emitFatalMessage (kind : Pipeline.MessageKind) (message : String)
    (file : System.FilePath) (loc : SourceRange := default) : Pipeline.PipelineM α := do
  if !kind.impact.isFatal then
    panic! s!"emitFatalMessage called with non-fatal impact {toString kind.impact}"
  emitMessage kind message file loc
  throw ()
```

(a) is cleaner; (b) preserves current call-site readability. Either way, the PR description's impact table becomes enforced rather than aspirational.

Note: `emitFatalMessage` currently has return type `PipelineM α` (polymorphic for expression-position use). Option (a) needs an `α`-polymorphic variant for callers that use the return value — the signature above returns `Unit`, which would break some existing call sites. A safer option-(a) is to keep `emitFatalMessage`'s signature but have it guard on impact internally.

tautschnig (commented):
Substantial pipeline refactor. The PR delivers what the description promises: `PipelineContext` + `PipelineM` with proper diagnostic accumulation, `withPhase` / `withRepeatedPhase` replacing ad-hoc `profileStep`, streaming JSONL metrics via `--metrics`, fatal-abort via `EIO Unit`, and cleaner module resolution. `lake build` passes. The structural-output-survives-timeout property is genuinely useful.

**1. The "impact drives behavior" claim in the PR description isn't enforced by code.** The PR body's table says:

```
| Impact                | Fatal? |
| configurationError    | yes    |
| internalError         | yes    |
| internalWarning       | no     |
| knownLimitation       | no     |
| userCodeIssue         | no     |
```

But whether a message aborts depends entirely on whether the caller uses `emitFatalMessage` vs `emitMessage` (see `Messages.lean:461-481`). Nothing in `PipelineM` enforces the mapping:

- `emitFatalMessage .overloadResolveWarning msg …` would throw despite the `.internalWarning` impact.
- `emitMessage .internalError msg …` would accumulate a fatal-impact message without aborting.

Today's in-tree callers are consistent with the table, but the invariant is tribal knowledge. Two low-cost options:

- **Derive behavior from impact** — make `emitMessage` check `if kind.impact.isFatal then throw ()` and drop `emitFatalMessage` entirely. This makes the table the contract.
- **Retain both but guard** — `emitFatalMessage` `panic!`s on non-fatal impact; `emitMessage` `panic!`s on fatal impact. Keeps the current call-site readability; enforces the invariant.

Related: `MessageImpact.isError` (Messages.lean:73) returns `true` for `.userCodeIssue` but the PR body's table marks it non-fatal. `isError` here means "display as 'error' tag", not "abort the pipeline". Calling the predicate `isError` for a display concern and leaving the abort concern nameless is confusing — consider `displayAsError` + `isFatal` as two explicit predicates.

**2. `PipelineM = ReaderT PipelineContext (EIO Unit)` throws `()` on abort.** The unit-typed error channel means any caller doing `try … catch e => …` gets zero information from `e` — they have to go read `ctx.messagesRef` to reconstruct what happened. Two suggestions:

- Throw a typed error (`PipelineAbort` with a reason or message ref) so `catch` is self-documenting.
- Or document the `()`-throwing convention explicitly in the `PipelineM` docstring, including the required "inspect `messagesRef` on abort" recovery pattern. Currently the docstring says "throws `()` to abort" but not how a caller should react.

**3. `metricsHandle` is never explicitly closed.** `StrataMain.lean:640` opens the file, threads it through `PipelineContext`, but nothing calls `h.close` / `withFile` / equivalent. Lean's GC will finalize the handle, but:

- If the `outcome` record emission at `StrataMain.lean:651-657` fails partway through, the handle is leaked.
- For long-running pipelines with the process still live, the outcome record may be buffered (even with per-record flush — the OS page cache is a separate layer).
- No guarantee the OS flushes before a `SIGKILL`.

The PR body claims "all records written up to that point remain on disk" — per-record `h.flush` at `Messages.lean:304` does `flush`, which is `fflush()` equivalent; on most systems that reaches the OS buffer but not the disk (that's `fsync`). For timeout resilience, `fsync` on each record is expensive but correct. Worth a docstring note clarifying the guarantee, and wrapping the handle in `IO.FS.withFile`-style resource management. Inline on `PyAnalyzeLaurel.lean:41`.

**4. Thread safety: `PipelineContext` uses plain `IO.Ref`s.** If the pipeline ever parallelizes (e.g., parallel per-obligation SMT), `messagesRef`, `timingRef`, `repeatedPhasesRef`, `currentPhaseRef` will race. Not a problem today, but the design choice should be stated in the docstring ("single-threaded; adopt `IO.Mutex` before parallelizing"). Inline on `Messages.lean:255`.

**5. `withPhase`'s `finally` may emit multiple `[end]` lines if the action itself called `withPhase` but threw before its own `finally` ran.** The `finally` in `withPhase` (Messages.lean:374-382) saves and restores the parent phase; but if a nested `withPhase` is interrupted by a throw before its own `finally`, the state is in a partial save-restore pattern. The outer `finally` restores to the outermost parent, which should be correct — but worth a test case that throws mid-nested-phase to confirm `[end]` lines match `[start]` lines in the profile output.

**6. Test coverage is thin for the new infrastructure.** The existing tests (`AnalyzeLaurelTest.lean`, `ToLaurelTest.lean`) exercise the happy path and a few error paths through the new `PipelineM`. Not exercised:

- Streaming JSONL output when the process is interrupted (can't be tested easily, but a fixture that verifies the JSONL is well-formed on normal exit would catch structural regressions).
- Multiple fatal errors accumulating before abort (the PR body notes "multiple error-impact messages may accumulate before or across aborts" but no test pins this down).
- `withRepeatedPhase` aggregation (count + total) — the test suite calls through to the verifier but doesn't assert on the aggregated output.
- Error path in `pyResolveOverloads` (T1 says a negative test was added; I see `ToLaurelTest.lean` grew by 250 lines — worth confirming the happy-and-error-path coverage is balanced).

**Proof-coverage suggestions** (each small):

```lean
/-- `emitFatalMessage` always throws. -/
theorem emitFatalMessage_throws (kind : MessageKind) (msg : String) (file : System.FilePath) :
    (emitFatalMessage kind msg file).run ctx = .error () ∧
    (∃ msgs', ctx.messagesRef.get = pure msgs' ∧ msgs'.size > 0)

/-- Metric file accumulates one record per `emitMetric` call. -/
theorem emitMetric_appends_line (ctx : PipelineContext) (j : Lean.Json) :
    ctx.metricsHandle.isSome →
    (emitMetric ctx j).run = append j.compress ++ "\n" to file

/-- `withPhase` restores current phase regardless of whether action threw. -/
theorem withPhase_restores_phase (ctx : PipelineContext) (name : String) (action : PipelineM α) :
    let before ← ctx.currentPhaseRef.get
    let _ ← try (withPhase name action).run ctx catch _ => pure ()
    ctx.currentPhaseRef.get = before
```

The third is the one I'd most want — it's the exception-safety invariant of `withPhase`, and without it, a throw inside the action can leave the phase hierarchy desynchronized, producing unmatched `[start]`/`[end]` in profile output.

**Refactor callouts:**

- `MessageKind` block (Messages.lean:102-215) has 30+ constants, each `{ category := "…", impact := .X }`. My earlier macro suggestion (T8) was declined and I'll drop it — but worth a `#guard` at the bottom of the block checking the category string matches the constant name for each (prevents `{ category := "unsupportdUnion", … }`-style typos).
- `Messages.lean:258` (`private mk ::`) + several `private fieldRef :` fields — good encapsulation, but the `create` constructor now has 9 fields. Consider `.mkDefault` with the common shape and individual setters for the rare overrides.
- `withPhase` / `withRepeatedPhase` / `emitMessage` / `emitFatalMessage` — four entry points with similar shapes. A `PipelineAction` typeclass or a single `runPipeline` combinator that takes the phase/message/abort intent explicitly would consolidate.

Strata/Pipeline/Messages.lean
joehendrix: Agreed the naming was conflating concerns. I've split this into two changes:

1. Renamed `isError` → `isFatal` — it's now used only at abort-decision sites. The verbose-mode display tag now uses `toString msg.kind.impact` directly (printing the actual impact level like `[internalError]`, `[knownLimitation]`, etc.) rather than collapsing to a binary `[error]`/`[warning]`.

2. Renamed `emitFatalMessage` → `emitMessageAndAbort` — the name describes control flow (emit + throw), not a required pairing with impact severity.

On collapsing to a single emit function that auto-aborts based on `isFatal`: I intentionally want the caller to choose whether to abort. A caller may accumulate multiple fatal-impact messages before aborting (giving the user a complete error list).  The `isFatal` docstring now notes that pipeline steps may want a custom abort strategy rather than relying on this predicate.

Strata/Pipeline/Messages.lean
joehendrix: Both addressed:
1. The `PipelineContext` docstring already documents the single-threaded assumption and warns about adopting `IO.Mutex` before parallelising.
2. Added a field docstring on `metricsHandle`: "Caller-owned handle for JSONL metrics. The pipeline appends and flushes per record but does not open or close the handle."

Strata/Pipeline/Messages.lean
joehendrix: Addressed together with Thread 1. The design is intentionally caller-controlled:

- `emitMessage` accumulates and continues — impact is for downstream consumers.
- `emitMessageAndAbort` accumulates and throws — the name describes the control flow.
- `isFatal` exists as a convenience predicate for callers who want the default abort policy, but pipeline steps may use a custom strategy (documented in `isFatal`'s docstring).

This avoids the coupling concern: a single auto-aborting emit function would prevent patterns like "scan all modules, collect all errors, then abort" which is what `readDispatchOverloads` does.

Strata/Pipeline/PyAnalyzeLaurel.lean
joehendrix: I've moved emitOutcome up so exceptions should not disrupt writing it.  Signals still could, but I think that's reasonable.  The standard out output would reflect the pipeline did not finish.

joehendrix: I think given the context, logging that may timeout, `fflush` is the right level of abstraction.  We aren't building a database with strong consistency guarantees against host crashes.  I could add the caveats about host crashes or OS level bugs impacting this, but it also seems relatively clear from the context.

aqjune-aws (dismissed):
Approving since the nested `[profile]` feature works now and `lake exe strata verify --profile Examples/HeapReasoning.core.st` prints a reasonable output

MikaelMayer (commented):
🤖🔍 Well-structured PR that introduces a coherent pipeline diagnostics backbone. The `PipelineContext`/`PipelineM` design is clean, the phase timing system is well-tested, and the streaming JSONL metrics are a clear improvement over the single-blob `--warning-summary`. The test coverage for phase timing is thorough.

A few issues worth addressing:

- `endPhaseNormal` has a latent index bug that will break if mode-N `withPhase` calls are ever nested (currently safe because all nesting uses `withRepeatedPhase`).
- `--verbose` lost the intermediate Laurel/Core program dumps.
- Several `MessageKind` constants are defined but unused.

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,630 @@
...
+      ("start_ms", .num 0), ("end_ms", .num (nsToMs data.totalNs)),
+      ("count", .num data.count)])
+    flushRepeatedEntries ctx subphase data.children
+
+/-- End the current phase in mode N: flush aggregated repeated subphases,
+    record end time, print [warnings] summary in profile mode. -/
+private def endPhaseNormal (ctx : PipelineContext) : BaseIO Unit := do
+  let currentPhase ← ctx.currentPhaseRef.get
+  let timing ← ctx.timingRef.get
+  let parentIdx := timing.size - 1
```
MikaelMayer: 🤖🔍 **Latent index bug in `parentIdx`.** This computes `timing.size - 1` assuming the current phase's entry is always the last one pushed to `timingRef`. This is only true if no child `withPhase` (in mode N) has completed and pushed its own entry after the parent's.

Currently safe because all nested phases use `withRepeatedPhase` (mode R), which doesn't push to `timingRef`. But if someone later nests `withPhase` inside `withPhase` in mode N, `parentIdx` will point to the last child's entry, not the parent's.

Consider saving the parent's index in `PhaseSavedState` at `startPhase` time (i.e., `parentIdx := timing.size` just before pushing), which is always correct regardless of nesting.

Strata/Pipeline/PyAnalyzeLaurel.lean
```diff
@@ -0,0 +1,127 @@
...
+        if msg.kind.impact.isFatal then throw ()
+      match coreOpt with
+      | some core => pure (core, stats)
+      | none =>
+        emitMessageAndAbort (file := uri) .laurelToCoreError s!"Laurel to Core translation failed: {diags}"
+    | .error e =>
+      emitMessageAndAbort (file := uri) .laurelToCoreError s!"Laurel translation error: {e}"
+
+  if config.skipVerification then
+    return (PyAnalyzeOutcome.verified #[] coreProgram, laurelPassStats)
```
MikaelMayer, line 84: 🤖🔍 **`--skip-verification` returns `.verified #[] coreProgram`** — this means the caller in `StrataMain.lean` enters the `.verified` branch and calls `printPyAnalyzeSummary` with an empty `vcResults` array. The old code had a dedicated `--skip-verification` path that printed diagnostic counts and exited with an appropriate code based on translation errors. Verify that the new behavior (printing "0 verified, 0 failed" via `printPyAnalyzeSummary`) is intentional.

StrataMain.lean
```diff
@@ -721,57 +669,101 @@ def pyAnalyzeLaurelCommand : Command where
...
+    -- Derive output mode from CLI flags.
+    let outputMode : Strata.Pipeline.OutputMode :=
+      if verbose then .verbose
+      else if profile then .profile
+      else if quiet then .quiet
+      else .default
+    let skipVerification := pflags.getBool "skip-verification"
+
+    -- Run the pipeline
+    let (outcome, laurelPassStats, pctx) ← Strata.Pipeline.runPyAnalyzePipeline {
```
MikaelMayer: 🤖🔍 **`--verbose` regression.** The old code printed the intermediate Laurel program (`==== Laurel Program ====`) and Core program (`==== Core Program ====`) when `--verbose` was set. This output is useful for debugging translation issues. The new pipeline no longer produces it — `outputMode == .verbose` only controls phase timing and per-message stderr output. Consider adding the program dumps back (e.g., in `runPipeline` after each translation step when `config.outputMode == .verbose`).

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,630 @@
...
+def laurelToCoreWarning : MessageKind :=
+  { category := "warning", impact := .internalWarning }
+def laurelToCoreError : MessageKind :=
+  { category := "error", impact := .internalError }
+def laurelToCoreNotImpl : MessageKind :=
+  { category := "notYetImplemented", impact := .knownLimitation }
+def laurelToCoreUserError : MessageKind :=
+  { category := "userError", impact := .userCodeIssue }
+
+-- Core transforms phase
```
MikaelMayer: 🤖🔍 Nit: `coreTransformWarning`, `coreTransformError`, `coreTransformNotImpl`, `coreTransformUserError`, `laurelToCoreWarning`, `laurelToCoreNotImpl`, `laurelToCoreUserError`, and `verificationWarning` are defined but never used in this PR. If they're forward-looking placeholders for future phases, a brief comment saying so would help. Otherwise they're dead code.

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,630 @@
...
+
+/--
+Whether this impact level typically warrants aborting the pipeline.
+
+N.B. Pipeline steps may want a custom abort strategy rather than
+relying on this predicate.
+-/
+def MessageImpact.isFatal : MessageImpact → Bool
+  | .internalError => true
+  | .configurationError => true
```
MikaelMayer, line 76: 🤖🔍 `userCodeIssue` is classified as fatal here, which means any user Python error (e.g., wrong keyword argument) aborts the pipeline immediately. The old `PipelineError.userCode` also aborted, so this preserves behavior — but the PR description says "Non-fatal diagnostics (warnings, known limitations) are added to `messagesRef` and execution continues." The name `isFatal` and the docstring ("typically warrants aborting") are accurate, but the fact that `userCodeIssue` is fatal means the pipeline can never accumulate multiple user errors in a single run. If that's intentional, fine — but it limits the usefulness of the streaming diagnostics for user-facing error reporting.

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,630 @@
...
+  /-- An unexpected failure that prevented some output from being generated
+      (e.g., a malformed overload entry that was skipped). -/
+  | internalError
+  /-- An unexpected condition that did not prevent output, but may indicate
+      a tool bug worth investigating. -/
+  | internalWarning
+  /-- A known, documented limitation that may cause specs to be incomplete
+      or imprecise. -/
+  | knownLimitation
+  /-- An issue detected in the user's Python source code. -/
```
atomb: This won't always be Python code.

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,630 @@
...
+def overloadArgNotStringLiteral : MessageKind :=
+  { category := "overloadArgNotStringLiteral", impact := .internalError }
+def overloadReturnArity : MessageKind :=
+  { category := "overloadReturnArity", impact := .internalError }
+def overloadReturnNotClass : MessageKind :=
+  { category := "overloadReturnNotClass", impact := .internalError }
+def overloadParamNameDisagreement : MessageKind :=
+  { category := "overloadParamNameDisagreement", impact := .internalError }
+
+-- PySpec parsing phase
```
atomb: When the Python-related code gets migrated to a separate package, we'll want to define these elsewhere, along with some of the others below.

Strata/Pipeline/Messages.lean
atomb, line 1: Nit: this filename seems overly narrow. There's a lot of pipeline stuff in here that's not really about messages.

StrataTest/Pipeline/PhaseTimingTest.lean
atomb, line 1: There's probably room to extract some abstractions from the code in this file, but it's not high priority.

atomb (dismissed):
I made a few minor stylistic comments, but nothing blocking.

Strata/Pipeline/Messages.lean
joehendrix: Good catch.  I removed the timingRef entirely.  It turns out it wasn't used anymore.

MikaelMayer (commented):
🤖🔍 The primary actionable issue (latent index bug in `parentIdx`) is fully resolved — the new commit removes `timingRef` entirely and passes `startNs` directly, eliminating the bug class. The remaining comments (unused `MessageKind` constants, `--verbose` program dumps, `--skip-verification` behavior, `userCodeIssue` fatality) were design observations/nits that remain unchanged but are not blockers. Reviewer sign-off still needed.

Strata/Pipeline/PyAnalyzeLaurel.lean
joehendrix: This is intentional.  The old code had a dedicated path but still generated the core program.  This preserves the behavior since the warnings/errors from earlier phases are captured in the PipelineContext and StrataMain will still return the appropriate error code.  The separate path just adds complexity.

Strata/Pipeline/Messages.lean
joehendrix: That's a good point.  `PythonToLaurel` currently is the phase capturing user code issues and it's design (`Except` monad) limits it to one error at a time.   This is intended to preserve existing behavior.

When that is changed we can change `userCodeIssue` to false (or implement a different abort strategy) as noted in the docstring.

MikaelMayer (commented):

StrataMain.lean
joehendrix: I've fixed the code to print out the programs.

Strata/Pipeline/Messages.lean
joehendrix: Remove all unused message kinds.

Strata/Pipeline/Messages.lean
joehendrix: Fixed

Strata/Pipeline/Messages.lean
joehendrix: Agreed.  I partitioned Messages into Messages and Context.

Strata/Pipeline/Messages.lean
joehendrix: Moved them.

MikaelMayer (commented):

- **Index bug** (fixed in prior commit): `timingRef` removed entirely, no longer applicable.
- **`--verbose` regression** (fixed in `b53969b2f`): Laurel and Core program dumps are now printed when `outputMode == .verbose`.
- **Unused `MessageKind` constants** (fixed in `39bd5c0a8`): old unused constants removed; new constants in `Specs/MessageKind.lean` are actively used by `ToLaurel.lean`.
- **`--skip-verification` behavior** and **`userCodeIssue` fatality**: design observations, unchanged — not blockers.

The `Messages.lean` → `Messages.lean` + `Context.lean` split is clean.

StrataTest/Pipeline/PhaseTimingTest.lean
joehendrix: Yes, I agree.  I haven't changed it though

MikaelMayer (commented):

MikaelMayer (commented):

aqjune-aws (dismissed):

Strata/Pipeline/Messages.lean
```diff
@@ -0,0 +1,133 @@
...
+  impact : MessageImpact
+  deriving BEq, DecidableEq, Hashable, Ord, Repr
+
+instance : ToString MessageKind where
+  toString mk := mk.category
+
+namespace MessageKind
+
+-- Laurel lowering phase
+def laurelLoweringError : MessageKind :=
```
MikaelMayer, line 102: These helpers are not specific to Laurel, is it possible to rename them so that their name does not depend on Laurel?

MikaelMayer (commented):
Just one comment and I'm good !

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -5,20 +5,21 @@
...
+import all Strata.DDM.Util.String
 import Strata.Languages.Laurel.FilterPrelude
 import Strata.Languages.Laurel.LaurelCompilationPipeline
-public import Strata.Util.Statistics
 public import Strata.Languages.Python.PythonToLaurel
 import Strata.Languages.Python.ReadPython
 import Strata.Languages.Python.PythonLaurelCorePrelude
 import Strata.Languages.Python.PythonRuntimeLaurelPart
 import Strata.Languages.Python.Specs
 import Strata.Languages.Python.Specs.DDM
```
shigoel, line 16: The old code had `existing.paramName <|> n.paramName`, which would pick up the new entry's `paramName` when the existing one was `none`. The new code unconditionally keeps `existing.paramName`, silently dropping `n.paramName` even if `existing.paramName = none`. Was this intentional?

shigoel (commented):
Minor comment on `mergeOverloads` semantics.

Strata/Pipeline/Messages.lean
joehendrix: These are specific to the Laurel lowering and Laurel-to-Core phases — the prefix indicates which pipeline phase emits the diagnostic. The phase context is also stamped on each message at emission time via `currentPhaseRef`, so the prefix is somewhat redundant, but it makes call sites self-documenting.

Happy to rename if you have a specific suggestion. The alternatives I see are generic names (`loweringError`, `notImpl`, `userError`) which could be confusing across phases, or keeping the phase-prefixed convention as-is.

Strata/Languages/Python/PySpecPipeline.lean
joehendrix: Yes, intentional. This PR made `FunctionOverloads.paramName` non-optional (changed from `Option String` to `String`), so the `<|>` fallback is no longer needed — both sides always have a valid `paramName`. Additionally, `pushOverloadEntry` (Specs/ToLaurel.lean:124) validates that `paramName` values agree for the same function, so `existing.paramName == n.paramName` is guaranteed by the time `mergeOverloads` runs.

atomb (dismissed):

MikaelMayer (commented):

- `--verbose` Laurel/Core program dumps: restored in `b53969b2f`
- Unused `MessageKind` placeholder constants: removed in `39bd5c0a8` (used ones moved to `Languages/Python/Specs/MessageKind.lean`)
- Latent index bug (`timingRef`): already resolved in earlier commit
- `--skip-verification` behavior and `userCodeIssue` fatality: design observations, unchanged and not blockers

atomb (approved):

MikaelMayer (approved):

PR #1096, opened 2026-05-01
feat: Add Split-Solve-Reconcile workflow for cloud-based SMT solving

olivier-aws:
## Summary

Add a Split-Solve-Reconcile (SSR) workflow that decouples VC generation from SMT solving, enabling parallel or cloud-based solver execution.

## Architecture

The key design decision is **manifest-free reconciliation**: instead of emitting a separate `manifest.json`, all obligation metadata is embedded directly in the `.smt2` files via SMT-LIB `set-info` directives. The reconcile phase parses these directives to reconstruct obligation metadata, then pairs each `.smt2` with its corresponding `.result` file.

### Three-phase workflow

1. **Generate** — `strata verify --no-solve --vc-directory ./vcs/` runs the full pipeline (parse, transform, symbolic eval, SMT encoding) and writes `.smt2` files with embedded metadata.
2. **Solve** — The user runs each `.smt2` file through a solver (locally, in parallel, or in the cloud) and captures stdout into a `.result` file.
3. **Reconcile** — `strata reconcile --vc-directory ./vcs/` reads `.smt2` files for obligation metadata, pairs them with `.result` files, and produces the final verification report.

### Embedded metadata (`set-info` directives)

Each `.smt2` file now includes:
- `(set-info :file "...")` — source file path
- `(set-info :start N)` / `(set-info :stop N)` — source location range
- `(set-info :final-message "...")` — obligation label/message
- `(set-info :property "...")` — property type (`assert`, `cover`, `divisionByZero`, `arithmeticOverflow`)
- `(set-info :resolved-sat "...")` / `(set-info :resolved-val "...")` — evaluator-resolved results (when the evaluator already decided a check before the solver ran)
- `(set-info :sat-message "...")` / `(set-info :unsat-message "...")` — presence indicates which checks were requested

### New files
- **`Strata/Languages/Core/Reconcile.lean`** — Reconcile module: parses `set-info` metadata from `.smt2` files, parses `.result` files, builds `VCResult`s using the shared `buildVCResult` function, and merges results via `mergeByAssertion`.
- **`Scripts/ssr_py.sh`** — End-to-end SSR helper script for Python analysis (generate → solve → reconcile).
- **`StrataTest/Languages/Python/run_py_ssr_test.sh`** — Integration test validating the SSR workflow against all Python test files.
- **`docs/design/SplitSolveReconcile.md`** — Design document.

### Modified files
- **`Strata/Languages/Core/Verifier.lean`**:
  - New `buildVCResult` function: single source of truth for turning `(satResult, valResult)` into a classified `VCResult`, used by both the integrated verifier and the reconcile path.
  - `encodeCore` / `dischargeObligation` / `getObligationResult`: extended with `property`, `resolvedSat`, `resolvedVal` parameters to emit `set-info` directives in `.smt2` files.
  - `verdictString` / `propertyString`: convert solver results and property types to short strings for `set-info` embedding.
  - `buildSolverLog`, `typedVarToSMTFn`: made public for reuse by reconcile.
  - `getObligationResult` refactored to use `buildVCResult`.
  - `verifySingleEnv`: threads evaluator-resolved results (`peSatResult?`, `peValResult?`) through to `getObligationResult`.
- **`Strata/SimpleAPI.lean`** — Minor refactor of `Core.verifyProgram` (do-notation).
- **`StrataMain.lean`** — New `reconcileCommand` CLI entry point; added to the Core command group. Namespace fix for `verify` → `Strata.verify`.
- **`StrataTest/Languages/Core/Tests/SMTEncoderTests.lean`** — Updated expected output to include the new `(set-info :property "assert")` directive.
- **`README.md`** — Documents the SSR workflow with usage examples.

## Testing

- Existing SMT encoder tests updated for new `set-info` output.
- Shell-based SSR integration test (`run_py_ssr_test.sh`) validates the full workflow end-to-end for Python test files: 202 passed, 0 failed, 14 skipped.


olivier-aws:
@keyboardDrummer-bot Please update the PR description and the design doc according to the latest commits in this PR

docs/design/SplitSolveAggregate.md
```diff
@@ -0,0 +1,256 @@
+# Design: Split-Solve-Reconcile for Cloud-Based SMT Solving
+
+## Motivation
+
+Strata's verification pipeline is bottlenecked by SMT solving, which can take
+seconds to minutes per query. For large programs, the total wall-clock time is
+dominated by sequential solver invocations. Cloud-based solving enables massive
+parallelism — all queries can be dispatched simultaneously — but requires
+decoupling the pipeline into three phases:
```
robin-aws, line 9: I like this phrasing of the motivation, but to nit-pick a bit, cloud-based solving could also be far more than just three phases. It could easily extend to a more general dynamically-constructed proof tree, with more iteration on proof goals.

I would just tweak this to point out that you need at least these three types of tasks, even if they aren't necessarily organized into three distinct phases (although they will be after this change)

docs/design/SplitSolveReconcile.md
```diff
@@ -0,0 +1,256 @@
...
+
+Strata's verification pipeline is bottlenecked by SMT solving, which can take
+seconds to minutes per query. For large programs, the total wall-clock time is
+dominated by sequential solver invocations. Cloud-based solving enables massive
+parallelism — all queries can be dispatched simultaneously — but requires
+decoupling the pipeline into three phases:
+
+1. **Generate** — run the Strata pipeline up to SMT file creation
+2. **Solve** — dispatch SMT queries to cloud solvers (external to Strata)
+3. **Reconcile** — read solver results and produce the final verification report
```
robin-aws: I don't love the term "reconcile" as it implies the guaranteed presence of some kind of conflicts. I would consider "report", "gather", "reduce" (as in Map/Reduce), "aggregate", etc.

docs/design/SplitSolveAggregate.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Key Decision: Manifest-Free Reconciliation
+
+Rather than emitting a separate `manifest.json` file alongside the `.smt2`
+files, all obligation metadata is embedded directly in each `.smt2` file using
+standard SMT-LIB `set-info` directives. This eliminates a separate serialization
+format, keeps each `.smt2` file self-contained, and avoids synchronization issues
+between manifest and SMT files.
+
+### Phase 1: Generate (`--no-solve`)
```
robin-aws, line 30: Some maintainers are understandably concerned that supporting this option (and this new command that really makes it a complete end-to-end feature) will be limiting in the future, both because it limits the cloud-based solving to a simple three-phase approach (see also my first comment) and because it will be more and more challenging to support new features, if they have to be able to capture all necessary state in `set-info` commands.

Can we mark both `--no-solve` and this new command as experimental, and/or something we might remove/completely rework in the future?

docs/design/SplitSolveAggregate.md
```diff
@@ -0,0 +1,256 @@
...
+Some obligations are trivially resolved by the evaluator (e.g., `assert true`
+→ valid). These still produce `.smt2` files (since the SMT encoding runs
+regardless), but the `resolved-sat` / `resolved-val` directives record the
+evaluator's verdict. The reconcile phase uses these stored results directly
+instead of relying on the solver output.
+
+### Phase 2: Solve (external)
+
+The user runs each `.smt2` file through a solver (locally, in the cloud, etc.)
+and captures the solver's stdout into a corresponding `.result` file:
```
robin-aws, line 74: For files with `resolved-sat` and `resolved-val` directives, is it legal to not run them and not create a `.result` file? Or are they expected to create the equivalent `.result` file? Or do they HAVE to call the solver anyway?

docs/design/SplitSolveAggregate.md
```diff
@@ -0,0 +1,256 @@
...
+```bash
+lake exe strata reconcile --vc-directory ./vcs/ [--check-mode deductive] [--check-level minimal] [--sarif]
+```
+
+This command:
+1. Scans the VC directory for `.smt2` files
+2. For each `.smt2` file:
+   - Parses `set-info` directives to extract obligation metadata (`SMT2Meta`)
+   - Determines which checks were requested (satisfiability, validity)
+   - If evaluator-resolved: uses the stored verdict directly
```
robin-aws, line 103: Ah here's the answer. Should be mentioned in the description of Phase 2 directly.

This isn't a strict either-or is it? Isn't it possible to resolve the sat check but not the validity check, or vice-versa?

docs/design/SplitSolveAggregate.md
```diff
@@ -0,0 +1,256 @@
...
+- `reconcileFromSMT2`: Combines parsed metadata and solver output into a
+  `VCResult`.
+- `reconcileDirectory`: Orchestrates the full reconcile over a directory.
+
+### `StrataMain.lean` — CLI Command
+
+The `reconcileCommand` accepts:
+- `--vc-directory` (required) — directory containing `.smt2` and `.result` files
+- `--check-mode` — verification mode (deductive, bugFinding, etc.)
+- `--check-level` — check level (minimal, minimalVerbose, full)
```
robin-aws, line 190: Is it sound to pass different values for these during phase 1 and phase 3?

docs/design/SplitSolveAggregate.md
```diff
@@ -0,0 +1,256 @@
...
+  validates SSR output matches direct verification for all Python test files.
+
+---
+
+## Design Decisions
+
+**Why embed metadata in `.smt2` files instead of a separate manifest?**
+A separate manifest introduces synchronization concerns (manifest and `.smt2`
+files can get out of sync), requires a custom JSON schema with serialization/
+deserialization code, and makes each `.smt2` file dependent on an external file
```
robin-aws, line 218: To be fair, we'll likely have to add more complex metadata in the future to support future features, and those will likely force using something like JSON or worse in individual `set-info` calls.

docs/design/SplitSolveReconcile.md
```diff
@@ -0,0 +1,256 @@
...
+self-contained, uses a standard SMT-LIB mechanism, and is ignored by solvers.
+
+**Why not re-run the pipeline during reconcile?**
+The symbolic evaluation step (path explosion, expression simplification) can be
+expensive for large programs. The embedded metadata avoids re-running it.
+
+**Why allow `checkMode`/`checkLevel` override in reconcile?**
+These only affect how outcomes are classified and displayed (pass/fail/warning).
+The underlying SMT results are mode-independent. This lets users re-classify
+results without re-solving — e.g., switch from deductive to bug-finding mode.
```
robin-aws: Ah that's fair, but in that case the initial generation phase shouldn't accept those options, if they have no effect until phase 3.

docs/design/SplitSolveReconcile.md
```diff
@@ -0,0 +1,256 @@
...
+
+**Why pass empty `phases` in reconcile?**
+The reconcile path calls `buildVCResult` with `phases = []` because phase
+validation (which demotes sat results to unknown for unvalidated abstractions)
+cannot be reconstructed from the `.smt2` metadata alone. This is acceptable
+because the current phase validators are all stubs that return `false`. When
+real validators are implemented, this will need revisiting.
```
robin-aws: High danger of not realizing this needs to change if that does happen. Can you add a meta-test that asserts they are all stubs with a comment describing what to do at that point?

docs/design/SplitSolveReconcile.md
```diff
@@ -0,0 +1,256 @@
...
+
+---
+
+## Risks and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|--------|------------|
+| Phase validators get real implementations | Reconcile cannot apply phase validation | Extend `set-info` metadata or add a lightweight sidecar file for phase decisions |
```
robin-aws: Previous comment is the simpler adequate mitigation IMO

docs/design/SplitSolveReconcile.md
```diff
@@ -0,0 +1,256 @@
...
+| SMT file naming collisions | Two obligations produce the same filename | Already handled: counter suffix (`_{N}`) ensures uniqueness |
+| Solver produces unexpected output format | Result parsing fails | Reuse existing verdict parsing; report clear errors for unparseable results |
+| User changes source between generate and reconcile | Source locations in `set-info` are stale | Document that generate and reconcile must use the same source |
```
robin-aws: I'm less worried about the source changing, and more about the `set-info` "schema" changing between generation and interpretation. I think we need to either ensure the changes are backwards compatible (which they are so far here) with mechanisms to enforce that, or add a schema version property of some kind that gets checked.

StrataTest/Languages/Python/run_py_ssr_test.sh
```diff
@@ -0,0 +1,152 @@
...
+      continue
+    fi
+    echo "FAIL (generate): $base_name"
+    failed=$((failed + 1))
+    rm -rf "$ssr_dir"
+    continue
+  fi
+
+  # Verify manifest was produced.
+  if [ ! -f "$ssr_dir/manifest.json" ]; then
```
robin-aws: I thought we didn't need a manifest?

StrataTest/Languages/Python/run_py_ssr_test.sh
robin-aws, line 1: Why doesn't this call `ssr_py.sh`?

Strata/SimpleAPI.lean
robin-aws, line 1: Pointless change?

StrataTest/Languages/Python/run_py_ssr_test.sh
```diff
@@ -0,0 +1,152 @@
...
+  # Capture both stdout and stderr; some tests produce expected errors.
+  generate_output=$("$STRATA" pyAnalyzeLaurel "$ion_file" \
+        --no-solve --vc-directory "$ssr_dir" --quiet 2>&1)
+  generate_rc=$?
+
+  if [ "$generate_rc" -ne 0 ]; then
+    # Check if the expected output indicates a known limitation or user error.
+    # These tests are expected to fail during generation — not an SSR issue.
+    if grep -qE 'Known limitation|User error|Internal error' "$expected_file"; then
+      echo "SKIP (expected error): $base_name"
```
robin-aws: Not how we should handle such errors. We need to record specifically what's expected for specific test inputs, not just silently swallow errors.

StrataTest/Languages/Python/run_py_ssr_test.sh
```diff
@@ -0,0 +1,152 @@
...
+#   2. Solves each .smt2 with cvc5
+#   3. Runs `strata reconcile` to produce the final report
+#   4. Compares the DETAIL/RESULT summary lines against the expected output
+#      from direct verification (the expected_laurel/*.expected files)
+#
+# Usage:
+#   ./run_py_ssr_test.sh [--filter <pattern>] [--solver <cmd>]
+# ------------------------------------------------------------------------------
+
+set -u
```
robin-aws: Off-topic issue: I'm concerned with how many shell scripts we're adding to this repository. LLMs love spitting them out, but they are also very sloppy about setting the right safety flags. We're also giving up on a great deal of safety by dropping from Lean to bash, even if we don't prove much about this kind of logic.

Scripts/ssr_py.sh
robin-aws, line 1: See comment below about bash scripts in general - this could easily be a `StrataMain.lean` command as well, even if it would need to shell-out for each phase.

robin-aws (changes requested):
Overall I like this approach to make `--no-solve` a complete feature. Added lots of high-level feedback, will take a closer look at the implementation once addressed.

docs/design/SplitSolveReconcile.md
```diff
@@ -0,0 +1,256 @@
...
+These directives are emitted by `encodeCore` in `Verifier.lean` at the end of
+the SMT encoding, after the `check-sat` commands. They are comments from the
+solver's perspective (solvers ignore unknown `set-info` keys) but are parsed by
+the reconcile phase.
+
+#### Evaluator-Resolved Obligations
+
+Some obligations are trivially resolved by the evaluator (e.g., `assert true`
+→ valid). These still produce `.smt2` files (since the SMT encoding runs
+regardless), but the `resolved-sat` / `resolved-val` directives record the
```
atomb: Note: generating SMT files for trivial obligations is optional.

Strata/Languages/Core/Verifier.lean
```diff
@@ -147,6 +158,21 @@ open Lambda Strata.SMT
 
 public section
 
+/-- Short verdict string for embedding in SMT2 `set-info` directives. -/
```
atomb: Aren't there already functions to convert `Imperative.SMT.Result` and `Imperative.PropertyType` to identical strings?

docs/design/SplitSolveReconcile.md
robin-aws: It seems like it might be required for this workflow to work, since the results need to be input into the reconcile phase somehow. i.e. we don't need SMT files, but we do need to remember the results somehow, and the PR currently only uses SMT files to remember anything. If you see a cleaner way, possibly by recording metadata separately (but not to the same extent as the earlier version of this PR did), we should consider it!

atomb: Oh, yes, I agree. Always generating SMT files for this workflow seems like the right way to go. The specific wording of the comment just implies that that's the only approach that's implemented.

PR #1097, opened 2026-05-01
Add --skip-verification flag to pyAnalyzeLaurel

joehendrix:
This adds a `--skip-verification` flag to the `pyAnalyzeLaurel` command that bypasses all the steps after converting to Core translation.  This flag enables fast detection of type-checker and translation bugs without that overhead of a full verification pass.

- Adds a `--skip-verification` boolean flag to the `pyAnalyzeLaurel` command
- When set, runs the full Python→Laurel→Core translation pipeline (including all 11 Laurel lowering passes) but skips SMT verification
- Reports translation diagnostics via the existing `RESULT`/`DETAIL` output format
- SARIF output is written (with empty VC results) when `--sarif` is also set, so the flag is a drop-in addition to existing invocations
- SMT-only flags (`--check-mode`, `--entry-point`, `--solver`, `--timeout`) are silently ignored

### DIFF
--- a/StrataMain.lean
+++ b/StrataMain.lean
@@ -571,7 +571,10 @@ def pyAnalyzeLaurelCommand : Command where
               takesArg := .arg "mode" },
             { name := "warning-summary",
               help := "Write PySpec warning summary as JSON to <file>.",
-              takesArg := .arg "file" }]
+              takesArg := .arg "file" },
+            { name := "skip-verification",
+              help := "Run Python-to-Laurel and Laurel-to-Core translation only (skip SMT verification).",
+              takesArg := .none }]
   help := "Verify a Python Ion program via the Laurel pipeline. Translates Python to Laurel to Core, then runs SMT verification."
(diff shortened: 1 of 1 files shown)


StrataMain.lean
```diff
@@ -650,6 +653,31 @@ def pyAnalyzeLaurelCommand : Command where
       IO.println "\n==== Core Program ===="
       IO.print (Core.formatProgram coreProgram)
 
+    -- When --skip-verification is set, report translation diagnostics and exit
+    -- without running SMT verification (stages 3-4).
```
tautschnig, line 657: Can we please have at least one test that shows that this works and what this looks like?

tautschnig (approved):

aqjune-aws (approved):

StrataMain.lean
tautschnig: #1129 for tracking.

PR #1100, opened 2026-05-02
Core: Sequence bounds preconditions

fabiomadge:
Blocker for Laurel Seq<T>/Array<T> in #1073.

## Bounds preconditions for `Sequence.select` / `update` / `take` / `drop`

Following the `Int.SafeDiv` pattern:

| Op                  | Precondition                        |
|---------------------|-------------------------------------|
| `Sequence.select`   | `0 <= i && i < Sequence.length(s)`  |
| `Sequence.update`   | `0 <= i && i < Sequence.length(s)`  |
| `Sequence.take`     | `0 <= n && n <= Sequence.length(s)` |
| `Sequence.drop`     | `0 <= n && n <= Sequence.length(s)` |

`Sequence.length` / `empty` / `append` / `build` / `contains` remain total.

`PrecondElim` picks the obligations up automatically at call sites in imperative code (via `transformStmt`) and in pure positions like `requires` / `ensures` / quantifier bodies / function bodies (via the synthetic `$$wf` procedures).

Obligations carry `propertyType = "outOfBoundsAccess"` (new `MetaData` constant, mirroring `divisionByZero`), flow through a new `PropertyType.outOfBoundsAccess` enum variant and the three metadata-to-PropertyType conversion sites, and render as `"out-of-bounds-access"` in SARIF output.

### Incidental fix

While wiring the SARIF classification, I noticed `propertyTypeToClassification` in `SarifOutput.lean` was pre-existing dead code: `vcResultToSarifResult` never set `properties.propertyType`, so every obligation defaulted to `"assert"` in SARIF output. This PR wires it up, so `divisionByZero` and `arithmeticOverflow` obligations now also classify correctly in SARIF alongside the new `outOfBoundsAccess`.

## Testing

- **`StrataTest/Transform/PrecondElim.lean` Test 10** — `collectPrecondAsserts` attaches `outOfBoundsAccess` metadata for all four partial ops plus a nested `Sequence.select(Sequence.update(...))` call. Mirrors the pattern in `OverflowCheckTest.lean`. Complemented by per-op pretty-print snapshots (`#guard_msgs`) that assert the exact obligation body string — these catch regressions that preserve count and metadata tag but corrupt the obligation body (e.g. passing `.Le` instead of `.Lt` to `mkSeqBoundsPrecond` at a call site, changing the bound-variable name, or swapping the lower/upper bound inside `mkSeqBoundsPrecond`).
- **`StrataTest/Languages/Core/Tests/SarifOutputTests.lean`** — property-type classification tests covering all five `PropertyType` variants.

Collateral updates to existing tests reflect the new obligations (`Seq.lean`) and updated `requires` on Sequence function signatures (`ProgramEvalTests.lean`). Note: the bounds obligations in `Seq.lean` appear as `true && 0 < length(...)` — the partial evaluator simplifies `0 <= 0` to `true` but does not further simplify `true && X` to `X`. This is a pre-existing evaluator gap, not newly introduced by this PR; the SMT solver discharges the obligation trivially.

## Known downstream impact

PR #1073 (Laurel Seq<T>/Array<T>) emits `Sequence.select` / `update` / `take` / `drop` calls from its translation. Its `T18_Sequences` test uses `#guard_msgs(drop info, error)` on a diagnostics-only pipeline, so the test assertions should still pass syntactically — but individual sequence-manipulation programs now require bounds guards to fully verify. PR #1073 has been adjusted accordingly.

### DIFF
--- a/Strata/DL/Imperative/CmdEval.lean
+++ b/Strata/DL/Imperative/CmdEval.lean
@@ -66,11 +66,7 @@ def Cmd.eval [BEq P.Ident] [EC : EvalContext P S] (σ : S) (c : Cmd P) : Cmd P �
       let e := EC.eval σ e
       let assumptions := EC.getPathConditions σ
       let c' := .assert label e md
-      let propType := match md.getPropertyType with
-        | some s => if s == MetaData.divisionByZero then .divisionByZero
-                    else if s == MetaData.arithmeticOverflow then .arithmeticOverflow
-                    else .assert
-        | none => .assert
+      let propType := convertMetaDataPropertyType md
       match EC.denoteBool e with
       | some true => -- Proved via evaluation.
         (c', EC.deferObligation σ (ProofObligation.mk label propType assumptions e md))
--- a/Strata/DL/Imperative/EvalContext.lean
+++ b/Strata/DL/Imperative/EvalContext.lean
@@ -102,12 +102,13 @@ inductive PropertyType where
   | assert
   | divisionByZero
   | arithmeticOverflow
+  | outOfBoundsAccess
   deriving Repr, DecidableEq
 
 /-- Whether an unreachable path counts as pass for this property type.
     Assertions pass vacuously when unreachable; covers fail. -/
 def PropertyType.passWhenUnreachable : PropertyType → Bool
-  | .assert | .divisionByZero | .arithmeticOverflow => true
+  | .assert | .divisionByZero | .arithmeticOverflow | .outOfBoundsAccess => true
   | .cover => false
 
 instance : ToFormat PropertyType where
@@ -116,6 +117,21 @@ instance : ToFormat PropertyType where
     | .assert => "assert"
     | .divisionByZero => "division by zero check"
     | .arithmeticOverflow => "arithmetic overflow check"
+    | .outOfBoundsAccess => "out-of-bounds access check"
+
+/-- Convert a `MetaData` entry's property-type classification string to the
+    `PropertyType` enum. Falls back to `.assert` when the metadata carries
+    no classification or an unrecognized string; callers that emit
+    propertyType classifications should add a matching arm here. -/
+def convertMetaDataPropertyType {P : PureExpr} [BEq P.Ident]
+    (md : MetaData P) : PropertyType :=
+  match md.getPropertyType with
+  | some s =>
+    if s == MetaData.divisionByZero then .divisionByZero
+    else if s == MetaData.arithmeticOverflow then .arithmeticOverflow
+    else if s == MetaData.outOfBoundsAccess then .outOfBoundsAccess
+    else .assert
+  | none => .assert
 
 /--
 A proof obligation can be discharged by some backend solver or a dedicated
--- a/Strata/Languages/Core/Factory.lean
+++ b/Strata/Languages/Core/Factory.lean
@@ -499,10 +499,53 @@ def seqAppendFunc : WFLFunc CoreLParams :=
               else #true))]
     ])
 
-/- A `Sequence` selection function with type `∀a. Sequence a → int → a`. -/
+/-! ### Sequence bounds preconditions
+
+`Sequence.select` / `update` / `take` / `drop` carry bounds
+preconditions; the other `Sequence.*` ops are total. -/
+
+/-- Choice of upper-bound operator in `mkSeqBoundsPrecond`: `Lt` (strict) for
+    `Sequence.select`/`update`, `Le` (non-strict) for `Sequence.take`/`drop`.
+    Restricting the parameter to this inductive rather than taking an
+    arbitrary `WFLFunc` or `LExpr` makes it impossible to attach a partial
+    operator (which would create a nested precondition obligation) by
+    accident. -/
+private inductive SeqBoundKind where | Lt | Le
+
+/-- Returns the *upper-bound* comparison for `mkSeqBoundsPrecond`.
+    The lower bound is always `0 ≤ x` (see `mkSeqBoundsPrecond`), so this
+    method characterises only the upper comparison. A future partial
+    Sequence op requiring a non-`int` comparison (e.g. a bitvector variant)
+    should introduce a separate helper rather than extend this enum. -/
+private def SeqBoundKind.upperOpExpr : SeqBoundKind → LExpr CoreLParams.mono
+  | .Lt => (intLtFunc (T := CoreLParams)).opExpr
+  | .Le => (intLeFunc (T := CoreLParams)).opExpr
+
+/-- Precondition `0 <= varName && varName `k.upperOpExpr` Sequence.length(s)`.
+
+    Assumes the enclosing function names its `Sequence a` input `"s"`.
+    Mismatches are caught at elaboration by `polyUneval`'s `h_precond`
+    free-vars check. -/
+private def mkSeqBoundsPrecond
+    (varName : String) (k : SeqBoundKind) :
+    Strata.DL.Util.FuncPrecondition (LExpr CoreLParams.mono) CoreLParams.Metadata :=
+  let sVar  : LExpr CoreLParams.mono := .fvar default "s" (some (seqTy mty[%a]))
+  let xVar  : LExpr CoreLParams.mono := .fvar default varName (some mty[int])
+  let zero  : LExpr CoreLParams.mono := .intConst default 0
+  let lenS  : LExpr CoreLParams.mono := .app default seqLengthFunc.opExpr sVar
+  let lower : LExpr CoreLParams.mono :=
+    .app default (.app default (intLeFunc (T := CoreLParams)).opExpr zero) xVar
+  let upper : LExpr CoreLParams.mono :=
+    .app default (.app default k.upperOpExpr xVar) lenS
+  ⟨.app default (.app default (boolAndFunc (T := CoreLParams)).opExpr lower) upper,
+   default⟩
+
+/- A `Sequence` selection function with type `∀a. Sequence a → int → a`.
+   Partial: requires `0 <= i && i < Sequence.length(s)`. -/
 def seqSelectFunc : WFLFunc CoreLParams :=
   polyUneval "Sequence.select" ["a"]
     [("s", seqTy mty[%a]), ("i", mty[int])] mty[%a]
+    (preconditions := [mkSeqBoundsPrecond "i" .Lt])
 
 /- A `Sequence` build (snoc) function with type `∀a. Sequence a → a → Sequence a`.
    `build(s, v)` appends a single element `v` to the end of `s`. -/
@@ -555,7 +598,8 @@ def seqBuildFunc : WFLFunc CoreLParams :=
     ])
 
 /- A `Sequence` update function with type `∀a. Sequence a → int → a → Sequence a`.
-   `update(s, i, v)` returns a sequence identical to `s` except at index `i` where the value is `v`. -/
+   `update(s, i, v)` returns a sequence identical to `s` except at index `i` where the value is `v`.
+   Partial: requires `0 <= i && i < Sequence.length(s)`. -/
 def seqUpdateFunc : WFLFunc CoreLParams :=
   polyUneval "Sequence.update" ["a"]
     [("s", seqTy mty[%a]), ("i", mty[int]), ("v", mty[%a])]
@@ -606,6 +650,7 @@ def seqUpdateFunc : WFLFunc CoreLParams :=
                   (((~Sequence.select : (Sequence %a) → int → %a) %3) %0)
                 else #true)))]
     ])
+    (preconditions := [mkSeqBoundsPrecond "i" .Lt])
 
 /- A `Sequence` contains function with type `∀a. Sequence a → a → bool`.
    `contains(s, v)` is true iff there exists an index `i` such that `select(s, i) == v`. -/
@@ -628,7 +673,8 @@ def seqContainsFunc : WFLFunc CoreLParams :=
     ])
 
 /- A `Sequence` take function with type `∀a. Sequence a → int → Sequence a`.
-   `take(s, n)` returns the first `n` elements of `s`. -/
+   `take(s, n)` returns the first `n` elements of `s`.
+   Partial: requires `0 <= n && n <= Sequence.length(s)`. -/
 def seqTakeFunc : WFLFunc CoreLParams :=
   polyUneval "Sequence.take" ["a"]
     [("s", seqTy mty[%a]), ("n", mty[int])]
@@ -664,9 +710,11 @@ def seqTakeFunc : WFLFunc CoreLParams :=
                 (((~Sequence.select : (Sequence %a) → int → %a) %2) %0)
               else #true))]
     ])
+    (preconditions := [mkSeqBoundsPrecond "n" .Le])
 
 /- A `Sequence` drop function with type `∀a. Sequence a → int → Sequence a`.
-   `drop(s, n)` returns the sequence with the first `n` elements removed. -/
+   `drop(s, n)` returns the sequence with the first `n` elements removed.
+   Partial: requires `0 <= n && n <= Sequence.length(s)`. -/
 def seqDropFunc : WFLFunc CoreLParams :=
   polyUneval "Sequence.drop" ["a"]
     [("s", seqTy mty[%a]), ("n", mty[int])]
@@ -709,6 +757,7 @@ def seqDropFunc : WFLFunc CoreLParams :=
                     (((~Int.Add : int → int → int) %0) %1))
               else #true))]
     ])
+    (preconditions := [mkSeqBoundsPrecond "n" .Le])
 
 def emptyTriggersFunc : WFLFunc CoreLParams :=
   nullaryUneval "Triggers.empty" mty[Triggers]
--- a/Strata/Languages/Core/Factory.lean
+++ b/Strata/Languages/Core/Factory.lean
@@ -521,15 +521,19 @@ private def SeqBoundKind.upperOpExpr : SeqBoundKind → LExpr CoreLParams.mono
   | .Lt => (intLtFunc (T := CoreLParams)).opExpr
   | .Le => (intLeFunc (T := CoreLParams)).opExpr
 
-/-- Precondition `0 <= varName && varName `k.upperOpExpr` Sequence.length(s)`.
+/-- Precondition `0 <= varName && varName `k.upperOpExpr` Sequence.length(seqName)`.
 
-    Assumes the enclosing function names its `Sequence a` input `"s"`.
-    Mismatches are caught at elaboration by `polyUneval`'s `h_precond`
-    free-vars check. -/
+    `seqName` defaults to `"s"` since all four current call sites
+    (`Sequence.select`/`update`/`take`/`drop`) name their `Sequence a` input
+    that way. The parameter exists so a future partial Sequence op with a
+    different input name need only pass it explicitly rather than rely on a
+    hidden string literal. Either way, mismatches between the function's
+    declared inputs and the names used here are caught at elaboration by
+    `polyUneval`'s `h_precond` free-vars check. -/
 private def mkSeqBoundsPrecond
-    (varName : String) (k : SeqBoundKind) :
+    (varName : String) (k : SeqBoundKind) (seqName : String := "s") :
     Strata.DL.Util.FuncPrecondition (LExpr CoreLParams.mono) CoreLParams.Metadata :=
-  let sVar  : LExpr CoreLParams.mono := .fvar default "s" (some (seqTy mty[%a]))
+  let sVar  : LExpr CoreLParams.mono := .fvar default seqName (some (seqTy mty[%a]))
   let xVar  : LExpr CoreLParams.mono := .fvar default varName (some mty[int])
   let zero  : LExpr CoreLParams.mono := .intConst default 0
   let lenS  : LExpr CoreLParams.mono := .app default seqLengthFunc.opExpr sVar
--- a/Strata/Languages/Core/ObligationExtraction.lean
+++ b/Strata/Languages/Core/ObligationExtraction.lean
@@ -55,11 +55,7 @@ def extractGo (pc : PathConditions Expression) : Statements →
   | s :: rest, acc =>
     match s with
     | .cmd (.cmd (.assert label e md)) =>
-      let propType := match md.getPropertyType with
-        | some s => if s == MetaData.divisionByZero then .divisionByZero
-                    else if s == MetaData.arithmeticOverflow then .arithmeticOverflow
-                    else .assert
-        | none => .assert
+      let propType := convertMetaDataPropertyType md
       extractGo pc rest (acc.push (ProofObligation.mk label propType pc e md))
 
     | .cmd (.cmd (.cover label e md)) =>
--- a/Strata/Languages/Core/StatementEval.lean
+++ b/Strata/Languages/Core/StatementEval.lean
@@ -320,11 +320,7 @@ private def createUnreachableAssertObligations
     Imperative.ProofObligations Expression :=
   asserts.toArray.map
     (fun (label, md) =>
-      let propType := match md.getPropertyType with
-        | some s => if s == Imperative.MetaData.divisionByZero then .divisionByZero
-                    else if s == Imperative.MetaData.arithmeticOverflow then .arithmeticOverflow
-                    else .assert
-        | _ => .assert
+      let propType := Imperative.convertMetaDataPropertyType md
       (Imperative.ProofObligation.mk label propType pathConditions (LExpr.true ()) md))
 
 /--
--- a/Strata/Transform/PrecondElim.lean
+++ b/Strata/Transform/PrecondElim.lean
@@ -76,6 +76,8 @@ private def classifyPrecondition (funcName : String) (precondIdx : Nat := 0) : O
   | .bv ⟨_, .SafeAdd⟩ | .bv ⟨_, .SafeSub⟩ | .bv ⟨_, .SafeMul⟩ | .bv ⟨_, .SafeNeg⟩
   | .bv ⟨_, .SafeUAdd⟩ | .bv ⟨_, .SafeUSub⟩ | .bv ⟨_, .SafeUMul⟩ | .bv ⟨_, .SafeUNeg⟩ =>
     some Imperative.MetaData.arithmeticOverflow
+  | .seq .Select | .seq .Update | .seq .Take | .seq .Drop =>
+    some Imperative.MetaData.outOfBoundsAccess
   | _ => none
 
 /--
--- a/StrataTest/Transform/PrecondElim.lean
+++ b/StrataTest/Transform/PrecondElim.lean
@@ -418,4 +418,80 @@ procedure test (inout g : int, y : int)
 #guard_msgs in
 #eval (Std.format (transformProgram loopGuardPrecondPgm))
 
+/-! ### Test 10: `collectPrecondAsserts` tags Sequence bounds obligations with `outOfBoundsAccess`
+
+Exercises the full `collectPrecondAsserts` path — the code called by
+`transformStmt` / `mkContractWFProc` / `mkFuncWFProc` — and inspects the
+metadata on the generated assert. Mirrors `OverflowCheckTest.lean`. -/
+
+section SeqBoundsObligations
+
+open Strata Core Lambda Core.PrecondElim Imperative
+
+/-- Shared fvar fixtures so each per-op case below is a one-liner. -/
+private def fxS : Core.Expression.Expr := .fvar () ⟨"s", ()⟩ (some (Core.seqTy .int))
+private def fxI : Core.Expression.Expr := .fvar () ⟨"i", ()⟩ (some .int)
+private def fxV : Core.Expression.Expr := .fvar () ⟨"v", ()⟩ (some .int)
+private def fxN : Core.Expression.Expr := .fvar () ⟨"n", ()⟩ (some .int)
+private def fxJ : Core.Expression.Expr := .fvar () ⟨"j", ()⟩ (some .int)
+
+/-- Check that `collectPrecondAsserts` produces exactly `expectedCount`
+    obligations from `expr`, each tagged with `outOfBoundsAccess`. -/
+private def assertOutOfBoundsObligations
+    (expr : Core.Expression.Expr) (expectedCount : Nat) : IO Unit := do
+  let stmts := collectPrecondAsserts Core.Factory expr "test" #[]
+  assert! stmts.length == expectedCount
+  for s in stmts do
+    let md : MetaData Core.Expression := match s with
+      | Statement.assert _ _ md => md | _ => #[]
+    assert! md.getPropertyType == some MetaData.outOfBoundsAccess
+
+-- Sequence.select / update / take / drop each emit one out-of-bounds obligation.
+#eval assertOutOfBoundsObligations (LExpr.mkApp () Core.seqSelectOp [fxS, fxI]) 1
+#eval assertOutOfBoundsObligations (LExpr.mkApp () Core.seqUpdateOp [fxS, fxI, fxV]) 1
+#eval assertOutOfBoundsObligations (LExpr.mkApp () Core.seqTakeOp   [fxS, fxN]) 1
+#eval assertOutOfBoundsObligations (LExpr.mkApp () Core.seqDropOp   [fxS, fxN]) 1
+
+-- Nested: `Sequence.select(Sequence.update(s, i, v), j)` emits two
+-- obligations (one per partial call), both tagged `outOfBoundsAccess`.
+#eval assertOutOfBoundsObligations
+  (LExpr.mkApp () Core.seqSelectOp [LExpr.mkApp () Core.seqUpdateOp [fxS, fxI, fxV], fxJ]) 2
+
+-- Sequence.length is total: no precondition obligations generated.
+#eval do
+  let stmts := collectPrecondAsserts Core.Factory
+    (LExpr.mkApp () Core.seqLengthOp [fxS]) "test" #[]
+  assert! stmts.isEmpty
+
+/-! #### Test 10a: Pretty-printed obligation shape per partial op
+
+Catches regressions that preserve count and metadata tag but corrupt the
+obligation body (e.g. swapping `.Lt` for `.Le` at a call site, changing
+the bound variable name, or swapping the lower/upper bound inside
+`mkSeqBoundsPrecond`). -/
+
+private def printFirstObligation (expr : Core.Expression.Expr) : IO Unit := do
+  let stmts := collectPrecondAsserts Core.Factory expr "test" #[]
+  match stmts.head? with
+  | some (Statement.assert _ e _) => IO.println s!"{Std.format e}"
+  | _ => IO.println "<unexpected>"
+
+/-- info: 0 <= i && i < Sequence.length(s) -/
+#guard_msgs in
+#eval printFirstObligation (LExpr.mkApp () Core.seqSelectOp [fxS, fxI])
+
+/-- info: 0 <= i && i < Sequence.length(s) -/
+#guard_msgs in
+#eval printFirstObligation (LExpr.mkApp () Core.seqUpdateOp [fxS, fxI, fxV])
+
+/-- info: 0 <= n && n <= Sequence.length(s) -/
+#guard_msgs in
+#eval printFirstObligation (LExpr.mkApp () Core.seqTakeOp [fxS, fxN])
+
+/-- info: 0 <= n && n <= Sequence.length(s) -/
+#guard_msgs in
+#eval printFirstObligation (LExpr.mkApp () Core.seqDropOp [fxS, fxN])
+
+end SeqBoundsObligations
+
 end PrecondElimTests
--- a/Strata/DL/Imperative/MetaData.lean
+++ b/Strata/DL/Imperative/MetaData.lean
@@ -320,6 +320,9 @@ def MetaData.divisionByZero : String := "divisionByZero"
 /-- Metadata value for arithmetic-overflow property type classification. -/
 def MetaData.arithmeticOverflow : String := "arithmeticOverflow"
 
+/-- Metadata value for out-of-bounds-access property type classification. -/
+def MetaData.outOfBoundsAccess : String := "outOfBoundsAccess"
+
 /-- Read the property type classification from metadata, if present. -/
 def MetaData.getPropertyType {P : PureExpr} [BEq P.Ident] (md : MetaData P) : Option String :=
   match md.findElem MetaData.propertyType with
--- a/Strata/DL/Lambda/IntBoolFactory.lean
+++ b/Strata/DL/Lambda/IntBoolFactory.lean
@@ -120,22 +120,27 @@ instance (n : Nat) : LambdaLeanType (.bitvec n) (BitVec n) where
 
 These build well-formed `WFLFunc`s that have no `concreteEval` or `body`. -/
 
-/-- General polymorphic unevaluated function with optional axioms.
-    Handles any arity and any number of type arguments. -/
+/-- General polymorphic unevaluated function with optional axioms and
+    preconditions. Handles any arity and any number of type arguments. -/
 @[inline]
 def polyUneval (n : T.Identifier) (typeArgs : List String)
     (inputs : List (T.Identifier × LMonoTy)) (output : LMonoTy)
     (axioms : List (LExpr T.mono) := [])
+    (preconditions :
+      List (FuncPrecondition (LExpr T.mono) T.Metadata) := [])
     (h_nodup : List.Nodup (inputs.map (·.1.name)) := by first | decide | grind)
     (h_ta_nodup : List.Nodup typeArgs := by grind)
     (h_inputs : ∀ ty, ty ∈ ListMap.values inputs →
       ty.freeVars ⊆ typeArgs := by first | decide | grind)
     (h_output : output.freeVars ⊆ typeArgs
       := by first | decide | grind)
+    (h_precond : ∀ p, p ∈ preconditions →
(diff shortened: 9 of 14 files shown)


tautschnig (dismissed):

Strata/Languages/Core/DDMTransform/FormatCore.lean
```diff
@@ -295,10 +295,12 @@ def handleZeroaryOps {M} [Inhabited M] (name : String)
   | .re .All => pure (.re_all default)
   | .re .AllChar => pure (.re_allchar default)
   | .re .None => pure (.re_none default)
-  -- TODO: seq_empty is not yet parseable (see Grammar.lean); handle here when added.
-  | _ => do
-    ToCSTM.logError "lopToExpr" "0-ary op not found" name
-    pure (.re_none default)
+  -- Any other 0-ary op (e.g. `Sequence.empty`, which is polymorphic and not
+  -- yet parseable in raw Core source) is rendered as a generic call. This
+  -- matches the fallback already used by `handleUnaryOps` and
```
joscoh: Comments should be stateless, should not mention "already used" and does not need to mention `re.none()`, as that is no longer relevant. I think we should also keep the TODO for `seq_empty`.

Strata/Languages/Core/Factory.lean
```diff
@@ -499,10 +499,38 @@ def seqAppendFunc : WFLFunc CoreLParams :=
               else #true))]
     ])
 
-/- A `Sequence` selection function with type `∀a. Sequence a → int → a`. -/
+/-! ### Sequence bounds preconditions
+
+`Sequence.select` / `update` / `take` / `drop` carry bounds
+preconditions; the other `Sequence.*` ops are total. PrecondElim picks
+the obligations up via `collectPrecondAsserts`, classifying them as
```
joscoh: Minor, but I don't think this needs to discuss `PrecondElim` here. Here we are defining the functions with preconditions, how they are translated to SMT is an implementation detail separate from the function definitions.

Strata/Languages/Core/Factory.lean
```diff
@@ -499,10 +499,38 @@ def seqAppendFunc : WFLFunc CoreLParams :=
...
+`outOfBoundsAccess` (see
+`Strata.Transform.PrecondElim.classifyPrecondition`). -/
+
+/-- Precondition `0 <= varName && varName `upperOp` Sequence.length(s)`.
+    `upperOp` must be a total `int → int → bool` operation: `intLtFunc`
```
joscoh: Where is it enforced/required that `upperOp` is total?

Strata/Transform/PrecondElim.lean
```diff
@@ -63,8 +63,11 @@ def wfProcName (name : String) : String := s!"{name}{wfSuffix}"
 
 /-- Classify a function precondition into a property type for SARIF reporting.
     For functions with multiple preconditions (e.g., SafeSDiv has both div-by-zero
-    and overflow), the precondition index distinguishes them. -/
-private def classifyPrecondition (funcName : String) (precondIdx : Nat := 0) : Option String :=
+    and overflow), the precondition index distinguishes them.
+
+    Exposed (not `private`) so that tests can verify the classification for each
```
joscoh: This does not need to be a in a comment. You can just change the visibility

StrataTest/Transform/PrecondElim.lean
```diff
@@ -418,4 +418,202 @@ procedure test (inout g : int, y : int)
...
+procedure test (s : Sequence int, i : int)
+{
+  assert [init_calls_Sequence.select_0]: 0 <= i && i < Sequence.length(s);
+  var x : int := Sequence.select(s, i);
+};
+-/
+#guard_msgs in
+#eval (Std.format (transformProgram seqSelectNoGuardPgm))
+
+/-! ### Test 10c: Sequence.select inside a requires clause generates a $$wf procedure
```
joscoh: Why does the numbering go from 10 to 10c?

StrataTest/Transform/PrecondElim.lean
```diff
@@ -418,4 +418,202 @@ procedure test (inout g : int, y : int)
 #guard_msgs in
 #eval (Std.format (transformProgram loopGuardPrecondPgm))
 
+/-! ### Test 10: Sequence.select without a bounds guard produces an out-of-bounds obligation
```
joscoh: I don't think these tests are actually testing anything new. The behavior of `PrecondElim` has not changed at all, this is just duplicating above tests but using a different function (`Sequence.select` instead of `safeDiv` or a safe destructor, etc).

StrataTest/Transform/PrecondElim.lean
```diff
@@ -418,4 +418,202 @@ procedure test (inout g : int, y : int)
...
+  assert! stmts.length == 2
+  -- Both obligations classified as outOfBoundsAccess.
+  let md0 : MetaData Core.Expression := match stmts[0]! with
+    | Statement.assert _ _ md => md | _ => #[]
+  assert! md0.getPropertyType == some MetaData.outOfBoundsAccess
+  let md1 : MetaData Core.Expression := match stmts[1]! with
+    | Statement.assert _ _ md => md | _ => #[]
+  assert! md1.getPropertyType == some MetaData.outOfBoundsAccess
+
+/-! ### Test 12: `Sequence.empty` printer regression
```
joscoh: This is not a precondition-related test. It should be moved to a `FormatTests` or similar.

Strata/Languages/Core/DDMTransform/FormatCore.lean
fabiomadge: Done — rewrote the fallback comment to be stateless and restored the `Sequence.empty` TODO.

Strata/Languages/Core/Factory.lean
fabiomadge: Done — dropped the `PrecondElim` / `classifyPrecondition` references. Section now just describes what is defined here.

Strata/Languages/Core/Factory.lean
fabiomadge: Fair point — nothing enforced it before. Refactored to take a private `SeqBoundKind` inductive (`.Lt` / `.Le`) instead of an arbitrary `WFLFunc`. That rules out accidentally passing a partial operator at the type level, and the two call-site spellings become `.Lt` / `.Le`.

Strata/Transform/PrecondElim.lean
fabiomadge: Done — reverted to `private` and dropped the stale "Exposed" docstring. No test actually needs it public (the metadata-check test uses `collectPrecondAsserts` + metadata inspection instead).

StrataTest/Transform/PrecondElim.lean
fabiomadge: Resolved by the fix to the next comment — Tests 10 / 10c / 10d were dropped, and the remaining Sequence-specific test is now renumbered to Test 10 so there is no gap.

StrataTest/Transform/PrecondElim.lean
fabiomadge: You are right. Tests 10, 10c, and 10d all reproduced pre-existing path coverage (Test 1 for var-init, Test 3 for `requires`, Test 5 for function body) with `Sequence.select` substituted for `safeDiv`. Dropped all three.

Kept the metadata-propagation test (now renumbered to Test 10) because it tests something genuinely new — that `collectPrecondAsserts` attaches the `outOfBoundsAccess` classification for all four partial ops and for nested calls. End-to-end coverage of the user-visible output lives in `StrataTest/Languages/Core/Examples/Seq.lean`.

StrataTest/Transform/PrecondElim.lean
fabiomadge: Done — moved to `StrataTest/Languages/Core/Tests/GenericCallFallbackTest.lean` following the existing `handleUnaryOps` / `handleBinaryOps` / `handleTernaryOps` / `lopToExpr` test pattern. It is the missing `handleZeroaryOps` case in that family.

Strata/DL/Imperative/CmdEval.lean
```diff
@@ -69,6 +69,7 @@ def Cmd.eval [BEq P.Ident] [EC : EvalContext P S] (σ : S) (c : Cmd P) : Cmd P 
       let propType := match md.getPropertyType with
         | some s => if s == MetaData.divisionByZero then .divisionByZero
                     else if s == MetaData.arithmeticOverflow then .arithmeticOverflow
+                    else if s == MetaData.outOfBoundsAccess then .outOfBoundsAccess
```
atomb: This PR didn't introduce this conditional chain, but I think it'd be nicer to have it in a separate function with a name something like `convertMetaDataPropertyType`.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -323,6 +323,7 @@ private def createUnreachableAssertObligations
       let propType := match md.getPropertyType with
         | some s => if s == Imperative.MetaData.divisionByZero then .divisionByZero
                     else if s == Imperative.MetaData.arithmeticOverflow then .arithmeticOverflow
+                    else if s == Imperative.MetaData.outOfBoundsAccess then .outOfBoundsAccess
```
atomb: Oh, this shows up twice! All the more reason to make it a separate function.

Strata/Languages/Core/ObligationExtraction.lean
```diff
@@ -58,6 +58,7 @@ def extractGo (pc : PathConditions Expression) : Statements →
       let propType := match md.getPropertyType with
         | some s => if s == MetaData.divisionByZero then .divisionByZero
                     else if s == MetaData.arithmeticOverflow then .arithmeticOverflow
+                    else if s == MetaData.outOfBoundsAccess then .outOfBoundsAccess
```
atomb: Oh, three times!

StrataTest/Transform/PrecondElim.lean
atomb, line 1: It's a minor thing, but these tests are very repetitive. Shared logic could be abstracted out.

atomb (dismissed):
This looks nice. There's a little room to reduce repetition, however.

Strata/DL/Imperative/CmdEval.lean
fabiomadge: Good suggestion — factored out `convertMetaDataPropertyType` in `EvalContext.lean` and replaced this and the two other sites with a call to it.

Strata/Languages/Core/StatementEval.lean
fabiomadge: This site now calls the shared `convertMetaDataPropertyType` as well.

Strata/Languages/Core/ObligationExtraction.lean
fabiomadge: Third site also replaced with the shared helper.

StrataTest/Transform/PrecondElim.lean
fabiomadge: Refactored — added shared `fx*` fvar fixtures and an `assertOutOfBoundsObligations` helper, so each per-op case is now a one-liner `#eval`.

Strata/Transform/PrecondElim.lean
```diff
@@ -63,8 +63,11 @@ def wfProcName (name : String) : String := s!"{name}{wfSuffix}"
...
+    partial op without having to inspect metadata on generated asserts. -/
+def classifyPrecondition (funcName : String) (precondIdx : Nat := 0) : Option String :=
```
tautschnig: Carryover from joscoh's thread at this location (https://github.com/strata-org/Strata/pull/1100#discussion_r3180728...). The reply says:

> Done — reverted to `private` and dropped the stale "Exposed" docstring. No test actually needs it public (the metadata-check test uses `collectPrecondAsserts` + metadata inspection instead).

But the code in `bbf0d27f7` still has `def classifyPrecondition` (public in the module sense) and still carries the "Exposed (not `private`)" paragraph. I grepped the tree: the only callers of `classifyPrecondition` are line 113 in this same file — the tests use `collectPrecondAsserts` + `md.getPropertyType` as the reply says — so `private` is genuinely safe. This is presumably an unpushed intent.

Suggested replacement for lines 64–70:

```suggestion
/-- Classify a function precondition into a property type for SARIF reporting.
    For functions with multiple preconditions (e.g., SafeSDiv has both div-by-zero
    and overflow), the precondition index distinguishes them. -/
private def classifyPrecondition (funcName : String) (precondIdx : Nat := 0) : Option String :=
```

Strata/Languages/Core/Factory.lean
```diff
@@ -499,10 +499,44 @@ def seqAppendFunc : WFLFunc CoreLParams :=
...
+
+private def SeqBoundKind.opExpr : SeqBoundKind → LExpr CoreLParams.mono
+  | .Lt => (intLtFunc (T := CoreLParams)).opExpr
+  | .Le => (intLeFunc (T := CoreLParams)).opExpr
+
+/-- Precondition `0 <= varName && varName `k.opExpr` Sequence.length(s)`. -/
+private def mkSeqBoundsPrecond
+    (varName : String) (k : SeqBoundKind) :
+    Strata.DL.Util.FuncPrecondition (LExpr CoreLParams.mono) CoreLParams.Metadata :=
+  let sVar  : LExpr CoreLParams.mono := .fvar default "s" (some (seqTy mty[%a]))
```
tautschnig: Implicit coupling: `mkSeqBoundsPrecond` hard-codes `"s"` here, which is only correct because the four callers (`seqSelectFunc`, `seqUpdateFunc`, `seqTakeFunc`, `seqDropFunc`) all happen to name their sequence input `"s"`. The `h_precond` proof obligation on `polyUneval` catches mismatches at definition time — so this isn't a correctness trap, but the constraint isn't documented at the point where it bites.

Either take a second `seqVarName` parameter (symmetric with the existing `varName` for the index/count):

```suggestion
/-- Precondition `0 <= varName && varName `k.opExpr` Sequence.length(seqVarName)`. -/
private def mkSeqBoundsPrecond
    (seqVarName : String) (varName : String) (k : SeqBoundKind) :
    Strata.DL.Util.FuncPrecondition (LExpr CoreLParams.mono) CoreLParams.Metadata :=
  let sVar  : LExpr CoreLParams.mono := .fvar default seqVarName (some (seqTy mty[%a]))
  let xVar  : LExpr CoreLParams.mono := .fvar default varName (some mty[int])
```

and update the four call sites to `mkSeqBoundsPrecond "s" "i" .Lt` etc. — slightly more verbose at the call sites, but the helper's contract is now self-describing.

Or, if you prefer to keep the helper's signature narrow, add a one-line doc note: `Assumes the enclosing function names its sequence input "s"`.

StrataTest/Transform/PrecondElim.lean
```diff
@@ -418,4 +418,51 @@ procedure test (inout g : int, y : int)
...
+/-- Check that `collectPrecondAsserts` produces exactly `expectedCount`
+    obligations from `expr`, each tagged with `outOfBoundsAccess`. -/
+private def assertOutOfBoundsObligations
+    (expr : Core.Expression.Expr) (expectedCount : Nat) : IO Unit := do
+  let stmts := collectPrecondAsserts Core.Factory expr "test" #[]
+  assert! stmts.length == expectedCount
+  for s in stmts do
+    let md : MetaData Core.Expression := match s with
+      | Statement.assert _ _ md => md | _ => #[]
+    assert! md.getPropertyType == some MetaData.outOfBoundsAccess
```
tautschnig, line 447: `assertOutOfBoundsObligations` checks the obligation count and the metadata tag, but not the obligation content. That means an off-by-one regression in `mkSeqBoundsPrecond` — say, someone flips `intLtFunc` to `intLeFunc` for the `.Lt` case, producing `... && i <= length(s)` for `Sequence.select` — slides past Test 10 entirely (count and tag both stay right).

The proof-coverage suggestion in the main review body (`mkSeqBoundsPrecond_form`) is the strong version. A lighter-weight test-level check that would catch this class of regression cheaply:

```lean
/-- Check that `collectPrecondAsserts` produces exactly `expectedCount`
    obligations from `expr`, each tagged with `outOfBoundsAccess` and
    each having the expected `boolAnd`-of-two-int-comparisons shape
    (so off-by-one regressions in `mkSeqBoundsPrecond` are caught). -/
private def assertOutOfBoundsObligations
    (expr : Core.Expression.Expr) (expectedCount : Nat) : IO Unit := do
  let stmts := collectPrecondAsserts Core.Factory expr "test" #[]
  assert! stmts.length == expectedCount
  for s in stmts do
    match s with
    | Statement.assert _ e md =>
      assert! md.getPropertyType == some MetaData.outOfBoundsAccess
      -- The obligation is `boolAnd (intLe 0 _) (intLt|intLe _ (seqLength _))`.
      match e with
      | LExpr.app _ (LExpr.app _ andOp _) _ =>
        assert! andOp.isBoolAnd  -- pseudo; real predicate will depend on LExpr shape
      | _ => assert! false
    | _ => assert! false
```

(Exact predicate `isBoolAnd` would need spelling out; the intent is to fail loudly if the top-level constructor stops being `boolAnd`.) Happy to prepare a concrete patch if you'd like.

Strata/DL/Imperative/EvalContext.lean
```diff
@@ -116,6 +117,21 @@ instance : ToFormat PropertyType where
...
+    propertyType classifications should add a matching arm here. -/
+def convertMetaDataPropertyType {P : PureExpr} [BEq P.Ident]
+    (md : MetaData P) : PropertyType :=
+  match md.getPropertyType with
+  | some s =>
+    if s == MetaData.divisionByZero then .divisionByZero
+    else if s == MetaData.arithmeticOverflow then .arithmeticOverflow
+    else if s == MetaData.outOfBoundsAccess then .outOfBoundsAccess
+    else .assert
+  | none => .assert
```
tautschnig, line 134: We now have three parallel representations of the same enum:

1. `MetaData.divisionByZero` / `arithmeticOverflow` / `outOfBoundsAccess` — `String` constants (`MetaData.lean:293–299`).
2. `propertyTypeToClassification : PropertyType → String` (`SarifOutput.lean:92`) — the *display* side, kebab-case.
3. `convertMetaDataPropertyType : MetaData → PropertyType` — the decode side, using the constants from (1).

Adding a new `PropertyType` variant requires changes in all three places plus a new kebab/camel string pair, and forgetting (1)+(3) silently degrades to `.assert` at runtime — there's no compile-time check that (1) and (3) stay in sync.

One way to consolidate: a single definition that tabulates the serialization plus inverse:

```lean
/-- Canonical mapping from `PropertyType` to the pair of (metadata tag, SARIF classification) strings. -/
def PropertyType.serialization : PropertyType → Option (String × String)
  | .divisionByZero     => some ("divisionByZero",     "division-by-zero")
  | .arithmeticOverflow => some ("arithmeticOverflow", "arithmetic-overflow")
  | .outOfBoundsAccess  => some ("outOfBoundsAccess",  "out-of-bounds-access")
  | .cover              => some ("cover",              "cover")
  | .assert             => none
```

Then `propertyTypeToClassification` and `convertMetaDataPropertyType` both derive from this single source, and the `MetaData.*` constants can go. A `#[simp]` round-trip lemma (`fromClassification? (toClassification p) = some p`) gives compile-time coverage of the set. See the proof-coverage suggestion in the main review body.

tautschnig (commented):
**Proof-coverage suggestions ("what to prove next").** These are all follow-ups, not blocking.

- `convertMetaDataPropertyType_inverse_of_classifyPrecondition` — for every `(fn, idx)` that `classifyPrecondition` maps to `some metaString`, `convertMetaDataPropertyType` applied to a `MetaData` tagged with that string returns a non-`.assert` `PropertyType`. Concretely: there shouldn't be a gap where `classifyPrecondition` tags a call but `convertMetaDataPropertyType` silently reverts to `.assert`. Trivial to state, would catch the string-mismatch class of regressions the triplication enables.

- `PropertyType`-to-SARIF round-trip — `(propertyTypeToClassification p).fromClassification? = some p` for all `p : PropertyType`, where `fromClassification?` is a matching deserializer (not yet written). Would make `PropertyType` the single source of truth and mechanically prune the `MetaData.*` string constants.

- `mkSeqBoundsPrecond_form` — a structural theorem about `(mkSeqBoundsPrecond x k).expr`: it equals `boolAnd (intLe 0 x) (k.opExpr x (seqLength s))` (or the concrete LExpr shape). Makes `Test 10`'s off-by-one blind spot (see inline comment #2) explicit and proved rather than merely tested. Small.

- `seqSelect_denote_partial` — semantic soundness of the bounds precondition with respect to Lean's `List.get?`: `0 ≤ i < s.length → (seqSelect s i).denote = .some (s.get! i)`. This is `FactoryCorrect.lean` territory (cf. PR #1103) rather than this PR, but worth tracking as the obvious continuation: the preconditions added here are only *useful* insofar as they correspond to a semantic partiality.

- `collectPrecondAsserts`-spec — a purely structural theorem that for every sub-expression that is a call to a function with `N` preconditions, `collectPrecondAsserts` produces exactly `N` `assert` statements, each tagged by `classifyPrecondition`. Test 10 verifies counts 1, 1, 1, 1, and 2 empirically; a theorem would subsume all of that and catch e.g. a future refactor that stops visiting nested calls.

**Refactoring opportunities (not blocking).**

- As noted in finding #3, consolidate `MetaData.{divisionByZero,...}` / `propertyTypeToClassification` / `convertMetaDataPropertyType` into a single table.
- The new `assertOutOfBoundsObligations` helper in the tests is a good pattern; the pre-existing overflow/division tests in the same file (`Test 1`, `Test 3`, `Test 5`) could share a sibling helper, reducing per-test boilerplate uniformly.

StrataTest/Languages/Core/Tests/GenericCallFallbackTest.lean
```diff
@@ -114,6 +114,21 @@ info: "assert [q1]: fourArgFn(a, b, c, d);\n\n-- Errors encountered during conve
   let e := mkCall4 "fourArgFn" (mkFvar "a") (mkFvar "b") (mkFvar "c") (mkFvar "d")
   IO.println (repr (fmtStmt (mkAssert "q1" e) #["a", "b", "c", "d"]))
 
+-- -----------------------------------------------------------------------
+-- Test: unknown 0-ary operation renders as generic call (e.g.
+-- `Sequence.empty`, which is registered in the factory but not yet
+-- parseable in raw Core source). Previously, the `handleZeroaryOps`
```
MikaelMayer: 🤖 Nit: "Previously, the `handleZeroaryOps` fallback silently substituted `re.none()`." — this is a non-stateless comment referencing the old behavior. Consider rephrasing to describe what the test verifies without referencing the prior state, e.g.:

```
-- Test: unknown 0-ary operation renders as a generic call (e.g.
-- `Sequence.empty`, which is registered in the factory but not yet
-- parseable in raw Core source).
```

MikaelMayer (commented):
🤖 Clean PR. The `convertMetaDataPropertyType` extraction is a good de-duplication, `SeqBoundKind` is a nice way to prevent nested-precondition accidents, and the SARIF property-type wiring fix is a genuine improvement (was dead code before). Test coverage is solid.

No new issues beyond what's already been flagged in existing threads.

Strata/Transform/PrecondElim.lean
fabiomadge: Thanks for catching this — you are right, the intent never actually landed. The test commit was silently reverting `private` back to `def` on every rebase (the change was stuck in that commit from an earlier iteration). Dropped the reversion from the test commit and re-applied the private.

Strata/Languages/Core/Factory.lean
fabiomadge: Added a docstring note — the `"s"` hardcoding is a convention that mirrors `yNeZeroPrecond`'s `"y"` in `IntBoolFactory.lean`, and `polyUneval`'s `h_precond` free-vars check catches mismatches at elaboration time. Kept the signature narrow rather than adding a parameter, matching the existing pattern.

StrataTest/Transform/PrecondElim.lean
fabiomadge: You are right about the gap — I confirmed locally that flipping `.Lt` to `.Le` slides past the current helper. Deferring the fix to follow-up per your "not blocking" framing. `mkSeqBoundsPrecond_form` (from the main review body) is the clean structural answer; a test-level pretty-print string comparison would be a cheaper partial substitute. Happy to take either in a follow-up PR.

Strata/DL/Imperative/EvalContext.lean
fabiomadge: Agreed on the architectural observation. Deferring per your "not blocking" framing — a `PropertyType.serialization` table with the round-trip lemma you sketched is the right shape, but it touches pre-existing constants (`MetaData.divisionByZero` etc.) and the SARIF classifier, so a standalone cleanup PR is cleaner than folding it in here.

StrataTest/Languages/Core/Tests/GenericCallFallbackTest.lean
fabiomadge: Done — dropped the "Previously, ..." sentence.

MikaelMayer (commented):

fabiomadge:
@tautschnig thanks for the careful review. Inline replies cover the four file-specific points. On the proof-coverage suggestions and refactoring in your review body — all acknowledged as legitimate follow-up work. The three most directly tied to this PR are:

- `convertMetaDataPropertyType_inverse_of_classifyPrecondition` — catches the string-mismatch class the triplication enables.
- `mkSeqBoundsPrecond_form` — closes the off-by-one test gap (your inline on `StrataTest/Transform/PrecondElim.lean:447`).
- Consolidating `MetaData.{divisionByZero,...}` / `propertyTypeToClassification` / `convertMetaDataPropertyType` into a single table with the round-trip lemma (your inline on `Strata/DL/Imperative/EvalContext.lean:134`).

Happy to file them as issues if you want them tracked separately; otherwise the review thread itself is a reasonable memorial.

MikaelMayer (commented):

Strata/Languages/Core/Factory.lean
```diff
@@ -499,10 +499,48 @@ def seqAppendFunc : WFLFunc CoreLParams :=
...
+  let xVar  : LExpr CoreLParams.mono := .fvar default varName (some mty[int])
+  let zero  : LExpr CoreLParams.mono := .intConst default 0
+  let lenS  : LExpr CoreLParams.mono := .app default seqLengthFunc.opExpr sVar
+  let lower : LExpr CoreLParams.mono :=
+    .app default (.app default (intLeFunc (T := CoreLParams)).opExpr zero) xVar
+  let upper : LExpr CoreLParams.mono :=
+    .app default (.app default k.opExpr xVar) lenS
+  ⟨.app default (.app default (boolAndFunc (T := CoreLParams)).opExpr lower) upper,
+   default⟩
```
tautschnig, line 545: **Proof coverage**: a `@[simp]` free-vars theorem here, parallel to `yNeZeroPrecond_freeVars` at `Strata/DL/Lambda/IntBoolFactory.lean:432`, would both document the invariant and trivially discharge `polyUneval`'s `h_precond` at every call site:

```lean
@[simp]
private theorem mkSeqBoundsPrecond_freeVars (varName : String) (k : SeqBoundKind) :
    ((mkSeqBoundsPrecond varName k).expr.freeVars).map (·.1.name) = ["s", varName] := by
  cases k <;>
  simp [mkSeqBoundsPrecond, LExpr.freeVars, SeqBoundKind.opExpr,
        LFunc.opExpr, intLeFunc, intLtFunc, boolAndFunc, seqLengthFunc]
```

(Exact simp lemmas will depend on how each helper is defined; the pattern in `IntBoolFactory.lean:432` gives the template. If the `.map (·.1.name) = ["s", varName]` equality is too tight — e.g. ordering-dependent — a `⊆` variant is the weaker-but-still-useful form.)

With that in place, the `"s"` hardcoding noted in your docstring becomes a *theorem*, not a convention maintained by vigilance. And the `h_precond` in each of the four `seqSelect/Update/Take/Drop` call sites discharges via `simp` rather than relying on `grind`'s opacity-sensitive behaviour.

Second, stronger theorem that I think is more valuable — see the inline on the test file — is a structural theorem `mkSeqBoundsPrecond_form` pinning the top-level AST to `boolAnd (intLe 0 _) (k.opExpr _ _)`. That one catches the off-by-one regression I flagged last round at elaboration time. Sketch:

```lean
theorem mkSeqBoundsPrecond_form (varName : String) (k : SeqBoundKind) :
    ∃ lo up,
      (mkSeqBoundsPrecond varName k).expr =
        .app default (.app default (boolAndFunc (T := CoreLParams)).opExpr lo) up ∧
      -- lo is `0 ≤ x` with `intLeFunc` (independent of `k`)
      lo = .app default (.app default (intLeFunc (T := CoreLParams)).opExpr (.intConst default 0))
             (.fvar default varName (some mty[int])) ∧
      -- up is `x ≤/< length(s)` with `k.opExpr`
      up = .app default (.app default k.opExpr
               (.fvar default varName (some mty[int])))
             (.app default seqLengthFunc.opExpr (.fvar default "s" (some (seqTy mty[%a])))) := by
  cases k <;> exact ⟨_, _, rfl, rfl, rfl⟩
```

If `rfl` holds as written (I'd expect it to, the definition is entirely concrete), this theorem is essentially free and closes the Test 10 content-check gap structurally.

Strata/DL/Imperative/EvalContext.lean
tautschnig, line 134: Per your "deferring to a follow-up PR" reply on the triple-representation thread — agreed on scope. One cheap intermediate measure worth adding *here*, before the full `PropertyType.serialization` consolidation lands, is a `@[simp]` roundtrip theorem. It doesn't change the code shape, but it gives you a compile-time error if someone adds a `MetaData.*` constant without a matching `convertMetaDataPropertyType` arm (or vice versa):

```lean
/-- Encoding `p` as a MetaData propertyType string and decoding it back
    yields `p` (for all non-`.assert` variants; `.assert` is the default). -/
@[simp]
theorem convertMetaDataPropertyType_roundTrip
    {P : PureExpr} [BEq P.Ident] (p : PropertyType) :
    p ≠ .assert →
    let md : MetaData P := MetaData.fromPair MetaData.propertyType
      (match p with
       | .divisionByZero     => MetaData.divisionByZero
       | .arithmeticOverflow => MetaData.arithmeticOverflow
       | .outOfBoundsAccess  => MetaData.outOfBoundsAccess
       | .cover              => "cover"    -- if .cover is ever tagged via MetaData
       | .assert             => "")
    convertMetaDataPropertyType md = p := by
  intro hne
  cases p <;> simp_all [convertMetaDataPropertyType, MetaData.getPropertyType, ...]
  -- the `.cover` case needs a MetaData string constant if you want it covered here,
  -- or split the theorem to non-`.cover`-and-non-`.assert` PropertyTypes.
```

(Exact simp set depends on how `MetaData.fromPair` / `getPropertyType` reduce; adjust from there. If `.cover` isn't meant to flow through MetaData, restrict the theorem's scope.)

When you do the full consolidation follow-up, this theorem becomes the inverse-of-forward lemma and is still useful. If a new `PropertyType` variant gets added without the corresponding `MetaData.*` + `convertMetaDataPropertyType` arm, `cases p` now produces an unhandled case and the proof stops compiling. That's the compile-time coverage the triple-representation is currently missing.

StrataTest/Transform/PrecondElim.lean
tautschnig, line 447: Per your "deferring to follow-up" reply — fine on the blocking call. Restating the concrete substitute so it's in one place:

The current helper catches *count* and *tag* regressions but not *content* regressions. An off-by-one — `.Lt ↔ .Le` in `mkSeqBoundsPrecond`, or worse, swapping `intLeFunc` for `intLtFunc` on the lower bound — slides through Test 10 entirely. I confirmed locally that flipping `mkSeqBoundsPrecond`'s upper-bound op on Select would not fail any test in this PR.

Two substitute shapes, in increasing order of strength:

**Cheap**: snapshot the pretty-printed form of each obligation (same shape as `Examples/Seq.lean` already does for the end-to-end case, but tighter-scoped to this helper):

```lean
/-- info: "0 <= i && i < Sequence.length(s)" -/
#guard_msgs in
#eval do
  let stmts := collectPrecondAsserts Core.Factory
    (LExpr.mkApp () Core.seqSelectOp [fxS, fxI]) "test" #[]
  match stmts.head? with
  | some (Statement.assert _ e _) => IO.println (repr (Std.format e))
  | _ => IO.println "<unexpected>"
```

Four of these (one per op, with `.Lt` vs `.Le` spelled in the snapshot) would catch any bound-operator regression. Pretty-print strings are a blunt instrument but cheap.

**Structural**: the `mkSeqBoundsPrecond_form` theorem I sketched in the inline comment on `Factory.lean`. If that `rfl`s, it obviates this test-level check — the regression becomes a *compile error* in the factory file, not a test failure. That's the better target if you can get it.

Happy to prep either patch if useful.

Strata/Languages/Core/Factory.lean
```diff
@@ -499,10 +499,48 @@ def seqAppendFunc : WFLFunc CoreLParams :=
...
+    `Sequence.select`/`update`, `Le` (non-strict) for `Sequence.take`/`drop`.
+    Restricting the parameter to this inductive rather than taking an
+    arbitrary `WFLFunc` or `LExpr` makes it impossible to attach a partial
+    operator (which would create a nested precondition obligation) by
+    accident. -/
+private inductive SeqBoundKind where | Lt | Le
+
+private def SeqBoundKind.opExpr : SeqBoundKind → LExpr CoreLParams.mono
+  | .Lt => (intLtFunc (T := CoreLParams)).opExpr
+  | .Le => (intLeFunc (T := CoreLParams)).opExpr
```
tautschnig, line 522: Minor: `SeqBoundKind.opExpr` hard-codes `intLtFunc` / `intLeFunc` for the current two kinds. Fine for today. If a third partial-Sequence op ever needs a different comparison (e.g. a `Nat`-typed variant, or bitvector sequences), the inductive grows and so does this match. Worth a one-line comment tying the enum to "comparison operators over `int`" so a future `bvLtFunc`-flavoured kind is clearly a separate thing rather than another arm here.

Also a naming nit: `SeqBoundKind.opExpr` returns the upper-bound comparison; `mkSeqBoundsPrecond` uses `k.opExpr` only for the upper bound (the lower is hardcoded to `intLeFunc`). A name like `SeqBoundKind.upperOpExpr` reads more self-documenting:

```suggestion
private def SeqBoundKind.upperOpExpr : SeqBoundKind → LExpr CoreLParams.mono
  | .Lt => (intLtFunc (T := CoreLParams)).opExpr
  | .Le => (intLeFunc (T := CoreLParams)).opExpr
```

tautschnig (dismissed):

Strata/Languages/Core/Factory.lean
fabiomadge: Done — renamed to `upperOpExpr` and added a scoping comment noting this is `int`-comparison-only and that a non-`int` variant (e.g. bitvector) should introduce a separate helper rather than extend this enum.

Strata/Languages/Core/Factory.lean
fabiomadge: Landed `mkSeqBoundsPrecond_form` per your sketch — `cases k <;> rfl` closes it. One subtlety: the `up` clause needs to pin the upper-bound op per-kind rather than abstract it as `k.upperOpExpr`, otherwise any change to the mapping inside `upperOpExpr` satisfies `rfl` trivially (both sides reduce through the same function). Verified locally that swapping `intLtFunc` → `intLeFunc` in `upperOpExpr` now breaks this proof.

Deferring `mkSeqBoundsPrecond_freeVars_subset` to a follow-up: I landed it locally and confirmed the `⊆ [varName, "s"]` form works (with `rw [mkSeqBoundsPrecond_form]` then `cases k <;> simp [LExpr.freeVars, WFLFunc.opExpr, LFunc.opExpr, List.subset_def]`). But I also probed stripping the `@[simp]` tag — the existing `polyUneval` default `first | decide | grind` was already closing `h_precond` at all four call sites, so the tag provides no observable benefit in this PR. Happy to land it separately if you want the theorem purely for documentation value.

Independently, `mkSeqBoundsPrecond_form` does _not_ catch the call-site regression (`.Lt` → `.Le` at a `preconditions := [mkSeqBoundsPrecond "i" _]`): the theorem is universally quantified over `k`, so a specific `k` at the call site is not pinned. That class is covered by the pretty-print snapshots on the test-file thread (see reply there).

StrataTest/Transform/PrecondElim.lean
fabiomadge: Landed the cheap pretty-print snapshot version — four `#guard_msgs` blocks (one per partial op: `Select`, `Update`, `Take`, `Drop`), each asserting the exact printed body of the first obligation. Verified locally that flipping `.Lt` → `.Le` at a call site (e.g. `seqSelectFunc`'s `preconditions`) now trips these tests with a clear diff.

Note that this complements, rather than is subsumed by, the structural `mkSeqBoundsPrecond_form` theorem I also landed: the theorem catches helper-implementation drift (changes to `mkSeqBoundsPrecond`'s body or `SeqBoundKind.upperOpExpr`'s mapping), while these snapshots catch call-site drift (a caller passing the wrong `SeqBoundKind`). The theorem is universally quantified over `k`, so it cannot pin the specific `k` at a call site.

Strata/DL/Imperative/EvalContext.lean
fabiomadge: Deferring. I prototyped the three variant round-trip theorems locally and they go through, but proving them requires adding `@[expose]` to about seven definitions in `Strata/DL/Imperative/MetaData.lean` (`pushElem`, `findElem`, `propertyType`, `getPropertyType`, `Field.beq`, and the three property-type string constants) plus `convertMetaDataPropertyType` itself in `EvalContext.lean`, because their bodies are otherwise opaque to `simp`/`decide` outside their defining file.

Exposing pre-existing `MetaData` internals is a public-API change that predates this PR's scope, and (as you note) the cleanest version is bundled with the full `PropertyType.serialization` consolidation. I'd rather do it once, cleanly, in the follow-up PR than slip the exposure in here.

Secondary note: the construction in your sketch uses `MetaData.fromPair`, which doesn't exist in the current codebase — the canonical form is `MetaData.empty.pushElem MetaData.propertyType (.msg MetaData.divisionByZero)`. Mentioning in case it's useful when you come back to this.

MikaelMayer (commented):

Strata/Languages/Core/Factory.lean
```diff
@@ -499,10 +499,76 @@ def seqAppendFunc : WFLFunc CoreLParams :=
...
+  ⟨.app default (.app default (boolAndFunc (T := CoreLParams)).opExpr lower) upper,
+   default⟩
+
+/-- Structural characterisation of `mkSeqBoundsPrecond`. Pins the obligation's
+    top-level AST so that changes to `mkSeqBoundsPrecond` or `SeqBoundKind.upperOpExpr`
+    surface as proof failures here rather than silent behavioural drift at the
+    call sites. The upper-bound comparison is pinned per-kind (strict `<` for
+    `.Lt`, non-strict `≤` for `.Le`) rather than abstracted as `k.upperOpExpr`,
+    so that swapping the mapping inside `upperOpExpr` breaks this proof. -/
+private theorem mkSeqBoundsPrecond_form (varName : String) (k : SeqBoundKind) :
```
joscoh: Nit: I personally don't think this theorem is all that necessary or helpful. IIUC, it reduces to proving that 
```
SeqBoundKind.upperOpExpr k = match k with
   | .Lt => (intLtFunc (T := CoreLParams)).opExpr
   | .Le => (intLeFunc (T := CoreLParams)).opExpr
```
Which is true, but it seems quite silly to enforce that functions in the code don't change by stating lemmas for each that prove they are definitionally equal to their unfolding. At this point, we are just writing the code twice.

MikaelMayer (dismissed):

Strata/Languages/Core/Factory.lean
fabiomadge: Agreed. You're right that it reduces to writing the code twice, and I verified locally that everything this theorem catches is already caught by the per-op pretty-print snapshots in `StrataTest/Transform/PrecondElim.lean` (Test 10a) — both the `upperOpExpr`-mapping regression and the lower/upper-swap regression inside `mkSeqBoundsPrecond`. The theorem wasn't adding a unique safety net, just earlier detection at the feat-commit boundary.

Dropped the theorem. Kept the rename to `upperOpExpr` and its scoping comment, since those are legitimate standalone improvements independent of the theorem question.

Strata/Languages/Core/Factory.lean
fabiomadge: Follow-up: after pushback from @joscoh [here](https://github.com/strata-org/Strata/pull/1100#discussion_r3210757329), I've dropped `mkSeqBoundsPrecond_form`. His argument is that the theorem reduces to restating the definition — writing code twice — and enforcing that a function is definitionally equal to its own unfolding.

I probed whether the theorem catches anything unique vs the per-op pretty-print snapshots (Test 10a) and confirmed it doesn't:

- `upperOpExpr .Lt => .Le` drift → all 4 snapshots fire.
- Swapping lower/upper inside `mkSeqBoundsPrecond`'s body → all 4 snapshots fire.
- Callsite `.Lt → .Le` → snapshot for that callsite fires (which the theorem never caught anyway, since it's quantified over `k`).

The theorem was only buying earlier detection at the feat-commit boundary, not unique coverage. Dropped to avoid the redundancy.

Keeping the `upperOpExpr` rename + scoping docstring (those are independent wins) and the snapshot tests (those do the coverage work).

MikaelMayer (commented):

MikaelMayer (dismissed):

MikaelMayer (commented):

atomb (approved):

MikaelMayer (approved):

MikaelMayer (commented):

PR #1103, opened 2026-05-04
Add correctness of the term creators in SMT/Factory.lean

aqjune-aws:
(Creating a new pull request because for some reason pushing to a branch didn't update https://github.com/strata-org/Strata/pull/1088)

This proves that the term creators in SMT (which are also called 'Factory') are sound with respect to the denotational semantics. Initially the patch was defining a cleaner inductive predicate `DenotePred` that is analogous to `denoteTerm` but avoids carrying data structures that are necessary to construct the Lean goals.
However, this turned out to introduce a big amount of boilerplate code because soundness/completeness of `DenotePred` w.r.t. `denoteTerm` turned out to be very big proofs.
Instead, this pull request directly uses the `denoteTerm` and its family functions to state & prove correctnesses.

There is one operation in Factory.lean which could not be proven: `quant`. The `quant` operation does the following simplification: `forall x, forall y, e` ==> `forall x y, e`. Proving this was hard because the result of `denoteTerm` is `TermDenoteResult ctx` which is a dependent type (`ctx : Context`)... And rewriting the dependent type position was challenging.


Strata/DL/SMT/FactoryCorrect.lean
```diff
@@ -0,0 +1,1212 @@
...
+/-!
+# Correctness of Factory optimizations
+
+This module proves that the simplifications performed by `Factory` functions
+preserve the denotational semantics directly in terms of the functional
+denotation (`denoteBoolTermAux`, `denoteIntTermAux`).
+
+For boolean operations, results use `∃ p', ... ∧ (p' ↔ expected)` because
+`denoteTerm` maps booleans to `Prop` and propositional double-negation
+elimination is not definitional.
```
joscoh: Confused what double-negation elimination has to do with this. I also don't think that this is accurate. Since Lean assumes propositional extensionality by default, `iff` coincides with definitional equality on `Prop`.

Strata/DL/SMT/FactoryCorrect.lean
```diff
@@ -0,0 +1,1212 @@
...
+
+For integer operations, results use direct equality.
+-/
+
+open Strata.SMT
+
+/-! ## Infrastructure -/
+
+/-- The unique `TermDenoteInput` for the empty context. -/
+private noncomputable abbrev tdi₀ : TermDenoteInput ({} : Context) :=
```
joscoh: What is `noncomputable` about this?

Strata/DL/SMT/FactoryCorrect.lean
```diff
@@ -0,0 +1,1212 @@
...
+  split at h
+  · rename_i _ _ fi _
+    exact ⟨fi, by grind, by grind⟩
+  · grind
+
+/-! ## Lemma: denoteBoolTermAux for .app .not -/
+
+/-- If `denoteBoolTermAux (.app .not [t'] ty) = some p`, then there exists `q`
+    such that `denoteBoolTermAux t' = some q` and `p ↔ ¬q`. -/
+private theorem denoteBoolTermAux_not_inv {t' : Term} {ty : TermType} {p : Prop}
```
joscoh, line 86: Larger comment/question: is there a particular reason to have everything be existentially quantified with `iff`? Because Lean assumes propositional extensionality by default, the following works:
```
private theorem denoteBoolTermAux_not_inv {t' : Term} {ty : TermType} {p : Prop}
    (h : denoteBoolTermAux (.app .not [t'] ty) = some p) :
    denoteBoolTermAux t' = some (¬ p):= by
  unfold denoteBoolTermAux at h ⊢
  conv at h => simp only [denoteTerm]
  revert h
  generalize denoteTerm {} t' = res'
  intro h
  match res' with
  | some ⟨.prim .bool, rfl, g⟩ => simp_all; grind
  | some ⟨.prim .int, _, _⟩ | some ⟨.prim .string, _, _⟩
  | some ⟨.prim (.bitvec _), _, _⟩ | some ⟨.prim .real, _, _⟩
  | some ⟨.prim .regex, _, _⟩ | some ⟨.prim .trigger, _, _⟩
  | some ⟨.option _, _, _⟩ | some ⟨.constr _ _, _, _⟩
  | none => grind
```
This is a design question, but it seems to me that non-quantified theorems like this will be much easier to work with (e.g. when rewriting).

joscoh (commented):
Haven't gone through it all yet but I'd love to know your thoughts on whether the theorem conclusions should be existentially quantified or direct.

Strata/DL/SMT/FactoryCorrect.lean
aqjune-aws: I thought what AI wrote here was something like this - double negation elimination was necessary while proving some bool-related Factory operations(I guess it was `Factory.not` because it optimizes double negation), but double negation elim is equivalent to excluded middle which is in the Classical module, and the agent wanted to avoid its usage.
But I found that Denote.lean was already using `Classical.propDecidable`, and IIUC all of the related theorems in Classical (which includes EM) share the common axiom: the axiom of choice. So I suggested just using Classical to AI and simplify theorem statements & proofs, and instead left a comment at Denote.lean as well as FactoryCorrect.lean.

Strata/DL/SMT/FactoryCorrect.lean
aqjune-aws: It was redundant, removed

Strata/DL/SMT/FactoryCorrect.lean
aqjune-aws: Initially I just left what the agent picked for the correctness statements (the `*_correct` are the top-level correctness statements of Factory ops). The `*_correct` theorems are switched to plain equality statements.

Strata/DL/SMT/FactoryCorrect.lean
```diff
@@ -0,0 +1,1159 @@
...
+  match res' with
+  | some ⟨.prim .bool, rfl, g⟩ =>
+    refine ⟨g tdi₀, by simp, ?_⟩
+    simp at h
+    rw [h]
+  | some ⟨.prim .int, _, _⟩ | some ⟨.prim .string, _, _⟩
+  | some ⟨.prim (.bitvec _), _, _⟩ | some ⟨.prim .real, _, _⟩
+  | some ⟨.prim .regex, _, _⟩ | some ⟨.prim .trigger, _, _⟩
+  | some ⟨.option _, _, _⟩ | some ⟨.constr _ _, _, _⟩
+  | none => grind
```
tautschnig: This is the last piece of joscoh's thread (https://github.com/strata-org/Strata/pull/1103#discussion_r3183...). The top-level `*_correct` theorems are now plain equalities, but `denoteBoolTermAux_not_inv` still has the `∃ q, ... ∧ p = ¬q` shape, which is exactly the form joscoh's suggestion replaced with `denoteBoolTermAux t' = some (¬ p)`.

I verified locally that the simpler form compiles (with the proof joscoh posted, modulo `simp_all; grind` for the `some ⟨.prim .bool, rfl, g⟩` case) and that it lets the three callers shorten:

- `Factory.not_correct` (line 216): `· obtain ⟨q, hq, hpq⟩ := denoteBoolTermAux_not_inv h; rw [hpq, Classical.not_not]; exact hq` collapses to `· exact denoteBoolTermAux_not_inv h`.
- `Factory.and_correct` (lines 263/266): the `obtain ⟨q, hq, heq⟩` + `denoteBoolTermAux_eq` + `rw [heq, hpq]` sequence collapses to a direct `Option.some.inj` on `h₁.symm.trans (denoteBoolTermAux_not_inv h₂)` followed by `subst`.
- `Factory.or_correct` (lines 303/310): analogous.

Suggested rewrite (one-click apply replaces lines 85–104):

```suggestion
/-- Invert `denoteBoolTermAux` of a negation: the inner term denotes `¬p`. -/
private theorem denoteBoolTermAux_not_inv {t' : Term} {ty : TermType} {p : Prop}
    (h : denoteBoolTermAux (.app .not [t'] ty) = some p) :
    denoteBoolTermAux t' = some (¬ p) := by
  unfold denoteBoolTermAux at h ⊢
  conv at h => simp only [denoteTerm]
  revert h
  generalize denoteTerm {} t' = res'
  intro h
  match res' with
  | some ⟨.prim .bool, rfl, g⟩ => simp_all; grind
  | some ⟨.prim .int, _, _⟩ | some ⟨.prim .string, _, _⟩
  | some ⟨.prim (.bitvec _), _, _⟩ | some ⟨.prim .real, _, _⟩
  | some ⟨.prim .regex, _, _⟩ | some ⟨.prim .trigger, _, _⟩
  | some ⟨.option _, _, _⟩ | some ⟨.constr _ _, _, _⟩
  | none => grind
```

The three callers will then need follow-up edits along the lines above; happy to prepare a patch if that helps.

Strata/DL/SMT/Denote.lean
```diff
@@ -523,7 +529,7 @@ def bindExistsVar : QuantVarBinder := fun {n} {ty} ctx hTy =>
         { sΓ := tdi.sΓ, hsΓ := tdi.hsΓ, tΓ := { ufs := tdi.tΓ.ufs, vs := vΓ' }, htΓ := { hv := hv', huf := tdi.htΓ.huf } }
       ft' tdi'
 
-def buildQuant (bindVar : QuantVarBinder) (ctx : Context) (vs : List TermVar)
+@[expose] def buildQuant (bindVar : QuantVarBinder) (ctx : Context) (vs : List TermVar)
```
tautschnig: This (and the `@[expose]` on `buildExists` at line 547, `buildForall` at line 554, and `rightAssoc` at line 918) are unused by `FactoryCorrect`. I checked by selectively removing them and rebuilding both `Strata.DL.SMT.Denote` and `Strata.DL.SMT.FactoryCorrect` — both still compile clean.

They were presumably added preemptively during exploration (especially for a future `Factory.quant_correct`), but since `quant` is explicitly deferred and `rightAssoc` isn't touched by any current proof, they don't need to be part of `Denote`'s exposed surface yet. `@[expose]` on `denoteSort` and `denotePrimSort` *is* needed — the new `denoteBVTermAux` / `denoteStringTermAux` definitions unfold into them, and removing those `@[expose]` attributes fails the Denote build with "definitions were not unfolded because their definition is not exposed". Those stay.

Recommend dropping `@[expose]` on just these four; re-add whenever the proofs that need them land.

Strata/DL/SMT/Factory.lean
```diff
@@ -129,6 +140,7 @@ theorem mkSimpleTriggerIsSimple: isSimpleTrigger (mkSimpleTrigger x ty) := by
   simp [isSimpleTrigger, mkSimpleTrigger]
 
 -- Note: we could coalesce nested quantifiers here, since SMT-Lib allows multiple variables to be bound at once.
+-- TODO: Its correctness could not be proven due to its complexity. Contribution is welcome
```
tautschnig, line 143: A free-text `TODO` comment is weak. Consider replacing it with a stub theorem in `FactoryCorrect.lean`:

```lean
/-- `Factory.quant` preserves `denoteTerm` semantics. Proving this requires
    rewriting under dependent types in `TermDenoteResult ctx`, which is not
    yet done. -/
theorem Factory.quant_correct (ctx : Context) (qk : QuantifierKind) (x : String)
    (ty : TermType) (tr : Term) (e : Term) :
    denoteTerm ctx (Factory.quant qk x ty tr e) = denoteTerm ctx (.quant qk x ty tr e) :=
  sorry
```

Benefits over a free-text TODO:
- The intended theorem statement is pinned down in Lean syntax and elaborates (so it won't bit-rot silently against `denoteTerm`'s signature).
- `#check Factory.quant_correct` surfaces the obligation, and Lean's `sorry` warning makes it explicit that the module relies on an unproven fact.
- It gives the next contributor a concrete starting point.

The comment on this line could then be shortened to `-- Correctness: Factory.quant_correct (currently sorry)`.

As noted in the top-level review, a tractable intermediate milestone is proving just the *non-coalescing* case (`rfl` over every `e` except `.quant _ _ _ _ _`, plus a case split on the condition inside the `.quant _` branch); that already gives a safety net for the common path.

Strata/DL/SMT/FactoryCorrect.lean
```diff
@@ -0,0 +1,1159 @@
...
+             · simp_all)
+      · obtain ⟨f₁, hdt₁, hiff₁⟩ := denoteBoolTermAux_extract h₁
+        obtain ⟨f₂, hdt₂, hiff₂⟩ := denoteBoolTermAux_extract h₂
+        have h1 := propext hiff₁
+        have h2 := propext hiff₂
+        subst h1; subst h2
+        simp only [denoteBoolTermAux, denoteTerm, Option.pure_def, Option.bind_eq_bind,
+                   Option.bind_some, hdt₁, hdt₂, denoteTerms, chainable, chainable.go,
+                   dif_pos trivial]
+        exact congrArg some (propext ⟨fun h => h ▸ Iff.rfl, propext⟩)
```
tautschnig, line 842: Non-blocking, for a follow-up. All four `Factory.eq_correct_{bool,int,bv,string}` theorems contain a structurally identical `iterate 3 · exfalso ...` block that rules out `.some t'` and `.none ty` sub-terms by unfolding `denote*TermAux`. The same pattern shows up inside the `Factory.ite_correct_*` family (lines ~985, ~1020, ~1055, ~1090).

Two tiny lemmas — stated once per `denote*TermAux` — would eliminate all of these:

```lean
private theorem denoteBoolTermAux_some (t : Term) : denoteBoolTermAux (.some t) = none := by
  simp [denoteBoolTermAux, denoteTerm]
  rcases hd : denoteTerm {} t with _ | ⟨ty', _, _⟩ <;> rw [hd] <;> simp

private theorem denoteBoolTermAux_none (ty : TermType) : denoteBoolTermAux (.none ty) = none := by
  simp [denoteBoolTermAux, denoteTerm]
```

Then each `exfalso` block collapses to a one-liner `simp_all [denoteBoolTermAux_some, denoteBoolTermAux_none]`, the `denote*TermAux`-specific variants extend the same pattern, and adding a new eq/ite variant in the future doesn't duplicate boilerplate.

Similar argument for the `Factory.{bvadd,bvsub,bvmul,bvshl,bvlshr}_correct` family (lines ~520–570): the `unfold Factory.bvadd Factory.bvapp; split; · next m b₁ b₂ => simp only [...] at hdt₁ hdt₂; obtain ⟨hmn₁, hf₁⟩ := hdt₁; ...; subst hmn₁; subst hmn₂; subst hf₁; subst hf₂; simp [...]` pattern repeats verbatim. A single helper tactic macro `bv_binop_correct` or a proof-producing `extract ⟨hmn, hf⟩ from hdt` tactic would remove the duplication without changing the proofs' semantic content.

Strata/DL/SMT/FactoryCorrect.lean
aqjune-aws: I fixed `denoteBoolTermAux_not_inv` and its user sites accordingly.

Strata/DL/SMT/Denote.lean
aqjune-aws: Good catch, I removed them.

Strata/DL/SMT/Factory.lean
aqjune-aws: I would prefer not having `sorry`s in our proofs (we once went through refactoring to remove all the sorrys in Strata by either removing them or proving them), so... I think the only choice is to leave a comment. :)

Strata/DL/SMT/FactoryCorrect.lean
aqjune-aws: It seems the provided theorems are not enough (fails at `rw [hd]`) and a few significant fixes are necessary..? Due to the implementation detail of these definitions automation is slightly tricky, but it might be me who thinks so. Might be a good exercise to people who want to contribute to FactoryCorrect.lean in the future :)

tautschnig (approved):

joscoh (approved):
I like how nice the theorem statements turn out!

PR #1105, opened 2026-05-04
fix : Issue 1102

julesmt:
## Summary

Fixes #1102. Two independent issues surfaced by the hypothesmith fuzzer (seed `1777894483`):

1. **`isinstance(...)` inside `assert` produced a Laurel type error.** The assert translator hoists `.Hole` conditions into a fresh `bool` variable, then wrapped the result in `Any_to_bool`, which is ill-typed (`Any → bool` applied to `bool`). Now the coercion is skipped when hoisting has produced a `bool`.

2. **Known Laurel limitations were reported as internal errors.** `pyAnalyzeLaurel` classified every Laurel-to-Core translation failure as `Internal error`, including legitimate `UserError`-level diagnostics like "calls to procedures are not supported in functions or contracts" (triggered here by `str(e)` on a caught exception, where `e : PythonError` is a composite and `str` resolves to a generated procedure). Now only `StrataBug` diagnostics map to `Internal error`; `UserError` and `NotYetImplemented` map to `Known limitation`.

### DIFF
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -1718,17 +1718,18 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
       | some (.Constant _ (.ConString _ str) _) => some str.val
       | _ => none
     -- Check if condition contains a Hole - if so, hoist to variable
-    let (condStmts, finalCondExpr, condCtx) :=
+    let (condStmts, finalCondExpr, condCtx, isHoisted) :=
       match condExpr.val with
       | .Hole =>
         let freshVar := s!"assert_cond_{test.toAst.ann.start.byteIdx}"
         let varType := mkHighTypeMd .TBool
         let varDecl := mkStmtExprMd (StmtExpr.LocalVariable freshVar varType (some condExpr))
         let varRef := mkStmtExprMd (StmtExpr.Identifier freshVar)
-        ([varDecl], varRef, { ctx with variableTypes := ctx.variableTypes ++ [(freshVar, "bool")] })
-      | _ => ([], condExpr, ctx)
+        ([varDecl], varRef, { ctx with variableTypes := ctx.variableTypes ++ [(freshVar, "bool")] }, true)
+      | _ => ([], condExpr, ctx, false)
 
-    let assertStmt := mkStmtExprMdWithLoc (StmtExpr.Assert { condition := Any_to_bool finalCondExpr, summary }) md
+    let coercedCond := if isHoisted then finalCondExpr else Any_to_bool finalCondExpr
+    let assertStmt := mkStmtExprMdWithLoc (StmtExpr.Assert { condition := coercedCond, summary }) md
 
     -- Wrap in block if we hoisted condition
     let result := if condStmts.isEmpty then
--- a/Strata/Languages/Python/PythonToLaurel.lean
+++ b/Strata/Languages/Python/PythonToLaurel.lean
@@ -1717,18 +1717,25 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
     let summary := match msg.val with
       | some (.Constant _ (.ConString _ str) _) => some str.val
       | _ => none
-    -- Check if condition contains a Hole - if so, hoist to variable
-    let (condStmts, finalCondExpr, condCtx, isHoisted) :=
+    let (condStmts, finalCondExpr, condCtx) :=
       match condExpr.val with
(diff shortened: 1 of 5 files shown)


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1718,17 +1718,18 @@ partial def translateStmt (ctx : TranslationContext) (s : Python.stmt SourceRang
...
         let varDecl := mkStmtExprMd (StmtExpr.LocalVariable freshVar varType (some condExpr))
         let varRef := mkStmtExprMd (StmtExpr.Identifier freshVar)
-        ([varDecl], varRef, { ctx with variableTypes := ctx.variableTypes ++ [(freshVar, "bool")] })
-      | _ => ([], condExpr, ctx)
+        ([varDecl], varRef, { ctx with variableTypes := ctx.variableTypes ++ [(freshVar, "bool")] }, true)
+      | _ => ([], condExpr, ctx, false)
 
-    let assertStmt := mkStmtExprMdWithLoc (StmtExpr.Assert { condition := Any_to_bool finalCondExpr, summary }) md
+    let coercedCond := if isHoisted then finalCondExpr else Any_to_bool finalCondExpr
+    let assertStmt := mkStmtExprMdWithLoc (StmtExpr.Assert { condition := coercedCond, summary }) md
```
tautschnig: The fix is correct — when the condition is a `.Hole` (unmodeled call like `isinstance`), it's hoisted into a `bool` variable, so applying `Any_to_bool` (which has type `Any → bool`) to the `bool`-typed variable reference is ill-typed.

However, the `isHoisted` flag is a proxy for "the expression is already `bool`-typed". If in the future another code path produces a `bool`-typed condition without hoisting, the coercion would still be incorrectly applied. A more robust approach would be to check the expression's type directly, e.g.:
```lean
let coercedCond := if finalCondExpr.val matches .Identifier _ then
  -- If it's a variable reference to a bool-typed var, skip coercion
  match condCtx.variableTypes.find? (fun (n, _) => n == ...) with
  | some (_, "bool") => finalCondExpr
  | _ => Any_to_bool finalCondExpr
else Any_to_bool finalCondExpr
```

**Missing test:** Please add a regression test, e.g. in `StrataTestExtra/Languages/Python/` or as a new `.py` test file:
```python
def main() -> None:
    x: int = 5
    assert isinstance(x, int)
```
This should verify that the program doesn't produce a type error.

StrataMain.lean
```diff
@@ -656,7 +656,13 @@ def pyAnalyzeLaurelCommand : Command where
...
       match coreProgramOption with
       | none =>
-        exitPyAnalyzeInternalError s!"Laurel to Core translation failed: {laurelTranslateErrors}"
+        let msg := s!"Laurel to Core translation failed: {laurelTranslateErrors}"
+        let hasStrataBug := laurelTranslateErrors.any
+          (fun d => d.type == DiagnosticType.StrataBug)
+        if hasStrataBug then
+          exitPyAnalyzeInternalError msg
+        else
+          exitPyAnalyzeKnownLimitation msg
```
tautschnig, line 667: Good fix. The classification is correct:
- `StrataBug` → "Internal error" (indicates a bug in Strata itself)
- `UserError` / `NotYetImplemented` → "Known limitation" (the program uses features Strata doesn't support yet)

This is consistent with the existing pattern at line 638 where `knownLimitation` and `internal` errors from the Python-to-Laurel pipeline are already distinguished.

One edge case: if `laurelTranslateErrors` contains BOTH a `StrataBug` and a `UserError`, the current code classifies the whole batch as "Internal error" (because `hasStrataBug` is true). This seems correct — if there's a Strata bug in the mix, we want to surface it — but it's worth a brief comment:
```suggestion
        let hasStrataBug := laurelTranslateErrors.any
          (fun d => d.type == DiagnosticType.StrataBug)
        -- If any diagnostic is a StrataBug, treat the whole failure as internal
        -- (a Strata bug takes priority over known limitations).
        if hasStrataBug then
```

tautschnig (commented):
Two concerns:

1. **No test coverage for either fix.** The `isinstance` in `assert` fix (PythonToLaurel.lean) should have a regression test. The diagnostic classification fix (StrataMain.lean) is harder to unit-test but could at least be covered by the fuzz seed that found it.

2. **The `isHoisted` flag is a proxy for type information.** The real invariant is "the condition expression is already `bool`-typed, so `Any_to_bool` should not be applied." Using a boolean flag works for this specific case but is fragile if the hoisting logic changes.

See inline comments.

julesmt:
@tautschnig Updated with your comments

PR #1107, opened 2026-05-05
Fix SMT translation of integer literals

Robertboy18:
*Issue #, if available:*

N/A. Found while working on leanprover/cslib#539.

*Description of changes:*

This fixes SMT translation of integer literals. Previously, `Strata.DL.SMT.Translate.translateTerm` translated `.prim (.int x)` as having Lean type `Prop`; this PR changes it to `Int`.

This matters for integer-valued SMT `ite` terms. In CSLib/Boole (Check: https://github.com/leanprover/cslib/pull/539), an expression like:

`if i + 1 == n then 0 else i + 1`

was producing a VC elaboration error because the branch literal `0` had type `Int` but was expected to have type `Prop`.

This PR also adds a regression test covering an integer-valued SMT `ite`.

Checks run:
- `lake env lean StrataTest/DL/SMT/TranslateTests.lean`
- `lake env lean StrataTest/DL/SMT/DenoteTests.lean`
- `lake env lean StrataTest/DL/SMT/EncoderTests.lean`
- `lake build Strata.DL.SMT.Translate StrataTest.DL.SMT.TranslateTests`

Let me know if you have any questions!

### DIFF
--- a/Strata/DL/SMT/Translate.lean
+++ b/Strata/DL/SMT/Translate.lean
@@ -325,7 +325,7 @@ def translateTerm (t : SMT.Term) : TranslateM (Expr × Expr) := do
     let (as, a) := ((a :: as).dropLast, (a :: as).getLast?.get rfl)
     return (mkProp, as.foldr mkArrow a)
   | .prim (.int x) =>
-    return (mkProp, toExpr x)
+    return (mkInt, toExpr x)
   | .app .neg [a] _ =>
     let (_, a) ← translateTerm a
     return (mkInt, .app mkIntNeg a)
--- a/Strata/DL/SMT/Translate.lean
+++ b/Strata/DL/SMT/Translate.lean
@@ -274,6 +274,14 @@ def translateSort (ty : TermType) : TranslateM Expr := do
     let as ← as.mapM translateSort
     return mkAppN t as.toArray
 
+/--
+Translate an SMT term to a Lean expression, together with its Lean type.
+
+The first component is the actual type of the second component, not just a
+hint. Consumers use it as the type argument for generated Lean constructs such
+as `Eq` and `ite`, and helpers such as `leftAssocOp` propagate it as the result
+type of compound expressions.
+-/
 def translateTerm (t : SMT.Term) : TranslateM (Expr × Expr) := do
   match t with
   | .var v =>
--- a/StrataTest/DL/SMT/TranslateTests.lean
+++ b/StrataTest/DL/SMT/TranslateTests.lean
@@ -14,6 +14,7 @@ open Elab Command
 def elabQuery (ctx : Core.SMT.Context) (assums : List SMT.Term) (conc : SMT.Term) : CommandElabM Unit := do
   runTermElabM fun _ => do
     let e ← translateQueryMeta ctx assums conc
+    Meta.check e
     logInfo e
 
 /-- info: ∀ (a : Int), 42 > a -/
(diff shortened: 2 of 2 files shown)


Strata/DL/SMT/Translate.lean
```diff
@@ -325,7 +325,7 @@ def translateTerm (t : SMT.Term) : TranslateM (Expr × Expr) := do
     let (as, a) := ((a :: as).dropLast, (a :: as).getLast?.get rfl)
     return (mkProp, as.foldr mkArrow a)
   | .prim (.int x) =>
-    return (mkProp, toExpr x)
+    return (mkInt, toExpr x)
```
tautschnig, line 336: Fix is correct: this now agrees with `translateSort (.prim .int) = mkInt` at line 258 and with `.prim (.bitvec (n := n) x) => (mkBitVec n, toExpr x)` at line 363. The other literal-returning arms (`.prim (.bool b)` → `(mkProp, ...)`) were already consistent with `translateSort`.

Worth noting for future readers — and worth committing to in a docstring on `translateTerm` — that the first component of the returned pair is not an incidental hint: it is the *actual* Lean type of the second component and is consumed as such by `.app .eq` (Lean `Eq` type arg), `.app .ite` (Lean `ite` type arg), and every arithmetic op via `leftAssocOp` (return type of the whole expression). That contract is currently carried only by convention, which is exactly why this slipped in; a short doc-comment making it explicit would make the next similar typo obvious at review time.

StrataTest/DL/SMT/TranslateTests.lean
```diff
@@ -41,3 +41,10 @@ info: ∀ (α : Type → Type → Type) [inst : ∀ (α_1 α_2 : Type), Nonempty
 #guard_msgs in
 #eval
   elabQuery {} [] (.app .eq [(.app .abs [(.prim (.int (-5)))] (.prim .int)), (.prim (.int 5))] (.prim .bool))
+
+/-- info: (if 0 = 0 then 0 else 1) = 0 -/
+#guard_msgs in
+#eval
+  let c := .app .eq [(.prim (.int 0)), (.prim (.int 0))] (.prim .bool)
+  let t := .app .ite [c, (.prim (.int 0)), (.prim (.int 1))] (.prim .int)
```
tautschnig, line 50: This test, as written, passes both with and without the fix. I verified locally by reverting `mkInt` back to `mkProp` on line 328 of `Translate.lean` and re-running just this block — the `#guard_msgs` still succeeds, printing `(if 0 = 0 then 0 else 1) = 0` as expected. The reason is that `elabQuery` only calls `logInfo e` and pretty-printing a malformed `Expr` (one whose declared Lean type contradicts its payload) does not trigger any type-check, so this is not actually a regression test for the bug the PR fixes.

There are two fixes to apply together:

**(a) Make the harness force a type-check.** The minimal change to `elabQuery` (lines 14–17, outside this hunk):

```lean
def elabQuery (ctx : Core.SMT.Context) (assums : List SMT.Term) (conc : SMT.Term) : CommandElabM Unit := do
  runTermElabM fun _ => do
    let e ← translateQueryMeta ctx assums conc
    Meta.check e
    logInfo e
```

Once that is in place, the existing test block fails without the fix with:

```
Application type mismatch: The argument
  0
has type
  Int
but is expected to have type
  Prop
in the application
  Eq 0
```

and passes cleanly with the fix. (The pre-existing tests above — including the `abs` one on line 43 — continue to pass under `Meta.check` both with and without the fix, because in those the raw int literal only appears as the second operand of `eq`, so its `α` is never consumed.)

**(b) Extend the test block to also exercise the other broken code paths.** The original bug also manifested when an int literal was the first operand of `.app .eq` or of any `leftAssocOp` arithmetic (`add`/`sub`/`mul`/`div`/`mod`), not just as an ITE branch. I verified locally that, under `Meta.check` and with the fix reverted, each of the extra tests below fails with an "Int vs Prop" mismatch; with the fix applied, all of them pass and produce the expected pretty-printed output.

```suggestion
/-- info: (if 0 = 0 then 0 else 1) = 0 -/
#guard_msgs in
#eval
  let c := .app .eq [(.prim (.int 0)), (.prim (.int 0))] (.prim .bool)
  let t := .app .ite [c, (.prim (.int 0)), (.prim (.int 1))] (.prim .int)
  elabQuery {} [] (.app .eq [t, (.prim (.int 0))] (.prim .bool))

-- Int literal as the *first* operand of `.app .eq`: previously built
-- `@Eq Prop 5 ...` because `.prim (.int 5)` returned `(mkProp, _)`.
/-- info: 5 = -5 -/
#guard_msgs in
#eval
  elabQuery {} [] (.app .eq [(.prim (.int 5)), (.app .neg [(.prim (.int 5))] (.prim .int))] (.prim .bool))

-- Int literal as the *first* operand of arithmetic: `leftAssocOp`
-- propagates the first operand's type as the whole expression's type,
-- so before the fix the entire `add` was typed `Prop`.
/-- info: 1 + 2 = 3 -/
#guard_msgs in
#eval
  elabQuery {} [] (.app .eq [(.app .add [(.prim (.int 1)), (.prim (.int 2))] (.prim .int)), (.prim (.int 3))] (.prim .bool))

-- Int literal as a branch of a nested ITE: exercises α propagation
-- across nesting.
/-- info: (if 0 = 0 then if 0 = 0 then 1 else 2 else 3) = 1 -/
#guard_msgs in
#eval
  let c := .app .eq [(.prim (.int 0)), (.prim (.int 0))] (.prim .bool)
  let inner := .app .ite [c, (.prim (.int 1)), (.prim (.int 2))] (.prim .int)
  let outer := .app .ite [c, inner, (.prim (.int 3))] (.prim .int)
  elabQuery {} [] (.app .eq [outer, (.prim (.int 1))] (.prim .bool))
```

Note that the one-click apply on the suggestion above only replaces lines 44–49; the `elabQuery` change in (a) still has to be made manually. Happy to split this into two suggestion blocks if that's easier.

Strata/DL/SMT/Translate.lean
Robertboy18: Thanks: ) and I addressed them and added the doc-comment to `translateTerm` spelling out that the first component of the returned pair is the actual Lean type of the second component!

StrataTest/DL/SMT/TranslateTests.lean
Robertboy18: Good point! I made the update that makes `elabQuery` call `Meta.check e`, so the tests type-check the generated expression rather than only pretty-printing it and extends the regression tests to cover integer literals in an `ite`, as the first operand of `eq`, as the first operand of arithmetic, and inside a nested `ite`!

Let me know if anything else is needed!

Robertboy18:
I reran all the tests and it all worked!

tautschnig (approved):

aqjune-aws (approved):

aqjune-aws:
We will have to go through the shepherding process. (https://github.com/strata-org/Strata/blob/main/CONTRIBUTING.md#two-step-review-and-finding-a-shepherd). I will create a separate pull request which will include preserved commits under your email.

aqjune-aws:
#1123

PR #1112, opened 2026-05-05
Escape SMT-LIB set-info strings per SMT-LIB 2.6+ (doubled quotes)

tautschnig:
`Verifier.encodeCore` was rendering `(set-info :final-message "...")` with C-style escaping (backslash + quote) for embedded double quotes and backslashes in the property summary.  That is invalid in SMT-LIB 2.6+: the backslash has no special meaning inside a string literal (except `\u{XXXX}` for Unicode escapes), so for input

    Expected len(kwargs["JobName"]) >= 1, got stringLen(kwargs[JobName])

the encoder emitted

    (set-info :final-message "Expected len(kwargs[\"JobName\"]) >= 1, got stringLen(kwargs[JobName])")

which an SMT-LIB parser reads as a string that ends after the first `\"` (a literal backslash followed by a string-closing quote), with the rest (`JobName\"...`) sitting outside any command and triggering a parse error.  Per the spec (§3.1.2), a literal double quote inside an SMT-LIB string must be written as two consecutive double quotes.

An `escapeSMTStringLit` helper that already does this (and also handles non-printable characters via `\u{XXXX}`) existed in `Strata/DDM/Util/String.lean` and is already used by the user-error path in `StrataMain.lean`.  Wire it into the `Solver` API as a typed `setInfoString` wrapper and use it everywhere a string-valued `set-info` attribute is emitted:

- `Verifier.lean`: the `:final-message` site (the site the reported failure hit) plus the `:sat-message` / `:unsat-message` sites in `addLocationInfo`.  The `sat-message` / `unsat-message` callers previously pre-quoted their argument; they now pass the raw Lean string.
- `SMTUtils.lean`: the `:file` site and the forwarded application- supplied `(name, value)` message.  The latter's contract changes from "pre-quoted value" to "raw Lean string" accordingly.

The plain `setInfo` remains available for attribute values that are already valid SMT-LIB tokens (integers, s-expressions), and is still used for `:start` / `:stop`.

Regression tests in `SMTEncoderTests.lean` cover both the reported case (embedded double quotes escaping to `""`) and the complementary case (embedded backslash staying literal, no escape).  Each fails on origin/main and passes here.

### DIFF
--- a/Strata/DL/Imperative/SMTUtils.lean
+++ b/Strata/DL/Imperative/SMTUtils.lean
@@ -275,17 +275,21 @@ def solverResult {P : PureExpr} [ToFormat P.Ident]
   | none =>
     return .error (mkError output)
 
+/-- Emit a block of `set-info` attributes describing the source location and
+    an optional application-specific `(name, rawValue)` pair. `message.snd` is
+    a raw Lean string; it will be quoted and SMT-LIB-escaped (doubled `""`)
+    before emission. -/
 def addLocationInfo {P : PureExpr} [BEq P.Ident]
   (md : Imperative.MetaData P) (message : String × String)
   : Strata.SMT.SolverM Unit := do
   match Imperative.getFileRange md with
     | .some fileRange => do
-      Strata.SMT.Solver.setInfo "file" s!"\"{format fileRange.file}\""
+      Strata.SMT.Solver.setInfoString "file" (toString (format fileRange.file))
       Strata.SMT.Solver.setInfo "start" s!"{fileRange.range.start}"
       Strata.SMT.Solver.setInfo "stop" s!"{fileRange.range.stop}"
       -- TODO: the following should probably be stored in metadata so it can be
       -- set in an application-specific way.
-      Strata.SMT.Solver.setInfo message.fst message.snd
+      Strata.SMT.Solver.setInfoString message.fst message.snd
     | .none => pure ()
 
 /--
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -103,14 +103,14 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
     -- Satisfiability check: P ∧ Q satisfiable?
     Solver.comment "Satisfiability"
     Imperative.SMT.addLocationInfo (P := Core.Expression) (md := md)
-      (message := ("sat-message", s!"\"Property can be satisfied\""))
+      (message := ("sat-message", "Property can be satisfied"))
     let obligationStr ← Solver.termToSMTString obligationId
     let _ ← Solver.checkSatAssuming [obligationStr] ids
 
     -- Validity check: P ∧ ¬Q satisfiable?
     Solver.comment "Validity"
     Imperative.SMT.addLocationInfo (P := Core.Expression) (md := md)
-      (message := ("unsat-message", s!"\"Property is always true\""))
+      (message := ("unsat-message", "Property is always true"))
     let negObligationStr := s!"(not {obligationStr})"
     let _ ← Solver.checkSatAssuming [negObligationStr] ids
   else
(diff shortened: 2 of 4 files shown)


Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -275,17 +275,21 @@ def solverResult {P : PureExpr} [ToFormat P.Ident]
   | none =>
     return .error (mkError output)
 
+/-- Emit a block of `set-info` attributes describing the source location and
+    an optional application-specific `(name, rawValue)` pair. `message.snd` is
+    a raw Lean string; it will be quoted and SMT-LIB-escaped (doubled `""`)
+    before emission. -/
```
atomb, line 281: Thanks for documenting this. :)

StrataTest/Languages/Core/Tests/SMTEncoderTests.lean
```diff
@@ -370,6 +370,71 @@ info: (set-logic ALL)
...
+/--
+info: (set-logic ALL)
+; Validity
+(assert false)
+(check-sat)
+(set-info :final-message "Expected len(kwargs[""JobName""]) >= 1, got stringLen(kwargs[JobName])")
+-/
+#guard_msgs in
+#eval show IO _ from do
+  let ctx : SMT.Context := SMT.Context.default
```
atomb: These two tests have almost identical bodies. Could you abstract the code out into a separate function?

atomb (dismissed):

StrataTest/Languages/Core/Tests/SMTEncoderTests.lean
tautschnig: Addressed in [c1dd44c](https://github.com/strata-org/Strata/pull/1112/commits/c1dd44ca26a556daa076c1f0e0401ffe344cd845).

Strata/Languages/Core/Verifier.lean
```diff
@@ -118,21 +118,24 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
       -- P ∧ Q satisfiable?
       Solver.comment "Satisfiability"
       Imperative.SMT.addLocationInfo (P := Core.Expression) (md := md)
-        (message := ("sat-message", s!"\"Property can be satisfied\""))
+        (message := ("sat-message", "Property can be satisfied"))
```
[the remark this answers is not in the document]
tautschnig: Addressed in [c1dd44c](https://github.com/strata-org/Strata/pull/1112/commits/c1dd44ca26a556daa076c1f0e0401ffe344cd845).

StrataTest/Languages/Core/Tests/SMTEncoderTests.lean
```diff
@@ -370,6 +370,71 @@ info: (set-logic ALL)
...
+  let obligationTerm := Term.prim (.bool true)
+  let msg := "Expected len(kwargs[\"JobName\"]) >= 1, got stringLen(kwargs[JobName])"
+  let md : Imperative.MetaData Core.Expression :=
+    Imperative.MetaData.empty.withPropertySummary msg
+  let b ← IO.mkRef { : IO.FS.Stream.Buffer }
+  let solver ← Strata.SMT.Solver.bufferWriter b
+  let _ ←
+    Strata.SMT.SolverM.run solver
+      (Strata.SMT.Encoder.encodeCore ctx (pure ()) [] obligationTerm md
+        (satisfiabilityCheck := false) (validityCheck := true) (label := "test"))
```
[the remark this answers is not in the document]
tautschnig: Addressed in [c1dd44c](https://github.com/strata-org/Strata/pull/1112/commits/c1dd44ca26a556daa076c1f0e0401ffe344cd845).

atomb (approved):

aqjune-aws (approved):
Indeed, `escapeSMTStringLit` was added from https://github.com/strata-org/Strata/pull/541 and it was used if `FormatOptions.smtStringEscaping` was set to true, but this wasn't covering the case when String was being directly printed like this `set-info` case.

PR #1116, opened 2026-05-05
Add resolution diagnostic for multi-output procedure in expression position

tautschnig:
When a StaticCall to a multi-output procedure appears in expression position (not as the RHS of an Assign), emit an error diagnostic: the extra outputs would be silently discarded, which is a semantic bug (e.g., error channels lost).

The diagnostic is suppressed when the call is the direct RHS of an assignment (where the target count check already validates arity).

Includes a test in ResolutionKindTests demonstrating the diagnostic.

### DIFF
--- a/Strata/Languages/Laurel/Resolution.lean
+++ b/Strata/Languages/Laurel/Resolution.lean
@@ -213,8 +213,10 @@ structure ResolveState where
   /-- When resolving inside an instance procedure, the owning composite type name.
       Used by `resolveFieldRef` to resolve `self.field` when `self` has type `Any`. -/
   instanceTypeName : Option String := none
-  /-- True when resolving the RHS of an Assign statement (suppresses multi-output diagnostic). -/
-  inAssignRhs : Bool := false
+  /-- True when resolving inside an expression where the value is used (e.g., as an
+      argument to another call or operator). Multi-output calls are only diagnosed
+      in value context, not in statement position or direct assignment RHS. -/
+  inValueContext : Bool := false
 
 @[expose] abbrev ResolveM := StateM ResolveState
 
@@ -404,10 +406,7 @@ def resolveStmtExpr (exprMd : StmtExprMd) : ResolveM StmtExprMd := do
         let ty' ← resolveHighType param.type
         let name' ← defineNameCheckDup param.name (.var param.name ty')
         pure (⟨.Declare ⟨name', ty'⟩, vs⟩ : VariableMd)
-    let savedFlag := (← get).inAssignRhs
-    modify fun s => { s with inAssignRhs := true }
     let value' ← resolveStmtExpr value
-    modify fun s => { s with inAssignRhs := savedFlag }
     -- Check that LHS target count matches the number of outputs from the RHS.
     -- This fires for procedure calls (which can have multiple outputs).
     -- Functions always have exactly 1 output in the model, so single-target function calls pass trivially.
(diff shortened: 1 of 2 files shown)


Strata/Languages/Laurel/Resolution.lean
```diff
@@ -402,7 +404,10 @@ def resolveStmtExpr (exprMd : StmtExprMd) : ResolveM StmtExprMd := do
         let ty' ← resolveHighType param.type
         let name' ← defineNameCheckDup param.name (.var param.name ty')
         pure (⟨.Declare ⟨name', ty'⟩, vs⟩ : VariableMd)
+    let savedFlag := (← get).inAssignRhs
+    modify fun s => { s with inAssignRhs := true }
```
keyboardDrummer: I think this won't work for issues that occur inside the RHS of an assignment. You need to set `inAssignRhs` to false when you find things that aren't assignments.

However, I think the easier to understand fix is to do type checking, which wouldn't rely on state. Could we ask a bot to add that to Resolution.lean? I don't think it's very difficult.

Something like:

Please add type checking to Resolution.lean. Let `resolveStmtExpr` return a `ResolveM (StmtExprMd x HighTypeMd)`and let the case for each StmtExpr constructor check that the types of its children are what is expected, if a particular type is expected.

Assigned this prompt to kbd-bot: https://github.com/strata-org/Strata/issues/1120

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -438,6 +443,22 @@ def resolveStmtExpr (exprMd : StmtExprMd) : ResolveM StmtExprMd := do
...
+      | some (_, .staticProcedure proc) =>
+        if proc.outputs.length > 1 then
+          let diag := diagnosticFromSource source
+            s!"Multi-output procedure '{callee'.text}' used in expression position; it returns {proc.outputs.length} values but only one can be used here. Use a multi-target assignment instead."
+          modify fun s => { s with errors := s.errors.push diag }
+      | some (_, .instanceProcedure _ proc) =>
+        if proc.outputs.length > 1 then
+          let diag := diagnosticFromSource source
+            s!"Multi-output procedure '{callee'.text}' used in expression position; it returns {proc.outputs.length} values but only one can be used here. Use a multi-target assignment instead."
+          modify fun s => { s with errors := s.errors.push diag }
```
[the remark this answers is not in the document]
tautschnig: Done in [352fdf8](https://github.com/strata-org/Strata/pull/1116/commits/352fdf81c6d974c37d10b08f52f53c5c72941476).

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -402,7 +404,10 @@ def resolveStmtExpr (exprMd : StmtExprMd) : ResolveM StmtExprMd := do
         let ty' ← resolveHighType param.type
         let name' ← defineNameCheckDup param.name (.var param.name ty')
         pure (⟨.Declare ⟨name', ty'⟩, vs⟩ : VariableMd)
+    let savedFlag := (← get).inAssignRhs
+    modify fun s => { s with inAssignRhs := true }
     let value' ← resolveStmtExpr value
+    modify fun s => { s with inAssignRhs := savedFlag }
```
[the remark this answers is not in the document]
tautschnig: Addressed in [352fdf8](https://github.com/strata-org/Strata/pull/1116/commits/352fdf81c6d974c37d10b08f52f53c5c72941476)[352fdf8](https://github.com/strata-org/Strata/pull/1116/commits/352fdf81c6d974c37d10b08f52f53c5c72941476).

Strata/Languages/Laurel/Resolution.lean
tautschnig: Done in [352fdf8](https://github.com/strata-org/Strata/pull/1116/commits/352fdf81c6d974c37d10b08f52f53c5c72941476)[352fdf8](https://github.com/strata-org/Strata/pull/1116/commits/352fdf81c6d974c37d10b08f52f53c5c72941476) (not the full type checking, though, which is left for #1120).

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -438,6 +443,22 @@ def resolveStmtExpr (exprMd : StmtExprMd) : ResolveM StmtExprMd := do
     let callee' ← resolveRef callee source
       (expected := #[.parameter, .staticProcedure, .datatypeConstructor, .constant])
     let args' ← args.mapM resolveStmtExpr
+    -- Multi-output procedures must not appear in expression position: the extra
+    -- outputs (e.g. error channels) would be silently discarded.
+    let s ← get
+    if !s.inAssignRhs then
+      match s.scope.get? callee'.text with
+      | some (_, .staticProcedure proc) =>
+        if proc.outputs.length > 1 then
```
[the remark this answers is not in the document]
tautschnig: Done.

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -437,10 +441,30 @@ def resolveStmtExpr (exprMd : StmtExprMd) : ResolveM StmtExprMd := do
...
+      let outputCount := match s.scope.get? callee'.text with
+        | some (_, .staticProcedure proc) => proc.outputs.length
+        | some (_, .instanceProcedure _ proc) => proc.outputs.length
+        | _ => 0
+      if outputCount > 1 then
+        let diag := diagnosticFromSource source
+          s!"Multi-output procedure '{callee'.text}' used in expression position; it returns {outputCount} values but only one can be used here. Use a multi-target assignment instead."
+        modify fun s => { s with errors := s.errors.push diag }
     pure (.StaticCall callee' args')
   | .PrimitiveOp op args =>
```
keyboardDrummer, line 468: Wouldn't there be more places that have a value context? What about an assert of assume statement, or the condition of an if-then-else?

tautschnig: Fixed in [1713b14](https://github.com/strata-org/Strata/pull/1116/commits/1713b147ed3e30758bf3540ef6b632249a59edaf)[1713b14](https://github.com/strata-org/Strata/pull/1116/commits/1713b147ed3e30758bf3540ef6b632249a59edaf).

keyboardDrummer:
How about instead of this PR we look at merging https://github.com/strata-org/Strata/pull/1121 ? I know this PR is a strict improvement, but I think it might be confusing to temporarily have this code. We'll have to make sure to remove it later as well. Since it's so easy to add a more long term solution, shouldn't we do that?

keyboardDrummer (approved):
Prefer to wait for https://github.com/strata-org/Strata/pull/1121, but approving in case you prefer not to wait

ssomayyajula (approved):

PR #1117, opened 2026-05-05
Fix multi-output calls in expression position (Python front-end)

tautschnig:
When translateExpr produces an expression containing a nested multi-output procedure call (e.g., PSub(base, timedelta_func(...))), the new extractMultiOutputCalls pass in translateAssign rewrites it into a preceding multi-target assignment and a variable reference:

  var $mo_N := ...; assign $mo_N, maybe_except := timedelta_func(...);
  delta := PSub(base, $mo_N)

This prevents multi-output calls from appearing in expression position, which would silently discard the error channel.

Key design choices:
- Counter seeded from source byte offset for globally unique temp variable names
- Branch-local multi-output calls are wrapped in Block nodes inside their respective IfThenElse branches (not hoisted unconditionally)
- Generated wrapper statements propagate source metadata from the original call
- Block-flattening in the lifter uses cons-based prepend for correct evaluation order
- containsBareAssignment skips Blocks so generated wrappers pass through assert/assume handlers
- extractMultiOutputCalls uses structural recursion (attach) rather than partial

Tested: existing tests pass, builds cleanly.


Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1336,6 +1336,79 @@ def freeVarExpr (name: String) := mkStmtExprMd (.Var (.Local name))
...
+      return (preamble, mkStmtExprMdWithLoc (.PrimitiveOp op newArgs) e.source)
+  | .IfThenElse cond thenBr elseBr =>
+    let (preCond, cond') ← extractMultiOutputCalls ctx cond
+    let (preThen, then') ← extractMultiOutputCalls ctx thenBr
+    let preElse ← elseBr.mapM (extractMultiOutputCalls ctx)
+    let thenExpr :=
+      if preThen.isEmpty then
+        then'
+      else
+        mkStmtExprMdWithLoc (.Block (preThen ++ [then'])) thenBr.source
```
MikaelMayer: 🤖🔍 Build failure: `.Block` requires a second `(label : Option String)` argument. Should be `.Block (preThen ++ [then']) none`. Same issue on line 1396.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1371,7 +1444,12 @@ partial def translateAssign  (ctx : TranslationContext)
     | .Attribute _ (.Name _ name _) _ _ => name.val == "self" && ctx.currentClassName.isSome
     | _ => false
   let rhsCtx := if isSelfFieldAssign then {ctx with suppressDispatch := true} else ctx
-  let rhs_trans ←  translateExpr rhsCtx rhs
+  let extractionSeedText :=
+    match pyRangeToFileRange rhs.ann with
```
MikaelMayer: 🤖🔍 Build failure: `pyRangeToFileRange` and `rangeToTempName` (line 1449) are not defined in this file or its imports. The seed computation doesn't compile.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1336,6 +1336,79 @@ def freeVarExpr (name: String) := mkStmtExprMd (.Var (.Local name))
...
+      if preThen.isEmpty then
+        then'
+      else
+        mkStmtExprMdWithLoc (.Block (preThen ++ [then'])) thenBr.source
+    let elseExpr := preElse.map fun (pre, else') =>
+      if pre.isEmpty then
+        else'
+      else
+        mkStmtExprMdWithLoc (.Block (pre ++ [else'])) else'.source
+    if preCond.isEmpty then
```
MikaelMayer: 🤖🔍 When `preCond` is empty but `preThen` or `preElse` is non-empty, this returns `([], e)` — discarding the branch rewrites. A multi-output call nested in a then/else branch (but not the condition) would be silently ignored. The guard should also check whether any branch preamble is non-empty.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1400,6 +1478,11 @@ partial def translateAssign  (ctx : TranslationContext)
         return (newctx, [initStmt] ++ exceptHavoc, true)
     | _ => return (ctx, [mkStmtExprMd .Hole] ++ exceptHavoc, false)
   }
+  -- When the RHS is a direct multi-output call (top-level), translateAssign
+  -- already handles it with multi-target assignment. Don't double-extract.
+  let moExtracts := match rhs_trans.val with
```
MikaelMayer: 🤖🔍 This guard is dead code. `extractMultiOutputCalls` already rewrites a top-level `withException` call into a `Var` reference, so `rhs_trans.val` can never be `.StaticCall callee _` with `withException ctx callee.text` true by the time this check runs.

MikaelMayer (commented):
🤖🔍 The latest commit (`7c11a98`) introduces build failures. Beyond those, the IfThenElse branch-extraction logic has a semantic gap.

MikaelMayer (commented):
🤖🔍 Two of four previous comments were addressed (`.Block` label argument and undefined `pyRangeToFileRange`/`rangeToTempName`). Two remain:

1. **Branch preamble discarded when `preCond.isEmpty`** (line ~1397): When `preCond` is empty but `preThen` or `preElse` is non-empty, the code still returns `([], e)` — the *original* expression — discarding the computed `thenExpr`/`elseExpr` that contain the branch-local extractions. A multi-output call nested only inside a then/else branch (not the condition) would be silently ignored. The fix should return the rewritten `IfThenElse` even when `preCond` is empty.

2. **Dead-code guard at line ~1484**: `extractMultiOutputCalls` already rewrites a top-level `withException` call into a `Var` reference, so `rhs_trans.val` can never match `.StaticCall callee _` with `withException ctx callee.text == true` by the time the guard runs. This is harmless but misleading — consider removing it or adding a comment explaining why it's kept as a safety net.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1371,7 +1431,7 @@ partial def translateAssign  (ctx : TranslationContext)
     | .Attribute _ (.Name _ name _) _ _ => name.val == "self" && ctx.currentClassName.isSome
     | _ => false
   let rhsCtx := if isSelfFieldAssign then {ctx with suppressDispatch := true} else ctx
-  let rhs_trans ←  translateExpr rhsCtx rhs
+  let (moExtracts, rhs_trans, _) ← translateExprExtractingCalls rhsCtx rhs 0
```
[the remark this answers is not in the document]
MikaelMayer: 🤖 Already addressed: the counter starts at `extractionSeed` (a hash of the source range text), not `0`. Each assignment site produces a unique seed, so temp names like `$mo_N` won't collide across assignments.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1336,6 +1336,66 @@ def freeVarExpr (name: String) := mkStmtExprMd (.Var (.Local name))
...
+        return ([], e)
+      else
+        return (preamble, mkStmtExprMdWithLoc (.StaticCall callee.text newArgs) e.source)
+  | .PrimitiveOp op args =>
+    let mut preamble : List StmtExprMd := []
+    let mut newArgs : List StmtExprMd := []
+    for arg in args do
+      let (pre, arg') ← extractMultiOutputCalls ctx arg
+      preamble := preamble ++ pre
+      newArgs := newArgs ++ [arg']
```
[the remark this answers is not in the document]
MikaelMayer: 🤖 Already addressed: the loops now build lists in reverse (`preambleRev`/`newArgsRev`) using cons, then reverse at the end — avoiding quadratic `++`.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -345,7 +363,23 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
+      -- Flatten generated multi-output call wrappers: Blocks of the form
+      -- [VarDecl, MultiAssign, VarRef] where the assign has 2+ targets.
+      match filtered with
+      | [decl, assign, last] =>
+        match decl.val, assign.val with
+        | .Assign [t] _, .Assign targets _ =>
+          match t.val with
+          | .Declare _ =>
+            if targets.length ≥ 2 then
+              modify fun s => { s with prependedStmts := s.prependedStmts ++ [decl, assign] }
```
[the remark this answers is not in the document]
MikaelMayer: 🤖 Already addressed: the Block-flattening now uses `prepend assign; prepend decl` (cons-based), which maintains correct evaluation order with the right-to-left traversal.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -345,7 +363,23 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
+              return last
+            else
+              return ⟨ .Block filtered labelOption, source⟩
+          | _ => return ⟨ .Block filtered labelOption, source⟩
+        | _, _ => return ⟨ .Block filtered labelOption, source⟩
+      | _ => return ⟨ .Block filtered labelOption, source⟩
```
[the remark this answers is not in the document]
MikaelMayer: 🤖 Safe: the flattening only triggers for the specific `[VarDecl, MultiAssign(2+ targets), VarRef]` pattern, which is exclusively generated by `extractMultiOutputCalls` with `none` label. User-authored labeled blocks won't match this pattern.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Already fixed in a previous commit — `.Block` now receives `none` as the second argument on both lines.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Already fixed in a previous commit — replaced `pyRangeToFileRange`/`rangeToTempName` with a simple hash computation using `rhs.ann.start`/`rhs.ann.stop` and `String.foldl`. Compiles cleanly now.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1336,6 +1336,66 @@ def freeVarExpr (name: String) := mkStmtExprMd (.Var (.Local name))
...
+  | .IfThenElse cond thenBr elseBr =>
+    let (preCond, cond') ← extractMultiOutputCalls ctx cond
+    let (preThen, then') ← extractMultiOutputCalls ctx thenBr
+    let preElse ← elseBr.mapM (extractMultiOutputCalls ctx)
+    let allPre := preCond ++ preThen ++ (preElse.map Prod.fst).getD []
+    if allPre.isEmpty then
+      return ([], e)
+    else
+      return (allPre, mkStmtExprMdWithLoc
+        (.IfThenElse cond' then' (preElse.map Prod.snd)) e.source)
```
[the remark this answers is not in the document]
MikaelMayer: 🤖 Fixed: the IfThenElse case now checks `anyRewrite` (whether any of preCond/preThen/preElse is non-empty) before deciding to return the rewritten expression. Branch-only multi-output calls are no longer discarded. Also propagated source metadata to generated wrapper statements. See 23cff9f9.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1336,6 +1336,66 @@ def freeVarExpr (name: String) := mkStmtExprMd (.Var (.Local name))
...
+      let mut preamble : List StmtExprMd := []
+      let mut newArgs : List StmtExprMd := []
+      for arg in args do
+        let (pre, arg') ← extractMultiOutputCalls ctx arg
+        preamble := preamble ++ pre
+        newArgs := newArgs ++ [arg']
+      if preamble.isEmpty then
+        return ([], e)
+      else
+        return (preamble, mkStmtExprMdWithLoc (.StaticCall callee.text newArgs) e.source)
```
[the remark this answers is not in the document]
MikaelMayer: 🤖 Fixed: source metadata is now propagated to the generated `assign` and `varRef` statements using `mkStmtExprMdWithLoc`. The rebuilt call also preserves the original source. See 23cff9f9.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Fixed: the condition now checks `anyRewrite` (preCond, preThen, or preElse non-empty) instead of only `preCond.isEmpty`. Multi-output calls in then/else branches are no longer silently discarded. See 23cff9f9.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Removed: the guard was dead code since `extractMultiOutputCalls` already rewrites top-level `withException` calls into `Var` references — `rhs_trans.val` can never be a `.StaticCall` with `withException` at that point. See 23cff9f9.

MikaelMayer (commented):

MikaelMayer (commented):

MikaelMayer (commented):

MikaelMayer (commented):

MikaelMayer (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1336,6 +1336,82 @@ def freeVarExpr (name: String) := mkStmtExprMd (.Var (.Local name))
 def maybeExceptVar := freeVarMd "maybe_except"
 def nullcall_var := freeVarMd "nullcall_ret"
 
+/-- Walk an expression tree and extract any nested multi-output procedure calls
+    into preceding multi-target assignments. Returns (preamble, rewritten expr).
+    Uses a mutable counter for unique variable names. -/
+partial def extractMultiOutputCalls (ctx : TranslationContext) (e : StmtExprMd)
```
joscoh: Is there a reason that this is `partial`? It seems to just be walking over `e.val` structurally (you can use `attach` for the list recursion as with many other functions throughout Strata)

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1336,6 +1336,82 @@ def freeVarExpr (name: String) := mkStmtExprMd (.Var (.Local name))
...
+        (mkStmtExprMdWithLoc (.StaticCall callee args) e.source)) e.source
+      let varRef := mkStmtExprMdWithLoc (StmtExpr.Var (.Local varName)) e.source
+      return ([varDecl, assign], varRef)
+    else
+      -- Recurse into arguments
+      let mut preambleRev : List StmtExprMd := []
+      let mut newArgsRev : List StmtExprMd := []
+      for arg in args do
+        let (pre, arg') ← extractMultiOutputCalls ctx arg
+        preambleRev := pre.foldl (fun acc stmt => stmt :: acc) preambleRev
```
joscoh: Nit: isn't this just `pre.reverse ++ preambleRev`?

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1371,7 +1447,16 @@ partial def translateAssign  (ctx : TranslationContext)
     | .Attribute _ (.Name _ name _) _ _ => name.val == "self" && ctx.currentClassName.isSome
     | _ => false
   let rhsCtx := if isSelfFieldAssign then {ctx with suppressDispatch := true} else ctx
-  let rhs_trans ←  translateExpr rhsCtx rhs
+  let extractionSeedText :=
+    if rhs.ann.isNone then
+      pyExprToString lhs ++ " <- " ++ pyExprToString rhs
+    else
+      s!"{rhs.ann.start}_{rhs.ann.stop}"
+  let extractionSeed := extractionSeedText.foldl (fun acc ch => acc * 131 + ch.toNat) 0
```
joscoh: This may not be globally unique. Is that a problem?

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Removed `partial` by using `attach` for list recursion and added a termination proof. Also simplified the preamble accumulation to use `foldl`/`map` on the results list directly. See 1e2ba1c3.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Simplified — the whole loop is now replaced by `args.attach.mapM` with `foldl`/`map` on the results, which avoids the manual reverse accumulation entirely. See 1e2ba1c3.

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Good catch — the hash was indeed not guaranteed unique. Now using `rhs.ann.start.byteIdx` directly as the seed (globally unique per source position). The hash fallback is only used when source location is unavailable. See 1e2ba1c3.

MikaelMayer (commented):

MikaelMayer (commented):

MikaelMayer (dismissed):

MikaelMayer (commented):

MikaelMayer (commented):

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -1356,35 +1351,25 @@ partial def extractMultiOutputCalls (ctx : TranslationContext) (e : StmtExprMd)
...
-      let mut preambleRev : List StmtExprMd := []
-      let mut newArgsRev : List StmtExprMd := []
-      for arg in args do
-        let (pre, arg') ← extractMultiOutputCalls ctx arg
-        preambleRev := pre.foldl (fun acc stmt => stmt :: acc) preambleRev
-        newArgsRev := arg' :: newArgsRev
-      let preamble := preambleRev.reverse
-      let newArgs := newArgsRev.reverse
+      let results ← args.attach.mapM fun ⟨arg, _⟩ => extractMultiOutputCalls ctx arg
+      let preamble := results.foldl (fun acc (pre, _) => acc ++ pre) []
```
joscoh: Nit (non-blocking): I believe this could be written more clearly as `results.map (fun <pre, _> => pre).flatten`.

joscoh (dismissed):

Strata/Languages/Python/PythonToLaurel.lean
MikaelMayer: 🤖 Done in 4fadb2b3 — replaced both occurrences with `.map (fun (pre, _) => pre)).flatten`.

MikaelMayer (approved):

joscoh (approved):

MikaelMayer (commented):

PR #1122, opened 2026-05-05
Add more documentation on the set up process following CI scripts

PROgram52bc:
*Description of changes:*

Added comprehensive set up instructions regarding required and optional dependencies (e.g., cvc5, z3, python strata, Java, ion) according to the CI workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,15 +19,88 @@ changes!**
 
 ## Prerequisites
 
-1. **Lean4**: Strata is built on Lean4; see the build specified in the
+1. **Lean4**: Strata is built on Lean4; see the version specified in the
    `lean-toolchain` file.
 
-   You can install Lean4 by following the instructions [here](https://lean-lang.org/).
+   Install Lean4 by following the instructions at [lean-lang.org](https://lean-lang.org/).
 
-2. **SMT Solver**: Analysis tools in Strata use SMT solvers for program
-   verification.
-   - Install an SMT solver. You can use any solver you want, but the unit
-     tests assume `cvc5` is on your `PATH` [cvc5](https://cvc5.github.io/).
+2. **SMT Solvers**: The verification pipeline and tests require SMT solvers
+   on your `PATH`.
+
+   - **cvc5** (required): [cvc5.github.io](https://cvc5.github.io/)
+   - **z3** (required by some tests): [github.com/Z3Prover/z3](https://github.com/Z3Prover/z3)
+
+3. **Python 3.11+** (required for Python-related tests and the `strata`
+   Python tooling):
+
+4. **Java JDK** (required for Java code generation tests):
+
+5. **ion-java jar** (required for the Java/Ion integration test):
+
+### Installing dependencies
+
+#### SMT Solvers
+
+Download static builds (single binary, no library dependencies):
+
+```bash
+# cvc5 (adjust URL for your OS/arch)
+# See https://github.com/cvc5/cvc5/releases for all platforms
+wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-Linux-x86_64-static.zip
+unzip cvc5-Linux-x86_64-static.zip
+cp cvc5-Linux-x86_64-static/bin/cvc5 /usr/local/bin/
+
+# z3
+# See https://github.com/Z3Prover/z3/releases for all platforms
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-x64-glibc-2.39.zip
+unzip z3-4.15.2-x64-glibc-2.39.zip
+cp z3-4.15.2-x64-glibc-2.39/bin/z3 /usr/local/bin/
+```
+
+On macOS (Apple Silicon), use the `arm64` variants and prefer the **static**
+build to avoid dynamic library issues:
+
+```bash
+# cvc5 static for macOS arm64
+wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.3.3/cvc5-macOS-arm64-static.zip
+unzip cvc5-macOS-arm64-static.zip
+cp cvc5-macOS-arm64-static/bin/cvc5 /usr/local/bin/
+
+# z3 for macOS arm64
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.16.0/z3-4.16.0-arm64-osx-15.7.3.zip
+unzip z3-4.16.0-arm64-osx-15.7.3.zip
+cp z3-4.16.0-arm64-osx-15.7.3/bin/z3 /usr/local/bin/
+```
+
+#### Python
+
+Python 3.11 or later is required. Install the Strata Python package:
+
+```bash
+pip install ./Tools/Python
+```
+
+This provides the `strata-python` tooling used by the Python verification
+pipeline and tests.
+
+#### Java (for code generation tests)
+
+A JDK (11+) providing `javac` must be on your `PATH`. Additionally,
+download the ion-java jar used by the Java/Ion integration test:
+
+```bash
+wget -q -O StrataTestExtra/DDM/Integration/Java/testdata/ion-java-1.11.11.jar \
+  https://github.com/amazon-ion/ion-java/releases/download/v1.11.11/ion-java-1.11.11.jar
+```
+
+### Verifying your setup
+
+```bash
+cvc5 --version    # should print version info
+z3 --version      # should print version info
+python3 --version # should be 3.11+
+javac --version   # should be 11+
+```
 
 ## Build
 
@@ -43,6 +116,19 @@ To build executable files only and omit proof checks that might take a long time
 
 ```bash
 lake build strata:exe strata StrataToCBMC StrataCoreToGoto
+```
+
+### Running specific test subsets
+
+```bash
+# Run all tests except Python (which requires the Python package)
+lake test -- --exclude Languages.Python
+
+# Run only Python tests (requires `pip install ./Tools/Python`)
+lake test -- Languages.Python
+
+# Run all tests
+lake test
 ```
 
 ## Running Analyses on Existing Strata Programs
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ cp z3-4.15.2-arm64-osx-15.7.3/bin/z3 /usr/local/bin/
 
 #### Python
 
-Python 3.11 or later is required. Install the Strata Python package:
+Python 3.11 or later is required. Install the `strata` Python package:
 
 ```bash
 pip install ./Tools/Python
--- a/README.md
+++ b/README.md
@@ -41,35 +41,43 @@ changes!**
 
 #### SMT Solvers
 
-Download static builds (single binary, no library dependencies):
+Download static builds (single binary, no library dependencies). The
+versions below are known to work; any newer release from the same links
+should also be fine.
+
+- cvc5 releases: https://github.com/cvc5/cvc5/releases
+- z3 releases: https://github.com/Z3Prover/z3/releases
+
+**Linux x86_64:**
 
 ```bash
-# cvc5 (adjust URL for your OS/arch)
-# See https://github.com/cvc5/cvc5/releases for all platforms
+# cvc5
 wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-Linux-x86_64-static.zip
 unzip cvc5-Linux-x86_64-static.zip
-cp cvc5-Linux-x86_64-static/bin/cvc5 /usr/local/bin/
+mkdir -p ~/.local/bin
+cp cvc5-Linux-x86_64-static/bin/cvc5 ~/.local/bin/
 
 # z3
-# See https://github.com/Z3Prover/z3/releases for all platforms
 wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-x64-glibc-2.39.zip
 unzip z3-4.15.2-x64-glibc-2.39.zip
-cp z3-4.15.2-x64-glibc-2.39/bin/z3 /usr/local/bin/
+cp z3-4.15.2-x64-glibc-2.39/bin/z3 ~/.local/bin/
+# Ensure ~/.local/bin is on your PATH (most modern distros include it by default).
 ```
 
-On macOS (Apple Silicon), use the `arm64` variants and prefer the **static**
+**macOS (Apple Silicon)** — use the `arm64` variants and prefer the **static**
 build to avoid dynamic library issues:
 
 ```bash
 # cvc5 static for macOS arm64
-wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.3.3/cvc5-macOS-arm64-static.zip
+wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-macOS-arm64-static.zip
 unzip cvc5-macOS-arm64-static.zip
 cp cvc5-macOS-arm64-static/bin/cvc5 /usr/local/bin/
 
 # z3 for macOS arm64
-wget https://github.com/Z3Prover/z3/releases/download/z3-4.16.0/z3-4.16.0-arm64-osx-15.7.3.zip
-unzip z3-4.16.0-arm64-osx-15.7.3.zip
-cp z3-4.16.0-arm64-osx-15.7.3/bin/z3 /usr/local/bin/
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-15.7.3.zip
+unzip z3-4.15.2-arm64-osx-15.7.3.zip
+cp z3-4.15.2-arm64-osx-15.7.3/bin/z3 /usr/local/bin/
+# On macOS, /usr/local/bin is typically writable without sudo.
 ```
 
 #### Python
@@ -127,7 +135,7 @@ lake test -- --exclude Languages.Python
 # Run only Python tests (requires `pip install ./Tools/Python`)
 lake test -- Languages.Python
 
-# Run all tests
+# Run all tests (requires the Python package above)
 lake test
 ```
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ changes!**
 
    - **cvc5** (required): [cvc5.github.io](https://cvc5.github.io/)
    - **z3** (required by some tests): [github.com/Z3Prover/z3](https://github.com/Z3Prover/z3)
-
+This provides the `strata` Python package (imported as `strata`) used by the
+Python verification pipeline and tests.
 3. **Python 3.11+** (required for Python-related tests and the `strata`
    Python tooling):
 
--- a/README.md
+++ b/README.md
@@ -25,12 +25,17 @@ changes!**
    Install Lean4 by following the instructions at [lean-lang.org](https://lean-lang.org/).
 
 2. **SMT Solvers**: The verification pipeline and tests require SMT solvers
+   (`cvc5` and `z3`). See [Installing dependencies → SMT Solvers](#smt-solvers)
+   below.
+
 3. **Python 3.11+**: required for Python-related tests and the `strata`
-   Python tooling.
+   Python tooling. See [Installing dependencies → Python](#python) below.
 
 4. **Java JDK (11 or later)**: required for Java code generation tests.
+   See [Installing dependencies → Java](#java-for-code-generation-tests) below.
 
 5. **ion-java jar (1.11.11)**: required for the Java/Ion integration test.
+   See [Installing dependencies → Java](#java-for-code-generation-tests) below.
 
 ### Installing dependencies
 
@@ -69,32 +74,32 @@ unzip cvc5-macOS-arm64-static.zip
 cp cvc5-macOS-arm64-static/bin/cvc5 /usr/local/bin/
 
 # z3 for macOS arm64
-wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-15.7.3.zip
-unzip z3-4.15.2-arm64-osx-15.7.3.zip
-cp z3-4.15.2-arm64-osx-15.7.3/bin/z3 /usr/local/bin/
-# On macOS, /usr/local/bin is typically writable without sudo.
+# z3 for macOS arm64
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-13.7.6.zip
+unzip z3-4.15.2-arm64-osx-13.7.6.zip
+sudo cp z3-4.15.2-arm64-osx-13.7.6/bin/z3 /usr/local/bin/
+# z3 for macOS arm64
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-13.7.6.zip
+unzip z3-4.15.2-arm64-osx-13.7.6.zip
+sudo cp z3-4.15.2-arm64-osx-13.7.6/bin/z3 /usr/local/bin/
+# Alternative: install into ~/.local/bin (no sudo), then ensure it's on your PATH.
 ```
 
 #### Python
 
-Python 3.11 or later is required. Install the `strata` Python package:
+Python 3.11 or later is required. Install the `strata` Python package inside a
+virtual environment (recommended; avoids PEP 668's `externally-managed-environment`
+error on Debian/Ubuntu 23.04+):
 
-```bash
-pip install ./Tools/Python
-```
-
-This provides the `strata-python` tooling used by the Python verification
-pipeline and tests.
 
 #### Java (for code generation tests)
 
 A JDK (11+) providing `javac` must be on your `PATH`. Additionally,
 download the ion-java jar used by the Java/Ion integration test:
 
-```bash
-wget -q -O StrataTestExtra/DDM/Integration/Java/testdata/ion-java-1.11.11.jar \
-  https://github.com/amazon-ion/ion-java/releases/download/v1.11.11/ion-java-1.11.11.jar
-```
+A JDK (11+) providing `javac` must be on your `PATH`. For running the
+Java/Ion integration test, download the ion-java jar:
+
 
 ### Verifying your setup
 
@@ -102,7 +107,6 @@ wget -q -O StrataTestExtra/DDM/Integration/Java/testdata/ion-java-1.11.11.jar \
 cvc5 --version    # should print version info
 z3 --version      # should print version info
 python3 --version # should be 3.11+
-javac --version   # should be 11+
 ```
 
 ## Build
@@ -123,16 +127,13 @@ lake build strata:exe strata StrataToCBMC StrataCoreToGoto
 
 ### Running specific test subsets
 
-```bash
-# Run all tests except Python (which requires the Python package)
-lake test -- --exclude Languages.Python
+Two kinds of tests coexist in this repo:
 
-# Run only Python tests (requires `pip install ./Tools/Python`)
-lake test -- Languages.Python
+- **Elaboration-time tests** (`#guard_msgs`) live under `StrataTest/` and run as
+  part of `lake build`. No output means they passed.
+- **Uncached extra tests** live under `StrataTestExtra/` and run via `lake test`.
+  These accept prefix filters:
 
-# Run all tests (requires the Python package above)
-lake test
-```
 
 ## Running Analyses on Existing Strata Programs
 
--- a/README.md
+++ b/README.md
@@ -41,35 +41,17 @@ changes!**
 
 #### SMT Solvers
 
-Download static builds (single binary, no library dependencies). The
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,15 +19,88 @@ changes!**
...
-2. **SMT Solver**: Analysis tools in Strata use SMT solvers for program
-   verification.
-   - Install an SMT solver. You can use any solver you want, but the unit
-     tests assume `cvc5` is on your `PATH` [cvc5](https://cvc5.github.io/).
+2. **SMT Solvers**: The verification pipeline and tests require SMT solvers
+   on your `PATH`.
+
+   - **cvc5** (required): [cvc5.github.io](https://cvc5.github.io/)
+   - **z3** (required by some tests): [github.com/Z3Prover/z3](https://github.com/Z3Prover/z3)
+
```
tautschnig, line 36: **Package name mismatch.** The PR body and this block both refer to the installed package as `strata-python`, but `Tools/Python/pyproject.toml:2` declares `name = "strata"` (version `0.0.1`). After `pip install ./Tools/Python`, the package appears on `pip list` as `strata`, not `strata-python`, and the importable module is `import strata` (see `Tools/Python/pyproject.toml:12`: `packages = ["strata"]`).

Suggested wording:

```suggestion
This provides the `strata` Python package (imported as `strata`) used by the
Python verification pipeline and tests.
```

Also related: the repo's developer-preferences steering note recommends installing Python packages in a venv (directory usually called `.venv`), which also sidesteps PEP 668 (`externally-managed-environment`) on Debian/Ubuntu 23.04+. A one-line preface would save every new contributor a detour:

```bash
python3 -m venv .venv
source .venv/bin/activate
pip install ./Tools/Python
```

README.md
```diff
@@ -19,15 +19,88 @@ changes!**
...
+```bash
+# cvc5 (adjust URL for your OS/arch)
+# See https://github.com/cvc5/cvc5/releases for all platforms
+wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-Linux-x86_64-static.zip
+unzip cvc5-Linux-x86_64-static.zip
+cp cvc5-Linux-x86_64-static/bin/cvc5 /usr/local/bin/
+
+# z3
+# See https://github.com/Z3Prover/z3/releases for all platforms
```
tautschnig: **Missing `sudo`.** Every CI job copies solver binaries as `sudo cp … /usr/local/bin/` — see `ci.yml:178`, `:300`, `:306`, `:311`. The README omits `sudo`, so any Linux developer without write access to `/usr/local/bin` (the default) hits "Permission denied" on these lines.

Three choices, pick one:

1. **Prefix with `sudo`** (matches CI exactly):
   ```bash
   sudo cp cvc5-Linux-x86_64-static/bin/cvc5 /usr/local/bin/
   ```
2. **Install to a user-local directory** and note the `PATH` requirement:
   ```bash
   mkdir -p ~/.local/bin
   cp cvc5-Linux-x86_64-static/bin/cvc5 ~/.local/bin/
   # Ensure ~/.local/bin is on your PATH (most modern distros already do this).
   ```
3. **Leave them unprefixed** and add a one-line note at the top of the block: `# These cp commands assume write access to /usr/local/bin; prefix with sudo if needed.`

Same applies to the z3 `cp` on the line below and to the macOS block (lines 57–64). The macOS arm64 block will also typically need `sudo` since `/usr/local/bin` is root-owned there too.

Related: the Linux snippet hard-codes `x86_64`. `ci.yml:38–49` auto-detects via `uname -m` and handles `aarch64` / `arm64` for ARM Linux. Worth copying the detection pattern here — ARM Linux users (ARM servers, Raspberry Pi, certain AWS/cloud instances) otherwise edit the command by hand.

README.md
```diff
@@ -19,15 +19,88 @@ changes!**
...
-   verification.
-   - Install an SMT solver. You can use any solver you want, but the unit
-     tests assume `cvc5` is on your `PATH` [cvc5](https://cvc5.github.io/).
+2. **SMT Solvers**: The verification pipeline and tests require SMT solvers
+   on your `PATH`.
+
+   - **cvc5** (required): [cvc5.github.io](https://cvc5.github.io/)
+   - **z3** (required by some tests): [github.com/Z3Prover/z3](https://github.com/Z3Prover/z3)
+
+3. **Python 3.11+** (required for Python-related tests and the `strata`
```
tautschnig: **Dangling `:` at end of each prerequisite.** Items 3, 4, 5 end with `(required for …):` followed by a blank line and no inline content. Items 1 and 2 include the install instruction after the colon, so the structure reads inconsistently and the reader expects more text.

Either (a) drop the trailing `:` and rely on the "Installing dependencies" section that follows:

```suggestion
3. **Python 3.11+** — required for Python-related tests and the `strata`
   Python tooling.

4. **Java JDK (11 or later)** — required for Java code generation tests.

5. **ion-java jar (1.11.11)** — required for the Java/Ion integration test.
```

or (b) fold the one-line install into each prerequisite (matches how Lean4 and SMT solvers are presented above).

While here: the Java prerequisite currently says "Java JDK" with no version; the "Verifying your setup" block at line 100 checks `javac --version` should be `11+`. Spelling the version floor in the prerequisites list matches that check.

tautschnig (commented):
Documentation-only PR, a strict improvement over the previous README which assumed readers could divine the solver / Python / Java / ion-java dependencies from CI scripts. 

**Small cleanup:**

- **macOS vs Linux version drift.** The README gives cvc5 `1.2.1` for Linux but `1.3.3` for macOS arm64, and z3 `4.15.2` for Linux but `4.16.0` for macOS arm64. No CI job uses those macOS versions, so they're informational only. That's fine, but picking one version and noting "or any newer release" is a cleaner signal than implicit drift that looks like a mistake. Pointers to the GitHub Releases pages already exist in the text, so "match the version your platform has a release for" is an easy framing.

- **`lake test -- --exclude Languages.Python` vs `lake test`.** The "Run all tests" variant says `lake test` unadorned. A reader who hasn't installed `./Tools/Python` will run `lake test` and see confusing failures for the Python suite. Either:
  - Make "Run all tests except Python" the default recommendation and move "Run all tests" under a note saying it requires the Python package.
  - Or add a one-line note: `# Running all tests requires the Python package above.`

- **Java JDK version line.** The verification block has `javac --version   # should be 11+`. The prerequisites section says "Java JDK (required for Java code generation tests)" without specifying a version. Calling out `JDK 11 or later` in the prerequisites list matches what the verification block shows and what the CI runners provide.

**Non-blocking suggestions:**

- The existing README doesn't mention `elan` or the Lean installer explicitly — the installation instructions just link to `lean-lang.org`. That's actually fine since the `lean-toolchain` file pins the version and `lake` auto-installs via elan on first run, but a one-line "Lean / elan will fetch the toolchain pinned in `lean-toolchain` automatically" would save the 'wait, which version do I need?' moment. Worth considering.

- **Arch architecture detection in the wget snippets.** The Linux snippet is hard-coded to `x86_64`. The CI script at `ci.yml:38–49` actually auto-detects via `uname -m` and handles `aarch64` / `arm64` too. Copy-pasting that detection pattern into the README would help ARM Linux users (e.g. Raspberry Pi, ARM servers, certain AWS instances) install without editing the command.

**No test or proof coverage expected** for a documentation-only PR, but the "Verifying your setup" section is itself a lightweight test harness — if it's worth including, it's worth making it the _only_ thing a reader needs to see "working setup" vs "broken setup". Two tiny additions would raise its signal:

- `lake --version` — catches missing-elan / wrong-toolchain situations before the slower `lake build` fires.
- `pip show strata | head -1` (or whatever the final package name ends up) — catches silent `pip install` failures.

Inline comments on the three blocking points below.

README.md
```diff
@@ -19,15 +19,96 @@ changes!**
...
 
-   You can install Lean4 by following the instructions [here](https://lean-lang.org/).
+   Install Lean4 by following the instructions at [lean-lang.org](https://lean-lang.org/).
 
-2. **SMT Solver**: Analysis tools in Strata use SMT solvers for program
-   verification.
-   - Install an SMT solver. You can use any solver you want, but the unit
-     tests assume `cvc5` is on your `PATH` [cvc5](https://cvc5.github.io/).
+2. **SMT Solvers**: The verification pipeline and tests require SMT solvers
+3. **Python 3.11+** — required for Python-related tests and the `strata`
```
atomb: nit: There's a tiny inconsistency between the use of a `-` here (and in later items) and a `:` above (and in earlier items).

atomb (dismissed):
I noted one small formatting nit, but this otherwise looks good. Thanks for doing it!

README.md
PROgram52bc: Ah, thanks for catching that!

README.md
```diff
@@ -19,15 +19,91 @@ changes!**
...
-   verification.
-   - Install an SMT solver. You can use any solver you want, but the unit
-     tests assume `cvc5` is on your `PATH` [cvc5](https://cvc5.github.io/).
+2. **SMT Solvers**: The verification pipeline and tests require SMT solvers
+3. **Python 3.11+**: required for Python-related tests and the `strata`
+   Python tooling.
+
+4. **Java JDK (11 or later)**: required for Java code generation tests.
+
+5. **ion-java jar (1.11.11)**: required for the Java/Ion integration test.
```
tautschnig, line 37: A few structural inconsistencies in this numbered list:

1. **Item 2 is missing a period and any install guidance.** Item 1 says "Install Lean4 by following the instructions at …", items 3–5 explain what each dep is for. Item 2 just trails off with `require SMT solvers` and no pointer to the `#### SMT Solvers` subsection below. Reader doesn't know where to go next.
2. **Missing blank line between item 2 and item 3.** Items 1↔2, 3↔4, 4↔5 are separated by blank lines; 2↔3 is not. On some markdown renderers this changes paragraph handling.
3. **Capitalisation.** After the colon, item 2 starts uppercase ("The verification..."), items 3–5 start lowercase ("required for..."). Pick one.

Concrete suggestion (fixes all three; also points readers at the subsection below and mirrors item 1's "Install X by following..." phrasing):

```suggestion
2. **SMT Solvers**: The verification pipeline and tests require SMT solvers
   (`cvc5` and `z3`). See [Installing dependencies → SMT Solvers](#smt-solvers)
   below.

3. **Python 3.11+**: required for Python-related tests and the `strata`
   Python tooling. See [Installing dependencies → Python](#python) below.

4. **Java JDK (11 or later)**: required for Java code generation tests.
   See [Installing dependencies → Java](#java-for-code-generation-tests) below.

5. **ion-java jar (1.11.11)**: required for the Java/Ion integration test.
   See [Installing dependencies → Java](#java-for-code-generation-tests) below.
```

README.md
```diff
@@ -19,15 +19,91 @@ changes!**
...
+```bash
+# cvc5 static for macOS arm64
+wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-macOS-arm64-static.zip
+unzip cvc5-macOS-arm64-static.zip
+cp cvc5-macOS-arm64-static/bin/cvc5 /usr/local/bin/
+
+# z3 for macOS arm64
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-15.7.3.zip
+unzip z3-4.15.2-arm64-osx-15.7.3.zip
+cp z3-4.15.2-arm64-osx-15.7.3/bin/z3 /usr/local/bin/
```
tautschnig: **Broken URL — this command 404s.** The asset `z3-4.15.2-arm64-osx-15.7.3.zip` does not exist. I verified against the z3 release API:

```
$ curl -sI https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-15.7.3.zip
HTTP/2 404
```

The actual macOS arm64 asset for this release is `z3-4.15.2-arm64-osx-13.7.6.zip` (per `gh api repos/Z3Prover/z3/releases/tags/z3-4.15.2`). Any macOS Apple Silicon developer following this block verbatim today gets a 404. Suggested fix:

```suggestion
# z3 for macOS arm64
wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-13.7.6.zip
unzip z3-4.15.2-arm64-osx-13.7.6.zip
sudo cp z3-4.15.2-arm64-osx-13.7.6/bin/z3 /usr/local/bin/
```

(See the next comment on line 75 for the `sudo` rationale.)

README.md
```diff
@@ -19,15 +19,91 @@ changes!**
...
+# On macOS, /usr/local/bin is typically writable without sudo.
```
tautschnig: **This claim is wrong on a fresh macOS.** `/usr/local/bin` is `root:wheel` / mode 755 by default on macOS; writing into it requires `sudo`. It only becomes writable by your regular user if Homebrew (or equivalent) has explicitly `chown`ed `/usr/local` to your user — and on Apple Silicon, Homebrew installs into `/opt/homebrew/` instead, so `/usr/local/bin` often doesn't even exist there.

In contrast, the `.github/workflows/ci.yml` (lines 178, 300, 306, 311) unconditionally uses `sudo cp … /usr/local/bin/`. I'd recommend this block match CI:

- prefix the two `cp` commands above with `sudo`, and
- drop this misleading one-liner.

If you want to keep a user-local option for parity with the Linux block, offer `~/.local/bin` as an alternative:

```suggestion
# z3 for macOS arm64
wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-13.7.6.zip
unzip z3-4.15.2-arm64-osx-13.7.6.zip
sudo cp z3-4.15.2-arm64-osx-13.7.6/bin/z3 /usr/local/bin/
# Alternative: install into ~/.local/bin (no sudo), then ensure it's on your PATH.
```

README.md
```diff
@@ -19,15 +19,91 @@ changes!**
...
+#### Python
+
+Python 3.11 or later is required. Install the `strata` Python package:
+
+```bash
+pip install ./Tools/Python
+```
+
+This provides the `strata-python` tooling used by the Python verification
+pipeline and tests.
```
tautschnig: Two carry-over items from the previous round that are currently showing as "resolved" but weren't actually applied to the file:

1. **Package name is still wrong.** Line 86 says `strata-python` tooling, but `Tools/Python/pyproject.toml:2` declares `name = "strata"` and `packages = ["strata"]`. After `pip install ./Tools/Python`, the package shows on `pip list` as `strata` (version `0.0.1`), and users `import strata`. There is no `strata-python` on PyPI or installed. (The string `strata_python` does appear in `Tools/Python/strata/gen.py:86` as an argparse `prog` label, but that's an internal label, not the installed package or CLI name.)
2. **`pip install` without venv fails on stock Debian/Ubuntu 23.04+** due to PEP 668 (`externally-managed-environment`), and the project preferences explicitly recommend venv for Python installs. One extra line would make this work on every supported platform.

Suggestion combining both:

```suggestion
Python 3.11 or later is required. Install the `strata` Python package inside a
virtual environment (recommended; avoids PEP 668's `externally-managed-environment`
error on Debian/Ubuntu 23.04+):

```bash
python3 -m venv .venv
source .venv/bin/activate
pip install ./Tools/Python
```

This provides the `strata` Python package (imported as `import strata`) used by
the Python verification pipeline and tests.
```

File: README.md [hunk 9]
```diff
@@ -19,15 +19,91 @@ changes!**
...
+  https://github.com/amazon-ion/ion-java/releases/download/v1.11.11/ion-java-1.11.11.jar
+```
+
+### Verifying your setup
+
+```bash
+cvc5 --version    # should print version info
+z3 --version      # should print version info
+python3 --version # should be 3.11+
+javac --version   # should be 11+
```
tautschnig: Nit: `javac --version # should be 11+` doesn't actually verify — it just prints whatever version is installed and the user has to eyeball the number. If we care about catching a wrong-version Java, a one-liner would fail loudly:

```bash
javac --version | awk '{split($2,a,"."); if (a[1] < 11) { print "Need JDK 11+"; exit 1 }}'
```

Same could apply to `python3 --version`. Otherwise, reword the comments as "check that …" rather than "should be", so the reader knows this is self-service:

```suggestion
```bash
cvc5 --version    # prints version info (any version OK)
z3 --version      # prints version info (any version OK)
python3 --version # check this reports 3.11 or later
javac --version   # check this reports javac 11 or later
```
```

File: README.md [hunk 10]
```diff
@@ -19,15 +19,91 @@ changes!**
...
+wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-macOS-arm64-static.zip
+unzip cvc5-macOS-arm64-static.zip
+cp cvc5-macOS-arm64-static/bin/cvc5 /usr/local/bin/
+
+# z3 for macOS arm64
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-15.7.3.zip
+unzip z3-4.15.2-arm64-osx-15.7.3.zip
+cp z3-4.15.2-arm64-osx-15.7.3/bin/z3 /usr/local/bin/
+# On macOS, /usr/local/bin is typically writable without sudo.
+```
```
tautschnig: **Architecture coverage gap.** The README covers `Linux x86_64` and `macOS Apple Silicon`, but `.github/workflows/ci.yml` (lines 37-49, 53-64) runs on both x86_64 and arm64 Linux, and upstream releases exist for macOS x86_64 (Intel Mac) too. Linux arm64 contributors (ARM servers, Raspberry Pi, some cloud VMs) and Intel-Mac contributors currently have to hand-translate URLs.

Two options — I'd strongly prefer (a):

**(a) Parameterise with `uname -m`** — mirrors what CI already does, so the README stays in sync with actual support:

```bash
# Detect platform
OS="$(uname -s)"      # Linux or Darwin
ARCH="$(uname -m)"    # x86_64, aarch64, arm64
case "$OS/$ARCH" in
  Linux/x86_64)         CVC5_ZIP=cvc5-Linux-x86_64-static   Z3_ZIP=z3-4.15.2-x64-glibc-2.39 ;;
  Linux/aarch64|Linux/arm64) CVC5_ZIP=cvc5-Linux-arm64-static Z3_ZIP=z3-4.15.2-arm64-glibc-2.34 ;;
  Darwin/arm64)         CVC5_ZIP=cvc5-macOS-arm64-static    Z3_ZIP=z3-4.15.2-arm64-osx-13.7.6 ;;
  Darwin/x86_64)        CVC5_ZIP=cvc5-macOS-x86_64-static   Z3_ZIP=z3-4.15.2-x64-osx-13.7.6 ;;
  *) echo "Unsupported: $OS/$ARCH"; exit 1 ;;
esac

wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/${CVC5_ZIP}.zip
unzip ${CVC5_ZIP}.zip
wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/${Z3_ZIP}.zip
unzip ${Z3_ZIP}.zip

# Install (choose one):
mkdir -p ~/.local/bin && cp ${CVC5_ZIP}/bin/cvc5 ${Z3_ZIP}/bin/z3 ~/.local/bin/
# or: sudo cp ${CVC5_ZIP}/bin/cvc5 ${Z3_ZIP}/bin/z3 /usr/local/bin/
```

**(b) Add the two missing OS/arch blocks** explicitly, paralleling the existing two.

(a) has the added benefit that the next time the solver versions bump, there is one line to edit, not four. I verified the four asset names above exist today via the GitHub release APIs.

README.md
```diff
@@ -19,15 +19,91 @@ changes!**
...
-     tests assume `cvc5` is on your `PATH` [cvc5](https://cvc5.github.io/).
+2. **SMT Solvers**: The verification pipeline and tests require SMT solvers
+3. **Python 3.11+**: required for Python-related tests and the `strata`
+   Python tooling.
+
+4. **Java JDK (11 or later)**: required for Java code generation tests.
+
+5. **ion-java jar (1.11.11)**: required for the Java/Ion integration test.
+
+### Installing dependencies
```
tautschnig, line 40: **Keeping this section honest over time.** Every URL and version number in this section is a time-bomb — when cvc5 or z3 cut a new release and retire old artifacts, users hit 404s (as happened for the z3 macOS arm64 URL in this very PR). Two low-cost options to consider:

- A lightweight CI job that runs the documented install commands in a throwaway container (`docker run --rm ubuntu:latest bash -c "$(the README's Linux block)"` ≈ 30 lines of yaml) and fails on non-zero exit. This catches 404s, missing dirs, and PATH issues before users file bugs.
- Or factor the install commands out of the README into a shell script (`scripts/install-deps.sh`) used both by CI and by this README, so the single source of truth is in-repo code, and the README `include`s / references it rather than duplicating.

This isn't blocking on this PR, but worth tracking — otherwise every future solver bump requires a manual README pass that's easy to forget.

README.md
```diff
@@ -45,6 +121,19 @@ To build executable files only and omit proof checks that might take a long time
...
+```bash
+# Run all tests except Python (which requires the Python package)
+lake test -- --exclude Languages.Python
+
+# Run only Python tests (requires `pip install ./Tools/Python`)
+lake test -- Languages.Python
+
+# Run all tests (requires the Python package above)
+lake test
+```
```
tautschnig, line 121: The framing here can confuse new contributors. `lake test` (via `Scripts/StrataTestMain.lean`) only enumerates `StrataTestExtra/` — it does not run the in-tree `#guard_msgs`-style tests that live under `StrataTest/`; those run implicitly during `lake build`.

So "# Run all tests" is slightly misleading: it runs all *extra* tests, and the `#guard_msgs` tests are only "run" (elaborated) as a side-effect of the build. Reader who does `lake test` thinking they get full coverage may miss the distinction.

A small clarification would help:

```suggestion
### Running specific test subsets

Two kinds of tests coexist in this repo:

- **Elaboration-time tests** (`#guard_msgs`) live under `StrataTest/` and run as
  part of `lake build`. No output means they passed.
- **Uncached extra tests** live under `StrataTestExtra/` and run via `lake test`.
  These accept prefix filters:

```bash
# Run all extra tests except Python (which requires the Python package)
lake test -- --exclude Languages.Python

# Run only Python extra tests (requires `pip install ./Tools/Python`)
lake test -- Languages.Python

# Run all extra tests (Python tests will fail without the Python package above)
lake test
```
```

(Also corrected the inner comment on the third form: without the Python package installed, `lake test` will *run* the Python tests and they will fail — it doesn't silently skip them.)

File: README.md [hunk 13]
```diff
@@ -19,15 +19,91 @@ changes!**
...
+
+#### Java (for code generation tests)
+
+A JDK (11+) providing `javac` must be on your `PATH`. Additionally,
+download the ion-java jar used by the Java/Ion integration test:
+
+```bash
+wget -q -O StrataTestExtra/DDM/Integration/Java/testdata/ion-java-1.11.11.jar \
+  https://github.com/amazon-ion/ion-java/releases/download/v1.11.11/ion-java-1.11.11.jar
+```
```
tautschnig, line 77: Minor: `StrataTestExtra/DDM/Integration/Java/regenerate-testdata.sh` already auto-downloads this jar on first run (see the `curl` block at lines 11-15 of that script). If a new contributor intends to regenerate test data, they'll download it twice; if they only want to run the lake test, this wget is the right call. Worth one sentence:

```suggestion
A JDK (11+) providing `javac` must be on your `PATH`. For running the
Java/Ion integration test, download the ion-java jar:

```bash
wget -q -O StrataTestExtra/DDM/Integration/Java/testdata/ion-java-1.11.11.jar \
  https://github.com/amazon-ion/ion-java/releases/download/v1.11.11/ion-java-1.11.11.jar
```

(If you run `StrataTestExtra/DDM/Integration/Java/regenerate-testdata.sh`, it
downloads the jar automatically on first run, and this manual step is
unnecessary.)
```

**Review by tautschnig — COMMENTED** [review]
Refactor note: the three OS/arch install snippets are near-copies of each other and diverge slightly (path, sudo, arch). A table or a single parameterised snippet with `uname -m` detection — mirroring `.github/workflows/ci.yml:37-49` — would be shorter, cover more platforms, and be easier to keep in sync with CI in the future.

File: README.md [hunk 14]
```diff
@@ -19,15 +19,95 @@ changes!**
...
+wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/cvc5-Linux-x86_64-static.zip
+unzip cvc5-Linux-x86_64-static.zip
+mkdir -p ~/.local/bin
+cp cvc5-Linux-x86_64-static/bin/cvc5 ~/.local/bin/
+
+# z3
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-x64-glibc-2.39.zip
+unzip z3-4.15.2-x64-glibc-2.39.zip
+cp z3-4.15.2-x64-glibc-2.39/bin/z3 ~/.local/bin/
+# Ensure ~/.local/bin is on your PATH (most modern distros include it by default).
```
PROgram52bc: remove specific version, and just say use a latest version.

README.md
```diff
@@ -19,15 +19,95 @@ changes!**
...
+
+# z3 for macOS arm64
+# z3 for macOS arm64
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-13.7.6.zip
+unzip z3-4.15.2-arm64-osx-13.7.6.zip
+sudo cp z3-4.15.2-arm64-osx-13.7.6/bin/z3 /usr/local/bin/
+# z3 for macOS arm64
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/z3-4.15.2-arm64-osx-13.7.6.zip
+unzip z3-4.15.2-arm64-osx-13.7.6.zip
+sudo cp z3-4.15.2-arm64-osx-13.7.6/bin/z3 /usr/local/bin/
```
PROgram52bc: same here, remove specific version with placeholder

README.md
PROgram52bc: Agreed. Added comments below to remove specific version numbers.

README.md
PROgram52bc: Will use option (a), but perhaps a shorter bash script. Describe on the high-level that the user should be mindful of the platform of their hardwares.

README.md
```diff
@@ -19,15 +19,88 @@ changes!**
...
+  Darwin/arm64)              CVC5_ZIP=cvc5-macOS-arm64-static    Z3_ZIP=z3-4.15.2-arm64-osx-13.7.6 ;;
+  Darwin/x86_64)             CVC5_ZIP=cvc5-macOS-x86_64-static   Z3_ZIP=z3-4.15.2-x64-osx-13.7.6 ;;
+  *) echo "Unsupported: $OS/$ARCH"; exit 1 ;;
+esac
+
+wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/${CVC5_ZIP}.zip
+unzip ${CVC5_ZIP}.zip
+wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/${Z3_ZIP}.zip
+unzip ${Z3_ZIP}.zip
+
```
tautschnig: **`wget` is not on a stock macOS.** The script uses `wget` twice; macOS ships with `curl` by default but `wget` requires `brew install wget` (or similar). A macOS developer following this README on a vanilla system hits "command not found: wget" on line 64.

Two low-cost fixes:

**(a) Replace `wget` with `curl -L -O`** — works out of the box on both Linux and macOS, same default behaviour (follow redirects, save with the original filename):

```suggestion
wget https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/${CVC5_ZIP}.zip || \
  curl -L -O https://github.com/cvc5/cvc5/releases/download/cvc5-1.2.1/${CVC5_ZIP}.zip
unzip ${CVC5_ZIP}.zip
wget https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/${Z3_ZIP}.zip || \
  curl -L -O https://github.com/Z3Prover/z3/releases/download/z3-4.15.2/${Z3_ZIP}.zip
unzip ${Z3_ZIP}.zip
```

The `|| curl` fallback avoids asking macOS users to install `wget` first while keeping the familiar form for Linux users who already have it.

**(b)** Or simply prefix the block with a note: `# Requires wget (install via 'brew install wget' on macOS).`

(a) is less friction. Worth also considering `cd /tmp` (or creating a scratch directory) before `wget`ing, so users don't end up with `cvc5-Linux-x86_64-static.zip` and `cvc5-Linux-x86_64-static/` clutter in their cwd after the script runs — see the top-level body.

tautschnig (approved):
Two small things worth flagging before merge:

**1. `wget` is not on a stock macOS (genuine friction for macOS users without Homebrew).** The auto-detect script uses `wget` twice. macOS ships with `curl` by default but *not* `wget` — a fresh macOS (especially one without Homebrew, or a CI runner using the vanilla system images) has to install it first. Detail inline.

**2. The script downloads into the user's current working directory** (`wget foo.zip` + `unzip foo.zip`) without any hint to `cd /tmp` first or clean up afterwards. A developer pasting this into their shell in their home directory ends up with two zip files and two extracted subdirectories lying around. Minor, but a one-line nudge in the code fence would save cleanup.

README.md
atomb: A previous commit had `java` here, too. That seems useful to me, though not essential.

atomb (approved):

PR #1124, opened 2026-05-05
Fix a bug in measuring running times in a few places, more fine-grained running time report

aqjune-aws:
Lean sometimes reorders invocation of current time function with other code:

```
let t1 ← IO.monoNanosNow
let res := (some heavyweight work)
let t2 ← IO.monoNanosNow
let elapsedTime := t2 - t1
```
into:
```
let res := (some heavyweight work)
let t1 ← IO.monoNanosNow
let t2 ← IO.monoNanosNow
let elapsedTime := t2 - t1 // This is trivially zero!
```

And this made the SMT encoding time (the ProofObligation object -> SMT Term, Definition, etc) appearing as zero ms, but actually it was taking around 20% of the whole VC dischare process.

Also, this pull request adds much more fine-grained reports on running times: it reports on (1) each Core transform, and (2) each SMT query generation.

```
[profile]     ProcedureInlining: 0ms             
[profile]     FilterProcedures: 0ms           
[profile]     CallElim: 0ms                   
[profile]     PrecondElim: 1ms  
[profile]     FilterProcedures: 0ms       
[profile]     LoopElim: 0ms
[profile]     typeCheck: 2ms
[profile]     symbolicEval: 2ms             
[profile]     ANFEncoder: 0ms  
[profile]   Program transformations: 5ms
[profile]   Build axiom relevance cache: 0ms
[profile]     Preprocess obligations: 3ms
[profile]     SMT encoding: 8ms
[profile]         defineFunTerms: 2ms
[profile]         emitDatatypes: 1ms
[profile]         encodeAssumptions: 0ms
[profile]         encodeAxioms: 0ms
[profile]         encodeFunctions: 1ms
[profile]         encodeObligation: 0ms
[profile]         encodeUFs: 3ms
[profile]         epilog: 3ms
[profile]         prelude: 0ms
[profile]       encodeCore: 24ms
[profile]       runSolver: 0ms
[profile]     Solver/file writing: 26ms
[profile]     Obligations: 20 total, 0 resolved by evaluator
[profile]   VC discharge: 38ms

``` 

### DIFF
--- a/Strata/DL/Imperative/SMTUtils.lean
+++ b/Strata/DL/Imperative/SMTUtils.lean
@@ -10,7 +10,8 @@ import Strata.DL.SMT.DDMTransform.Parse
 import Strata.DL.SMT.DDMTransform.Translate
 import Strata.DDM.Elab
 import Strata.DDM.Format
-import Strata.Util.Profile
+public import Strata.Util.Profile
+import Strata.Util.Statistics
 public import Strata.DL.Imperative.PureExpr
 public import Strata.DL.Imperative.EvalContext
 
@@ -293,6 +294,13 @@ def addLocationInfo {P : PureExpr} [BEq P.Ident]
       Strata.SMT.Solver.setInfoString message.fst message.snd
     | .none => pure ()
 
+/-- Timing keys tracked by `dischargeObligation`. -/
+inductive Timing where
+  | encodeSMT
+  | runSolver
+
+#derive_prefixed_toString Timing "dischargeObligation"
+
 /--
 Writes the proof obligation to file, discharge the obligation using SMT solver,
 and parse the output of the SMT solver.
@@ -300,24 +308,29 @@ and parse the output of the SMT solver.
 When two-sided checking is enabled, the generated SMT file will contain two
 `(check-sat-assuming)` commands, one for `P ∧ Q` and one for `P ∧ ¬Q`,
 and the return value includes both decisions.
+
+Returns a `TimingInfo` map recording nanosecond durations for the encoding
+(`dischargeObligation.encodeSMT (encodeCore)`) and solver execution
+(`dischargeObligation.runSolver`) phases, merged with sub-timings from
+`encodeCore`.
 -/
 def dischargeObligation {P : PureExpr} [ToFormat P.Ident] [BEq P.Ident]
-  (encodeSMT : Strata.SMT.SolverM (List String × Strata.SMT.EncoderState × Std.HashMap String Nat))
+  (encodeSMT : Strata.SMT.SolverM (List String × Strata.SMT.EncoderState × TimingInfo))
   (typedVarToSMTFn : P.Ident → P.Ty → Except Format (String × Strata.SMT.TermType))
   (vars : List P.TypedIdent)
   (smtsolver filename : String)
   (solver_options : Array String) (printFilename : Bool)
   (satisfiabilityCheck validityCheck : Bool)
   (skipSolver : Bool := false) :
-  IO (Except SolverError (Result P.Ident × Result P.Ident × Strata.SMT.EncoderState × Std.HashMap String Nat)) := do
+  IO (Except SolverError (Result P.Ident × Result P.Ident × Strata.SMT.EncoderState × TimingInfo)) := do
   let handle ← IO.FS.Handle.mk filename IO.FS.Mode.write
   let solver ← Strata.SMT.Solver.fileWriter handle
 
-  let timing : Std.HashMap String Nat := {}
+  let timing : TimingInfo := {}
 
   -- encodeSMT (which calls encodeCore) emits check-sat commands internally
   let (((_ids, estate, encTiming), _solverState), timing) ←
-    recordNanos "dischargeObligation.encodeSMT (encodeCore)" timing do
+    recordNanos (toString Timing.encodeSMT) timing do
       encodeSMT.run solver
   -- Merge encodeCore's internal timing into ours
   let timing := encTiming.fold (init := timing) fun acc k v => acc.insert k v
@@ -327,7 +340,7 @@ def dischargeObligation {P : PureExpr} [ToFormat P.Ident] [BEq P.Ident]
   if skipSolver then
     return .ok (.unknown, .unknown, estate, timing)
 
-  let (solver_output, timing) ← recordNanos "dischargeObligation.runSolver" timing do
+  let (solver_output, timing) ← recordNanos (toString Timing.runSolver) timing do
     runSolver smtsolver (#[filename] ++ solver_options)
 
   match ← solverResult typedVarToSMTFn vars solver_output estate smtsolver satisfiabilityCheck validityCheck with
--- a/Strata/DL/Imperative/SMTUtils.lean
+++ b/Strata/DL/Imperative/SMTUtils.lean
@@ -333,7 +333,7 @@ def dischargeObligation {P : PureExpr} [ToFormat P.Ident] [BEq P.Ident]
     recordNanos (toString Timing.encodeSMT) timing do
       encodeSMT.run solver
   -- Merge encodeCore's internal timing into ours
-  let timing := encTiming.fold (init := timing) fun acc k v => acc.insert k v
+  let timing := TimingInfo.mergeAdd timing encTiming
 
   if printFilename then IO.println s!"Wrote problem to {filename}."
 
--- a/Strata/DL/Imperative/SMTUtils.lean
+++ b/Strata/DL/Imperative/SMTUtils.lean
@@ -10,6 +10,7 @@ import Strata.DL.SMT.DDMTransform.Parse
 import Strata.DL.SMT.DDMTransform.Translate
 import Strata.DDM.Elab
 import Strata.DDM.Format
+import Strata.Util.Profile
 public import Strata.DL.Imperative.PureExpr
 public import Strata.DL.Imperative.EvalContext
 
@@ -301,29 +302,37 @@ When two-sided checking is enabled, the generated SMT file will contain two
 and the return value includes both decisions.
 -/
 def dischargeObligation {P : PureExpr} [ToFormat P.Ident] [BEq P.Ident]
-  (encodeSMT : Strata.SMT.SolverM (List String × Strata.SMT.EncoderState))
+  (encodeSMT : Strata.SMT.SolverM (List String × Strata.SMT.EncoderState × Std.HashMap String Nat))
   (typedVarToSMTFn : P.Ident → P.Ty → Except Format (String × Strata.SMT.TermType))
   (vars : List P.TypedIdent)
   (smtsolver filename : String)
   (solver_options : Array String) (printFilename : Bool)
   (satisfiabilityCheck validityCheck : Bool)
   (skipSolver : Bool := false) :
-  IO (Except SolverError (Result P.Ident × Result P.Ident × Strata.SMT.EncoderState)) := do
+  IO (Except SolverError (Result P.Ident × Result P.Ident × Strata.SMT.EncoderState × Std.HashMap String Nat)) := do
   let handle ← IO.FS.Handle.mk filename IO.FS.Mode.write
   let solver ← Strata.SMT.Solver.fileWriter handle
 
+  let timing : Std.HashMap String Nat := {}
+
   -- encodeSMT (which calls encodeCore) emits check-sat commands internally
-  let ((_ids, estate), _solverState) ← encodeSMT.run solver
+  let (((_ids, estate, encTiming), _solverState), timing) ←
+    recordNanos "dischargeObligation.encodeSMT (encodeCore)" timing do
+      encodeSMT.run solver
+  -- Merge encodeCore's internal timing into ours
+  let timing := encTiming.fold (init := timing) fun acc k v => acc.insert k v
 
   if printFilename then IO.println s!"Wrote problem to {filename}."
 
   if skipSolver then
-    return .ok (.unknown, .unknown, estate)
+    return .ok (.unknown, .unknown, estate, timing)
 
-  let solver_output ← runSolver smtsolver (#[filename] ++ solver_options)
+  let (solver_output, timing) ← recordNanos "dischargeObligation.runSolver" timing do
+    runSolver smtsolver (#[filename] ++ solver_options)
+
   match ← solverResult typedVarToSMTFn vars solver_output estate smtsolver satisfiabilityCheck validityCheck with
   | .error e => return .error e
-  | .ok (satResult, validityResult) => return .ok (satResult, validityResult, estate)
+  | .ok (satResult, validityResult) => return .ok (satResult, validityResult, estate, timing)
 
 ---------------------------------------------------------------------
 end -- public section
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -1104,7 +1104,7 @@ def verifySingleEnv (oblProgram : Program)
     let t0 ← IO.monoNanosNow
     let (obligation, peSatResult?, peValResult?) ← preprocessObligation obligation p options satisfiabilityCheck validityCheck axiomCache axiomNames axiomProgram
     let t1 ← IO.monoNanosNow
-    timingNs := timingNs.insert (toString Verifier.Timing.preprocessObligations) (timingNs.getD (toString Verifier.Timing.preprocessObligations) 0 + (t1 - t0))
+    timingNs := timingNs.add (toString Verifier.Timing.preprocessObligations) (t1 - t0)
 
     -- If evaluator resolved both checks, we're done, unless we always want to generate SMT queries
     if not options.alwaysGenerateSMT then
@@ -1133,7 +1133,7 @@ def verifySingleEnv (oblProgram : Program)
     let needValCheck := validityCheck && peValResult?.isNone
     let (maybeTerms, encNs) ← measureNanos fun () =>
       ProofObligation.toSMTTerms E obligation { SMT.Context.default with uniqueBoundNames := options.uniqueBoundNames } options.useArrayTheory
-    timingNs := timingNs.insert (toString Verifier.Timing.smtEncoding) (timingNs.getD (toString Verifier.Timing.smtEncoding) 0 + encNs)
+    timingNs := timingNs.add (toString Verifier.Timing.smtEncoding) encNs
 
     match maybeTerms with
     | .error err =>
@@ -1155,10 +1155,10 @@ def verifySingleEnv (oblProgram : Program)
                     counter tempDir needSatCheck needValCheck (externalPhases ++ corePhases)
                     (varDefinitions := varDefs) (varDeclarations := varDecls)
       let t5 ← IO.monoNanosNow
-      timingNs := timingNs.insert (toString Verifier.Timing.solverFileWriting) (timingNs.getD (toString Verifier.Timing.solverFileWriting) 0 + (t5 - t4))
+      timingNs := timingNs.add (toString Verifier.Timing.solverFileWriting) (t5 - t4)
 
       for (key, val) in timing do
-        timingNs := timingNs.insert key (timingNs.getD key 0 + val)
+        timingNs := timingNs.add key val
 
       -- Merge evaluator results with solver results
       let result := match result.outcome with
@@ -1220,6 +1220,8 @@ def verify (program : Program)
     let mut current := program
     let mut state : Transform.CoreTransformState := { Transform.CoreTransformState.emp with factory := some factory }
     let mut step := 0
+    -- Required by `measureNanos`'s `[Inhabited α]` constraint; the default value
+    -- is never observed (it only initializes an IO.Ref that is overwritten before read).
     have : Inhabited (Except Transform.Err Program × Transform.CoreTransformState) :=
       ⟨(.error default, Transform.CoreTransformState.emp)⟩
     for pp in pipelinePhases do
--- a/Strata/Languages/Core/Verifier.lean
+++ b/Strata/Languages/Core/Verifier.lean
@@ -32,6 +32,20 @@ open Strata
 
 public section
 
+/-- Timing keys tracked by `encodeCore`. -/
+inductive Timing where
+  | prelude
+  | emitDatatypes
+  | encodeUFs
+  | encodeFunctions
+  | encodeAxioms
+  | defineFunTerms
+  | encodeAssumptions
+  | encodeObligation
+  | epilog
+
+#derive_prefixed_toString Timing "encodeCore"
+
 /-- Encode a verification condition into SMT-LIB format.
 
 This function encodes the path conditions (P) and obligation (Q) into SMT,
@@ -55,15 +69,15 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
     (label : String)
     (varDefinitions : List Core.VarDefinition := [])
     (varDeclarations : List Core.VarDeclaration := []) :
-    SolverM (List String × EncoderState × Std.HashMap String Nat) := do
+    SolverM (List String × EncoderState × TimingInfo) := do
   Solver.setLogic "ALL"
 
-  let timing : Std.HashMap String Nat := {}
+  let timing : TimingInfo := {}
 
-  let ((), timing) ← recordNanos "encodeCore.prelude" timing do
+  let ((), timing) ← recordNanos (toString Timing.prelude) timing do
     prelude
 
-  let ((), timing) ← recordNanos "encodeCore.emitDatatypes" timing do
+  let ((), timing) ← recordNanos (toString Timing.emitDatatypes) timing do
     let _ ← ctx.sorts.mapM (fun s => Solver.declareSort s.name s.arity)
     ctx.emitDatatypes
 
@@ -71,14 +85,14 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
   let varDeclNames := varDeclarations.map (·.name)
   let managedNames := varDefNames ++ varDeclNames
 
-  let (estate, timing) ← recordNanos "encodeCore.encodeUFs" timing do
+  let (estate, timing) ← recordNanos (toString Timing.encodeUFs) timing do
     -- Filter out managed variables from UF declarations (they will be emitted separately)
     let ufsToDecl := if managedNames.isEmpty then ctx.ufs
       else ctx.ufs.filter fun uf => !managedNames.contains uf.id
     let (_ufs, estate) ← ufsToDecl.mapM (fun uf => encodeUF uf) |>.run EncoderState.init
     pure estate
 
-  let (estate, timing) ← recordNanos "encodeCore.encodeFunctions" timing do
+  let (estate, timing) ← recordNanos (toString Timing.encodeFunctions) timing do
     -- Pre-populate encoder state with managed variable names so encodeTerm
     -- recognizes them without emitting declare-fun
     let estate := if managedNames.isEmpty then estate
@@ -89,7 +103,7 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
     let (_ifs, estate) ← ctx.ifs.mapM (fun fn => encodeFunction fn.uf fn.body) |>.run estate
     pure estate
 
-  let ((_axms, estate), timing) ← recordNanos "encodeCore.encodeAxioms" timing do
+  let ((_axms, estate), timing) ← recordNanos (toString Timing.encodeAxioms) timing do
     ctx.axms.mapM (fun ax => encodeTerm ax) |>.run estate
 
   for id in _axms do
@@ -98,25 +112,25 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
   for decl in varDeclarations do
     Solver.declareFun decl.name [] decl.ty
 
-  let (estate, timing) ← recordNanos "encodeCore.defineFunTerms" timing do
+  let (estate, timing) ← recordNanos (toString Timing.defineFunTerms) timing do
     -- Emit variable definitions as define-fun (macro expansions, not constraints)
     varDefinitions.foldlM (init := estate) fun estate def_ => do
       let (bodyEnc, estate) ← (encodeTerm def_.body) |>.run estate
       Solver.defineFunTerm def_.name [] def_.ty bodyEnc
       pure estate
 
-  let ((assumptionIds, estate), timing) ← recordNanos "encodeCore.encodeAssumptions" timing do
+  let ((assumptionIds, estate), timing) ← recordNanos (toString Timing.encodeAssumptions) timing do
     -- Assert assumption terms
     assumptionTerms.mapM (encodeTerm) |>.run estate
 
   for id in assumptionIds do
     Solver.assert id
 
-  let ((obligationId, estate), timing) ← recordNanos "encodeCore.encodeObligation" timing do
+  let ((obligationId, estate), timing) ← recordNanos (toString Timing.encodeObligation) timing do
     -- Encode the obligation term Q (not negated)
     (encodeTerm obligationTerm) |>.run estate
 
-  let (ids, timing) ← recordNanos "encodeCore.epilog" timing do
+  let (ids, timing) ← recordNanos (toString Timing.epilog) timing do
     let ids := estate.ufs.toList.filterMap fun (uf, id) =>
       if uf.args.isEmpty && !managedNames.contains uf.id then some id else none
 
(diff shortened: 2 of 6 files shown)


Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -301,29 +302,37 @@ When two-sided checking is enabled, the generated SMT file will contain two
 and the return value includes both decisions.
```
joehendrix, line 310: Could you update docstring to discuss HashMap addition (maybe use an alias as well)?

Strata/Util/Profile.lean
```diff
@@ -5,8 +5,30 @@
...
+    Takes a thunk to ensure the expression is not evaluated before timing starts.
+    The `t1 == 0` trick prevents hoisting `result` above t1.
+    Writing `result` to an `IO.Ref` forces its evaluation before t2 is read. -/
+public def measureNanos [Inhabited α] (expr : Unit → α) : BaseIO (α × Nat) := do
+  let ref ← IO.mkRef (default : α)
+  let t1 ← IO.monoNanosNow
+  let result := if t1 == 0 then default else expr ()
+  ref.set result
+  let t2 ← IO.monoNanosNow
+  pure (result, t2 - t1)
```
joehendrix, line 43: This looks really hacky.  How about something annotated with noinline?

```
/-- Measure the wall-clock nanoseconds taken by an IO action.
    Writing `result` to an `IO.Ref` forces its evaluation before t2 is read. -/
@[noinline] def measureNanosM (act : BaseIO α) : BaseIO (α × Nat) := do
  let t1 ← IO.monoNanosNow
  let result ← act
  let t2 ← IO.monoNanosNow
  pure (result, t2 - t1)
```

Then `measureNanos` would become:

```
/-- Measure the wall-clock nanoseconds taken by a pure expression.
    Takes a thunk to ensure the expression is not evaluated before timing starts.
    The `t1 == 0` trick prevents hoisting `result` above t1.
    Writing `result` to an `IO.Ref` forces its evaluation before t2 is read. -/
@[noinline]
public def measureNanos (expr : Unit → α) : BaseIO (α × Nat) :=
  measureNanosM (pure (expr ())
```

The second noinline is necessary so that the expr doesn't get pre-evaluated.

Strata/Util/Profile.lean
```diff
@@ -5,8 +5,30 @@
...
+
+/-- Run an action and return its result along with the elapsed nanoseconds. -/
+@[inline] public def recordNanos {m α} [Monad m] [MonadLiftT BaseIO m]
+    (key : String) (timing : Std.HashMap String Nat) (action : m α) : m (α × Std.HashMap String Nat) := do
```
joehendrix: I think using an abbreviation with a name (`NanoTimings`?) instead of a hashmap directly would clarify code.

joehendrix (changes requested):

Strata/Util/Profile.lean
aqjune-aws: Sounds great, I named it as `TimingInfo`. Like the collected stats do, I added simple inductive data structure that represents the items and created `ToString` instances using existing `#derive_prefixed_toString`.

Strata/Util/Profile.lean
aqjune-aws: It seems the version still has the problem with measuring running times. `lake exe strata verify --profile Examples/HeapReasoning.core.st` prints 0ms on the Core transformations.
I tried another version of `measureNanos` which uses `@[noinline]` but also relies on `IO.mkRef`. I think this looks better than the previous one because there is an order between `IO.monoNanosNow` and `ref.set`.

aqjune-aws:
Diffs on Github will look better if ignoring whitespace changes is toggled on.

Strata/Util/Profile.lean
```diff
@@ -5,8 +5,33 @@
...
+  pure (result, t2 - t1)
+
+/-- Run an action and record its elapsed nanoseconds into a `TimingInfo` under the given key. -/
+@[inline] public def recordNanos {m α} [Monad m] [MonadLiftT BaseIO m]
+    (key : String) (timing : TimingInfo) (action : m α) : m (α × TimingInfo) := do
+  let t0 ← IO.monoNanosNow
+  let result ← action
+  let t1 ← IO.monoNanosNow
+  pure (result, timing.insert key (t1 - t0))
+
```
MikaelMayer, line 63: 🤖 `recordNanos` is marked `@[inline]` and measures a monadic `action` using the same `t0/t1` bracketing pattern that the PR description says Lean can reorder for pure code. Since `action` here is already monadic (sequenced by `>>=`), the reordering issue shouldn't apply — but the `@[inline]` means the compiler *could* inline this and then reorder the timestamp reads relative to the inlined body in theory. Is there a reason this isn't also `@[noinline]` like `measureNanos`? If the monadic bind is sufficient to prevent reordering, a brief comment explaining why would help future readers understand the asymmetry.

Strata/Languages/Core/Verifier.lean
```diff
@@ -1054,7 +1104,8 @@ def verifySingleEnv (oblProgram : Program)
     let t0 ← IO.monoNanosNow
     let (obligation, peSatResult?, peValResult?) ← preprocessObligation obligation p options satisfiabilityCheck validityCheck axiomCache axiomNames axiomProgram
     let t1 ← IO.monoNanosNow
-    preprocessNs := preprocessNs + (t1 - t0)
+    timingNs := timingNs.insert (toString Verifier.Timing.preprocessObligations) (timingNs.getD (toString Verifier.Timing.preprocessObligations) 0 + (t1 - t0))
+
     -- If evaluator resolved both checks, we're done, unless we always want to generate SMT queries
     if not options.alwaysGenerateSMT then
       if let (some peSat, some peVal) := (peSatResult?, peValResult?) then
```
MikaelMayer, line 1111: 🤖 The pattern `timingNs.insert key (timingNs.getD key 0 + val)` is repeated 4 times in this function. Consider a small helper like:
```lean
def TimingInfo.add (t : TimingInfo) (key : String) (ns : Nat) : TimingInfo :=
  t.insert key (t.getD key 0 + ns)
```
This would reduce noise and make the accumulation intent clearer.

MikaelMayer (commented):
🤖 Good PR — the reordering bug is real and the fix via `measureNanos` with `@[noinline]` + `IO.Ref` is sound. The fine-grained timing breakdown is useful.

A few observations below.

Strata/Languages/Core/Verifier.lean
```diff
@@ -1155,10 +1220,15 @@ def verify (program : Program)
     let mut current := program
     let mut state : Transform.CoreTransformState := { Transform.CoreTransformState.emp with factory := some factory }
     let mut step := 0
+    have : Inhabited (Except Transform.Err Program × Transform.CoreTransformState) :=
+      ⟨(.error default, Transform.CoreTransformState.emp)⟩
```
MikaelMayer, line 1224: 🤖 This `have : Inhabited ...` instance is needed because `measureNanos` requires `[Inhabited α]`. Since `Transform.runWith` is presumably a pure function being timed, this makes sense — but the `Inhabited` instance uses `.error default` which means if the `IO.Ref` default value somehow leaks through, you'd get a silent error result rather than a crash. Worth a brief comment explaining this is only needed to satisfy `measureNanos`'s constraint and the default is never observed.

Strata/Languages/Core/Verifier.lean
```diff
@@ -1006,8 +1046,19 @@ def getObligationResult (assumptionTerms : List Term) (obligationTerm : Term)
...
+    return (result, timing)
+
+namespace Verifier
 
+/-- Timing keys tracked by `verifySingleEnv`. -/
+inductive Timing where
+  | preprocessObligations
+  | smtEncoding
+  | solverFileWriting
+
```
MikaelMayer, line 1058: 🤖 Nit: The key `"verifySingleEnv.solverFileWriting"` is misleading — it measures the entire `getObligationResult` call which includes both SMT file writing *and* solver execution. The sub-timings (`dischargeObligation.encodeSMT`, `dischargeObligation.runSolver`) break it down correctly, but the parent key name suggests only file I/O. Consider renaming to something like `solverTotal` or `dischargeTotal`.

Strata/Util/Profile.lean
aqjune: Good point, removed `@[noline]`.

Strata/Languages/Core/Verifier.lean
aqjune: This is intentional since once switched to FFI there won't be file writing :)

Strata/Languages/Core/Verifier.lean
aqjune: Fixed, thanks

Strata/Languages/Core/Verifier.lean
aqjune: Add a comment

MikaelMayer (commented):
🔍 LGTM — all comments addressed. The `TimingInfo.add` helper, the `Inhabited` comment, and the `@[inline]` removal from `recordNanos` all look good. The `solverFileWriting` naming nit was minor and I'm fine leaving it as-is since the sub-timings break it down correctly.

MikaelMayer (commented):
🔍 LGTM — all comments addressed. The `TimingInfo.add` helper is used in all 4 spots, `@[inline]` removed from `recordNanos`, and the `Inhabited` comment clarifies intent.

Strata/Util/Profile.lean
```diff
@@ -5,8 +5,37 @@
...
+    The `@[noinline]` prevents the compiler from hoisting `expr ()` out of
+    this function. The `IO.Ref.set` forces evaluation of `expr ()` between
+    the two timestamp reads (IO actions are sequenced). -/
+@[noinline] public def measureNanos [Inhabited α] (expr : Unit → α) : BaseIO (α × Nat) := do
+  let ref ← IO.mkRef (default : α)
+  let t1 ← IO.monoNanosNow
+  ref.set (expr ())
+  let t2 ← IO.monoNanosNow
+  let result ← ref.get
+  pure (result, t2 - t1)
```
tautschnig, line 43: The docstring could be more precise about what evaluation is actually forced. `IO.Ref.set` stores whatever value is passed to it; Lean's evaluation strategy will reduce `expr ()` to WHNF before the `set` call runs, but anything lazy *inside* the returned value (e.g., unevaluated thunks in a `List`, an `Array` of unforced elements, a lazy closure field) will not be touched by this function. For simple types like `Nat` or small ADTs in the leaf position this is fine, but callers measuring e.g. the cost of building a large structure need to either ensure the structure is strict or arrange deep evaluation themselves. Worth a one-line caveat:

```suggestion
/-- Measure the wall-clock nanoseconds taken by a pure expression.
    The `@[noinline]` prevents the compiler from hoisting `expr ()` out of
    this function. The `IO.Ref.set` forces evaluation of `expr ()` between
    the two timestamp reads (IO actions are sequenced).

    Note: only WHNF of `expr ()` is forced. Lazy sub-structures of the returned
    value are not evaluated and their cost is not counted. If you need to time
    deep evaluation, make the returned value strict (or force it yourself before
    the second timestamp read). -/
@[noinline] public def measureNanos [Inhabited α] (expr : Unit → α) : BaseIO (α × Nat) := do
  let ref ← IO.mkRef (default : α)
  let t1 ← IO.monoNanosNow
  ref.set (expr ())
  let t2 ← IO.monoNanosNow
  let result ← ref.get
  pure (result, t2 - t1)
```

Strata/Util/Profile.lean
```diff
@@ -5,8 +5,37 @@
...
+
+/-- Run an action and record its elapsed nanoseconds into a `TimingInfo` under the given key. -/
+public def recordNanos {m α} [Monad m] [MonadLiftT BaseIO m]
+    (key : String) (timing : TimingInfo) (action : m α) : m (α × TimingInfo) := do
+  let t0 ← IO.monoNanosNow
+  let result ← action
+  let t1 ← IO.monoNanosNow
+  pure (result, timing.insert key (t1 - t0))
```
tautschnig: Two concerns here, both about the contract:

1. **Reordering safety comment still owed.** @MikaelMayer asked in the previous round why this one doesn't need `@[noinline]` while `measureNanos` does. The annotation was removed but the explanation wasn't added. A short comment would help future readers understand the asymmetry — the answer, I believe, is that because `action : m α` is a monadic bind, the IO primitives it wraps are sequenced by `>>=`, so reordering cannot move `IO.monoNanosNow` across `action`. But that's a non-obvious property of the compiler.

2. **Latent pure-thunk hazard, symmetric with the `measureNanos` bug.** If a caller writes
   ```lean
   recordNanos "foo" t (pure expensivePureComputation)
   ```
   then `expensivePureComputation` is a pure thunk inside `pure`; the bind sequences the *IO* part but doesn't force the thunk. The recorded time will be near-zero even though the actual work hasn't happened yet. All current call sites are fine (they wrap IO-heavy work), but this is exactly the trap this PR exists to close elsewhere, so documenting it — or, better, forcing via `IO.Ref` with `[Inhabited α]` like `measureNanos` does — would prevent future regressions.

Minor: `recordNanos` uses `timing.insert` (overwrite), but the parallel merge loop in `verifySingleEnv` uses `TimingInfo.add` (accumulate). If the same key ever gets measured twice in the same `timing` scope (currently not the case, but easy to introduce), `recordNanos` would silently throw away the earlier reading. Prefer `TimingInfo.add` here too; for a first-write `.getD k 0 = 0`, so semantics are preserved, and repeated-key usage becomes safe.

```suggestion
/-- Run an action and record its elapsed nanoseconds into a `TimingInfo` under
    the given key. The measurement is accumulated via `TimingInfo.add`, so
    repeated calls with the same key sum their durations.

    Unlike `measureNanos`, this does not need `@[noinline]`: `action` is a
    monadic computation, so its IO primitives are sequenced by `>>=` and cannot
    be reordered across `IO.monoNanosNow`.

    Caveat: if the `α` returned by `action` is a lazy pure value (e.g.
    `pure (expensivePureComputation)`), the thunk is not forced here, and the
    recorded time will reflect only the IO sequencing, not the pure work. For
    timing pure computations, use `measureNanos` (which forces via `IO.Ref`). -/
public def recordNanos {m α} [Monad m] [MonadLiftT BaseIO m]
    (key : String) (timing : TimingInfo) (action : m α) : m (α × TimingInfo) := do
  let t0 ← IO.monoNanosNow
  let result ← action
  let t1 ← IO.monoNanosNow
  pure (result, timing.add key (t1 - t0))
```

Strata/Util/Profile.lean
```diff
@@ -5,8 +5,37 @@
...
 module
 
+public import Std.Data.HashMap.Basic
+
+/-- A map from timing label to elapsed nanoseconds. -/
+public abbrev TimingInfo := Std.HashMap String Nat
+
+/-- Accumulate nanoseconds into an existing key (defaulting to 0). -/
+@[inline] public def TimingInfo.add (t : TimingInfo) (key : String) (ns : Nat) : TimingInfo :=
+  t.insert key (t.getD key 0 + ns)
```
tautschnig, line 16: Two adjacent requests:

**(a) Add a `mergeAdd` helper.** The codebase now has two places merging sub-timings into a parent `TimingInfo`, using different shapes:

- `Strata/DL/Imperative/SMTUtils.lean:336` — `encTiming.fold (init := timing) fun acc k v => acc.insert k v` (overwrites on collision)
- `Strata/Languages/Core/Verifier.lean:1160` — `for (key, val) in timing do timingNs := timingNs.add key val` (accumulates on collision)

These two behaviours are *not* equivalent (the first loses information when a key repeats across obligations). The `Verifier.lean` version is the semantically correct one for an accumulator spanning multiple calls. A shared helper would make the intent explicit and eliminate the asymmetry:

```lean
/-- Merge `b` into `a` by summing values for overlapping keys. -/
@[inline] public def TimingInfo.mergeAdd (a b : TimingInfo) : TimingInfo :=
  b.fold (init := a) fun acc k v => acc.add k v
```

**(b) Pin down `add`'s semantics with a couple of tiny theorems.** Since `TimingInfo.add` is load-bearing for every per-phase total reported to the user, it's worth stating — and, ideally, proving — its key lemmas. Rough shape (names and exact `Std.HashMap` lemma references will need adjustment):

```lean
@[simp] theorem TimingInfo.add_getD_self (t : TimingInfo) (k : String) (v : Nat) :
    (t.add k v).getD k 0 = t.getD k 0 + v := by
  simp [TimingInfo.add, Std.HashMap.getD_insert_self]

@[simp] theorem TimingInfo.add_getD_of_ne (t : TimingInfo) (k k' : String) (v : Nat)
    (h : k ≠ k') : (t.add k v).getD k' 0 = t.getD k' 0 := by
  simp [TimingInfo.add, Std.HashMap.getD_insert_of_ne h]

theorem TimingInfo.mergeAdd_getD (a b : TimingInfo) (k : String) :
    (TimingInfo.mergeAdd a b).getD k 0 = a.getD k 0 + b.getD k 0 := by
  -- induction on the fold; uses the two lemmas above
  sorry
```

Even just the first two `@[simp]` lemmas would catch future typos in a rewrite of `add` and are a natural place to put `#guard`-style sanity tests in `StrataTest/Util/`.

Strata/Util/Profile.lean
tautschnig, line 43: **Missing regression test for the PR's headline fix.** The whole point of this PR is that the previous naïve pattern silently reported 0 ms for heavyweight pure work. Nothing in `StrataTest/` would catch if a future edit to `measureNanos` (e.g. dropping the `@[noinline]` or the `IO.Ref` indirection) reintroduced the bug.

Proposed test in a new `StrataTest/Util/Profile.lean`:

```lean
import Strata.Util.Profile

/-- Regression: `measureNanos` must actually observe time for CPU-bound pure
    work. If this reports 0 ns, the `@[noinline]` / `IO.Ref.set` trick has
    regressed and timing will silently be lost again. -/
def measureNanos_nonzero : BaseIO Bool := do
  let (_, ns) ← measureNanos fun () =>
    -- Large enough to comfortably exceed monoNanosNow resolution on any host.
    (List.range 200000).foldl (· + ·) 0
  pure (ns > 0)

/-- `#guard` an `IO Bool` by running it. -/
#eval show IO Unit from do
  unless (← measureNanos_nonzero.toIO) do
    throw <| IO.userError "measureNanos regressed to 0 ns"
```

And a small guard-level test for `TimingInfo.add`:

```lean
#guard
  let t := ({} : TimingInfo).add "a" 5 |>.add "b" 7 |>.add "a" 3
  t.getD "a" 0 == 8 && t.getD "b" 0 == 7 && t.getD "missing" 0 == 0
```

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -292,38 +294,58 @@ def addLocationInfo {P : PureExpr} [BEq P.Ident]
...
 
+  let timing : TimingInfo := {}
+
   -- encodeSMT (which calls encodeCore) emits check-sat commands internally
-  let ((_ids, estate), _solverState) ← encodeSMT.run solver
+  let (((_ids, estate, encTiming), _solverState), timing) ←
+    recordNanos (toString Timing.encodeSMT) timing do
+      encodeSMT.run solver
+  -- Merge encodeCore's internal timing into ours
+  let timing := encTiming.fold (init := timing) fun acc k v => acc.insert k v
```
tautschnig: If the `TimingInfo.mergeAdd` helper I suggested on `Profile.lean` lands, this becomes:

```suggestion
  -- Merge encodeCore's internal timing into ours
  let timing := TimingInfo.mergeAdd timing encTiming
```

Semantic note: using `insert` here (overwrite) happens to be OK *today* because each `encodeCore` key is set exactly once per call and the outer `timing` doesn't yet contain any `encodeCore.*` key. But it's the wrong operation for a generic merge — it silently loses the parent value on any key collision — and `mergeAdd` would stay correct if the invariant ever changes.

Strata/Languages/Core/Verifier.lean
```diff
@@ -55,89 +69,114 @@ def encodeCore (ctx : Core.SMT.Context) (prelude : SolverM Unit)
...
+    pure estate
+
+  let ((_axms, estate), timing) ← recordNanos (toString Timing.encodeAxioms) timing do
+    ctx.axms.mapM (fun ax => encodeTerm ax) |>.run estate
+
   for id in _axms do
     Solver.assert id
   -- Emit variable declarations as declare-fun
   for decl in varDeclarations do
     Solver.declareFun decl.name [] decl.ty
```
tautschnig, line 113: These two `for` loops (the `Solver.assert id` pass for axioms and the `Solver.declareFun` pass for variable declarations) aren't wrapped in any `recordNanos` block, so their cost isn't attributed to any sub-timing. On `HeapReasoning.core.st` I see `encodeCore: 267ms` but the sum of its sub-timings is ~199ms — ~67ms is unattributed. Either:

- wrap these two loops in `recordNanos` (e.g., fold them into `encodeAxioms` / a new `declareVars` bucket), or
- add a comment explaining that `encodeCore` total intentionally includes unattributed overhead and the sub-timings are not expected to sum.

I'd lean toward the former for consistency — otherwise the indented output makes users mentally attempt the sum and come up short.

Strata/Languages/Core/Verifier.lean
tautschnig, line 1058: Revisiting the earlier thread on this name: I understand the forward-looking argument (post-FFI there is no file writing), but the current code makes this key cover a `t4…t5` window around `getObligationResult`, which encompasses SMT encoding *and* solver execution *and* file I/O. A reader of the profile output today sees `Solver/file writing: 1605ms` next to `runSolver: 1313ms` and reasonably wonders how file writing alone could take ~300ms.

Concrete suggestion: pick a name that is already accurate today *and* stays accurate post-FFI, e.g. `dischargeTotal` or `obligationTotal`. The print label can be changed independently later. This is a small rename; doesn't need to block the PR, but leaving it makes the output confusing in the meantime.

Strata/Languages/Core/Verifier.lean
```diff
@@ -1100,11 +1151,15 @@ def verifySingleEnv (oblProgram : Program)
...
-      let result ← getObligationResult assumptionTerms obligationTerm ctx obligation p options
+      let (result, timing) ← getObligationResult assumptionTerms obligationTerm ctx obligation p options
                     counter tempDir needSatCheck needValCheck (externalPhases ++ corePhases)
                     (varDefinitions := varDefs) (varDeclarations := varDecls)
       let t5 ← IO.monoNanosNow
-      solverNs := solverNs + (t5 - t4)
+      timingNs := timingNs.add (toString Verifier.Timing.solverFileWriting) (t5 - t4)
+
+      for (key, val) in timing do
+        timingNs := timingNs.add key val
```
tautschnig: With a `TimingInfo.mergeAdd` helper (see comment on `Profile.lean`), this becomes a one-liner and matches the pattern used in `dischargeObligation`:

```suggestion
      timingNs := timingNs.add (toString Verifier.Timing.solverFileWriting) (t5 - t4)
      timingNs := TimingInfo.mergeAdd timingNs timing
```

Strata/Languages/Core/Verifier.lean
```diff
@@ -1120,10 +1175,20 @@ def verifySingleEnv (oblProgram : Program)
...
+    let encodeCoreEntries := timingNs.toList.filter (·.1.startsWith "encodeCore.")
+      |>.mergeSort (fun a b => a.1 < b.1)
+    for (key, val) in encodeCoreEntries do
+      let suffix := key.drop "encodeCore.".length
+      let _ ← (IO.println s!"[profile]         {suffix}: {nsToMs val}ms" |>.toBaseIO)
+    -- Print dischargeObligation-level timings (encodeCore total, runSolver)
+    let _ ← (IO.println s!"[profile]       encodeCore: {nsToMs (timingNs.getD (toString Imperative.SMT.Timing.encodeSMT) 0)}ms" |>.toBaseIO)
+    let _ ← (IO.println s!"[profile]       runSolver: {nsToMs (timingNs.getD (toString Imperative.SMT.Timing.runSolver) 0)}ms" |>.toBaseIO)
+    let _ ← (IO.println s!"[profile]     Solver/file writing: {nsToMs (timingNs.getD (toString Verifier.Timing.solverFileWriting) 0)}ms" |>.toBaseIO)
     let _ ← (IO.println s!"[profile]     Obligations: {obligations.size} total, {peResolvedCount} resolved by evaluator" |>.toBaseIO)
```
tautschnig, line 1190: The printed hierarchy is confusing. On `HeapReasoning.core.st` this produces:

```
[profile]     SMT encoding: 300ms
[profile]         defineFunTerms: 152ms
[profile]         emitDatatypes: 0ms
[profile]         encodeAssumptions: 2ms
...
[profile]       encodeCore: 267ms
[profile]       runSolver: 1313ms
[profile]     Solver/file writing: 1605ms
```

The `encodeCore.*` sub-entries appear under — and indented deeper than — `SMT encoding`, but they actually belong to `encodeCore` (which belongs to `Solver/file writing`). `SMT encoding` (= `ProofObligation.toSMTTerms`, building `Term` objects in memory) is a sibling, not a parent, of `encodeCore` (= emitting SMT-LIB from those `Term` objects). The current order is also children-before-parent, which is unusual.

Suggested reordering, parent-first with children indented consistently beneath:

```suggestion
  if profile then
    let _ ← (IO.println s!"[profile]     Preprocess obligations: {nsToMs (timingNs.getD (toString Verifier.Timing.preprocessObligations) 0)}ms" |>.toBaseIO)
    let _ ← (IO.println s!"[profile]     SMT term construction: {nsToMs (timingNs.getD (toString Verifier.Timing.smtEncoding) 0)}ms" |>.toBaseIO)
    let _ ← (IO.println s!"[profile]     Solver/file writing: {nsToMs (timingNs.getD (toString Verifier.Timing.solverFileWriting) 0)}ms" |>.toBaseIO)
    let _ ← (IO.println s!"[profile]       encodeCore: {nsToMs (timingNs.getD (toString Imperative.SMT.Timing.encodeSMT) 0)}ms" |>.toBaseIO)
    let encodeCorePrefix := "encodeCore."
    let encodeCoreEntries := timingNs.toList.filter (·.1.startsWith encodeCorePrefix)
      |>.mergeSort (fun a b => a.1 < b.1)
    for (key, val) in encodeCoreEntries do
      let suffix := key.drop encodeCorePrefix.length
      let _ ← (IO.println s!"[profile]         {suffix}: {nsToMs val}ms" |>.toBaseIO)
    let _ ← (IO.println s!"[profile]       runSolver: {nsToMs (timingNs.getD (toString Imperative.SMT.Timing.runSolver) 0)}ms" |>.toBaseIO)
    let _ ← (IO.println s!"[profile]     Obligations: {obligations.size} total, {peResolvedCount} resolved by evaluator" |>.toBaseIO)
  return (results, stats)
```

I also renamed the user-visible label from `SMT encoding` to `SMT term construction` to disambiguate it from `encodeCore` (both phases are forms of "SMT encoding"). And I lifted `"encodeCore."` into a local to avoid the literal being duplicated on two lines.

Strata/Languages/Core/Verifier.lean
```diff
@@ -1120,10 +1175,20 @@ def verifySingleEnv (oblProgram : Program)
...
         if options.stopOnFirstError then break
+
   if profile then
-    let _ ← (IO.println s!"[profile]     Preprocess obligations: {nsToMs preprocessNs}ms" |>.toBaseIO)
-    let _ ← (IO.println s!"[profile]     SMT encoding: {nsToMs smtEncodeNs}ms" |>.toBaseIO)
-    let _ ← (IO.println s!"[profile]     Solver/file writing: {nsToMs solverNs}ms" |>.toBaseIO)
+    let _ ← (IO.println s!"[profile]     Preprocess obligations: {nsToMs (timingNs.getD (toString Verifier.Timing.preprocessObligations) 0)}ms" |>.toBaseIO)
+    let _ ← (IO.println s!"[profile]     SMT encoding: {nsToMs (timingNs.getD (toString Verifier.Timing.smtEncoding) 0)}ms" |>.toBaseIO)
+    -- Print encodeCore.* sub-timings
+    let encodeCoreEntries := timingNs.toList.filter (·.1.startsWith "encodeCore.")
```
tautschnig, line 1181: The string `"encodeCore."` is also the prefix argument to `#derive_prefixed_toString Timing "encodeCore"` at line 47. If the enum prefix is ever renamed (say, to `smtEncode`), the profiler silently drops its sub-timings and no one notices until someone eyeballs the output. Either:

- factor out a `def Timing.prefix : String := "encodeCore."` next to the `inductive Timing`, or
- derive the prefix from a representative constructor's `toString` at runtime.

The `#derive_prefixed_toString` elaborator could also emit such a constant alongside the `ToString` instance — worth considering if any other `Stats` enum in the codebase develops the same coupling.

Strata/DL/Imperative/SMTUtils.lean
```diff
@@ -292,38 +294,58 @@ def addLocationInfo {P : PureExpr} [BEq P.Ident]
       Strata.SMT.Solver.setInfoString message.fst message.snd
     | .none => pure ()
 
+/-- Timing keys tracked by `dischargeObligation`. -/
+inductive Timing where
+  | encodeSMT
+  | runSolver
+
```
tautschnig, line 301: Naming nit: this local `Timing` enum lives in the same file-local scope as the other `SMTUtils` helpers and shares a plain name with `Strata.SMT.Encoder.Timing` (at `Verifier.lean:34`) and `Strata.Languages.Core.Verifier.Timing` (at `Verifier.lean:1053`). All three are disambiguated by namespace and by their `#derive_prefixed_toString` prefix, so there's no correctness issue, but callers end up writing `Imperative.SMT.Timing.encodeSMT` / `Verifier.Timing.smtEncoding` / bare `Timing.prelude` depending on context, which is noisy. If you want to collapse the noise, a single `module Strata.Util.ProfileKeys` collecting all profile enums (or at least more specific per-use names like `DischargeTiming` / `EncodeCoreTiming` / `VerifyTiming`) would help.

MikaelMayer (commented):
🔍 LGTM — all comments addressed. The new docstring on `recordNanos` clearly explains the monadic sequencing argument and the caveat about lazy pure values. Using `timing.add` in `recordNanos` is a nice consistency improvement.

Strata/Languages/Core/Verifier.lean
aqjune-aws: These were intentionally left because their running times are negligible - similarly line 84 ~ 86 are unwrapped.

Strata/Languages/Core/Verifier.lean
aqjune-aws: I think this can be addressed in the following pull request which will give more structure to the profiling infrastructure by @joehendrix - otherwise there can be a conflict.

Strata/Languages/Core/Verifier.lean
aqjune-aws: I think this can be also considered after profile implementation restructuring. Would like to minimize conflicts...

Strata/Languages/Core/Verifier.lean
aqjune-aws: "encodeCore." seems unique and easy to update - I would like to keep the current version.

Strata/DL/Imperative/SMTUtils.lean
aqjune-aws: It seems `Timing` in SMTUtils.lean is in Imperative.SMT whereas `Timing` in Verifier.lean line 34 is in Strata.SMT.Encoder, so... I hope this is ok.

Strata/Util/Profile.lean
aqjune-aws: Add the `measureNanos_nonzero` test :)

Strata/Util/Profile.lean
aqjune-aws: It seems `TimingInfo.mergeAdd_getD` needs extra layer of lemmas to prove - I guess converting `HashMap` to `List` and doing induction on the first argument `a` is necessary, but can be slightly lengthy. I added the first two lemmas, thank you for suggestion :)

MikaelMayer (commented):
🔍 LGTM — all comments addressed. The new docstring on `recordNanos` clearly explains the monadic sequencing argument and the caveat about lazy pure values. Using `timing.add` in `recordNanos` is a nice consistency improvement.

Strata/Util/Profile.lean
aqjune-aws: This is actually a valid point... `expr ()` in `measureNanos` could have been evaluated in advanced before `let t1 ← IO.monoNanosNow` is run. @joehendrix I am again leaning toward the simple `t1 == 0` hack to add a false data dependency... What do you think?

aqjune-aws:
This edit will be piled on top of #1093

Issue #650, opened 2026-03-24
StrataVerify Panic in Type inference


yizhou7:
I seem to run into a `PANIC` during type inferencing when running `StrataVerify`. Here is a small example Core program that would trigger this bug:
```
program Core;

datatype Option (a : Type) { None(), Some(val: a) };

datatype MethodSetting () {
  MethodSetting_Cons(LoggingLevel: Option string)
};

datatype Stage () {
  Stage_Cons(MethodSettings: Option (Sequence MethodSetting))
};

function method_ok(ms: MethodSetting): bool {
  (MethodSetting..LoggingLevel(ms) == Some("ERROR"))
  || (MethodSetting..LoggingLevel(ms) == Some("INFO"))
}

function stage_ok(stage: Stage): bool {
  forall i: int ::
    (Stage..MethodSettings(stage) != None()
     && 0 <= i
     && i < Sequence.length(Option..val(Stage..MethodSettings(stage))))
    ==>
    method_ok(Sequence.select(Option..val(Stage..MethodSettings(stage)), i))
}

const s: Stage;
const grouped: Sequence Stage;
axiom Sequence.length(grouped) == 1;
axiom Sequence.select(grouped, 0) == s;

procedure Check() returns ()
{
  assert [check]:
    forall i: int :: (0 <= i && i < Sequence.length(grouped)) ==>
      stage_ok(Sequence.select(grouped, i));
};
```
Here is the error I am getting:
```
PANIC at Strata.Elab.inferType Strata.DDM.Elab.Core:988:11: Cannot instantiate type Strata.TypeExprF.bvar { start := { byteIdx := 3018 }, stop := { byteIdx := 3019 } } 2 with args #[Strata.ArgF.op
    { ann := { start := { byteIdx := 0 }, stop := { byteIdx := 0 } },
      name := { dialect := "", name := "" },
      args := #[] },
  Strata.ArgF.expr
    (Strata.ExprF.app
      { start := { byteIdx := 622 }, stop := { byteIdx := 662 } }
      (Strata.ExprF.fvar { start := { byteIdx := 622 }, stop := { byteIdx := 633 } } 5)
      (Strata.ArgF.expr
        (Strata.ExprF.app
          { start := { byteIdx := 634 }, stop := { byteIdx := 661 } }
          (Strata.ExprF.fvar { start := { byteIdx := 634 }, stop := { byteIdx := 655 } } 15)
[...]

PR #1125, opened 2026-05-05
Add dialect_option typecheck off to bypass DDM type checker

joehendrix:
## Summary

The DDM type checker uses unification to infer implicit type arguments, but this fails in some cases — notably when template-generated accessors with unresolved tvar return types are composed with polymorphic functions. This PR adds a per-dialect option to bypass type checking entirely, filling implicit type parameter slots with anonymous type placeholders instead.

## How to use

In a dialect definition, add `dialect_option typecheck off;` to disable type checking for all programs using that dialect:

```
#dialect
dialect MyDialect;
dialect_option typecheck off;

type Foo;
fn bar (A : Type, x : A) : A => x;
...
#end
```

Programs using the dialect will skip `inferType` and `unifyTypes` for expression arguments. Variable name resolution and global context population still work normally — only type inference is bypassed.

## Test changes

`StrataTest/DDM/TypecheckSkip.lean` defines an inline dialect with parameterized types, polymorphic functions, and perField accessor templates. It demonstrates that the accessor-into-polymorphic-fn pattern (from issue #650 / PR #734) produces type errors with typecheck on, and elaborates successfully with typecheck off.

## Details

- **AST.lean**: Add `TypeExprF.skip` (anonymous type placeholder `.tvar loc ""`), `Dialect.typecheck : Bool := true`, refactor `Program.globalContext` into explicit `computeGlobalContext` function
- **Elab/Core.lean**: Add `ElabContext.typecheck`, skip `inferType`/`unifyTypes` in `elabSyntaxArg`, fill unfilled type param slots in `runSyntaxElaborator` post-pass, skip pre-registration in `elabOperation`, fix `applyNArgs` to handle tvar types
- **Elab/DeclM.lean**: Add `DeclContext.typecheck`
- **Elab/DialectM.lean**: Add `elabSetOptionCommand` handler dispatching on `"typecheck"` option
- **BuiltinDialects/StrataDDL.lean**: Add `setOptionCommand` with syntax `dialect_option <name> <value> ;`
- **Elab.lean**: Thread `Dialect.typecheck` into `DeclContext` in `elabProgramRest`
- **Ion.lean, Python/Specs/DDM.lean**: Update `Program` construction for `globalContext` refactor


Strata/DDM/Elab/Core.lean
```diff
@@ -111,6 +111,10 @@ def applyNArgs (tctx : TypingContext) (e : TypeExpr) (n : Nat) := aux #[] e
     if argsLt : args.size < n then
       match tctx.hnf e with
       | .arrow _ a r => aux (args.push a) r
+      | .tvar ann _ =>
+        let placeholder := .placeholder ann
+        let tvars := Array.replicate (n - args.size) placeholder
+        .ok (⟨args ++ tvars, by simp [tvars]; omega⟩, placeholder)
```
tautschnig, line 121: **Unconditional change.** This new `.tvar` arm runs regardless of the `typecheck` flag — it's a general relaxation of `applyNArgs`, not part of the opt-in path. A `typecheck = true` program that previously errored with `.error (args, .tvar _ _)` will now succeed with placeholder arguments.

Is that deliberate? If the intent is to preserve existing behaviour for typecheck-on programs and only relax for typecheck-off:

```suggestion
      | .tvar ann _ =>
        -- Producing placeholder arguments for tvars is required by the
        -- `typecheck off` path (so that arity-dependent slots can be filled
        -- without unification); it is also a relaxation for typecheck-on
        -- programs where `applyNArgs` previously errored. If this broader
        -- semantics is undesired outside the typecheck-off path, gate here.
        let placeholder := .placeholder ann
        let tvars := Array.replicate (n - args.size) placeholder
        .ok (⟨args ++ tvars, by simp [tvars]; omega⟩, placeholder)
```

Either way, a comment explaining the scope of the change would help — I reached for the PR description to discover that this was general.

A small proof to lock the contract of this branch (see the main review body):

```lean
theorem applyNArgs_tvar_placeholder
    (tctx : TypingContext) (ann : _) (name : String) (n : Nat) :
    applyNArgs tctx (.tvar ann name) n =
    .ok (Vector.replicate n (.placeholder ann), .placeholder ann) := by
  …
```

Strata/DDM/Elab/DialectM.lean
```diff
@@ -909,6 +909,19 @@ def elabMdCommand : DialectElab := fun tree => do
...
+def elabSetOptionCommand : DialectElab := fun tree => do
+  let .isTrue _ := checkTreeSize tree 2
+    | logError tree.info.loc "setOptionCommand: unexpected tree size"; return
+  let nameInfo := tree[0].info.asIdent!
+  let valueInfo := tree[1].info.asIdent!
+  match nameInfo.val with
+  | "typecheck" =>
+    match valueInfo.val with
+    | "on" => modifyDialect fun d => { d with typecheck := true }
+    | "off" => modifyDialect fun d => { d with typecheck := false }
```
tautschnig, line 921: Two small improvements to make this scale as more options land:

1. **Dispatch via a data-driven table.** The nested `match` on `("typecheck", "on"/"off")` is already two levels deep; a third option adds a third. Extract a `DialectOption` sum with `fromKeyValue`:

```lean
inductive DialectOption where
  | typecheck (value : Bool)

def DialectOption.fromKeyValue (key : String) (val : String)
    : Except String DialectOption :=
  match key, val with
  | "typecheck", "on" => .ok (.typecheck true)
  | "typecheck", "off" => .ok (.typecheck false)
  | "typecheck", v => .error s!"Expected 'on' or 'off' for option 'typecheck', got '{v}'."
  | key, _ => .error s!"Unknown option '{key}'."

def DialectOption.apply : DialectOption → Dialect → Dialect
  | .typecheck b, d => { d with typecheck := b }
```

Then `elabSetOptionCommand` is just a parse-then-apply pipeline and error messages live in one place.

2. **Include the bad value in the error** — currently the error `"Expected 'on' or 'off' for option 'typecheck'."` doesn't echo back what the user typed, which is less helpful than e.g. `"Expected 'on' or 'off' for option 'typecheck', got 'maybe'."`.

Also: add a `#guard_msgs` fixture to lock these error messages so a future rewrite doesn't silently change them — the tests in `TypecheckSkip.lean` only cover the successful `off` case.

docs/verso/DDMDoc.lean
```diff
@@ -191,6 +191,23 @@ declared.  This includes transitive imports of the dialect being imported.
...
+Sets the dialect option _name_ to _value_.
+:::
+
+The following options are supported:
+
+- `typecheck` (`on` | `off`, default `on`): When set to `off`, the elaborator
+  skips type inference and unification for expression arguments. Implicit type
+  parameter slots are filled with anonymous type placeholders. Variable name
+  resolution and global context population still operate normally.
+
```
tautschnig, line 210: The doc says what `typecheck off` does. It doesn't warn about what it *doesn't* do. Worth one paragraph covering the failure modes that are moved downstream:

```suggestion
## Dialect Options

Dialect options configure elaboration behavior for programs using the dialect.

:::paragraph
`dialect_option` _name_ _value_`;`

Sets the dialect option _name_ to _value_.
:::

The following options are supported:

- `typecheck` (`on` | `off`, default `on`): When set to `off`, the elaborator
  skips type inference and unification for expression arguments. Implicit type
  parameter slots are filled with anonymous type placeholders. Variable name
  resolution and global context population still operate normally.

  This is intended as a workaround for cases where the type checker cannot
  infer implicit type arguments — notably when template-generated accessors
  with unresolved tvar return types are composed with polymorphic functions
  (see issue #650).

  ⚠️  With `typecheck off`, type errors in a program are not detected at
  elaboration. They will surface at later pipeline stages (VC generation,
  symbolic evaluation, SMT encoding) with less-helpful diagnostics, or — in
  the worst case — will silently produce ill-typed intermediate terms. Only
  use this flag when the elaboration error is known to be a type-checker
  limitation rather than a real type mismatch.
```

Without this paragraph, a user who turns the flag on because they saw it in an example is likely to discover the downside the hard way when their downstream pipeline fails opaquely.

StrataTest/DDM/TypecheckSkip.lean
```diff
@@ -0,0 +1,173 @@
...
+
+datatype Maybe (a : Type) { Nothing(), Just(val: a) };
+
+const m: Maybe (Lst Inte);
+
+procedure Test () returns ()
+{
+  assert [t1]: Lst.sel(Maybe..val(m), 0) == 0;
+};
+#end
```
tautschnig, line 175: The file covers the positive case (typecheck-off succeeds on the motivating shape). A few additional fixtures would pin the feature's contract:

1. **Typecheck-off still catches unresolved identifiers** — pin the "only type inference is bypassed" promise from the PR description:
   ```lean
   /-- Even with typecheck off, an undefined identifier must be rejected. -/
   /--
   error: Unknown identifier 'undefined_name'  -- or whatever the actual message is
   -/
   #guard_msgs in
   def typecheckOffStillCatchesUndefined :=
   #strata
   program TestTCOff;
   procedure Test () returns () {
     assert [t]: undefined_name == 0;
   };
   #end
   ```

2. **Invalid option values produce clean errors**:
   ```lean
   /-- `dialect_option typecheck maybe` should fail with the expected message. -/
   /-- error: Expected 'on' or 'off' for option 'typecheck'. -/
   #guard_msgs in
   #dialect
   dialect BadOption;
   dialect_option typecheck maybe;
   #end

   /-- `dialect_option nonsense on` should fail with the expected message. -/
   /-- error: Unknown option 'nonsense'. -/
   #guard_msgs in
   #dialect
   dialect BadOption2;
   dialect_option nonsense on;
   #end
   ```

3. **Primary-dialect semantics across import boundary** — what happens when `TestTCOn` (typecheck on) *imports* `TestTCOff`, or vice versa? `elabProgramRest` uses only the primary dialect's flag. Worth a fixture plus a one-line docstring where the flag is read.

4. **Ion round-trip** — if a dialect-with-`typecheck-off` is serialized to Ion and re-read, does the flag survive? I didn't see serialization support for `Dialect.typecheck` in the Ion changes. If it doesn't persist, that's a correctness gap for workflows that round-trip through Ion.

Strata/DDM/AST.lean
```diff
@@ -113,6 +113,9 @@ deriving BEq, Inhabited, Repr
 
 namespace TypeExprF
 
+/-- An anonymous type placeholder used when type checking is skipped. -/
+def placeholder {α} (loc : α) : TypeExprF α := .tvar loc ""
```
tautschnig, line 117: Update the docstring to reflect that `placeholder` is used in two places, not one:

```suggestion
/-- An anonymous type placeholder (`.tvar loc ""`) used when a concrete type
    cannot be inferred. Currently produced by:
    - `runSyntaxElaborator` post-pass when `typecheck off` leaves type slots unfilled,
    - `applyNArgs` when the type of the applied expression is itself a tvar
      (the latter runs regardless of the `typecheck` flag; see Elab/Core.lean:114).

    Represented as an unnamed tvar — compatible with `TypingContext.hnf`
    and `unifyTypes`, which treat empty-name tvars as fresh / matching-any. -/
def placeholder {α} (loc : α) : TypeExprF α := .tvar loc ""
```

Also worth a trivial `@[simp]` definitional theorem so downstream proofs don't need to unfold:

```lean
@[simp] theorem placeholder_def {α} (loc : α) :
    placeholder loc = TypeExprF.tvar loc "" := rfl
```

joscoh (dismissed):

tautschnig (commented):
I'm concerned about the *shape* of the feature (not any specific details of the implementation). `dialect_option typecheck off;` is an **escape hatch around a type-checker bug we don't yet understand**. Merging this is fine and pragmatic, but I'd push for the following so the escape hatch doesn't silently become permanent technical debt:

1. **File a tracking issue for the underlying type-checker bug.** The PR description mentions "accessor-into-polymorphic-fn" as the specific failure mode from #650/#734, but there's no issue with "remove `typecheck off` once the root cause is fixed". Without that, any future user who hits a type error is going to reach for `typecheck off` as a Band-Aid, and we won't have a way to tell whether the root cause has been addressed.

2. **Document which categories of program are expected to need this.** The test file (`TypecheckSkip.lean`) names exactly one shape (accessor → polymorphic fn). If that's the only expected use, say so in the DDM manual. If there are others, enumerate them.

3. **Warn about `typecheck off`'s blast radius in the docs.** `docs/verso/DDMDoc.lean` says only what the option does, not what it *doesn't* do. With `typecheck off`, a program with a genuine type error (e.g., passing a `Lst Inte` where `Inte` is expected) will elaborate silently. Downstream consumers — the VC generator, the symbolic evaluator, the SMT encoder — will then encounter the type error and produce confusing diagnostics ("Expression has type X when Y expected" at an IR level the user hasn't seen) or, worse, will silently proceed with ill-typed terms. A paragraph in the doc saying "turning this off means type errors surface at later pipeline stages with less-helpful messages" is fair warning.

4. **Audit `applyNArgs`'s new `.tvar` case — it's a *general* change, not gated on `typecheck`.** At `Elab/Core.lean:114-117`, the new branch
   ```lean
   | .tvar ann _ =>
     let placeholder := .placeholder ann
     let tvars := Array.replicate (n - args.size) placeholder
     .ok (⟨args ++ tvars, by simp [tvars]; omega⟩, placeholder)
   ```
   runs regardless of the `typecheck` flag. Previously the function errored on `tvar`; now it produces placeholder arguments. This means **programs elaborated with `typecheck = true` (the default) can now succeed where they previously errored** — the broadening is not confined to the opt-in path. Is that intended? If so, worth a comment; if not, gate on `(← read).typecheck`.

5. **`Program.globalContext` refactor is a cleanly-scoped improvement** — removing the panic-in-default and exposing `computeGlobalContext` + `Program.create` is strictly better. I verified no other callers in-tree reach into `Program.mk` directly after this change. Good.

**Test coverage:**

The one test file covers the positive case (typecheck-off succeeds where typecheck-on fails). Missing cases that would pin the feature's behaviour:

- **Negative test: typecheck off still catches unresolved identifiers.** The PR description promises "Variable name resolution and global context population still work normally — only type inference is bypassed". A test that demonstrates an undefined identifier (e.g., `assert [t]: undefined_name == 0;`) still fails under `typecheck off` would pin this contract.
- **Test: `typecheck off` across a dialect import boundary.** What happens when a `typecheck = true` dialect imports a `typecheck = false` dialect? `elabProgramRest` threads `d.typecheck` into `DeclContext` — where `d` is the *primary* dialect. So the imported dialect's flag is ignored. Worth a test (and a docstring note) to pin this semantics.
- **Test: invalid `dialect_option` values fail cleanly.** `dialect_option typecheck maybe;` or `dialect_option nonsense on;` should produce the explicit "expected 'on' or 'off'" / "Unknown option" errors — worth a `#guard_msgs` fixture to lock those error messages.
- **Test: option flag survives Ion round-trip.** A DDM program with `typecheck off` serialized to Ion and re-read should preserve the flag. If it doesn't, the `typecheck` field may need serialization support.

**Proof-coverage suggestions:**

Three small theorems pinning the load-bearing invariants:

```lean
/-- `placeholder` is definitionally a `tvar` with empty name. -/
@[simp] theorem TypeExprF.placeholder_def {α} (loc : α) :
    TypeExprF.placeholder loc = TypeExprF.tvar loc "" := rfl

/-- `applyNArgs` on any tvar produces exactly `n` placeholder-typed
    arguments — contract for the new `.tvar` branch. -/
theorem applyNArgs_tvar_placeholder
    (tctx : TypingContext) (ann : _) (name : String) (n : Nat) :
    applyNArgs tctx (.tvar ann name) n =
    .ok (Vector.replicate n (.placeholder ann), .placeholder ann) := by
  cases n <;> simp [applyNArgs, applyNArgs.aux, …]

/-- Elaboration with `typecheck = false` still rejects unresolved
    identifier references (only type unification is bypassed). -/
theorem elabProgram_typecheck_off_preserves_resolution
    (d : Dialect) (h : d.typecheck = false) (cmds : Array Operation) :
    elabProgram d cmds = .ok p →
    ∀ op ∈ p.commands, ∀ id ∈ op.idents, identResolves p.globalContext id
```

The first is trivial; the second locks the new branch; the third is the user-visible contract ("name resolution still works") the PR description promises. Even as `sorry`-terminated stubs these would make the contract explicit.

**Refactor / style:**

- `elabSyntaxArg` has three copies of `if !typecheck then return trees.set argIdx (some tree)` at `:1174`, `:1195` (and conceptually a third for the unnamed preType arm). A tiny helper
  ```lean
  private def bypassIfTypecheckOff (tree : Tree) (trees : …) : ElabM _ := do
    if !(← read).typecheck then return trees.set argIdx (some tree) else …
  ```
  would collapse the duplication and make future maintenance of the skip condition land in one place.
- `elabSetOptionCommand` uses a nested `match` on string pairs (`"typecheck" × "on"/"off"`). If more options are added, this will balloon. A `DialectOption` sum type with `fromKeyValue : String → String → Except String DialectOption` would centralize the parsing.
- `dialect_option` is a nice name, but the current implementation only recognizes one option (`typecheck`). If this *is* the only expected option long-term, consider a narrower syntax like `@[typecheck off] dialect …`. If more are coming, say so in a comment.
- **Minor: `TypeInfo` construction at `Elab/Core.lean:1266-1273`** uses `inputCtx := tctx0` (the initial context) rather than `t.resultContext`. That's intentional (the placeholder doesn't depend on earlier args) but worth a one-line comment — a reader stepping through the generic flow will expect the per-arg context.
- **The renamed `TypeExprF.skip` → `TypeExprF.placeholder`** (commit 9449f1c65) is an improvement. The docstring now says "An anonymous type placeholder used when type checking is skipped." — but `applyNArgs` now uses placeholder regardless of the typecheck flag. Update the docstring:
  ```
  /-- An anonymous type placeholder (`.tvar loc ""`) used when a concrete type
      cannot be inferred — either because type checking is skipped at the
      dialect level, or because `applyNArgs` encountered a tvar where an
      arrow type was expected. -/
  ```

joehendrix:
I'll take a look at more detail.  But I'm going to react to this:

> I'm concerned about the shape of the feature (not any specific details of the implementation). dialect_option typecheck off; is an escape hatch around a type-checker bug we don't yet understand.

In fact, I have a pretty good understanding of the issue #734 is a fix.  However, there is a desire from the Strata Core folks to have this feature rather than expanding higher order type checking.

I'll look into the more detailed review soon, and I appreciate it, but I'm kind of stuck here.  We have three options from my perspective:

1. Fix the bug (e.g., #734 or a better PR).
2. Leave as is and just panic (basically won't fix).
3. Provide an escape hatch (this PR).

At this point, I am just ashamed how much work is going into **not** fixing an issue.  I'll review the PR comments, but if it's too much work, then I'm going to move to option 2.

joscoh:
I disagree that this PR is problematic. I am fine mentioning this in the docs, but this option/escape hatch should be used when there is a separate typechecker that will be run on the resulting programs (as is the case for Core and Boole). It does not introduce any soundness/mistranslation issues, but rather reduces typechecking to a single function that can be reasoned about, avoiding some duplication.

atomb:
> I disagree that this PR is problematic. I am fine mentioning this in the docs, but this option/escape hatch should be used when there is a separate typechecker that will be run on the resulting programs (as is the case for Core and Boole). It does not introduce any soundness/mistranslation issues, but rather reduces typechecking to a single function that can be reasoned about, avoiding some duplication.

I second this.

Strata/DDM/AST.lean
joehendrix: This suggestion is to duplicate the right hand side of the `placeholder` by adding an extra theorem, and explain the places it is used.

I don't understand why either are a good idea.  I could add `@[simp]` to `placeholder` def if that's useful for a theorem., but I'd rather do that if/when it is needed. 

I also don't think documenting every where a definition is used is good practice.  What if someone uses `.tvar` in another place?

Strata/DDM/Elab/DialectM.lean
joehendrix: Maybe let's add this if/when we create other options?  That's easy to do.

Strata/DDM/Elab/Core.lean
joehendrix: `tvar` has always been a type variable that was created to bypass typechecking rather than implement higher order type checking in the DDM.  The Strata Core dialect has its own typechecker, and that's what the Strata Core developers want.  

The behavior here is consistent with that type variable use case.  I don't think we need additional redundancy to "lock the contract".

docs/verso/DDMDoc.lean
joehendrix: The documentation string you propose is not accurate.  It is not intended as a workaround, but rather something intended for the Strata Core.  Essentially, Strata Core started using the Expr/Type, the developers outgrew its capabilities and decided to implement their own typechecker, but didn't want to port everything to avoid Type/Expr at the DDM layer, and this enables that without bugs.

Yes eventually, I'd like a richer type system, but framing this so negatively isn't right in my view either.

StrataTest/DDM/TypecheckSkip.lean
joehendrix: Thanks for the test suggestions. Here's where things stand:

1. **Undefined identifiers** — added (`typecheckOffStillCatchesUndefined` in TypecheckSkip.lean).
2. **Invalid option values** — added (`BadOptionValue` and `BadOptionName` fixtures with `#guard_msgs`).
3. **Cross-import boundary** — the typecheck flag is a property of the program's primary dialect; imported dialects' flags aren't consulted. I've documented this in the test file and in the DDM docs rather than adding a separate test — it's hard to regress on this without breaking the feature entirely.
4. **Ion round-trip** — implemented in commit a51e8097. The `Dialect.typecheck` flag is now serialized as `{type: option, name: "typecheck", value: "off"}` when non-default, and deserialized back. Python tooling updated as well.

joehendrix:
Responding to the top-level review:

This isn't an escape hatch around a type-checker bug — it's a deliberate design choice. The Strata Core team implemented their own type checker downstream and doesn't want/need the DDM elaborator's type inference. They haven't ported everything away from Type/Expr at the DDM layer, and `typecheck off` enables that without issues. The DDM type system was always limited (tvars were already placeholders for "we don't know the type"), and this makes that explicit.

Re: the specific suggestions:
- **Tracking issue** — I don't think one is needed. This isn't a bug to fix; it's a deliberate opt-in for dialects that manage their own type checking downstream.
- **Revert path** — the feature is additive and backward-compatible (typecheck defaults to `on`), so reverting is straightforward if ever needed, but I don't expect it to be.
- **Broader testing** — covered by the fixtures added (see inline comment response on TypecheckSkip.lean).

Strata/DDM/Ion.lean
```diff
@@ -1503,14 +1515,28 @@ def fromIonFragment (dialect : DialectName) (f : Ion.Fragment) : Except String D
...
+      let some (_, nameVal) := fields.find? (·.fst == nameId)
+        | throw "Could not find option name"
+      let optName ← FromIonM.asString "Option name" nameVal ctx
+      let some (_, valueVal) := fields.find? (·.fst == valueId)
+        | throw "Could not find option value"
+      let optValue ← FromIonM.asString "Option value" valueVal ctx
+      match optName, optValue with
+      | "typecheck", "off" => pure (imports, decls, false)
+      | "typecheck", "on" => pure (imports, decls, true)
+      | _, _ => throw s!"Unknown option '{optName}' = '{optValue}'"
```
MikaelMayer: 🤖🔍 The wildcard `| _, _` catches both unknown option names and known options with invalid values. For example, `("typecheck", "maybe")` would produce `"Unknown option 'typecheck' = 'maybe'"` — but the option *is* known, only the value is invalid.

The dialect elaborator in `DialectM.lean` distinguishes these cases (`"Expected 'on' or 'off'…"` vs `"Unknown option…"`). Consider matching the same structure here for consistent diagnostics:

```lean
| "typecheck", v => throw s!"Expected 'on' or 'off' for option 'typecheck', got '{v}'"
| name, _ => throw s!"Unknown option '{name}'"
```

Strata/Languages/Python/Specs/DDM.lean
```diff
@@ -7,20 +7,15 @@ module
...
+import Strata.DDM.Format
 import Strata.DDM.Ion
 
-public section
-
 namespace Strata.Python
 
 /-- Converts a Python identifier to an annotated string for DDM serialization. -/
-private def PythonIdent.toDDM (d : PythonIdent) : Ann String SourceRange :=
+def PythonIdent.toDDM (d : PythonIdent) : Ann String SourceRange :=
```
MikaelMayer, line 18: 🤖🔍 The visibility changes in this file (removing `private` from ~15 definitions, removing the `public section`/`end` wrapper, adding explicit `public` to `readDDM`/`writeDDM`) are unrelated to the typecheck-skip feature. They appear to be a prerequisite for the `meta import` in `SystemSymbolIds.lean` or the `SpecsTest.lean` fix, but the PR description doesn't explain why these definitions need to be visible outside the module now.

If this was needed to fix a build issue from the merge, a one-line note in the PR description would help reviewers understand the motivation.

Strata/DDM/Elab/Core.lean
```diff
@@ -1119,7 +1129,12 @@ partial def elabOperation (tctx : TypingContext) (stx : Syntax) : ElabM Tree :=
   if not success then
     return default
   let getKind i := .ofArgDeclKind argDecls[i].kind
+  let typecheck := (← read).typecheck
   let ((args, newCtx), success) ← runChecked <|
+    if !typecheck then do
```
MikaelMayer, line 1136: 🤖🔍 When `typecheck` is off, `elabOperation` skips `elaborateWithPreRegistrationCore` entirely — meaning `preRegisterTypesScope` and `preRegisterFunctionsScope` are not consulted. This means datatype/function pre-registration (populating the global context before full elaboration) is also skipped.

Is that intentional? The test uses `command_datatypes` which has `@[preRegisterTypes(datatypes)]`, so pre-registration must still work for the test to pass. Looking more carefully, `runSyntaxElaborator` still calls `elabSyntaxArg` which calls `catElaborator` for category args — so the scope elaboration still happens. But the *pre-registration* phase (extracting type/function info and adding it to the global context before elaborating the body) is skipped.

If this is safe because `computeGlobalContext` already populates the global context at program creation time, it would be worth a comment explaining that.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -169,17 +169,13 @@ public def buildPySpecLaurel (pyspecEntries : Array (String × String))
       | .Constrained ct => ct.name.text
       | .Datatype dt => dt.name.text
       | .Alias ta => ta.name.text
-    match seenTypes.get? name with
-    | some prevFile =>
+    if let some prevFile := seenTypes[name]? then
```
MikaelMayer, line 172: 🤖🔍 Nit: This `match` → `if let` refactoring is a style improvement but unrelated to the typecheck feature. Fine to include, just noting it's a drive-by cleanup.

MikaelMayer (commented):
🤖🔍 Overall this is a well-structured feature addition. The threading of the `typecheck` flag through the elaboration pipeline is clean, the Ion round-trip is handled correctly, and the test coverage demonstrates the core use case. A few items below.

Positives:
- The `computeGlobalContext` extraction is a good refactor — it makes `Program.create` the single place where the panic lives, and `globalContext` is now an explicit field rather than a computed default.
- The `SymbolTable` privacy hardening and the `SystemSymbolIds.lean` extraction are clean improvements.
- Test file is well-commented and covers the key scenarios (typecheck-on fails, typecheck-off succeeds, name resolution still works, invalid options error).

Strata/Languages/Python/Specs/DDM.lean
joehendrix: This didn't make them visible.  Removing the public section changed the default visibility to private eliminate the need to add private.

Strata/DDM/Ion.lean
joehendrix: Good catch. Split the match to distinguish the two cases:

```lean
| "typecheck", v => throw s!"Expected 'on' or 'off' for option 'typecheck', got '{v}'"
| name, _ => throw s!"Unknown option '{name}'"
```

Done in 994676ad.

Strata/DDM/Elab/Core.lean
joehendrix: Good question. `computeGlobalContext` (called during program creation) already populates the global context with type and function information before elaboration runs, so the pre-registration pass isn't needed when typecheck is off. Added a comment explaining this in 994676ad.

Strata/Languages/Python/PySpecPipeline.lean
joehendrix: Yep, drive-by cleanup from the merge.

MikaelMayer (commented):

- Ion error messages now properly distinguish unknown options from invalid values (commit 994676ad) ✅
- Pre-registration skip has an explanatory comment (commit 994676ad) ✅
- Visibility changes and PySpecPipeline nit were informational only, no action needed ✅

docs/verso/DDMDoc.lean
joehendrix: It looks like my agent just decided to just accept your suggested text.  I am fine with this as it will help me merge sooner.

shigoel (approved):

MikaelMayer (approved):

PR #1127, opened 2026-05-05
feat(transform): Add optional SSA transformation for Strata Core

sagjoshi:
Introduce a Static Single Assignment (SSA) transformation that converts
procedure bodies so every variable is assigned exactly once via init.
Mutations (set, havoc) become fresh init declarations. At if-then-else
join points, conditional init expressions merge divergent variable versions.

Uses the standard CoreTransformM monad with CoreGenState.gen for fresh
name generation (ssa_ prefix), aligning with transform API conventions.

Integration:
- Available as \`--pass ssa\` in the transform CLI command
- Available as \`TransformPass.ssa\` in the SimpleAPI
- Not part of the default verification pipeline (opt-in)

Files:
- \`Strata/Transform/SSA.lean\`: transformation implementation
- \`StrataTest/Transform/SSA.lean\`: 16 tests
- \`Strata/SimpleAPI.lean\`: TransformPass.ssa + applyPass wiring
- \`StrataMain.lean\`: ssa added to valid transform passes

### DIFF
--- a/Strata/Transform/SSA.lean
+++ b/Strata/Transform/SSA.lean
@@ -104,28 +104,28 @@ def transformCmd (env : Env) (cmd : Command) : CoreTransformM (Env × List State
     | some info =>
       let freshId ← genSSAIdent name.name
       let some mty := LTy.toMonoType? info.ty
-        | throw s!"SSA: type of '{name.name}' is not a monotype"
+        | throw (Strata.DiagnosticModel.fromMessage s!"SSA: type of '{name.name}' is not a monotype")
       let env' := env.insert name.name { ident := freshId, ty := info.ty }
       incrementStat s!"{Stats.renamedVars}"
       return (env', [Statement.init freshId (.forAll [] mty) (.det expr') smd])
-    | none => throw s!"SSA: variable '{name.name}' not found in environment (set)"
+    | none => throw (Strata.DiagnosticModel.fromMessage s!"SSA: variable '{name.name}' not found in environment (set)")
   | .cmd (.set name .nondet smd) =>
     match env.get? name.name with
     | some info =>
       let freshId ← genSSAIdent name.name
       let some mty := LTy.toMonoType? info.ty
-        | throw s!"SSA: type of '{name.name}' is not a monotype"
+        | throw (Strata.DiagnosticModel.fromMessage s!"SSA: type of '{name.name}' is not a monotype")
       let env' := env.insert name.name { ident := freshId, ty := info.ty }
       incrementStat s!"{Stats.renamedVars}"
       return (env', [Statement.init freshId (.forAll [] mty) .nondet smd])
-    | none => throw s!"SSA: variable '{name.name}' not found in environment (havoc)"
+    | none => throw (Strata.DiagnosticModel.fromMessage s!"SSA: variable '{name.name}' not found in environment (havoc)")
   | .cmd (.assert label b amd) =>
     return (env, [Statement.assert label (rewriteExpr env b) amd])
   | .cmd (.assume label b amd) =>
     return (env, [Statement.assume label (rewriteExpr env b) amd])
   | .cmd (.cover label b cmd') =>
     return (env, [Statement.cover label (rewriteExpr env b) cmd'])
-  | .call _ _ _ => throw "SSA: unexpected call command (callElim should have run first)"
+  | .call _ _ _ => throw (Strata.DiagnosticModel.fromMessage "SSA: unexpected call command (callElim should have run first)")
 
 private def collectAllKeys (envs : List Env) : List String :=
   let hs := envs.foldl (fun acc env =>
@@ -149,7 +149,7 @@ def emitJoinMerges (condVar : Expression.Ident)
       let elseId := getIdOr elseInfo preInfo origName
       let ty := getTyOr thenInfo elseInfo preInfo
       let some mty := LTy.toMonoType? ty
-        | throw s!"SSA: type of '{origName}' is not a monotype at join point"
+        | throw (Strata.DiagnosticModel.fromMessage s!"SSA: type of '{origName}' is not a monotype at join point")
       let freshId ← genSSAIdent origName
       let iteExpr : Expression.Expr :=
         Lambda.LExpr.ite () (createFvar condVar) (createFvar thenId) (createFvar elseId)
@@ -176,7 +176,7 @@ def emitNondetJoinHavocs (preEnv thenEnv elseEnv : Env)
     if varChanged thenInfo preInfo || varChanged elseInfo preInfo then
       let ty := getTyOr thenInfo elseInfo preInfo
       let some mty := LTy.toMonoType? ty
-        | throw s!"SSA: type of '{origName}' is not a monotype at nondet join"
+        | throw (Strata.DiagnosticModel.fromMessage s!"SSA: type of '{origName}' is not a monotype at nondet join")
       let freshId ← genSSAIdent origName
       havoces := havoces ++ [Statement.init freshId (.forAll [] mty) .nondet md]
       env := env.insert origName { ident := freshId, ty := ty }
@@ -214,7 +214,7 @@ partial def transformStmt (env : Env) (s : Statement) : CoreTransformM (Env × L
       let (mergedEnv, havoces) ← emitNondetJoinHavocs env thenEnv elseEnv md
       return (mergedEnv, [Stmt.ite .nondet thenStmts' elseStmts' md] ++ havoces)
   | .loop _ _ _ _ _ =>
-    throw "SSA: unexpected loop statement (loopElim should have run first)"
+    throw (Strata.DiagnosticModel.fromMessage "SSA: unexpected loop statement (loopElim should have run first)")
   | .exit label md => return (env, [Stmt.exit label md])
   | .funcDecl decl md => return (env, [Stmt.funcDecl decl md])
   | .typeDecl tc md => return (env, [Stmt.typeDecl tc md])
--- a/Strata/Transform/SSA.lean
+++ b/Strata/Transform/SSA.lean
@@ -1,0 +1,273 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+public import Strata.Languages.Core.PipelinePhase
+
+/-! # SSA (Static Single Assignment) Transformation
+
+Converts Strata Core procedure bodies into SSA form where every variable is
+assigned exactly once via `init`. At if-then-else join points, conditional
+`init` expressions merge divergent variable versions.
+
+Preconditions: runs after `callElim` and `loopElim`.
+SSA is semantics-preserving, so the pipeline phase uses `modelPreserving`.
+-/
+
+namespace Core
+namespace SSA
+
+open Imperative Lambda
+
+public section
+
+/-- Statistics keys tracked by the SSA transformation. -/
+inductive Stats where
+  | renamedVars
+  | joinPointMerges
+
+#derive_prefixed_toString Stats "SSA"
+
+structure VarInfo where
+  ident : Expression.Ident
+  ty    : Expression.Ty
+  deriving Repr, BEq
+
+abbrev Env := Std.HashMap String VarInfo
+
+structure SSAState where
+  counter : Nat := 0
+  env : Env := {}
+  statistics : Statistics := {}
+
+abbrev SSAM := ExceptT String (StateM SSAState)
+
+def freshIdent (baseName : String) : SSAM Expression.Ident := do
+  let st ← get
+  let freshName := s!"{baseName}_{st.counter}"
+  set { st with counter := st.counter + 1 }
+  return ⟨freshName, ()⟩
+
+def bindVar (origName : String) (newIdent : Expression.Ident) (ty : Expression.Ty) : SSAM Unit :=
+  modify fun st => { st with env := st.env.insert origName { ident := newIdent, ty := ty } }
+
+def incrStat (key : String) (n : Nat := 1) : SSAM Unit :=
+  modify fun st => { st with statistics := st.statistics.increment key n }
+
+def mkFvar (id : Expression.Ident) : Expression.Expr :=
+  Lambda.LExpr.fvar ((): ExpressionMetadata) id none
+
+def rewriteExpr (e : Expression.Expr) : SSAM Expression.Expr := do
+  let env := (← get).env
+  if env.isEmpty then return e
+  let sm : Map Expression.Ident Expression.Expr :=
+    env.fold (init := Map.empty) fun acc origName info =>
+      acc.insert ⟨origName, ()⟩ (mkFvar info.ident)
+  return Lambda.LExpr.substFvars e sm
+
+def rewriteExprOrNondet (e : ExprOrNondet Expression) : SSAM (ExprOrNondet Expression) :=
+  match e with
+  | .det expr => return .det (← rewriteExpr expr)
+  | .nondet => return .nondet
+
+def getEnv : SSAM Env := return (← get).env
+
+def setEnv (env : Env) : SSAM Unit :=
+  modify fun st => { st with env := env }
+
+/-- Helper to get the identifier from a VarInfo option, with fallback. -/
+private def getIdOr (info : Option VarInfo) (fallback : Option VarInfo) (origName : String)
+    : Expression.Ident :=
+  match info with
+  | some i => i.ident
+  | none => match fallback with
+    | some i => i.ident
+    | none => ⟨origName, ()⟩
+
+/-- Helper to get the type from multiple VarInfo options. -/
+private def getTyOr (a b c : Option VarInfo) : Expression.Ty :=
+  match a with
+  | some i => i.ty
+  | none => match b with
+    | some i => i.ty
+    | none => match c with
+      | some i => i.ty
+      | none => .forAll [] .bool
+
+/-- Check if a variable changed between two environments. -/
+private def varChanged (info pre : Option VarInfo) : Bool :=
+  match info, pre with
+  | some t, some p => t.ident != p.ident
+  | some _, none => true
+  | none, _ => false
+
+def transformCmd (cmd : Command) : SSAM (List Statement) := do
+  match cmd with
+  | .cmd (.init name ty rhs imd) =>
+    let rhs' ← rewriteExprOrNondet rhs
+    let freshId ← freshIdent name.name
+    bindVar name.name freshId ty
+    incrStat s!"{Stats.renamedVars}"
+    return [Statement.init freshId ty rhs' imd]
+  | .cmd (.set name (.det expr) smd) =>
+    let expr' ← rewriteExpr expr
+    match (← get).env.get? name.name with
+    | some info =>
+      let freshId ← freshIdent name.name
+      let some mty := LTy.toMonoType? info.ty
+        | throw s!"SSA: type of '{name.name}' is not a monotype"
+      bindVar name.name freshId info.ty
+      incrStat s!"{Stats.renamedVars}"
+      return [Statement.init freshId (.forAll [] mty) (.det expr') smd]
+    | none => throw s!"SSA: variable '{name.name}' not found in environment (set)"
+  | .cmd (.set name .nondet smd) =>
+    match (← get).env.get? name.name with
+    | some info =>
+      let freshId ← freshIdent name.name
+      let some mty := LTy.toMonoType? info.ty
+        | throw s!"SSA: type of '{name.name}' is not a monotype"
+      bindVar name.name freshId info.ty
+      incrStat s!"{Stats.renamedVars}"
+      return [Statement.init freshId (.forAll [] mty) .nondet smd]
+    | none => throw s!"SSA: variable '{name.name}' not found in environment (havoc)"
+  | .cmd (.assert label b amd) => return [Statement.assert label (← rewriteExpr b) amd]
+  | .cmd (.assume label b amd) => return [Statement.assume label (← rewriteExpr b) amd]
+  | .cmd (.cover label b cmd') => return [Statement.cover label (← rewriteExpr b) cmd']
+  | .call _ _ _ => throw "SSA: unexpected call command (callElim should have run first)"
+
+private def collectAllKeys (envs : List Env) : List String :=
+  let hs := envs.foldl (fun acc env =>
+    env.fold (init := acc) fun acc k _ => acc.insert k)
+    (Std.HashSet.emptyWithCapacity (α := String))
+  hs.toList
+
+def emitJoinMerges (condVar : Expression.Ident)
+    (preEnv thenEnv elseEnv : Env)
+    (md : Imperative.MetaData Expression) : SSAM (List Statement) := do
+  let allKeys := collectAllKeys [preEnv, thenEnv, elseEnv]
+  let mut merges : List Statement := []
+  for origName in allKeys do
+    let preInfo := preEnv.get? origName
+    let thenInfo := thenEnv.get? origName
+    let elseInfo := elseEnv.get? origName
+    if varChanged thenInfo preInfo || varChanged elseInfo preInfo then
+      let thenId := getIdOr thenInfo preInfo origName
+      let elseId := getIdOr elseInfo preInfo origName
+      let ty := getTyOr thenInfo elseInfo preInfo
+      let some mty := LTy.toMonoType? ty
+        | throw s!"SSA: type of '{origName}' is not a monotype at join point"
+      let freshId ← freshIdent origName
+      let iteExpr : Expression.Expr :=
+        Lambda.LExpr.ite () (mkFvar condVar) (mkFvar thenId) (mkFvar elseId)
+      merges := merges ++ [Statement.init freshId (.forAll [] mty) (.det iteExpr) md]
+      bindVar origName freshId ty
+      incrStat s!"{Stats.joinPointMerges}"
+  return merges
+
+def emitNondetJoinHavocs (preEnv thenEnv elseEnv : Env)
+    (md : Imperative.MetaData Expression) : SSAM (List Statement) := do
+  let allKeys := collectAllKeys [preEnv, thenEnv, elseEnv]
+  let mut havoces : List Statement := []
+  for origName in allKeys do
+    let preInfo := preEnv.get? origName
+    let thenInfo := thenEnv.get? origName
+    let elseInfo := elseEnv.get? origName
+    if varChanged thenInfo preInfo || varChanged elseInfo preInfo then
+      let ty := getTyOr thenInfo elseInfo preInfo
+      let some mty := LTy.toMonoType? ty
+        | throw s!"SSA: type of '{origName}' is not a monotype at nondet join"
+      let freshId ← freshIdent origName
+      havoces := havoces ++ [Statement.init freshId (.forAll [] mty) .nondet md]
+      bindVar origName freshId ty
+      incrStat s!"{Stats.joinPointMerges}"
+  return havoces
+
+def initEnvFromProcedure (proc : Procedure) : SSAM Unit := do
+  let _ ← (proc.header.inputs : List _).mapM fun (id, mty) =>
+    bindVar id.name id (.forAll [] mty)
+  let _ ← (proc.header.outputs : List _).mapM fun (id, mty) =>
+    bindVar id.name id (.forAll [] mty)
+
+mutual
+partial def transformStmt (s : Statement) : SSAM (List Statement) := do
+  match s with
+  | .cmd cmd => transformCmd cmd
+  | .block label stmts md =>
+    let stmts' ← transformBlock stmts
+    return [Stmt.block label stmts' md]
+  | .ite cond thenStmts elseStmts md =>
+    match cond with
+    | .det condExpr =>
+      let condExpr' ← rewriteExpr condExpr
+      let condVar ← freshIdent "$ssa_cond"
+      let condInit := Statement.init condVar (.forAll [] LMonoTy.bool) (.det condExpr') md
+      let preEnv ← getEnv
+      let thenStmts' ← transformBlock thenStmts
+      let thenEnv ← getEnv
+      setEnv preEnv
+      let elseStmts' ← transformBlock elseStmts
+      let elseEnv ← getEnv
+      let merges ← emitJoinMerges condVar preEnv thenEnv elseEnv md
+      let iteStmt := Stmt.ite (.det (mkFvar condVar)) thenStmts' elseStmts' md
+      return [condInit, iteStmt] ++ merges
+    | .nondet =>
+      let preEnv ← getEnv
+      let thenStmts' ← transformBlock thenStmts
+      let thenEnv ← getEnv
+      setEnv preEnv
+      let elseStmts' ← transformBlock elseStmts
+      let elseEnv ← getEnv
+      let havoces ← emitNondetJoinHavocs preEnv thenEnv elseEnv md
+      return [Stmt.ite .nondet thenStmts' elseStmts' md] ++ havoces
+  | .loop _ _ _ _ _ =>
+    throw "SSA: unexpected loop statement (loopElim should have run first)"
+  | .exit label md => return [Stmt.exit label md]
+  | .funcDecl decl md => return [Stmt.funcDecl decl md]
+  | .typeDecl tc md => return [Stmt.typeDecl tc md]
+
+partial def transformBlock (stmts : List Statement) : SSAM (List Statement) := do
+  let mut result : List Statement := []
+  for s in stmts do
+    let stmts' ← transformStmt s
+    result := result ++ stmts'
+  return result
+end
+
+def transformProcedure (proc : Procedure) : SSAM Procedure := do
+  if proc.body.isEmpty then return proc
+  initEnvFromProcedure proc
+  let body' ← transformBlock proc.body
+  return { proc with body := body' }
+
+def ssaTransform (p : Program) : Except String (Program × Statistics) := do
+  let decls ← p.decls.mapM fun d =>
+    match d with
+    | .proc proc md =>
+      let initState : SSAState := {}
+      let (result, finalState) := StateT.run (transformProcedure proc) initState
+      match result with
+      | .ok proc' => .ok (.proc proc' md, finalState.statistics)
+      | .error e => .error s!"SSA error in procedure '{proc.header.name}': {e}"
+    | other => .ok (other, {})
+  let (decls', stats) := decls.foldl (fun (acc, stats) (d, s) =>
+    (acc ++ [d], stats.merge s)) ([], {})
+  .ok ({ decls := decls' }, stats)
+
+def ssaTransform' (p : Program) : Transform.CoreTransformM (Bool × Program) := do
+  match ssaTransform p with
+  | .ok (p', stats) =>
+    modify fun σ => { σ with statistics := σ.statistics.merge stats }
+    return (true, p')
+  | .error e => throw e
+
+def ssaPipelinePhase : PipelinePhase where
+  transform := ssaTransform'
+  phase.name := "SSA"
+  phase.getValidation _ := .modelPreserving
+
+end -- public section
+end SSA
+end Core
--- a/Strata/Transform/SSA.lean
+++ b/Strata/Transform/SSA.lean
@@ -21,6 +21,7 @@ namespace Core
 namespace SSA
 
 open Imperative Lambda
+open Core.Transform
 
 public section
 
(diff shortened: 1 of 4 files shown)


Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,257 @@
...
+
+/-- Helper to get the type from multiple VarInfo options. -/
+private def getTyOr (a b c : Option VarInfo) : Expression.Ty :=
+  match a with
+  | some i => i.ty
+  | none => match b with
+    | some i => i.ty
+    | none => match c with
+      | some i => i.ty
+      | none => .forAll [] .bool
```
aqjune-aws: Can we make `getTyOr` return `Option Expression.Ty` and return .none in this case? The caller of `getTyOr` can check whether the result is none and either throw with a informative message or fallback to `.forAll [] .bool`.

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,257 @@
...
+  | .nondet => .nondet
+
+/-- Helper to get the identifier from a VarInfo option, with fallback. -/
+private def getIdOr (info fallback : Option VarInfo) (origName : String)
+    : Expression.Ident :=
+  match info with
+  | some i => i.ident
+  | none => match fallback with
+    | some i => i.ident
+    | none => ⟨origName, ()⟩
```
aqjune-aws: This can be simplified to:
```
((info <|> fallback).map (·.ident)).getD ⟨origName, ()⟩
```
`<|>` is a notation for `Option.or`. :) `·.ident` means `fun x => x.ident`.

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,257 @@
...
+    let sm : Map Expression.Ident Expression.Expr :=
+      env.fold (init := Map.empty) fun acc origName info =>
+        acc.insert ⟨origName, ()⟩ (createFvar info.ident)
+    Lambda.LExpr.substFvars e sm
+
+/-- Rewrite free variables in an `ExprOrNondet`. -/
+def rewriteExprOrNondet (env : Env) (e : ExprOrNondet Expression) : ExprOrNondet Expression :=
+  match e with
+  | .det expr => .det (rewriteExpr env expr)
+  | .nondet => .nondet
```
aqjune-aws: You can use
```
  e.map (rewriteExpr env ·)
```

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,257 @@
...
+      let condInit := Statement.init condVar (.forAll [] LMonoTy.bool) (.det condExpr') md
+      let (thenEnv, thenStmts') ← transformBlock env thenStmts
+      let (elseEnv, elseStmts') ← transformBlock env elseStmts
+      let (mergedEnv, merges) ← emitJoinMerges condVar env thenEnv elseEnv md
+      let iteStmt := Stmt.ite (.det (createFvar condVar)) thenStmts' elseStmts' md
+      return (mergedEnv, [condInit, iteStmt] ++ merges)
+    | .nondet =>
+      let (thenEnv, thenStmts') ← transformBlock env thenStmts
+      let (elseEnv, elseStmts') ← transformBlock env elseStmts
+      let (mergedEnv, havoces) ← emitNondetJoinHavocs env thenEnv elseEnv md
```
aqjune-aws: I think this is too aggressive - no? We can still do `emitJoinMerges`, but with `condVar` initialized to `Statement.init condVar (.forAll [] LMonoTy.bool) .nondet`. This will be more precise than the current version. Also we can only leave one implementation `emitJoinMerges` and drop `emitNondetJoinHavocs` :)

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,257 @@
...
+      return (mergedEnv, [Stmt.ite .nondet thenStmts' elseStmts' md] ++ havoces)
+  | .loop _ _ _ _ _ =>
+    throw (Strata.DiagnosticModel.fromMessage "SSA: unexpected loop statement (loopElim should have run first)")
+  | .exit label md => return (env, [Stmt.exit label md])
```
aqjune-aws: I think `.exit` is complicated - let's throw as we do in `.loop`.

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,257 @@
...
+  let mut havoces : List Statement := []
+  for origName in allKeys do
+    let preInfo := preEnv.get? origName
+    let thenInfo := thenEnv.get? origName
+    let elseInfo := elseEnv.get? origName
+    if varChanged thenInfo preInfo || varChanged elseInfo preInfo then
+      let ty := getTyOr thenInfo elseInfo preInfo
+      let some mty := LTy.toMonoType? ty
+        | throw (Strata.DiagnosticModel.fromMessage s!"SSA: type of '{origName}' is not a monotype at nondet join")
+      let freshId ← genSSAIdent origName
```
aqjune-aws: Can we have `"ssa_phi_{id}"` as a prefix of the var name in this case? :)

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,257 @@
...
+    | .proc proc md => return .proc (← transformProcedure proc) md
+    | other => return other
+  return (true, { decls := decls })
+
+/-- SSA pipeline phase: converts procedure bodies to SSA form.
+    SSA is semantics-preserving, so models are preserved. -/
+def ssaPipelinePhase : PipelinePhase where
+  transform := ssaTransform
+  phase.name := "SSA"
+  phase.getValidation _ := .modelPreserving
```
aqjune-aws, line 256: At the moment this isn't `.modelPreserving` because `emitNondetJoinHavocs` overapproximates it. Converting it to using `emitJoinMerges` will make `.modelPreserving` OK.

aqjune-aws (changes requested):

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,240 @@
...
+public import Strata.Languages.Core.PipelinePhase
+
+/-! # SSA (Static Single Assignment) Transformation
+
+Converts Strata Core procedure bodies into SSA form where every variable is
+assigned exactly once via `init`. At if-then-else join points, conditional
+`init` expressions merge divergent variable versions.
+
+Preconditions: runs after `callElim` and `loopElim`.
+SSA is semantics-preserving, so the pipeline phase uses `modelPreserving`.
```
tautschnig: This doc comment asserts `modelPreserving`, but the current transform isn't — see the inline on `transformCmd` for the concrete counterexample (output `y` never gets written in the SSA body, so `ensures y == x + 1` flips from ✅ to ❌). Either the docstring needs to lose the claim or the transform needs to preserve final assignments to outputs.

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,240 @@
...
+        acc.insert ⟨origName, ()⟩ (createFvar info.ident)
+    Lambda.LExpr.substFvars e sm
+
+/-- Rewrite free variables in an `ExprOrNondet`. -/
+def rewriteExprOrNondet (env : Env) (e : ExprOrNondet Expression) : ExprOrNondet Expression :=
+  e.map (rewriteExpr env ·)
+
+/-- Get the identifier from VarInfo options, with fallback to original name. -/
+private def getIdOr (info fallback : Option VarInfo) (origName : String)
+    : Expression.Ident :=
```
tautschnig: The `⟨origName, ()⟩` fallback is dangerous: it produces an identifier for a name that has never been declared at this scope, which becomes a free-variable reference in the emitted phi. This is how Bug 2 (see review body) manifests — a variable declared only inside the then-branch gets phi-merged with the else-branch's missing info, and the emitted phi says `ite c ssa_x_1 x` where `x` is the original name and was never declared.

I'd prefer this function to return `Option Expression.Ident` and let callers throw a diagnostic when a phi is requested for a variable that wasn't present in `preEnv` (or handle the pre-ITE absence case with an explicit havoc / skip-the-merge). A `dbg_trace`/`panic!`-style unreachability marker would at least surface the bug during testing instead of emitting a silently-invalid program.

Suggested refactor sketch:

```lean
private def phiEntry (origName : String) (preEnv thenEnv elseEnv : Env)
    : Option (Expression.Ident × Expression.Ident × Expression.Ty) := do
  let preInfo := preEnv.get? origName
  let thenInfo := thenEnv.get? origName
  let elseInfo := elseEnv.get? origName
  -- Only merge variables that were in scope before the ITE.
  let pre ← preInfo
  if !varChanged thenInfo preInfo && !varChanged elseInfo preInfo then none
  else
    let thenId := (thenInfo.map (·.ident)).getD pre.ident
    let elseId := (elseInfo.map (·.ident)).getD pre.ident
    some (thenId, elseId, pre.ty)
```

This makes the scoping invariant ("only merge what was in scope before") explicit and eliminates the unsafe string-fallback branch.

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,240 @@
...
+/-- Transform a single command in SSA form. Returns updated env and new statements. -/
+def transformCmd (env : Env) (cmd : Command) : CoreTransformM (Env × List Statement) := do
+  match cmd with
+  | .cmd (.init name ty rhs imd) =>
+    let rhs' := rewriteExprOrNondet env rhs
+    let freshId ← genSSAIdent name.name
+    let env' := env.insert name.name { ident := freshId, ty := ty }
+    incrementStat s!"{Stats.renamedVars}"
+    return (env', [Statement.init freshId ty rhs' imd])
+  | .cmd (.set name (.det expr) smd) =>
```
tautschnig, line 102: **Correctness bug (blocking, see review body).** `.set` on an output parameter gets rewritten to `var ssa_r_N : int := …`, which is a *local* declaration rather than an assignment to the declared output `r`. As a consequence the procedure's output is never written and every `ensures (r == …)` that held before SSA is violated after.

Repro:

```
procedure inc(x : int, out y : int) spec { ensures (y == x + 1); }
{ y := x + 1; };
```

`verify pgm` → `inc_ensures_0 : ✅ pass` on the original; `verify (runTransforms pgm [.ssa])` → `inc_ensures_0 : ❌ fail` on the transformed body.

One fix: track which env entries correspond to the procedure's declared outputs (and `inout`s), then at the end of `transformProcedure` emit a `set` back to each such output from its final SSA name:

```lean
for (outName, _) in proc.header.outputs do
  match env.get? outName.name with
  | some info when info.ident != outName =>
    result := result ++ [Statement.set outName (.det (createFvar info.ident)) md]
  | _ => pure ()
```

Alternatively, keep `.set` on outputs as-is (don't rewrite the target, only the RHS) — that weakens SSA but preserves semantics.

Either way, please also add a VC-outcome regression test (see the companion inline on the test file).

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,240 @@
...
+    (preEnv thenEnv elseEnv : Env)
+    (md : Imperative.MetaData Expression) : CoreTransformM (Env × List Statement) := do
+  let allKeys := collectAllKeys [preEnv, thenEnv, elseEnv]
+  let mut env := thenEnv
+  let mut merges : List Statement := []
+  for origName in allKeys do
+    let preInfo := preEnv.get? origName
+    let thenInfo := thenEnv.get? origName
+    let elseInfo := elseEnv.get? origName
+    if varChanged thenInfo preInfo || varChanged elseInfo preInfo then
```
tautschnig: The condition `varChanged thenInfo preInfo || varChanged elseInfo preInfo` will fire for any variable that appears in *either* branch's env — including variables that were declared inside a single branch and were never in scope before the ITE (`preInfo = none`, one branch = `some`).

In that case:

- `thenId`/`elseId` point at SSA names declared inside the branch, which are not visible after the ITE (Strata Core blocks are lexically scoped; see `translateBlock` in `DDMTransform/Translate.lean:1332`).
- For the branch that did not declare the variable, `getIdOr` falls back to the raw pre-SSA name, producing an undeclared free variable.

Concretely, `if (c) { var x : int := 42; }` plus a later statement produces `var ssa_phi_x_2 : int := if ssa_$ssa_cond_0 then ssa_x_1 else x;` where `ssa_x_1` is block-local and `x` was never declared. (Strata's current type checker accepts this, but that's its own bug — the program is ill-scoped.)

Please restrict phi generation to variables that are present in `preEnv` (i.e. were in scope before the ITE). See the `phiEntry` sketch in the inline on `getIdOr`. If you also want to SSA-ize branch-locally-declared vars, that's a larger design discussion — it would require lifting their declarations out of the blocks, not just referencing them.

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,240 @@
...
+  match s with
+  | .cmd cmd => transformCmd env cmd
+  | .block label stmts md =>
+    let (env', stmts') ← transformBlock env stmts
+    return (env', [Stmt.block label stmts' md])
+  | .ite cond thenStmts elseStmts md =>
+    match cond with
+    | .det condExpr =>
+      let condExpr' := rewriteExpr env condExpr
+      let condVar ← genSSAIdent "$ssa_cond"
```
tautschnig: Nit: `genSSAIdent "$ssa_cond"` runs through `ssaVarPrefix` which prepends `ssa_`, so you end up with `ssa_$ssa_cond_N` — double-prefix and the `$` becomes `ssa_$`, which is visually confusing in printed programs (e.g. `var ssa_$ssa_cond_0 : bool := c;`). Consider `genSSAIdent "cond"` (→ `ssa_cond_N`) or introduce a separate generator that doesn't re-prefix. Same applies to `"$ssa_nondet_cond"` below.

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,240 @@
...
+  let mut result : List Statement := []
+  for s in stmts do
+    let (env', stmts') ← transformStmt curEnv s
+    curEnv := env'
+    result := result ++ stmts'
+  return (curEnv, result)
+end
+
+/-- Transform a single procedure into SSA form. -/
+def transformProcedure (proc : Procedure) : CoreTransformM Procedure := do
```
tautschnig, line 227: After the body is transformed, the outputs declared in `proc.header.outputs` have likely been renamed to `ssa_<out>_N` in the env but are never written back to their original names. This is where the output-parameter fix (see inline on `transformCmd`) needs to land: after `let (_, body') ← transformBlock env proc.body`, append a final `set outName (.det (createFvar env[outName].ident))` for every output whose env ident has diverged from its original.

It would also be worth documenting why `initEnvFromProcedure` treats inputs and outputs symmetrically — the symmetry is fine, but today the comment doesn't spell out that outputs depend on this final re-assignment pass for semantic preservation.

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,240 @@
...
+    | .proc proc md => return .proc (← transformProcedure proc) md
+    | other => return other
+  return (true, { decls := decls })
+
+/-- SSA pipeline phase: converts procedure bodies to SSA form.
+    SSA is semantics-preserving, so models are preserved. -/
+def ssaPipelinePhase : PipelinePhase where
+  transform := ssaTransform
+  phase.name := "SSA"
+  phase.getValidation _ := .modelPreserving
```
tautschnig, line 256: Given Bug 1 (output parameters), this should not claim `.modelPreserving` yet. Suggestion: downgrade to `.unchecked` (or similar — whichever matches the PipelinePhase API's vocabulary for transforms whose correctness is unverified) until the output-preservation fix and at least a VC-outcome round-trip test are in place. Re-upgrade to `.modelPreserving` once the transform is actually semantics-preserving.

It would also be a good place to cite the proposed theorems in a comment so future readers know what `.modelPreserving` would mean formally for this transform.

StrataTest/Transform/SSA.lean
```diff
@@ -0,0 +1,322 @@
...
+};
+#end
+
+/-- info: true -/
+#guard_msgs in
+#eval do
+  let pgm := translate SSATest1
+  let result := runSSA pgm
+  -- After SSA, the program should be different (variables renamed)
+  return toString (Std.format result) != toString (Std.format pgm)
```
tautschnig, line 48: These 14 tests all use the oracle `toString (Std.format result) != toString (Std.format pgm)`, which only checks that the transform did *something*. A transform that deleted every procedure would pass these tests; so would the two correctness bugs I reported (see review body).

Please add at least two stronger oracles:

1. **VC-outcome round-trip.** Pick a small representative program with `ensures` that passes on the original, run `.ssa`, and check that the ensures still passes. This catches Bug 1 (output-parameter loss). Sketch:

    ```lean
    def IncRoundTrip := #strata
    program Core;
    procedure inc(x : int, out y : int) spec { ensures (y == x + 1); } { y := x + 1; };
    #end

    /-- info: inc_ensures_0 : ✅ pass -/
    #guard_msgs in
    #eval do
      let pgm := translate IncRoundTrip
      let .ok ssaPgm := Core.runTransforms pgm [.ssa] | return "ssa error"
      -- verify with proceduresToVerify := some ["inc"] and print the single VC outcome
      ...
    ```

2. **No undefined identifiers.** Walk the transformed body collecting free-variable references and declared identifiers; assert that every free-variable reference resolves to something declared in an enclosing scope. This catches Bug 2 (phi merges referring to branch-local / never-declared names).

Tests #2, #5, #6, #7 (`SSATestNestedIte`, the pipeline test, the one-branch-only tests) are particularly good targets for both oracles.

tautschnig (changes requested):
### Bug 1 — output parameters lose their final assignment

In `transformCmd`, the `.set r …` case generates `var ssa_r_N : int := …` instead of preserving the write to the output parameter `r`. After SSA, the declared output is never assigned.

Minimal repro:

```
procedure inc(x : int, out y : int)
spec { ensures (y == x + 1); }
{ y := x + 1; };
```

Before SSA: `inc_ensures_0 : ✅ pass`. After `[.ssa]`: body becomes `var ssa_y_0 : int := x + 1;` — `y` is never written — and `inc_ensures_0 : ❌ fail`.

See the inline comment on `transformCmd`. The fix is probably to emit a final `r := ssa_r_N` for every output whose env ident diverged from its original ident, at the end of each procedure body (and similarly for `inout` parameters).

### Bug 2 — phi merges can reference out-of-scope or undefined names

Strata Core blocks introduce lexical scope (see `Translate.lean:1332` — `translateBlock` does not propagate `boundVars` out of the block). `emitJoinMerges` emits phi inits at the join point that reference the branches' in-block SSA names, which are no longer in scope after the ITE. And when a variable is declared only in one branch (and wasn't in `preEnv`), the other branch's `getIdOr` falls back to `⟨origName, ()⟩` — the raw pre-SSA name, which may never have been declared at all.

Minimal repro:

```
procedure f(c : bool, out r : int) {
  if (c) { var x : int := 42; }
  r := 0;
};
```

SSA output:

```
  var ssa_$ssa_cond_0 : bool := c;
  if (ssa_$ssa_cond_0) { var ssa_x_1 : int := 42; }
  var ssa_phi_x_2 : int := if ssa_$ssa_cond_0 then ssa_x_1 else x;
  var ssa_r_3 : int := 0;
```

Both `ssa_x_1` (declared inside the then-block) and `x` (never declared) appear in the outer scope. Strata's current type checker happens to accept this (that's probably its own bug), but it's clearly wrong. The logic should either skip join-merges for variables that weren't in `preEnv`, or lift the declarations themselves out of their branches to procedure scope before emitting the phi.

See the inline comment on `emitJoinMerges` / `getIdOr`.

### Pipeline validation is wrong

Given Bug 1, the `.modelPreserving` annotation on `ssaPipelinePhase` is false: there exist programs where SSA changes verification outcomes from pass to fail. Until Bug 1 is fixed, this phase shouldn't claim model preservation. Consider `.unchecked` or a new `.transformModels` validation with explicit documentation — or `.notImplemented` while the transform is still experimental / opt-in.

### Test coverage is too weak to catch either bug

All 16 tests use one of three weak oracles — see inline on `StrataTest/Transform/SSA.lean`. A single VC-outcome round-trip test would have caught Bug 1; a simple "no undefined identifiers" check on the output would have caught Bug 2.

## Refactoring notes

- `getIdOr`'s fallback to `⟨origName, ()⟩` silently produces potentially-undefined references. It should either be removed (in favor of throwing a diagnostic at the call site) or, if a fallback really is needed, asserted unreachable via a `panic!`-replacement that reports the bug.
- `genSSAIdent "$ssa_cond"` yields `ssa_$ssa_cond_N`, a double-`ssa_` / double-`$` name that is ugly to read in traces. Pick `"cond"` (→ `ssa_cond_N`) or bypass the `ssa_` prefix for these internal names.
- `initEnvFromProcedure` seeds the env for outputs identically to inputs — document that outputs require a final re-assignment (once Bug 1 is fixed) so a future reader doesn't try to "simplify" the initialization.
- `collectAllKeys : List Env → List String` is only ever called with a 3-element list; either specialize it or keep it general and add a 0/1/2/empty test.
- `rewriteExpr` shortcircuits on `env.isEmpty`, but `substFvars` with an empty map is already a no-op; the check is harmless but redundant.
- The `rewriteExpr`/`rewriteExprOrNondet` split plus `getIdOr`/`getTyOr`/`varChanged` makes `emitJoinMerges` hard to follow — consider extracting a `phiEntry : String → Env → Env → Env → Option (Ident × Ident × Ty)` that returns `none` for vars that don't need a phi and handles the scoping / lookup concerns in one place.

## Theorems / tests worth locking in

All of these would either prevent re-occurrence of the two bugs or push the transform toward the `.modelPreserving` guarantee:

1. **Single-assignment**: for all `p : Program`, every procedure body produced by `ssaTransform` has distinct `init` identifiers (modulo nested blocks). This is the defining invariant of SSA form.
2. **Scoping**: for all `p : Program`, every free-variable reference in the body produced by `ssaTransform p` resolves to a `init`-declared identifier in an enclosing scope. (Would have failed for Bug 2.)
3. **Output preservation**: for all `p : Program` and every procedure `P` with output `o`, the final value of `o` under the transformed body equals the final value of `o` under the original body. (Would have failed for Bug 1.)
4. **VC-outcome preservation**: a property-test (plausible / random Core procedures) that `verify p` and `verify (runTransforms p [.ssa])` produce identical VC outcomes per label.
5. **Well-typedness preservation**: a simpler prerequisite to #4 — if the original type-checks, so does the SSA output.

#1 and #2 should be tractable as pencil-and-paper proofs over the structure of `transformStmt` / `transformCmd`. #3 and #4 are harder but the payoff is high: they would justify the `.modelPreserving` claim that currently isn't true.

aqjune-aws (approved):

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,260 @@
...
+      -- variable selects which branch's values are used via the phi.
+      let condExpr' := rewriteExpr env condExpr
+      let condVar ← genSSAIdent "cond"
+      let condInit := Statement.init condVar (.forAll [] LMonoTy.bool) (.det condExpr') md
+      let (thenEnv, thenStmts') ← transformBlock env thenStmts
+      let (elseEnv, elseStmts') ← transformBlock env elseStmts
+      let (mergedEnv, merges) ← emitJoinMerges condVar env thenEnv elseEnv md
+      -- Flatten: condition init, then-branch stmts, else-branch stmts, phi merges.
+      -- All at the same scope level so phi references are valid.
+      return (mergedEnv, [condInit] ++ thenStmts' ++ elseStmts' ++ merges)
```
tautschnig, line 198: **Soundness bug: branch-specific asserts lose their path-condition guard.** I reproduced this on the current branch:

```lean
procedure check_branch_assert(x : int) {
  if (x > 0) { assert [pos]: (x > 0); }
};
```

Original: verifies. After `--pass ssa`:

```
procedure check_branch_assert (x : int)
{
  var ssa_cond_0 : bool := x > 0;
  assert [pos]: x > 0;
};
```

SSA'd: **fails** (`x = -5` is a spurious counterexample). The `if` is flattened away but the `assert` isn't guarded by `ssa_cond_0`, so the verifier checks `x > 0` unconditionally.

Two standard fixes:

1. **Guard branch asserts at flattening time.** In `transformBlock` (called with the branch env), whenever an assert is encountered inside a then/else branch, rewrite it as `assert (condVar → original_assert)`. This preserves VC semantics while still flattening.
2. **Don't flatten — keep the `.ite` intact and only add phi inits at the join.** Emit the ITE as `.ite cond' (thenStmts' ++ localPhis) (elseStmts' ++ localPhis) md ++ outerPhis`. The phis for variables changed in only one branch use the `preEnv` value on the other side, which is what `phiEntry` already computes. This is the standard SSA-with-control-flow representation.

Option (2) is less invasive — the flattening was the source of the bug. Option (1) is simpler to implement on top of the current code.

Until one of these lands, please downgrade `.modelPreserving` at line 256 to `.modelToValidate` so the solver validates models it produces — see inline comment there.

Also add the reproducer to `SSACorrectnessTests`:

```lean
def SSABranchAssertRoundTrip :=
#strata
program Core;
procedure check_branch_assert(x : int) {
  if (x > 0) { assert [pos]: (x > 0); }
};
#end

/-- info: true -/
#guard_msgs in
#eval show IO Bool from do
  let pgm := translate SSABranchAssertRoundTrip
  let origResults ← … -- verify, should pass
  let .ok ssaPgm := Core.runTransforms pgm [.ssa] | return false
  let ssaResults ← … -- verify ssaPgm, should ALSO pass (currently fails)
  return origPass && ssaPass
```

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,260 @@
...
+  let outputAssigns := emitOutputAssignments proc finalEnv MetaData.empty
+  return { proc with body := body' ++ outputAssigns }
+
+/-- SSA transformation on an entire program, using CoreTransformM. -/
+def ssaTransform (p : Program) : CoreTransformM (Bool × Program) := do
+  let decls ← p.decls.mapM fun d =>
+    match d with
+    | .proc proc md => return .proc (← transformProcedure proc) md
+    | other => return other
+  return (true, { decls := decls })
```
tautschnig, line 241: Nit: `return (true, { decls := decls })` always reports `changed = true`, even when `p.decls` contains no procedures (or every procedure had an empty body, short-circuited at `:228`). Downstream consumers that run passes to a fixpoint will loop forever on SSA applied to an axiom-only program.

```suggestion
  -- Track whether any procedure was actually transformed. An empty / proc-less
  -- program is a no-op; reporting `changed = true` would force fixpoint loops
  -- in the transform driver to iterate indefinitely.
  let mut anyChanged := false
  let decls ← p.decls.mapM fun d =>
    match d with
    | .proc proc md =>
      if proc.body.isEmpty then pure (.proc proc md)
      else do
        anyChanged := true
        return .proc (← transformProcedure proc) md
    | other => return other
  return (anyChanged, { decls := decls })
```

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,260 @@
...
+    procedure's contract semantics. -/
+def emitOutputAssignments (proc : Procedure) (finalEnv : Env)
+    (md : Imperative.MetaData Expression) : List Statement :=
+  (proc.header.outputs : List _).filterMap fun (outId, _) =>
+    match finalEnv.get? outId.name with
+    | some info =>
+      if info.ident != outId then
+        some (Statement.set outId (createFvar info.ident) md)
+      else none
+    | none => none
```
tautschnig, line 175: Worth documenting the intentional SSA-invariant exception here: after `emitOutputAssignments`, the output parameter gets assigned via `set outId := ssa_y_N` *in addition to* its `var ssa_y_N : int := …` init. That is a genuine double-assignment on `outId` — once implicitly at the procedure boundary (when we seed it in `initEnvFromProcedure`), once here.

A reader checking "does every variable have exactly one init?" would find this confusing.

```suggestion
/-- Emit final `set` statements to write SSA-renamed values back to the
    original output/inout parameter identifiers. This preserves the
    procedure's contract semantics.

    Note (SSA-invariant exception): the output parameter's original name is
    reassigned here via `set` even though SSA nominally permits exactly one
    assignment per name. This is deliberate: the procedure contract references
    outputs by their declared name, so the SSA form has to write the final
    SSA value back before returning. All *local* variables remain strictly
    single-assignment. -/
def emitOutputAssignments (proc : Procedure) (finalEnv : Env)
    (md : Imperative.MetaData Expression) : List Statement :=
  (proc.header.outputs : List _).filterMap fun (outId, _) =>
    match finalEnv.get? outId.name with
    | some info =>
      -- Skip when the output was never assigned in the body (its SSA id
      -- equals the original seed from `initEnvFromProcedure`).
      if info.ident != outId then
        some (Statement.set outId (createFvar info.ident) md)
      else none
    | none => none
```

Strata/Transform/SSA.lean
```diff
@@ -0,0 +1,260 @@
...
+    annotation is justified because:
+    - Every variable is assigned exactly once (SSA invariant)
+    - Output parameters receive their final SSA value via explicit set
+    - Phi merges only reference variables in scope before the ITE
+
+    TODO: formal proof of single-assignment, scoping, and output preservation. -/
+def ssaPipelinePhase : PipelinePhase where
+  transform := ssaTransform
+  phase.name := "SSA"
+  phase.getValidation _ := .modelPreserving
```
tautschnig, line 256: **Downgrade to `.modelToValidate` until the branch-assert soundness bug is fixed.** Per the inline on `:184-198` and the top-level body: the current `.ite` flattening produces spurious counterexamples (verified on this branch: the `check_branch_assert` reproducer fails in SSA form where it passes in original). `.modelPreserving` means "sat results are sound" — which is exactly what this flattening breaks.

```suggestion
/-- SSA pipeline phase: converts procedure bodies to SSA form.

    Correctness status (NOT YET `.modelPreserving`): the current `.ite`
    handling flattens branches without guarding branch-specific asserts by
    the branch condition. This introduces spurious counterexamples for
    programs like `if c then assert b`, where the SSA form checks `b`
    unconditionally even though the original only reaches the assert on
    the `c = true` path. See the TODO on `transformStmt`'s `.ite` arm.

    Until the guard-or-preserve fix lands, models must be validated — a
    counterexample reported by the solver after SSA may be spurious.

    TODO: once the fix lands, flip back to `.modelPreserving` and file a
    formal proof of model-preservation. -/
def ssaPipelinePhase : PipelinePhase where
  transform := ssaTransform
  phase.name := "SSA"
  phase.getValidation _ := .modelToValidate (fun _ => false)
```

(The concrete validator `fun _ => false` conservatively rejects every model — safe default. A sharper validator that re-runs the original program on the candidate counterexample is the follow-up work.)

StrataTest/Transform/SSA.lean
```diff
@@ -0,0 +1,423 @@
...
+#guard_msgs in
+#eval do
+  let pgm := translate SSAScopingTest
+  let result := runSSA pgm
+  let s := toString (Std.format result)
+  -- The output should NOT contain "ssa_phi_x" since x was only declared
+  -- inside the then-branch and wasn't in scope before the ITE.
+  return (s.splitOn "ssa_phi_x").length == 1
+
+end SSACorrectnessTests
```
tautschnig, line 423: Two gaps in the correctness suite:

1. **Branch-specific asserts.** Add the reproducer from my review body — `if c then assert b`-shaped procedures — so the soundness bug on `SSA.lean:184-198` is caught by the test suite:

```lean
def SSABranchAssertRoundTrip :=
#strata
program Core;
procedure check_branch_assert(x : int) {
  if (x > 0) { assert [pos]: (x > 0); }
};
#end

/-- info: true -/
#guard_msgs in
#eval show IO Bool from do
  let pgm := translate SSABranchAssertRoundTrip
  let origResults ← …  (as SSAIncRoundTrip)
  let origPass := origResults.all Core.VCResult.isSuccess
  let .ok ssaPgm := Core.runTransforms pgm [.ssa] | return false
  let ssaResults ← …
  let ssaPass := ssaResults.all Core.VCResult.isSuccess
  return origPass && ssaPass  -- fails today; passes once SSA guards branch asserts
```

2. **Structural / snapshot tests.** The 14 SSA tests at `:43-322` use the oracle `toString fmt result != toString fmt pgm` — they only check that *something* changed. A transform that deletes every procedure passes all 14; so does a transform that emits the empty program with one global `var`. Consider replacing at least one of these (e.g. `SSATest1`) with a `#guard_msgs` snapshot of the expected SSA output, so a regression in the shape of the produced program surfaces directly rather than only via a downstream verification failure.

tautschnig (changes requested):
**Remaining soundness bug: flattened ITE loses the path-condition guard for branch-specific asserts.**

Reproducer I ran locally on this branch:

```lean
procedure check_branch_assert(x : int) {
  if (x > 0) {
    assert [pos]: (x > 0);
  }
};
```

Original: verifies (path condition for the assert is `x > 0` from the branch guard — trivially entails the assert).

After `--pass ssa`, the program becomes (copied verbatim from the output):

```
procedure check_branch_assert (x : int)
{
  var ssa_cond_0 : bool := x > 0;
  assert [pos]: x > 0;
};
```

SSA'd: **fails** — `x` is an input, the assert is now unconditional, and `x = -5` is a counterexample the solver picks up. The `if (x > 0) { … }` wrapper has been flattened away, and the assert no longer has `ssa_cond_0 → …` as its guard.

This falsifies the `.modelPreserving` claim (`SSA.lean:256`): the SSA'd program produces a counterexample (`x = -5`) that is **not** a counterexample in the original (where `x = -5` simply skips the then-branch). So an SSA run in deductive mode reports false-alarm failures, and an SSA run in bug-finding mode reports spurious counterexamples. Both are the kind of unsoundness `.modelPreserving` is meant to rule out.

The `.ite .det` arm at `:186-198` flattens both branches inline:

```lean
return (mergedEnv, [condInit] ++ thenStmts' ++ elseStmts' ++ merges)
```

The standard fix in SSA is either (a) guard each branch-local assert with an implication `ssa_cond_0 → assert_expr` before flattening, or (b) keep the `.ite` intact and only add phi inits at the join (no flattening). Details inline on `transformStmt`. Worst-case, route this under `.modelToValidate` until the fix lands.

**Other observations (all smaller):**

- **`ssaTransform` always reports `changed = true`** (`:241`) even when no procedures were transformed (e.g., a program of only axioms/types). Pipeline consumers treat `changed` as "something changed"; reporting `true` unconditionally makes downstream fixpoint loops non-terminating.
- **`emitOutputAssignments` reintroduces `set`** into the output program (`:173`). This breaks the SSA invariant ("every variable assigned exactly once") for output parameters specifically — the output gets `var ssa_y_N := …` followed by `set y := ssa_y_N`. Worth a docstring note that this exception is intentional to preserve the procedure contract, so a reader doesn't misread "SSA" as literally meaning every variable is init-once.
- **`emitOutputAssignments` skips outputs whose final SSA id equals the original.** Fine for the common case, but for an output parameter that a procedure never assigns, we skip — which is correct. Worth a comment saying so.
- **Duplicated `.ite .det` / `.ite .nondet` arms** (`:186-205`) differ only in the RHS of `condInit` (`.det condExpr'` vs `.nondet`). Extract a helper `buildIteLocal (condInitRhs : ExprOrNondet Expression)` that emits condVar + branches + merges. Not blocking.
- **`rewriteExpr`'s `substFvars` handles bound variables correctly** (standard capture-avoiding), but there's no test that exercises SSA inside a quantifier (the only `LExpr` binders in the asserts that might appear in procedure bodies). Worth a fixture: `assert [q]: ∀ i : int. x > i` — verify SSA renames `x` but leaves `i` alone.

**Test-coverage gaps:**

- The new `SSACorrectnessTests` section (`:324-423`) adds 3 round-trip fixtures; none of them exercise branch-specific asserts, so the soundness bug above is invisible to the test suite. Adding my `check_branch_assert` reproducer would catch it.
- 14 of the 16 tests use the oracle `toString fmt result != toString fmt pgm` (unchanged since thread T14). Those tests only check that the transform did *something*; they pass if a transform deletes every procedure, or if it emits spurious counterexamples. I'd like to see at least one of them replaced with a structural `#guard_msgs` check on the expected output shape.

**Proof suggestions:**

Three theorems would formalize the SSA invariants and (importantly) would fail to prove today because of the soundness bug — making the bug provable-as-a-lemma, which is the right place for it:

```lean
/-- Single-assignment invariant: after `ssaTransform`, no variable is assigned more than once. -/
theorem ssaTransform_single_assignment
    (p : Program) (p' : Program) (h : Core.Transform.run p (Core.SSA.ssaTransform · |>.map (·.2)) = .ok p') :
    ∀ proc ∈ p'.decls.filterMap (·.getProc?),
      (∀ name, (Core.SSA.countInits name proc.body) + (Core.SSA.countSets name proc.body) ≤ 1)
```
(The output-writeback `set` is the one documented exception — the statement would need an explicit carve-out for procedure output parameters.)

```lean
/-- Scoping invariant: every fvar referenced in `ssaTransform p` is either an input,
    an output, or introduced by an SSA-generated init earlier in the same scope. -/
theorem ssaTransform_closed_fvars (p p') : …
```

```lean
/-- Model-preservation: if the SSA'd program has a counterexample for obligation Q,
    the original program has one too.  Fails under the current `.ite` flattening
    — proving this is the right gate for flipping back from `.modelToValidate`. -/
theorem ssaTransform_preserves_model (p p' : Program) (Q : ProofObligation Expression) (m : Model) :
    Core.Transform.run p (Core.SSA.ssaTransform · |>.map (·.2)) = .ok p' →
    Q ∈ obligationsOf p' → isModel Q m → ∃ Q', Q' ∈ obligationsOf p ∧ isModel Q' m
```

Filing these even as `sorry`-terminated stubs would make the soundness story visible at the type level and would prevent silent regressions when a future edit touches the `.ite` arm.

PR #1131, opened 2026-05-06
Unify pyAnalyzeLaurel error classification

julesmt:
The two pipeline stages (Python→Laurel, Laurel→Core) each had their own error-handling block, and the Laurel→Core one misreported user-facing errors as internal. Now both stages throw List DiagnosticModel and route through a single classifier that maps DiagnosticType to exit category. Preserves existing user_errors.txt and DETAIL formatting.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Strata/Languages/Python/PySpecPipeline.lean
+++ b/Strata/Languages/Python/PySpecPipeline.lean
@@ -390,34 +390,16 @@ public def translateCombinedLaurel (combined : Laurel.Program)
   let (coreOption, errors, _, _) ← translateCombinedLaurelWithLowered combined (profile := profile)
   return (coreOption, errors)
 
-/-- Errors from the pyAnalyzeLaurel pipeline: a batch of diagnostics.
-
-Each entry carries a `DiagnosticType` (UserError, NotYetImplemented, StrataBug,
-or Warning) so the caller can classify the batch into a single exit category
-using `classifyDiagnostics`.  This type replaces the prior three-case
-`PipelineError` inductive; routing is now driven by `DiagnosticType` rather
-than by sum-type constructors, so the Python→Laurel and Laurel→Core pipeline
-stages share a single classifier. -/
+/-- A batch of diagnostics emitted by the pyAnalyzeLaurel pipeline. Routing
+to an exit category is driven by each entry's `DiagnosticType` via
+`classifyDiagnostics`. -/
 public abbrev PipelineError := List DiagnosticModel
 
-/-- Helper: create a user-code error diagnostic.
-
-The optional `filePath`/`range` populate the `FileRange` used by
-`classifyDiagnosticsAndExit` to format locations and by the
-`user_errors.txt` side effect to emit SMT `set-info` lines. -/
 public def mkUserCodeDiagnostic (filePath : String) (range : SourceRange) (msg : String)
     : DiagnosticModel :=
   { fileRange := { file := .file filePath, range := range }
     message := msg
     type := .UserError }
-
-/-- Helper: create a known-limitation diagnostic (intentionally unsupported). -/
-public def mkKnownLimitationDiagnostic (msg : String) : DiagnosticModel :=
-  { fileRange := FileRange.unknown, message := msg, type := .NotYetImplemented }
-
-/-- Helper: create an internal-error diagnostic (tool bug). -/
-public def mkInternalDiagnostic (msg : String) : DiagnosticModel :=
-  { fileRange := FileRange.unknown, message := msg, type := .StrataBug }
 
 /-- Run the pyAnalyzeLaurel pipeline: read a Python Ion program,
     resolve overloads from dispatch files, load PySpec declarations,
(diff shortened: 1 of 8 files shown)


julesmt:
@tautschnig Then I can use that to solve your suggestions in PR https://github.com/strata-org/Strata/pull/1118

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -390,20 +390,34 @@ public def translateCombinedLaurel (combined : Laurel.Program)
...
-
-public instance : ToString PipelineError where
-  toString
-    | .userCode _ msg => s!"User code error: {msg}"
-    | .knownLimitation msg => s!"Known limitation: {msg}"
-    | .internal msg => msg
+/-- Errors from the pyAnalyzeLaurel pipeline: a batch of diagnostics.
+
+Each entry carries a `DiagnosticType` (UserError, NotYetImplemented, StrataBug,
+or Warning) so the caller can classify the batch into a single exit category
```
MikaelMayer: 🔍 This docstring references "the prior three-case `PipelineError` inductive" — comments should be stateless and not reference previous versions of the code. Consider rephrasing to describe what this type *is* without referencing what it replaced.

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -390,20 +390,34 @@ public def translateCombinedLaurel (combined : Laurel.Program)
...
+`user_errors.txt` side effect to emit SMT `set-info` lines. -/
+public def mkUserCodeDiagnostic (filePath : String) (range : SourceRange) (msg : String)
+    : DiagnosticModel :=
+  { fileRange := { file := .file filePath, range := range }
+    message := msg
+    type := .UserError }
+
+/-- Helper: create a known-limitation diagnostic (intentionally unsupported). -/
+public def mkKnownLimitationDiagnostic (msg : String) : DiagnosticModel :=
+  { fileRange := FileRange.unknown, message := msg, type := .NotYetImplemented }
```
MikaelMayer: 🔍 `mkKnownLimitationDiagnostic` and `mkInternalDiagnostic` duplicate `DiagnosticModel.fromMessage` (defined in `FileRange.lean`) — they're equivalent to `DiagnosticModel.fromMessage msg (type := .NotYetImplemented)` and `DiagnosticModel.fromMessage msg (type := .StrataBug)` respectively. `mkUserCodeDiagnostic` adds the file/range so it has a reason to exist, but the other two could be inlined or replaced by the existing helper.

StrataMain.lean
```diff
@@ -506,6 +507,64 @@ private def exitPyAnalyzeKnownLimitation {α} (message : String) : IO α := do
...
+    match fileMap with
+    | some fm =>
+      let pos := fm.toPosition d.fileRange.range.start
+      s!" at line {pos.line}, col {pos.column}"
+    | none => ""
+  s!"{d.message}{location}"
+
+/-- Classify diagnostics and exit with the matching `pyAnalyzeLaurel` category.
+    `prefixMsg` is prepended to internal-error and known-limitation details;
+    user-error details use the diagnostic's own message + line/column only. -/
```
MikaelMayer: 🔍 Docstring says "`prefixMsg` is prepended to internal-error and known-limitation details" but it's only used in the `internalError` branch. The `knownLimitation` branch uses `d.message` directly.

StrataMain.lean
```diff
@@ -506,6 +507,64 @@ private def exitPyAnalyzeKnownLimitation {α} (message : String) : IO α := do
...
+    exitPyAnalyzeKnownLimitation msg
+
+/-- Emit SMT `set-info` lines and write `user_errors.txt` for the first
+    user-code diagnostic in the batch. -/
+private def emitUserErrorsFile (diags : List DiagnosticModel) : IO Unit := do
+  let userDiags := diags.filter (·.type == DiagnosticType.UserError)
+  match userDiags.head? with
+  | none => pure ()
+  | some d =>
+    let filePath := match d.fileRange.file with | .file p => p
```
MikaelMayer: 🔍 `let filePath := match d.fileRange.file with | .file p => p` — this is fine today since `Uri` has a single constructor, but if `Uri` ever gains another case (e.g., a URI scheme), this will silently break. A destructuring `let .file filePath := d.fileRange.file` would make the intent clearer and produce a compile error if `Uri` grows.

MikaelMayer (commented):
🔍 Clean unification — the single classifier approach is a good simplification and the new test demonstrates the Laurel→Core user-error path that was previously misreported. A few items below.

julesmt:
@MikaelMayer thanks for your review. I tried to apply your suggestions :).

MikaelMayer (commented):

julesmt:
@tautschnig could you be my second reviewer for merge, like that I can rewrite the other PR this afternoon

MikaelMayer (dismissed):

joehendrix:
This conflicts with #1093.  I think unifying the concepts is good, but Laurel to core failures should be treated as internal errors.  We should stop generating invalid laurel rather than rely on it to catch bugs.

PR #1132, opened 2026-05-06
Add Concrete Syntax for Unstructured Programs in Strata Core

PROgram52bc:
# Add Support for Unstructured Programs in Strata Core
## Description of changes

- Adds syntax and parser support for unstructured (CFG-based) procedure bodies in Strata Core. For example,
```c
cfg entry { 
    // Deterministic block (default): 
    entry: {  
      x := 0; 
      if (x < n) goto entry else done;       
    } 
    done: {
      return; 
    }
  }
  
cfg entry {
    // Nondeterministic block:   
    entry: {   
      y := y + 1;     
      goto entry, done;   // nondeterministic choice 
    }
    done: {
      return; 
    }
  }
```
Prior to this PR, the only way to obtain unstructured programs was to apply `StructuredToUnstructured` transformation. 
- Adds `.core.st` file examples illustrating the syntax of unstructured programs.
- Introduces Procedure.Body as a sum type:
```lean
  inductive Procedure.Body where
    | structured : List Statement → Procedure.Body
    | cfg : DetCFG → Procedure.Body
```
- Adapts uses of `Procedure.body` to handle the `cfg` case based on the context
- **Adds [metadata](https://github.com/strata-org/Strata/pull/1132/changes/4d29cde63ccb04832b99eb6dc90a03ab2faefdcb) support to unstructured programs (new since initial PR draft)**
- **Propagates loop invariants/decreases measure to unstructured CFG during transformation (new since initial PR draft)** 
- **Adds [CFG-based Core-to-GOTO pipeline](https://github.com/strata-org/Strata/pull/1132/changes/63181c964dc0fe526583db18c94302bab891677f) alongside direct path (new since initial PR draft)** 

## Adaptation Methods for `Procedure.body` Uses

1. **CFG (Implemented)** — Functions that handle both structured and CFG bodies with meaningful separate logic for each case.
1. **CFG (N/A)** — Operations where CFGs legitimately contribute nothing (e.g., no local funcDecls, no structured loops), so returning [] or skipping is semantically correct.                                          
1. **CFG (deferred)** — Operations that currently error on CFGs or produce placeholder results, where proper CFG support is desirable but requires non-trivial additional work (e.g., dominator analysis, two-stage    
  pipelines, graph-level inlining).
1. **Proof (Implemented)** — Formal correctness proofs adapted to the `Procedure.Body` sum type.
1. **Proof (deferred)** — Proof obligations where CFG correctness is not yet captured; the current formulation is trivially true for CFGs and is left for future PRs. 

(**Bolded items are updates since the initial PR draft**)

| Category                | Specific Strategy                                                                                                                        | File                                          | Function/Location                                     |
| ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------- | ----------------------------------------------------- |
| **Proof (Implemented)** | **`postconditionsValid` adapted with `CoreBodyExec`**                                                                                    | `Transform/CoreSpecification.lean`            | `ProcedureCorrect.postconditionsValid`                |
|                         | Proves structured if transform succeeds                                                                                                  | `Transform/ProcBodyVerifyCorrect.lean`        | `procToVerifyStmt_is_structured` (new helper theorem) |
|                         | Unifies two `ss` witnesses via `subst`                                                                                                   | `Transform/ProcBodyVerifyCorrect.lean`        | `procBodyVerify_procedureCorrect`                     |
|                         | Adjusted case splits                                                                                                                     | `Languages/Core/ObligationExtraction.lean`    | `extractGo_ok` proof                                  |
| **CFG (Implemented)**   | **Symbolic evaluation of CFG bodies using fuel measure, without path merging.**                                                          | `Languages/Core/ProcedureEval.lean`           | `eval`                                                |
|                         | **Type-check CFG bodies (labels, vars, targets) implemented.**                                                                           | `Languages/Core/ProcedureType.lean`           | `typeCheck`, `checkModificationRights`                |
|                         | **Alternative two-stage pipeline implemented by `procedureToGotoCtxViaCFG`**                                                             | `Backends/CBMC/GOTO/CoreToCProverGOTO.lean`   | `transformToGoto`                                     |
|                         | **Alternative two-stage pipeline implemented by `procedureToGotoCtxViaCFG`**                                                             | `Backends/CBMC/GOTO/CoreToGOTOPipeline.lean`  | `procedureToGotoCtx`                                  |
|                         | **Alternative two-stage pipeline implemented by `procedureToGotoCtxViaCFG`**                                                             | `StrataTest/.../E2E_CoreToGOTO.lean`          | `coreToGotoJsonWithSummary`                           |
|                         | **Alternative two-stage pipeline implemented by `procedureToGotoCtxViaCFG`**                                                             | `Backends/CBMC/CoreToCBMC.lean`               | `createImplementationSymbolFromAST`                   |
|                         | `extractCallsFromStatements` vs `extractCallsFromDetCFG`                                                                                 | `Languages/Core/CallGraph.lean`               | `extractCallsFromProcedure`                           |
|                         | `blockToCST` vs `detCFGToCST`                                                                                                            | `Languages/Core/DDMTransform/FormatCore.lean` | `procToCST`                                           |
|                         | `runStmt` vs `runCFG`                                                                                                                    | `Languages/Core/StatementEval.lean`           | `Command.runCall`                                     |
|                         | Use `CoreBodyExec` with two constructors, and `CoreStepStar` vs `CoreCFGStepStar`                                                        | `Languages/Core/StatementSemantics.lean`      | `EvalCommand.call_sem`                                |
|                         | `Lang.core` vs `Lang.coreCFG`                                                                                                            | `Transform/CoreSpecification.lean`            | `AssertValidInProcedure`                              |
|                         | `transformStmts` vs `transformDetCFG`                                                                                                    | `Transform/PrecondElim.lean`                  | `precondElim`                                         |
|                         | `eraseTypes`, `stripMetaData`, `getVars` implemented and used for CFG                                                                    | `Languages/Core/Procedure.lean`               | `eraseTypes`, `stripMetaData`, `getVars`              |
|                         | `extractFromStatements` vs `extractFromDetCFG`                                                                                           | `Languages/Core/ObligationExtraction.lean`    | `extractObligations`                                  |
| **CFG (N/A)**           | `return .none` on CFG, skip inlining, because the inlined procedure could be structured or unstructured. Semantics is largely undefined. | `Transform/ProcedureInlining.lean`            | `inlineCallCmd`                                       |
|                         | `stmtsToCFG` vs identity, latter is a no-op.                                                                                             | `StrataTest/.../Loops.lean`                   | `singleCFG`                                           |
|                         | Remove loops in structured, pass CFG through because it has no loops                                                                     | `Transform/LoopElim.lean`                     | `loopElim`                                            |
|                         | `[]` for CFG (no local funcDecls in CFG bodies)                                                                                          | `Languages/Core/Core.lean`                    | `buildEnv`                                            |
| **CFG (deferred)**      | Dominator-based path-condition propagation to be implemented                                                                             | `Languages/Core/ObligationExtraction.lean`    | `extractFromDetCFG`                                   |
|                         | `throw` on CFG, due to incompatible return type of `List Statement` instead of `List Command`                                            | `Transform/CoreTransform.lean`                | `runProgram`                                          |
|                         | `throw` on CFG, VC for CFG bodies to be implemented, proof to be adapted as well.                                                        | `Transform/ProcBodyVerify.lean`               | `procToVerifyStmt`                                    |
|                         | Encode structured procedures, handling for CFG procedures to be implemented.                                                             | `Transform/ANFEncoder.lean`                   | `anfEncodeProgram`                                    |
|                         | CFG body renaming for inlining to be implemented                                                                                         | `Transform/ProcedureInlining.lean`            | `renameAllLocalNames`                                 |
|                         | Call extraction for CFG bodies to be implemented                                                                                         | `StrataTest/Boole/global_readonly_call.lean`  | `callHelper`                                          |
| **Proof (deferred)**    | WF Properties `wfstmts`, `wfloclnd`, `bodyExitsCovered` conditioned on structured procedures, props for CFG to be implemented            | `Languages/Core/WF.lean`                      | `WFProcedureProp`                                     |

### DIFF
--- a/Strata/Backends/CBMC/CoreToCBMC.lean
+++ b/Strata/Backends/CBMC/CoreToCBMC.lean
@@ -346,7 +346,9 @@ def createImplementationSymbolFromAST (func : Core.Procedure) : Except String CB
 
   -- For now, keep the hardcoded implementation but use function name from AST
   let loc : SourceLoc := { functionName := (func.header.name.toPretty), lineNum := "1" }
-  let bodyStmts := match func.body with | .structured ss => ss | .cfg _ => []
+  let bodyStmts ← match func.body with
+    | .structured ss => .ok ss
+    | .cfg _ => .error "Cannot translate unstructured CFG body to CBMC JSON format"
   let stmtJsons ← (bodyStmts.mapM (stmtToJson (I:=CoreLParams) · loc))
 
   let implValue := Json.mkObj [
--- a/Strata/Backends/CBMC/CoreToCBMC.lean
+++ b/Strata/Backends/CBMC/CoreToCBMC.lean
@@ -346,9 +346,7 @@ def createImplementationSymbolFromAST (func : Core.Procedure) : Except String CB
 
   -- For now, keep the hardcoded implementation but use function name from AST
   let loc : SourceLoc := { functionName := (func.header.name.toPretty), lineNum := "1" }
-  let bodyStmts ← match func.body with
-    | .structured ss => .ok ss
-    | .cfg _ => .error "Cannot translate unstructured CFG body to CBMC JSON format"
+  let bodyStmts ← func.body.getStructured
   let stmtJsons ← (bodyStmts.mapM (stmtToJson (I:=CoreLParams) · loc))
 
   let implValue := Json.mkObj [
--- a/Strata/Backends/CBMC/CoreToCBMC.lean
+++ b/Strata/Backends/CBMC/CoreToCBMC.lean
@@ -346,7 +346,7 @@ def createImplementationSymbolFromAST (func : Core.Procedure) : Except String CB
 
   -- For now, keep the hardcoded implementation but use function name from AST
   let loc : SourceLoc := { functionName := (func.header.name.toPretty), lineNum := "1" }
-  let stmtJsons ← (func.body.mapM (stmtToJson (I:=CoreLParams) · loc))
+  let stmtJsons ← (func.body.toStmts.mapM (stmtToJson (I:=CoreLParams) · loc))
 
   let implValue := Json.mkObj [
     ("id", "code"),
--- a/Strata/Backends/CBMC/CoreToCBMC.lean
+++ b/Strata/Backends/CBMC/CoreToCBMC.lean
@@ -346,7 +346,8 @@ def createImplementationSymbolFromAST (func : Core.Procedure) : Except String CB
 
   -- For now, keep the hardcoded implementation but use function name from AST
   let loc : SourceLoc := { functionName := (func.header.name.toPretty), lineNum := "1" }
-  let stmtJsons ← (func.body.toStmts.mapM (stmtToJson (I:=CoreLParams) · loc))
+  let bodyStmts := match func.body with | .structured ss => ss | .cfg _ => []
+  let stmtJsons ← (bodyStmts.mapM (stmtToJson (I:=CoreLParams) · loc))
 
   let implValue := Json.mkObj [
     ("id", "code"),
--- a/Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
@@ -201,6 +201,9 @@ def transformToGoto (cprog : Core.Program) : Except Format CProverGOTO.Context :
       if !p.header.typeArgs.isEmpty then
         throw f!"[transformToGoto] Translation for polymorphic Strata Core procedures is unimplemented."
 
+      -- TODO: This pass could be split into a two-stage transformation:
+      -- 1. structured → cfg (via StructuredToUnstructured)
+      -- 2. cfg → CProverGOTO (always operates on CFG, no pattern matching needed)
       let bodyStmts ← match p.body with
         | .structured ss => pure ss
         | .cfg _ => throw f!"CFG body not supported for GOTO translation"
--- a/Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
@@ -204,9 +204,7 @@ def transformToGoto (cprog : Core.Program) : Except Format CProverGOTO.Context :
       -- TODO: This pass could be split into a two-stage transformation:
       -- 1. structured → cfg (via StructuredToUnstructured)
       -- 2. cfg → CProverGOTO (always operates on CFG, no pattern matching needed)
-      let bodyStmts ← match p.body with
-        | .structured ss => pure ss
-        | .cfg _ => throw f!"CFG body not supported for GOTO translation"
+      let bodyStmts ← p.body.getStructured.mapError fun s => f!"{s}"
       let cmds ← bodyStmts.mapM
         (fun b => match b with
           | .cmd (.cmd c) => return c
--- a/Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
@@ -201,7 +201,10 @@ def transformToGoto (cprog : Core.Program) : Except Format CProverGOTO.Context :
       if !p.header.typeArgs.isEmpty then
         throw f!"[transformToGoto] Translation for polymorphic Strata Core procedures is unimplemented."
 
-      let cmds ← p.body.toStmts.mapM
+      let bodyStmts ← match p.body with
+        | .structured ss => pure ss
+        | .cfg _ => throw f!"CFG body not supported for GOTO translation"
+      let cmds ← bodyStmts.mapM
         (fun b => match b with
           | .cmd (.cmd c) => return c
           | _ => throw f!"[transformToGoto] We can process Strata Core commands only, not statements! \
@@ -218,7 +221,7 @@ def transformToGoto (cprog : Core.Program) : Except Format CProverGOTO.Context :
       let formals_renamed := formals.zip new_formals
       let formals_tys : Map String CProverGOTO.Ty := formals.zip formals_tys
 
-      let locals := (Imperative.Block.definedVars p.body.toStmts).map Core.CoreIdent.toPretty
+      let locals := (Imperative.Block.definedVars bodyStmts).map Core.CoreIdent.toPretty
       let new_locals := locals.map (fun l => CProverGOTO.mkLocalSymbol pname l)
       let locals_renamed := locals.zip new_locals
 
--- a/Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
@@ -201,7 +201,7 @@ def transformToGoto (cprog : Core.Program) : Except Format CProverGOTO.Context :
       if !p.header.typeArgs.isEmpty then
         throw f!"[transformToGoto] Translation for polymorphic Strata Core procedures is unimplemented."
 
-      let cmds ← p.body.mapM
+      let cmds ← p.body.toStmts.mapM
         (fun b => match b with
           | .cmd (.cmd c) => return c
           | _ => throw f!"[transformToGoto] We can process Strata Core commands only, not statements! \
@@ -218,7 +218,7 @@ def transformToGoto (cprog : Core.Program) : Except Format CProverGOTO.Context :
       let formals_renamed := formals.zip new_formals
       let formals_tys : Map String CProverGOTO.Ty := formals.zip formals_tys
 
-      let locals := (Imperative.Block.definedVars p.body).map Core.CoreIdent.toPretty
+      let locals := (Imperative.Block.definedVars p.body.toStmts).map Core.CoreIdent.toPretty
       let new_locals := locals.map (fun l => CProverGOTO.mkLocalSymbol pname l)
       let locals_renamed := locals.zip new_locals
 
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -40,12 +40,12 @@ namespace Strata
 
 public section
 
-private def renameIdent (rn : Std.HashMap String String) (id : Core.CoreIdent) : Core.CoreIdent :=
+def renameIdent (rn : Std.HashMap String String) (id : Core.CoreIdent) : Core.CoreIdent :=
   match rn[id.name]? with
   | some new => ⟨new, id.metadata⟩
   | none => id
 
-private partial def renameExpr
+partial def renameExpr
     (rn : Std.HashMap String String)
     : Core.Expression.Expr → Core.Expression.Expr
   | .fvar m name ty => .fvar m (renameIdent rn name) ty
@@ -56,7 +56,7 @@ private partial def renameExpr
   | .eq m e1 e2 => .eq m (renameExpr rn e1) (renameExpr rn e2)
   | e => e
 
-private def renameCmd
+def renameCmd
     (rn : Std.HashMap String String)
     : Imperative.Cmd Core.Expression → Imperative.Cmd Core.Expression
   | .init name ty e md => .init (renameIdent rn name) ty (e.map (renameExpr rn)) md
@@ -102,7 +102,7 @@ private def hasCallStmt : List Core.Statement → Bool
 Collect all funcDecl statements from a procedure body (recursively)
 and return them as Core.Functions, stripping them from the body.
 -/
-private def collectFuncDecls : List Core.Statement →
+def collectFuncDecls : List Core.Statement →
     Except Std.Format (List Core.Function × List Core.Statement)
   | [] => return ([], [])
   | .funcDecl decl _ :: rest => do
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -259,7 +259,7 @@ def procedureToGotoCtx
     : Except Std.Format
         (CoreToGOTO.CProverGOTO.Context × List Core.Function) := do
   -- Lift local function declarations out of the body
-  let (liftedFuncs, body) ← collectFuncDecls p.body
+  let (liftedFuncs, body) ← collectFuncDecls p.body.toStmts
   let pname := Core.CoreIdent.toPretty p.header.name
   if !p.header.typeArgs.isEmpty then
     .error f!"[procedureToGotoCtx] Polymorphic procedures unsupported."
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -262,9 +262,7 @@ def procedureToGotoCtx
   -- TODO: This pass could be split into a two-stage transformation:
   -- 1. structured → cfg (via StructuredToUnstructured)
   -- 2. cfg → CProverGOTO (always operates on CFG, no pattern matching needed)
-  let bodyStmts ← match p.body with
-    | .structured ss => pure ss
-    | .cfg _ => throw f!"CFG body not supported for GOTO pipeline"
+  let bodyStmts ← p.body.getStructured.mapError fun s => f!"{s}"
   let (liftedFuncs, body) ← collectFuncDecls bodyStmts
   let pname := Core.CoreIdent.toPretty p.header.name
   if !p.header.typeArgs.isEmpty then
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -259,6 +259,9 @@ def procedureToGotoCtx
     : Except Std.Format
         (CoreToGOTO.CProverGOTO.Context × List Core.Function) := do
   -- Lift local function declarations out of the body
+  -- TODO: This pass could be split into a two-stage transformation:
+  -- 1. structured → cfg (via StructuredToUnstructured)
+  -- 2. cfg → CProverGOTO (always operates on CFG, no pattern matching needed)
   let bodyStmts ← match p.body with
     | .structured ss => pure ss
     | .cfg _ => throw f!"CFG body not supported for GOTO pipeline"
--- a/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
+++ b/Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
@@ -259,7 +259,10 @@ def procedureToGotoCtx
     : Except Std.Format
         (CoreToGOTO.CProverGOTO.Context × List Core.Function) := do
   -- Lift local function declarations out of the body
-  let (liftedFuncs, body) ← collectFuncDecls p.body.toStmts
+  let bodyStmts ← match p.body with
+    | .structured ss => pure ss
+    | .cfg _ => throw f!"CFG body not supported for GOTO pipeline"
+  let (liftedFuncs, body) ← collectFuncDecls bodyStmts
   let pname := Core.CoreIdent.toPretty p.header.name
   if !p.header.typeArgs.isEmpty then
     .error f!"[procedureToGotoCtx] Polymorphic procedures unsupported."
--- a/Strata/Languages/Core/DDMTransform/FormatCore.lean
+++ b/Strata/Languages/Core/DDMTransform/FormatCore.lean
@@ -869,6 +869,44 @@ partial def invariantsToCST {M} [Inhabited M]
     let restCST ← invariantsToCST rest
     pure (.consInvariants default labelAnn exprCST restCST)
 
+/-- Convert a `DetTransferCmd` to a CST statement (if-then-else for condGoto, skip for finish). -/
+partial def transferToCST {M} [Inhabited M]
+    (t : Imperative.DetTransferCmd String Expression) : ToCSTM M (Option (CoreDDM.Statement M)) := do
+  match t with
+  | .condGoto cond lt lf _ => do
+    let condCST ← lexprToExpr cond 0
+    let gotoTrue : CoreDDM.Statement M := .exit_statement default ⟨default, lt⟩
+    let gotoFalse : CoreDDM.Statement M := .exit_statement default ⟨default, lf⟩
+    let thenBlock : CoreDDM.Block M := .block default ⟨default, #[gotoTrue]⟩
+    let elseBlock : Else M := .else1 default (.block default ⟨default, #[gotoFalse]⟩)
+    pure (some (.if_statement default (.condDet default condCST) thenBlock elseBlock))
+  | .finish _ => pure none
+
+/-- Convert a single `DetBlock` to a CST block (commands + transfer). -/
+partial def detBlockToCST {M} [Inhabited M]
+    (blk : Imperative.DetBlock String Core.Command Expression)
+    : ToCSTM M (CoreDDM.Block M) := do
+  modify ToCSTContext.pushScope
+  let cmdStmts ← blk.cmds.toArray.mapM (stmtToCST ∘ Imperative.Stmt.cmd)
+  let transferStmt ← transferToCST blk.transfer
+  let allStmts := match transferStmt with
+    | some s => cmdStmts.push s
+    | none => cmdStmts
+  modify ToCSTContext.popScope
+  pure (.block default ⟨default, allStmts⟩)
+
+/-- Convert a `DetCFG` to a CST block (sequence of labeled blocks). -/
+partial def detCFGToCST {M} [Inhabited M] (cfg : Core.DetCFG)
+    : ToCSTM M (CoreDDM.Block M) := do
+  modify ToCSTContext.pushScope
+  let mut stmts : Array (CoreDDM.Statement M) := #[]
+  for (label, blk) in cfg.blocks do
+    let labelAnn : Ann String M := ⟨default, label⟩
+    let blockCST ← detBlockToCST blk
+    stmts := stmts.push (.block_statement default labelAnn blockCST)
+  modify ToCSTContext.popScope
+  pure (.block default ⟨default, stmts⟩)
+
 partial def measureToCST {M} [Inhabited M]
     (measure : Option (Lambda.LExpr CoreLParams.mono)) :
     ToCSTM M (Ann (Option (Measure M)) M) := do
@@ -940,7 +978,7 @@ def procToCST {M} [Inhabited M] (proc : Core.Procedure) : ToCSTM M (Command M) :
       ⟨default, some (Spec.spec_mk default specAnn)⟩
   let bodyCST ← match proc.body with
     | .structured ss => blockToCST ss
-    | .cfg _ => blockToCST []
+    | .cfg c => detCFGToCST c
   let body : Ann (Option (CoreDDM.Block M)) M := ⟨default, some bodyCST⟩
   modify ToCSTContext.popScope
   pure (.command_procedure default name typeArgs arguments spec body)
--- a/Strata/Languages/Core/DDMTransform/FormatCore.lean
+++ b/Strata/Languages/Core/DDMTransform/FormatCore.lean
@@ -938,7 +938,7 @@ def procToCST {M} [Inhabited M] (proc : Core.Procedure) : ToCSTM M (Command M) :
       ⟨default, none⟩
     else
       ⟨default, some (Spec.spec_mk default specAnn)⟩
-  let bodyCST ← blockToCST proc.body
+  let bodyCST ← blockToCST proc.body.toStmts
   let body : Ann (Option (CoreDDM.Block M)) M := ⟨default, some bodyCST⟩
   modify ToCSTContext.popScope
   pure (.command_procedure default name typeArgs arguments spec body)
--- a/Strata/Languages/Core/DDMTransform/FormatCore.lean
+++ b/Strata/Languages/Core/DDMTransform/FormatCore.lean
@@ -869,43 +869,48 @@ partial def invariantsToCST {M} [Inhabited M]
     let restCST ← invariantsToCST rest
     pure (.consInvariants default labelAnn exprCST restCST)
 
-/-- Convert a `DetTransferCmd` to a CST statement (if-then-else for condGoto, skip for finish). -/
+/-- Convert a `DetTransferCmd` to its CFG-specific CST `Transfer` node. -/
 partial def transferToCST {M} [Inhabited M]
-    (t : Imperative.DetTransferCmd String Expression) : ToCSTM M (Option (CoreDDM.Statement M)) := do
+    (t : Imperative.DetTransferCmd String Expression) : ToCSTM M (CoreDDM.Transfer M) := do
   match t with
-  | .condGoto cond lt lf _ => do
-    let condCST ← lexprToExpr cond 0
-    let gotoTrue : CoreDDM.Statement M := .exit_statement default ⟨default, lt⟩
-    let gotoFalse : CoreDDM.Statement M := .exit_statement default ⟨default, lf⟩
-    let thenBlock : CoreDDM.Block M := .block default ⟨default, #[gotoTrue]⟩
-    let elseBlock : Else M := .else1 default (.block default ⟨default, #[gotoFalse]⟩)
-    pure (some (.if_statement default (.condDet default condCST) thenBlock elseBlock))
-  | .finish _ => pure none
+  | .condGoto cond lt lf _ =>
+    if lt == lf then
+      pure (.transfer_goto default ⟨default, lt⟩)
+    else
+      match cond with
+      | .fvar _ id _ =>
+        if id.name.startsWith "$__nondet_" then
+          pure (.transfer_nondet_goto default ⟨default, lt⟩ ⟨default, lf⟩)
+        else
+          let condCST ← lexprToExpr cond 0
+          pure (.transfer_cond_goto default condCST ⟨default, lt⟩ ⟨default, lf⟩)
+      | _ =>
+        let condCST ← lexprToExpr cond 0
+        pure (.transfer_cond_goto default condCST ⟨default, lt⟩ ⟨default, lf⟩)
+  | .finish _ => pure (.transfer_return default)
 
-/-- Convert a single `DetBlock` to a CST block (commands + transfer). -/
+/-- Convert a single `DetBlock` to a CST `CFGBlock`. -/
 partial def detBlockToCST {M} [Inhabited M]
-    (blk : Imperative.DetBlock String Core.Command Expression)
-    : ToCSTM M (CoreDDM.Block M) := do
+    (label : String) (blk : Imperative.DetBlock String Core.Command Expression)
+    : ToCSTM M (CoreDDM.CFGBlock M) := do
   modify ToCSTContext.pushScope
   let cmdStmts ← blk.cmds.toArray.mapM (stmtToCST ∘ Imperative.Stmt.cmd)
-  let transferStmt ← transferToCST blk.transfer
-  let allStmts := match transferStmt with
-    | some s => cmdStmts.push s
-    | none => cmdStmts
+  let transfer ← transferToCST blk.transfer
   modify ToCSTContext.popScope
-  pure (.block default ⟨default, allStmts⟩)
+  pure (.cfg_block default ⟨default, label⟩ ⟨default, cmdStmts⟩ transfer)
 
-/-- Convert a `DetCFG` to a CST block (sequence of labeled blocks). -/
+/-- Convert a `DetCFG` to a CST `CFGBody`. -/
 partial def detCFGToCST {M} [Inhabited M] (cfg : Core.DetCFG)
-    : ToCSTM M (CoreDDM.Block M) := do
-  modify ToCSTContext.pushScope
-  let mut stmts : Array (CoreDDM.Statement M) := #[]
-  for (label, blk) in cfg.blocks do
-    let labelAnn : Ann String M := ⟨default, label⟩
-    let blockCST ← detBlockToCST blk
-    stmts := stmts.push (.block_statement default labelAnn blockCST)
-  modify ToCSTContext.popScope
-  pure (.block default ⟨default, stmts⟩)
+    : ToCSTM M (CoreDDM.CFGBody M) := do
+  let cfgBlocks ← cfg.blocks.mapM fun (label, blk) => detBlockToCST label blk
+  let blocks := cfgBlocks.foldr (init := none) fun blk acc =>
+    match acc with
+    | none => some (.cfg_blocks_one default blk)
+    | some rest => some (.cfg_blocks_cons default blk rest)
+  match blocks with
+  | some bs => pure (.cfg_body default ⟨default, cfg.entry⟩ bs)
+  | none => pure (.cfg_body default ⟨default, cfg.entry⟩
+      (.cfg_blocks_one default (.cfg_block default ⟨default, cfg.entry⟩ ⟨default, #[]⟩ (.transfer_return default))))
 
 partial def measureToCST {M} [Inhabited M]
     (measure : Option (Lambda.LExpr CoreLParams.mono)) :
@@ -976,12 +981,16 @@ def procToCST {M} [Inhabited M] (proc : Core.Procedure) : ToCSTM M (Command M) :
       ⟨default, none⟩
     else
       ⟨default, some (Spec.spec_mk default specAnn)⟩
-  let bodyCST ← match proc.body with
-    | .structured ss => blockToCST ss
-    | .cfg c => detCFGToCST c
-  let body : Ann (Option (CoreDDM.Block M)) M := ⟨default, some bodyCST⟩
+  let cmd ← match proc.body with
+    | .structured ss =>
+      let bodyCST ← blockToCST ss
+      let body : Ann (Option (CoreDDM.Block M)) M := ⟨default, some bodyCST⟩
+      pure (.command_procedure default name typeArgs arguments spec body)
+    | .cfg c =>
+      let cfgBody ← detCFGToCST c
+      pure (.command_cfg_procedure default name typeArgs arguments spec cfgBody)
   modify ToCSTContext.popScope
-  pure (.command_procedure default name typeArgs arguments spec body)
+  pure cmd
 
 -- Recreate enough of `GlobalContext` from `ToCSTContext` obtained from
 -- `programToCST`, purely for formatting.
--- a/Strata/Languages/Core/DDMTransform/FormatCore.lean
+++ b/Strata/Languages/Core/DDMTransform/FormatCore.lean
@@ -938,7 +938,9 @@ def procToCST {M} [Inhabited M] (proc : Core.Procedure) : ToCSTM M (Command M) :
       ⟨default, none⟩
     else
       ⟨default, some (Spec.spec_mk default specAnn)⟩
-  let bodyCST ← blockToCST proc.body.toStmts
+  let bodyCST ← match proc.body with
+    | .structured ss => blockToCST ss
+    | .cfg _ => blockToCST []
   let body : Ann (Option (CoreDDM.Block M)) M := ⟨default, some bodyCST⟩
   modify ToCSTContext.popScope
   pure (.command_procedure default name typeArgs arguments spec body)
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -455,6 +455,52 @@ op datatype_decl (name : Ident,
 op command_datatypes (datatypes : NewlineSepBy DatatypeDecl) : Command =>
   datatypes ";\n";
 
+// =====================================================================
+// CFG (Unstructured Control Flow) Syntax
+// =====================================================================
+
+// Transfer commands: how a basic block ends
+category Transfer;
+
+// Goto: one target = unconditional, multiple targets = nondeterministic choice
+op transfer_goto (labels : CommaSepBy Ident) : Transfer =>
+  "goto " labels ";";
+
+// Conditional goto (deterministic: condition selects between two targets)
+op transfer_cond_goto (c : Expr, lt : Ident, lf : Ident) : Transfer =>
+  "branch (" c ") goto " lt " else " lf ";";
+
+// Return/finish (terminate execution)
+op transfer_return : Transfer =>
+  "return;";
+
+// A single CFG basic block: label, commands, transfer
+category CFGBlock;
+@[scope(cmds)]
+op cfg_block (label : Ident, cmds : Seq Statement, tr : Transfer) : CFGBlock =>
+  label ":" " {\n" indent(2, cmds) "  " tr "\n}";
+
+// A list of CFG blocks
+category CFGBlocks;
+op cfg_blocks_one (b : CFGBlock) : CFGBlocks => b;
+op cfg_blocks_cons (b : CFGBlock, rest : CFGBlocks) : CFGBlocks =>
+  b "\n" rest;
+
+// CFG body: entry label + blocks
+category CFGBody;
+op cfg_body (entry : Ident, blocks : CFGBlocks) : CFGBody =>
+  "cfg " entry " {\n" indent(2, blocks) "\n}";
+
+// Procedure with CFG body
+op command_cfg_procedure (name : Ident,
+                          typeArgs : Option TypeArgs,
+                          @[scope(typeArgs)] b : Bindings,
+                          @[scope(b)] s : Option Spec,
+                          @[scope(b)] body : CFGBody) :
+  Command =>
+  @[prec(10)] "procedure " name typeArgs b "\n"
+              s body ";\n";
+
 #end
 
 ---------------------------------------------------------------------
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -467,6 +467,9 @@ op transfer_goto (labels : CommaSepBy Ident) : Transfer =>
   "goto " labels ";";
 
 // Conditional goto (deterministic: condition selects between two targets)
+// NOTE: We use "branch" instead of "if" to avoid ambiguity with the
+// structured if-statement syntax. The DDM parser registers tokens globally,
+// so "if (" in Transfer would conflict with "if (" in Statement.
 op transfer_cond_goto (c : Expr, lt : Ident, lf : Ident) : Transfer =>
   "branch (" c ") goto " lt " else " lf ";";
 
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -25,7 +25,7 @@ namespace Strata
 
 -- Sequence operations and lambda/application syntax increase the grammar size enough
 -- to require higher recursion and heartbeat limits.
-set_option maxRecDepth 10000
+set_option maxRecDepth 20000
 set_option maxHeartbeats 400000
 
 /- DDM support for parsing and pretty-printing Strata Core -/
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -463,7 +463,9 @@ op command_datatypes (datatypes : NewlineSepBy DatatypeDecl) : Command =>
 // Transfer commands: how a basic block ends
 category Transfer;
 
-// Goto: one target = unconditional, multiple targets = nondeterministic choice
+// Goto: one target = unconditional, two targets = nondeterministic choice.
+// The grammar accepts any number of comma-separated labels, but the translator
+// currently supports at most 2 (see translateTransfer in Translate.lean).
 op transfer_goto (labels : CommaSepBy Ident) : Transfer =>
   "goto " labels ";";
 
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -463,11 +463,13 @@ op command_datatypes (datatypes : NewlineSepBy DatatypeDecl) : Command =>
 // Transfer commands: how a basic block ends
 category Transfer;
 
-// Goto: one target = unconditional, two targets = nondeterministic choice.
-// The grammar accepts any number of comma-separated labels, but the translator
-// currently supports at most 2 (see translateTransfer in Translate.lean).
-op transfer_goto (labels : CommaSepBy Ident) : Transfer =>
-  "goto " labels ";";
+// Unconditional goto: exactly one target.
+op transfer_goto (label : Ident) : Transfer =>
+  "goto " label ";";
+
+// Nondeterministic goto: exactly two targets chosen nondeterministically.
+op transfer_nondet_goto (label1 : Ident, label2 : Ident) : Transfer =>
+  "goto " label1 ", " label2 ";";
 
 // Conditional goto (deterministic: condition selects between two targets)
 // NOTE: We use "branch" instead of "if" to avoid ambiguity with the
--- a/Strata/Languages/Core/DDMTransform/Translate.lean
+++ b/Strata/Languages/Core/DDMTransform/Translate.lean
@@ -1628,7 +1628,9 @@ partial def translateTransfer (p : Program) (bindings : TransBindings) (arg : Ar
     | [l] =>
       let label ← translateIdent String l
       return .condGoto (Lambda.LExpr.boolConst () Bool.true) label label
-    | l1 :: l2 :: _ =>
+    | l1 :: l2 :: rest =>
+      if !rest.isEmpty then
+        TransM.error s!"translateTransfer: goto with more than 2 targets is not supported"
       let label1 ← translateIdent String l1
       let label2 ← translateIdent String l2
       let condName := s!"$__nondet_{bindings.gen.var_def}"
--- a/Strata/Languages/Core/DDMTransform/Translate.lean
+++ b/Strata/Languages/Core/DDMTransform/Translate.lean
@@ -1611,18 +1611,19 @@ def translateBlockCommand (p : Program) (bindings : TransBindings) (op : Operati
 
 /-- Translate a transfer command from the CFG syntax -/
 
+private instance : Inhabited TransBindings := ⟨{}⟩
 private instance : Inhabited (Imperative.DetTransferCmd String Core.Expression) := ⟨.finish⟩
 private instance : Inhabited (Imperative.BasicBlock (Imperative.DetTransferCmd String Core.Expression) Core.Command) := ⟨⟨[], .finish⟩⟩
 private instance : Inhabited (Imperative.CFG String (Imperative.DetBlock String Core.Command Core.Expression)) := ⟨⟨"", []⟩⟩
 
 partial def translateTransfer (p : Program) (bindings : TransBindings) (arg : Arg) :
-  TransM (Imperative.DetTransferCmd String Core.Expression) := do
+  TransM (Imperative.DetTransferCmd String Core.Expression × TransBindings) := do
   let .op op := arg
     | TransM.error s!"translateTransfer expected op {repr arg}"
   match op.name with
   | q`Core.transfer_goto =>
     let label ← translateIdent String op.args[0]!
-    return .condGoto (Lambda.LExpr.boolConst () Bool.true) label label
+    return (.condGoto (Lambda.LExpr.boolConst () Bool.true) label label, bindings)
   | q`Core.transfer_nondet_goto =>
     let label1 ← translateIdent String op.args[0]!
     let label2 ← translateIdent String op.args[1]!
@@ -1633,14 +1634,15 @@ partial def translateTransfer (p : Program) (bindings : TransBindings) (arg : Ar
     -- will error on this, which is expected — nondeterministic gotos are only
     -- meaningful under symbolic execution.
     let condName := s!"$__nondet_{bindings.gen.var_def}"
-    return .condGoto (Lambda.LExpr.fvar () ⟨condName, ()⟩ none) label1 label2
+    let bindings := incrNum .var_def bindings
+    return (.condGoto (Lambda.LExpr.fvar () ⟨condName, ()⟩ none) label1 label2, bindings)
   | q`Core.transfer_cond_goto =>
     let cond ← translateExpr p bindings op.args[0]!
     let lt ← translateIdent String op.args[1]!
     let lf ← translateIdent String op.args[2]!
-    return .condGoto cond lt lf
+    return (.condGoto cond lt lf, bindings)
   | q`Core.transfer_return =>
-    return .finish
+    return (.finish, bindings)
   | _ => TransM.error s!"translateTransfer: unknown transfer {repr op.name}"
 
 /-- Translate a single CFG block -/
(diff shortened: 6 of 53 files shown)


Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
```diff
@@ -201,7 +201,10 @@ def transformToGoto (cprog : Core.Program) : Except Format CProverGOTO.Context :
       if !p.header.typeArgs.isEmpty then
         throw f!"[transformToGoto] Translation for polymorphic Strata Core procedures is unimplemented."
 
-      let cmds ← p.body.mapM
+      let bodyStmts ← match p.body with
+        | .structured ss => pure ss
+        | .cfg _ => throw f!"CFG body not supported for GOTO translation"
+      let cmds ← bodyStmts.mapM
```
PROgram52bc, line 208: add a comment saying maybe this pass could be split into a two-stage transformation. Then a hypothetical first-stage transformation would be from `.structured` to `.cfg`, and second stage from `.cfg` to `CProverGOTO`, with no need for pattern matching.

Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
```diff
@@ -259,7 +259,10 @@ def procedureToGotoCtx
     : Except Std.Format
         (CoreToGOTO.CProverGOTO.Context × List Core.Function) := do
   -- Lift local function declarations out of the body
-  let (liftedFuncs, body) ← collectFuncDecls p.body
+  let bodyStmts ← match p.body with
+    | .structured ss => pure ss
+    | .cfg _ => throw f!"CFG body not supported for GOTO pipeline"
```
PROgram52bc: Add comments aimilar to the above.

Strata/Backends/CBMC/GOTO/CoreToCProverGOTO.lean
PROgram52bc: Addressed — added TODO comment noting the two-stage pipeline refactor opportunity.

Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
PROgram52bc: Addressed — added the same TODO comment here as well.

Strata/Backends/CBMC/CoreToCBMC.lean
```diff
@@ -346,7 +346,8 @@ def createImplementationSymbolFromAST (func : Core.Procedure) : Except String CB
 
   -- For now, keep the hardcoded implementation but use function name from AST
   let loc : SourceLoc := { functionName := (func.header.name.toPretty), lineNum := "1" }
-  let stmtJsons ← (func.body.mapM (stmtToJson (I:=CoreLParams) · loc))
+  let bodyStmts := match func.body with | .structured ss => ss | .cfg _ => []
```
PROgram52bc: throw an error instead of returning [], noting that we can't turn unstructured CFGs into structured JSON format in general.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -812,8 +812,10 @@ def Command.runCall (lhs : List Expression.Ident) (procName : String) (args : Li
               hasError := fun E => E.error.isSome
               addError := fun E msg => CmdEval.updateError E (.Misc msg)
             }
+            let bodyStmts := match proc.body with
+              | .structured ss => ss | .cfg _ => []
```
PROgram52bc: interpret the CFG as well

Strata/Languages/Core/Procedure.lean
```diff
@@ -278,21 +279,55 @@ def Procedure.Spec.updateCheckExprs
...
+  /-- An unstructured body: a control-flow graph of deterministic basic blocks.
+      Labels are strings; each block contains Core commands and ends with a
+      deterministic transfer (conditional goto or finish). -/
+  | cfg : Imperative.CFG String (Imperative.DetBlock String Command Expression) → Procedure.Body
+  deriving Inhabited
+
+/-- Extract the structured statements, or error if the body is a CFG. -/
+def Procedure.Body.getStructured : Procedure.Body → Except String (List Statement)
+  | .structured ss => .ok ss
+  | .cfg _ => .error "expected structured body, got CFG"
```
PROgram52bc, line 301: add `.getCfg` method as well

Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -455,6 +455,52 @@ op datatype_decl (name : Ident,
...
+// Transfer commands: how a basic block ends
+category Transfer;
+
+// Goto: one target = unconditional, multiple targets = nondeterministic choice
+op transfer_goto (labels : CommaSepBy Ident) : Transfer =>
+  "goto " labels ";";
+
+// Conditional goto (deterministic: condition selects between two targets)
+op transfer_cond_goto (c : Expr, lt : Ident, lf : Ident) : Transfer =>
+  "branch (" c ") goto " lt " else " lf ";";
```
PROgram52bc, line 488: use `if` instead of `branch`.

Strata/Backends/CBMC/CoreToCBMC.lean
PROgram52bc: Done — now throws an Except error instead of returning [].

Strata/Languages/Core/StatementEval.lean
PROgram52bc: Done — CFG bodies are now interpreted by linearizing blocks (with a TODO for full control-flow-following interpretation).

Strata/Languages/Core/Procedure.lean
PROgram52bc: Done — added Body.getCfg that mirrors getStructured.

Strata/Languages/Core/DDMTransform/Grammar.lean
PROgram52bc: Cannot use 'if' — the DDM parser registers tokens globally, so 'if (' in the Transfer category conflicts with the existing if-statement syntax. Added a comment in the grammar explaining this. Keeping 'branch' as the keyword.

Strata/Languages/Core/Procedure.lean
```diff
@@ -278,21 +279,63 @@ def Procedure.Spec.updateCheckExprs
...
+
+/-- Extract the CFG, or error if the body is structured. -/
+def Procedure.Body.getCfg : Procedure.Body → Except String DetCFG
+  | .cfg c => .ok c
+  | .structured _ => .error "expected CFG body, got structured"
+
+/-- Extract statements from a structured body, or `[]` for CFG.
+    Intended for use in Prop contexts where `Except` is inconvenient. -/
+@[simp, expose] def Procedure.Body.stmts : Procedure.Body → List Statement
```
PROgram52bc: remove this.

Strata/Languages/Core/WF.lean
```diff
@@ -138,15 +138,15 @@ structure WFAxiomDeclarationProp (p : Program) (f : Axiom) : Prop where
...
+  wfstmts : WFStatementsProp p d.body.stmts
+  wfloclnd : (HasVarsImp.definedVars (P:=Expression) d.body.stmts).Nodup
   inputsNodup : (ListMap.keys d.header.inputs).Nodup
   outputsNodup : (ListMap.keys d.header.outputs).Nodup
   ioNotOld : ∀ id ∈ ListMap.keys d.header.inputs ++ ListMap.keys d.header.outputs,
       ∀ x, id ≠ CoreIdent.mkOld x
   wfspec : WFSpecProp p d.spec d
   -- There is no exit statement that cannot be caught by any block in the procedure.
-  bodyExitsCovered : Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks [] d.body
+  bodyExitsCovered : Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks [] d.body.stmts
```
PROgram52bc: remove the `.stmts` methods, and use a case split based on whether d.body is structured or unstructured.

Strata/Languages/Core/Procedure.lean
```diff
@@ -278,21 +279,63 @@ def Procedure.Spec.updateCheckExprs
...
+  | .cfg _ => []
+
+/-- Get variables referenced in the body. -/
+def Procedure.Body.getVars : Procedure.Body → List Expression.Ident
+  | .structured ss => ss.flatMap Imperative.HasVarsPure.getVars
+  | .cfg c => c.blocks.flatMap fun (_, blk) =>
+    blk.cmds.flatMap Imperative.HasVarsPure.getVars
+
+/-- Is this body empty (abstract)? -/
+def Procedure.Body.isEmpty : Procedure.Body → Bool
```
PROgram52bc: The interface needs to be refined to distinguish between abstract procedures and unstructured body.

StrataTest/Transform/ProcedureInlining.lean
```diff
@@ -214,19 +214,59 @@ def alphaEquivStatement (s1 s2: Core.Statement) (map:IdMap)
 
 end
 
-private def alphaEquiv (p1 p2:Core.Procedure):Except Format Bool := do
-  if p1.body.length ≠ p2.body.length then
-    .error (s!"# statements do not match: in {p1.header.name}, "
-        ++ s!"inlined fn one has {p1.body.length}"
-        ++ s!" whereas the answer has {p2.body.length}")
+private def getStmts (p : Core.Procedure) : List Core.Statement :=
```
PROgram52bc: remove this method

StrataTest/Languages/Boole/global_readonly_call.lean
```diff
@@ -66,7 +66,7 @@ private def callHelper (p : Strata.Program) : Except String (List String) := do
   return cp.decls.filterMap fun d =>
     match d with
     | .proc p _ =>
-      p.body.findSome? fun
+      p.body.stmts.findSome? fun
```
PROgram52bc: use explicit pattern matching, and add comment for cfg case explaining it is to be implemented.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -762,6 +762,34 @@ Interpret a single procedure call.
...
+private def runCFG (cfg : Core.DetCFG) (fuel : Nat) (env : Env)
+    (ops : Imperative.RunOps Expression Command Env) : Env :=
+  go cfg.entry fuel env
+where
+  go (label : String) (fuel : Nat) (env : Env) : Env :=
+    match fuel with
+    | 0 => env
+    | fuel' + 1 =>
+      match cfg.blocks.lookup label with
+      | none => env
```
PROgram52bc: return an error here

Strata/Languages/Core/WF.lean
PROgram52bc: Done — removed .stmts, WFProcedureProp now uses bodyIsStructured premise with case-split fields.

Strata/Languages/Core/Procedure.lean
PROgram52bc: Done — replaced isEmpty with isAbstract/isStructured/isCfg to distinguish abstract procedures from CFG bodies.

StrataTest/Transform/ProcedureInlining.lean
PROgram52bc: Done — removed the unused getStmts helper.

StrataTest/Languages/Boole/global_readonly_call.lean
PROgram52bc: Done — using explicit pattern matching with comment explaining CFG case.

Strata/Languages/Core/WF.lean
```diff
@@ -138,15 +138,16 @@ structure WFAxiomDeclarationProp (p : Program) (f : Axiom) : Prop where
 structure WFDistinctDeclarationProp (p : Program) (l : Expression.Ident) (es : List (Expression.Expr)) : Prop where
 
 structure WFProcedureProp (p : Program) (d : Procedure) : Prop where
```
PROgram52bc, line 146: TODO: add WF properties for unstructured programs:   
-- * verify block labels are unique
  -- * all variables used are declared/initialized
  -- * target labels of transfer commands exist

Strata/Transform/ProcedureInlining.lean
```diff
@@ -94,13 +94,14 @@ private def renameAllLocalNames (c:Procedure)
   let proc_name := c.header.name.name
 
   -- Make a map for renaming local variables
-  let lhs_vars := List.flatMap (fun (s:Statement) => s.definedVars) c.body
+  let bodyStmts := match c.body with | .structured ss => ss | .cfg _ => []
```
PROgram52bc, line 100: Should extract local names from unstructured body as well. Implement this and add comments explaining although ProcedureInlining does not support unstructured bodies, this could be a generic facility that support unstructured bodies.

StrataTest/Languages/Core/Examples/Loops.lean
```diff
@@ -15,7 +15,7 @@ def singleCFG (p : Program) (n : Nat) : Imperative.CFG String
   let corePgm : Core.Program := TransM.run Inhabited.default (translateProgram p) |>.fst
   let proc := match corePgm.decls[n]? with
               | .some (.proc p _) => p | _ => Inhabited.default
-  Imperative.stmtsToCFG proc.body
+  Imperative.stmtsToCFG (match proc.body with | .structured ss => ss | .cfg _ => [])
```
PROgram52bc: Should return the CFG as is if the program is already unstructured.

StrataTest/Backends/CBMC/GOTO/E2E_CoreToGOTO.lean
```diff
@@ -350,7 +350,7 @@ private def coreToGotoJsonWithSummary (p : Strata.Program) (summary : String) :
   let Env := Lambda.TEnv.default
   let procs := cprog.decls.filterMap fun d => d.getProc?
   let p := procs[0]!
-  let p' : Core.Procedure := { p with body := injectPropertySummary p.body summary }
+  let p' : Core.Procedure := { p with body := .structured (injectPropertySummary (match p.body with | .structured ss => ss | .cfg _ => []) summary) }
```
PROgram52bc: add comment saying we could split the transformation into two-stage. For now, if body is unstructured, transformation from unstructured -> GOTO is left unimplemented.

PROgram52bc: Returns an error if the procedure is unstructured.

Strata/Transform/ProcBodyVerify.lean
```diff
@@ -88,8 +88,9 @@ open Core Imperative Transform
   -- Convert preconditions to assumes
   let assumes := requiresToAssumes proc.spec.preconditions
 
+  let bodyStmts := match proc.body with | .structured ss => ss | .cfg _ => []
```
PROgram52bc: return an error if the body is a cfg, since we expect structured input programs. Also add a comment explaining why this is an error.

Strata/Transform/LoopElim.lean
```diff
@@ -241,8 +241,11 @@ def loopElim (p : Program) : Program × Statistics :=
   let (decls, stats) := p.decls.foldl (fun (acc, stats) d =>
     match d with
     | .proc proc md =>
-      let (body, st) := StateT.run (Block.removeLoopsM proc.body) {}
-      ((.proc { proc with body := body } md) :: acc, stats.merge st.statistics)
+      match proc.body with
+      | .structured ss =>
+        let (body, st) := StateT.run (Block.removeLoopsM ss) {}
+        ((.proc { proc with body := .structured body } md) :: acc, stats.merge st.statistics)
+      | .cfg _ => ((.proc proc md) :: acc, stats)
```
PROgram52bc, line 249: add comments saying that there's no loop in cfg

Strata/Transform/CoreTransform.lean
```diff
@@ -332,10 +332,13 @@ def runProgram
           currentProcedureName := .some proc.header.name.1
         })
 
-        let (changed, new_body) ← runStmtsRec f proc.body
+        let bodyStmts ← match proc.body with
+          | .structured ss => pure ss
+          | .cfg _ => pure []
+        let (changed, new_body) ← runStmtsRec f bodyStmts
```
PROgram52bc, line 345: add a comment explaining we could add `runCFGRec` counterpart, and call it here based on the procedure body. However, since f returns Option (List Statement) and a CFG block holds List Command, we need to unwrap each returned statement as a .cmd. If f ever returns nested       
  structures (blocks, if-then-else, loops), those can't be represented in a basic block.

Strata/Languages/Core/Procedure.lean
```diff
@@ -306,21 +354,42 @@ def Procedure.modifiedVars (p : Procedure) : List Expression.Ident :=
...
+    | .structured ss => HasVarsImp.definedVars ss
+    | .cfg cfgBody => cfgBody.blocks.flatMap fun (_, blk) =>
+        blk.cmds.filterMap fun | .cmd (.init n _ _ _) => some n | _ => none
+  modifiedVars b := match b with
+    | .structured ss => HasVarsImp.modifiedVars ss
+    | .cfg cfgBody => cfgBody.blocks.flatMap fun (_, blk) =>
+        blk.cmds.filterMap fun
+          | .cmd (.set n _ _) => some n
+          | .cmd (.init n _ _ _) => some n
+          | _ => none
```
PROgram52bc: Define a HasVarsImp instance on `DetCFG` and use it here.

Strata/Languages/Core/Procedure.lean
```diff
@@ -306,21 +354,42 @@ def Procedure.modifiedVars (p : Procedure) : List Expression.Ident :=
...
+
 instance : HasVarsImp Expression Procedure where
   definedVars := Procedure.definedVars
   modifiedVars := Procedure.modifiedVars
 
 def Procedure.eraseTypes (p : Procedure) : Procedure :=
-  { p with body := Statements.eraseTypes p.body, spec := p.spec }
+  let body' := match p.body with
+    | .structured ss => .structured (Statements.eraseTypes ss)
+    | .cfg c => .cfg c
```
PROgram52bc: define `DetCFG.eraseTypes` and call it here.

Strata/Languages/Core/Procedure.lean
```diff
@@ -306,21 +354,42 @@ def Procedure.modifiedVars (p : Procedure) : List Expression.Ident :=
...
+  { p with body := body', spec := p.spec }
 
-/-- Remove all metadata from procedure. -/
 def Procedure.stripMetaData (p : Procedure) : Procedure :=
-  { p with body := Imperative.Block.stripMetaData p.body }
+  let body' := match p.body with
+    | .structured ss => .structured (Imperative.Block.stripMetaData ss)
+    | .cfg c => .cfg c
```
PROgram52bc: add `DetCFG.stripMetaData`, even if it doesn't have anything to strip right now. Comment on this.

Strata/Languages/Core/ProcedureEval.lean
```diff
@@ -113,7 +113,8 @@ def eval (E : Env) (p : Procedure) : Env × Statistics :=
       /- the assumptions from preconditions are set to have empty metadata  -/
       (.assume label check.expr check.md))
       p.spec.preconditions
-  let (ssEs, evalStats) := Statement.eval E old_g_subst (precond_assumes ++ p.body ++ postcond_asserts)
+  let bodyStmts := match p.body with | .structured ss => ss | .cfg _ => []
```
PROgram52bc: Return error on cfg case, add comment saying it would increase complexity and require control flow tracking. Leave it unimplemented now.

Strata/Transform/CoreTransform.lean
```diff
@@ -332,10 +332,19 @@ def runProgram
...
-        let (changed, new_body) ← runStmtsRec f proc.body
+        -- TODO: A `runCFGRec` counterpart could be added to apply command-level
+        -- transforms to CFG bodies. Since `f` returns `Option (List Statement)`
+        -- and a CFG block holds `List Command`, each returned statement would
+        -- need to be unwrapped as `.cmd`. If `f` returns nested structures
+        -- (blocks, if-then-else, loops), those can't be represented in a basic
+        -- block, so such cases would need to be rejected or left untransformed.
+        let bodyStmts ← match proc.body with
+          | .structured ss => pure ss
+          | .cfg _ => pure []
```
PROgram52bc: raise an error when trying to transform a cfg body

Strata/Transform/CoreSpecification.lean
```diff
@@ -143,7 +143,7 @@ structure ProcedureCorrect (proc : Procedure) (p : Program) : Prop where
     WF.WFProcedureProp p proc →
     ∀ (ρ₀ ρ' : Env Expression),
       ProcEnvWF proc ρ₀ →
-      CoreStepStar π φ (.stmts proc.body ρ₀) (.terminal ρ') →
+      CoreStepStar π φ (.stmts (match proc.body with | .structured ss => ss | .cfg _ => []) ρ₀) (.terminal ρ') →
```
PROgram52bc: Use `CoreCFGStepStar` for the unstructured body

Strata/Transform/CoreSpecification.lean
```diff
@@ -82,7 +82,7 @@ variable (φ : CoreEval → PureFunc Expression → CoreEval)
     (proc : Procedure)
     (a : Imperative.AssertId Expression) : Prop :=
   Imperative.Specification.AssertValidWhen (Specification.Lang.core π φ)
-    (ProcEnvWF proc) (Stmt.block "" proc.body #[]) a
+    (ProcEnvWF proc) (Stmt.block "" (match proc.body with | .structured ss => ss | .cfg _ => []) #[]) a
```
PROgram52bc: Define a separate Lang.coreCFG that uses CFGConfig as its CfgT, DetCFG as its StmtT, CoreCFGStepStar as its star, and a CFG-specific isAtAssert. Then AssertValidInProcedure would dispatch:
    - .structured ss → use Lang.core with Stmt.block "" ss #[]
    - .cfg cfg → use Lang.coreCFG with cfg

Strata/Languages/Core/Procedure.lean
```diff
@@ -306,21 +354,60 @@ def Procedure.modifiedVars (p : Procedure) : List Expression.Ident :=
...
+-- DetCFG.stripMetaData clears metadata from transfer commands. Currently,
+-- commands inside blocks carry no standalone metadata field, so only
+-- transfer metadata is stripped. This mirrors Block.stripMetaData for
+-- structured bodies and ensures a uniform interface.
+def DetCFG.stripMetaData (cfg : DetCFG) : DetCFG :=
+  { cfg with blocks := cfg.blocks.map fun (lbl, blk) =>
+      (lbl, { blk with transfer := match blk.transfer with
+                          | .condGoto p lt lf _ => .condGoto p lt lf .empty
+                          | .finish _ => .finish .empty }) }
+
```
PROgram52bc, line 405: move `DetCFG.eraseTypes` and `DetCFG.stripMetaData` to `BasicBlock.lean`

Strata/Languages/Core/Procedure.lean
```diff
@@ -306,21 +354,60 @@ def Procedure.modifiedVars (p : Procedure) : List Expression.Ident :=
...
   getVars := Procedure.getVars
 
+instance : HasVarsPure Expression Procedure.Body where
+  getVars := Procedure.Body.getVars
+
+instance : HasVarsImp Expression DetCFG where
+  definedVars cfg := cfg.blocks.flatMap fun (_, blk) =>
+    blk.cmds.flatMap Command.definedVars
+  modifiedVars cfg := cfg.blocks.flatMap fun (_, blk) =>
+    blk.cmds.flatMap Command.modifiedVars
```
PROgram52bc, line 379: move this to `BasicBlock.lean`

Strata/Languages/Core/Procedure.lean
PROgram52bc: Done — .stmts removed in earlier commit.

Strata/Languages/Core/StatementEval.lean
PROgram52bc: Done — runCFG now returns an error via CmdEval.updateError when block not found.

Strata/Languages/Core/WF.lean
PROgram52bc: Done — added the TODO comment.

Strata/Transform/ProcedureInlining.lean
PROgram52bc: Done — lhs_vars now extracts defined variables from CFG blocks via Command.definedVars, with a comment explaining the generic facility.

StrataTest/Languages/Core/Examples/Loops.lean
PROgram52bc: Done — singleCFG now returns the CFG as-is for unstructured bodies.

Strata/Transform/ProcBodyVerify.lean
PROgram52bc: Done — throws a DiagnosticModel error with a comment explaining why CFG bodies require a different verification strategy.

Strata/Transform/LoopElim.lean
PROgram52bc: Done — added comment.

Strata/Transform/CoreTransform.lean
PROgram52bc: Done — added comment explaining the runCFGRec counterpart and the unwrapping constraint.

Strata/Languages/Core/Procedure.lean
PROgram52bc: Done — defined HasVarsImp instance on DetCFG; Body instance now delegates to it.

Strata/Languages/Core/Procedure.lean
PROgram52bc: Done — defined DetCFG.eraseTypes that erases types from commands and transfer conditions.

Strata/Languages/Core/Procedure.lean
PROgram52bc: Done — added DetCFG.stripMetaData (delegates to generic CFG.stripDetMetaData in BasicBlock.lean). Comment explains only transfer metadata is stripped currently.

Strata/Languages/Core/ProcedureEval.lean
PROgram52bc: Done — returns error with comment explaining symbolic eval of CFGs would require control-flow-following with path merging.

Strata/Transform/CoreTransform.lean
PROgram52bc: Done — runProgram now throws an error on CFG bodies instead of returning [].

Strata/Transform/CoreSpecification.lean
PROgram52bc: Done — postconditionsValid now uses CoreStepStar with the structured body (extracted via h_body_match). AssertValidInProcedure dispatches to Lang.coreCFG for CFG bodies.

Strata/Transform/CoreSpecification.lean
PROgram52bc: Done — defined Lang.coreCFG with CoreCFGSpecConfig, coreCFGIsAtAssert, and CoreCFGStepStar. AssertValidInProcedure dispatches: .structured → Lang.core, .cfg → Lang.coreCFG.

Strata/Languages/Core/Procedure.lean
PROgram52bc: Partially addressed — generic stripMetaData moved to BasicBlock.lean (DetTransferCmd.stripMetaData, DetBlock.stripMetaData, CFG.stripDetMetaData). DetCFG.eraseTypes remains in Procedure.lean because it depends on Core-specific Command.eraseTypes/LExpr.eraseTypes which aren't available in BasicBlock.lean.

Strata/Languages/Core/Procedure.lean
PROgram52bc: Partially addressed — the generic strip logic (DetTransferCmd.stripMetaData, DetBlock.stripMetaData, CFG.stripDetMetaData) was moved to BasicBlock.lean. The HasVarsImp instance must stay in Procedure.lean because it uses Command.definedVars/modifiedVars which are Core-specific types not available in BasicBlock.lean.

Strata/Transform/CoreSpecification.lean
```diff
@@ -138,12 +184,16 @@ variable (φ : CoreEval → PureFunc Expression → CoreEval)
...
+      For structured bodies, termination is witnessed by `CoreStepStar`.
+      For CFG bodies, use `CoreCFGStepStar` (via `Lang.coreCFG`). -/
   postconditionsValid :
     WF.WFProcedureProp p proc →
     ∀ (ρ₀ ρ' : Env Expression),
       ProcEnvWF proc ρ₀ →
-      CoreStepStar π φ (.stmts proc.body ρ₀) (.terminal ρ') →
+      CoreStepStar π φ
+        (.stmts (match proc.body with | .structured ss => ss | .cfg _ => []) ρ₀)
+        (.terminal ρ') →
```
PROgram52bc: This looks wrong. Should use CoreCFGStepStar for unstructured procedure bodies.

PROgram52bc (commented):
## Code Review

Overall this is a well-structured PR that cleanly extends the AST with CFG bodies. The grammar design, parser, formatter, and test coverage are solid. I have a few concerns, mostly around soundness in the specification layer and a body-erasure bug in type-checking.

### Summary of issues:

1. **[Bug] Type-checker erases CFG bodies** — `ProcedureType.lean` extracts `[]` for CFG bodies, type-checks the empty list, then wraps the result back as `.structured annotated_body`. This silently replaces the CFG body with an abstract procedure.

2. **[Soundness] `postconditionsValid` is vacuous for CFGs** — The `match proc.body with | .structured ss => ss | .cfg _ => []` in `ProcedureCorrect` means postcondition validity for CFG procedures is trivially satisfied (empty statement list always terminates immediately).

3. **[Silent data loss] Nondeterministic goto drops labels** — `translateTransfer` handles `goto l1, l2;` but silently ignores any labels after the second.

4. **Minor issues** — duplicate comment, garbled docstring, silent non-error in `runCFG`.

See inline comments for details.

Strata/Languages/Core/ProcedureType.lean
```diff
@@ -110,8 +111,16 @@ def typeCheck (C : Core.Expression.TyContext) (Env : Core.Expression.TyEnv) (p :
...
-  let (annotated_body, finalEnv) ← Statement.typeCheck C envAfterPostconds p (.some proc) proc.body
+  -- Type check body.
+  let bodyStmts := match proc.body with
+    | .structured ss => ss
+  -- For now, we skip checking CFG bodies
+  -- potential TODOs for CFGs:
+  -- * verify block labels are unique
+  -- * all variables used are declared/initialized
+  -- * target labels of transfer commands exist
+    | .cfg _ => []
```
PROgram52bc: **[Bug]** This silently erases CFG bodies. When `proc.body` is `.cfg _`, you extract `[]`, type-check that empty list, then at line 135 wrap it back as `.structured annotated_body`. The procedure exits type-checking as an abstract (bodyless) procedure, losing the entire CFG.

Consider either:
- Returning early for CFG bodies (skip type-checking but preserve the body as-is)
- Or implementing CFG-specific checks and returning `.cfg` unchanged

```lean
let (annotated_body, finalEnv) ← match proc.body with
  | .structured ss => Statement.typeCheck C envAfterPostconds p (.some proc) ss
  | .cfg _ => pure ([], envAfterPostconds) -- but then preserve original body below
...
body := match proc.body with
  | .structured _ => .structured annotated_body
  | .cfg c => .cfg c
```

Strata/Transform/CoreSpecification.lean
```diff
@@ -138,12 +184,16 @@ variable (φ : CoreEval → PureFunc Expression → CoreEval)
...
+  /-- (2) The postconditions hold on termination.
+      For structured bodies, termination is witnessed by `CoreStepStar`.
+      For CFG bodies, use `CoreCFGStepStar` (via `Lang.coreCFG`). -/
   postconditionsValid :
     WF.WFProcedureProp p proc →
     ∀ (ρ₀ ρ' : Env Expression),
       ProcEnvWF proc ρ₀ →
-      CoreStepStar π φ (.stmts proc.body ρ₀) (.terminal ρ') →
+      CoreStepStar π φ
+        (.stmts (match proc.body with | .structured ss => ss | .cfg _ => []) ρ₀)
```
PROgram52bc: This looks wrong. Should use CoreCFGStepStar for unstructured procedure bodies.

Strata/Transform/PrecondElim.lean
```diff
@@ -87,19 +87,12 @@ so expression-level metadata carries no source location. We therefore inherit
...
   let wfObs := Lambda.collectWFObligations F e
-  -- Strip propertySummary: the enclosing statement's user-facing message
-  -- (e.g., a Python assert message) should not propagate to generated
-  -- precondition checks for called functions.
   let md := md.eraseAllElems Imperative.MetaData.propertySummary
-  -- Use modulo to cycle the precondition index correctly across call sites.
-  -- For nested calls like SafeSDiv(SafeSDiv(x,y),z), obligations arrive as
-  -- [inner-0, inner-1, outer-0, outer-1] with the same funcName throughout.
-  -- Without modulo, the index would be 0,1,2,3 instead of 0,1,0,1.
-  let (_, _, result) := wfObs.foldl (init := ("", 0, ([] : List Statement)))
```
PROgram52bc, line 103: Preserve comments here

Strata/Transform/CoreSpecification.lean
PROgram52bc: **[Soundness]** `postconditionsValid` is vacuous for CFG bodies. The match `(match proc.body with | .structured ss => ss | .cfg _ => [])` produces `[]` for CFGs, so `CoreStepStar π φ (.stmts [] ρ₀) (.terminal ρ')`  is immediately satisfied — postconditions trivially hold for any CFG procedure.

Since you already defined `CoreBodyExec` which unifies both cases, consider using it here:

```lean
postconditionsValid :
    WF.WFProcedureProp p proc →
    ∀ (ρ₀ : Env Expression) (σ_final : CoreStore) (failed : Bool),
      ProcEnvWF proc ρ₀ →
      CoreBodyExec π φ proc.body ρ₀.store ρ₀.eval σ_final failed →
      ...
```

Alternatively, if CFG correctness proofs are out of scope for now, gate this behind `WFProcedureProp.bodyIsStructured` and add a comment explaining that postcondition validity for CFGs is deferred.

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -1602,13 +1602,128 @@ def translateBlockCommand (p : Program) (bindings : TransBindings) (op : Operati
...
+  match op.name with
+  | q`Core.transfer_goto =>
+    let .seq _ .comma labels := op.args[0]!
+      | TransM.error s!"translateTransfer goto expected comma-sep labels"
+    match labels.toList with
+    | [] => return .finish
+    | [l] =>
+      let label ← translateIdent String l
+      return .condGoto (Lambda.LExpr.boolConst () Bool.true) label label
+    | l1 :: l2 :: _ =>
```
PROgram52bc: **[Silent data loss]** The `l1 :: l2 :: _` pattern silently drops any labels after the second. Since the grammar (`CommaSepBy Ident`) accepts `goto a, b, c;`, a user writing 3+ targets would get no error — labels are just ignored.

Since `DetTransferCmd.condGoto` only supports two targets, either:
1. Emit a translation error here: `if labels.toList.length > 2 then TransM.error "goto with more than 2 targets not supported"`
2. Or restrict the grammar to accept at most 2 identifiers

Strata/Languages/Core/ProcedureType.lean
```diff
@@ -110,8 +111,16 @@ def typeCheck (C : Core.Expression.TyContext) (Env : Core.Expression.TyEnv) (p :
 
   -- Type check body.
   -- Note that `Statement.typeCheck` already reports source locations in
-  -- error messages.
-  let (annotated_body, finalEnv) ← Statement.typeCheck C envAfterPostconds p (.some proc) proc.body
+  -- Type check body.
```
PROgram52bc: **[Nit]** Duplicate comment — `-- Type check body.` appears on both line 112 and line 114. The second occurrence should be removed (it replaced the old `-- error messages.` line).

Strata/Languages/Core/StatementEval.lean
```diff
@@ -762,6 +762,34 @@ Interpret a single procedure call.
...
+        let cmdStmts := blk.cmds.map (Imperative.Stmt.cmd ·)
+        match Imperative.runStmt ops fuel' (.stmts cmdStmts env) with
+        | .terminal env' =>
+          match blk.transfer with
+          | .finish _ => env'
+          | .condGoto cond lt lf _ =>
+            match ops.evalExpr env' cond with
+            | some (.boolConst _ true) => go lt fuel' env'
+            | some (.boolConst _ false) => go lf fuel' env'
+            | _ => env'
```
PROgram52bc: **[Bug]** When the condition does not evaluate to a boolean constant, `runCFG` silently returns the current env with no error. This means a malformed or unevaluable branch condition causes execution to silently "succeed" at that point.

Should report an error:
```lean
| _ => CmdEval.updateError env' (.Misc s!"runCFG: branch condition in block '{label}' did not evaluate to a boolean")
```

Strata/Languages/Core/StatementEval.lean
```diff
@@ -762,6 +762,34 @@ Interpret a single procedure call.
...
+Execute a CFG by following control flow from the entry block.
+    Returns the list of commands executed (as statements) for compatibility
+    with the structured interpreter. -/
+private def runCFG (cfg : Core.DetCFG) (fuel : Nat) (env : Env)
+    (ops : Imperative.RunOps Expression Command Env) : Env :=
+  go cfg.entry fuel env
+where
+  go (label : String) (fuel : Nat) (env : Env) : Env :=
+    match fuel with
+    | 0 => env
```
PROgram52bc: **[Bug]** Fuel exhaustion silently returns the current env with no error indication. This means infinite loops in CFG bodies silently "succeed" — the caller sees a normal result.

Compare with `Command.runCall` which sets `.OutOfFuel` on fuel exhaustion. Should be:
```lean
| 0 => CmdEval.updateError env (.Misc s!"runCFG: fuel exhausted (possible infinite loop)")
```
Or use the `.OutOfFuel` error variant if available.

Strata/Languages/Core/ObligationExtraction.lean
```diff
@@ -92,6 +98,16 @@ def extractFromStatements
   extractGo pathConditions ss #[]
 end
 
+/-- Extract proof obligations from a deterministic CFG by walking all blocks. -/
+def extractFromDetCFG (pc : PathConditions Expression) (cfg : DetCFG)
+    : Except String (Array (ProofObligation Expression)) :=
+  let obs := cfg.blocks.foldl (init := #[]) fun acc (_, blk) =>
+    let (_, blockObs) := blk.cmds.foldl (init := (pc, #[])) fun (curPc, obs) cmd =>
+      let (newPc, newObs) := extractFromCmd curPc cmd
+      (newPc, obs ++ newObs)
```
PROgram52bc, line 112: **[Semantic gap]** `extractFromDetCFG` restarts path conditions from the global `pc` for each block independently. Assumes within one block do not propagate to successor blocks — so if block A has `assume x > 0` followed by a goto to block B, the assert in B won't have `x > 0` in its path conditions.

This is conservative (no false negatives for verification), but it means proof obligations for CFG asserts will be harder to discharge than necessary. Worth a TODO noting that a proper dominator-based path-condition propagation would improve precision.

Strata/Languages/Core/WF.lean
```diff
@@ -137,16 +137,21 @@ structure WFAxiomDeclarationProp (p : Program) (f : Axiom) : Prop where
 
 structure WFDistinctDeclarationProp (p : Program) (l : Expression.Ident) (es : List (Expression.Expr)) : Prop where
 
+-- TODO: add WF properties for unstructured programs:
+--   * verify block labels are unique
```
PROgram52bc, line 143: **[Design note]** `bodyIsStructured` makes `WFProcedureProp` unconstructible for CFG procedures. This means `ProcedureCorrect` (which requires `WFProcedureProp` in `postconditionsValid`) can never be proven for CFG procedures.

This is fine as a temporary measure, but worth an explicit comment here noting that CFG well-formedness is deferred and what the plan is (e.g., a separate `WFCFGProcedureProp` or relaxing this field to `Option`).

Strata/Languages/Core/ProcedureType.lean
```diff
@@ -30,8 +30,9 @@ private def checkNoDuplicates (proc : Procedure) (sourceLoc : FileRange) :
 
 private def checkModificationRights (proc : Procedure) (sourceLoc : FileRange) :
     Except DiagnosticModel Unit := do
-  let modifiedVars := (Imperative.Block.modifiedVars proc.body).eraseDups
-  let definedVars := (Imperative.Block.definedVars proc.body).eraseDups
+  let bodyStmts ← proc.body.getStructured.mapError (fun e => DiagnosticModel.withRange sourceLoc f!"{e}")
```
PROgram52bc: Is this necessary?

Strata/Transform/CoreSpecification.lean
```diff
@@ -138,12 +184,20 @@ variable (φ : CoreEval → PureFunc Expression → CoreEval)
...
+        (.terminal ρ') →
       (∀ (label : CoreLabel) (check : Procedure.Check),
         (label, check) ∈ proc.spec.postconditions.toList →
         check.attr = Procedure.CheckAttr.Default →
```
tautschnig: **Soundness gap that survived the resolution of threads 28/30/32.** The match
```lean
CoreStepStar π φ
  (.stmts (match proc.body with | .structured ss => ss | .cfg _ => []) ρ₀)
  (.terminal ρ')
```
reduces to `CoreStepStar π φ (.stmts [] ρ₀) (.terminal ρ')` when `proc.body = .cfg _`. That derivation forces `ρ' = ρ₀` immediately (via `step_stmts_nil` → `terminal`), so the conjunction below holds trivially given `ProcEnvWF proc ρ₀` — *no claim about the CFG body is ever checked*. Combined with `WFProcedureProp` being unconstructible for CFGs (thread 38: `bodyIsStructured` can't be satisfied when `body = .cfg _`), the CFG branch of `ProcedureCorrect` is vacuous in two independent ways, which is invisible to a caller who reads the signature.

`AssertValidInProcedure` got routed to `Lang.coreCFG` correctly (thread 25) — asserts inside the CFG are fine. It's specifically postconditions on the procedure boundary that are not.

Pick one of these before merge (I'd prefer (1) or (2); (3) is a minimum):

**1. Route postconditions through `CoreBodyExec`** — the infrastructure is already in place. The TODO at :187-193 cites "`CoreBodyExec` only exposes terminal store/hasFailure, not eval" as the blocker; expose eval by either enriching `CoreBodyExec` with a `δ'` or using `CoreCFGStepStar` directly with the CFG's env. This is the Right Thing and matches thread 24's ask.

**2. Split into two fields:**
```lean
postconditionsValid_structured :
  ∀ ss, proc.body = .structured ss →
  WF.WFProcedureProp p proc →
  ∀ (ρ₀ ρ' : Env Expression),
    ProcEnvWF proc ρ₀ →
    CoreStepStar π φ (.stmts ss ρ₀) (.terminal ρ') → …
postconditionsValid_cfg :
  ∀ cfg, proc.body = .cfg cfg →
  ∀ (ρ₀ ρ' : Env Expression),
    ProcEnvWF proc ρ₀ →
    CoreCFGStepStar π φ cfg
      (.cont cfg.entry ρ₀.store false)
      (.terminal ρ'.store ρ'.hasFailure) → …
```
so the CFG case has a *stated but unproven* slot to fill in a future PR. `CoreBodyExec` could still be the eventual unifier.

**3. Minimum: add an explicit `sorry` / `axiom` for the CFG case so the hole is visible.** A doc-only TODO, as the current code has, is easy to miss when someone adds a caller that depends on `postconditionsValid` holding for both body kinds.

Concrete suggestion for (2):
```suggestion
  /-- (2a) The postconditions hold on termination for *structured* bodies.
      CFG-bodied procedures are covered by `postconditionsValid_cfg` below. -/
  postconditionsValid_structured :
    WF.WFProcedureProp p proc →
    ∀ ss, proc.body = .structured ss →
    ∀ (ρ₀ ρ' : Env Expression),
      ProcEnvWF proc ρ₀ →
      CoreStepStar π φ (.stmts ss ρ₀) (.terminal ρ') →
      (∀ (label : CoreLabel) (check : Procedure.Check),
        (label, check) ∈ proc.spec.postconditions.toList →
        check.attr = Procedure.CheckAttr.Default →
        ρ'.eval ρ'.store check.expr = some HasBool.tt) ∧
      ρ'.hasFailure = Bool.false
  /-- (2b) Postconditions hold on termination for *CFG* bodies.
      TODO: currently unimplemented — any CFG procedure satisfies this
      trivially because the premise is False. When CFG verification lands,
      replace with a `CoreCFGStepStar`-based formulation. -/
  postconditionsValid_cfg : True
```

Strata/Languages/Core/StatementEval.lean
```diff
@@ -762,6 +762,35 @@ Interpret a single procedure call.
...
+        | .terminal env' =>
+          match blk.transfer with
+          | .finish _ => env'
+          | .condGoto cond lt lf _ =>
+            match ops.evalExpr env' cond with
+            | some (.boolConst _ true) => go lt fuel' env'
+            | some (.boolConst _ false) => go lf fuel' env'
+            | _ => CmdEval.updateError env'
+                (.Misc s!"runCFG: branch condition in block '{label}' did not evaluate to a boolean")
+        | _ => env
```
tautschnig: **Silent fallthrough on non-terminal inner result.** When `Imperative.runStmt ops fuel' (.stmts cmdStmts env)` returns anything other than `.terminal env'` (e.g. `.stmts remaining env''` when the inner fuel is exhausted mid-block, or `.exiting` from an unexpected `exit` statement), this branch returns the *input* env silently.

Consequence in practice: if a block's command list is long enough to exhaust the inner `fuel'`, `runCFG` reports success with the pre-block env — the block's effects are lost with no diagnostic. Compare the sister cases on lines 786 (`block not found` → error) and 789-790 (`fuel exhausted` → error) which do surface the failure.

```suggestion
        match Imperative.runStmt ops fuel' (.stmts cmdStmts env) with
        | .terminal env' =>
          match blk.transfer with
          | .finish _ => env'
          | .condGoto cond lt lf _ =>
            match ops.evalExpr env' cond with
            | some (.boolConst _ true) => go lt fuel' env'
            | some (.boolConst _ false) => go lf fuel' env'
            | _ => CmdEval.updateError env'
                (.Misc s!"runCFG: branch condition in block '{label}' did not evaluate to a boolean")
        | _ =>
          CmdEval.updateError env
            (.Misc s!"runCFG: block '{label}' did not reach terminal (possibly inner-loop fuel exhaustion or unexpected exit)")
```

This mirrors the existing error-surfacing pattern in the other arms and pins down behaviour if a future edit adds early-exits / labeled exits that bubble through `runStmt`.

Strata/Languages/Core/ObligationExtraction.lean
```diff
@@ -92,6 +98,20 @@ def extractFromStatements
...
+    acc ++ blockObs
+  .ok obs
+
 /-- Extract proof obligations from a program. Axioms become global assumptions
```
tautschnig, line 116: Thread 37 marked as resolved-with-TODO — I'd refine the comment slightly. The current wording says:

> Assumes within one block do not propagate to successor blocks. This is conservative (no false negatives) but means obligations may be harder to discharge than necessary.

"No false negatives" is true *at the verifier's claim level* (no obligation is dropped). But from the user's perspective, an obligation in block B that needs `assume x > 0` from block A to discharge *will fail* — which the user experiences as a false alarm (the tool says "can't prove" for a property that actually holds).

Suggest rephrasing so the user-visible consequence is clear:

```suggestion
/-- Extract proof obligations from a deterministic CFG by walking all blocks.
    Path conditions restart from the global `pc` for each block independently:
    `assume` and `init` effects are visible within the same block but do not
    propagate to successor blocks.

    Verifier-level soundness: every obligation in the CFG is extracted, so this
    is conservative — no missed bugs.

    User-visible consequence: obligations in a block B that depend on an
    `assume` in a predecessor block A will fail to discharge, even when the
    property holds on every execution reaching B. The fix is dominator-based
    path-condition propagation along the control-flow graph — tracked as
    follow-up work. See #… -/
def extractFromDetCFG (pc : PathConditions Expression) (cfg : DetCFG)
    : Except String (Array (ProofObligation Expression)) :=
```

(Fill in a tracking issue number when you file one.) The point is: a reader skimming this should see both the soundness story and the precision story.

StrataTest/Languages/Core/Tests/CFGParseTests.lean
```diff
@@ -0,0 +1,197 @@
...
+      match p.body with
+      | .cfg c =>
+        IO.println s!"  CFG entry: {c.entry}, {c.blocks.length} blocks"
+        for (lbl, blk) in c.blocks do
+          let transferDesc := match blk.transfer with
+            | .condGoto _ l1 l2 _ => if l1 == l2 then s!"goto {l1}" else s!"branch → {l1}/{l2}"
+            | .finish _ => "return"
+          IO.println s!"  Block '{lbl}': {blk.cmds.length} cmds, {transferDesc}"
+      | .structured _ => IO.println "  ERROR: expected CFG body"
+    | _ => pure ()
```
tautschnig, line 404: The parse tests cover syntactic shape. Two cases not exercised that would pin down the deferred end-to-end story, and would fail explicitly when the deferred work lands:

1. **Verification rejection test**: a `#guard_msgs` fixture confirming that `strata verify Examples/CFGSimple.core.st` (or `Examples/CFGNondet.core.st`) produces the expected diagnostic — today `procToVerifyStmt` throws `"procToVerifyStmt: expected structured body, got CFG"`. A test that matches that diagnostic guarantees the error path stays connected (doesn't regress to silent success), and serves as a natural `xfail`-style anchor: when VC generation for CFGs lands, the diagnostic changes and the author has to come back and update the snapshot.

2. **Type-check success test**: `Procedure.typeCheck` now walks CFG bodies only with the minimal `inp_mty_sig`/`out_mty_sig` annotations (it explicitly skips body type-checking — ProcedureType.lean:121). A fixture that (a) typechecks a CFG procedure, (b) verifies the resulting `proc.body` is still `.cfg cfg'` (not silently collapsed to `.structured []`, which was the bug thread 29 caught), would pin down the invariant.

Concrete shape for (1):

```lean
/-- info: error: procToVerifyStmt: expected structured body, got CFG -/
#guard_msgs (error, drop info) in
#eval (do
  let prog ← parseCoreText Examples.CFGSimple.source
  -- attempt verification, expect CFG-unsupported error
  …)
```

Also worth: a negative-path parse test where a CFG body is malformed (e.g., `goto a, b, c;` with 3+ targets) and confirming thread 33's fix emits the `"translateTransfer: goto with more than 2 targets is not supported"` error rather than silently dropping the extras.

Strata/Languages/Core/StatementSemantics.lean
```diff
@@ -285,6 +286,86 @@ inductive CoreStepStar
...
+    CoreBodyExec π φ (.structured ss) σ δ ρ'.store ρ'.hasFailure
+  | cfg :
+    CoreCFGStepStar π φ cfg
+      (.cont cfg.entry σ false)
+      (.terminal σ' failed) →
+    CoreBodyExec π φ (.cfg cfg) σ δ σ' failed
+
 inductive EvalCommand (π : String → Option Procedure) (φ : CoreEval → PureFunc Expression → CoreEval) : CoreEval →
   CoreStore → Command → CoreStore → Bool → Prop where
   | cmd_sem {δ σ c σ' f} :
```
tautschnig, line 331: `CoreBodyExec` is the right shape for a body-polymorphic execution predicate. Two observations:

1. **It's not used where it would matter most** — `Core.Specification.ProcedureCorrect.postconditionsValid` still uses `CoreStepStar` on a `.structured`-shaped argument (see the inline comment on `CoreSpecification.lean:205`). The PR's own TODO at CoreSpecification.lean:187-193 names this. Please file a tracking issue for the switchover, otherwise this constructor has one in-tree user (`EvalCommand.call_sem`) and risks bitrot.

2. **`CoreBodyExec` drops the terminal `eval`.** The structured case reduces to `CoreStepStar … (.terminal ρ')` and extracts `ρ'.store`/`ρ'.hasFailure`, discarding `ρ'.eval`. For contexts that need the final `eval` (post-funcDecl extensions), this loses information. If the intent is to support postconditions eventually, the constructor should expose `ρ'` (or add a `δ'` field), mirroring the structured-body shape that the existing `postconditionsValid` uses.

Proof opportunity to make the equivalence explicit for the structured case:

```lean
theorem CoreBodyExec.structured_iff_CoreStepStar
    (π φ ss σ δ σ' f) :
    CoreBodyExec π φ (.structured ss) σ δ σ' f ↔
    ∃ ρ', CoreStepStar π φ (.stmts ss ⟨σ, δ, false⟩) (.terminal ρ') ∧
          ρ'.store = σ' ∧ ρ'.hasFailure = f
```

Straightforward by constructor inversion. Once this is in place, the drop-in for `postconditionsValid` is much cleaner (route through `CoreBodyExec` and invoke the direction you need).

tautschnig (commented):
Two things that are classified as "CFG (deferred)" / "Proof (deferred)" in the table, but whose current code shape is load-bearing for what downstream clients currently believe about verification results — I think both deserve a more visible marker before merge, because the tool will *silently* accept CFG procedures with neither an error nor a warning that "your CFG procedure was not verified":

**1. `ProcedureCorrect.postconditionsValid` is vacuously true for CFG procedures.** At `CoreSpecification.lean:198-201`, the match
```lean
CoreStepStar π φ
  (.stmts (match proc.body with | .structured ss => ss | .cfg _ => []) ρ₀)
  (.terminal ρ')
```
yields `CoreStepStar π φ (.stmts [] ρ₀) (.terminal ρ')` for CFG procedures. That derivation forces `ρ' = ρ₀` immediately, so the conclusion
```lean
  (∀ label check, … → ρ'.eval ρ'.store check.expr = some HasBool.tt) ∧
  ρ'.hasFailure = Bool.false
```
becomes "postconditions hold on ρ₀ and ρ₀.hasFailure = false" — which is just a restatement of `ProcEnvWF` plus the no-failure hypothesis. Nothing about the CFG body was actually verified.

`AssertValidInProcedure` *is* correctly dispatched to `Lang.coreCFG` for CFG bodies (great — that came out of thread 25), so asserts inside the CFG are covered. But postconditions on the *procedure boundary* are not, and the stacked vacuity (plus `WFProcedureProp` being unconstructible for CFG per thread 38) makes this hole invisible to anyone who reads only `ProcedureCorrect`'s signature and assumes it means what it says for both body kinds.

Concrete asks (details inline):
- Introduce `ProcedureCorrect.postconditionsValid` specifically for CFG via `CoreBodyExec` / `CoreCFGStepStar`, *or*
- Add an explicit `postconditionsValid_CFG_is_vacuous` theorem as a stability marker so that the gap shows up at proof-development time if someone adds CFG-specific reasoning elsewhere, *or*
- At minimum, split the current `postconditionsValid` into a `_structured` and `_cfg := True` so a future CFG postcondition proof lands in an unambiguous slot rather than overwriting the old one.

**2. `runCFG` silently returns the current env on a non-terminal interpreter result.** At `StatementEval.lean:791` the final `| _ => env` arm catches the case where `Imperative.runStmt … (.stmts cmdStmts env)` returns `.stmts …` (remaining statements, i.e., inner fuel exhausted or not-fully-reducible). That silently discards the computation. Thread 36 (fuel exhaustion) was addressed only for the outer `runCFG` fuel; the inner `runStmt`'s incomplete-reduction path still falls through. Match needs to propagate the error. Inline.

**Other smaller items:**

- **`extractFromDetCFG` resets path conditions per block (thread 37, marked resolved-with-TODO).** The current comment says "conservative (no false negatives)". That's true in the VC-generation sense (no assert is dropped), but it *does* produce false alarms: `assume x > 0` in block A followed by `goto B` does not constrain the verifier in B, so an obligation in B that needs `x > 0` will fail to discharge. The comment is slightly misleading — flag the user-visible consequence.
- **Examples don't have end-to-end verification tests.** `Examples/CFGSimple.core.st` / `Examples/CFGNondet.core.st` only appear in `CFGParseTests.lean` (parse-shape checks). Since `procToVerifyStmt` throws on CFG, they cannot be `strata verify`-ed today. Worth a `#guard_msgs` fixture that confirms `strata verify <cfg example>` produces a clean diagnostic ("CFG verification not yet supported") rather than silently succeeding or crashing — this pins down the current state and flags when the deferred work lands.
- **`CoreBodyExec` is partially dead.** Introduced as the unifier for body execution, used in `EvalCommand.call_sem` (line 398 of StatementSemantics.lean), but *not* used in `postconditionsValid`. If the plan is to route `postconditionsValid` through it once the `eval` accessibility obstacle is resolved (per the TODO at CoreSpecification.lean:187-193), consider filing a tracking issue so the API isn't left half-wired.
- **`.match proc.body with | .structured ss => ss | .cfg _ => []` pattern appears in at least three places.** `CoreSpecification.lean:199`, `ProcBodyVerifyCorrect.lean:661-662` (the new `h_body_match`), `Core.lean:98-102`. An `Procedure.Body.structuredOrEmpty : Body → List Statement` helper would centralize the semantics-of-CFG-as-empty and make the vacuity visible at one source location. Trade-off: it also makes the vacuity easier to miss. I'd lean toward extracting + adding a short docstring saying "returns `[]` for CFG; this is the semantics-of-absence used in several structured-only passes pending proper CFG handling."

**Proof-coverage suggestions:**

Three small theorems that would make the soundness gap (and any future closing of it) explicit:

```lean
/-- CFG procedures satisfy `postconditionsValid` trivially because the
    structured-only predicate is instantiated on an empty statement list.
    This theorem is **not** a progress statement — it is a stability marker
    for the current (incomplete) verification. See TODO at :187-193. -/
theorem postconditionsValid_CFG_is_vacuous
    {π : String → Option Procedure}
    {φ : CoreEval → PureFunc Expression → CoreEval}
    (proc : Procedure) (_ : proc.body.isCfg)
    (p : Program)
    (h_wf : WF.WFProcedureProp p proc)
    (ρ₀ ρ' : Env Expression)
    (h_env : ProcEnvWF proc ρ₀)
    (h_step : CoreStepStar π φ
      (.stmts (match proc.body with | .structured ss => ss | .cfg _ => []) ρ₀)
      (.terminal ρ')) :
    ρ' = ρ₀ := by
  cases hb : proc.body with
  | structured => simp [Procedure.Body.isCfg, hb] at *
  | cfg _ => simp [hb] at h_step; …
```

```lean
/-- Round-trip for `DetCFG.stripMetaData`: semantics is unchanged by stripping. -/
theorem DetCFG.stripMetaData_preserves_eval (π φ cfg σ δ σ' f) :
    CoreCFGStepStar π φ cfg (.cont cfg.entry σ false) (.terminal σ' f) ↔
    CoreCFGStepStar π φ cfg.stripMetaData (.cont cfg.stripMetaData.entry σ false)
      (.terminal σ' f)
```

```lean
/-- `transformDetCFG` is a no-op on the CFG shape (only inserts assert
    commands, never changes control flow). -/
theorem transformDetCFG_preserves_entry_and_blocks (F cfg) :
    let (_, cfg') := transformDetCFG F cfg
    cfg'.entry = cfg.entry ∧
    cfg'.blocks.map (·.1) = cfg.blocks.map (·.1)
```

All three are `simp` / `rfl`-shaped once the constructors unfold. The first makes the current soundness hole provable-as-a-lemma (and hence impossible to forget when someone later extends the verifier).

**Refactor callouts:**

- `CoreEvalCmds` (StatementSemantics.lean:320-328) duplicates the list-of-commands fold that `EvalCommand` does statement-by-statement. Might be worth a general `CoreEvalCmds_of_EvalCommand` lemma so that properties of `EvalCommand` lift automatically.
- The CFG-related instances on `Procedure.Body` (`HasVarsImp`, `HasVarsPure`) were split so that the generic `HasVarsImp Expression DetCFG` lives next to the `Procedure.Body` instance in `Procedure.lean`. Thread 26/27 asked for `DetCFG.eraseTypes`/`stripMetaData` in `BasicBlock.lean`; the generic parts moved but `DetCFG.eraseTypes` remained in `Procedure.lean` (with the reason: dependency on `Command.eraseTypes`). Consider an intermediate file `Strata/Languages/Core/DetCFG.lean` so the Core-flavoured but generic-over-statement parts live together.

Strata/Languages/Core/ProcedureType.lean
```diff
@@ -111,7 +112,13 @@ def typeCheck (C : Core.Expression.TyContext) (Env : Core.Expression.TyEnv) (p :
   -- Type check body.
   -- Note that `Statement.typeCheck` already reports source locations in
   -- error messages.
-  let (annotated_body, finalEnv) ← Statement.typeCheck C envAfterPostconds p (.some proc) proc.body
+  -- For now, we skip checking CFG bodies. Potential TODOs for CFGs:
```
atomb: Given that we need to implement this placeholder otherwise, I think it makes sense to include type checking as part of this PR. It shouldn't be too tricky. The properties listed seem right, along with all expressions in blocks being type correct.

Strata/Languages/Core/StatementSemantics.lean
```diff
@@ -285,6 +286,86 @@ inductive CoreStepStar
     ----
     CoreStepStar π φ c₁ c₃
 
+/-- Evaluate the commands in a deterministic basic block, then transfer
```
atomb: Is there something misaligned with the definitions in `CFGSemantics` that means they can be directly used when instantiated for Core? If there's a way in which they don't work, let's see if we can generalize or adapt them so they can be used directly.

Strata/Transform/ProcedureInlining.lean
```diff
@@ -269,9 +277,16 @@ def inlineCallCmd
               Statement.set lhs_var (.fvar () out_var (.none)) md)
             outs_lhs_and_sig
 
+        -- Cannot inline unstructured (CFG) bodies into structured code.
+        -- CFG-level inlining is a separate, more complex pass that operates
+        -- entirely in the CFG domain (graph splicing).
+        let procBodyStmts ← match proc.body with
+        | .cfg _ => return .none
```
atomb: I'd prefer to generate an error here rather than silently inlining an empty body.

Strata/Languages/Core/DDMTransform/FormatCore.lean
```diff
@@ -869,6 +869,44 @@ partial def invariantsToCST {M} [Inhabited M]
     let restCST ← invariantsToCST rest
     pure (.consInvariants default labelAnn exprCST restCST)
 
+/-- Convert a `DetTransferCmd` to a CST statement (if-then-else for condGoto, skip for finish). -/
```
atomb: Now that we have concrete syntax for unstructured programs, we could use that for the results of the functions added here, instead of `Statement` and `Block`.

StrataTest/Languages/Core/Tests/CFGParseTests.lean
```diff
@@ -0,0 +1,197 @@
...
+    m := y;
+    goto done;
+  }
+  done: {
+    return;
+  }
+};
+"
+  for d in prog.decls do
+    match d with
```
atomb: This loop is essentially duplicated for all four tests and could be extracted into a named definition to be reused.

Strata/Languages/Core/DDMTransform/FormatCore.lean
PROgram52bc: Yeah. Thanks for spotting that. I have made some change.

Strata/Transform/ProcedureInlining.lean
PROgram52bc: Okay. Should be fixed.

StrataTest/Languages/Core/Tests/CFGParseTests.lean
PROgram52bc: Agreed. Fixed.

Strata/Languages/Core/StatementSemantics.lean
PROgram52bc: The short answer is that Lean doesn't like nested inductives that do not _directly_ use the parameterized inductive in its constructor. [Here](https://github.com/strata-org/Strata/compare/htd/unstructured-procedure...htd/unstructured-procedure-experiment) is an experiment explaining what would be needed to reuse `EvalDetBlock` in `CoreEvalDetBlock`.

PROgram52bc:
superseded by #1196 , targeting main2

PR #1019, opened 2026-04-22
Python -> Laurel translation: unmodeled method calls to havoc non-receiver arguments too

ssomayyajula:
*Description of changes:*

Havocs locals passed as arguments to unmodeled calls. The call is a black box that could have side effects on its arguments, and (non-composite) locals are not reachable via heap havoc + havoc'ing receiver.                                                                                                                                                                                                                                                                                 

Adds a test case (xs2.unknown_method_that_may_modify_arguments(ys)) to test_havoc_callee_after_hole_call exercising the new behavior.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1133, opened 2026-05-06
Lift unlabeled block expressions emitted by Python-to-Laurel

tautschnig:
he Python-to-Laurel translator emits unlabeled `Block` expressions in expression position for several patterns, including modeled method calls with precondition tag checks (`{ asserts...; Call }`) and unmodeled method calls (PR #1019: `{ havocs...; Hole }`). These patterns trigger `block expression should have been lowered in a separate pass` errors in `LaurelToCoreTranslator` for nested cases, because `LiftImperativeExpressions` was not fully lowering blocks/assignments embedded inside declare-with-init initializers.

`LaurelToCoreTranslator` does already handle several block-in-expression patterns natively:

  * `.Block [single]` — singleton blocks
  * `.Block (.Assign [.Declare ⟨name, ty⟩] init :: rest)` — translated as a let-binding (`app(abs(name, body), value)`)

so this PR is now scoped narrowly to make sure only valid expression blocks reach the translator, rather than collapsing block structure unconditionally.

Changes in `LiftImperativeExpressions.lean`:

1. **`transformExpr.Assign` for `.Declare` targets without substitution** now recursively transforms the initializer with a fresh prepend stack and reassembles state as `outerPrepends ++ innerPrepends ++ [liftedDecl]`, returning a `Var(.Local name)` reference as the expression's value. Previously the case did `return expr` early, leaving inner blocks / assignments / nondeterministic holes in the initializer unlowered. The `outerPrepends`-first ordering preserves correct execution order when a substitution-introducing snapshot decl (from a later-in-source assignment, processed earlier in the right-to-left traversal) is referenced from the transformed initializer.

2. **New helper `containsAssignmentOutsideUnlabeledBlock`** is used by the `.Assert` / `.Assume` cases of `transformStmt`. These cases previously bailed out whenever the condition contained any assignment; they now proceed when the assignment is confined to an unlabeled block — since the `.Block` case of `transformExpr` together with `onlyKeepSideEffectStmtsAndLast` will safely lift those side effects as prepends.

   The original `containsAssignment` is preserved for the ITE case, which correctly isolates branch-local prepends when a branch contains assignments.

3. **`transformExpr.Block`** is unchanged in its structural behavior: sub-expressions are recursively transformed, side effects are extracted via `onlyKeepSideEffectStmtsAndLast`, and the result is wrapped back as a Block. We deliberately do not collapse Block structure here so that singleton blocks and `var x := init`-prefixed blocks remain in shape for the translator's let-binding case.

A regression test (`nestedBlockInDeclInit`) covering the `{ var t := { x := 1; x }; t + 1 }` pattern is added to `LiftExpressionAssignmentsTest.lean`. Existing test expected outputs are kept consistent with this approach (singleton blocks `{ x }` / `{ y }` / `{ 0 }` / `{ t + 1 }` rather than unwrapped values).

Eliminates all `block expression should have been lowered` errors in the internal benchmark suite.


keyboardDrummer:
Hey, thanks for this PR. You're right that Laurel should support the code that PythonToLaurel is generating here.

One thing though, the existing error "block expression should have been lowered" is phrased too strongly. Sorry about that. LaurelToCoreTranslator is intended to support some blocks in expressions, namely the ones that can be translated to Core expressions. The intention is for a block like `{ var x: int := <expr1>; <expr2> }` to be modeled as a Core let binding.

Could we investigate making sure that only valid expression blocks make it to `LaurelToCoreTranslator`, instead of preventing all expression blocks from reaching there? Lifting all blocks and not having any let expressions seems like an overly aggressive transformation that won't benefit the performance/debugability of our pipeline. Let's look at the two problematic Python cases mentioned by this PR.

> The Python-to-Laurel translator emits unlabeled Block expressions in expression position for two common patterns:

> Modeled method calls with precondition tag checks: { asserts...; Call }

Can we lift the asserts? The Laurel->Core translator will handle a singleton block `{ Call }`

> Unmodeled method calls (PR https://github.com/strata-org/Strata/pull/1019): havocs receiver + Any-typed args, then yields a Hole: { havoc_stmt...; Hole }

Can you say more about what `havoc_stmt` looks like? If it's a destructive assignment, like `x := ?`, then it should have been lifted. If it's a variable declaration with (non-det) initializer, like `var x: int := ?`, then that's something that `LaurelToCoreTranslator` is intended to support. Note the case `| .Block (⟨ .Assign [⟨ .Declare ⟨name, ty ⟩, _source⟩] initializer, innerSrc⟩ :: rest) label => do ... `

Maybe there should be a separate case for a hole assignment in a block: `| .Block (⟨ .Assign [⟨ .Declare ⟨name, ty ⟩, _source⟩] .Hole, innerSrc⟩ :: rest) label => do ... `

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -344,8 +375,25 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
+          -- declare-assigns, drops plain assigns already processed), and
+          -- return the last element as the expression value. This lowers
+          -- the block structure so the Laurel-to-Core translator no longer
+          -- sees a Block in expression position. Pattern is commonly emitted
+          -- by PythonToLaurel: { asserts; Call } or { havoc; Hole }
+          -- (PR #1019 for unmodeled calls).
+          let _ ← onlyKeepSideEffectStmtsAndLast (head :: rest)
+          return (head :: rest).getLast (by simp)
+      else
+        return ⟨ .Block (← onlyKeepSideEffectStmtsAndLast newStmts) labelOption, source⟩
```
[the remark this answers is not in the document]
tautschnig: Fixed in 977afc1e: `transformExpr.Assign` for `.Declare` without substitution now recursively transforms `value`, then reassembles state as `outerPrepends ++ innerPrepends ++ [liftedDecl]` so any inner block / assignment / nondeterministic hole in the initializer is fully lowered before reaching the translator. `outerPrepends`-first preserves correct ordering when a substitution-introducing snapshot decl from a later-position prepend is referenced by the transformed initializer.

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
```diff
@@ -23,11 +23,20 @@ namespace Strata.Laurel
...
   var y: int := { assert x == 0; x := 1; x };
   assert y == 1
 };
+
+procedure condAssign(x: int)
+  opaque
+{
+  var y: int := 0;
+  var z: int := (if x > 0 then { y := y + 1 } else { 0 }) + y;
+  assert z == 2
```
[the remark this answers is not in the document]
tautschnig: Added in 977afc1e as procedure `nestedBlockInDeclInit` (`var y := { var t := { x := 1; x }; t + 1 }`), which fails on the prior commit and passes after the fix above.

tautschnig:
@keyboardDrummer:

Good point. In 977afc1e I reverted the wholesale `.Block` unwrapping so let-binding-shaped blocks (`{ var x := init; ...; e }`) and singletons reach `LaurelToCoreTranslator` intact.

Re: `havoc_stmt` — for the unmodeled-call pattern (`PythonToLaurel.translateCall`) it's a destructive `Assign [.Local recv] .Hole` over an existing local; after `EliminateHoles` runs the trailing hole becomes `hole(Unknown)`, so by lift time the pattern is `{ x := hole(Unknown); …; hole(Unknown) }`, which the existing `.Assign[.Local]` lift path reduces to a singleton `{ hole(Unknown) }` — the translator already handles it. The `{ asserts; Call }` case is similarly handled (asserts hoisted, residual `{ Call }` is a singleton). No new translator case needed.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -157,7 +157,38 @@ private def computeType (expr : StmtExprMd) : LiftM HighTypeMd := do
...
 
-/-- Check if an expression contains any assignments (recursively). -/
+/-- Check if an expression contains any assignments that are NOT inside an
+unlabeled block. Unlabeled blocks in expression position are lowered by the
+`.Block` case of `transformExpr` (prepends side-effects + returns the last
+element as the value), so assignments inside them do not need special
+rejection logic. This is used for assert/assume conditions, which should
+proceed with transformation even when an unlabeled block embeds an
+assignment (e.g. havoc from PR #1019's unmodeled call handling). -/
+def containsAssignmentOutsideUnlabeledBlock (expr : StmtExprMd) : Bool :=
```
ssomayyajula, line 169: Is the catch-all to false actually correct here? I'm having trouble understanding what happens if it returns a "false negative" to `transformExpr`. Is it "accidentally correct" since `transformExpr` has a catch-all that doesn't recurse into the same variants?

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
```diff
@@ -23,11 +23,28 @@ namespace Strata.Laurel
...
 def blockStmtLiftingProgram : String := r"
 procedure assertInBlockExpr()
+  opaque
 {
   var x: int := 0;
   var y: int := { assert x == 0; x := 1; x };
   assert y == 1
 };
+
+procedure condAssign(x: int)
```
keyboardDrummer: These tests, once fixed, and placed in a testing file that runs deductive verification, already pass in main:
```
procedure condAssign(x: int)
  opaque
{
  var y: int := 0;
  var z: int := (if x > 0 then { y := y + 1 } else { 0 }) + y;
  assert z == if x > 0 then 2 else 0
};

procedure nestedBlockInDeclInit()
  opaque
{
  var x: int := 0;
  var y: int := { var t: int := { x := 1; x }; t + 1 };
  assert y == 2
};
```

so I don't think the tests in this PR test its changes.

Could you put in the test file `T2_ImpureExpressions`, cases that are currently not passing on main but pass with the changes from this PR? The test file `LiftExpressionAssignmentsTest.lean` deserves to be renamed into `LiftExpressionAssignmentsIdiomaticityTest` since it is not meant to test the correctness but rather the quality of the expression lifting pass.

keyboardDrummer (changes requested):

keyboardDrummer:
> for the unmodeled-call pattern (`PythonToLaurel.translateCall`) it's a destructive `Assign [.Local recv] .Hole` over an existing local; after `EliminateHoles` runs the trailing hole becomes `hole(Unknown)`, so by lift time the pattern is `{ x := hole(Unknown); …; hole(Unknown) }`, which the existing `.Assign[.Local]` lift path reduces to a singleton `{ hole(Unknown) }` — the translator already handles it. The `{ asserts; Call }` case is similarly handled (asserts hoisted, residual `{ Call }` is a singleton). No new translator case needed.

If the holes are already covered, what cases do you still need block lifting for? Is it just assertions/assumptions that occur in expressions? Can we not lift those then? My feeling is that block lifting, while probably effective for correctness, is not the most idiomatic solution.

You can add 

```
  | .Assume cond =>
      let prepends ← takePrepends
      let seqCond ← transformExpr cond
      let argPrepends ← takePrepends
      modify fun s => { s with prependedStmts := argPrepends ++ [⟨.Assume seqCond, source⟩] ++ prepends }
      default

  | .Assert cond =>
      let prepends ← takePrepends
      let seqCond ← transformExpr cond.condition
      let argPrepends ← takePrepends
      modify fun s => { s with prependedStmts := argPrepends ++ [⟨.Assert { cond with condition := seqCond }, source⟩] ++ prepends }
      default
```

to `transformExpr` to lift those.

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
tautschnig: Both points addressed in f5bd0f71.

* `T2_ImpureExpressions.lean` now has three deductive-verification regression cases — `assertWithBlockExpr`, `assumeWithBlockExpr`, `nestedBlockInDeclInit` — that I verified all fail on `origin/main` with "block expression should have been lowered in a separate pass" and pass with this PR.

* `LiftExpressionAssignmentsTest.lean` is renamed to `LiftExpressionAssignmentsIdiomaticityTest.lean`, with the module docstring rewritten to make the quality-of-output purpose explicit. `condAssign` is dropped from that file — your observation is right, it already passes on main and isn't testing this PR.

tautschnig:
> If the holes are already covered, what cases do you still need block lifting for? Is it just assertions/assumptions that occur in expressions? Can we not lift those then? My feeling is that block lifting, while probably effective for correctness, is not the most idiomatic solution.
[...]

Good call — done in ad373f98. `transformExpr` now has `.Assert`/`.Assume` cases that recursively transform the condition and lift the assert/assume to the prepend stack with a leaf placeholder, replacing the special-casing in `onlyKeepSideEffectStmtsAndLast`. As a side benefit this also fixes a latent case my earlier draft missed — an `assert`/`assume` *inside* an expression-position block whose condition contains another unlabeled block with an assignment — covered by a new `nestedAssertInBlockExpr` regression test.

The `containsAssignmentOutsideUnlabeledBlock` guard is kept so a top-level destructive assignment in a condition (e.g. `assert (x := 2) == 2`) still surfaces the existing `destructive assignments are not supported in functions or contracts` diagnostic from `LaurelToCoreTranslator` rather than getting silently lifted.

StrataTest/Languages/Laurel/Examples/Fundamentals/T2_ImpureExpressions.lean
```diff
@@ -46,6 +46,57 @@ procedure conditionalAssignmentInExpression(x: int)
   }
 };
 
+// Regression for #1133: an unlabeled-block expression that contains a
+// destructive assignment must be admissible inside an `assert` condition.
```
keyboardDrummer, line 50: > // Regression for #1133: an unlabeled-block expression that contains a
// destructive assignment must be admissible inside an `assert` condition.

Why must that be admissible? What is the use-case? I was planning to update Resolution.lean so it explicitly rejects assignments to variables that occur inside assert/assume but were the variable was defined outside of it.

atomb:
Can you retarget this PR to `main2`?

PR #1134, opened 2026-05-06
InferHoleTypes: recover param types for datatype destructors/testers

tautschnig:
Holes in arguments to datatype destructors (e.g. `Any..as_ListAny!(<?>)`) and testers (e.g. `Any..isfrom_str(<?>)`) were left untyped after `InferHoleTypes`, producing "could not infer type" diagnostics and downstream `Unknown`-typed `$hole_N` functions.

The fix surfaces the parent datatype through the semantic model so `InferHoleTypes` can recover parameter types via `uniqueId`-based model lookups (per the Laurel steering doc, post-resolution passes should rely on model lookups, not textual names).

Key changes:

- New `ResolvedNode.datatypeDestructor` variant carrying the parent type name and field parameter, with corresponding updates to `ResolvedNodeKind`, `.kind`, `.getType`, and `SemanticModel.isFunction`.
- Datatype scope registration chains `defineNameCheckDup` calls so safe/unsafe destructors and tester/constructor names share a single `uniqueId`.
- `buildRefToDef` registers each constructor parameter as `.datatypeDestructor`.
- `calleeParamTypes` matches `.datatypeConstructor` and `.datatypeDestructor` directly, synthesizing the correct `.UserDefined` input type.
- `ResolvedNode.getType` now uses explicit pattern matches instead of a catch-all, so new constructors trigger compile errors rather than silently falling through.
- `SemanticModel.refToDef` is now `private`; external code uses `SemanticModel.get?` (returning `Option ResolvedNode`) or `SemanticModel.get` to look up nodes by `Identifier`, preventing arbitrary `Nat` key access.

Tests: three new `#guard_msgs` cases for holes inside destructors/testers, plus updated expected diagnostics in `ResolutionKindTests`. Existing tests pass.


Strata/Languages/Laurel/Resolution.lean
ssomayyajula: Why the catch-all here, why not go to `Unknown`?

ssomayyajula: This kind of freaks me out - because it's not a map from names keyed on their unique identifiers to resolved nodes, that implies that you could either pass an arbitrary nat (rather than the ones generated by resolution) or that there are ways to map identifiers to resolved nodes that do not involve `refToDef`. If you marked this private so that `get` were the only accessor, then I can see that being fine. But should `get` return `default` or error out?

MikaelMayer: 🤖 Good point. Replaced the catch-all with explicit matches for `staticProcedure`, `instanceProcedure`, `compositeType`, `constrainedType`, `datatypeDefinition`, and `typeAlias` that all return `Unknown`. This way new constructors will trigger a compile error instead of silently hitting a catch-all. See 15f2288d9.

MikaelMayer: 🤖 Good call. Made `refToDef` a `private` field so external code can no longer access it directly. Added `SemanticModel.get?` (returns `Option ResolvedNode`) as the public accessor alongside the existing `get` (which returns `default` on miss). Updated the translator to use `get?`.

Commit: 09f9ae04c

MikaelMayer (approved):

ssomayyajula (approved):

PR #1135, opened 2026-05-07
ANFEncoder: iterate to fixpoint to eliminate nested duplicates

tautschnig:
The ANF encoder extracts duplicated subexpressions into fresh variables, but the existing single-pass implementation can leave large duplicated sub-subexpressions behind.

Root cause: `removeSubsumed` drops candidate duplicates that are subexpressions of other (larger) candidate duplicates, to avoid creating redundant variable declarations. But this means that if only the outer expression appears at the top level, the inner dupes are hidden inside the lifted var declaration and never extracted.

Example (from PyAnalyzeLaurel benchmark check_storage_costs):

  Original: assert Any_to_bool(Any_get(response, "Datapoints")) ...

  After partial evaluation, Any_to_bool inlines its 7-branch body,
  each branch referencing the argument. With Any_get also inlined
  as an `ite (is-DictStrAny response) (DictStrAny_get ...)
  (List_get ...)`, the Any_get expression ends up duplicated 62
  times inside a single assert.

  Old ANF output:
    var $__anf.0 : bool := <9KB body with 62 duplicates of Any_get>;
    assert $__anf.0

  New ANF output (after iteration):
    var $__anf.3 : Any := Any_get(response, "Datapoints");
    var $__anf.0 : bool := <body using $__anf.3>;
    assert $__anf.0

Effect: VC file size for VCs on one benchmark drops from 32KB to 11KB (~65% reduction), and another benchmark now completes verification at 36s where it previously hit a 60s timeout.


Strata/Transform/ANFEncoder.lean
```diff
@@ -192,9 +192,17 @@ private def findANFEncoderTargets (exprs : List Expression.Expr) :
...
+    always denote the same value within a procedure body.
+
+    Iterates to fixpoint because extracting one level of duplicates exposes
+    further duplicates inside the newly-created var declarations (a large
+    expression may hide subexpression dupes that `removeSubsumed` skipped
+    on the first pass). -/
+partial def anfEncodeBody (body : Statements) (startIdx : Nat) : Statements × Nat :=
   let targets := findANFEncoderTargets ((Statements.collectExprs body).flatMap collectSubexprs)
+  if targets.isEmpty then
+    (body, startIdx)
```
MikaelMayer: 🤖 Nit: consider adding a unit test that exercises multi-pass extraction (e.g., an expression where a large duplicate contains a smaller duplicate that only becomes extractable after the first pass). The existing tests all resolve in a single iteration, so the fixpoint loop is currently untested at the unit level.

MikaelMayer (commented):
🤖 Clean, well-motivated change. The fixpoint iteration is a natural solution to the nested-duplicate problem, and the early-exit on `targets.isEmpty` guarantees termination in practice (each pass extracts at least one duplicate, strictly reducing the pool of non-leaf subexpressions).

One suggestion below about test coverage for the new iteration behavior.

Strata/Transform/ANFEncoder.lean
```diff
@@ -192,9 +192,17 @@ private def findANFEncoderTargets (exprs : List Expression.Expr) :
     the next available dedup index.
     Assumes single-assignment (SSA-like) property of the post-PE Core IR:
     variables are assigned only once, so structurally equal expressions
-    always denote the same value within a procedure body. -/
-def anfEncodeBody (body : Statements) (startIdx : Nat) : Statements × Nat :=
+    always denote the same value within a procedure body.
+
+    Iterates to fixpoint because extracting one level of duplicates exposes
+    further duplicates inside the newly-created var declarations (a large
```
joscoh: I don't quite understand this. How does exposing further duplicates mean that it reaches a fixpoint? It will be necessary (to prove soundness) to give a termination proof, so it might as well be done here. Or if the proof effort is too large for now, there should at least be a clear explanation of why this converges.

Strata/Transform/ANFEncoder.lean
```diff
@@ -209,7 +217,10 @@ def anfEncodeBody (body : Statements) (startIdx : Nat) : Statements × Nat :=
   ) ([], ({} : Std.HashMap UInt64 (Expression.Expr × Expression.Expr)), startIdx)
   -- Single pass: replace all targets at once
   let body' := Statements.mapExprs (replaceExprs replacements) body
-  (revDecls.reverse ++ body', nextIdx)
+  let newBody := revDecls.reverse ++ body'
+  -- Iterate: the newly-inserted var declarations may themselves contain
+  -- duplicated subexpressions that `removeSubsumed` dropped in this pass.
+  anfEncodeBody newBody nextIdx
```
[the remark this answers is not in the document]
tautschnig: Fixed in 6789d548: the `replacements` map is now `Std.HashMap UInt64 (List (Expression.Expr × Expression.Expr))`. Inserts append a `(target, replacement)` pair to the bucket list, and `replaceExprs.check` walks the list with structural `==`, so colliding hashes never displace each other. Bottom-up O(n) hash computation is preserved.

Strata/Transform/ANFEncoder.lean
tautschnig: Added in 6789d548 as `nestedDupProg` in `StrataTest/Transform/ANFEncoderTests.lean` — `(x + 1) * 2` (outer) subsumes `x + 1` (inner) in pass 1; pass 2 extracts the inner once it surfaces in the new var-decl init and the third assert.

Strata/Transform/ANFEncoder.lean
tautschnig: Addressed in 6789d548: `anfEncodeBody` is now a real `def` (no longer `partial`) parameterised by a structurally-decreasing `fuel`, set to the total expression size of the initial body — a sound upper bound on the iteration count. The docstring states the termination argument explicitly: `S(body)` (distinct non-leaf, no-bvar subexpressions) is finite and monotonically non-increasing, since fresh `fvar`s are leaves filtered from all future `S`, and each new var-decl init is drawn from `S(body)` itself.

MikaelMayer (commented):
🤖🔍 LGTM — all comments addressed.

joscoh (dismissed):

Strata/Transform/ANFEncoder.lean
```diff
@@ -83,8 +83,14 @@ private def findDuplicates (exprs : List Expression.Expr) : List Expression.Expr
 
 /-- Replace all occurrences of any target with its corresponding replacement
     in an expression. Computes hashes bottom-up to avoid redundant traversals.
-    The map stores (target, replacement) pairs keyed by hash. -/
-def replaceExprs (replacements : Std.HashMap UInt64 (Expression.Expr × Expression.Expr))
+
+    The map values are lists of (target, replacement) pairs so that distinct
+    expressions sharing the same `LExpr.hashExpr` do not displace each other
+    on insertion (cf. PR #1135 review). On lookup we walk the list with
```
MikaelMayer: We should not have " (cf. PR #1135 review)". Comments should have authority. Only typical accepted GitHub links are links to unresolved GitHub issues.

MikaelMayer (commented):

Strata/Transform/ANFEncoder.lean
tautschnig: Addressed in [a88dace](https://github.com/strata-org/Strata/pull/1135/commits/a88dace613373d45e9ca7dd7deae79dc7047acc7).

joscoh (approved):

MikaelMayer (approved):

PR #1136, opened 2026-05-07
Document the Python front-end

tautschnig:
Adds two complementary documents:

* docs/PythonFrontend.md — a narrative, human-oriented reference describing how Python source is turned into a Core program. Covers the .py → Laurel → Core → SMT/GOTO pipeline, the Any-boxed value encoding (with the datatype definition), the PySpec mechanism for external libraries such as boto3, the supported Python subset and the constructs that are rejected or collapsed, and the soundness posture (tag checks are the contract, PySpec is trusted, aliasing is a known unsoundness). It includes a 'where to look in the code' table linking to the main Lean files involved at each stage.

* .kiro/steering/python-frontend-debugging.md — a compact playbook for agents debugging pyAnalyzeLaurel / pyAnalyzeLaurelToGoto failures. Scoped via fileMatch so it activates only when editing Strata/Languages/Python/**, Strata/Languages/Laurel/**, or the corresponding tests. Lists: a pipeline map correlating error prefixes with the stage that emits them; the seven invariants of the Any-boxed encoding; catalogued failure classes ('Resolution failed', 'could not infer type', 'calls to procedures in functions or contracts', 'holes should have been eliminated', 'N passed, M inconclusive', false alarms on clean benchmarks, GOTO-side issues) with where-to-look pointers; a pre-edit checklist; and the architectural limits coming from Strata's type system that should not be worked around locally.

### DIFF
--- a/docs/PythonFrontend.md
+++ b/docs/PythonFrontend.md
@@ -1,0 +1,377 @@
+# Strata's Python Front-End
+
+## Overview
+
+The Python front-end translates annotated Python programs into Strata's Core
+intermediate representation so that they can be analysed by Strata's deductive
+verifier (`pyAnalyzeLaurel`) or compiled through to CBMC GOTO programs
+(`pyAnalyzeLaurelToGoto`).
+
+The front-end is intentionally scoped. It does not attempt to model arbitrary
+Python. Its target is the subset of Python that AI coding agents typically
+produce when gluing together calls to external libraries: fully type-annotated
+procedural code that invokes library APIs described by PySpec, raises a
+handful of exception types, and avoids the more dynamic corners of the
+language. Within that subset the aim is *sound* analysis — if the tool
+reports "no bugs" then, modulo known specification gaps, there really are
+none.
+
+This document describes:
+
+1. The pipeline from a Python `.py` file to a Core program.
+2. The `Any`-boxed value encoding that underpins everything the front-end
+   produces.
+3. The PySpec mechanism that describes external library functions.
+4. What is supported and what is *not*.
+5. Where soundness hinges on translator-inserted tag checks rather than
+   Strata's own type system.
+
+Read this together with [`Architecture.md`](Architecture.md) for background
+on Strata's dialects and Core type system.
+
+## The pipeline at a glance
+
+```
+┌──────────────┐   ReadPython        ┌──────────────────┐
+│ Python .py   │ ───────────────────▶│ Python AST       │
+└──────────────┘   (cpython-derived) │ (DDM dialect)    │
+                                     └────────┬─────────┘
+                                              │ PythonToLaurel
+                                              ▼
+┌──────────────┐   pySpecToLaurel    ┌──────────────────┐
+│ PySpec .ion  │ ───────────────────▶│ Laurel program   │
+│ (libraries)  │                     │ (Any-boxed)      │
+└──────────────┘                     └────────┬─────────┘
+                                              │ Laurel→Core
+                                              │ (resolution,
+                                              │  heap param.,
+                                              │  hole elim., ...)
+                                              ▼
+                                     ┌──────────────────┐
+                                     │ Core program     │
+                                     └────────┬─────────┘
+                                              │
+                      ┌───────────────────────┴───────────────────┐
+                      ▼                                           ▼
+            ┌────────────────────┐                     ┌────────────────────┐
+            │ Core Verifier      │                     │ GOTO emission      │
+            │ (VCG + SMT)        │                     │ (CBMC backend)     │
+            └────────────────────┘                     └────────────────────┘
+```
+
+The two user-facing commands are:
+
+* `strata pyAnalyzeLaurel` — runs the full pipeline into SMT-based deductive
+  verification.
+* `strata pyAnalyzeLaurelToGoto` — stops after Core and emits CBMC GOTO JSON.
+
+Both consume the same `.python.st.ion` file (produced by the cpython-based
+`Tools/Python` reader).
+
+## Stage 1: Python to Laurel
+
+Code: [`Strata/Languages/Python/PythonToLaurel.lean`](../Strata/Languages/Python/PythonToLaurel.lean)
+(~3000 lines).
+
+This is the single largest piece of the front-end. It walks the Python AST and
+produces Laurel statements. Each Python construct is translated under the
+following conventions:
+
+### Every user value is `Any`
+
+All user-defined variables, procedure arguments, and return values are typed
+`Any` at the Laurel/Core level, regardless of their Python annotations. `Any`
+is an inductive datatype defined in
+[`PythonLaurelCorePrelude.lean`](../Strata/Languages/Python/PythonLaurelCorePrelude.lean):
+
+```
+datatype Any () {
+  from_None (),
+  from_bool (as_bool : bool),
+  from_int  (as_int : int),
+  from_float (as_float : real),
+  from_str (as_string : string),
+  from_datetime (as_datetime : int),
+  from_bytes (as_bytes : string),
+  from_DictStrAny (as_Dict : DictStrAny),
+  from_ListAny (as_ListAny : ListAny),
+  from_ClassInstance (classname : string, instance_attributes : DictStrAny),
+  from_Slice (start : int, stop : OptionInt),
+  exception (get_error : Error)
+}
+```
+
+Under this encoding:
+
+* A Python `int` literal `42` becomes `from_int(42)`.
+* `x + y` on Python `int`s becomes
+  `from_int(Any..as_int!(x) + Any..as_int!(y))`, bracketed by tag checks.
+* A class instance is `from_ClassInstance("MyClass", <attrs>)`.
+* An exception is `exception(<error>)` so that error propagation can be
+  threaded through assignments without branching.
+
+### Type checking = tag checking
+
+Strata Core uses Hindley–Milner inference (see
+[`LTy.lean`](../Strata/DL/Lambda/LTy.lean)). At the Core level the checker
+only sees `Any` everywhere, so the real "type checking" is done by the
+translator: before it emits `Any..as_int!(x)` it emits an assertion of
+`Any..isfrom_int(x)`, and after computing a Python expression whose annotated
+type is `int` it rewraps the result with `from_int`.
+
+This is sound *if* the translator inserts all the right tag checks. Strata's
+type system cannot catch a translator bug that drops one. Treat the translator
+as the trusted part of the chain.
+
+### Statements and control flow
+
+The translator supports the following statement forms:
+
+* Assignments, including augmented assignments (`x += 1`).
+* `if/elif/else`.
+* `while` and `for` (for-loops with `range(...)` or iterable expressions are
+  desugared to counter-driven while-loops).
+* `try/except`. The translator threads a sentinel `maybe_except : Error`
+  through the block: after every statement that might raise, an `if
+  isError(maybe_except) then exit exception_handlers_N` is inserted, and the
+  `except` handlers are emitted as labelled blocks that the exit jumps to.
+* `return`, `break`, `continue`, `pass`.
+* Class and function definitions (see below).
+
+### Procedure calls
+
+Every Python call goes through `translateCall`, which distinguishes:
+
+1. **User-defined functions** in the same file — translated as Laurel
+   procedure calls using the locally gathered signatures.
+2. **PySpec-modelled functions** (external libraries) — translated using the
+   spec's pre/postconditions and return shape.
+3. **Unmodelled calls** — translated to a `.Hole` placeholder, which the
+   eventual hole-elimination pass turns into a fresh uninterpreted function
+   call. Receiver objects and `Any`-typed arguments are havoc'd because the
+   call may mutate them.
+
+The translator inserts exception-check assertions around calls that may raise
+— for a call `f(...)` that is declared to possibly return `Any|exception`,
+the translator emits
+`assert !Any..isexception(<result-variable>)` *after* the assignment, so that
+downstream code does not silently continue with an exception in hand. When
+the expression `f(...)` appears in a position where an embedded assert would
+violate purity rules (e.g. inside an `if` condition, a `while` guard, a `for`
+iterable, or another assignment RHS), the call is first extracted into a
+temporary variable (see PR #1012 for the precise transformation).
+
+### Classes
+
+Classes are translated as Laurel composite types. A Python class becomes:
+
+* A composite declaration listing each `self.field` encountered during
+  translation (fields are inferred, not declared upfront, so classes without
+  `__init__` initialisers still produce a complete composite).
+* One procedure per method. Method names are mangled as
+  `ClassName@methodName` to avoid clashes across classes.
+* A heap parameterization pass later turns `self.field` reads/writes into
+  `readField`/`updateField` on a common `Heap` parameter.
+
+Inheritance is *not* modelled with subtyping. Classes involved in an
+inheritance hierarchy are flagged in `classesInHierarchy`; method calls on
+such classes are conservatively turned into holes (see PR #1019), because
+dynamic dispatch is beyond what the `Any` encoding can recover after the
+translation step.
+
+### Holes
+
+The translator emits `StmtExpr.Hole` whenever it cannot faithfully model a
+construct. Later passes (`InferHoleTypes`, `EliminateHoles`) annotate each
+hole with a type recovered from the surrounding context and replace it with
+a call to a fresh, uninterpreted `$hole_N` procedure. The verifier can then
+either havoc the result (bug-finding mode) or treat it as an arbitrary value
+consistent with its type (deductive mode).
+
+## Stage 2: PySpec — describing external libraries
+
+Code: [`Strata/Languages/Python/Specs/`](../Strata/Languages/Python/Specs),
+pipeline in
+[`PySpecPipeline.lean`](../Strata/Languages/Python/PySpecPipeline.lean).
+
+PySpec is the companion format that describes the external surface the user
+code talks to: third-party libraries and parts of the standard library. A
+PySpec file is itself a Strata program (in the `PythonSpecs` dialect) written
+as an `.ion`. Large specs are typically generated offline from an upstream
+API description (type stubs, interface definition files, service models) and
+then compiled to `.pyspec.st.ion` by Strata.
+
+A PySpec declaration is roughly:
+
+```
+composite KeyValueStore {
+  function get_item(self : KeyValueStore, **kwargs : dict(container : str, key : str, ...)) -> dict(value : bytes, ...) {
+    requires kwargs.container.length >= 3
+    requires regex(kwargs.key, "^[^/]+(/.+)?$")
+    ensures response.value is_bytes
+    ...
+  };
+  ...
+}
+```
+
+At translation time the front-end:
+
+1. Reads the compiled PySpec Ion files selected by the `--pyspec` and
+   `--dispatch` module-name flags (and any auxiliary modules they pull in),
+   resolving each module name against the directory passed to `--spec-dir`.
+2. Indexes overload tables so that a Python call like `store.get_item(...)`
+   can be dispatched to the right PySpec procedure based on the argument shape
+   (upstream API descriptions frequently have overloaded signatures).
+3. Produces a Laurel procedure per PySpec procedure. The pre/post-conditions
+   become Laurel `requires`/`ensures` clauses; the declared parameter
+   constraints (regex, length, range, enum literals) become assertions inside
+   the requires.
+
+Because a library's full spec can easily run to thousands of procedures,
+PySpec Ion files are compiled offline and cached on disk. In this repository,
+the compilation step is exposed as the `pySpecs` subcommand:
+
+```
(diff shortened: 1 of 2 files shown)


docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
+# Strata's Python Front-End
+
+## Overview
+
+The Python front-end translates annotated Python programs into Strata's Core
```
remi-delmas-3000, line 5: programs annotated with what kind of information ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+intermediate representation so that they can be analysed by Strata's deductive
+verifier (`pyAnalyzeLaurel`) or compiled through to CBMC GOTO programs
+(`pyAnalyzeLaurelToGoto`).
+
+The front-end is intentionally scoped. It does not attempt to model arbitrary
```
remi-delmas-3000, line 10: suggestion: the front-end is scoped down to a fragment of python

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+Python. Its target is the subset of Python that AI coding agents typically
+produce when gluing together calls to external libraries: fully type-annotated
+procedural code that invokes library APIs described by PySpec, raises a
+handful of exception types, and avoids the more dynamic corners of the
+language. Within that subset the aim is *sound* analysis — if the tool
+reports "no bugs" then, modulo known specification gaps, there really are
```
remi-delmas-3000, line 16: > modulo known specification gaps
this introduces doubt, are these gaps in the semantics of the tool accepts as input ? If we know the gaps then why not scope down the fragment even further to reject these constructs and give unconditionally sound results ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+none.
+
+This document describes:
+
+1. The pipeline from a Python `.py` file to a Core program.
+2. The `Any`-boxed value encoding that underpins everything the front-end
+   produces.
+3. The PySpec mechanism that describes external library functions.
```
remi-delmas-3000, line 24: So the pyspec .ion files are contracts for all external dependencies of the .py file on this diagram ?

remi-delmas-3000: why this asymetry between .py and .ion files ? why not use regular python signatures with contract annotations to specify external dependencies ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+  from_Slice (start : int, stop : OptionInt),
+  exception (get_error : Error)
+}
+```
+
+Under this encoding:
+
+* A Python `int` literal `42` becomes `from_int(42)`.
+* `x + y` on Python `int`s becomes
+  `from_int(Any..as_int!(x) + Any..as_int!(y))`, bracketed by tag checks.
```
remi-delmas-3000, line 108: is any of that boxing/unboxing statically resolved before or during the translation to SMT or GOTO or do we generate quantifier heavy SMT encodings and let the SMT solver resolve all boxing and unboxing ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+Strata Core uses Hindley–Milner inference (see
+[`LTy.lean`](../Strata/DL/Lambda/LTy.lean)). At the Core level the checker
+only sees `Any` everywhere, so the real "type checking" is done by the
+translator: before it emits `Any..as_int!(x)` it emits an assertion of
+`Any..isfrom_int(x)`, and after computing a Python expression whose annotated
+type is `int` it rewraps the result with `from_int`.
+
+This is sound *if* the translator inserts all the right tag checks. Strata's
+type system cannot catch a translator bug that drops one. Treat the translator
+as the trusted part of the chain.
```
remi-delmas-3000, line 124: would there be a way to prove using lean that the translator actually inserts these checks where they need to be ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+Every Python call goes through `translateCall`, which distinguishes:
+
+1. **User-defined functions** in the same file — translated as Laurel
+   procedure calls using the locally gathered signatures.
+2. **PySpec-modelled functions** (external libraries) — translated using the
+   spec's pre/postconditions and return shape.
+3. **Unmodelled calls** — translated to a `.Hole` placeholder, which the
+   eventual hole-elimination pass turns into a fresh uninterpreted function
+   call. Receiver objects and `Any`-typed arguments are havoc'd because the
```
remi-delmas-3000, line 151: So side effects on call arguments and receiver objects are captured via havocs but what about statefulness of such calls ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+   call may mutate them.
+
+The translator inserts exception-check assertions around calls that may raise
+— for a call `f(...)` that is declared to possibly return `Any|exception`,
+the translator emits
+`assert !Any..isexception(<result-variable>)` *after* the assignment, so that
```
remi-delmas-3000, line 157: The doc may need a section about exceptions are modeled in Laurel. I would have thought it's ok for code to raise exceptions.

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+downstream code does not silently continue with an exception in hand. When
+the expression `f(...)` appears in a position where an embedded assert would
+violate purity rules (e.g. inside an `if` condition, a `while` guard, a `for`
```
remi-delmas-3000, line 160: thats a lot to unpack in a single sentence: What is an "embedded assert" ? Who imposes purity rules (laurel or python) ? are asserts embedded in terms that end up embedded in `if`, `while`, `for` conditions ?

Seems like purity rules are Laurel requirements but it needs to be made more explicit. Do we have something like GOTO where all side effects need to be captured by explicit assignment/call statements involving pure expressions ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+on Strata's dialects and Core type system.
+
+## The pipeline at a glance
+
+```
+┌──────────────┐   ReadPython        ┌──────────────────┐
+│ Python .py   │ ───────────────────▶│ Python AST       │
+└──────────────┘   (cpython-derived) │ (DDM dialect)    │
+                                     └────────┬─────────┘
+                                              │ PythonToLaurel
```
remi-delmas-3000, line 39: Can we explain where python desugaring happens ? the python compiler does some desugaring, do we translate the initial AST or the desugared AST ?

remi-delmas-3000: Can we also say more about how the front-end decides to reject inputs it knows it cannot soundly support ? is it based on pattern matching basic language constructs in isolation, or is it based on pattern matching interactions of different language constructs we know we cannot soundly model ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+Classes are translated as Laurel composite types. A Python class becomes:
+
+* A composite declaration listing each `self.field` encountered during
+  translation (fields are inferred, not declared upfront, so classes without
+  `__init__` initialisers still produce a complete composite).
+* One procedure per method. Method names are mangled as
+  `ClassName@methodName` to avoid clashes across classes.
+* A heap parameterization pass later turns `self.field` reads/writes into
+  `readField`/`updateField` on a common `Heap` parameter.
```
remi-delmas-3000, line 174: what about setattr/getattr ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+Inheritance is *not* modelled with subtyping. Classes involved in an
+inheritance hierarchy are flagged in `classesInHierarchy`; method calls on
+such classes are conservatively turned into holes (see PR #1019), because
```
remi-delmas-3000, line 178: are we modeling the MRO in any way today ? https://docs.python.org/3/reference/datamodel.html#object.__mro_entries__

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+translation step.
+
+### Holes
+
+The translator emits `StmtExpr.Hole` whenever it cannot faithfully model a
+construct. Later passes (`InferHoleTypes`, `EliminateHoles`) annotate each
+hole with a type recovered from the surrounding context and replace it with
+a call to a fresh, uninterpreted `$hole_N` procedure. The verifier can then
+either havoc the result (bug-finding mode) or treat it as an arbitrary value
+consistent with its type (deductive mode).
```
remi-delmas-3000, line 189: Why two models of the same thing ? the deductive mode treatment of holes seems unsound

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+inheritance hierarchy are flagged in `classesInHierarchy`; method calls on
+such classes are conservatively turned into holes (see PR #1019), because
+dynamic dispatch is beyond what the `Any` encoding can recover after the
+translation step.
+
+### Holes
+
+The translator emits `StmtExpr.Hole` whenever it cannot faithfully model a
+construct. Later passes (`InferHoleTypes`, `EliminateHoles`) annotate each
+hole with a type recovered from the surrounding context and replace it with
```
remi-delmas-3000, line 186: Inferred using which type system ?  Laurel-level type inference or python-level type inference here ? can laurel infer anything but Any today ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+## Stage 2: PySpec — describing external libraries
+
+Code: [`Strata/Languages/Python/Specs/`](../Strata/Languages/Python/Specs),
+pipeline in
+[`PySpecPipeline.lean`](../Strata/Languages/Python/PySpecPipeline.lean).
+
+PySpec is the companion format that describes the external surface the user
+code talks to: third-party libraries and parts of the standard library. A
+PySpec file is itself a Strata program (in the `PythonSpecs` dialect) written
```
remi-delmas-3000, line 199: How does this PythonSpec dialect relate to Laurel ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+    ...
+  };
+  ...
+}
+```
+
+At translation time the front-end:
+
+1. Reads the compiled PySpec Ion files selected by the `--pyspec` and
+   `--dispatch` module-name flags (and any auxiliary modules they pull in),
```
remi-delmas-3000, line 221: why two flags ? how do they differ ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+2. Indexes overload tables so that a Python call like `store.get_item(...)`
+   can be dispatched to the right PySpec procedure based on the argument shape
+   (upstream API descriptions frequently have overloaded signatures).
+3. Produces a Laurel procedure per PySpec procedure. The pre/post-conditions
+   become Laurel `requires`/`ensures` clauses; the declared parameter
+   constraints (regex, length, range, enum literals) become assertions inside
+   the requires.
+
+Because a library's full spec can easily run to thousands of procedures,
+PySpec Ion files are compiled offline and cached on disk. In this repository,
```
remi-delmas-3000, line 232: Is that the PySpec to Laurel compilation ? Why not just translate what is needed on the fly ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+which walks the source directory for `.py` stub files and emits one
+`<module>.pyspec.st.ion` per module into the output directory. `--module`
+restricts the run to selected modules and `--skip` excludes specific
+top-level definitions. The resulting directory is then what gets passed to
+`pyAnalyzeLaurel` as `--spec-dir`.
+
+### The `--dispatch` flag
+
+When the spec tree contains a top-level `<package>.pyspec.st.ion` (or
```
remi-delmas-3000, line 247: is that "directory tree" or "AST" of the spec  ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+From Laurel onwards the pipeline is common to all front-ends:
+
+1. **Resolution** annotates identifier uses with the declarations they refer
+   to ([`Resolution.lean`](../Strata/Languages/Laurel/Resolution.lean)).
+2. **Heap parameterization** lowers `self.field` access to
+   `readField`/`updateField`
+   ([`HeapParameterization.lean`](../Strata/Languages/Laurel/HeapParameterization.lean)).
+3. **Hole elimination** assigns types to `Hole` expressions and replaces them
+   with fresh procedures
```
remi-delmas-3000, line 263: in earlier sections we say uninterpreted functions, not procedures

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+   [`EliminateHoles.lean`](../Strata/Languages/Laurel/EliminateHoles.lean)).
+4. **Laurel-to-Core translation**
+   ([`LaurelToCoreTranslator.lean`](../Strata/Languages/Laurel/LaurelToCoreTranslator.lean))
+   yields a Core program that is then type-checked and verified (or compiled
+   to GOTO).
+
+For deductive verification, the verifier then performs standard Core VC
+generation and discharges the VCs through the SMT solver. Bug-finding mode is
+similar but uses non-deterministic havoc instead of quantified universals on
+the uninstantiated hole procedures.
```
remi-delmas-3000, line 274: why these two distinct treatments of holes ?

remi-delmas-3000: the way this is explained makes it seem like we are using a sound model of holes (havoc) for bug finding and an unsound modeling of holes for verification (fully correct value of expected types but no side effects)

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+## Supported Python subset
+
+What the front-end handles well:
```
remi-delmas-3000, line 278: well as in "sound" or well as in "60% of the time it works every time" ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+   ([`LaurelToCoreTranslator.lean`](../Strata/Languages/Laurel/LaurelToCoreTranslator.lean))
+   yields a Core program that is then type-checked and verified (or compiled
+   to GOTO).
+
+For deductive verification, the verifier then performs standard Core VC
+generation and discharges the VCs through the SMT solver. Bug-finding mode is
+similar but uses non-deterministic havoc instead of quantified universals on
+the uninstantiated hole procedures.
+
+## Supported Python subset
```
remi-delmas-3000, line 276: This could be moved earlier in the doc as it really helps understand why the translation works

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+* `list`, `dict[str, Any]` with value semantics (see caveat below).
+* Classes with `__init__`, regular methods, simple field access.
+* `try/except` for a known exception type, and raising exceptions by
+  instantiating them.
+* External library calls (via PySpec), including auto-generated input
+  validation and result narrowing.
+* The `__name__ == "__main__"` idiom (kept as a plain `if`).
+* Arithmetic, comparisons, and boolean combinators on boxed `Any` values.
+
+## What is not supported (and why)
```
remi-delmas-3000, line 290: May I suggest the following categorization:
- what we support now
- what we will definitely never support
- what we plan to support in the near future (we know we can model soundly)
- what we don't know yet wether we can support or not

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+5. Where soundness hinges on translator-inserted tag checks rather than
+   Strata's own type system.
+
+Read this together with [`Architecture.md`](Architecture.md) for background
+on Strata's dialects and Core type system.
+
+## The pipeline at a glance
+
+```
+┌──────────────┐   ReadPython        ┌──────────────────┐
```
remi-delmas-3000, line 35: Where would an external third party python type-checker like pyright fit in this picture ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+* Arbitrary decorators.
+* Multiple inheritance / MRO-dependent behaviour.
+
+These are not representable in the `Any` encoding without a heap-allocated
+code model which Strata does not have.
+
+### Collapsed to untyped
+
+* Union types (`X | Y`), `Optional[X]`. The translator encodes these by
+  erasing to `Any`; individual PySpec stubs occasionally introduce hand-rolled
```
remi-delmas-3000, line 307: so we don't retain any information  from these type annotations ? or do they someone get translated to type tag assertions/assumptions ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+  pairs such as `StrOrNone`, but there is no general union machinery.
+* `TypedDict`, `Protocol`, `@overload`, `Literal`, `Callable`,
+  `TypeGuard`/`TypeIs`, `Self`, `ParamSpec`. The Core type system has no
+  subtyping, variance, row types, or intersection/constraint solving, so
+  these annotations are stripped and replaced with tag checks where
+  possible.
```
remi-delmas-3000, line 313: We should try to tell exactly in which case this is or isn't possible

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+* Flow-sensitive narrowing (`if isinstance(x, int): x + 1` does not give `x`
+  the type `int` in the branch). The Core checker is flow-insensitive;
+  narrowing must be encoded manually via tag checks.
```
remi-delmas-3000, line 316: it seems like only the front end could generate these tag checks ? is it already doing it ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+### Unusable libraries
+
+Frameworks that rely on metaclasses, descriptors, dynamic class generation,
+runtime reflection, or non-trivial C extensions are effectively outside the
+supported subset: Django, SQLAlchemy, pydantic, NumPy, pandas, PyTorch, attrs,
+FastAPI/Flask at the framework layer, pytest fixtures. Using any of these
+typically leads to unmodelled calls being havoc'd into holes, which in turn
+drives SMT verification toward "inconclusive".
```
remi-delmas-3000, line 325: Does that mean our verification would still be sound (but very imprecise) on programs that use these libraries ?

remi-delmas-3000: Reading the whole document it is still not clear wether the front-end rejects unsupported inputs or just ingests inputs but models them super conservatively. Could we try to distinguish inputs that we can soundly overapproximate using holes and still ingest ((e.g. call to an external dep we don't have a precise contract for), from inputs that are outside of anything we can remotely model and simply reject (e.g. metaclasses that generate classes dynamically)?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+
+### Aliasing and identity
+
+Mutable containers (`list`, `dict`) are encoded as *value-semantics*
+inductive datatypes (`ListAny`, `DictStrAny`), not as heap references. This
+means the model does not capture aliasing: after `c = a; a.append(4)`, the
```
remi-delmas-3000, line 331: is this a hard limitation or do we have a path to soundness on this point ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+the Strata model has `c` unchanged. Similarly, Python's `is` and user-defined
+`__eq__`/`__hash__` are not modelled. This is a known latent unsoundness that
+can affect verification results on code that relies on aliasing.
+
+## Soundness posture
+
+* **Tag checks are the contract.** Every place a Core expression consumes a
+  specific constructor (e.g. `as_int`) must be preceded by a Strata assertion
+  that the corresponding `isfrom_*` holds. The translator is responsible for
+  inserting these; a missing one is a soundness bug.
```
remi-delmas-3000, line 342: what's our path to soundness here ? lean proofs about how the translator inserts tag checking assertions ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+* **PySpec correctness is assumed.** When a PySpec procedure states a
+  postcondition, the verifier trusts it. Specs generated from upstream API
+  descriptions inherit the same trust.
+* **Holes are sound by construction.** Each hole becomes a fresh
+  uninterpreted procedure; the verifier treats its outputs as arbitrary
```
remi-delmas-3000, line 347: uninterpreted as in x = y => f(x) = f(y) ? is this really sound ? what if the external dep is stateful ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+  values consistent with their type. This is sound but often imprecise — a
+  typical "inconclusive" result traces back to a hole.
+* **Aliasing is unsound** as noted above.
```
remi-delmas-3000, line 350: what's the path to soundness there ?

docs/PythonFrontend.md
```diff
@@ -0,0 +1,377 @@
...
+* **Everything else follows from Strata's Core type system**, whose
```
remi-delmas-3000, line 351: can we elaborate on what the "everything else" means here from a python semantics perspective ? does it mean that once we've reached strata core nothing else can go wrong ? i.e. any unexpected results is necessarily a bug in the upper layers ?

PR #1137, opened 2026-05-07
ci: extract install-cvc5, install-z3, restore-lake-cache into composite actions

tautschnig:
The same ~15-line blocks for installing cvc5, installing z3, and restoring the lake cache were duplicated across ci.yml's three jobs (build_and_test_lean, check_pending_python, build_python), cbmc.yml, and python-fuzz.yml. Robin asked for this to be consolidated as part of the #984 review (flagged non-blocking there, handled here).

Introduce three composite actions under .github/actions/:

- install-cvc5: downloads the static cvc5 release zip, makes the binary available either on $GITHUB_PATH ('path', default) or in /usr/local/bin ('system'). Version defaults to 1.2.1.

- install-z3: same shape for z3, version defaults to 4.15.2. This fixes a latent bug: cbmc.yml and build_and_test_lean used different aarch64 archive names ('arm64-glibc-2.34' vs. 'arm64-win'); the consolidated action uses the correct glibc-2.34 one for both.

- restore-lake-cache: wraps actions/cache/restore@v5 with the established key pattern lake-<os>-<arch>-<toolchain>-<manifest>-<all .st>-<sha> plus the three fallback keys dropping each trailing component. Exposes a 'fail-on-cache-miss' input for jobs that depend on a cache saved for the same SHA by build_and_test_lean.

Rewire ci.yml and cbmc.yml to use the new actions:

- ci.yml: build_and_test_lean: install-cvc5, install-z3, restore-lake-cache. check_pending_python: install-cvc5 (system), restore-lake-cache (fail-on-cache-miss: true).
    build_python:        install-cvc5 (system), restore-lake-cache
      (fail-on-cache-miss: true).
      (The pip-based z3 install here is deliberately left as-is —
      it is a different install flow.)
- cbmc.yml:
    install-cvc5, install-z3, restore-lake-cache (fail-on-cache-miss: true).

Net change: -108 lines of duplicated shell, +17 lines of action usage across the two workflow files, with the reusable definitions ready to be adopted by python-fuzz.yml and any future workflows.

actionlint reports no new warnings (the pre-existing save-always warning on cbmc.yml line 22 is unrelated).


.github/actions/restore-lake-cache/action.yml
```diff
@@ -0,0 +1,56 @@
...
+  using: composite
+  steps:
+    - name: Restore lake cache
+      id: restore
+      uses: actions/cache/restore@v5
+      with:
+        path: ${{ inputs.path }}
+        key: ${{ inputs.key-prefix }}-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
+        restore-keys: |
+          ${{ inputs.key-prefix }}-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}
```
shigoel, line 71: The `restore-keys` are always included, but the downstream jobs (`cbmc`, `check_pending_python`, `build_python`) previously had **no** `restore-keys` — intentionally, so that `fail-on-cache-miss: true` would reject anything except the exact-SHA cache (per #952). With these fallbacks present, a stale prefix match will satisfy the restore and the step won't fail, undermining the safety net.

Consider adding an input (e.g. `use-restore-keys`, default `true`) that omits the `restore-keys` block when `false`.

shigoel (commented):
Nice cleanup — the z3 aarch64 bug fix is a good catch. One behavioral concern flagged inline.

.github/actions/restore-lake-cache/action.yml
tautschnig: Fixed in [2bd4be7](https://github.com/strata-org/Strata/pull/1137/commits/2bd4be77ed17dce121796951f8259e0afef3c6ba) - thank you for catching this!!

shigoel (dismissed):
Clean refactoring that eliminates ~108 lines of duplicated CI shell and introduces well-parameterized composite actions. Notably fixes a latent bug where ci.yml used `arm64-win` as the z3 archive name on aarch64 instead of the correct `arm64-glibc-2.34`. The `use-restore-keys` design on restore-lake-cache is a nice touch for preserving the #952 safety net. LGTM.

.github/workflows/ci.yml
aqjune-aws: The save step in build_and_test_lean is still inline:
  key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}

  The restore action generates the same key from inputs.key-prefix (default "lake") + the same hash components. This works today, but there's a drift risk: if someone edits the save step's key pattern but forgets to update the composite action (or vice versa), the downstream jobs will silently get cache misses.

  Suggestion: Consider either (a) also extracting the save into a companion action, or (b) adding a comment on the save step pointing to the action as the canonical key definition.

atomb (commented):
## Code Review: PR #1137 — Extract CI composite actions

### Summary

Extracts duplicated shell blocks for installing cvc5, z3, and restoring the lake cache into three reusable composite actions under `.github/actions/`. Rewires `ci.yml` and `cbmc.yml` to use them. Net change: -108 lines of duplicated shell, +17 lines of action usage. Also fixes a latent bug where `ci.yml` used the wrong archive name (`arm64-win` instead of `arm64-glibc-2.34`) for z3 on aarch64.

---

### Features

1. **Fixes a real bug.** The existing `ci.yml` downloads `z3-4.15.2-arm64-glibc-2.34.zip` but sets `ARCHIVE_NAME="z3-4.15.2-arm64-win"` — the subsequent `unzip` would fail on aarch64 runners. The consolidated action uses the correct name.

2. **Well-designed input interface.** `install-to` (`path` vs `system`) cleanly handles the two installation modes. `fail-on-cache-miss` and `use-restore-keys` on `restore-lake-cache` correctly model the two cache usage patterns (upstream builder vs downstream consumer).

3. **Good documentation.** Each action has a clear description explaining its purpose, provenance, and future adoption plans. The `restore-lake-cache` action documents the rationale for `use-restore-keys` with a link to #952.

4. **Correct cache key pattern.** Reproduces the exact same key structure and fallback hierarchy as the original inline blocks.

5. **Appropriate scoping.** The `build_doc` job's cache (different prefix, multiple paths, multiple toolchain files) is correctly left untouched.

6. **`set -eu` in shell scripts.** Stricter than the original inline scripts — failures will be caught immediately.

---

### Potential Issues

See inline comments for the significant items. Minor items:

- **No cleanup of downloaded archives** — zip files and extracted directories accumulate in `$GITHUB_WORKSPACE` (~50-100MB). Same as before, not a regression.
- **`wget -q` consistency** — The new actions consistently use `wget -q`, which is an improvement but means less verbose error output if a download fails in `build_and_test_lean`.
- **Version pinning discoverability** — Default versions are now in `.github/actions/*/action.yml` rather than the workflow files. A comment in the workflows pointing to where versions are configured would help.
- **`build_python` z3 version skew** — Uses z3 4.13.4 (pip) while other jobs use 4.15.2 (GitHub release). Not introduced by this PR, just noting.
- **`python-fuzz.yml` not yet converted** — Consider whether to do it here or track as follow-up.

.github/actions/restore-lake-cache/action.yml
```diff
@@ -0,0 +1,80 @@
...
+    required: false
+    default: "true"
+
+outputs:
+  cache-hit:
+    description: Whether a cache entry was restored (see actions/cache/restore@v5).
+    value: ${{ steps.restore-with-fallback.outputs.cache-hit || steps.restore-exact.outputs.cache-hit }}
+
+runs:
+  using: composite
```
atomb, line 61: **`cache-hit` output may be empty when both conditional steps are skipped**

The output uses `||` to combine two conditional steps:
```yaml
value: ${{ steps.restore-with-fallback.outputs.cache-hit || steps.restore-exact.outputs.cache-hit }}
```

If `inputs.use-restore-keys` is neither exactly `'true'` nor `'false'` (e.g., a caller passes boolean `true` instead of string `"true"`), the `if` conditions use string comparison (`!= 'false'` and `== 'false'`), so a non-string value would hit the first branch. This is fine for current callers but could surprise future users.

More importantly: if a caller passes `use-restore-keys: "false"` but the `restore-exact` step fails (e.g., due to a permissions issue), the output would be empty rather than signaling failure. Consider adding a validation step or documenting the expected input values more explicitly in the description.

.github/workflows/ci.yml
```diff
@@ -74,14 +43,7 @@ jobs:
         # Only use the caches on PRs because there is a risk of stale results:
```
atomb, line 43: **Cache key coupling with `Save lake cache` step**

The `restore-lake-cache` action now owns the cache key pattern, but the "Save lake cache" step in `build_and_test_lean` (line ~93 of ci.yml, not touched by this PR) still uses the raw `actions/cache/save@v5` with a hardcoded key:

```yaml
key: lake-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('lean-toolchain') }}-${{ hashFiles('lake-manifest.json') }}-${{ hashFiles('**/*.st') }}-${{ github.sha }}
```

If the `restore-lake-cache` action's key pattern ever drifts from this save step's key (e.g., someone changes the `key-prefix` default or adds a new hash component), caches would silently miss.

Consider either:
- Adding a comment in `ci.yml` next to the save step noting the coupling ("key must match `.github/actions/restore-lake-cache/action.yml`")
- Or as a follow-up, extracting a `save-lake-cache` companion action that uses the same key construction logic

atomb (dismissed):
My agent's review above highlights some potential future improvements (mostly to things this PR didn't introduce), but I don't think any are blocking.

.github/workflows/ci.yml
tautschnig: Addressed in [0656bf8](https://github.com/strata-org/Strata/pull/1137/commits/0656bf8c3f1f13c014f64b123c59d7b4d4cf4a6a).

.github/actions/restore-lake-cache/action.yml
tautschnig: Addressed in [0656bf8](https://github.com/strata-org/Strata/pull/1137/commits/0656bf8c3f1f13c014f64b123c59d7b4d4cf4a6a).

.github/workflows/ci.yml
tautschnig: Addressed in [0656bf8](https://github.com/strata-org/Strata/pull/1137/commits/0656bf8c3f1f13c014f64b123c59d7b4d4cf4a6a).

aqjune-aws (approved):

shigoel (approved):

PR #1141, opened 2026-05-07
The small-step semantics of Imperative must have scoped var init, remove unlabeled exit from Core

### LINKED ISSUE #372: Semantics should be scoped
The semantics say that the following program executes correctly
```
if B {
  init x := 2;
} else {
  init x := 3;
}
assert x > 1;
```

This is because the small-step statement semantics inline the statements of a block when they are reached.
However, everyone I asked declared that this program should not be accepted because x is declared locally.
We need to make sure the semantics capture that.
This will be helpful to support shadowing because otherwise shadowing would not have been possible with this program:

```
init x := 2;
if B {
  init x := 3;
}
assert x == 2;
```

With shadowing semantics, this program verifies, but if the init of the blocks leaks, then this program no longer verifies.
I checked with many of us and we all agreed the program above, if given a meaning, it should be that x is shadowed, not overwritten.

aqjune-aws:
*Issue #, if available:*

Closes https://github.com/strata-org/Strata/issues/372                                                                                                                                                                                            

*Description of changes:*

Formalizing scoped variable initialization:

  - Adds scoped environment semantics to blocks: step_block_done now projects the store through the parent store via projectStore, discarding block-local variables on exit. This applies to both StepStmt (deterministic) and StepKleene           
  (nondeterministic).
  - Adds PostWF helper definition in Specification.lean for postcondition stability under projectStore.                                                     
  - Adds .block constructor to KleeneStmt and corresponding step_block/step_block_body/step_block_done to StepKleene, mirroring the deterministic block semantics.
  - Changes StmtToKleeneStmt (.block _ bss _) to produce .block b (wrapping in a Kleene block), so the Kleene translation preserves scoping.

Removing unlabeled `exit` command (`exit;`, not `exit lbl;`):

  - The unlabeled `exit` command doesn't have clear meaning when it is inside `while`. In `while cond { ... exit; ... }`, is `exit` equivalent to `continue` or `break` in C/Python? No translation to Core was introducing the unlabeled `exit`, and the small-step semantics wasn't clear about the meaning of `exit` inside a loop.
  - The type checker of Core (Strata/Languages/Core/StatementType.lean) fails when an unlabeled `exit` appears inside the body of a while loop / if statement (unless it is wrapped by another nested block), but the DDM syntax of core allows it, so from user's perspective this gap is kind of surprising.
  - This patch removes the unlabeled `exit` case because it can always be simulated by a labeled block + labeled exit.

On top of this, this also renames `touchedVars` to `modifiedOrDefinedVars` for clarity, and instead makes `touchedVars` all vars that are read + modified + defined.

### How to review?

- Strata/DL/Imperative/StmtSemantics.lean has the most important update: the `.block` constructor now has the input store which will be used to 'project' the output store to the variables that have been defined before. Also, any block-exiting small steps like `step_block_done` will do the projection, to define variables which were defined inside the inner scope.
- Correspondingly, the Kleene language was updated to add a notion of block and scoping to its syntax and semantics, otherwise DetToKleene doesn't prove. (Strata/DL/Imperative/KleeneStmt.lean, Strata/DL/Imperative/KleeneStmtSemantics.lean)
- StrataTest/DL/Imperative/StepStmtTest.lean has a few new tests that show scoping works well.
- DetToKleeneCorrect.lean and ProcBodyVerifyCorrect.lean shows that the top-level statements are not touched after this update.
- Strata/DL/Imperative/HasVars.lean has the `touchedVars` update and Strata/Languages/Core/StatementSemantics.lean has some additional well-formedness about evaluator extension that are useful.

### DIFF
--- a/Strata/DL/Imperative/Stmt.lean
+++ b/Strata/DL/Imperative/Stmt.lean
@@ -43,10 +43,9 @@ inductive Stmt (P : PureExpr) (Cmd : Type) : Type where
   | loop     (guard : ExprOrNondet P) (measure : Option P.Expr)
              (invariants : List (String × P.Expr))
              (body : List (Stmt P Cmd)) (md : MetaData P)
-  /-- An exit statement that transfers control out of the nearest enclosing
-  block with the given label. If no label is provided, exits the nearest
-  enclosing block. -/
-  | exit     (label : Option String) (md : MetaData P)
+  /-- An exit statement that transfers control out of the enclosing block
+  with the given label. -/
+  | exit     (label : String) (md : MetaData P)
   /-- A function declaration within a statement block. -/
   | funcDecl (decl : PureFunc P) (md : MetaData P)
   /-- A type declaration within a statement block. -/
@@ -80,7 +79,7 @@ def Stmt.inductionOn {P : PureExpr} {Cmd : Type}
       (body : List (Stmt P Cmd)) (md : MetaData P),
       (∀ s, s ∈ body → motive s) →
       motive (Stmt.loop guard measure invariant body md))
-    (exit_case : ∀ (label : Option String) (md : MetaData P),
+    (exit_case : ∀ (label : String) (md : MetaData P),
       motive (Stmt.exit label md))
     (funcDecl_case : ∀ (decl : PureFunc P) (md : MetaData P),
       motive (Stmt.funcDecl decl md))
@@ -373,9 +372,7 @@ def formatStmt (P : PureExpr) (s : Stmt P C)
       let beforeBody := nestD f!"{line}{guard}{line}({measure}){line}{invFmt}"
       let children := group f!"{beforeBody}{line}{body}"
       f!"{md}while{children}"
-  | .exit label md => match label with
-    | some l => f!"{md}exit {l}"
-    | none => f!"{md}exit"
+  | .exit label md => f!"{md}exit {label}"
   | .funcDecl _ md => f!"{md}funcDecl <function>"
   | .typeDecl tc md => f!"{md}type {tc.name} (arity {tc.numargs})"
 
@@ -407,29 +404,24 @@ instance [ToFormat P.Ident] [ToFormat P.Expr] [ToFormat P.Ty] [ToFormat C]
 by an enclosing `block` — either within `s` itself or with a label in
 `labels` (representing blocks that enclose `s` externally).
 
-When `s.exitsCoveredByBlocks []`, execution of `s` can never produce `.exiting`.
+When `s.exitsCoveredByBlocks []`, execution of `s` can never produce `.exiting`. -/
 
-The labels have type `Option String` (not `String`) so that `exit` without
-destination block label can be considered as covered even when it is surrounded
-by unlabeled blocks (`[None]`). -/
-
-@[expose] def Stmt.exitsCoveredByBlocks : List (Option String) → Stmt P CmdT → Prop
+@[expose] def Stmt.exitsCoveredByBlocks : List String → Stmt P CmdT → Prop
   | _, .cmd _ => True
-  | labels, .block l ss _ => Block.exitsCoveredByBlocks (.some l :: labels) ss
+  | labels, .block l ss _ => Block.exitsCoveredByBlocks (l :: labels) ss
   | labels, .ite _ tss ess _ => Block.exitsCoveredByBlocks labels tss ∧ Block.exitsCoveredByBlocks labels ess
   | labels, .loop _ _ _ body _ => Block.exitsCoveredByBlocks labels body
-  | labels, .exit none _ => labels.length > 0
-  | labels, .exit (some l) _ => .some l ∈ labels
+  | labels, .exit l _ => l ∈ labels
   | _, .funcDecl _ _ => True
   | _, .typeDecl _ _ => True
 where
-  Block.exitsCoveredByBlocks : List (Option String) → List (Stmt P CmdT) → Prop
+  Block.exitsCoveredByBlocks : List String → List (Stmt P CmdT) → Prop
     | _, [] => True
     | labels, s :: ss => Stmt.exitsCoveredByBlocks labels s ∧ Block.exitsCoveredByBlocks labels ss
 
 theorem block_exitsCoveredByBlocks_append
     {P : PureExpr} {CmdT : Type}
-    (labels : List (Option String)) (ss₁ ss₂ : List (Stmt P CmdT))
+    (labels : List String) (ss₁ ss₂ : List (Stmt P CmdT))
     (h₁ : Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks labels ss₁)
     (h₂ : Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks labels ss₂) :
     Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks labels (ss₁ ++ ss₂) := by
@@ -441,7 +433,7 @@ theorem block_exitsCoveredByBlocks_append
     can only help. -/
 theorem exitsCoveredByBlocks_weaken
     {P : PureExpr} {CmdT : Type}
-    (labels₁ labels₂ : List (Option String))
+    (labels₁ labels₂ : List String)
     (hsub : ∀ l, l ∈ labels₁ → l ∈ labels₂) :
     (∀ (s : Stmt P CmdT),
       s.exitsCoveredByBlocks labels₁ → s.exitsCoveredByBlocks labels₂) ∧
@@ -466,8 +458,8 @@ theorem exitsCoveredByBlocks_weaken
   | cmd _ => intros; trivial
   | block l ss _ ih =>
     intro labels₁ labels₂ hsub h
-    show Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks (.some l :: labels₂) ss
-    exact ih (.some l :: labels₁) (.some l :: labels₂)
+    show Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks (l :: labels₂) ss
+    exact ih (l :: labels₁) (l :: labels₂)
       (fun x hx => by cases hx with
         | head => exact .head _
         | tail _ hm => exact .tail _ (hsub x hm))
@@ -478,14 +470,9 @@ theorem exitsCoveredByBlocks_weaken
   | loop _ _ _ body _ ih =>
     intro labels₁ labels₂ hsub h
     exact ih labels₁ labels₂ hsub h
-  | exit label _ =>
+  | exit l _ =>
     intro labels₁ labels₂ hsub h
-    cases label with
-    | none =>
-      show labels₂.length > 0
-      exact List.length_pos_iff_exists_mem.mpr
-        (let ⟨x, hx⟩ := List.length_pos_iff_exists_mem.mp h; ⟨x, hsub x hx⟩)
-    | some l => exact hsub (.some l) h
+    exact hsub l h
   | funcDecl _ _ => intros; trivial
   | typeDecl _ _ => intros; trivial
   | nil => intros; trivial
@@ -497,7 +484,7 @@ theorem exitsCoveredByBlocks_weaken
     for any labels (since `.cmd` has no exit statements). -/
 theorem all_cmd_exitsCoveredByBlocks
     {P : PureExpr} {CmdT : Type}
-    (labels : List (Option String)) (ss : List (Stmt P CmdT))
+    (labels : List String) (ss : List (Stmt P CmdT))
     (h : ∀ s ∈ ss, ∃ c, s = Stmt.cmd c) :
     Stmt.exitsCoveredByBlocks.Block.exitsCoveredByBlocks labels ss := by
   induction ss with
--- a/Strata/DL/Imperative/Stmt.lean
+++ b/Strata/DL/Imperative/Stmt.lean
@@ -107,7 +107,7 @@ def Stmt.inductionOn {P : PureExpr} {Cmd : Type}
 /-! ### sizeOf -/
 
 mutual
-@[simp, expose]
+@[simp]
 def Stmt.sizeOf (s : Imperative.Stmt P C) : Nat :=
   match s with
   | .cmd c => 1 + SizeOf.sizeOf c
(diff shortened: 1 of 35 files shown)


MikaelMayer (commented):
🤖 Overall this is a well-structured PR that correctly implements block-scoped variable semantics via `projectStore`. The Kleene language extension mirrors the deterministic semantics cleanly, and the new tests provide good coverage of the scoping behavior.

A few items to consider:

- The `PostWF` definition quantifies over *all* `σ_parent`, which is stronger than needed (in practice only the actual parent store matters). This is conservative and safe, but worth documenting that it means postconditions must not reference block-local variables.
- There is significant code duplication in the new `core_step_preserves_wf*` / `core_wf*_preserved` theorems — all six follow the exact same proof pattern as `core_step_preserves_wfBool`.
- `Stmt.labels` / `Block.labels` appear to be dead code (not referenced anywhere outside their definition).

Strata/Languages/Core/StatementSemanticsProps.lean
```diff
@@ -2326,6 +2306,123 @@ theorem core_wfBool_preserved
   | step _ _ _ hstep _ ih =>
     exact ih (core_step_preserves_wfBool π φ h_wf_ext _ _ hwf₀ hstep)
 
+theorem core_step_preserves_wfVar
```
MikaelMayer, line 2309: 🤖 `core_step_preserves_wfVar`, `core_wfVar_preserved`, `core_step_preserves_wfCong`, `core_wfCong_preserved`, `core_step_preserves_wfExprCongr`, `core_wfExprCongr_preserved` are all structurally identical to `core_step_preserves_wfBool` / `core_wfBool_preserved` — only the property type and the `WFEvalExtension` field differ. Consider extracting a generic helper parameterized by the property and the field accessor to avoid this 6× duplication.

Strata/DL/Imperative/Stmt.lean
```diff
@@ -304,31 +308,66 @@ mutual
...
   -- order matters for Havoc, so needs to override the default
-  touchedVars := Block.touchedVars
+  modifiedOrDefinedVars := Block.modifiedOrDefinedVars
+
+---------------------------------------------------------------------
+
+/-! ### Labels -/
+
+mutual
+@[expose] def Stmt.labels (s : Stmt P C) : List String :=
```
MikaelMayer: 🤖 `Stmt.labels` and `Block.labels` are defined here but not referenced anywhere else in the codebase. Is this dead code left over from development, or is it intended for future use? If the latter, a brief docstring explaining the intended purpose would help.

Strata/Transform/Specification.lean
```diff
@@ -287,18 +287,32 @@ theorem seq_cons {s : Stmt P CmdT} {ss : List (Stmt P CmdT)}
           exact h₂ ρ₁ ρ' hmid (hwfb_preserved ρ₁ hterm_s) hf₁ (.inr ⟨lbl, hexit_ss⟩)
 
 omit [HasVal P] in
-/-- Lift a `TripleBlock` to a `Triple` by wrapping in a block. -/
+/-- A postcondition is well-formed if it is stable under `projectStore`. -/
+def PostWF (Post : Env P → Prop) : Prop :=
```
MikaelMayer, line 291: 🤖 `PostWF` quantifies over *all* `σ_parent`, meaning the postcondition must hold even when projected through a store that defines no variables (making everything `none`). In practice, `TripleBlock.toTriple` only applies projection with the actual parent store at block entry. Would it be more precise (and easier for callers to discharge) to parameterize `PostWF` by the specific `σ_parent` that will be used? Or at minimum, document that this means the postcondition must only reference variables that are guaranteed to be defined in the parent scope.

MikaelMayer (commented):
🔍 The new commit is a merge of main — no changes to the files I previously commented on. My three suggestions (deduplicating the wf-preservation proofs, clarifying whether `Stmt.labels`/`Block.labels` are dead code, and documenting `PostWF`'s universal quantification over `σ_parent`) remain unaddressed. These were non-blocking suggestions so LGTM overall, but it would be good to get a response on the `labels` question in particular.

Strata/Transform/Specification.lean
aqjune-aws: This had to be forall `σ_parent` because in general we don't know which variables are defined in this scope unless we know the information of the exact context. A Hoare triple is only aware of the code block inside the pre-postconditions, but can't know the code outside of it:
```
c1
{pre}
c2      <-- A Hoare triple must only be aware of this c2; we can't give information about c1 or c3 (which will have .init) to post.
{post}
c3
```

Strata/Languages/Core/StatementSemanticsProps.lean
aqjune-aws: Yes, indeed they have very similar but slightly different proofs.... I have tried this but resolving the duplication was non-trivial.

Strata/DL/Imperative/Stmt.lean
aqjune-aws: Removed, thanks :)

MikaelMayer (commented):
🔍 Verified the new commits since last review:

- **`.labels` dead code**: ✅ Fixed — removed in commit 917ea90.
- **Deduplicating wf-preservation proofs**: Not addressed (non-blocking, still a nice-to-have).
- **`PostWF` documentation**: Not addressed (non-blocking).

The one actionable item was properly fixed. The "Disallow unlabeled exit" commit is a clean refactor that simplifies the semantics. LGTM.

MikaelMayer (commented):
🔍 LGTM — the latest commit (`3a875b08`) is a minor `omit` annotation cleanup. My previous non-blocking suggestions (deduplicating wf-preservation proofs, documenting `PostWF`) remain open but are not blockers.

MikaelMayer (commented):
🔍 LGTM — the latest commit is a merge of main with no changes to PR files. Previous non-blocking suggestions (deduplicating wf-preservation proofs, documenting `PostWF`) remain open but are not blockers.

MikaelMayer:
I think this is an excellent move to have scoped semantics. It will likely make proofs simpler

MikaelMayer (dismissed):

MikaelMayer (commented):
🔍 LGTM — the latest commit removes `@[simp]` from variable-tracking definitions, keeping only `@[expose]`. Clean change. Previous non-blocking suggestions (deduplicating wf-preservation proofs, documenting `PostWF`) remain open but are not blockers.

Strata/DL/Imperative/Stmt.lean
```diff
@@ -108,7 +107,7 @@ def Stmt.inductionOn {P : PureExpr} {Cmd : Type}
 /-! ### sizeOf -/
 
 mutual
-@[simp]
+@[simp, expose]
```
joscoh: Curious why these (and all the subsequent definitions) need to be exposed when they weren't before.

Strata/DL/Imperative/Stmt.lean
```diff
@@ -304,31 +307,44 @@ mutual
...
   match ss with
   | [] => []
-  | s :: srest => Stmt.touchedVars s ++ Block.touchedVars srest
+  | s :: srest => Stmt.modifiedOrDefinedVars s ++ Block.modifiedOrDefinedVars srest
+end
+
+mutual
+/-- Get all variables touched (modified, defined, or read) by the statement `s`. -/
+@[expose]
+def Stmt.touchedVars [HasVarsImp P C] [HasVarsPure P P.Expr] [HasVarsPure P C]
```
joscoh, line 320: Not a blocker for this PR, but we should really be using sets and set-union for these things. The same variable might appear many times within a statement/block/etc.

Strata/Languages/Core/StatementSemanticsProps.lean
```diff
@@ -2282,33 +2282,13 @@ theorem core_step_preserves_wfBool
     (hstep : CoreStep π φ c₁ c₂) :
     WellFormedSemanticEvalBool c₂.getEnv.eval := by
   induction hstep with
-  | step_cmd hcmd =>
-    cases hcmd with
-    | cmd_sem _ => simp [Config.getEnv]; exact hwf
-    | @call_sem _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ =>
-        simp only [Config.getEnv]; exact hwf
+  | step_cmd hcmd => cases hcmd with
+    | cmd_sem _ | @call_sem _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ =>
```
joscoh: Nit: why do you need the `@` if all arguments are wildcards anyway?

Strata/Languages/Core/StatementSemanticsProps.lean
```diff
@@ -2326,6 +2306,123 @@ theorem core_wfBool_preserved
...
+theorem core_wfExprCongr_preserved
+    (h_wf_ext : WFEvalExtension φ)
+    (c₁ c₂ : CoreConfig)
+    (hwf₀ : @Imperative.WellFormedSemanticEvalExprCongr Expression _ c₁.getEnv.eval)
+    (hstar : CoreStepStar π φ c₁ c₂) :
+    @Imperative.WellFormedSemanticEvalExprCongr Expression _ c₂.getEnv.eval := by
+  suffices ∀ c₁ c₂, @Imperative.WellFormedSemanticEvalExprCongr Expression _ c₁.getEnv.eval →
+      Imperative.StepStmtStar Expression (EvalCommand π φ) (EvalPureFunc φ) c₁ c₂ →
+      @Imperative.WellFormedSemanticEvalExprCongr Expression _ c₂.getEnv.eval from
+    this c₁ c₂ hwf₀ (CoreStepStar_to_StepStmtStar hstar)
```
joscoh: We should not use auto-generated Lean names (e.g. `this`)

Strata/Transform/DetToKleeneCorrect.lean
```diff
@@ -474,38 +495,58 @@ private def loop_sim_kleene
...
+          | .refl _ => rfl
+          | .step _ _ _ h _ => exact nomatch h
+      subst hρ_x_eq
+      have hterm_body_eq : StepStmtStar P (EvalCmd P) extendEval (.stmts body ρ₀) (.terminal ρ_inner) :=
+        hρ₀_eq ▸ reflTransT_to_prop h_inner_term
+      have h_sim_body : StepKleeneStar P (EvalCmd P) (.stmt b ρ₀) (.terminal ρ_inner) :=
+        sim_body ρ₀ ρ_inner hwfb hwfv hterm_body_eq
+      have hwfv_inner : WellFormedSemanticEvalVal ρ_inner.eval := by
+        have := block_noFuncDecl_preserves_eval P (EvalCmd P) extendEval body ρ₀ ρ_inner hnofd_body hterm_body_eq
+        rw [this]; exact hwfv
```
joscoh, line 521: Same here about `this`

StrataTest/DL/Imperative/StepStmtTest.lean
```diff
@@ -104,42 +104,85 @@ def noCmd : EvalCmdParam MiniPureExpr CmdT := fun _ _ _ _ _ => False
...
-/-! ## Test: `loop { exit }` exactly exits the loop, not the outer block.
+/-! ## Test: `block "L" { loop { exit "L" } }` exits the loop via labeled exit.
 
-A minimal program `loop { exit }` is shown to step to `.terminal`.  This
-verifies that an unlabeled `exit` inside the body terminates just the
-loop (and not the enclosing block).
+The `exit "L"` propagates out of body's per-iteration block and the loop's
+recursive step (mismatch propagates), reaching the labeled outer block.
+
+With the new per-iteration loop semantics (each body wrapped in its own block),
```
joscoh: Nit: comments should be stateless

joscoh:
Also, could you update the docs please? Right now it says "An exit statement transfers control out of the nearest enclosing block with a matching label, or, if no label is provided, the nearest enclosing block.".

Strata/DL/Imperative/Stmt.lean
aqjune-aws: It is copied from the loop elimination proof, but at least in this PR `@[expose]` wasn't necessary so I removed them.

Strata/Languages/Core/StatementSemanticsProps.lean
aqjune-aws: Fixed them

aqjune-aws:
Fixed, thanks. Setifying the result of touchedVars and relevant helper functions make sense to me (their item order in the list shouldn't matter).

MikaelMayer (commented):
🔍 LGTM — the "address comments" commit properly addresses feedback: `@[expose]` removed from variable-tracking and sizeOf defs, anonymous `suffices` hypotheses named (`h_gen`), `@call_sem` simplified to named-argument style, outdated test comment removed, and docs updated to reflect unlabeled exit removal. Previous non-blocking suggestions (deduplicating wf-preservation proofs, documenting `PostWF`) remain open but are not blockers.

joscoh (approved):

Strata/DL/Imperative/Stmt.lean
joscoh: Makes sense. I know it is often unavoidable, but I think we should try to avoid exposing things when possible to get at least some of the benefits of the module system.

MikaelMayer (approved):

Strata/DL/Imperative/Stmt.lean
```diff
@@ -304,31 +303,40 @@ mutual
...
-@[simp]
-def Block.touchedVars [HasVarsImp P C] (ss : Block P C) : List P.Ident :=
+def Block.modifiedOrDefinedVars [HasVarsImp P C] (ss : Block P C) : List P.Ident :=
   match ss with
   | [] => []
-  | s :: srest => Stmt.touchedVars s ++ Block.touchedVars srest
+  | s :: srest => Stmt.modifiedOrDefinedVars s ++ Block.modifiedOrDefinedVars srest
+end
+
+mutual
```
PROgram52bc, line 318: Mutual block unnecessary?

PROgram52bc (approved):
I think a [similar Option construct](https://github.com/strata-org/Strata/blob/jlee/scoped_env/Strata/DL/Imperative/ToCProverGOTO.lean#L60) for the GOTO exits could be simplified as well. This is non-blocking, but perhaps they can be addressed in future PRs.

PR #1143, opened 2026-05-08
Qualify nested-class references when lowering pyspec class fields

tautschnig:
Some PySpec producers emit references to a class's nested class using the 'typeClassNoArgs' DDM grammar operator, which carries only a bare identifier with no module path.  On round-trip through 'DDM.SpecType.fromDDM', the reconstructed 'PythonIdent' has an empty 'pythonModule', so 'specTypeToLaurelType' renders the reference with its bare name rather than the qualified path under which the nested class is emitted.

When multiple sibling classes each declare a nested class of the same bare name, the resulting references collide, are filtered out by 'filterPrelude' as dangling (the nested class itself lives under a qualified name), and the Laurel-to-Core translator bails out with 'Resolution failed: <name> is not defined'.

Fix at the PySpec-to-Laurel lowering site:

* Add 'SpecType.mapIdentNames' for rewriting the PythonIdent on each 'SpecIdent' within a SpecType.
* In 'classDefToLaurel', before lowering each field, rewrite any bare ident whose name matches one of the current class's 'subclasses' to carry the parent's prefixed name as its 'pythonModule'.  After rewriting, 'PythonIdent.toLaurelName' produces, for an 'Outer' class with a nested 'Inner', the name 'Outer_Inner', which matches the name 'classDefToLaurel' assigns to the nested class.
* When recursing into nested classes, push the parent's prefixed name as the module prefix so the nested class's own emitted name aligns with how its parent references it.

Add regression tests in 'StrataTest/Languages/Python/ToLaurelTest.lean' covering: a single outer class with a nested-class field reference, two sibling outer classes each declaring a nested class of the same bare name (no collision), and a bare reference to an unrelated top-level class (left unqualified).

### DIFF
--- a/Strata/Languages/Python/Specs/Decls.lean
+++ b/Strata/Languages/Python/Specs/Decls.lean
@@ -342,6 +342,12 @@ def union (loc : SourceRange) (x y : SpecType) : SpecType :=
 def ident (loc : SourceRange) (i : PythonIdent) (args : Array SpecType := #[]) : SpecType :=
   { empty loc with idents := #[{ name := i, args }] }
 
+/-- Apply `f` to the `PythonIdent` of each `SpecIdent` in this `SpecType`.
+    Literal atoms and typed-dicts are left untouched. Useful for rewriting
+    bare identifiers that lost their module context during DDM round-trip. -/
+def mapIdentNames (f : PythonIdent → PythonIdent) (ty : SpecType) : SpecType :=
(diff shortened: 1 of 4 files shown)


joehendrix:
My first inclination is to say there's a bug that `moduleName` is empty.  Do you have an example?

We could just add a `moduleName \ne ""` constraint to `PythonIdent`, and make sure we are catching all empty module names at `PythonIdent` construction time.

PR #1144, opened 2026-05-08
Document the design of Laurel

tautschnig:
Extend the rendered Laurel manual and cross-link it from the repository entry points, so that Laurel's design rationale is discoverable and sits alongside the auto-extracted type reference.

* docs/verso/LaurelDoc.lean — four new top-level sections:
  - Design Principles (shared features only; imperative code need not hide behind a functional spec; one AST for statements and expressions; one AST node serves many back-ends; constructs are not defined by their Core encoding).
  - Conventions and Invariants (holes as the uniform 'unknown' marker; reserved `$`-prefixed names; uniqueId-based resolution; determinism of procedure calls).
  - Error Correction via Holes.
  - Relationship with the Python Front-End, and Known Limits and Roadmap. The Translation Pipeline section is expanded from four named passes to all eleven Laurel-to-Laurel passes plus CoreGroupingAndOrdering and the final Laurel-to-Core translation.

* docs/Architecture.md — the Laurel subsection is promoted from a two-sentence pointer to a summary of Laurel's distinguishing design choices, with links to the manual and to PythonFrontend.md.

* README.md — adds a pointer to the Laurel manual and the Python front-end guide next to the existing Architecture and Getting Started links.

* .kiro/steering/laurel-development.md — an agent-facing playbook, scoped via fileMatch to Strata/Languages/Laurel/** and its tests. Covers the scope rule, explicit do/don't lists, the pipeline order with its invariants, reserved identifier prefixes, error-reporting rules, and a common-bugs table.

The manual still builds (`lake build laurel`).


.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+* **Do not resolve identifiers by textual name** in any pass after
+  resolution. Use `uniqueId`.
+* **Do not encode Python-specific type information in Laurel types.**
+  Python's `Any` lives in `Strata/Languages/Python/`; Laurel sees it as
+  a single composite type with constructors, and that is the intended
+  abstraction.
+* **Do not lift procedure calls into assert/contract bodies.** Pure
+  contexts must remain pure. If a front-end wants to check a property
+  about a procedure call's result, it must first bind the result to a
+  local.
```
olivier-aws: I would challenge that one. I think it is find to assume that users in many programing languages will want to write something like `assert(foo()==3)` where `foo` could modify the heap. And I think we should accept that at the source level. Then, given that it is a construct that can occur in Python/Java/TS, we should accept it in Laurel per Laurel's design principles.

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+## Pipeline order (`LaurelCompilationPipeline.lean`)
+
+```
+FilterNonCompositeModifies → EliminateValueReturns →
+HeapParameterization (needsResolves) → TypeHierarchyTransform (needsResolves) →
+ModifiesClausesTransform (needsResolves) → InferHoleTypes → EliminateHoles →
+DesugarShortCircuit → LiftExpressionAssignments →
+EliminateReturns (needsResolves) → ConstrainedTypeElim (needsResolves) →
+CoreGroupingAndOrdering → LaurelToCoreTranslator
+```
```
olivier-aws, line 101: Maybe add a `Do` that the order in this doc must be updated if one adds a new transformation pass

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+  `LaurelFormat`, and `ConcreteToAbstractTreeTranslator` +
+  `AbstractToConcreteTreeTranslator`. Missing one produces the classic
+  "holes should have been eliminated before translation" error or
+  equivalent.
+* **Mark passes that invalidate resolution with `needsResolves := true`**
+  in `LaurelCompilationPipeline.lean`. Otherwise subsequent passes will
+  trip over dangling `uniqueId`s.
+* **Keep the translation to Core "dumb".** If a Laurel → Core case needs
+  non-trivial work, the work probably belongs in an earlier
+  Laurel → Laurel pass.
```
olivier-aws: Love that. I don't know if the order of the bullets matter to the coding agent, but this should come first IMO

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+  minimum: `MapStmtExpr`, `Resolution`, `HeapParameterization`,
+  `InferHoleTypes` / `EliminateHoles` (if the constructor can contain a
+  hole), `LiftImperativeExpressions`, `LaurelToCoreTranslator`,
+  `LaurelFormat`, and `ConcreteToAbstractTreeTranslator` +
+  `AbstractToConcreteTreeTranslator`. Missing one produces the classic
+  "holes should have been eliminated before translation" error or
+  equivalent.
+* **Mark passes that invalidate resolution with `needsResolves := true`**
+  in `LaurelCompilationPipeline.lean`. Otherwise subsequent passes will
+  trip over dangling `uniqueId`s.
```
olivier-aws: How do we know that a pass invalidates resolution?

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+## Do
+
+* **Use `.Hole` to recover from errors in translators.** Emit a
+  diagnostic and a hole in place of the problematic sub-expression; the
+  rest of the program still compiles and the user sees a useful error.
+* **When you add a `StmtExpr` constructor, update every pass.** At
+  minimum: `MapStmtExpr`, `Resolution`, `HeapParameterization`,
+  `InferHoleTypes` / `EliminateHoles` (if the constructor can contain a
+  hole), `LiftImperativeExpressions`, `LaurelToCoreTranslator`,
+  `LaurelFormat`, and `ConcreteToAbstractTreeTranslator` +
```
olivier-aws, line 67: Can we have a build time check that enforces that?

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+* **Do not lift procedure calls into assert/contract bodies.** Pure
+  contexts must remain pure. If a front-end wants to check a property
+  about a procedure call's result, it must first bind the result to a
+  local.
+
+## Do
+
+* **Use `.Hole` to recover from errors in translators.** Emit a
+  diagnostic and a hole in place of the problematic sub-expression; the
+  rest of the program still compiles and the user sees a useful error.
```
olivier-aws: Does that mean that we never reject an input program but always emit a `.Hole` with a warning?

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Reserved identifier prefixes
+
+Emit these from passes; do **not** accept them as user identifiers:
+
+* `$heap`, `$heap_in` — heap parameters.
+* `$hole_N` — uninterpreted functions generated by `EliminateHoles`.
+* `$__ty_unused_N` — fresh type variables used as "unknown type"
+  placeholders.
+* `$` prefix in general — reserved for translator-generated names.
```
olivier-aws, line 126: In Java, one can start an identifier with `$`: https://www.w3schools.com/java/java_variables_identifiers.asp. This may cause name conflict with `$heap`.

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+## Error reporting
+
+* Collect diagnostics; do not abort on the first error if you can
+  meaningfully continue with a hole.
+* Resolution errors abort the pipeline before Core to avoid duplicate
+  diagnostics. If your pass produces resolution errors, add them to the
+  Laurel diagnostics list and let the pipeline do the early return.
+* The Laurel pipeline is meant to produce user-actionable diagnostics.
+  "Not yet implemented" is an acceptable diagnostic for an unreachable
+  case; a panic is not.
```
olivier-aws, line 142: Can we add a principle that the error diagnostic must come with a source localization, to be really actionable?

olivier-aws (commented):
First pass at review.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
 like conditionals and variable declarations. Each StmtExpr has a user facing type, which for
 statement-like constructs could be void.
 
+# Design Principles
+
+These principles recur in design discussions on Laurel pull requests and issues. Proposed
+features or refactorings that violate them tend to be pushed back.
+
+## Shared between source languages, or don't add it
```
keyboardDrummer, line 92: I feel this paragraph does not add much on top of what is said in the introduction earlier in this file.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+The defining criterion for a Laurel feature is that it is "useful for Java and something
+else". Features that make sense only for one source language (for example, nominal
+subtyping specifically for Java generics, or Python's `__getattr__`) belong in the
+source-language translator, not in Laurel. Front-ends are expected to encode
+language-specific constructs using Laurel's primitives.
+
+Concrete instances of the principle:
+
+- Inheritance is in Laurel because Java, Python and TypeScript all use it.
+- Closures and thunks are not first-class in Laurel; they are modelled through the dynamic
```
keyboardDrummer: This contradicts what is said in the introduction, that higher-order procedures and procedure types are part of the language.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+
+The defining criterion for a Laurel feature is that it is "useful for Java and something
+else". Features that make sense only for one source language (for example, nominal
+subtyping specifically for Java generics, or Python's `__getattr__`) belong in the
+source-language translator, not in Laurel. Front-ends are expected to encode
+language-specific constructs using Laurel's primitives.
+
+Concrete instances of the principle:
+
+- Inheritance is in Laurel because Java, Python and TypeScript all use it.
```
keyboardDrummer, line 101: We'll have to see how well this works out for Python and TypeScript so maybe let's not mention this yet : )

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+- Closures and thunks are not first-class in Laurel; they are modelled through the dynamic
+  dispatch mechanism — a caller can encode a closure as a composite with a named method and
+  use `InstanceCall`.
+- Manual memory management is out of scope in the short term.
```
keyboardDrummer: This does not seem like a strong example of the principle, since it occurs in none of the mentioned languages.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+
+## Imperative code need not hide behind a functional specification
```
keyboardDrummer, line 103: This is already covered earlier in the document

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+Sometimes the imperative code is as readable as, or more readable than, a functional
+specification would be.
+
+Consequences of this decision:
+
+- A Laurel procedure body may be transparent to its callers. There is no requirement that a
+  caller only see an `ensures` clause; if the body has no postcondition the caller can see
+  (and reason about) the implementation. This is the origin of the transparent / opaque /
+  `Body.External` distinction.
+- Statements are allowed in expression positions in principle. The compilation pipeline has
```
keyboardDrummer, line 113: This seems more related to the next paragraph.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+  to lift them out eventually (see Lifting Expression Assignments below), but the AST
+  admits programs in which an assignment appears as part of an expression. The fact that
+  Core cannot represent such programs is not Laurel's problem.
+
+## One AST for statements and expressions
```
keyboardDrummer: This seems already covered by the existing "Implementation choices" part of the doc

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+
+The same choice drives why blocks can occur inside expressions: the machinery that handles
+a block of statements is exactly the machinery that handles a block used as an
+expression-with-side-effects.
+
+## One AST node, many back-ends
+
+Because Laurel targets multiple analyses — deductive verification, and in the future model
+checking, property-based testing, and data-flow analysis — semantic features that are
+specific to one discharge strategy must not leak into the AST. Opacity, for example, is a
```
keyboardDrummer: > semantic features that are
specific to one discharge strategy must not leak into the AST

I think I disagree. Opaqueness and assume statements are relevant for symbolic but not concrete execution, but they are first class language concepts, so they "leak" into the AST.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,72 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+concept connected to deductive verification: it is about how to generate VCs that don't
+include too much irrelevant information. Analyses other than deductive verification should
+not need to take opacity into account.
+
+## Constructs are not defined by their Core encoding
```
keyboardDrummer, line 127: Is there are reason to think the contrary? It seems overly verbose to mention this.

docs/verso/LaurelDoc.lean
```diff
@@ -146,11 +212,94 @@ A Laurel program consists of procedures, global variables, type definitions, and
 
 {docstring Strata.Laurel.Program}
 
+# Conventions and Invariants
+
+## Holes are the uniform "unknown" marker
```
keyboardDrummer, line 197: I will move any usage information related to holes to the documentation of the Hole constructor.

docs/verso/LaurelDoc.lean
```diff
@@ -146,11 +212,94 @@ A Laurel program consists of procedures, global variables, type definitions, and
...
+
+Two pipeline passes close the loop: `InferHoleTypes` labels every hole with the type its
+context demands, and `EliminateHoles` replaces deterministic holes with calls to freshly
+generated uninterpreted functions, and non-deterministic holes with havocs.
+
+The invariant is that any hole that survives past `EliminateHoles` is a pipeline bug. If a
+new translator case introduces holes inside a constructor the elimination pass does not
+traverse, that is where the fix has to go.
+
+## Reserved identifier prefixes
```
keyboardDrummer, line 212: This is a bug that's documented in this issue: https://github.com/strata-org/Strata/issues/1176

docs/verso/LaurelDoc.lean
```diff
@@ -146,11 +212,94 @@ A Laurel program consists of procedures, global variables, type definitions, and
...
+
+## Identifier resolution by unique ID
+
+After the Resolution pass, every identifier reference carries a `uniqueId : Option Nat`
+that matches some declaration. Passes must not resolve identifiers by looking at textual
+names; post-resolution, the `uniqueId` is authoritative. If a pass needs to know the owning
+type of a field, it should use the field's `uniqueId` to look up its declaration in the
+model rather than walking back through the composite type hierarchy.
+
+## Determinism of procedure calls
```
keyboardDrummer: This should be part of the documentation of the deterministic property of a procedure, which right now is commented out because it's not implemented yet.

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
+---
+inclusion: fileMatch
```
keyboardDrummer, line 2: To me its unclear whether such steering docs are useful. There's no mechanism that is keeping this document in sync with the code its describing, so wouldn't it become counter-productive over time?

keyboardDrummer (commented):
I went through this PR and extracted some things to another PR: https://github.com/strata-org/Strata/pull/1222 other things I have commented on.

docs/verso/LaurelDoc.lean
```diff
@@ -146,11 +212,94 @@ A Laurel program consists of procedures, global variables, type definitions, and
...
+the program can still compile and produce useful diagnostics), and by specification
+pipelines that cannot materialise a precondition.
+
+Two pipeline passes close the loop: `InferHoleTypes` labels every hole with the type its
+context demands, and `EliminateHoles` replaces deterministic holes with calls to freshly
+generated uninterpreted functions, and non-deterministic holes with havocs.
+
+The invariant is that any hole that survives past `EliminateHoles` is a pipeline bug. If a
+new translator case introduces holes inside a constructor the elimination pass does not
+traverse, that is where the fix has to go.
```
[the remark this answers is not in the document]
tautschnig: Fixed in 7278a820a: only deterministic holes become UF calls; non-deterministic holes are preserved and lowered to havocs later by `LiftExpressionAssignments`.

docs/verso/LaurelDoc.lean
```diff
@@ -146,11 +212,94 @@ A Laurel program consists of procedures, global variables, type definitions, and
...
+   returned.
+3. `HeapParameterization` (invalidates resolution) — add `$heap` and `$heap_in` parameters
+   to procedures that read or mutate the heap, and rewrite field reads/writes.
+4. `TypeHierarchyTransform` (invalidates resolution) — add a `TypeTag` field to composites
+   and emit `extends` facts for the inheritance hierarchy.
+5. `ModifiesClausesTransform` (invalidates resolution) — translate modifies clauses into
+   ensures clauses.
+6. `InferHoleTypes` — assign a type to every `Hole` from its context.
+7. `EliminateHoles` — replace deterministic holes with calls to freshly generated
+   uninterpreted functions; non-deterministic holes become havocs.
```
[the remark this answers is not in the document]
tautschnig: Fixed in 3410a570f: step 7 now says non-deterministic holes are lowered to havocs later by `LiftExpressionAssignments`, not in `EliminateHoles`.

docs/Architecture.md
```diff
@@ -58,8 +58,16 @@ The C_Simp dialect ([`Strata.Languages.C_Simp`](../Strata//Languages/C_Simp/)) i
...
+- **A unified `StmtExpr` AST.** Statements and expressions share a single type, eliminating duplication for shared constructs such as `IfThenElse` and `Block`.
+- **Imperative code may be transparent.** Unlike some IVLs, Laurel does not require imperative code to hide behind a functional specification; a Laurel procedure's body can be visible to its callers.
+- **Translator-neutral semantics.** Features that are specific to one back-end (e.g. opacity, which is a deductive-verification concept) are carried as annotations but not baked into the core AST, so that model checkers and data-flow analyses do not need to know about them.
+
+For full documentation — design principles, conventions, the compilation pipeline, and all `{docstring}`-extracted type references — see **[The Laurel Language manual](./verso/LaurelDoc.lean)** (rendered HTML under `docs/api/.lake/build/doc/Strata/Languages/Laurel/Laurel.html` after running `docs/verso/generate.sh`). For how the Python front-end uses Laurel, see [`docs/PythonFrontend.md`](PythonFrontend.md).
```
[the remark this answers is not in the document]
tautschnig: Fixed in 3f8129936: now points to the rendered HTML under `docs/verso/_out/laurel/html-single/`.

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
...
+EliminateReturns (needsResolves) → ConstrainedTypeElim (needsResolves) →
+CoreGroupingAndOrdering → LaurelToCoreTranslator
+```
+
+Invariants across passes:
+
+1. After `Resolution`, every identifier reference has a `uniqueId` that
+   matches some declaration.
+2. After `EliminateHoles`, the program contains no deterministic holes.
+   Non-deterministic holes are translated to havocs.
```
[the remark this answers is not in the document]
tautschnig: Fixed in fec0a5f91: invariant #2 now says non-deterministic holes are preserved and lowered later.

.kiro/steering/laurel-development.md
```diff
@@ -0,0 +1,139 @@
+---
+inclusion: fileMatch
+fileMatchPattern: ['Strata/Languages/Laurel/**/*.lean', 'StrataTest/Languages/Laurel/**/*.lean']
+---
+
+# Working on Laurel
+
+Pointers for agents editing Laurel code. For background, read
+[the rendered Laurel manual](../../docs/verso/LaurelDoc.lean)
```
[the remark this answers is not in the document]
tautschnig: Fixed in 3f8129936: relabeled to “Laurel manual source” and added the rendered-HTML path.

docs/verso/LaurelDoc.lean
```diff
@@ -146,11 +212,94 @@ A Laurel program consists of procedures, global variables, type definitions, and
...
+8. `DesugarShortCircuit` — desugar `&&` and `||` into conditional expressions.
+9. `LiftExpressionAssignments` — hoist statement-like `StmtExpr`s out of expression
+   positions.
+10. `EliminateReturns` (invalidates resolution) — remove `Return` from any remaining
+    expression contexts.
+11. `ConstrainedTypeElim` (invalidates resolution) — peel constrained types down to their
```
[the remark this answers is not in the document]
tautschnig: Fixed in 3410a570f: `TypeAliasElim` is now called out as running after `resolve`, before the numbered passes.

.kiro/steering/laurel-development.md
tautschnig: Addressed in bf9fdbc05: the source-level `assert(foo()==3)` is accepted; the rule is only that the front-end binds the call to a local before the pure context.

.kiro/steering/laurel-development.md
tautschnig: Added in e58975622.

.kiro/steering/laurel-development.md
tautschnig: Done in d5cb20698: “keep the translation to Core dumb” is now the first Do bullet.

.kiro/steering/laurel-development.md
tautschnig: Clarified in ccdc53ec3: a pass invalidates resolution when it introduces new declarations/references or rewrites references.

.kiro/steering/laurel-development.md
tautschnig: Implemented in ddf2caf88: a build-time check in `MapStmtExpr.lean` fails `lake build` if the central traversal skips a constructor's `StmtExpr` children (d20ca05f8 documents its scope).

.kiro/steering/laurel-development.md
tautschnig: Clarified in ab28f254f: `.Hole` is for recoverable errors; unrecoverable ones (e.g. resolution failures) still abort.

.kiro/steering/laurel-development.md
tautschnig: Noted in dc5f4d1d3: front-ends for languages that allow a leading `$` must escape user `$`-identifiers before emitting Laurel.

.kiro/steering/laurel-development.md
tautschnig: Added in 6ae7d84ae: every user-facing diagnostic must carry a source range via `DiagnosticModel.withRange`.

docs/verso/LaurelDoc.lean
tautschnig: Trimmed and cross-referenced the intro in bb052e1a8.

docs/verso/LaurelDoc.lean
tautschnig: Removed the closures bullet in 5cc16ec24.

docs/verso/LaurelDoc.lean
tautschnig: Removed in 5cc16ec24.

docs/verso/LaurelDoc.lean
tautschnig: Removed the manual-memory bullet in 5cc16ec24.

docs/verso/LaurelDoc.lean
tautschnig: Reduced to a cross-reference in bb052e1a8.

docs/verso/LaurelDoc.lean
tautschnig: Removed the “One AST for statements and expressions” subsection in bb052e1a8 (it duplicated the Implementation Choices paragraph).

docs/verso/LaurelDoc.lean
tautschnig: Removed the duplicate subsection in bb052e1a8.

docs/verso/LaurelDoc.lean
tautschnig: Reworded in d9acc454b: verification constructs (opacity, `assume`) are first-class in the AST; the rule is now that analyses which don't need them can ignore them.

docs/verso/LaurelDoc.lean
tautschnig: Trimmed to a single sentence in bb052e1a8.

docs/verso/LaurelDoc.lean
tautschnig: Done in 7278a820a: the LaurelDoc prose is now a pointer; hole-usage detail stays on the `StmtExpr`/`Hole` docstring.

docs/verso/LaurelDoc.lean
tautschnig: Reworded in 7278a820a: the survival invariant is now scoped to deterministic holes and references #1176.

docs/verso/LaurelDoc.lean
tautschnig: Removed the determinism subsection in 18d6bf416, pending the (currently commented-out) `deterministic` property.

.kiro/steering/laurel-development.md
tautschnig: Fair concern. The most drift-prone parts are now either corrected (hole behaviour 7278a820a, invariant #2 fec0a5f91) or machine-checked (the traversal-coverage claim, ddf2caf88); what remains is stable design guidance. Happy to fold the do/don'ts into module docstrings instead if you'd prefer.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,52 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+
+The introduction's "Verification design choices" section notes that Laurel does not require
+imperative code to be encapsulated behind a functional specification. Two consequences for
+the design:
+
+- A Laurel procedure body may be transparent to its callers. There is no requirement that a
+  caller only see an `ensures` clause; if the body has no postcondition the caller can see
+  (and reason about) the implementation. This is the origin of the transparent / opaque /
+  `Body.External` distinction.
+- Statements are allowed in expression positions in principle. The compilation pipeline has
```
keyboardDrummer, line 113: I think it would be better not to mention implementation choices, like the lifting pass or Core, here.

What does this item, "Statements are allowed in expression positions in principle..." add on top of the "## Implementation Choices" section? (which I think should get a different name, since the merging of statements and expression is quite the functional choice). 

Lastly, I think this item doesn't fit under the description of "## Imperative code need not hide behind a functional specification".

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,52 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+Beyond the shared features listed in the introduction, the operative rule is that a Laurel
+feature must be "useful for Java and something else". Constructs that make sense only for one
+source language (for example, nominal subtyping for Java generics, or Python's `__getattr__`)
+belong in that front-end's translator, encoded using Laurel's primitives.
+
+Concrete instances of the principle:
+
+- Inheritance is in Laurel because Java, Python and TypeScript all use it.
+
+## Imperative code need not hide behind a functional specification
```
keyboardDrummer, line 103: This section has duplication with a paragraph that comes slightly before this starting with "A peculiar choice of Laurel is that it does not require imperative code to be encapsulated using a functional specification...". Can the two be merged?

I think we could have a section title `## Specifications are optional`. In contrast to what the current documentation says, spec and implementation are not about functional versus imperative per se, although often the spec is functional and the implementation imperative. 

I think more generally, specs are useful because sometimes we write code in a certain way because it performs well, and in these cases there is a specification of that code which performs worse but is more readable. In these cases encapsulating the implementation using a contract is gives you more correctness and faster validation on the call-site.

Although rare, the spec and implementation can both be functional, like with a merge sort.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,52 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
 like conditionals and variable declarations. Each StmtExpr has a user facing type, which for
 statement-like constructs could be void.
 
+# Design Principles
+
+These principles recur in design discussions on Laurel pull requests and issues. Proposed
```
keyboardDrummer, line 89: I would leave both of these lines out. Of the first the relevance is unclear and the second seems like common sense.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,52 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+features or refactorings that violate them tend to be pushed back.
+
+## Shared between source languages, or don't add it
+
+Beyond the shared features listed in the introduction, the operative rule is that a Laurel
+feature must be "useful for Java and something else". Constructs that make sense only for one
+source language (for example, nominal subtyping for Java generics, or Python's `__getattr__`)
+belong in that front-end's translator, encoded using Laurel's primitives.
+
+Concrete instances of the principle:
```
keyboardDrummer, line 99: I think it would be interesting to categorize which features are used by which source languages, but then I would like to do that holistically.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,52 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+# Design Principles
+
+These principles recur in design discussions on Laurel pull requests and issues. Proposed
+features or refactorings that violate them tend to be pushed back.
+
+## Shared between source languages, or don't add it
+
+Beyond the shared features listed in the introduction, the operative rule is that a Laurel
+feature must be "useful for Java and something else". Constructs that make sense only for one
+source language (for example, nominal subtyping for Java generics, or Python's `__getattr__`)
```
keyboardDrummer, line 96: It seems dangerous to mention specific source language features because what's unique to a source language can change if we support more of them.

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,52 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
 
+# Design Principles
+
+These principles recur in design discussions on Laurel pull requests and issues. Proposed
+features or refactorings that violate them tend to be pushed back.
+
+## Shared between source languages, or don't add it
+
+Beyond the shared features listed in the introduction, the operative rule is that a Laurel
+feature must be "useful for Java and something else". Constructs that make sense only for one
```
keyboardDrummer, line 95: I wouldn't focus on Java, that's not by design.

docs/verso/LaurelDoc.lean
keyboardDrummer: That Laurel lowers to Core is part of the implementation. Shouldn't we fully cover the user facing design because mentioning implementation?

docs/verso/LaurelDoc.lean
```diff
@@ -146,11 +192,88 @@ A Laurel program consists of procedures, global variables, type definitions, and
...
 
+# Conventions and Invariants
+
+## Holes are the uniform "unknown" marker
+
+The `.Hole` constructor of `StmtExpr` is Laurel's single representation for an expression
+whose identity is not yet known. The `StmtExpr` documentation above describes when
+front-ends and error-recovery paths emit one.
+
+`InferHoleTypes` labels every hole with the type its context demands. `EliminateHoles` then
```
keyboardDrummer, line 203: The passes are covered somewhere else.

docs/verso/LaurelDoc.lean
```diff
@@ -146,11 +192,88 @@ A Laurel program consists of procedures, global variables, type definitions, and
 
 {docstring Strata.Laurel.Program}
 
+# Conventions and Invariants
```
keyboardDrummer, line 195: Is part `# Conventions and Invariants` part of the language definition or part of the implementation?

docs/verso/LaurelDoc.lean
keyboardDrummer: "Reserved identifier prefixes" is a bug that's documented in the referenced issue. I think this entire section is unnecessary.

docs/verso/LaurelDoc.lean
keyboardDrummer: I'm still not really excited about this section. I would be excited if it had a table that tracks which features relate to which source languages, but that's difficult to get because even splitting the language into independent features is not easy.

Removing everything from the section that's about source languages and not about Laurel, the only part that I would keep is what the introduction already says "Laurel tries to include any features that are common to those three languages."

docs/verso/LaurelDoc.lean
```diff
@@ -84,6 +84,52 @@ share a single implementation type, the StmtExpr. This reduces duplication for c
...
+- A Laurel procedure body may be transparent to its callers. There is no requirement that a
+  caller only see an `ensures` clause; if the body has no postcondition the caller can see
+  (and reason about) the implementation. This is the origin of the transparent / opaque /
+  `Body.External` distinction.
+- Statements are allowed in expression positions in principle. The compilation pipeline has
+  to lift them out eventually (see Lifting Expression Assignments below), but the AST
+  admits programs in which an assignment appears as part of an expression. The fact that
+  Core cannot represent such programs is not Laurel's problem.
+
+## One AST node, many back-ends
```
keyboardDrummer, line 118: I like to phrase things in terms of the user value. I think the term "AST" and "back-end" are irrelevant to the user. What the user wants to know is what features the tool has.

I guess what I'm saying is that for `LaurelDoc` I would like to start out with information that's relevant for the user of Laurel.

There can be a different section that says what's relevant to the designer of Laurel, but that should really come after. We should probably put that in a different file even to keep the two well separated.

But then even in a `LaurelDesignerNotes` doc I would phrase this section as `One language, many analyses` since even for the Laurel language design, the terms AST and the back-ends are implementation details.

So, to conclude. I think it's best to have three separate documentation files:
1. LaurelUserGuide
1. LaurelDesignerGuide
1. LaurelImplementationGuide

Right now, also due to myself, all three of those parts are mixed into this doc. This PR is adding a lot of (2) and (3), but I would like to keep that separate from (1).

Strata/Languages/Laurel/MapStmtExpr.lean
```diff
@@ -135,4 +135,143 @@ def mapProgram (f : StmtExprMd → StmtExprMd) (program : Program) : Program :=
   mapProgramM (m := Id) f program
 
 end -- public section
+
+/-! ## Build-time traversal-coverage check
```
keyboardDrummer, line 139: Doesn't Lean already guarantee that we're handling all constructors of the type tree under `Laurel.Program` ?

keyboardDrummer (commented):
Hey, left some more comments. Sorry about pushing back so much. I think documentation is a rather subjective topic to begin with, and on top of that one that I'm quite particular about. If you want I can also contribute with changes to this PR, instead of just comments.

PR #1145, opened 2026-05-08
Fix bug: ADT constructors do not change `Map` to `Array` when using useArrayTheory option.

thanhnguyen-aws:
*Description of changes:*

**Bug:** When `useArrayTheory` is enabled, the SMT encoder correctly converts `Map` types to `Array` in variable declarations and function signatures, but it did *not* perform this conversion for fields inside ADT (algebraic data type) constructor declarations. This caused a type mismatch: a datatype field would be declared with type `Map` while the rest of the encoding used `Array`.

**Fix:** Thread the `useArrayTheory` flag through the datatype emission pipeline:
- `lMonoTyToTermType` now accepts `useArrayTheory` and converts `Map` to `Array` when enabled.
- `datatypeConstructorsToSMT` passes the flag to `lMonoTyToTermType`.
- `SMT.Context.emitDatatypes` accepts and forwards the flag.
- `encodeCore` in `Verifier.lean` passes `options.useArrayTheory` to `emitDatatypes`.

**Test:** Added a unit test in `SMTEncoderDatatypeTest.lean` that verifies a datatype with a `Map`-typed field emits `(Array Int Int)` when `useArrayTheory=true` and `(Map Int Int)` when `useArrayTheory=false`.

### DIFF
--- a/StrataTest/Languages/Core/Tests/SMTEncoderDatatypeTest.lean
+++ b/StrataTest/Languages/Core/Tests/SMTEncoderDatatypeTest.lean
@@ -71,13 +71,13 @@ def treeDatatype : LDatatype Unit :=
 Convert an expression to full SMT string including datatype declarations.
 `blocks` is a list of mutual blocks (each block is a list of mutually recursive datatypes).
 -/
-def toSMTStringWithDatatypeBlocks (e : LExpr CoreLParams.mono) (blocks : List (List (LDatatype Unit))) : IO String := do
+def toSMTStringWithDatatypeBlocks (e : LExpr CoreLParams.mono) (blocks : List (List (LDatatype Unit))) (useArrayTheory : Bool := false): IO String := do
   match Env.init.addDatatypes blocks with
   | .error msg => return s!"Error creating environment: {msg}"
   | .ok env =>
     -- Set the TypeFactory for correct datatype emission ordering
     let ctx := SMT.Context.default.withTypeFactory env.datatypes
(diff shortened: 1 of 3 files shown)


thanhnguyen-aws:
@keyboardDrummer-bot Please add more PR description and add a test.

StrataTest/Languages/Core/Tests/SMTEncoderDatatypeTest.lean
```diff
@@ -511,6 +511,68 @@ info: (declare-datatype IntList (
   [[intListDatatype]]
   listLenFunc
 
+/-! ## useArrayTheory Tests for Datatype Constructors -/
+
+/-- Helper that emits datatypes with useArrayTheory=true -/
+def toSMTStringWithDatatypesArrayTheory (e : LExpr CoreLParams.mono) (datatypes : List (LDatatype Unit)) : IO String := do
```
joscoh: Is it possible to just add a flag to the existing `toSMTStringWithDatatypeBlocks`/`toSMTStringWithDatatypes` rather than copying the function?

joscoh (commented):
Can you please add a test using the testers/destructors? This PR does not change their types, but it seems like it might need to.

aqjune-aws (approved):

joscoh (approved):

aqjune-aws:
I will merge this PR after the internal bnechmark CI check passes on https://github.com/strata-org/Strata/pull/1178.

PR #1149, opened 2026-05-08
Fix namespace collision and SMACK assert encoding in BoogieToStrata

### LINKED ISSUE #1148: BoogieToStrata: Blockers preventing SMACK-generated Boogie verification
Testing the SMACK → BoogieToStrata → Strata pipeline on 12 C programs reveals systematic issues that block end-to-end verification. All programs that pass translation (8/12) fail at verification time.

See [`Tools/BoogieToStrata/Docs/aws-c-common-test-results.md`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Docs/aws-c-common-test-results.md) on branch `htd/smack` for full test results.

### 1. Namespace collision: constant and procedure sharing the same name

SMACK emits `const main : ref;` (function pointer address) alongside `procedure main(...)`. Strata Core has a single namespace, rejecting this as a duplicate declaration.

**Impact**: 7/12 programs
**Fix**: Rename or skip constants that collide with procedure names in `StrataGenerator.cs`.

Details: [`Tools/BoogieToStrata/Docs/issues/issue-namespace-collision.md`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Docs/issues/issue-namespace-collision.md)
Reproducer: [`Tools/BoogieToStrata/Tests/NamespaceCollision.bpl`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Tests/NamespaceCollision.bpl)

---

### 2. Type synonym chain not resolved for comparison/arithmetic operators

Strata's DDM translator panics on `<=`, `<`, `>=`, `>`, `+`, `-`, `*`, `div`, `mod` when operands have a type that is a synonym of `int` (e.g., `ref := i64 := int`). Equality (`==`) is unaffected.

**Impact**: 1/12 programs (blocks more once issue 1 is fixed)
**Fix**: Make `dealiasTypeExpr` in `Strata/Languages/Core/DDMTransform/Translate.lean:452` resolve transitively.

Details: [`Tools/BoogieToStrata/Docs/issues/issue-type-synonym-comparison.md`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Docs/issues/issue-type-synonym-comparison.md)
Reproducer: [`Tools/BoogieToStrata/Tests/TypeSynonymComparison.core.st`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Tests/TypeSynonymComparison.core.st)

---

### 3. SMACK's `assert_` call pattern not recognized as verification condition

SMACK encodes C `assert(expr)` as `call assert_.i32(cond)`. BoogieToStrata emits this as an opaque procedure call, so zero VCs are generated — even `assert(false)` passes silently.

**Impact**: All programs (latent — only visible once issues 1 and 2 are fixed)
**Fix**: In `VisitCallCmd`, intercept calls matching `assert_.*` and emit `assert (arg != 0);`. The `VisitAssertCmd` infrastructure already exists.

Details: [`Tools/BoogieToStrata/Docs/issues/issue-smack-assert-encoding.md`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Docs/issues/issue-smack-assert-encoding.md)
Reproducer: [`Tools/BoogieToStrata/Tests/SmackAssert.bpl`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Tests/SmackAssert.bpl)

### LINKED ISSUE #1152: BoogieToStrata: Sanitization maps distinct Boogie names to the same Strata identifier
`SanitizeNameForStrata` replaces `@`, `.`, `#`, `^`, and `$` with `_`. This means distinct Boogie identifiers that differ only in these characters collide in the Strata output, causing duplicate definition errors.

Unlike the cross-namespace collision fixed in #1148, this affects entities in the **same** namespace (e.g., two functions both named `_add_i32` after sanitization).

### Example

```boogie
function {:inline} $add.i32(i1: i32, i2: i32) returns (i32) { i1 + i2 }
function {:inline} $add_i32(i1: i32, i2: i32) returns (i32) { i1 + i2 }
```

Both sanitize to `_add_i32`, producing:

```
error: '_add_i32' already defined
```

### Reproducer

[`Tools/BoogieToStrata/Tests/SanitizationCollision.bpl`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Tests/SanitizationCollision.bpl)

### Details & suggested fixes

[`Tools/BoogieToStrata/Docs/issues/issue-sanitization-collision.md`](https://github.com/strata-org/Strata/blob/htd/smack/Tools/BoogieToStrata/Docs/issues/issue-sanitization-collision.md)

PROgram52bc:
*Issue #, if available:* #1148

*Description of changes:*

1. Namespace collision (issue 1 in #1148): Constants that share a name with a procedure are now prefixed with __const_ in the Strata output.  The rename is applied consistently in declarations and references.

2. Recursive synonym resolution (Fixes issue 2 in #1148): DealiasTypeExpr now recurses on its result, resolving synonym chains like ref := i64 := int fully to the base type. Previously it resolved only one level, causing panics on comparison/arithmetic operators applied to multi-level type synonyms.

3. SMACK assert encoding (issue 3 in #1148): Procedures named assert_.* now get a requires (arg != 0) precondition emitted. Call elimination generates VCs at each call site, making SMACK assertions verifiable.

4. InferModifies: SMACK-generated Boogie often omits explicit `modifies` clauses on procedures that mutate globals. Setting `InferModifies = true` in BoogieToStrata has two effects: (1) it runs `ModSetCollector.CollectModifies(program)` to populate empty modifies clauses, and (2) through `CheckModifies` in Boogie's resolution context, it suppresses typechecking of modifies clauses. Without this, `ResolveAndTypecheck` would reject SMACK programs that omit modifies clauses.

### DIFF
--- a/Strata/Languages/Core/DDMTransform/Translate.lean
+++ b/Strata/Languages/Core/DDMTransform/Translate.lean
@@ -453,8 +453,8 @@ partial def dealiasTypeExpr (p : Program) (te : TypeExpr) : TypeExpr :=
   match te with
   | (.fvar _ idx #[]) =>
     match p.globalContext.kindOf! idx with
-    | .expr te => te
-    | .type [] (.some te) => te
+    | .expr te => dealiasTypeExpr p te
+    | .type [] (.some te) => dealiasTypeExpr p te
     | _ => te
   | _ => te
 
--- a/Tools/BoogieToStrata/Source/BoogieToStrata.cs
+++ b/Tools/BoogieToStrata/Source/BoogieToStrata.cs
@@ -3,29 +3,59 @@ using Microsoft.Boogie;
 namespace BoogieToStrata;
 
 public static class BoogieToStrata {
+    private const string Usage = "Usage: BoogieToStrata [--smack] <inputFile>";
+
+    private static bool _smack;
+
     private static void PrintResolvedProgram(ExecutionEngineOptions options, ProcessedProgram prog) {
         var writer = new TokenTextWriter(Console.Out, options);
-        // smack: true here preserves pre-task behavior; Task 3 wires the real value
-        // from the parsed CLI flag.
-        StrataGenerator.EmitProgramAsStrata(options, prog.Program, writer, smack: true);
+        StrataGenerator.EmitProgramAsStrata(options, prog.Program, writer, _smack);
+    }
+
+    /// <summary>
+    /// Parse args into (smack, filename). Returns false on any malformed
+    /// invocation (zero or two-plus positional args, unknown flags); the
+    /// caller should print Usage and return exit code 1.
+    /// </summary>
+    private static bool TryParseArgs(string[] args, out bool smack, out string filename) {
+        smack = false;
+        filename = "";
+        string? positional = null;
+        foreach (var arg in args) {
+            if (arg == "--smack") {
+                smack = true;
+            } else if (arg.StartsWith("--")) {
+                return false; // unknown flag
+            } else if (positional == null) {
+                positional = arg;
+            } else {
+                return false; // two positional args
+            }
+        }
+        if (positional == null) return false; // no positional arg
+        filename = positional;
+        return true;
     }
 
     public static int Main(string[] args) {
-        if (args.Length != 1) {
-            Console.Error.WriteLine("Usage: BoogieToStrata <inputFile>");
+        if (!TryParseArgs(args, out var smack, out var filename)) {
+            Console.Error.WriteLine(Usage);
             return 1;
         }
-
-        var filename = args[0];
+        _smack = smack;
 
         var options = new CommandLineOptions(Console.Out, new ConsolePrinter()) {
             Verify = false,
             TypeEncodingMethod = CoreOptions.TypeEncoding.Predicates,
-            // SMACK-generated Boogie often omits explicit `modifies` clauses on procedures that mutate globals.
-            // InferModifies runs ModSetCollector.CollectModifies to populate empty modifies clauses and
-            // suppresses modifies-clause typechecking (via CheckModifies), so that ResolveAndTypecheck
+            // Under --smack, SMACK-generated Boogie often omits explicit
+            // `modifies` clauses on procedures that mutate globals.
+            // InferModifies runs ModSetCollector.CollectModifies to populate
+            // empty modifies clauses and suppresses modifies-clause
+            // typechecking (via CheckModifies), so that ResolveAndTypecheck
             // does not reject SMACK programs missing modifies clauses.
-            InferModifies = true
+            // For strict Boogie input (no --smack), this stays false and
+            // missing modifies clauses are reported as typecheck errors.
+            InferModifies = smack
         };
 
         var boogieEngine = ExecutionEngine.CreateWithoutSharedCache(options);
--- a/Tools/BoogieToStrata/Source/BoogieToStrata.cs
+++ b/Tools/BoogieToStrata/Source/BoogieToStrata.cs
@@ -5,7 +5,9 @@ namespace BoogieToStrata;
 public static class BoogieToStrata {
     private static void PrintResolvedProgram(ExecutionEngineOptions options, ProcessedProgram prog) {
         var writer = new TokenTextWriter(Console.Out, options);
-        StrataGenerator.EmitProgramAsStrata(options, prog.Program, writer);
+        // smack: true here preserves pre-task behavior; Task 3 wires the real value
+        // from the parsed CLI flag.
+        StrataGenerator.EmitProgramAsStrata(options, prog.Program, writer, smack: true);
     }
 
     public static int Main(string[] args) {
--- a/Tools/BoogieToStrata/Source/BoogieToStrata.cs
+++ b/Tools/BoogieToStrata/Source/BoogieToStrata.cs
@@ -18,7 +18,9 @@ public static class BoogieToStrata {
 
         var options = new CommandLineOptions(Console.Out, new ConsolePrinter()) {
             Verify = false,
-            TypeEncodingMethod = CoreOptions.TypeEncoding.Predicates
+            TypeEncodingMethod = CoreOptions.TypeEncoding.Predicates,
+            InferModifies = true,
+            Prune = true
         };
 
         var boogieEngine = ExecutionEngine.CreateWithoutSharedCache(options);
--- a/Tools/BoogieToStrata/Source/BoogieToStrata.cs
+++ b/Tools/BoogieToStrata/Source/BoogieToStrata.cs
@@ -19,8 +19,7 @@ public static class BoogieToStrata {
         var options = new CommandLineOptions(Console.Out, new ConsolePrinter()) {
             Verify = false,
             TypeEncodingMethod = CoreOptions.TypeEncoding.Predicates,
-            InferModifies = true,
-            Prune = true
+            InferModifies = true
         };
 
         var boogieEngine = ExecutionEngine.CreateWithoutSharedCache(options);
--- a/Tools/BoogieToStrata/Source/BoogieToStrata.cs
+++ b/Tools/BoogieToStrata/Source/BoogieToStrata.cs
@@ -19,6 +19,10 @@ public static class BoogieToStrata {
         var options = new CommandLineOptions(Console.Out, new ConsolePrinter()) {
             Verify = false,
             TypeEncodingMethod = CoreOptions.TypeEncoding.Predicates,
+            // SMACK-generated Boogie often omits explicit `modifies` clauses on procedures that mutate globals.
+            // InferModifies runs ModSetCollector.CollectModifies to populate empty modifies clauses and
+            // suppresses modifies-clause typechecking (via CheckModifies), so that ResolveAndTypecheck
+            // does not reject SMACK programs missing modifies clauses.
             InferModifies = true
         };
 
--- a/Tools/BoogieToStrata/Source/StrataGenerator.cs
+++ b/Tools/BoogieToStrata/Source/StrataGenerator.cs
@@ -56,6 +56,13 @@ public class StrataGenerator : ReadOnlyVisitor {
     // declaration. Keyed by Boogie Declaration object to avoid ambiguity
     // when two entities share the same original name (e.g., const main and
     // procedure main). First-seen wins; later entities get prefixed.
+    //
+    // Registration order determines who wins a collision:
+    //   1. Procedures  — registered first, always keep their name.
+    //   2. Implementations — claimed defensively (they share names with
+    //      their procedures, but claiming guards against edge cases).
+    //   3. Constants, Functions, Globals — registered last; in a
+    //      proc-vs-const collision the constant is always renamed.
     private readonly Dictionary<Declaration, string> _renames = new();
 
     private StrataGenerator(VCGenOptions options, TokenTextWriter writer, Program program) {
@@ -87,30 +94,22 @@ public class StrataGenerator : ReadOnlyVisitor {
             // First-seen wins; colliding entities get a suffix (_2, _3, ...).
             var claimed = new HashSet<string>();
 
-            void ClaimOrRename(Declaration decl, string originalName, string prefix) {
-                var sanitized = SanitizeNameForStrata(originalName);
-                if (claimed.Add(sanitized)) return;
-                var candidate = $"{prefix}{sanitized}";
-                if (!claimed.Add(candidate)) {
-                    var i = 2;
-                    while (!claimed.Add($"{candidate}_{i}")) i++;
-                    candidate = $"{candidate}_{i}";
-                }
-                generator._renames[decl] = candidate;
-            }
-
             foreach (var proc in p.Procedures)
-                ClaimOrRename(proc, proc.Name, "__proc_");
+                ClaimOrRename(proc, proc.Name, "__proc_", claimed, generator._renames);
+            // Defensive: implementations share names with their corresponding
+            // procedures, so they would normally never collide. Claiming them
+            // here guards against edge cases (e.g., an implementation whose
+            // procedure was pruned or renamed upstream).
             foreach (var impl in p.Implementations) {
                 var sanitized = SanitizeNameForStrata(impl.Name);
                 claimed.Add(sanitized);
             }
             foreach (var c in liveDeclarations.OfType<Constant>())
-                ClaimOrRename(c, c.TypedIdent.Name, "__const_");
+                ClaimOrRename(c, c.TypedIdent.Name, "__const_", claimed, generator._renames);
             foreach (var f in liveDeclarations.OfType<Function>())
-                ClaimOrRename(f, f.Name, "__func_");
+                ClaimOrRename(f, f.Name, "__func_", claimed, generator._renames);
             foreach (var g in p.GlobalVariables)
-                ClaimOrRename(g, g.Name, "__var_");
+                ClaimOrRename(g, g.Name, "__var_", claimed, generator._renames);
 
             var typeConstructors = p.TopLevelDeclarations.OfType<TypeCtorDecl>().ToList();
             if (typeConstructors.Count != 0) {
@@ -241,6 +240,29 @@ public class StrataGenerator : ReadOnlyVisitor {
             .Replace('#', '_')
             .Replace('^', '_')
             .Replace("$", "_");
+    }
+
+    /// <summary>
+    /// Claim a sanitized name for <paramref name="decl"/>, or rename it if the
+    /// name is already taken. The first declaration to claim a name wins;
+    /// subsequent colliders get a prefixed (and possibly suffixed) name recorded
+    /// in <paramref name="renames"/>.
+    /// </summary>
+    private static void ClaimOrRename(
+        Declaration decl,
+        string originalName,
+        string prefix,
+        HashSet<string> claimed,
+        Dictionary<Declaration, string> renames) {
+        var sanitized = SanitizeNameForStrata(originalName);
+        if (claimed.Add(sanitized)) return;
+        var candidate = $"{prefix}{sanitized}";
+        if (!claimed.Add(candidate)) {
+            var i = 2;
+            while (!claimed.Add($"{candidate}_{i}")) i++;
+            candidate = $"{candidate}_{i}";
+        }
+        renames[decl] = candidate;
     }
 
     private void AddUniqueConst(Type t, string name) {
--- a/Tools/BoogieToStrata/Source/StrataGenerator.cs
+++ b/Tools/BoogieToStrata/Source/StrataGenerator.cs
@@ -1850,14 +1850,17 @@ public class StrataGenerator : ReadOnlyVisitor {
 
     public override Procedure VisitProcedure(Procedure node) {
         if (!_program.Implementations.Any(i => i.Name.Equals(node.Name))) {
-            // SMACK encodes C assert(expr) as a call to assert_.*(cond).
-            // Inject a synthetic requires precondition so the call-elimination
-            // pass generates a VC checking the condition is non-zero.
-            // We add it to node.Requires so WriteProcedureHeader emits it
-            // inside a single spec block alongside any existing specs.
+            // Under --smack, SMACK encodes C assert(expr) as a call to
+            // assert_.*(cond). Inject a synthetic requires precondition so the
+            // call-elimination pass generates a VC checking the condition is
+            // non-zero. We add it to node.Requires so WriteProcedureHeader
+            // emits it inside a single spec block alongside any existing
+            // specs. The injection always fires when the name pattern matches
+            // (no Requires.Count == 0 guard) — if the procedure already has a
+            // hand-written requires, both clauses appear in the merged spec
+            // block, preserving the SMACK invariant unconditionally.
             Requires? syntheticReq = null;
-            if (node.Name.StartsWith("assert_.") && node.InParams.Count > 0
-                && node.Requires.Count == 0) {
+            if (_smack && node.Name.StartsWith("assert_.") && node.InParams.Count > 0) {
                 var param = node.InParams[0];
                 var paramExpr = new IdentifierExpr(param.tok, param);
                 var zero = new LiteralExpr(param.tok, Microsoft.BaseTypes.BigNum.FromInt(0));
--- a/Tools/BoogieToStrata/Source/StrataGenerator.cs
+++ b/Tools/BoogieToStrata/Source/StrataGenerator.cs
@@ -1816,22 +1816,30 @@ public class StrataGenerator : ReadOnlyVisitor {
 
     public override Procedure VisitProcedure(Procedure node) {
         if (!_program.Implementations.Any(i => i.Name.Equals(node.Name))) {
-            WriteProcedureHeader(node);
             // SMACK encodes C assert(expr) as a call to assert_.*(cond).
-            // Emit a requires precondition so the call-elimination pass
-            // generates a VC checking the condition is non-zero.
+            // Inject a synthetic requires precondition so the call-elimination
+            // pass generates a VC checking the condition is non-zero.
+            // We add it to node.Requires so WriteProcedureHeader emits it
+            // inside a single spec block alongside any existing specs.
+            Requires? syntheticReq = null;
             if (node.Name.StartsWith("assert_.") && node.InParams.Count > 0
                 && node.Requires.Count == 0) {
-                var firstParam = Name(node.InParams[0].TypedIdent.Name);
-                WriteLine("spec {");
-                IncIndent();
-                Indent();
-                WriteLine($"requires ({firstParam} != 0);");
-                DecIndent();
-                WriteText("}");
+                var param = node.InParams[0];
+                var paramExpr = new IdentifierExpr(param.tok, param);
+                var zero = new LiteralExpr(param.tok, Microsoft.BaseTypes.BigNum.FromInt(0));
+                var neqExpr = Expr.Neq(paramExpr, zero);
+                syntheticReq = new Requires(false, neqExpr);
+                node.Requires.Add(syntheticReq);
             }
+
+            WriteProcedureHeader(node);
             WriteLine(";");
             WriteLine();
+
+            // Remove the synthetic requires to avoid mutating the shared AST.
+            if (syntheticReq != null) {
+                node.Requires.Remove(syntheticReq);
+            }
         }
 
         return node;
--- a/Tools/BoogieToStrata/Source/StrataGenerator.cs
+++ b/Tools/BoogieToStrata/Source/StrataGenerator.cs
@@ -962,7 +962,7 @@ public class StrataGenerator : ReadOnlyVisitor {
 
     private void EmitSimpleAssign(SimpleAssignLhs lhs, Expr rhs) {
         Indent();
-        WriteText($"{Name(lhs.AssignedVariable.Name)} := ");
+        WriteText($"{NameOf(lhs.AssignedVariable.Decl, lhs.AssignedVariable.Name)} := ");
         VisitExpr(rhs);
         WriteLine(";");
     }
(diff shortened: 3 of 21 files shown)


Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -453,8 +453,8 @@ partial def dealiasTypeExpr (p : Program) (te : TypeExpr) : TypeExpr :=
   match te with
   | (.fvar _ idx #[]) =>
     match p.globalContext.kindOf! idx with
-    | .expr te => te
-    | .type [] (.some te) => te
+    | .expr te => dealiasTypeExpr p te
+    | .type [] (.some te) => dealiasTypeExpr p te
     | _ => te
```
tautschnig, line 458: The recursive fix is the right change. Two follow-ups, both worth considering:

1. **Termination.** The function is `partial def`, so Lean doesn't check termination. Boogie's resolver rejects cyclic synonyms (verified: `type A = B; type B = A;` is rejected with `type synonym could not be resolved because of cycles`), so on well-formed input this is safe. But other ingress paths to `Translate.lean` don't necessarily have that guarantee. For defence-in-depth, consider a `fuel : Nat` parameter or a `HashSet FreeVarIndex` of visited indices.

2. **Proof opportunities.** Two natural lemmas that would make this change load-bearing rather than opportunistic:
   ```lean
   theorem dealiasTypeExpr_idempotent (p : Program) (te : TypeExpr) :
       dealiasTypeExpr p (dealiasTypeExpr p te) = dealiasTypeExpr p te

   -- "result is not a resolvable alias"
   theorem dealiasTypeExpr_fully_unwound (p : Program) (te : TypeExpr) :
       match dealiasTypeExpr p te with
       | .fvar _ idx #[] =>
           match p.globalContext.kindOf! idx with
           | .expr _ | .type [] (.some _) => False
           | _ => True
       | _ => True
   ```
   Both require either a termination measure or proofs against a bounded-fuel variant. If that's out of scope for this PR, at minimum please add a Lean-side `#guard` test on a synthesised `Program` with a 3-level chain, and a corresponding `.bpl`/`.expect` end-to-end test (see separate comment).

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -453,8 +453,8 @@ partial def dealiasTypeExpr (p : Program) (te : TypeExpr) : TypeExpr :=
   match te with
   | (.fvar _ idx #[]) =>
     match p.globalContext.kindOf! idx with
-    | .expr te => te
-    | .type [] (.some te) => te
+    | .expr te => dealiasTypeExpr p te
+    | .type [] (.some te) => dealiasTypeExpr p te
```
tautschnig, line 457: There's no test exercising this code path in the PR. Issue #1148 referenced `TypeSynonymComparison.core.st` as the reproducer, but it wasn't carried over. The existing `TypeSynonyms{1,2}.bpl` only use single-level synonyms, so the old (non-recursive) behaviour would still pass them. Please add at least a minimal `.bpl`:

```boogie
// Tools/BoogieToStrata/Tests/TypeSynonymChain.bpl
type i64 = int;
type ref = i64;
procedure main() {
  var a: ref; var b: ref;
  a := 1; b := 2;
  assert (a < b);
  assert (a + b == 3);
}
```

with a matching `.expect` that contains `All 2 goals passed.` so CI catches regressions.

Tools/BoogieToStrata/Source/BoogieToStrata.cs
```diff
@@ -18,7 +18,8 @@ public static int Main(string[] args) {
 
         var options = new CommandLineOptions(Console.Out, new ConsolePrinter()) {
             Verify = false,
-            TypeEncodingMethod = CoreOptions.TypeEncoding.Predicates
+            TypeEncodingMethod = CoreOptions.TypeEncoding.Predicates,
+            InferModifies = true
```
tautschnig: `InferModifies = true` isn't mentioned in the PR description, and it has two effects that are load-bearing for SMACK but not obvious:

1. It runs `ModSetCollector.CollectModifies(program)`, which **mutates** the Boogie `Program` to add inferred `modifies` clauses to every procedure.
2. Through `CheckModifies => Proc != null && (!Options?.InferModifies ?? true)` in Boogie's resolution context, it **suppresses** typechecking of modifies clauses.

Both make sense for SMACK-generated Boogie (which can omit `modifies`), but please (a) call this out in the PR body and motivate it, and (b) add a `.bpl` test with a procedure that mutates a global and has no `modifies` clause — something like:

```boogie
var g: int;
procedure p() { g := 1; }  // modifies g;  -- intentionally omitted
```

with an `.expect` that the translation completes and the output has `inout g` for `p`, so any future regression to this assumption (e.g. Boogie stopping inferring) is caught.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -52,6 +52,11 @@ public class StrataGenerator : ReadOnlyVisitor {
     // Global variables collected from the program, used to convert them
     // into inout/input parameters on procedure headers and call sites.
     private List<GlobalVariable> _globalVariables = [];
+    // Renames for declarations whose sanitized name collides with another
+    // declaration. Keyed by Boogie Declaration object to avoid ambiguity
+    // when two entities share the same original name (e.g., const main and
+    // procedure main). First-seen wins; later entities get prefixed.
+    private readonly Dictionary<Declaration, string> _renames = new();
```
tautschnig, line 66: Please document the ordering of `ClaimOrRename` calls (procs → impls → consts → funcs → globals — see lines 102-113) here. It's the tie-breaker for which declaration keeps its sanitized name in a collision, and the current wording "first-seen wins" leaves the definition of "first" implicit.

Also worth noting: a proc-vs-const collision always renames the const, never the proc. That's a design choice, not forced by anything — document it as such.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -74,6 +79,39 @@ public static void EmitProgramAsStrata(VCGenOptions options, Program p, TokenTex
...
+            foreach (var impl in p.Implementations) {
+                var sanitized = SanitizeNameForStrata(impl.Name);
+                claimed.Add(sanitized);
+            }
+            foreach (var c in liveDeclarations.OfType<Constant>())
+                ClaimOrRename(c, c.TypedIdent.Name, "__const_");
+            foreach (var f in liveDeclarations.OfType<Function>())
+                ClaimOrRename(f, f.Name, "__func_");
+            foreach (var g in p.GlobalVariables)
+                ClaimOrRename(g, g.Name, "__var_");
```
tautschnig: Two small points on this block:

1. The `ClaimOrRename` closure would be easier to review and test as a named `private static` helper taking the `claimed` set and `_renames` dict as parameters. Nested closures with mutable captures are harder to reason about, especially when one of the captures (`_renames`) is on another object.

2. The `impl` loop at 104-107 is effectively a no-op: an implementation in Boogie always has a matching procedure declaration that was processed at lines 102-103, so `claimed.Add(sanitizedImplName)` returns `false` and nothing happens. If this is defensive for implementations without matching procedures, add a comment; otherwise consider removing it.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -310,7 +354,7 @@ private void EmitOldExpr(Expr expr) {
         switch (expr) {
             case IdentifierExpr identExpr:
                 WriteText("old ");
-                WriteText(Name(identExpr.Name));
+                WriteText(identExpr.Decl != null ? NameOf(identExpr.Decl, identExpr.Name) : Name(identExpr.Name));
```
tautschnig: Two observations on the `identExpr.Decl != null ? ... : Name(...)` fallback (same pattern repeats at line 593):

- Post-Boogie-resolution, `Decl` should always be non-null. If it can be null in practice, please add a comment explaining when.
- Silently falling back to `Name(identExpr.Name)` emits the unrenamed identifier, which is **semantically wrong in the collision case** — the identifier might refer to an entity that was renamed to e.g. `__var_x`, but we'd emit `x`. Consider an explicit assertion (`Debug.Assert(identExpr.Decl != null)`) or throwing a `StrataConversionException`, rather than failing silently.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -953,7 +997,7 @@ public override Cmd VisitCallCmd(CallCmd node) {
         var modifiesNames = new HashSet<string>(callee.Modifies.Select(m => m.Name));
 
         Indent("call ");
-        WriteText($"{Name(callee.Name)}(");
+        WriteText($"{NameOf(callee, callee.Name)}(");
```
tautschnig, line 1034: Call-site handling is only _half_ of the rename refactor. The following name-emission sites are all still on the pre-rename path:

- **line 1005** (`inout {Name(g.Name)}`) and **line 1010** (`{Name(g.Name)}`): call-site arguments for globals. Globals that got renamed via `_renames` appear here under their original sanitized name.
- **line 965** (`{Name(lhs.AssignedVariable.Name)} := `, in `EmitSimpleAssign`): assignment LHS uses pre-rename name.
- **line 1030** (`havoc {Name(x.Name)};`, in `VisitHavocCmd`): havoc target uses pre-rename name. `x.Decl` is available (used two lines down in `EmitWhereAssumption(x.Decl.TypedIdent)`).
- **line 1847** (`{prefix}{Name(name)} : `, in `WriteFormals`): parameter names emitted here. When `WriteProcedureHeader` (line 1775) calls `WriteFormals(modifiesGlobals, ...)`, the `inout` parameter for a renamed global still uses the pre-rename name while the body (via `IdentifierExpr`) uses the post-rename name — the two desync.

Reproducer demonstrating the desync — `strata verify` on the generated output reports `Unknown expr identifier __var__a_b`:

```boogie
const $a.b: int;   axiom $a.b > 0;
var   $a_b: int;   // both sanitize to _a_b
procedure main() returns (r: int)
  modifies $a_b;
{
  $a_b := 1;
  havoc $a_b;
  r := $a.b + $a_b;
}
```

Suggested fix: route each of these through `NameOf(decl, originalName)`. At the `WriteFormals` site, this requires passing the `Variable` rather than just its name — either lift the globals-as-formals emission into `WriteProcedureHeader` directly, or pass a rename lookup into `WriteFormals`.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -1773,6 +1817,19 @@ private void WriteProcedureHeader(Procedure proc) {
     public override Procedure VisitProcedure(Procedure node) {
         if (!_program.Implementations.Any(i => i.Name.Equals(node.Name))) {
             WriteProcedureHeader(node);
+            // SMACK encodes C assert(expr) as a call to assert_.*(cond).
+            // Emit a requires precondition so the call-elimination pass
+            // generates a VC checking the condition is non-zero.
+            if (node.Name.StartsWith("assert_") && node.InParams.Count > 0
```
tautschnig: `node.Name.StartsWith("assert_")` is too permissive. Reproducer:

```boogie
procedure assert_helper(p: int) returns (r: int);
procedure main() returns (r: int) { call r := assert_helper(0); }
```

With this PR, `assert_helper` gets `requires (p != 0)` injected and any caller passing 0 fails verification even though the user never wrote such a precondition. Any Boogie program with a procedure named `assert_helper`, `assert_invariant`, `assert_equal`, etc. starts failing.

SMACK's actual encoding is `assert_.<type>` (with a literal dot). Tighten the test:

```csharp
if (node.Name.StartsWith("assert_.") && node.InParams.Count > 0
    && node.Requires.Count == 0) { ... }
```

or use `Regex.IsMatch(node.Name, @"^assert_\.")`. Please also cover this with a negative test case (a user-named `assert_helper` whose `.expect` does **not** contain a bogus `requires` failure).

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -1773,6 +1817,19 @@ private void WriteProcedureHeader(Procedure proc) {
...
+                && node.Requires.Count == 0) {
+                var firstParam = Name(node.InParams[0].TypedIdent.Name);
+                WriteLine("spec {");
+                IncIndent();
+                Indent();
+                WriteLine($"requires ({firstParam} != 0);");
+                DecIndent();
+                WriteText("}");
+            }
             WriteLine(";");
```
tautschnig: This block duplicates the `spec { requires ...; ensures ...; }` emission pattern in `WriteProcedureHeader` (around lines 1781-1800). Consider injecting a synthetic `Requires` (from the Boogie API: `new Requires(false, cond)`) into `node.Requires` **before** calling `WriteProcedureHeader`, so the existing spec-block emission logic handles it uniformly and this ad-hoc second `spec {}` block goes away.

Minor: `var firstParam = Name(node.InParams[0].TypedIdent.Name);` uses `Name(...)` rather than `NameOf(node.InParams[0], ...)`. InParams are fresh local declarations so they shouldn't collide, but for consistency with the rest of the rename refactor, use `NameOf`.

Tools/BoogieToStrata/Tests/SmackAssert.bpl
```diff
@@ -0,0 +1,21 @@
...
+
+procedure assert_.i32(p.0: i32) returns ($r: i32);
+
+procedure main() returns ($r: i32)
+{
+  // assert(false) — should fail verification
+  call $r := assert_.i32(0);
+  $r := 0;
+  return;
+}
```
tautschnig, line 18: **Critical coverage gap.** This test has no `.expect` file, so `VerifyTestFile` adds `--check` and the strata verifier emits `Skipping verification.` — exactly what the harness is matching against. The verification failure that this fix is designed to produce (`call $r := assert_.i32(0)` → `❌ fail`) is never asserted, so CI passes even if the fix regresses.

I re-ran this manually and the fix does work:

```
SmackAssert.core.st(21, 6) [callElimAssert_assert__i32_requires_0_2]: ❌ fail
Finished with 0 goals passed, 1 failed.
```

Please add a `SmackAssert.expect` file matching this output (the word `failed` in the body makes the harness set `expectedExitCode = 2`, so it actually enforces the failure).

Tools/BoogieToStrata/Tests/NamespaceCollision.bpl
```diff
@@ -0,0 +1,25 @@
...
+{
+  var y: int;
+  x := 42;
+  assert x == 42;
+  // Use the constant in an expression that doesn't require the axiom to verify
+  y := 0;
+  if (main == 1000) { y := 1; }
+  // This assertion is trivially true regardless of the axiom
+  assert y == 0 || y == 1;
+}
```
tautschnig, line 25: Same coverage gap as `SmackAssert.bpl`: no `.expect` → `VerifyTestFile` skips verification via `--check`, so the two `assert` statements aren't actually checked in CI. The renamed constant `__const_main` happens to work end-to-end (I confirmed `All 2 goals passed.`), but without an `.expect` file we only verify that the translator doesn't crash.

Please add `NamespaceCollision.expect` with the expected `All 2 goals passed.` output.

Tools/BoogieToStrata/Tests/SanitizationCollision.bpl
```diff
@@ -0,0 +1,25 @@
...
+ensures r == 5;
+{
+  var a: i32;
+  var b: i32;
+  a := $add.i32(2, 3);
+  b := $add_i32(2, 3);
+  assert a == 5;
+  assert b == 5;
+  r := a;
+}
```
tautschnig, line 25: Same coverage gap — no `.expect`, so CI only checks translation + typechecking. Please add `SanitizationCollision.expect` containing the `All 3 goals passed.` output I observed manually so the actual post-rename semantics are pinned down.

(Side note on the generated name: the `__func_` prefix is applied to the already-sanitized name, which produces `__func__add_i32` with a double underscore. It works, but it's a bit awkward-looking — happy to leave it alone, but worth a comment in `ClaimOrRename` explaining the concatenation rule.)

tautschnig (commented):
Thanks for tackling all three #1148 issues in one PR — I ran the three fixes end-to-end on the included reproducers and each of them does what the description claims. The direction looks right, and I especially like that fix (3) follows atomb's suggestion of an opaque procedure with a precondition rather than inlining `assert`. I have a mix of blocking and non-blocking comments, grouped below:

### Blocking

- **Test harness gap.** `VerifyTestFile` adds `--check` when no `.expect` file exists, which literally emits `Skipping verification.` without invoking the solver. I re-ran the three new tests: with no `.expect`, `SmackAssert.bpl` passes CI even though the whole point of the test is that `call $r := assert_.i32(0);` should now fail verification. Please add `.expect` files for all three new tests (`SmackAssert.expect` crucially must contain a `failed` line so the harness flips `expectedExitCode` to 2 and actually catches regressions).
- **Fix (2) has no regression test.** `dealiasTypeExpr` now recurses, but there is no `.bpl` test exercising a multi-level synonym chain (the issue's reproducer `TypeSynonymComparison.core.st` wasn't carried over). Existing `TypeSynonyms{1,2}.bpl` only use single-level synonyms. Without a test, this fix has no CI coverage. A 10-line `.bpl` using `type i64 = int; type ref = i64;` with `<=` / `+` over `ref` would close this.
- **Rename mechanism is not applied consistently.** Four of the places that emit variable names still call `Name(...)` instead of `NameOf(decl, ...)`. This is only triggered when a global gets renamed (i.e. its sanitized name collides with a proc/const/func), but once it does, the output is broken. Reproducer:

  ```boogie
  const $a.b: int;   axiom $a.b > 0;
  var   $a_b: int;   // both sanitize to _a_b
  procedure main() returns (r: int)
    modifies $a_b;
  {
    $a_b := 1;
    havoc $a_b;
    r := $a.b + $a_b;
  }
  ```

  Produces (inout parameter name and havoc target still use the pre-rename identifier, while the identifier expression uses the renamed one):

  ```
  procedure main(inout _a_b : int, out r : int)
  {
    _a_b := 1;
    havoc _a_b;
    r := (_a_b + __var__a_b);   // __var__a_b is undefined
  }
  ```

  ...and `strata verify` reports `Unknown expr identifier __var__a_b`. See the inline comment at `VisitCallCmd` for the full list of affected sites.

- **`StartsWith("assert_")` is too permissive** and triggers on user-written procedures. Reproducer:

  ```boogie
  procedure assert_helper(p: int) returns (r: int);
  procedure main() returns (r: int) { call r := assert_helper(0); }
  ```

  emits `requires (p != 0);` on `assert_helper` even though the user never wrote it. Any Boogie program with a procedure whose name happens to start with `assert_` (e.g. `assert_helper`, `assert_invariant`, `assert_equal`) would start failing verification. SMACK's actual pattern is `assert_.<type>` (literal dot); please tighten to match that specifically.

### Non-blocking

- **`InferModifies = true` isn't mentioned in the PR body.** It's not just a knob — it runs `ModSetCollector` (mutating the program) and, via `CheckModifies`, suppresses modifies-clause typechecking. Please document the motivation and add a `.bpl` test with a procedure that mutates a global without an explicit `modifies` clause, so regressions in this new assumption are caught.
- **Proof opportunities for `dealiasTypeExpr`.** Since this is `partial def`, strong termination isn't provable without an acyclicity argument on `p.globalContext`. Natural lemmas that the rest of the file could rely on: `dealiasTypeExpr_idempotent : dealiasTypeExpr p (dealiasTypeExpr p te) = dealiasTypeExpr p te`, and a "result is fully unwound" lemma: if `dealiasTypeExpr p te` is `.fvar _ idx #[]`, then `p.globalContext.kindOf! idx` is neither `.expr _` nor `.type [] (.some _)`. If you don't want to prove these, at least consider a Lean-side `#guard` test constructing a three-level synonym `Program` and checking the result is `.int`.
- **Refactoring the rename builder.** The `ClaimOrRename` logic would be more readable as a named static helper that takes the `claimed` set and the `_renames` dict as parameters, rather than a closure inside `EmitProgramAsStrata`. The ordering "procs → impls → consts → funcs → globals" is a deliberate tie-breaker (first-seen wins) — please document it at the `_renames` field, since it's a semantic choice with implications (e.g. a name collision between a proc and a const always renames the const, never the proc).
- **Duplicated spec emission for `assert_.*`.** The new `spec { requires ... }` block in `VisitProcedure` duplicates the emission pattern in `WriteProcedureHeader`. Consider injecting a synthetic `Requires` into `node.Requires` just before calling `WriteProcedureHeader` so the existing code path handles it uniformly.
- **`IdentifierExpr` fallback.** `identExpr.Decl != null ? NameOf(...) : Name(...)`: post-Boogie-resolution, `Decl` should always be non-null. Silently falling back to `Name(...)` emits the unrenamed name and is wrong in the collision case. Consider asserting non-null instead, or at least add a comment about when `Decl` can be null and why the fallback is safe.

Strata/Languages/Core/DDMTransform/Translate.lean
PROgram52bc: Since the function has been declared as `partial def`, I think this issue deems another PR.

PROgram52bc: Agreed — deferring termination proofs to a separate PR. The function is `partial def`, so Lean doesn't enforce termination today. Boogie's resolver rejects cyclic type synonyms (verified: `type A = B; type B = A;` errors with "type synonym could not be resolved"), so in practice the recursion always terminates on well-formed input. A follow-up PR can add the idempotency and full-unwinding lemmas, plus a `#guard` test with a three-level synonym chain.

Tools/BoogieToStrata/Source/BoogieToStrata.cs
PROgram52bc: Addressed in commits `3a7fb909` and `f568f65f`:
- Added a 4-line code comment above `InferModifies = true` in BoogieToStrata.cs documenting both effects (ModSetCollector + CheckModifies suppression).
- Added `InferModifiesGlobal.bpl` regression test — a procedure that mutates a global with no explicit `modifies` clause, asserting the output contains `inout g`.
- Updated PR body with a new item covering InferModifies motivation.

Tools/BoogieToStrata/Source/StrataGenerator.cs
PROgram52bc: Addressed in commit `3a7fb909`. Expanded the `_renames` field comment (lines 55-66) to document the registration order: procs first (always keep their name), impls claimed defensively, then consts/funcs/globals. In a proc-vs-const collision, the constant is always renamed.

Tools/BoogieToStrata/Source/StrataGenerator.cs
PROgram52bc: Addressed in commit `3a7fb909`:
1. Extracted `ClaimOrRename` closure into `private static void ClaimOrRename(Declaration, string, string, HashSet<string>, Dictionary<Declaration, string>)` at line 251.
2. Added defensive comment before the implementation loop (lines 99-102) explaining that impls share names with their procedures but claiming guards against edge cases.

Tools/BoogieToStrata/Source/StrataGenerator.cs
PROgram52bc: Addressed in commit `7cf444c2`. Both fallback sites (EmitOldExpr and VisitExpr) now throw `StrataConversionException` instead of silently falling back to `Name()`. Added `OldExprRenameCollision.bpl` regression test verifying `old()` uses the renamed name on collision.

Tools/BoogieToStrata/Source/StrataGenerator.cs
PROgram52bc: Addressed in commit `46fde0d7`. All five identified call sites now route through `NameOf()` instead of `Name()`. The `GlobalVarRenameCollision.bpl` test covers the inout parameter, havoc target, and expression paths.

Tools/BoogieToStrata/Source/StrataGenerator.cs
PROgram52bc: Addressed in commit `0880ae33`. Replaced the ad-hoc `spec { requires ... }` block with a synthetic `Requires` AST node injected into `node.Requires` before calling `WriteProcedureHeader`, then removed after to avoid mutating the shared AST. Added `SmackAssertDuplicateSpec.bpl` test asserting exactly one `spec {` block in the output.

PROgram52bc:
Thanks @tautschnig for your detailed comments. I have addressed everything except for the `dealiasTypeExpr` proof, which could be completed in a separate PR, especially in light of the refactoring that's going to take place.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -1772,9 +1844,33 @@ private void WriteProcedureHeader(Procedure proc) {
...
-            WriteProcedureHeader(node);
-            WriteLine(";");
-            WriteLine();
+            // SMACK encodes C assert(expr) as a call to assert_.*(cond).
+            // Inject a synthetic requires precondition so the call-elimination
+            // pass generates a VC checking the condition is non-zero.
+            // We add it to node.Requires so WriteProcedureHeader emits it
+            // inside a single spec block alongside any existing specs.
+            Requires? syntheticReq = null;
+            if (node.Name.StartsWith("assert_.") && node.InParams.Count > 0
```
atomb: We might want more of a generic mechanism for adapting contracts from the input file in the long run, but I think this is fine for now.

atomb (dismissed):

MikaelMayer (commented):
🤖🔍 Good progress since the last round. The `.expect` files have been added for all test cases, `ClaimOrRename` is now a static helper, and the `assert_.` pattern (with the dot) correctly avoids false positives like `assert_helper`. The rename mechanism is well-structured and the test coverage is solid.

One remaining issue below.

Tools/BoogieToStrata/Source/StrataGenerator.cs
```diff
@@ -1772,9 +1844,33 @@ private void WriteProcedureHeader(Procedure proc) {
...
+                && node.Requires.Count == 0) {
```
MikaelMayer: 🤖🔍 The `Requires.Count == 0` guard means the synthetic `requires (p != 0)` is NOT injected when the procedure already has any requires clause. But `SmackAssertDuplicateSpec.bpl` tests the case where there's an existing `ensures` (not `requires`) — that still triggers injection correctly.

However: if a SMACK `assert_.i32` procedure ever has a pre-existing `requires` (e.g., from a manual annotation or another tool pass), the `!= 0` check silently disappears. Consider documenting this design choice in a comment, or changing the guard to always inject and deduplicate instead.

Tools/BoogieToStrata/Tests/SmackAssert.bpl
```diff
@@ -0,0 +1,21 @@
+// Minimal test case for SMACK assert_ pattern recognition.
+// SMACK encodes C assert(expr) as a call to assert_.i32(cond).
+// BoogieToStrata should recognize this pattern and emit:
+//   assert (cond != 0);
+// instead of an opaque procedure call.
+//
+// This test passes 0 (false) to assert — it should be a verification
```
MikaelMayer: That was true before this PR, but after this PR the error message changed.

MikaelMayer (commented):

Tools/BoogieToStrata/Source/StrataGenerator.cs
PROgram52bc: Done in 6fb6f9c19: dropped the `Requires.Count == 0` guard so the synthetic `requires (p.0 != 0)` is always added, given that `--smack` is provided. Tests extended in 5a1beaaf9.

Tools/BoogieToStrata/Tests/SmackAssert.bpl
PROgram52bc: Good point! Removed the stale paragraph in 5297e8dbf.

Tools/BoogieToStrata/Source/StrataGenerator.cs
PROgram52bc: I think that's a good idea. I started with a `--smack` option for now, and perhaps later we can have options for contracts coming from other tools.

MikaelMayer (approved):

atomb (approved):

PR #1151, opened 2026-05-08
Replace empty module name strings with validated ModuleName type

joehendrix:
## Summary

`PythonIdent.pythonModule` was a raw `String`, allowing empty module names to propagate silently through the system and cause downstream failures (see #1143). This PR changes `pythonModule` to use `ModuleName` — which now guarantees at least one non-empty component — and adopts it more widely through the PySpec pipeline, eliminating this class of bugs at the type level.

## Approach

Rather than adding runtime checks at each use site, `ModuleName` gains a strengthened invariant via a private constructor enforcing non-empty components. `PythonIdent.ofComponent` uses `decide` to verify non-emptiness at compile time for literal module names (used in `Decls.lean`). Runtime parsing goes through `ModuleName.ofString?` which returns `Option`.

## Key changes

- **PythonIdent.lean**: `PythonIdent.pythonModule` changed from `String` to `ModuleName`. New `ofComponent` constructor with compile-time proof for single-component literals. `ModuleName` API extended: `back`, `parent (n)`, `push`, `append`/`++`, `toString (sep)`
- **PythonIdent.lean**: New `ModuleName.ofRelativePath` provides a single canonical function for deriving a `ModuleName` from a `.py` file path relative to a search root, handling both regular files and `__init__.py` packages. This eliminates redundant ad-hoc implementations scattered across `discoverModules`, `RelativeImportTest`, and `findInPath` — each of which had slightly different `__init__.py` handling, creating inconsistency and latent bugs (e.g., `pySpecToLaurelCommand` silently failed for package modules)
- **Specs.lean**: `findInPath` simplified to return `(FilePath × Bool)` instead of `(FilePath × Array ModuleComponent)` — the `Bool` is `isInit`, and callers use `ModuleOfPath.modulePrefix` to derive the package prefix. Deleted unused `ofFile` function
- **Specs/DDM.lean**: Converted `fromDDM` functions to monadic `FromDDM` (`Except (SourceRange × String)`) for proper error propagation instead of panics
- **Specs/ToLaurel.lean**: `signaturesToLaurel` takes `ModuleName` instead of `String` prefix. Overload extraction factored into dedicated `ExtractM` monad (no dependency on `ToLaurelContext`)
- **PySpecPipeline.lean**: `buildPySpecLaurel` takes `Array (ModuleName × String)` instead of `Array (String × String)`
- **Specs.lean**: Import resolution uses `ModuleName` throughout; `translateFile` requires a `ModuleName` (no longer optional)
- **Specs/Decls.lean**: Builtin idents use `ofComponent` with compile-time proof

### Unrelated cleanup

- Deleted `Strata.Python.containsSubstr` helper and replaced all call sites with `String.contains` (which accepts substring patterns in Lean 4.29)


MikaelMayer (commented):
🤖🔍 Well-designed PR that eliminates a class of bugs at the type level. The `ModuleName` invariant is clean, `ofRelativePath` consolidates scattered logic, and the `FromDDM` monad properly replaces `panic!` with structured errors. A few issues below.

Strata/Languages/Python/PythonIdent.lean
```diff
@@ -0,0 +1,226 @@
...
+  ofStringAux mod #[] mod.startPos mod.startPos
+
+/--
+Parses a dot-separated module name string (e.g., "typing.List")
+and panics if parsing fails.
+-/
+def ofString! (mod : String) : ModuleName :=
+  match ofStringAux mod #[] mod.startPos mod.startPos with
+  | .some m => m
+  | .none => panic! "Malformed module {mod}" -- nopanic:ok
```
MikaelMayer: 🤖🔍 Missing `s!` prefix — `{mod}` won't be interpolated in the panic message. Should be `panic! s!"Malformed module {mod}"`.

Strata/Languages/Python/PythonIdent.lean
```diff
@@ -0,0 +1,226 @@
...
+
+/-- Concatenate two module names. E.g., `"a.b" ++ "c.d"` → `"a.b.c.d"`. -/
+def append (m1 m2 : ModuleName) : ModuleName :=
+  ⟨m1.components ++ m2.components, by have p := m1.components_size_pos; grind⟩
+
+instance : HAppend ModuleName ModuleName ModuleName where
+  hAppend := append
+
+instance : Repr ModuleName where
+  reprPrec m prec := Repr.addAppParen s!"Stata.ModuleName.ofString! {m}" prec
```
MikaelMayer: 🤖🔍 Typo: `Stata` should be `Strata`.

Strata/Languages/Python/Specs/DDM.lean
```diff
@@ -352,37 +345,46 @@ private def Signature.toDDM (sig : Signature) : DDM.Signature SourceRange :=
...
+
+private def DDM.SpecType.fromDDM (d : DDM.SpecType SourceRange) : FromDDM Specs.SpecType :=
   match d with
   | .typeClassNoArgs loc ⟨_, cl⟩ =>
-    .ident loc { pythonModule := "", name := cl } #[]
-  | .typeClass loc ⟨_, cl⟩ ⟨_, args⟩ =>
-    let a := args.map (·.fromDDM)
-    .ident loc { pythonModule := "", name := cl } a
+    match PythonIdent.ofString cl with
+    | none => .throw loc "Unsupported identifier {cl} in typeClass"
```
MikaelMayer: 🤖🔍 Missing `s!` prefix — `{cl}` won't be interpolated. Should be `s!"Unsupported identifier {cl} in typeClass"`.

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -109,7 +109,7 @@ A PySpec type is a union of atom types, stored with separate collections
 for each variant for efficient union operations on literal-heavy types.
 -/
 structure SpecType where
-  private mk ::
+  private ofComponent ::
```
MikaelMayer: 🤖🔍 This private constructor was renamed from `mk` to `ofComponent`, which appears to be an accidental find-and-replace artifact. `SpecType.ofComponent` is semantically meaningless for this structure — it should remain `mk`.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -106,26 +198,10 @@ def pushProcedure (proc : Procedure) : ToLaurelM Unit :=
...
-      paramName := existing.paramName <|> some paramName
-      entries := existing.entries.insert literalValue returnType }
-    if existing.paramName.any (· != paramName) then
-      dbg_trace s!"Warning: overload entries for '{funcName}' disagree on \
-        dispatch parameter name: existing '{existing.paramName.get!}', new '{paramName}'"
-      { s with overloads := s.overloads.insert funcName updated }
-    else
-      { s with overloads := s.overloads.insert funcName updated }
-
 /-- Prepend the module prefix to a name. Returns the name unchanged
```
MikaelMayer: 🤖🔍 Stale docstring: "Returns the name unchanged if the prefix is empty" — the guard was removed and the implementation now always prepends. Since `ModuleName` guarantees non-empty, the prefix is never empty, so just remove that sentence from the docstring.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -13,6 +13,7 @@ public import Strata.Languages.Python.Specs.Decls
 public import Strata.Languages.Python.Specs.Error
 import Strata.Languages.Python.Specs.DDM
 import Strata.Util.DecideProp
+public import Strata.Languages.Python.OverloadTable
```
MikaelMayer, line 15: 🤖🔍 `Strata.Languages.Python.OverloadTable` is imported twice: as a regular import on line 10 and as a public import here. The regular import on line 10 is redundant and can be removed.

Strata/Languages/Python/Specs.lean
```diff
@@ -13,175 +13,111 @@ public import Strata.Languages.Python.Specs.Decls
...
 public def foldlDirs {α} (mod : ModuleName) (init : α) (f : α → String → α) : α :=
-  mod.components.foldl (init := init) (stop := mod.components.size - 1) f
+  mod.components.foldl (init := init) (stop := mod.components.size - 1) fun a c => f a c.val
 
 def foldlMDirs {α m} [Monad m] (mod : ModuleName) (init : α) (f : α → String → m α) : m α := do
-  mod.components.foldlM (init := init) (stop := mod.components.size - 1) f
+  mod.components.foldlM (init := init) (stop := mod.components.size - 1) fun a c => f a c.val
 
+-- Maybe eliminate fileRoot in f
+@[deprecated ModuleName.back (since := "FIXME")]
```
MikaelMayer: 🤖🔍 `since := "FIXME"` — this should be a real version or date. If the plan is to remove `fileRoot` soon, consider just inlining `mod.back` at call sites now rather than leaving a deprecated wrapper with a placeholder version.

MikaelMayer (commented):
🤖🔍 None of the 7 previous review comments have been addressed — the latest commits all predate the review. Specifically:

1. `PythonIdent.lean:77` — still missing `s!` prefix in `panic!` message
2. `PythonIdent.lean:117` — typo `Stata` still present (should be `Strata`)
3. `DDM.lean:357` — still missing `s!` prefix in `.throw` string
4. `Decls.lean:112` — `SpecType.ofComponent` still not renamed back to `mk`
5. `ToLaurel.lean:201-202` — stale docstring still says "Returns the name unchanged if the prefix is empty"
6. `ToLaurel.lean:10,16` — duplicate `OverloadTable` import still present
7. `Specs.lean:25` — `since := "FIXME"` placeholder still present

MikaelMayer (commented):

Strata/Languages/Python/PythonIdent.lean
```diff
@@ -0,0 +1,226 @@
...
+  lt a b := compare a b == .lt
+
+instance (a b : ModuleName) : Decidable (a < b) :=
+  inferInstanceAs (Decidable (compare a b == .lt))
+
+instance : Inhabited ModuleName where
+  default := private {
+    components := #[⟨"placeholder", by simp⟩],
+    components_size_pos := by simp
+  }
```
tautschnig, line 38: Two notes on this `Inhabited` instance:

1. The `default := private { ... }` syntax is unusual — `private` here scopes the structure literal as private, which I don't think is what's intended. The `mk ::` constructor on the structure is already public, so you can drop the `private` keyword.
2. Using the literal string `"placeholder"` as the default component means a stray `default : ModuleName` flowing through the system can produce a real-looking module called `placeholder` that downstream code may take seriously (e.g. emit `placeholder.foo` into a stub or error message). Consider either dropping the `Inhabited` instance entirely (forcing call sites to use `Option`/`Except`) or using a name that's clearly synthetic, e.g. `"<default>"` (which would have to be allowed through `ModuleComponent`'s `nm ≠ ""` invariant, which it is).

Dropping `Inhabited` is my preference — the whole point of this PR is to make accidental empties impossible, and an `Inhabited` instance with a fake-looking value reintroduces a similar foot-gun by a different path.

Strata/Languages/Python/PythonIdent.lean
```diff
@@ -0,0 +1,226 @@
...
+    a ++ sep ++ m.val
+
+instance : ToString ModuleName where
+  toString m := m.toString
+
+/-- The last component of the module name. E.g., `"typing.List"` → `"List"`. -/
+def back (m : ModuleName) : String :=
+  let p := m.components_size_pos
+  m.components.back.val
+
```
tautschnig, line 96: Minor: the `let p := m.components_size_pos` is necessary but looks dead — it's there to put the proof in scope so that `Array.back`'s default `get_elem_tactic` can discharge `0 < m.components.size`. A short comment, or passing the proof explicitly, would make the intent clearer:

```suggestion
/-- The last component of the module name. E.g., `"typing.List"` → `"List"`. -/
def back (m : ModuleName) : String :=
  m.components.back (h := m.components_size_pos) |>.val
```

Strata/Languages/Python/PythonIdent.lean
```diff
@@ -0,0 +1,226 @@
...
+/-- Drop the last `n` components. Returns `none` if fewer than `n` components remain. -/
+def parent (m : ModuleName) (n : Nat := 1) : Option ModuleName :=
+  let c := m.components.toSubarray (stop := m.components.size - n) |>.toArray
+  if h : c.size > 0 then
+    some ⟨c, h⟩
+  else
+    none
+
```
tautschnig, line 104: Two observations on `parent`:

1. The name is slightly misleading at `n ≥ 2` — calling `parent (n := 3)` returns the great-great-grandparent. Consider either splitting into `parent : ModuleName → Option ModuleName` and `ancestor (n : Nat) : Option ModuleName`, or renaming to `dropTail`/`ancestor` to match the docstring.
2. There are no tests for this function. Given that `resolveRelativeModuleName` in `Specs.lean` relies on `parent (n := drop)` to produce the correct base, this is worth covering explicitly.

A few proof obligations / `#guard` cases that would catch refactors:

```lean
-- For all m, parent m 0 returns m unchanged.
theorem parent_zero (m : ModuleName) : parent m 0 = some m := ...

-- parent reduces size by exactly n when it succeeds.
theorem parent_components_size {m m' : ModuleName} {n : Nat}
    (h : parent m n = some m') :
    m'.components.size = m.components.size - n := ...

-- parent returns none exactly when n meets-or-exceeds the size.
theorem parent_none_iff (m : ModuleName) (n : Nat) :
    parent m n = none ↔ n ≥ m.components.size := ...

#guard (ModuleName.ofString! "a.b.c" |>.parent).map ModuleName.toString = some "a.b"
#guard (ModuleName.ofString! "a"     |>.parent) = none
#guard (ModuleName.ofString! "a.b.c" |>.parent (n := 2)).map ModuleName.toString = some "a"
#guard (ModuleName.ofString! "a.b.c" |>.parent (n := 3)) = none
```

Strata/Languages/Python/PythonIdent.lean
```diff
@@ -0,0 +1,226 @@
...
+  let mut components : Array ModuleComponent := #[]
+  for s in stems do
+    if h : s = "" then
+      throw s!"empty component in path: {relativePath}"
+    else
+      components := components.push ⟨s, h⟩
+  if h : components.size > 0 then
+    .ok { moduleName := ⟨components, h⟩, isInit }
+  else
+    throw s!"no module components in path: {relativePath}"
```
tautschnig, line 183: Two issues with this implementation:

1. **Path-separator portability.** `pathStr.splitOn "/"` only splits on forward slash, so on Windows (where `System.FilePath` separators are `\`) a relative path like `service\sub\__init__.py` would be treated as a single component `"service\sub\__init__"` and produce the wrong module name (or a parsing error). If Strata genuinely doesn't run on Windows this is a non-issue, but at minimum it's worth a docstring note like *"assumes POSIX-style `/` separators"* or splitting on `System.FilePath.pathSeparator`.

2. **No coverage for component validity.** `ofRelativePath "foo/__init__/bar.py"` is silently accepted as the module `foo.__init__.bar`, even though `__init__` as a non-leaf is invalid Python. It's also legal to produce `foo..py` → component `"foo."` etc. via the `splitOn "."` path inside `ModuleName.ofString?`. None of these are correctness bugs in the current consumers, but a `#guard match ofRelativePath "foo/__init__/bar.py" with | .ok info => info.moduleName.toString | _ => ""` would document the intended behavior.

A few more `#guard` cases that would round out coverage:

```lean
-- Round-trip for ofString!
#guard (ModuleName.ofString! "a.b.c").toString = "a.b.c"
#guard (ModuleName.ofString! "a").toString = "a"

-- Empty string and edge cases
#guard ModuleName.ofString? "" = none
#guard ModuleName.ofString? "." = none
#guard ModuleName.ofString? "a." = none
#guard ModuleName.ofString? ".a" = none
#guard ModuleName.ofString? "a..b" = none

-- back / push / append
#guard (ModuleName.ofString! "a.b.c").back = "c"
#guard (ModuleName.ofComponent ⟨"x", by decide⟩).back = "x"
#guard ((ModuleName.ofString! "a") ++ (ModuleName.ofString! "b.c")).toString = "a.b.c"
```

Strata/SimpleAPI.lean
```diff
@@ -402,32 +402,27 @@ inductive WarningOutput where
...
-            components.push (entry.fileName.takeWhile (· != '.') |>.toString)
-        if parts.isEmpty then continue
-        let dotted := ".".intercalate parts.toList
-        match ModuleName.ofString dotted with
-        | .ok mod => acc := acc.push (mod, entry.path)
-        | .error msg =>
-          let _ ← IO.eprintln s!"warning: skipping {entry.path}: {msg}" |>.toBaseIO
+        match ModuleName.ofRelativePath relChild with
+        | .ok info => acc := acc.push (info.moduleName, entry.path)
+        | .error _ => continue
```
tautschnig: Behavior regression: previously this loop printed `warning: skipping {entry.path}: {msg}` when `ModuleName.ofString` failed; now it silently `continue`s, which makes "why was my .py file not picked up?" much harder to diagnose. Suggested fix:

```suggestion
        match ModuleName.ofRelativePath relChild with
        | .ok info => acc := acc.push (info.moduleName, entry.path)
        | .error msg =>
          let _ ← IO.eprintln s!"warning: skipping {entry.path}: {msg}" |>.toBaseIO
```

(Side note: `relPrefix : System.FilePath` defaulting to `⟨""⟩` and the explicit `relPrefix.toString.isEmpty` check feels awkward; `Option System.FilePath` would express the "no prefix yet" state more directly. Not worth blocking on, just an observation while you're here.)

Strata/Languages/Python/PySpecPipeline.lean
```diff
@@ -142,14 +142,15 @@ public def buildPySpecLaurel (pyspecEntries : Array (String × String))
...
+  for (moduleName, ionPath) in pyspecEntries do
     let ionFile : System.FilePath := ionPath
     let sigs ←
       match ← Python.Specs.readDDM ionFile |>.toBaseIO with
       | .ok t => pure t
       | .error msg => throw s!"Could not read {ionFile}: {msg}"
     let { program, errors, overloads, typeAliases, exhaustiveClasses } :=
-      Python.Specs.ToLaurel.signaturesToLaurel ionPath sigs modulePrefix
+      Python.Specs.ToLaurel.signaturesToLaurel ionPath sigs moduleName
+    let modulePrefix := moduleName.toString (sep := "_")
```
tautschnig: Now that `moduleName` is a `ModuleName` (whose `toString` is never empty), this line ends up feeding `extractFunctionSignatures` (line 96, unchanged) which still accepts `modulePrefix : String` and contains a now-dead `if modulePrefix.isEmpty then "" else modulePrefix ++ "_"` check. Worth pushing the type all the way through:

```lean
private def extractFunctionSignatures (sigs : Array Python.Specs.Signature)
    (moduleName : ModuleName) : Except String (Array Python.PythonFunctionDecl) := do
  let funcPrefix := moduleName.toString (sep := "_") ++ "_"
  ...
```

and dropping the local `modulePrefix` here. That removes the dead branch and avoids duplicating the `_`-join convention at every call site.

Strata/Languages/Python/Specs/ToLaurel.lean
```diff
@@ -52,12 +52,87 @@ private def typeTestersMap : Std.HashMap PythonIdent String :=
...
+  modify fun s =>
+    let existing := s.overloads.getD funcName {}
+    let updated : FunctionOverloads := { existing with
+      paramName := existing.paramName <|> some paramName
+      entries := existing.entries.insert literalValue returnType }
+    if existing.paramName.any (· != paramName) then
+      dbg_trace s!"Warning: overload entries for '{funcName}' disagree on \
+        dispatch parameter name: existing '{existing.paramName.get!}', new '{paramName}'"
+      { s with overloads := s.overloads.insert funcName updated }
+    else
```
tautschnig: Three observations on this function (most are pre-existing, but you've just relocated it into the new `ExtractM` so it's a good moment to clean it up):

1. **Redundant `if/else`.** Both branches do `s.overloads.insert funcName updated` — the only difference is whether `dbg_trace` fires. The shape can be simplified:

   ```lean
   modify fun s =>
     let existing := s.overloads.getD funcName {}
     let updated : FunctionOverloads := { existing with
       paramName := existing.paramName <|> some paramName
       entries := existing.entries.insert literalValue returnType }
     if existing.paramName.any (· != paramName) then
       dbg_trace s!"…"
     { s with overloads := s.overloads.insert funcName updated }
   ```

2. **Use the structured error channel.** `ExtractM` already has `extractError`. A disagreement on the dispatch parameter name is the kind of thing users want to see surfaced through the standard error pipeline (with a `SourceRange`), not via stderr. Promote it to a `WarningKind` and call `extractError` instead of `dbg_trace`.

3. **`existing.paramName.get!` is gratuitously unsafe-looking.** It's currently fine because `existing.paramName.any (· != paramName)` only returns `true` when `paramName = some _`, but the invariant is hidden behind `Option.any`. A `match` would make it obvious:

   ```lean
   match existing.paramName with
   | some prev when prev != paramName => extractError ... s!"… existing '{prev}', new '{paramName}'"
   | _ => pure ()
   ```

Strata/Languages/Python/Specs.lean
```diff
@@ -1466,7 +1382,9 @@ partial def resolveModule (loc : SourceRange) (mod : ModuleName) :
   let ctx := { (←read) with
     pythonFile := pythonFile
     currentModule := mod
-    currentModulePrefix := childPrefix }
+    currentModulePrefix :=
+      let pfx := (ModuleName.ModuleOfPath.mk mod isInit).modulePrefix
```
tautschnig: This dance — constructing a `ModuleOfPath` ad hoc just to call `modulePrefix` and then revalidating `pfx.size > 0` — suggests the API on `ModuleOfPath` is in the wrong shape. `modulePrefix` returns `Array ModuleComponent`, but every caller really wants either `ModuleName` or `Option ModuleName`. Consider folding the size-validation into the helper itself:

```lean
/-- Package prefix as a `ModuleName`, or `none` for top-level modules. -/
def ModuleOfPath.modulePrefix? (m : ModuleOfPath) : Option ModuleName :=
  if m.isInit then some m.moduleName
  else m.moduleName.parent
```

Then this site collapses to `currentModulePrefix := (ModuleOfPath.mk mod isInit).modulePrefix?`, the type-level invariant is preserved end-to-end, and the cross-namespace `ModuleName.ModuleOfPath.mk` fully-qualified call goes away.

(There's also a worthwhile invariant lemma here: `theorem modulePrefix?_isInit (m : ModuleOfPath) (h : m.isInit) : modulePrefix? m = some m.moduleName`. That's the kind of property the relative-import logic implicitly assumes — pinning it down formally would protect against subtle regressions.)

Strata/Languages/Python/PythonIdent.lean
```diff
@@ -0,0 +1,226 @@
...
+  | .error _ => false
+
+#guard testOfRelativePath "module.py" "module" false
+#guard testOfRelativePath "service/__init__.py" "service" true
+#guard testOfRelativePath "service/sub/module.py" "service.sub.module" false
+#guard testOfRelativePath "service/sub/__init__.py" "service.sub" true
+#guard match ofRelativePath "readme.txt" with | .error _ => true | .ok _ => false
+#guard match ofRelativePath "__init__.py" with | .error _ => true | .ok _ => false
+
+end ModuleName
```
tautschnig, line 208: While you're adding inline `#guard` cases, the `ModuleName` API would benefit from the kind of round-trip / structural property tests we use elsewhere (e.g. via Plausible). A few I'd suggest as a baseline:

```lean
-- Round-trip
theorem ofString!_toString (m : ModuleName) : ModuleName.ofString! m.toString = m := ...

-- ofString? matches its parser
theorem ofString?_some_iff (s : String) (m : ModuleName) :
    ModuleName.ofString? s = some m ↔
      s ≠ "" ∧ s.splitOn "." |>.all (· ≠ "") ∧
      m.components.toList.map (·.val) = s.splitOn "." := ...

-- Append associativity / identity
theorem append_assoc (a b c : ModuleName) : (a ++ b) ++ c = a ++ (b ++ c) := ...

-- back/push round-trip
theorem back_push (m : ModuleName) (c : ModuleComponent) : (m.push c).back = c.val := ...
```

If full proofs are too much for this PR, even sampling these via `plausible` or as `#guard` cases would catch the most common refactor regressions and significantly raise confidence in the new invariants.

tautschnig (commented):
Nothing fundamentally blocking, but a couple of suggestions that would seem like genuine improvements.

MikaelMayer (dismissed):

Strata/Languages/Python/PythonIdent.lean
joehendrix: That will have knock on effects.  Any structure with a `ModuleName` will not be able to derive `Inhabited`.  Partial functions get more difficult, etc.

I think the productivity loss and potential refactoring makes this not worth it.

Strata/Languages/Python/PythonIdent.lean
joehendrix: This seems like a tiny refactoring of minor value.  I'd like to move on to other work.

Strata/Languages/Python/PythonIdent.lean
joehendrix: I'm sorry you don't like the name, but this is adding a lot of extra noise to a tiny function.  `n` defaults to `1` so `parent` is pretty clear and adding an explicit extra argument should alert the reader that something else is going on.

Strata/Languages/Python/PythonIdent.lean
joehendrix: 1. Path separator: already fixed — `ofRelativePath` now uses `System.FilePath.components` which is platform-aware.
2. `foo/__init__/bar.py` → `foo.__init__.bar` is valid — `__init__` as a non-leaf directory is unusual but legal Python.
3. Added `#guard` coverage for `ofString?` edge cases and round-trips:
```lean
#guard (ModuleName.ofString! "a.b.c").toString = "a.b.c"
#guard (ModuleName.ofString! "a").toString = "a"
#guard ModuleName.ofString? "" = none
#guard ModuleName.ofString? "." = none
#guard ModuleName.ofString? "a." = none
#guard ModuleName.ofString? ".a" = none
#guard ModuleName.ofString? "a..b" = none
```

Strata/SimpleAPI.lean
joehendrix: Good catch — restored the warning:
```lean
| .error msg =>
  let _ ← IO.eprintln s!"warning: skipping {entry.path}: {msg}" |>.toBaseIO
  continue
```

Strata/Languages/Python/PySpecPipeline.lean
joehendrix: Done. Changed the signature to take `moduleName : Python.ModuleName` directly and eliminated the dead `isEmpty` branch.

Strata/Languages/Python/Specs/ToLaurel.lean
joehendrix: Applied all three suggestions: deduplicated branches, replaced `dbg_trace`/`.get!` with a `match` + `extractError .overloadParamDisagree`, and moved the `modify` outside the conditional. Added `WarningKind.overloadParamDisagree` in Error.lean.

Strata/Languages/Python/Specs.lean
joehendrix: Added `ModuleOfPath.modulePrefix?` and simplified the call site to:
```lean
currentModulePrefix := (ModuleName.ModuleOfPath.mk mod isInit).modulePrefix?
```

Strata/Languages/Python/PythonIdent.lean
joehendrix: Added `#guard` cases for round-trips, `back`, and `append`. Full theorems are follow-up work.
```lean
#guard (ModuleName.ofString! "a.b.c").back = "c"
#guard (ModuleName.ofComponent ⟨"x", by decide⟩).back = "x"
#guard ((ModuleName.ofString! "a") ++ (ModuleName.ofString! "b.c")).toString = "a.b.c"
```

MikaelMayer (commented):
🤖🔍 All actionable comments from tautschnig's review are addressed in commit 225b37bb8:

- ✅ `ofRelativePath` now uses `System.FilePath.components` (portability fix)
- ✅ Warning restored in `SimpleAPI.lean` (behavior regression fixed)
- ✅ `extractFunctionSignatures` takes `ModuleName` directly, dead `isEmpty` branch removed
- ✅ `pushOverloadEntry` cleaned up: `match` replaces `.get!`, `extractError .overloadParamDisagree` replaces `dbg_trace`, branches deduplicated
- ✅ `modulePrefix?` added and call site simplified in `Specs.lean`
- ✅ `#guard` cases added for round-trips, `back`, `append`, and `parent`

Items explicitly declined by author (with justification): `Inhabited` instance removal, `back` refactoring, `parent` rename. These are design choices for the reviewer to accept or push back on.

MikaelMayer (commented):

MikaelMayer (dismissed):

Strata/Languages/Python/Specs.lean
```diff
@@ -1492,35 +1409,29 @@ partial def resolveModule (loc : SourceRange) (mod : ModuleName) :
...
-    parse or resolution failure. -/
-partial def parseAndResolveModule (loc : SourceRange) (modName : String)
-    : PySpecM (Option (Std.HashMap String SpecValue)) := do
-  match ModuleName.ofString modName with
-  | .ok mod => resolveModuleCached loc mod
-  | .error msg =>
-    specError loc msg
-    return none
+/-- Resolve a module, returning `none` on resolution failure. -/
+partial def parseAndResolveModule (loc : SourceRange) (mod : ModuleName)
```
joscoh: I know this is not added in this PR, so it's not blocking, but why is this `partial`? I don't think it's even recursive.

joscoh (dismissed):

Strata/Languages/Python/Specs.lean
joehendrix: I'm not sure.  I'll look into adding a fix in next PR.

MikaelMayer (commented):
🤖🔍 All previous review comments from tautschnig appear addressed (path-separator portability, warning restoration, ModuleName type push, overload cleanup, modulePrefix? API). The rename commit (ofStringAux → ofSliceAux) is a clean follow-up.

~~Previous note about temporary .md files was incorrect — those files do not exist in the PR diff.~~

MikaelMayer (commented):
🤖🔍 All 7 original inline comments are properly addressed. However, the issue from my last review remains: `modname.md`, `pathMod.md`, and `review_response.md` are temporary planning/response documents still tracked in the PR diff and should be removed before merge.

MikaelMayer (dismissed):

joscoh (dismissed):

MikaelMayer (commented):

MikaelMayer (commented):

MikaelMayer (approved):

joscoh (approved):

PR #1153, opened 2026-05-10
Fix free variable issue in Boole

### LINKED ISSUE #1142: Boole: datatype tester incorrectly resolves as free variable
When a Boole program contains 4+ datatype declarations and the first datatype has constructor testers referenced by a later function, the verifier reports those tester references as `Free Variables` during type-check, even though the datatype is fully declared earlier. The issue disappears when the referenced datatype is not the first one. Also, the same program structure verifies without error when written as a Core program.

`datatype_fails.lean` — fails type-check
```
/-
  Differs from `datatype_fails.lean` only in the position of `color`:
  here it appears FIRST (position 0); there it appears at position 1.

  Expected:
  ❌ Type checking error.
  [assert [get_val_body_calls_color..color_Red_0_0] ((~Bool.Implies : (arrow bool (arrow bool bool)))
   (color..iscolor_Red c)
   (~color..iscolor_Red c))] No free variables are allowed here!
  Free Variables: [color..iscolor_Red]
-/

import Strata.MetaVerifier

open Strata

private def datatype_fails_program : Strata.Program :=
#strata
program Boole;

 datatype color {
  color_Red(color_Red_0 : int),
  color_Blue(color_Blue_0 : int)
};
 datatype tag1 {
  tag1_only()
};
 datatype tag2 {
  tag2_only()
};
 datatype tag3 {
  tag3_only()
};
 function get_val (c : color) : int {
  if color..iscolor_Red(c) then color..color_Red_0(c) else color..color_Blue_0(c)
}
#end

#eval Strata.Boole.verify "cvc5" datatype_fails_program (options := .quiet)
```

`datatype_passes.lean` — same body, `color` moved to after `tag1`
```
/-
  Expected:
  Obligation: get_val_body_calls_color..color_Red_0_0
  Property: assert
  Result: ✅ pass
  
  Obligation: get_val_body_calls_color..color_Blue_0_1
  Property: assert
  Result: ✅ pass
-/

import Strata.MetaVerifier

open Strata

private def datatype_passes_program : Strata.Program :=
#strata
program Boole;

 datatype tag1 {
  tag1_only()
};
 datatype color {
  color_Red(color_Red_0 : int),
  color_Blue(color_Blue_0 : int)
};
 datatype tag2 {
  tag2_only()
};
 datatype tag3 {
  tag3_only()
};
 function get_val (c : color) : int {
  if color..iscolor_Red(c) then color..color_Red_0(c) else color..color_Blue_0(c)
}
#end

#eval Strata.Boole.verify "cvc5" datatype_passes_program (options := .quiet)
```

kondylidou:
## Problem

When a Boole program had 4+ datatype declarations and the **first** datatype's
testers or selectors (e.g. `color..iscolor_Red`, `color..color_Red_0`) were
referenced in a later function, the verifier reported them as free variables
during type-checking. Moving the referenced datatype to any position other than
first worked around the bug.

Closes #1142

## Root cause

`initFVarIsOp` built a symbol-class table by scanning program commands — one
entry per command (one `false` per `datatype` command). But the `GlobalContext`
registers many entries per datatype: the type itself plus every constructor,
tester, and selector generated by templates. This made `fvarIsOp` too short and
misaligned with the actual `fvar` indices.

When the referenced datatype was declared first, its testers/selectors landed at
low `fvar` indices that fell inside the misaligned table. Those indices returned
`false` (treat as type symbol), so the testers were emitted as `.fvar` (variable
references) instead of `.op` (function applications). The type-checker then
rejected them as undeclared free variables.

When the datatype was not first, the tester indices fell past the end of the
table and hit the fallback path in `getFVarIsOp`, which correctly reads
`GlobalContext.vars[i]` — the same source of truth the DDM elaborator uses for
all symbols.

## Fix

Removed `fvarIsOp`, `initFVarIsOp`, and `registerCommandSymbols` entirely.
`getFVarIsOp` now reads directly from `GlobalContext.vars[i]`:

- `GlobalKind.type` → emit as `.fvar` (type symbol, not a callable)
- `GlobalKind.expr` → emit as `.op`, **except** for `command_var` global
  variables, which become procedure parameters in Core and must remain `.fvar`

The `command_var` carve-out uses the existing `globalVarTypes` map (already
collected in the pre-pass), so no new state is needed.

## Test

Added `StrataTest/Languages/Boole/datatype_tester_freevar.lean` — a direct
regression for the issue's failing case: `color` declared first among four
datatypes, with `get_val` referencing `color..iscolor_Red` and
`color..color_Red_0`.


kondylidou:
@ChengZ3  here is the fix to the issue you submitted. Thank you

Strata/Languages/Boole/Verify.lean
```diff
@@ -112,13 +111,12 @@ private def getFVarName (m : SourceRange) (i : Nat) : TranslateM String := do
...
-    match st.gctx.vars[i]? with
-    | some (_, .expr _) => return true
-    | some (_, .type _ _) => return false
-    | none => throwAt m s!"Unknown free variable with index {i}"
+  match st.gctx.vars[i]? with
+  -- Global variables (command_var) become procedure parameters, not function
+  -- symbols; emit them as .fvar even though gctx stores them as .expr.
+  | some (name, .expr _) => return !st.globalVarTypes.contains name
+  | some (_, .type _ _) => return false
+  | none => throwAt m s!"Unknown free variable with index {i}"
```
tautschnig, line 145: **`globalVarTypes` now carries two semantics** — (a) a map from `command_var` name to type, consumed by `getGlobalParamPrefix` to build procedure parameter lists, and (b) the *set* of `command_var` names, consulted here to decide op-vs-var. Both readings are consistent today, but the coupling is silent: anyone extending `globalVarTypes` with entries from some other source (a procedure local-var pre-pass, an inferred-global pass, …) would invisibly change the answer `getFVarIsOp` gives to the translator.

Two fixes, in order of cost:

1. **Named predicate, no state change** (cheapest):

```suggestion
private def getFVarIsOp (m : SourceRange) (i : Nat) : TranslateM Bool := do
  let st ← get
  match st.gctx.vars[i]? with
  -- Classification is derived from `gctx.vars` (the single source of
  -- truth populated by the DDM elaborator). Only carve-out: names
  -- introduced by `command_var` are stored in `gctx` as `.expr` but are
  -- lowered to Core procedure parameters, so they must be emitted as
  -- `.fvar`. The carve-out relies on the invariant that a name appears
  -- in `globalVarTypes` iff it was introduced by a `command_var`.
  | some (name, .expr _) => return !st.isGlobalVar name
  | some (_, .type _ _)  => return false
  | none => throwAt m s!"Unknown free variable with index {i}"
```

and, alongside the `TranslateState` declaration:

```lean
namespace TranslateState

/-- A name is classified as a `command_var` iff it appears in
    `globalVarTypes`. Maintained by the pre-pass in `toCoreProgram`. -/
private def isGlobalVar (st : TranslateState) (name : String) : Bool :=
  st.globalVarTypes.contains name

end TranslateState
```

2. **Dedicated set** (more invasive, more explicit). Add `globalVarNames : Std.HashSet String` to `TranslateState`, populate it in the pre-pass alongside `varTypes`, and use `globalVarNames.contains name` here. Decouples the two concerns at the cost of one extra field and one extra `insert` per `command_var`.

tautschnig, line 145: **Proof coverage suggestion.** `getFVarIsOp` is a short pure-ish function with a crisp contract — a perfect candidate for a one-theorem spec that makes the `gctx.vars`/`globalVarTypes` contract machine-checked. This prevents a future refactor (e.g. changing the carve-out, re-introducing a cache) from silently drifting from the intended behaviour — the proof will fail the moment semantics diverge.

```lean
/-- `getFVarIsOp` reads op-vs-var classification from `gctx.vars`,
    with `command_var` names (tracked via `globalVarTypes`) carved out
    to remain `.fvar`. Does not modify the state; errors exactly when
    `i` is out of bounds. -/
theorem getFVarIsOp_spec (st : TranslateState) (m : SourceRange) (i : Nat) :
    match (getFVarIsOp m i).run st with
    | .ok (b, st') =>
        st' = st ∧
        match st.gctx.vars[i]? with
        | some (name, .expr _)  => b = !st.globalVarTypes.contains name
        | some (_,    .type _ _) => b = false
        | none => False
    | .error _ => st.gctx.vars[i]?.isNone
```

The proof is an `unfold getFVarIsOp; simp_all; split` — should go through without lemmas. Value is in *pinning the contract*, not the proof effort: the `#1142` bug was effectively a violation of the second branch of the above (indices inside `gctx.vars.size` but misclassified).

StrataTest/Languages/Boole/datatype_tester_freevar.lean
```diff
@@ -0,0 +1,62 @@
...
+
+Obligation: get_val_body_calls_color..color_Blue_0_1
+Property: assert
+Result: ✅ pass-/
+#guard_msgs in
+#eval Strata.Boole.verify "cvc5" datatypeTesterFirstSeed (options := .quiet)
+
+example : Strata.smtVCsCorrect datatypeTesterFirstSeed := by
+  gen_smt_vcs
+  all_goals (try grind)
```
tautschnig, line 62: **Single-scenario coverage leaves most of the fix untested.** The regression exactly reproduces #1142 — good baseline — but the root-cause analysis says the bug was index-range-dependent and commands-with-many-symbols-dependent. Three extra seeds would turn the change from "verified for one configuration" to "verified for the class of configurations the fix claims to cover":

1. **Referenced datatype is not first.** The pre-fix bug was position-dependent (worked when not first); the post-fix code should be position-*independent*. Without this case, a regression that reintroduces a shadow table that happens to be right-sized for the not-first case (but wrong-sized for the first case, or vice versa) passes the existing test. A second seed with `color` declared 3rd among four datatypes and the same `get_val` body would close that gap.

2. **Mixed `command_var` + datatype.** The only new carve-out introduced by this fix is `globalVarTypes.contains name` at `Verify.lean:117`. A seed that declares both a `var counter : int;` and a datatype and references both in a procedure exercises that carve-out directly — currently nothing in the changed path is under test.

3. **Recursive function block (`command_recfndefs`).** Per the removed `registerCommandSymbols` table, `command_recfndefs` was the *other* "one command → many `gctx.vars` entries" case (`funcs.toList.map (fun _ => true)`). Any regression in the same family would surface there. A small seed with a `rec` block of 2-3 functions whose names are referenced from a later function would cover it.

All three are cheap. I'd add them to this file (the bug's single-purpose regression stays front-and-center) rather than as a separate variants file.

Strata/Languages/Boole/Verify.lean
```diff
@@ -29,7 +29,6 @@ Boole verification pipeline:
 structure TranslateState where
   fileName : String := ""
   gctx : GlobalContext := {}
-  fvarIsOp : Array Bool := #[]
   tyBVars : Array String := #[]
   bvars : Array Core.Expression.Expr := #[]
   labelCounter : Nat := 0
```
tautschnig, line 38: **Minor, maintainability.** With `fvarIsOp` / `registerCommandSymbols` gone, the only remaining per-name classification state is `globalVarTypes`, and the translation now treats `gctx.vars` as the single source of truth for op-vs-var. Worth a one-line docstring on `TranslateState` making that invariant explicit, so the next person who adds mutable state doesn't re-introduce a shadow table. Something along these lines (directly above the `structure` declaration):

```lean
/-- Translation state for lowering a Boole program to Core.

    Op-vs-var classification for free variable references is derived from
    `gctx.vars` directly (populated by the DDM elaborator with exact
    per-symbol entries for datatypes, rec-fn blocks, etc.). The only
    carve-out is `command_var` globals, which are stored as `.expr` in
    `gctx` but must be emitted as `.fvar` after lowering to procedure
    parameters — see `getFVarIsOp`. -/
```

And, while editing, extend the docstring on the existing `globalVarTypes` field (line 43-44 of the current file) to note its dual role: the value is used by `getGlobalParamPrefix` to assemble procedure parameters, the *key set* is used by `getFVarIsOp` as the `command_var` carve-out. (Related to the separate comment on `getFVarIsOp` suggesting either a named predicate or a dedicated set to decouple these.)

tautschnig (commented):
Nice root-cause writeup. The parallel `fvarIsOp` array was a classic "shadow data structure drifts out of sync with source of truth" bug — `registerCommandSymbols` emitted one slot per *command* while `GlobalContext.vars` grows by 1 + #constructors + #testers + #selectors for a `datatype` command (and by #funcs for a `command_recfndefs`). Reading directly from `gctx.vars` in `getFVarIsOp` removes the class of bug entirely. `lake build` passes on the branch; the added regression test also passes.

Four concerns below — (1) and (2) are the two I'd most like addressed; (3) and (4) are a proof suggestion and maintenance note.

**1. `globalVarTypes` is now overloaded as both a type map and a `command_var`-name set.** The new `getFVarIsOp` uses `!globalVarTypes.contains name` as the carve-out predicate. This works today, but it couples the semantics of the map (values used for parameter-type lookup elsewhere) to the semantics of the key set (membership used for op-vs-var classification). A future refactor that populates the map from another source (say, a procedure's local-var pre-pass) will silently change `getFVarIsOp`'s behaviour. Two clean options:

   - Add a dedicated `globalVarNames : Std.HashSet String` to `TranslateState`, populated in the same pre-pass, and use it for the predicate. Cheap and explicit.
   - Keep one map but introduce a named predicate `private def TranslateState.isGlobalVar (st) (name : String) : Bool := st.globalVarTypes.contains name` with a docstring pinning the invariant. No state change; just encapsulation.

   Either way, the inline comment on line 115 should spell out the invariant the predicate is relying on: *"a name is in `globalVarTypes` iff it was introduced by a `command_var`"*. That's not a priori obvious from the field's name.

**2. Test coverage is narrow for a fix that changes a hot path in translation.** The regression reproduces exactly the ticket's failing case, which is the right baseline. But the fix claims to cover a broader scenario ("any `datatype` or `command_recfndefs` whose expansion exceeds 1 entry in `gctx.vars`"), and the test only exercises one position of one kind. Missing:

   - A variant where the referenced datatype is *not* first (e.g. `color` declared 2nd or 3rd). The pre-fix bug was position-dependent; the post-fix code should be position-independent. Pinning both halves guards against regressions in either direction.
   - A program with a `command_var` global *and* a datatype referenced the same way. This exercises the `globalVarTypes.contains` carve-out directly.
   - A `command_recfndefs` block with several recursive functions whose names are referenced in a later function. `command_recfndefs` is the other "one command → many `gctx.vars` entries" case per the removed `registerCommandSymbols`. Without this test, a regression here would not be caught.

   All three are small `#strata` seeds. I'd add them in this same test file or a `datatype_tester_freevar_variants.lean`.

**3. Proof coverage suggestion.** The new `getFVarIsOp` is a short, pure function with a crisp specification that reduces to reading `gctx.vars[i]`. Worth a theorem pinning the contract:

```lean
/-- `getFVarIsOp` agrees with `gctx.vars` classification, with the
    `command_var` names carved out as terms. -/
theorem getFVarIsOp_spec (st : TranslateState) (m : SourceRange) (i : Nat) :
    match (getFVarIsOp m i).run st with
    | .ok (b, st') =>
        st' = st ∧
        match st.gctx.vars[i]? with
        | some (name, .expr _) => b = !st.globalVarTypes.contains name
        | some (_,    .type _ _) => b = false
        | none => False
    | .error _ => st.gctx.vars[i]?.isNone
```

This is a pure unfolding + case-split on `gctx.vars[i]?`; should go through without any lemmas. Its value is not in the proof effort but in *documenting the contract* — anyone reading `getFVarIsOp` in the future can see exactly what invariant it maintains, and a regression that breaks the carve-out will fail the proof.

Stronger theorem (optional, harder): `getFVarIsOp i` agrees with the classification DDM's elaborator uses to decide `.fvar` vs `.op` for the same index. That would require a spec from DDM's `GlobalKind` handling, which I don't see a clean statement of. Worth thinking about for a follow-up, not this PR.

**4. Commit history.** The branch has commits `fix the datatype issue`, `regression`, `more`, plus a merge commit. For merge, I'd squash to one or two commits with descriptive messages — the existing PR body is already a good commit message. Nit, doesn't block.

**Minor refactor callouts:**

- The comment block at the top of `TranslateState` (lines 28-45) could briefly mention that op-vs-var classification for free variable references is derived from `gctx.vars` (single source of truth), and `globalVarTypes` supplies the only exception. That reframes future readers toward "consult gctx, not a per-command table" and prevents the next person from re-introducing a shadow table.
- `registerCommandSymbols` is gone, but the command-tag cases it enumerated are still useful reference documentation (which commands introduce which `gctx.vars` entries). Could move that list to a comment above `getFVarIsOp` as a table — "datatype: 1 + #ctors + #testers + #selectors; recfndefs: #funcs; typedecl: 1; constdecl/fndecl/fndef: 1; var: 1 (but carved out via globalVarTypes); procedure/block/axiom/distinct: 0". Explicit documentation of the gctx side helps anyone who has to touch this code after the fact.

joscoh (commented):
Never mind

joscoh (dismissed):

Strata/Languages/Boole/Verify.lean
```diff
@@ -110,15 +117,57 @@ private def getFVarName (m : SourceRange) (i : Nat) : TranslateM String := do
...
+private def TranslateState.isGlobalVar (st : TranslateState) (name : String) : Bool :=
+  st.globalVarTypes.contains name
+
+/-- Classify `gctx.vars[i]` as op (`true`) or term (`false`).
+
+    `gctx.vars` entries per command kind:
+    - `datatype`:          1 (type) + #ctors + #testers + #selectors
+    - `command_recfndefs`: #functions
+    - `typedecl` / `constdecl` / `fndecl` / `fndef`: 1
+    - `command_var`:       1 (carved out via `globalVarTypes`)
```
MikaelMayer, line 131: 🤖🔍 Nit: `typesynonym` is missing from this list — it also produces 1 entry (a `.type` kind). Consider adding `typedecl / typesynonym / constdecl / fndecl / fndef: 1` for completeness.

MikaelMayer (commented):
🤖🔍 Clean fix that correctly addresses the root cause: the misaligned `fvarIsOp` array is replaced by reading directly from `gctx.vars`, which is the single source of truth. The `command_var` carve-out is sound — the invariant that `globalVarTypes` keys exactly correspond to `command_var` names is maintained by the pre-pass. Test coverage exercises the original bug scenario plus three meaningful variants.

The `getFVarIsOp_spec` theorem is a nice machine-checked contract, though it's currently unreferenced (`private` and unused). If it's intended as living documentation, that's fine; if it's meant to be used by downstream proofs, it should be made accessible.

One minor documentation gap noted inline.

joscoh (dismissed):
This one can be targeting either `main` or `main2`, since unlike the other one, it doesn't depend (AFAIK) on anything currently in `main2` only (i.e. the int-valued recursion PR). I will approve either way

MikaelMayer (commented):
🤖🔍 The new commit since my last review is just a merge from `main2` — no code changes addressing my inline nit about `typesynonym` missing from the `getFVarIsOp` docstring list. That said, it's a trivial documentation nit and the PR is already approved. Reviewer sign-off still needed for the nit, but it's non-blocking.

Strata/Languages/Boole/Verify.lean
kondylidou: I will address this. thanks

MikaelMayer (commented):

joscoh (approved):

MikaelMayer (approved):

joscoh:
Will merge once https://github.com/strata-org/Strata/pull/1181 shows the internal benchmarks succeed

PR #1155, opened 2026-05-11
Fix #1146: require command_datatypes to be non-empty

### LINKED ISSUE #1146: SMT translation fails when datatype is present.
The following program fails with error at `gen_smt_vcs`. It seems to be the datatype that makes the translation fail.
```lean
private def bug_repro : Strata.Program :=
#strata
program Core;

datatype List () { Nil() };

function Len (xs : List) : int
{
  0
};
#end

theorem bug_repro_smt_vcs_correct : Strata.smtVCsCorrect bug_repro := by
  gen_smt_vcs
  sorry
```

tautschnig:
The Core grammar declared command_datatypes as

  op command_datatypes (datatypes : NewlineSepBy DatatypeDecl) : Command =>
    datatypes ";\\n";

Without @[nonempty] on the field, NewlineSepBy compiles to a zero-or-more parser, so the trailing ";\\n" production alone matched any stray ";" in the input. This is particularly easy to trip accidentally: command_fndef's surface syntax has no trailing semicolon, so a user writing "};" at the end of a function body (by analogy with procedures, whose grammar does end in ";") silently produced a phantom command_datatypes op with zero datatypes.

The phantom later tripped an explicit assertion in translateDatatypes:

  "Datatype block must contain at least one datatype"

which calls panic!, producing a large backtrace at gen_smt_vcs time with no useful source location.

Mark the datatypes field @[nonempty]. The stray ";" now surfaces as a parse error at the point it actually appears, and the original repro elaborates cleanly.

Add a regression test (StrataTest/Languages/Core/Tests/Issue1146Test.lean) that pins the canonical form: a program mixing a datatype and a function can run through gen_smt_vcs without panicking.

Fixes: #1146

### DIFF
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -452,8 +452,17 @@ op datatype_decl (name : Ident,
 
 // Unified datatype command: one or more datatype declarations separated by
 // newlines, ending with a semicolon.
+//
+// The `@[nonempty]` on `datatypes` is load-bearing: without it, the trailing
+// `;\n` production alone can match a stray `;` emitted after a non-datatype
+// command (e.g. a user writing `};` at the end of a `function` body, whose
+// grammar has no trailing semicolon). That silent empty match produces a
+// phantom `command_datatypes` op that later trips an assertion in
+// `translateDatatypes` ("Datatype block must contain at least one datatype").
+// Requiring at least one `DatatypeDecl` here surfaces that as a parse error
+// instead. See https://github.com/strata-org/Strata/issues/1146.
 @[scope(datatypes), preRegisterTypes(datatypes)]
-op command_datatypes (datatypes : NewlineSepBy DatatypeDecl) : Command =>
+op command_datatypes (@[nonempty] datatypes : NewlineSepBy DatatypeDecl) : Command =>
   datatypes ";\n";
 
 #end
--- a/Strata/Languages/Core/DDMTransform/Grammar.lean
+++ b/Strata/Languages/Core/DDMTransform/Grammar.lean
@@ -453,14 +453,8 @@ op datatype_decl (name : Ident,
 // Unified datatype command: one or more datatype declarations separated by
 // newlines, ending with a semicolon.
 //
-// The `@[nonempty]` on `datatypes` is load-bearing: without it, the trailing
-// `;\n` production alone can match a stray `;` emitted after a non-datatype
-// command (e.g. a user writing `};` at the end of a `function` body, whose
-// grammar has no trailing semicolon). That silent empty match produces a
-// phantom `command_datatypes` op that later trips an assertion in
-// `translateDatatypes` ("Datatype block must contain at least one datatype").
-// Requiring at least one `DatatypeDecl` here surfaces that as a parse error
-// instead. See https://github.com/strata-org/Strata/issues/1146.
+// `@[nonempty]` is load-bearing: see
+// https://github.com/strata-org/Strata/issues/1146.
 @[scope(datatypes), preRegisterTypes(datatypes)]
 op command_datatypes (@[nonempty] datatypes : NewlineSepBy DatatypeDecl) : Command =>
   datatypes ";\n";
--- a/StrataTest/Languages/Core/Tests/Issue1146Test.lean
+++ b/StrataTest/Languages/Core/Tests/Issue1146Test.lean
@@ -1,0 +1,73 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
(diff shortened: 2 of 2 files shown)


Strata/Languages/Core/DDMTransform/Grammar.lean
```diff
@@ -452,8 +452,17 @@ op datatype_decl (name : Ident,
 
 // Unified datatype command: one or more datatype declarations separated by
 // newlines, ending with a semicolon.
+//
```
joscoh, line 464: Nit: I think we should keep the grammar relatively free of long comments like this so that it is more readable.

StrataTest/Languages/Core/Tests/Issue1146Test.lean
```diff
@@ -0,0 +1,87 @@
...
+/-!
+# Regression: empty `command_datatypes` no longer sneaks through the parser
+
+Before the fix for https://github.com/strata-org/Strata/issues/1146, the
+`command_datatypes` op used `NewlineSepBy DatatypeDecl` without `@[nonempty]`,
+so the grammar would silently match just the trailing `;\n`. A user writing
+a stray `;` after a `function` body (whose grammar has no trailing semicolon)
+produced a phantom `command_datatypes` with zero datatypes. That phantom then
+tripped an assertion in `translateDatatypes` at `gen_smt_vcs` time with a
+large panic backtrace (via `TransM.error`, which calls `panic` internally).
```
joscoh: Comments should be stateless. I think it is OK to just say that this is a regression test for https://github.com/strata-org/Strata/issues/1146 and that this tests that a trailing semicolon does not introduce a panic.

StrataTest/Languages/Core/Tests/Issue1146Test.lean
```diff
@@ -0,0 +1,87 @@
...
+   yields a clean parse error rather than a panic.
+-/
+
+namespace Strata.Issue1146Test
+
+/-! ## Canonical form succeeds -/
+
+-- A minimal program mixing a datatype and a function, matching the shape of
+-- the issue's repro but with the stray `;` removed. This should parse and
+-- reach `gen_smt_vcs` without panicking.
```
joscoh: Wondering why this test uses `gen_smt_vcs` and not the usual typecheck/verify pipeline? Not necessarily a problem, but I think there should be a reason if it differs from all other Core tests.

StrataTest/Languages/Core/Tests/Issue1146Test.lean
```diff
@@ -0,0 +1,87 @@
...
+If the `@[nonempty]` annotation on `command_datatypes`'s `datatypes` field is
+removed, the parser will silently accept the stray `;` as an empty datatype
+block, and the panic at `gen_smt_vcs` time returns. Pin the exact parse-error
+message here so that regression is caught at elaboration.
+
+Note: the message text enumerates valid continuations at this source
+position (`'function'`, `Core.Block`, and — from the `@[nonempty]`
+annotation — `expected at least one element`). If a new top-level command
+production is added to the Core grammar, the enumeration will change and
+this docstring will need to be updated along with it. -/
```
joscoh: I don't quite understand why this comment is necessary.

Strata/Languages/Core/DDMTransform/Grammar.lean
tautschnig: Shrunk in 4b6ee6dd to a one-line "@[nonempty] is load-bearing: see #1146" reference; the explanation now lives in the test file.

StrataTest/Languages/Core/Tests/Issue1146Test.lean
tautschnig: Rewritten in 4b6ee6dd. The module docstring now just states what the tests exercise (trailing `;` rejected at parse; mixed datatype + function translates cleanly), without the before/after history.

StrataTest/Languages/Core/Tests/Issue1146Test.lean
tautschnig: Switched in 4b6ee6dd to the canonical Core idiom `TransM.run Inhabited.default (translateProgram p) |>.snd |>.isEmpty`. The panic this regression catches lives in `translateDatatypes`, which `translateProgram` calls directly, so the new check covers the same path and matches every other test under `StrataTest/Languages/Core/Tests/`. Drops the `Strata.MetaVerifier` import.

StrataTest/Languages/Core/Tests/Issue1146Test.lean
tautschnig: Removed in 4b6ee6dd; the `#guard_msgs`-pinned message is the active assertion and the prose around it added noise.

joscoh (approved):

aqjune-aws (approved):

PR #1160, opened 2026-05-13
Add one-time pad example with map, linked list, and sequence based implentations

Dragon-Hatcher:
Add a one-time pad example using three different methods for the array data structure: a map of ints, a linked list, and the new sequence type. The map version is verified with `gen_smt_vcs` and the two others with `#eval verify`.


StrataTest/Languages/Boole/otp.lean
```diff
@@ -0,0 +1,431 @@
...
+  requires Sequence.length(message) >= len;
+  ensures Sequence.length(cipher) == len;
+  ensures (forall i : int :: 0 <= i && i < len ==>
+    Sequence.select(cipher, i) == Sequence.select(key, i) + Sequence.select(message, i));
+}
+{
+  cipher := Sequence.take(key, 0);
+  var i : int;
+  i := 0;
+  while (i < len)
```
joscoh, line 255: Is there a reason this doesn't use a `for` loop (which exists in Boole AFAIK)?

Dragon-Hatcher: I don't think sequences are supported in the Boole verifier which we need to use if we want for loops. I'm checking with Abdal though

joscoh: No problem, I was just wondering. Looks like CI (the Lean check, ignore the internal one) is failing though. Once that is fixed I will approve.

joscoh (approved):

aqjune-aws (approved):

aqjune-aws:
Umm, I was about to create a draft PR that includes the updates in this PR to check whether internal benchmark check passes, but this PR is in a merge queue and for some reason it can't be removed. I will create a separate draft PR one and check in parallel.

PR #1167, opened 2026-05-14
Add int-valued recursion with termination checking

joscoh:
*Issue #, if available:*

*Description of changes:*

  - Support `decreases <int expr>` termination measures alongside structural ADT recursion
  - Int-recursive functions are pure UFs (no definitional axioms); termination obligations assert non-negativity and strict decrease at each call site
  - Compound measures supported (e.g., `decreases m + n`)
  - Mixed structural/int-valued mutual blocks are rejected
  - New `inlineIfAllCanonical` attribute enables concrete evaluation of int-recursive functions

Core logic in `Strata/Transform/TerminationCheck.lean` — new `DecreasesKind` type classifies each function's measure, unified obligation generation via callback.

Tests in `StrataTest/Languages/Core/Tests/IntRecursionTests.lean` and `StrataTest/Languages/Core/Tests/RecursiveFunctionErrorTests.lean`.

CC @kondylidou 


docs/RecursiveFunctions.md
shigoel, line 1: Any reason why we still maintain this file, instead of porting this over to docs/verso/LangDefDoc.lean?

docs/verso/LangDefDoc.lean
```diff
@@ -464,8 +469,41 @@ rec function zipLen (@[cases] xs : IntList, ys : IntList) : int
...
+  else diagonal(m - 1, n)
+};
+```
+
+#### General Rules
+
 Every recursive function must have at least a `@[cases]` annotation or a
 `decreases` clause; Strata rejects recursive functions without a termination hint.
+If a function has both `@[cases]` and an int-valued `decreases` clause,
+`@[cases]` is used for unfolding in the SMT solver, while `decreases` is used
```
shigoel: I'd recommend describing the semantics independent of an external tool, like an SMT solver. E.g., `@[cases]` enables per-constructor axiom generation and PE, while `decreases` is used for termination checking.

docs/RecursiveFunctions.md
```diff
@@ -55,20 +54,32 @@ rec function zipLen (@[cases] xs : IntList, ys : IntList) : int
...
+
+- If `decreases` is omitted, the `@[cases]` parameter is used as the measure.
+- If `decreases` names a datatype-typed parameter, structural termination is
+  used. The `@[cases]` parameter still controls axiom generation independently.
+- If `decreases` is an expression of type `int`, int-valued termination is used.
+  `@[cases]` is not required.
+- The expression may be compound (e.g., `decreases m + n`).
+- If both `@[cases]` and an int-valued `decreases` are present, `@[cases]` is
+  used for unfolding in the SMT solver, while `decreases` is used for
+  termination checking.
```
shigoel: I think all of this info. is missing from the language definition doc.

StrataTest/Languages/Core/Tests/IntRecursionTests.lean
```diff
@@ -0,0 +1,605 @@
...
+
+---------------------------------------------------------------------
+-- Test 8: mutual int recursion
+---------------------------------------------------------------------
+
+def mutualIntRecPgm : Program :=
+#strata
+program Core;
+
+rec function isEven (n : int) : bool
```
shigoel, line 356: I'd recommend a test that'd check whether the evaluator correctly unfolds mutual int-recursive functions to concrete values (i.e., testing `inlineIfAllCanonical`'s behavior for this case).

E.g., does `isEven(4) == true` provable entirely during PE?

shigoel: Another test recommendation: does `decreases 10` fail, as it's supposed to?

shigoel:
> Int-recursive functions are pure UFs (no definitional axioms);

I'm wondering if it makes sense to provide a mode where int-recursive functions can produce definitional axioms if requested with a way to avoid matching loops in the SMT solver. That'd allow proving e.g., `fib(n) == fib(n-1) + fib(n-2)` without users typing in those assumptions (which can get error-prone for more complicated functions).

And you'd already have seen this, but AR tools, like Dafny and Verus, address this problem by providing a `reveal` keyword and/or fuel-based unrolling. See https://github.com/dafny-lang/dafny/issues/3806, https://verus-lang.github.io/verus/guide/recursion.html#fuel-and-reasoning-about-recursive-functions, etc.

Out of scope of this PR though.

kondylidou:
@joscoh nice!! thanks

docs/RecursiveFunctions.md
joscoh: I wasn't sure how detailed the `LangDefDoc` should be, and this doc keeps track of planned features, etc which I didn't think it made sense to include in the language reference. But I agree it is largely redundant now, and I can remove it if you prefer.

joscoh:
> I'm wondering if it makes sense to provide a mode where int-recursive functions can produce definitional axioms if requested with a way to avoid matching loops in the SMT solver. That'd allow proving e.g., fib(n) == fib(n-1) + fib(n-2) without users typing in those assumptions (which can get error-prone for more complicated functions).

> And you'd already have seen this, but AR tools, like Dafny and Verus, address this problem by providing a reveal keyword and/or fuel-based unrolling. See https://github.com/dafny-lang/dafny/issues/3806, https://verus-lang.github.io/verus/guide/recursion.html#fuel-and-reasoning-about-recursive-functions, etc.

> Out of scope of this PR though.

Yes, I had a discussion with Mikael where we agreed that some kind of `reveal`/`unfold` mechanism (which desugars to "assume the unfolding axiom on the concrete input given") is likely the best idea for now. According to Mikael, fuel-based unrolling in Dafny suffered from proof stability issues. The remaining question (for a future PR) is whether this should be part of Core syntax or whether it should somehow be stored in metadata and reconstructed (which is possible but a bit convoluted).

docs/RecursiveFunctions.md
joscoh: Removed this file

Strata/Transform/TerminationCheck.lean
```diff
@@ -46,13 +50,45 @@ def termProcName (name : String) : String := s!"{name}{termSuffix}"
...
+  | some _ =>
+    .error s!"recursive function '{func.name.name}': decreasing parameter must have a datatype type"
+  | none =>
+    .error s!"recursive function '{func.name.name}': decreasing parameter index {idx} is out of range"
+
+/-- Determine the `DecreasesKind` for a recursive function, validating types
+    against the TypeFactory. -/
+private def getDecreasesKind (func : Function) (tf : @TypeFactory Unit)
+    : Except String DecreasesKind := do
+  let casesIdx := FuncAttr.findInlineIfConstr func.attr
```
MikaelMayer: 🔍 `casesIdx` is computed here but only used in the `| none =>` branch at line 90. Consider moving it inside that branch to make the data flow clearer and avoid computing it when `func.measure` is `some`.

Strata/Transform/TerminationCheck.lean
```diff
@@ -176,32 +226,56 @@ private def mkTySubst (tf : @TypeFactory Unit) (concreteTy : LMonoTy) : Subst :=
...
+      match computeCallMeasure calleeName calleeKind calleeFormals calleeInputTys callArgs with
+      | .ok callMeasure =>
+        if callMeasure.hasBVar then
+          .error s!"termination checking '{func.name.name}': decreasing argument contains a bound variable"
+        else if containsOpCall callMeasure recFuncNames then
+          .error s!"termination checking '{func.name.name}': decreasing argument contains a recursive call"
+        else
+          let decrease := LExpr.mkApp () (@intLtFunc CoreLParams).opExpr
+            [callMeasure, callerMeasure]
+          if needsNonNeg then
```
MikaelMayer, line 255: 🔍 Nit: `needsNonNeg` is determined by the caller's kind, which works because mixed mutual blocks are rejected. A brief comment noting this invariant ("safe because mixed blocks are rejected above") would help future readers understand why this doesn't need to check the callee's kind.

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -1731,7 +1731,7 @@ partial def translateRecFnDecl (p : Program) (preBindings : TransBindings)
   let bodyBindings := { preBindings with boundVars := bbindings }
   let casesAttr := match casesIdx with
     | some i => #[Strata.DL.Util.FuncAttr.inlineIfConstr i]
-    | none => #[]
+    | none => #[Strata.DL.Util.FuncAttr.inlineIfAllCanonical]
```
MikaelMayer, line 1734: 🔍 When `@[cases]` is absent, the function unconditionally gets `inlineIfAllCanonical`. This means *all* recursive functions without `@[cases]` will attempt concrete evaluation when all args are canonical — even if the function doesn't terminate for those inputs (e.g., a function that fails its termination check). Is there a risk of the evaluator looping on non-terminating int-recursive functions with canonical arguments? The fuel-based evaluator (`eval n`) should prevent infinite loops, but worth confirming the fuel is always bounded.

Strata/Languages/Core/FunctionType.lean
```diff
@@ -61,6 +61,19 @@ def typeCheck (C: Core.Expression.TyContext) (Env : Core.Expression.TyEnv) (func
     -- Apply the outer unification substitution back to the body so that type
     -- variables introduced inside it are resolved.
     let bodya := LExpr.applySubstT bodya S.subst
+    -- Validate the measure expression type for int-recursive functions.
+    -- Only validates non-fvar measures (fvar measures are validated in TermCheck
+    -- using the TypeFactory, which has ADT information).
+    match func.measure with
```
MikaelMayer, line 67: 🔍 The comment says "Only validates non-fvar measures (fvar measures are validated in TermCheck using the TypeFactory, which has ADT information)". This split validation is a bit fragile — if someone adds a new code path that skips TermCheck, fvar measures would go unvalidated. Consider adding a brief note in the TermCheck validation explaining that it's the single source of truth for fvar measure validation.

MikaelMayer (commented):
🔍 Well-structured PR that cleanly extends the termination checking infrastructure to support int-valued measures. The `DecreasesKind` discriminated union is a good design choice that makes the two modes explicit. The obligation generation via callback (`mkObligations`) is elegant and avoids code duplication between structural and int-valued paths. The test coverage is thorough, including both positive and negative cases.

Positive highlights:
- The rejection of mixed structural/int-valued mutual blocks is a sensible simplification.
- `inlineIfAllCanonical` is a clean mechanism for enabling concrete evaluation of int-recursive functions.
- The compound measure support (e.g., `m + n`) via `substFvarsLifting` is well-thought-out.
- Good error messages that guide users toward the correct fix.

Strata/Languages/Core/DDMTransform/Translate.lean
joscoh: I did think about this, and I am not sure there is a great solution (we cannot change the evaluation behavior based on the result of the termination check, since this only occurs at verification time). The fuel-based evaluation does prevent true infinite loops, but unfortunately the way that the pipeline is set up currently, we cannot reject evaluation of functions that turn out not to be terminating.

Strata/Languages/Core/FunctionType.lean
joscoh: I can add a comment, but it is worth noting that this split is necessary. We need type context information (in `FunctionType`) to typecheck arbitrary `int`-valued expressions, but we need type factory information to check datatypes. These don't live in the same place currently.

MikaelMayer (commented):

MikaelMayer (approved):

Strata/Transform/TerminationCheck.lean
```diff
@@ -71,10 +71,11 @@ private def validateStructuralParam (func : Function) (tf : @TypeFactory Unit)
     .error s!"recursive function '{func.name.name}': decreasing parameter index {idx} is out of range"
 
 /-- Determine the `DecreasesKind` for a recursive function, validating types
-    against the TypeFactory. -/
+    against the TypeFactory. This is the single source of truth for validating
```
shigoel, line 74: Nice! Thanks for adding this comment.

shigoel (approved):

PR #1177, opened 2026-05-18
Merge `main` into `main2`

joscoh:
Keeps `main2` in sync with `main`

### DIFF
--- a/Strata/DL/Imperative/StmtSemantics.lean
+++ b/Strata/DL/Imperative/StmtSemantics.lean
@@ -66,12 +66,14 @@ inductive Config (P : PureExpr) (CmdT : Type) : Type where
   /-- A terminal configuration, indicating that execution has finished. -/
   | terminal : Env P → Config P CmdT
   /-- An exiting configuration, indicating that an exit statement was encountered.
-      The optional label identifies which block to exit to. -/
-  | exiting : Option String → Env P → Config P CmdT
+      The label identifies which block to exit to. -/
+  | exiting : String → Env P → Config P CmdT
   /-- A block context: execute the inner config, then consume matching exits.
       The label is `Option String` — `none` denotes an unnamed block that only
-      catches unlabeled exits. -/
-  | block : Option String → Config P CmdT → Config P CmdT
+      catches unlabeled exits.  The `SemanticStore P` is the parent store at
+      block entry; on exit, the result is projected through it so that
+      variables initialized inside the block are not visible outside. -/
+  | block : Option String → SemanticStore P → Config P CmdT → Config P CmdT
   /-- A sequence context: execute the first statement (as a sub-config), then
       continue with the remaining statements. -/
   | seq : Config P CmdT → List (Stmt P CmdT) → Config P CmdT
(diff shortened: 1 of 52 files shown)


Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -66,12 +66,14 @@ inductive Config (P : PureExpr) (CmdT : Type) : Type where
   /-- A terminal configuration, indicating that execution has finished. -/
   | terminal : Env P → Config P CmdT
   /-- An exiting configuration, indicating that an exit statement was encountered.
-      The optional label identifies which block to exit to. -/
-  | exiting : Option String → Env P → Config P CmdT
+      The label identifies which block to exit to. -/
+  | exiting : String → Env P → Config P CmdT
   /-- A block context: execute the inner config, then consume matching exits.
       The label is `Option String` — `none` denotes an unnamed block that only
```
MikaelMayer, line 72: 🤖🔍 Nit: The docstring says "The label is `Option String` — `none` denotes an unnamed block that only catches unlabeled exits." Since unlabeled exits no longer exist (exit now requires a `String`), this should be updated. A `.none` block now simply never matches any exit (it only provides scoping via `projectStore`).

Strata/Transform/Specification.lean
```diff
@@ -287,18 +287,32 @@ theorem seq_cons {s : Stmt P CmdT} {ss : List (Stmt P CmdT)}
           exact h₂ ρ₁ ρ' hmid (hwfb_preserved ρ₁ hterm_s) hf₁ (.inr ⟨lbl, hexit_ss⟩)
 
 omit [HasVal P] in
-/-- Lift a `TripleBlock` to a `Triple` by wrapping in a block. -/
+/-- A postcondition is well-formed if it is stable under `projectStore`. -/
+def PostWF (Post : Env P → Prop) : Prop :=
```
MikaelMayer, line 291: 🤖🔍 `PostWF` requires the postcondition to be stable under `projectStore` *and* that `hasFailure` is preserved. The `hasFailure` part is trivially true since `projectStore` only touches the store, not `hasFailure` (the `with store := ...` syntax leaves other fields unchanged). Consider whether the `hasFailure` conclusion adds value or just forces callers to prove something trivial.

MikaelMayer (commented):
🤖🔍 This is a substantial merge bringing two major features from `main` into `main2`:

1. **Provenance type** (#1140): Replaces raw `FileRange` metadata with a structured `Provenance` type that distinguishes real source locations from synthesized origins. Clean migration across all translators.

2. **Scoped var init + mandatory exit labels** (#1141, #1167): Removes unlabeled `exit`, makes `exit` require a `String` label, adds `projectStore` for block-scoped variable initialization, and restructures loop semantics so each iteration's body runs in its own block scope.

The changes are well-structured. The `projectStore` mechanism is clean and the proofs are updated consistently. The `WFEvalExtension` structure is properly extended with the new preservation obligations. The Kleene semantics correctly mirrors the new block scoping.

A few observations below.

joscoh:
These comments are about previous merged PRs and should be addressed in separate followup PRs rather than a merge PR IMO (especially since they should be upstreamed to `main`).

atomb (approved):
Looks good. It's probably worth replacing our current PR description warning with a real message.

MikaelMayer (approved):

PR #1179, opened 2026-05-18
Split StrataMain.lean into importable library and thin executable root

### LINKED ISSUE #1163: Split StrataMain.lean into importable library and thin executable root

MikaelMayer:
Fixes #1163

Moves all library code from `StrataMain.lean` into a new `StrataMainLib` module, leaving `StrataMain.lean` as a thin executable entry point (`import StrataMainLib`). This makes the library importable from tests and other tools without pulling in the executable entry point.


StrataMainLib.lean
```diff
@@ -0,0 +1,1560 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+-- Executable with utilities for working with Strata files.
```
MikaelMayer: 🤖🔍 Nit: This comment says "Executable" but this file is now a library. Consider updating to e.g. `-- Library with utilities for working with Strata files.`

MikaelMayer (commented):
🤖🔍 Clean mechanical split. The code is moved verbatim, visibility is correctly widened for the definitions that `runCommandMap` needs, and the new `runCommandMap` abstraction makes the library reusable from custom executables. One minor nit below.

StrataMainLib.lean
MikaelMayer: 🤖 Fixed — updated the comment to "Library with utilities for working with Strata files."

atomb (approved):

joehendrix (approved):
I'm not sure this is right approach to what is being addressed, but approving to unblock changes.

joscoh (approved):

PR #1190, opened 2026-05-19
fix(eval): isolate Env.error at procedure boundary (issue #1185)

### LINKED ISSUE #1185: `Program.eval` propagates `Env.error` across procedures, silently dropping obligations from later procedures
## Summary

`Strata/Languages/Core/ProgramEval.lean:Program.eval` threads a single `Env` through every procedure declaration in source order. If any procedure's evaluation sets `Env.error`, the error is left in place by `Procedure.fixupError` (`Strata/Languages/Core/ProcedureEval.lean:23-27`) and the errored env is passed as the input env to subsequent procedures. Inside `Statement.eval` and `evalAuxGo`, every statement short-circuits when `env.error.isSome` is true (`Strata/Languages/Core/StatementEval.lean:625-637`), so subsequent procedures' bodies produce no `deferred` obligations. `extractObligations` then walks empty bodies and reports `All 0 goals passed`, even when the affected procedures contain asserts that should produce failing VCs.

The error is never surfaced — there is no warning in stdout, in `--profile` output, or in `--verbose` output. The user sees a clean green pass.

## Severity

Soundness. A real assertion that should fail is silently reported as passing whenever the verifier happens to evaluate a contaminating procedure earlier in the program. The trigger is not exotic — the funcDecl conflict in the reproduction below is a 17-line program that uses only documented Core syntax.

## Reproduction

Save the following as `repro.core.st` and run `strata verify --check-mode deductive repro.core.st`:

```
program Core;

procedure proc_a()
{
  function foo(y : int) : int { y + 1 }
};

procedure proc_b()
{
  function foo(y : int) : int { y + 2 }
};

procedure proc_c()
{
  assert false;
};
```

Output:

```
Successfully parsed.
All 0 goals passed.
```

`proc_c` contains a literal `assert false` and there is no enclosing assumption strong enough to discharge it. The verifier should report a failing VC. Instead it reports a clean pass.

### Why

`proc_a` is evaluated first and adds `foo` to `Env.exprEnv`'s factory via the `funcDecl` arm of `Statement.evalOneStmt`/`evalAuxGo` (`Strata/Languages/Core/StatementEval.lean:609-625`). `proc_b` is then evaluated against the *same* threaded `Env`; its `function foo` declaration calls `Env.addFactoryFunc foo` (`Strata/Languages/Core/Env.lean:241-247`), which fails because `foo` is already in the factory. `evalAuxGo` writes `Env.error := some (.Misc "A function of name foo already exists! …")` (`StatementEval.lean:625`). `proc_c` is then evaluated against the errored env; every command short-circuits on `env.error.isSome` (`StatementEval.lean:637`), so the `assert false` never reaches `deferObligation` and `proc_c` produces zero deferred obligations. `Program.eval` returns this single accumulated env to `toCoreProofObligationProgram` (`Strata/Languages/Core/Core.lean:135-152`), which extracts an empty obligations list, and the verifier prints `All 0 goals passed.`

### Empirical confirmation

A trace `dbg_trace` added at `ProgramEval.lean:62` to print `Env.error` and `Env.deferred.size` after each `Procedure.eval`:

```
[ProgramEval] after proc_a: error=none                                   deferred.size=0
[ProgramEval] after proc_b: error=Misc(A function of name foo …)         deferred.size=0
[ProgramEval] after proc_c: error=Misc(A function of name foo …)         deferred.size=0
```

`proc_b`'s funcDecl conflict sets the error; `proc_c`'s eval inherits it and contributes no obligations. `deferred.size` stays at 0 through every subsequent procedure.

### Reordering rescues the obligation

If `proc_c` is moved ahead of the conflicting pair:

```
program Core;

procedure proc_c()
{
  assert false;
};

procedure proc_a()
{
  function foo(y : int) : int { y + 1 }
};

procedure proc_b()
{
  function foo(y : int) : int { y + 2 }
};
```

Output:

```
Successfully parsed.
repro.core.st(7, 2) [assert_0]: ❌ fail
Finished with 0 goals passed, 1 failed.
```

Same procedures, same error — but `proc_c`'s obligation is collected *before* the error is set, so it survives.

### Trace with reordered input

```
[ProgramEval] after proc_c: error=none                                   deferred.size=1
[ProgramEval] after proc_a: error=none                                   deferred.size=1
[ProgramEval] after proc_b: error=Misc(A function of name foo …)         deferred.size=1
```

## Root cause

`Strata/Languages/Core/ProgramEval.lean` walks `program.decls` sequentially:

```lean
| .proc proc _md =>
  let (E, procStats) := Procedure.eval declsE proc
  go rest E (stats.merge procStats)
```

The post-eval `Env` is passed as the input env for the next procedure.

`Procedure.fixupError` (`ProcedureEval.lean:23-27`) intentionally preserves the error:

```lean
def fixupError (E : Env) : Env :=
  match E.error with
  | none => { E with exprEnv.state := E.exprEnv.state.pop,
                     pathConditions := E.pathConditions.pop }
  | some _ => E
```

Combined with `evalAuxGo`'s `if good.isEmpty then (Ewns, noStats, …)` short-circuit (`StatementEval.lean:637`), every command in every subsequent procedure becomes a no-op once any earlier procedure has errored. The error is never surfaced to the user — it sits in `Env`, silently shaping the rest of the run.

## Where the error is swallowed

Three distinct consumers each have an opportunity to surface the error and none of them do.

**1. `Procedure.fixupError`** (`Strata/Languages/Core/ProcedureEval.lean:23-27`). First chance to drop the error. The clean path pops the procedure's scope frame and returns; the error path returns the env *unchanged*:

```lean
def fixupError (E : Env) : Env :=
  match E.error with
  | none => { E with exprEnv.state := E.exprEnv.state.pop,
                     pathConditions := E.pathConditions.pop }
  | some _ => E   -- error preserved, scope NOT popped, no signal raised
```

Doesn't clear `error`, doesn't lift it into an `Except`, doesn't even pop the procedure's scope frame. The error rides along inside the env.

**2. `Program.eval`'s `go` recursion** (`Strata/Languages/Core/ProgramEval.lean:61-63`). Where the contamination becomes cross-procedural:

```lean
| .proc proc _md =>
  let (E, procStats) := Procedure.eval declsE proc
  go rest E (stats.merge procStats)   -- passes errored E to next proc
```

The function's return type is `Except Strata.DiagnosticModel (List Env × Statistics)`. The `Except` channel is reserved for *DDM-level* failures: axiom-stack mismatch (line 50), top-level `function`/`recFuncBlock` factory conflict (lines 66, 71). Per-procedure PE errors land in `Env.error` instead, and `go` doesn't promote them into the `Except`. So `Program.eval` returns `.ok` even when every later procedure was contaminated.

**3. `toCoreProofObligationProgram`** (`Strata/Languages/Core/Core.lean:135-152`). The final swallow point. After `Program.eval`:

```lean
let (pEs, evalStats) ← Program.eval E
...
let postEvalEnv ← match pEs with
  | [e] => pure e
  | _ => throw (DiagnosticModel.fromMessage s!"... expected exactly 1 evaluation Env, got {pEs.length}")
let blocks := postEvalEnv.deferred.toList.map fun ob => ...
```

Reads `postEvalEnv.deferred` to build the obligations program; **never inspects `postEvalEnv.error`**. The error has been sitting in `Env.error` since `StatementEval.lean:622-625`, but this consumer walks past it. Without the deferred obligations the contaminated procedures would have produced, `extractObligations` walks empty bodies and the verifier prints `All N goals passed.`

The pattern across all three: the error is real data sitting in a field, but no consumer ever checks `Env.error`. Inside `evalAuxGo` itself the error short-circuits subsequent statements (`StatementEval.lean:637`), but at every procedure or program-level boundary above that, it is read past silently.

The funcDecl conflict above is the cleanest user-facing trigger that passes the upstream guards (callElim's pre-PE existence check, precondElim's `already in factory.` check). The contamination mechanism, however, is independent of the trigger — any current or future PE error path has the same effect.

## Suggested fix

Reset `Env.error` (and the per-procedure scope state) at the boundary between procedures so per-procedure failures stay local to that procedure. Two reasonable places:

1. **In `Procedure.eval`'s return**: drop `error` and pop the leftover procedure scope before handing the env back. `fixupError` already does the no-error case; the question is whether to also pop on the error path or to just clear `error` so the next procedure starts clean.

2. **In `Program.eval`'s `go`**: between procedures, take the input env's `factory`, `program`, top-frame `pathConditions`, and any axioms/distincts already accumulated, and discard everything else (`error`, leftover scope frames). Ensures cross-procedure independence by construction.

Either way, the user should also see *some* signal that a procedure errored during PE — at minimum a stat (`Evaluator.peErrorByProc: …`), ideally a stderr warning. Today the error is silently dropped on the floor and a clean-looking `All N goals passed` is the only output.

## Files involved

- `Strata/Languages/Core/ProgramEval.lean:61-63` — the threading.
- `Strata/Languages/Core/ProcedureEval.lean:23-27` — `fixupError` keeps errors.
- `Strata/Languages/Core/StatementEval.lean:625-637` — `evalAuxGo`'s `good.isEmpty` short-circuit.
- `Strata/Languages/Core/StatementEval.lean:609-625` — funcDecl evaluation, where the empirical trigger above writes `Env.error`.
- `Strata/Languages/Core/Env.lean:358-364` — `Env.merge` correctly short-circuits to error envs, which makes ITE branches contaminate the merged env the same way procedures contaminate their successors.

PROgram52bc:
## Summary

Fixes issue #1185 — `Program.eval` propagated `Env.error` across procedures, silently dropping later procedures' obligations and reporting "All 0 goals passed" even with a literal `assert false` in scope.

Two surgical fixes restore local invariants:

- `Procedure.fixupError` (`Strata/Languages/Core/ProcedureEval.lean:23-31`) now unconditionally pops scope/PC frames AND clears `Env.error`, so each procedure's evaluation starts from a clean error state.
- `Env.merge` (`Strata/Languages/Core/Env.lean:358-373`) now unions `deferred` across the error wall when one ITE branch errors, preserving the clean branch's already-accumulated obligations.

Scope is deliberately minimal: visibility/UX (stderr warnings, per-procedure error stat, `peErrors` field) was considered and explicitly cut to a follow-up. No new types, helpers, fields, or signature changes.

## Test plan

- [x] `lake build StrataTest` passes (verified locally — 608/608 jobs).
- [x] `StrataTest.Languages.Core.Tests.ProgramEvalErrorIsolation` (new) — issue's exact 17-line repro reports `assert_0: ❌ fail`; reorder regression still passes.
- [x] `StrataTest.Languages.Core.Tests.EnvMergeErrorPreservesDeferred` (new) — 5 cases covering errored/clean combinations of `Env.merge`, including the both-errored asymmetry contract.
- [x] `Examples/IsolatedProcEvalError.core.st` (new CLI golden) — `Examples/run_examples.sh` reports `Test passed`.
- [x] No existing test regressed.

## Notes for reviewers

- **Existing-proof impact:** none. Confirmed via grep across `*WF.lean`, `VerifierProofs.lean`, and `StatementSemanticsProps.lean` (4091 lines combined): no proof references `fixupError`, `mergeResults`, `Procedure.eval`, `Program.eval`, `Env.merge`, `performMerge`, or `Env.error`. The `merge` references in `VerifierProofs.lean` are about `SMT.Result.merge` (solver-result lattice join), unrelated.
- **Public signatures unchanged.**
- **Soundness argument:** `ProofObligation.assumptions` is `PathConditions Expression` (a value-typed list of assumptions baked in at obligation creation), so an obligation carried across an error boundary remains independently provable.
- **Trade-off accepted:** the underlying PE error (e.g., "function foo already exists") is now silently swallowed. The user sees the assertion failure correctly, but the function-redefinition error itself is no longer surfaced. Visibility work is a small additive follow-up if this becomes a concern in practice.

## Commits

- `ecf402211` test: failing repro for cross-procedure leak
- `d55ac1c33` fix: clear Env.error and pop frames in fixupError unconditionally
- `caec30bd4` test: failing unit tests for Env.merge deferred preservation
- `eeb0dfa3d` fix: union deferred obligations across Env.merge error wall
- `2e260c4b8` docs: fix stale StatementEval line reference in fixupError comment
- `d22edcf9f` test: CLI golden regression
- `490d5cbb8` test/docs: document Env.merge both-errored asymmetry

Closes #1185


Strata/Languages/Core/ProcedureEval.lean
```diff
@@ -21,17 +21,23 @@ open Std
 open Statement Lambda LExpr
 
 def fixupError (E : Env) : Env :=
```
tautschnig, line 22: Two suggestions:

1. **Rename**: `fixupError` is now misleading — the function is no longer error-conditional, it unconditionally pops frames and clears `error`. `closeProcedureScope` (or `finalizeProcedureEnv`) describes the new behavior; `fixupError` reads as "only does something if there's an error".

2. **State the soundness invariant as a theorem**, not just prose. The PR description argues:
   > `ProofObligation.assumptions` is `PathConditions Expression` … so an obligation carried across an error boundary remains independently provable.
   
   That's exactly the kind of claim future refactors are likely to silently break. Worth promoting to:
   ```lean
   theorem fixupError_preserves_deferred (E : Env) :
       (fixupError E).deferred = E.deferred ∧
       (fixupError E).error = none
   theorem deferred_obligations_self_contained (E : Env) (ob : ProofObligation _) :
       ob ∈ E.deferred → ob.assumptions = ob.assumptions  -- i.e. no env reference
   ```
   Combined with a `Procedure.eval` lemma stating the post-procedure `Env` has the same `deferred` regardless of whether evaluation errored, this gives a regression-resistant statement of the property the issue asks for.

Strata/Languages/Core/Env.lean
```diff
@@ -356,10 +356,16 @@ def Env.performMerge (cond : Expression.Expr) (E1 E2 : Env)
   { E1 with exprEnv := exprEnv, pathConditions := pcs, deferred := deferred }
 
 def Env.merge (cond : Expression.Expr) (E1 E2 : Env) : Env :=
+  -- When one branch errors, take that side's structure (state, scope, error,
+  -- pathConditions) but UNION `deferred`, so the sibling's already-accumulated
+  -- obligations survive the error wall. Dropping the non-errored side's path
+  -- conditions is sound: the merged env carries an error and downstream
+  -- evaluation short-circuits before PCs are consulted again. When both sides
```
tautschnig, line 364: Two notes on this branch:

1. The both-errored asymmetry (E1's error wins, E2's silently dropped) is documented in the comment but is invisible to users today — and inconsistent with the visibility follow-up the issue calls out. If the visibility work lands later, please revisit this so both errors propagate (e.g. `error := some (.Combined #[e1, e2])` or just appending into a list field).

2. The argument "dropping the non-errored side's path conditions is sound: the merged env carries an error and downstream evaluation short-circuits before PCs are consulted again" is correct *given* the current short-circuit in `evalAuxGo`, but it's a load-bearing global invariant that's easy to miss. Worth a comment cross-referencing `StatementEval.lean:632` (the `partition` on `env.error.isSome`) so the next reader knows what to preserve.

StrataTest/Languages/Core/Tests/EnvMergeErrorPreservesDeferred.lean
```diff
@@ -0,0 +1,135 @@
...
+-- E1 wins arbitrarily — the merged env carries an error flag either way,
+-- so downstream short-circuits regardless of which side's value is kept.
+/--
+info: some "[ERROR] error A"
+-/
+#guard_msgs in
+#eval ((Env.merge dummyCond E_err_a E_err_b).error.map fun e => f!"{e}").map fun f => Std.Format.pretty f
+
+
+end Core
```
tautschnig, line 135: Test gap: the issue's exact scenario is *errored procedure that had **partially** accumulated deferred obligations before erroring* — e.g., `proc_b` could have a successful `assert P; function foo …` where the `assert P` produced a deferred obligation *before* the `function foo` redefinition triggered the error. This PR's fix preserves those, but no test fixture here exercises that combination.

Please add a Case 6: `E_err_with_oblig` (errored, deferred ≠ ∅) merged with `E_clean_with_oblig` — assert the merged `deferred.size` is the union (3 in your existing helpers' shape). Without it, a future regression that drops `E1.deferred` on the errored branch (the obvious-looking "simplification") would still pass current tests.

tautschnig (commented):
### Note: this target _main_

One deliberate trade-off worth raising: the original issue (#1185) explicitly asks for *both* contamination removal *and* error visibility ("the user should also see *some* signal that a procedure errored during PE — at minimum a stat, ideally a stderr warning"). This PR fixes the first half and *punts the second*, swallowing PE errors silently. The user now sees the assertion failure but has no signal that `proc_b`'s `function foo already exists` happened. Please open a tracking issue for the visibility follow-up before merge so the regression doesn't get lost.

PROgram52bc:
Superseded by CR-285951652

PR #1196, opened 2026-05-20
Add Concrete Syntax for Unstructured Programs in Strata Core

PROgram52bc:
# Add Support for Unstructured Programs in Strata Core
## Description of changes

- Adds syntax and parser support for unstructured (CFG-based) procedure bodies in Strata Core. For example,
```c
cfg entry { 
    // Deterministic block (default): 
    entry: {  
      x := 0; 
      if (x < n) goto entry else done;       
    } 
    done: {
      return; 
    }
  }
  
cfg entry {
    // Nondeterministic block:   
    entry: {   
      y := y + 1;     
      goto entry, done;   // nondeterministic choice 
    }
    done: {
      return; 
    }
  }
```
Prior to this PR, the only way to obtain unstructured programs was to apply `StructuredToUnstructured` transformation. 
- Adds `.core.st` file examples illustrating the syntax of unstructured programs.
- Introduces Procedure.Body as a sum type:
```lean
  inductive Procedure.Body where
    | structured : List Statement → Procedure.Body
    | cfg : DetCFG → Procedure.Body
```
- Adapts uses of `Procedure.body` to handle the `cfg` case based on the context
- **Adds [metadata](https://github.com/strata-org/Strata/pull/1132/changes/4d29cde63ccb04832b99eb6dc90a03ab2faefdcb) support to unstructured programs (new since initial PR draft)**
- **Propagates loop invariants/decreases measure to unstructured CFG during transformation (new since initial PR draft)** 
- **Adds [CFG-based Core-to-GOTO pipeline](https://github.com/strata-org/Strata/pull/1132/changes/63181c964dc0fe526583db18c94302bab891677f) alongside direct path (new since initial PR draft)** 

## Adaptation Methods for `Procedure.body` Uses

1. **CFG (Implemented)** — Functions that handle both structured and CFG bodies with meaningful separate logic for each case.
1. **CFG (N/A)** — Operations where CFGs legitimately contribute nothing (e.g., no local funcDecls, no structured loops), so returning [] or skipping is semantically correct.                                          
1. **CFG (deferred)** — Operations that currently error on CFGs or produce placeholder results, where proper CFG support is desirable but requires non-trivial additional work (e.g., dominator analysis, two-stage    
  pipelines, graph-level inlining).
1. **Proof (Implemented)** — Formal correctness proofs adapted to the `Procedure.Body` sum type.
1. **Proof (deferred)** — Proof obligations where CFG correctness is not yet captured; the current formulation is trivially true for CFGs and is left for future PRs. 

(**Bolded items are updates since the initial PR draft**)

| Category                | Specific Strategy                                                                                                                        | File                                          | Function/Location                                     |
| ----------------------- | ---------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------- | ----------------------------------------------------- |
| **Proof (Implemented)** | **`postconditionsValid` adapted with `CoreBodyExec`**                                                                                    | `Transform/CoreSpecification.lean`            | `ProcedureCorrect.postconditionsValid`                |
|                         | Proves structured if transform succeeds                                                                                                  | `Transform/ProcBodyVerifyCorrect.lean`        | `procToVerifyStmt_is_structured` (new helper theorem) |
|                         | Unifies two `ss` witnesses via `subst`                                                                                                   | `Transform/ProcBodyVerifyCorrect.lean`        | `procBodyVerify_procedureCorrect`                     |
|                         | Adjusted case splits                                                                                                                     | `Languages/Core/ObligationExtraction.lean`    | `extractGo_ok` proof                                  |
| **CFG (Implemented)**   | **Symbolic evaluation of CFG bodies using fuel measure, without path merging.**                                                          | `Languages/Core/ProcedureEval.lean`           | `eval`                                                |
|                         | **Type-check CFG bodies (labels, vars, targets) implemented.**                                                                           | `Languages/Core/ProcedureType.lean`           | `typeCheck`, `checkModificationRights`                |
|                         | **Alternative two-stage pipeline implemented by `procedureToGotoCtxViaCFG`**                                                             | `Backends/CBMC/GOTO/CoreToCProverGOTO.lean`   | `transformToGoto`                                     |
|                         | **Alternative two-stage pipeline implemented by `procedureToGotoCtxViaCFG`**                                                             | `Backends/CBMC/GOTO/CoreToGOTOPipeline.lean`  | `procedureToGotoCtx`                                  |
|                         | **Alternative two-stage pipeline implemented by `procedureToGotoCtxViaCFG`**                                                             | `StrataTest/.../E2E_CoreToGOTO.lean`          | `coreToGotoJsonWithSummary`                           |
|                         | **Alternative two-stage pipeline implemented by `procedureToGotoCtxViaCFG`**                                                             | `Backends/CBMC/CoreToCBMC.lean`               | `createImplementationSymbolFromAST`                   |
|                         | `extractCallsFromStatements` vs `extractCallsFromDetCFG`                                                                                 | `Languages/Core/CallGraph.lean`               | `extractCallsFromProcedure`                           |
|                         | `blockToCST` vs `detCFGToCST`                                                                                                            | `Languages/Core/DDMTransform/FormatCore.lean` | `procToCST`                                           |
|                         | `runStmt` vs `runCFG`                                                                                                                    | `Languages/Core/StatementEval.lean`           | `Command.runCall`                                     |
|                         | Use `CoreBodyExec` with two constructors, and `CoreStepStar` vs `CoreCFGStepStar`                                                        | `Languages/Core/StatementSemantics.lean`      | `EvalCommand.call_sem`                                |
|                         | `Lang.core` vs `Lang.coreCFG`                                                                                                            | `Transform/CoreSpecification.lean`            | `AssertValidInProcedure`                              |
|                         | `transformStmts` vs `transformDetCFG`                                                                                                    | `Transform/PrecondElim.lean`                  | `precondElim`                                         |
|                         | `eraseTypes`, `stripMetaData`, `getVars` implemented and used for CFG                                                                    | `Languages/Core/Procedure.lean`               | `eraseTypes`, `stripMetaData`, `getVars`              |
|                         | `extractFromStatements` vs `extractFromDetCFG`                                                                                           | `Languages/Core/ObligationExtraction.lean`    | `extractObligations`                                  |
| **CFG (N/A)**           | `return .none` on CFG, skip inlining, because the inlined procedure could be structured or unstructured. Semantics is largely undefined. | `Transform/ProcedureInlining.lean`            | `inlineCallCmd`                                       |
|                         | `stmtsToCFG` vs identity, latter is a no-op.                                                                                             | `StrataTest/.../Loops.lean`                   | `singleCFG`                                           |
|                         | Remove loops in structured, pass CFG through because it has no loops                                                                     | `Transform/LoopElim.lean`                     | `loopElim`                                            |
|                         | `[]` for CFG (no local funcDecls in CFG bodies)                                                                                          | `Languages/Core/Core.lean`                    | `buildEnv`                                            |
| **CFG (deferred)**      | Dominator-based path-condition propagation to be implemented                                                                             | `Languages/Core/ObligationExtraction.lean`    | `extractFromDetCFG`                                   |
|                         | `throw` on CFG, due to incompatible return type of `List Statement` instead of `List Command`                                            | `Transform/CoreTransform.lean`                | `runProgram`                                          |
|                         | `throw` on CFG, VC for CFG bodies to be implemented, proof to be adapted as well.                                                        | `Transform/ProcBodyVerify.lean`               | `procToVerifyStmt`                                    |
|                         | Encode structured procedures, handling for CFG procedures to be implemented.                                                             | `Transform/ANFEncoder.lean`                   | `anfEncodeProgram`                                    |
|                         | CFG body renaming for inlining to be implemented                                                                                         | `Transform/ProcedureInlining.lean`            | `renameAllLocalNames`                                 |
|                         | Call extraction for CFG bodies to be implemented                                                                                         | `StrataTest/Boole/global_readonly_call.lean`  | `callHelper`                                          |
| **Proof (deferred)**    | WF Properties `wfstmts`, `wfloclnd`, `bodyExitsCovered` conditioned on structured procedures, props for CFG to be implemented            | `Languages/Core/WF.lean`                      | `WFProcedureProp`                                     |


Examples/CFGSimple.core.st
```diff
@@ -0,0 +1,48 @@
...
+// A simple deterministic CFG: compute max of two integers.
+procedure Max(x : int, y : int, out m : int)
+spec {
+  ensures (m >= x);
+  ensures (m >= y);
+  ensures (m == x || m == y);
+}
+cfg entry {
+  entry: {
+    branch (x >= y) goto then_branch else else_branch;
```
atomb: Does it cause any grammar ambiguities to have this be `branch (x >= y) then goto then_branch else goto else_branch`? I feel like that reads better.

Strata/DL/Imperative/BasicBlock.lean
```diff
@@ -40,8 +40,8 @@ inductive DetTransferCmd (Label : Type) (P : PureExpr) where
 model it instead using `condGoto`. By defining this function, we can easily
 create unconditional jumps, and future proof against the possibility of adding
 it as a constructor in the future.  -/
-def DetTransferCmd.goto [HasBool P] (l : Label) : DetTransferCmd Label P :=
-  condGoto HasBool.tt l l
+def DetTransferCmd.goto [HasBool P] (l : Label) (md : MetaData P := .empty) : DetTransferCmd Label P :=
```
atomb: We're moving in the direction of avoiding default values for metadata arguments. It might mean callers will need extra arguments, and that may cascade, but I think it's ultimately the better approach to require it to be explicitly provided. We're less likely to forget it that way.

Strata/Languages/Boole/Verify.lean
```diff
@@ -1009,8 +1009,10 @@ def toCoreDecls (cmd : BooleDDM.Command SourceRange) : TranslateM (List Core.Dec
     return [.proc {
       header := { name := mkIdent topLevelBlockProcedureName, typeArgs := [], inputs := [], outputs := [] }
       spec := { preconditions := [], postconditions := [] }
-      body := ← toCoreBlock b
+      body := .structured (← toCoreBlock b)
     } .empty]
+  | .command_cfg_procedure _ _ _ _ _ _ =>
+    throwAt default "CFG procedures are not supported in Boole dialect"
```
atomb: Tiny nit:
```suggestion
    throwAt default "CFG procedures are not supported in the Boole dialect"
```

Strata/Languages/Core/StatementEval.lean
```diff
@@ -758,6 +740,36 @@ Interpret a single procedure call.
 
 Importantly, this creates a separate Env to execute the body of the procedure with,
 which initially only contains input/output variables.
+
+Execute a CFG by following control flow from the entry block.
+    Returns the list of commands executed (as statements) for compatibility
```
atomb: This comments seems a little out of date relative to the code. I think it'd be more accurate to say "For each, block, translate the commands within it to a sequence of command statements and evaluate those using the structured interpreter".

Strata/Languages/Core/StatementEval.lean
```diff
@@ -808,10 +820,15 @@ def Command.runCall (lhs : List Expression.Ident) (procName : String) (args : Li
...
-            let configAfter := Imperative.runStmt ops fuel' config
-            match configAfter with
+            let callEnvAfter := match proc.body with
+              | .structured ss =>
+                let config : Imperative.RunConfig Expression Command Env :=
+                  .stmts ss callEnv
+                Imperative.runStmt ops fuel' config
+              | .cfg cfgBody =>
+                -- Interpret CFG by following control flow from the entry block.
+                .terminal (runCFG cfgBody fuel' callEnv ops)
```
atomb, line 828: Oh, nice! I didn't realize this was already implemented for structured statements. It's great to have it for unstructured CFGs, too.

Strata/Transform/LoopElim.lean
```diff
@@ -241,8 +241,12 @@ def loopElim (p : Program) : Program × Statistics :=
   let (decls, stats) := p.decls.foldl (fun (acc, stats) d =>
     match d with
     | .proc proc md =>
-      let (body, st) := StateT.run (Block.removeLoopsM proc.body) {}
-      ((.proc { proc with body := body } md) :: acc, stats.merge st.statistics)
+      match proc.body with
+      | .structured ss =>
+        let (body, st) := StateT.run (Block.removeLoopsM ss) {}
+        ((.proc { proc with body := .structured body } md) :: acc, stats.merge st.statistics)
+      -- CFG bodies have no structured loops; pass through unchanged.
```
atomb: There is a loop elimination transformation we could perform on CFGs, but I don't think we need to implement that now.

StrataTest/Backends/CBMC/GOTO/E2E_CoreToGOTO.lean
```diff
@@ -344,13 +344,18 @@ private def injectPropertySummary (stmts : List Core.Statement) (msg : String)
       .cmd (.cmd (.assert label b (md.withPropertySummary msg)))
     | other => other
 
+-- TODO: This could be split into a two-stage transformation:
+-- 1. structured → cfg (via StructuredToUnstructured)
+-- 2. cfg → CProverGOTO (always operates on CFG, no pattern matching needed)
+-- For now, unstructured bodies are not supported in this test helper.
```
atomb: You do implement that two-stage pipeline in this PR, though.

Strata/Languages/Core/ProcedureType.lean
```diff
@@ -50,6 +50,73 @@ private def setupInputEnv (C : Core.Expression.TyContext) (Env : Core.Expression
...
+        let (pa, newEnv) ← LExpr.resolve C currentEnv p
+          |>.mapError DiagnosticModel.fromFormat
+        currentEnv := newEnv
+        if pa.toLMonoTy != mty[bool] then
+          .error <| DiagnosticModel.withRange sourceLoc
+            f!"[{proc.header.name}]: Branch condition in block \"{lbl}\" \
+               is not of type bool, got {pa.toLMonoTy}"
+        pure (Imperative.DetTransferCmd.condGoto pa.unresolved lt lf md)
+      | .finish md => pure (.finish md)
+    annotatedBlocks := annotatedBlocks ++ [(lbl, { cmds := cmds', transfer := transfer' })]
```
atomb, line 116: 🤖  says:

These are `List` appends in a loop, which is O(n^2). For typical CFG sizes this is negligible, but if large CFGs are expected, consider using `Array` or building in reverse and calling `.reverse` at the end.

Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
```diff
@@ -40,12 +40,12 @@ namespace Strata
 
 public section
 
-private def renameIdent (rn : Std.HashMap String String) (id : Core.CoreIdent) : Core.CoreIdent :=
+def renameIdent (rn : Std.HashMap String String) (id : Core.CoreIdent) : Core.CoreIdent :=
```
atomb, line 49: 🤖 says:

**`renameIdent`, `renameExpr`, `renameCmd` visibility changed from `private` to public.**

   These were made public so `CoreCFGToGOTOPipeline.lean` can reuse them. This is fine, but they lack docstrings. Since they're now part of the public API, brief documentation would be helpful.

Strata/Languages/Core/StatementEval.lean
```diff
@@ -758,6 +740,36 @@ Interpret a single procedure call.
...
+        match Imperative.runStmt ops fuel' (.stmts cmdStmts env) with
+        | .terminal env' =>
+          match blk.transfer with
+          | .finish _ => env'
+          | .condGoto cond lt lf _ =>
+            match ops.evalExpr env' cond with
+            | some (.boolConst _ true) => go lt fuel' env'
+            | some (.boolConst _ false) => go lf fuel' env'
+            | _ => CmdEval.updateError env'
+                (.Misc s!"runCFG: branch condition in block '{label}' did not evaluate to a boolean")
```
atomb: 🤖 says:

**`runCFG` in `StatementEval.lean` does not handle nondeterministic gotos gracefully.**

   `./Strata/Languages/Core/StatementEval.lean:762-770`
   ```lean
   | some (.boolConst _ true) => go lt fuel' env'
   | some (.boolConst _ false) => go lf fuel' env'
   | _ => CmdEval.updateError env'
       (.Misc s!"runCFG: branch condition in block '{label}' did not evaluate to a boolean")
   ```
   The concrete interpreter errors on nondeterministic gotos (where the condition is a free variable). The `Translate.lean` comment explains this is expected, but the error message says "did not evaluate to a boolean" which is misleading — the condition *is* boolean, it's just symbolic/unresolved. A more precise message like "branch condition is symbolic (nondeterministic goto); concrete execution requires deterministic conditions" would help debugging.

Strata/Backends/CBMC/GOTO/CoreCFGToGOTOPipeline.lean
```diff
@@ -0,0 +1,304 @@
...
+/-! ### Core-specific CFG-to-GOTO translation -/
+
+/--
+Translate a Core `DetCFG` to CProver GOTO instructions.
+
+Like `detCFGToGotoTransform` but handles `Core.Command` (which includes
+`CmdExt.call`). The CFG should already have identifiers renamed via
+`renameCoreDetCFG`.
+-/
+def coreCFGToGotoTransform
```
atomb, line 80: 🤖 says:

**`coreCFGToGotoTransform` duplicates significant logic from `detCFGToGotoTransform`.**

   `./Strata/Backends/CBMC/GOTO/CoreCFGToGOTOPipeline.lean:82-180` vs `./Strata/DL/Imperative/CFGToCProverGOTO.lean:74-137`

   The two functions share the same structure (label map, pending patches, loop contract annotation, two-pass resolution). The Core-specific version adds handling for `CmdExt.call` but otherwise duplicates the control-flow translation. Consider whether `detCFGToGotoTransform` could be parameterized by a command translator callback to avoid this duplication. This is a design concern, not a bug.

Strata/Languages/Core/ProcedureType.lean
```diff
@@ -50,6 +50,73 @@ private def setupInputEnv (C : Core.Expression.TyContext) (Env : Core.Expression
...
+        .error <| DiagnosticModel.withRange sourceLoc
+          f!"[{procName}]: Block \"{lbl}\" branches to unknown label \"{lf}\". \
+             Available labels: {labels}"
+    | .finish _ => pure ()
+
+-- Type environment flows sequentially through all blocks in list order,
+-- regardless of control-flow reachability.  This is a sound over-approximation:
+-- it may accept a program that uses an uninitialized variable at runtime (the
+-- verifier will still catch the error), but it never misses a real type error.
+-- Per-block scoping would require a dataflow fixpoint over the CFG.
```
atomb, line 89: 🤖 says:

**Type checking CFG bodies uses sequential flow regardless of control flow.**

   `./Strata/Languages/Core/ProcedureType.lean:89-92` (comment in diff)
   ```lean
   -- Type environment flows sequentially through all blocks in list order,
   -- regardless of control-flow reachability.
   ```
   This means a variable initialized in block B (which appears after block A in the list) will be considered "in scope" for blocks listed after B, even if B is unreachable from those blocks. The comment says this is a sound over-approximation, but it could also *miss* errors: if block C uses variable `x` that is only initialized in block B, but C is reachable without going through B, the type checker won't catch it. The verifier will catch it at runtime, but a type error would be more helpful. This is acceptable for now but worth noting as a known limitation.

Strata/Languages/Core/StatementSemantics.lean
```diff
@@ -285,6 +286,46 @@ inductive CoreStepStar
     ----
     CoreStepStar π φ c₁ c₃
 
+/-- Reflexive-transitive closure of CFG steps for the Core language.
+    Each step looks up a block by label and evaluates it using the generic
+    `Imperative.EvalDetBlock` instantiated with `EvalCommand`. This works
+    because `EvalDetBlock` has a `_enableNesting` constructor that directly
+    references `EvalCmd`, satisfying the Lean kernel's nested inductive
+    requirement. See `StrataTest/Languages/Core/Tests/NestedInductiveRestriction.lean`. -/
```
atomb: 🤖 says:

**`CoreCFGStepStar` docstring references non-existent `_enableNesting` constructor.**

   `./Strata/Languages/Core/StatementSemantics.lean:289-292`
   ```lean
   /-- ... This works because `EvalDetBlock` has a `_enableNesting` constructor
       that directly references `EvalCmd`, satisfying the Lean kernel's nested
       inductive requirement. ...
   ```
   There is no `_enableNesting` constructor in `EvalDetBlock`. The relevant constructor is `cmd`. This should say something like "because `EvalDetBlock` has a `cmd` constructor that directly references `EvalCmd`".

Strata/Languages/Core/DDMTransform/FormatCore.lean
```diff
@@ -919,6 +919,49 @@ partial def invariantsToCST {M} [Inhabited M]
...
+partial def transferToCST {M} [Inhabited M]
+    (t : Imperative.DetTransferCmd String Expression) : ToCSTM M (CoreDDM.Transfer M) := do
+  match t with
+  | .condGoto cond lt lf _ =>
+    if lt == lf then
+      pure (.transfer_goto default ⟨default, lt⟩)
+    else
+      match cond with
+      | .fvar _ id _ =>
+        if id.name.startsWith "$__nondet_" then
```
atomb: I'd say leave out this case and always generate a deterministic CST given a `DetTransferCmd`. We could have an implementation for `NondetTransferCmd` that generates `.transfer_nondet_goto`.

StrataTest/Languages/Core/Tests/NestedInductiveRestriction.lean
atomb, line 1: Although this is a useful explanation for why things are the way they are, I'm not sure whether it's something we should test as part of the test suite. It's really something we wish wasn't true, rather than something we want to be true (which is what tests generally focus on).

atomb (changes requested):
I think this mostly looks great, but it's big enough that my list of comments and small requests got kind of long. :)

Examples/CFGSimple.core.st
PROgram52bc: No, there is no ambiguities with the grammar. I updated it to `branch (cond) goto X else **goto** Y` and updated the examples and parse tests accordingly.

Strata/DL/Imperative/BasicBlock.lean
PROgram52bc: Got it. Removed the `:= .empty` default on `DetTransferCmd.goto`. Cascaded explicit metadata to the three call sites in `StructuredToUnstructured.lean` (`detCmdBlock`, `flushCmds`, the loop's measure-decrease block), all using the existing `synthesizedMd` provenance. The other transfer constructors (`condGoto`, `finish`, `NondetTransferCmd.goto`) still carry the default though; happy to apply the same treatment in a follow-up if desired.

Strata/Languages/Boole/Verify.lean
PROgram52bc: Applied.

Strata/Languages/Core/StatementEval.lean
PROgram52bc: Updated. Thanks for noticing that!

Strata/Languages/Core/StatementEval.lean
PROgram52bc: :)

Strata/Transform/LoopElim.lean
PROgram52bc: Acknowledged. Leaving as-is for this PR; CFG-level loop elimination can be added when needed.

StrataTest/Backends/CBMC/GOTO/E2E_CoreToGOTO.lean
PROgram52bc: Rewrote the comment. The TODO predated the two-stage pipeline; replaced it with a note explaining why this test helper uses the direct path (its `injectPropertySummary` pattern-matches on `Core.Statement`, so adding CFG support means an analogous injection over `DetCFG.blocks`).

Strata/Languages/Core/ProcedureType.lean
PROgram52bc: Deferred to a follow-up PR. Not on the critical path for this change.

Strata/Backends/CBMC/GOTO/CoreToGOTOPipeline.lean
PROgram52bc: Added brief docstrings to `renameIdent`, `renameExpr`, and `renameCmd`.

Strata/Languages/Core/StatementEval.lean
PROgram52bc: Reworded the error message to explicitly mention the symbolic / nondeterministic-goto case.

Strata/Backends/CBMC/GOTO/CoreCFGToGOTOPipeline.lean
PROgram52bc: That's right. We are keeping the duplication for this PR, so the user-facing utilities stay unchanged. Once the two-stage pipeline is stable (testing benchmarks in progress on `htd/smack`), we can unify the algorithm.

Strata/Languages/Core/ProcedureType.lean
PROgram52bc: Acknowledged. The dataflow-aware refinement belongs in a future type-checker pass; deferred to the same follow-up as the O(n^2) note.

Strata/Languages/Core/StatementSemantics.lean
PROgram52bc: Fixed. Constructor is `cmd`. Also removed the dead reference to the deleted `NestedInductiveRestriction.lean` from this docstring.

Strata/Languages/Core/DDMTransform/FormatCore.lean
PROgram52bc: Made the change. `transferToCST` now always emits a deterministic CST given a `DetTransferCmd` (`transfer_goto` when targets coincide, else `transfer_cond_goto`). Added a separate `nondetTransferToCST` for `NondetTransferCmd`. The `$__nondet_` sniff is dropped; nondet-goto round-tripping via the Det path is no longer special-cased.

StrataTest/Languages/Core/Tests/NestedInductiveRestriction.lean
PROgram52bc: Yeah. I never meant to commit this file to the PR, but I did. Just removed it; also removed the reference to it from `CoreCFGStepStar`'s docstring.

PROgram52bc (commented):
Thanks for reading through such as tedious PR :) I addressed the comments you gave. Please let me know if there's anything you'd like to see differently.

Strata/DL/Imperative/BasicBlock.lean
atomb: Yeah, let's continue to propagate this change at least to the boundary of code touched by this PR.

atomb (dismissed):
I would like to continue propagating the change to remove default metadata, but that could be in a separate PR. Otherwise I think this looks great.

PROgram52bc:
> I would like to continue propagating the change to remove default metadata, but that could be in a separate PR. Otherwise I think this looks great.

Thanks! Drafted in #1305. Could combine into this PR as well.

atomb (dismissed):

atomb (dismissed):

StrataTest/Transform/ProcedureInlining.lean
```diff
@@ -214,19 +214,56 @@ def alphaEquivStatement (s1 s2: Core.Statement) (map:IdMap)
...
+private def alphaEquivCFG (cfg1 cfg2 : Core.DetCFG) (map : IdMap)
+    : Except Format IdMap := do
+  let map ← IdMap.updateLabel map cfg1.entry cfg2.entry
+  if cfg1.blocks.length ≠ cfg2.blocks.length then
+    .error f!"CFG block count mismatch: {cfg1.blocks.length} vs {cfg2.blocks.length}"
+  else
+    (cfg1.blocks.zip cfg2.blocks).foldlM (fun map ((lbl1, blk1), (lbl2, blk2)) => do
+      let map ← IdMap.updateLabel map lbl1 lbl2
+      let map ← alphaEquivCmds blk1.cmds blk2.cmds map
+      alphaEquivTransfer blk1.transfer blk2.transfer map) map
```
aqjune-aws, line 246: I think alpha-equivalence of two unstructured CFG is not necessary for now because ProcedureInlining doesn't support it - WDYT?

Strata/Transform/ANFEncoder.lean
```diff
@@ -274,8 +274,14 @@ def anfEncodeProgram (p : Program) : Bool × Program :=
...
-      let (body', idx') := anfEncodeBody proc.body idx
-      (.proc { proc with body := body' } md :: acc, idx', changed || idx' > idx)
+      match proc.body with
+      | .structured ss =>
+        let (body', idx') := anfEncodeBody ss idx
+        (.proc { proc with body := .structured body' } md :: acc, idx', changed || idx' > idx)
+      | .cfg _ =>
+        -- CSE on CFGs would require dominator analysis to determine where to
+        -- place hoisted var declarations. Skipped for now.
+        (.proc proc md :: acc, idx, changed)
```
aqjune-aws, line 284: Instead of silently passing, I think it is better to explicitly throw a message. It seems it is a good timing to change the type of `anfEncodeProgram` to use `CoreTransformM`? :)

Strata/Languages/Core/StatementSemantics.lean
```diff
@@ -284,6 +286,45 @@ inductive CoreStepStar
...
+    `EvalCmd`, satisfying the Lean kernel's nested inductive requirement. -/
+inductive CoreCFGStepStar
+    (π : String → Option Procedure)
+    (φ : CoreEval → PureFunc Expression → CoreEval) :
+    DetCFG → CFGConfig String Expression →
+    CFGConfig String Expression → Prop where
+  | refl : CoreCFGStepStar π φ cfg c c
+  | step :
+    List.lookup t cfg.blocks = .some b →
+    Imperative.EvalDetBlock Expression (EvalCommand π φ) (EvalPureFunc φ) σ b config →
```
aqjune-aws, line 344: It seems this is big-step semantics because it runs the block from begin to end at once. Comparing to `CoreStepStar`, even if `CoreStepStar` is a transitive closure of steps, the individual `.step` is a small step.

Strata/Languages/Core/StatementSemantics.lean
```diff
@@ -284,6 +286,45 @@ inductive CoreStepStar
...
+  | structured :
+    CoreStepStar π φ
+      (.stmts ss ⟨σ, δ, false⟩)
+      (.terminal ρ') →
+    CoreBodyExec π φ (.structured ss) σ δ ρ'.store ρ'.eval ρ'.hasFailure
+  | cfg :
+    CoreCFGStepStar π φ cfg
+      (.cont cfg.entry σ false)
+      (.terminal σ' failed) →
+    CoreBodyExec π φ (.cfg cfg) σ δ σ' δ failed
```
aqjune-aws, line 373: Even if CFG doesn't extend `eval`, I think there is a benefit in sharing the `Env` type with the structured control flow case - WDYT..? It will make the type of `CoreBodyExec` ... -> Env -> Env -> Prop very nicely.

aqjune-aws (commented):
I have a small suggestion about the overall direction re. updates in semantics in correctness of transformation. I think there are many interesting points to discuss, and if we split the PR into (1) the syntactic updates (which is already almost complete?) and (2) the following semantic update (which I can help of course), then it can be a truly nice extension of the Core language.

StrataTest/Transform/ProcedureInlining.lean
PROgram52bc: Right. I found it tricky to decide where this PR should stop growing. For ProcedureInlining of unstructured programs, I reasoned that it was complicated enough to deserve a separate PR, so I didn't implement it. But for alpha-equivalence, it was straightforward enough for me (using an agent) to write it, so I decided to include it. 
Imo, if we have the intention to support procedure inlining for CFGs in the future, it doesn't hurt to have it here for now, but I'm happy to go either way.

Strata/Transform/ANFEncoder.lean
PROgram52bc: Right, I think that makes sense. Thanks for pointing it out. I tried to be explicit about errors, but somehow missed this one. The signature change would also be good.

Strata/Languages/Core/StatementSemantics.lean
PROgram52bc: I think it can be seen as small step in the sense that it evaluates one block per step. The reason for this design is mostly due to the simplicity of `CFGConfig` that it doesn't hold the "rest of the commands". I would prefer keeping it unless there is strong reason to modify, since some of the ongoing proof depends on this, and changing it to a "true small step" semantics would seem to make the proof longer...

Strata/Languages/Core/StatementSemantics.lean
PROgram52bc: Yeah, I can see how it will make the signature cleaner. I'm not sure what it will do to the proof though, especially now I think we can eliminate `extendEval` completely from `EvalDetBlock` and `EvalNondetBlock`, which seems to simplify things in the opposite way... 
But at the minimum, I do think it's something worth experimenting.

StrataTest/Transform/ProcedureInlining.lean
aqjune-aws: Yeah, I think it is safer to remove it because unexecuted code is more likely to have a bug than routinely executed one... And if someone wants to use this, they might find that actually this had a bug.

Strata/Languages/Core/StatementSemantics.lean
aqjune-aws: Probably we can split the semantics-related part out as a separate PR to continue the discussion. But my take is that small-step semantics typically allows us to do something more interesting, and as you could see from the CallElim proof, the size of proofs won't actually change very much. :)
For example, let's assume that we have a basic block:
```
BB1:
  instruction1;
  instruction2;
```
And we split it into two smaller BBs:
```
BBA:
  instruction1;
  goto BBB;
BBB:
  instruction2;
```
We know that BB1 is equivalent to BBA + BBB. A straightforward way to prove this equivalence is that the 'intermediate state' after instruction1 in BB1 is equivalent to the program state at the entrance of BBB. However, if bigstep is used, it is not possible to describe the program state in the middle of BB1 (you can create a temporary basic block that contains instruction1 only, but this isn't syntactically equal to BB1.)
What you will need is just an extra field in a program state that describes instructions that are remaining to run in the block. `StepStmt` already uses this style.

atomb:
We've changed plans and won't merge this PR as-is. Instead, we merged part of it yesterday (#1342) and will merge two other parts (#1347, #1348) after the restructure is finished.

PROgram52bc:
Merged into Brazil mainline.

PR #1203, opened 2026-05-21
laurel: wire old() to Core two-state semantics

julesmt:
laurel: wire old() to Core two-state semantics

Modifies-clause frame conditions now flow through Core's `old`-prefixed
identifiers via Core's native two-state semantics, replacing the previous
synthetic `$heap_in` parameter.

1. HeapParameterization: heap-writing procedures take `$heap` as a true
   inout parameter (same name in inputs and outputs) rather than a
   `$heap_in` / `$heap` pair with a synthesized assignment prelude.

2. ModifiesClauses: frame condition references `old($heap)` via
   `StmtExpr.Old` instead of a separate `$heap_in` variable.

3. New PushOldInward Laurel-to-Laurel pass: distributes `StmtExpr.Old`
   through its sub-expressions until each `Old` immediately wraps a Local
   Var. Warns if `old(...)` does not mention any inout parameter.

4. LaurelToCoreTranslator: `Old (Var (Local n))` translates directly to
   `fvar (mkOld n)`. Inout parameters at call sites are detected and emit
   `.inoutArg` rather than paired `.inArg` + `.outArg`. Call-arg
   construction is shared between the two StaticCall sites via a new
   buildCallArgs helper.

Tests: T9_OldHeapTwoState drives the feature end-to-end — `old` of heap
field reads, arithmetic/unary/comparison sub-expressions, `old(old(...))`,
`old` inside quantifiers and if-then-else, multiple modifies/ensures,
modifies-wildcard, wrong-body negative cases, the no-inout warning, and a
caller asserting the post-state. PushOldInward's normalization shape (every
`Old` wraps an inout Local Var) is enforced at translate time in
LaurelToCoreTranslator: the `.Old` arm emits a `StrataBug` diagnostic if the
invariant is ever violated.


Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -473,22 +513,36 @@ def translateStmt (stmt : StmtExprMd)
         exprAsUnusedInit stmt md
       else
         let coreArgs ← args.mapM (fun a => translateExpr a)
-        -- Generate throwaway LHS variables for all outputs so Core arity
-        -- checking passes (lhs.length == outputs.length).
-        let outputs := match model.get callee with
-          | .staticProcedure proc => proc.outputs
-          | .instanceProcedure _ proc => proc.outputs
-          | _ => []
+        let (calleeInputs, calleeOutputs) := match model.get callee with
```
keyboardDrummer: Can you extract the duplication between the two cases that handle StaticCall, the bare call and the assigning call?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -135,6 +138,18 @@ def throwExprDiagnostic (d : DiagnosticModel): TranslateM Core.Expression.Expr :
   let id ← freshId
   return LExpr.fvar () (⟨s!"DUMMY_VAR_{id}", ()⟩) none
 
+private def oldifyExpr (inoutNames : List String) : Core.Expression.Expr → Core.Expression.Expr
```
keyboardDrummer: Should there be a warning if old is pushed inward but hits no inout parameter?

Interesting to have this transformation at the Core level. Alternatively you can have a "PushOldInward" phase at the Laurel level that pushes `old` inwards until it hits identifiers, so `LaurelToCoreTranslator` only needs to support `Old` that immediately wraps an identifier. I think if you use `mapStmtExprM` then `PushOldInward` could be about as short as this function.

I think I would prefer that since it would move a little code out of `LaurelToCoreTranslator` and I prefer to keep this transformation as dumb as possible. Computing the Core in/inout/out arguments in `LaurelToCoreTranslator` is unfortunate but seems unavoidable unless we change Laurel.

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
```diff
@@ -0,0 +1,92 @@
+/-
```
keyboardDrummer, line 1: Since this PR is implicitly also adding support for inout parameters to Laurel, could you add a separate test for that? Could you also check how Resolution.lean currently deals with in and out parameters that have the same name? How come it is not reporting duplicate names in this PR ?

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
```diff
@@ -0,0 +1,92 @@
...
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+/-
+End-to-end coverage of the two-state path that runs through Core's
```
keyboardDrummer: I'm confused about this test because the programs being tested do not use the `old` keyword. The tests here are for things that I think are already tested in T2_ModifiesClauses. I see that this PR also simplifies the implementation of ModifiesClauses.lean, but since that's a refactoring it shouldn't need additional testing. Could you replace the tests here with tests that use `StmtExpr.Old` ? Also, given your implementation, you could use `old` both for the implicit heap parameter, and for explicit inout parameters, so we should test both.

Secondly, this comment describes part of the implementation of StmtExpr.Old, which I think shouldn't be part of a test.

keyboardDrummer (changes requested):
Nice!! This feature is super useful to have. Love the reliance on Core's support for old.

keyboardDrummer:
Note: this PR should target main2 given the ongoing Strata refactoring

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
julesmt: Yes will do

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
julesmt: `resolveParameter` calls `defineNameCheckDup` on every input and output, so a user-written procedure with the same name in both does get a duplicate-definition error at the initial resolve.

Is there's a user-facing way to declare an `inout` parameter today? I don't really know how to add a "user-declared inout" test... @keyboardDrummer

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
julesmt: You're right that the test programs don't write old(...) — turns out .Old has no Laurel surface syntax today. My bad

julesmt:
@keyboardDrummer I added changes, let me know what you think

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
keyboardDrummer: Please add the grammar for `old` and write tests that depend on user programs that use `old`

StrataTest/Languages/Laurel/PushOldInwardTest.lean
```diff
@@ -0,0 +1,454 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+/-
+Unit tests for `pushOldInward`. Builds StmtExpr AST nodes directly (since
```
keyboardDrummer: Wouldn't be nicer to have Laurel programs written in Laurel syntax, instead of a test like this?

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
keyboardDrummer: I think we should add a test that contains a program like:

```
procedure hasInout(x: int) returns (x: int) {
  x := x + 1
}

procedure hasInoutCaller() {
  var x: int := hasInout(3);
}
```

and also one that fails somewhere:

```
procedure hasBadInout(x: int) returns (x: bool) {
  x := x + 1
}
```

Strata/Languages/Laurel/PushOldInward.lean
```diff
@@ -0,0 +1,137 @@
...
+  | .Quantifier mode param trigger body =>
+      let trigger' ← trigger.mapM pushOld
+      return wrap (.Quantifier mode param trigger' (← pushOld body))
+  | .Old inner => pushOld inner -- old(old(e)) ≡ old(e)
+  | _ => return expr
+
+/-- Top-down rewrite: every `Old e` is replaced by the result of distributing
+    `Old` through `e`. Warns once per user-written `Old` that does not mention
+    any inout parameter. -/
+partial def pushOldInwardExpr (expr : StmtExprMd) : PushOldM StmtExprMd := do
```
keyboardDrummer: I think this file can be a bit simpler.

We want to first find all the `old` expressions using a generic traversal that is able to stop when we want it to. The current `mapStmtExpr` can't stop, but we can add another version that can. See here: https://github.com/keyboardDrummer/Strata/blob/4c7e39d472382025d31583a9bdc9af776d5c7fba/Strata/Languages/Laurel/MapStmtExpr.lean#L115

So we'll have one call to `mapStmtExprPrePostM` that as a pre function, returns `none` for all non-old nodes, and when it sees an old, calls `mapStmtExpr` with a handler for `.Var` that applies the important logic of this PR, and tracks whether we applied a change. After the call to `mapStmtExpr` resolves, we check whether a change was made, and if not report the `old(...) has no effect` warning. The handler passed to `mapStmtExpr` also has a handler for old, which will report a warning that `old` is used inside another `old`, which has no effect. 

I hope that this way, we won't need `mentionsAnyInout`, handlers for anything other than `.Var` and `.Old`, or a recursive call to pushOldInwardExpr (which currently results in a partial keyword because termination is unproven).

StrataTest/Languages/Laurel/PushOldInwardTest.lean
```diff
@@ -0,0 +1,454 @@
+/-
```
keyboardDrummer: I think I'd prefer not having this test-file. Do you think it's still adding value?

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
```diff
@@ -0,0 +1,198 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+
+/-
+End-to-end coverage of user-written `old(...)` in Laurel postconditions.
```
keyboardDrummer, line 8: Nice! The warning you added isn't tested though. Would it be possible to test? It'll be the first Laurel warning so the testing framework might not support it well.

StrataTest/Languages/Laurel/PushOldInwardTest.lean
julesmt: Well I caught some bugs with it, but now it's not that useful no...

Strata/Languages/Laurel/PushOldInward.lean
julesmt: Ok I think I understood

StrataTest/Languages/Laurel/Examples/Objects/T9_OldHeapTwoState.lean
julesmt: Yes and it doesn't work, I will just stick to Errors

tautschnig:

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -301,7 +301,16 @@ def translateExpr (expr : StmtExprMd)
 
   | .AsType target _ => throwExprDiagnostic $ diagnosticFromSource expr.source "AsType expression translation" DiagnosticType.NotYetImplemented
   | .Assigned _ => throwExprDiagnostic $ diagnosticFromSource expr.source "assigned expression translation" DiagnosticType.NotYetImplemented
-  | .Old value => throwExprDiagnostic $ diagnosticFromSource expr.source "old expression translation" DiagnosticType.NotYetImplemented
+  | .Old value =>
+      -- After PushOldInward, every `Old` immediately wraps a Local Var of an inout parameter.
+      match value.val with
+      | .Var (.Local name) =>
+          let coreTy ← translateType (model.get name).getType
+          return .fvar () (Core.CoreIdent.mkOld name.text) (some coreTy)
```
tautschnig: **Soundness invariant trusted, not verified.** The translation `Old (Var (Local n)) ↦ fvar () (CoreIdent.mkOld n)` silently produces an `fvar` referring to `mkOld n` even when `n` is *not* an inout parameter. If `PushOldInward` ever has a bug (or the AST is touched between PushOldInward and this translator — e.g. by a future pass that synthesizes `Old (Var Local)`), this site emits an unbound-or-misbound name with no diagnostic.

Two options:

1. **Cheap defense:** check `n` against the current procedure's inout-name list and `throwExprDiagnostic … StrataBug` if absent. Catches regressions at translate time.
2. **Stronger:** state and prove the `PushOldInward` output invariant in Lean and discharge the obligation here. Sketch:
   ```lean
   theorem pushOldInward_old_normalized (p : Program) :
       ∀ proc ∈ (pushOldInward p).fst.staticProcedures,
       ∀ e ∈ proc.allStmtExprs,
       ∀ subOld ∈ e.subOlds,
       ∃ n, subOld = .Old ⟨.Var (.Local n), _⟩ ∧
            n.text ∈ procInoutNames proc
   ```
   With this lemma, the `_ =>` arm here can be eliminated entirely (proven unreachable), and the inout assumption becomes structural rather than commented.

Strata/Languages/Laurel/PushOldInward.lean
```diff
@@ -0,0 +1,92 @@
...
+  | .Old inner =>
+    warn expr.source "nested `old(...)` has no effect"
+    return inner
+  | _ => return expr
+
+/-- Top-down: when we hit an `Old`, distribute via `mapStmtExprM` and signal
+    that recursion should stop here (children handled internally). For all
+    other nodes we return `none` so the generic traversal recurses normally. -/
+private def visitOld (expr : StmtExprMd) : PushOldM (Option StmtExprMd) := do
+  match expr.val with
```
tautschnig, line 49: Minor: the `changed` field is shared mutable state — `visitOld` resets it, runs `mapStmtExprM insideOld`, then reads it. This works only because `mapStmtExprPrePostM` returns `some` for the outer `Old`, ensuring the generic traversal does not recurse into siblings while we're inspecting `changed`. That ordering invariant is load-bearing but invisible at this site.

Two cleaner alternatives:

1. Have `insideOld` return `Bool × StmtExprMd` (whether it wrapped) and thread through a custom traversal — eliminates the shared state.
2. At minimum, a doc-line above `visitOld` cross-referencing `mapStmtExprPrePostM`'s short-circuit semantics so the next reader knows why the `modify ... reset` / read pattern is sound.

julesmt: I divided the proof in sub lemmas, I will see if claude can grind through them, or not ...

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
julesmt: The theorem only talks about **pushOldInward**'s direct output, but the program reaching the translator goes through 6 more passes after **PushOldInward**: eliminating the `_ => arm` would require additionally proving each of those passes preserves the canonical shape

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -552,7 +552,19 @@ partial def toSMTOp (E : Env) (fn : CoreIdent) (fnty : LMonoTy) (ctx : SMT.Conte
     | .bv ⟨_, .SGe⟩  => .ok (.app Op.bvsge,      .bool,   ctx)
     | .bv ⟨n, .Concat⟩ => .ok (.app Op.bvconcat,    .bitvec (n * 2), ctx)
     | .bv ⟨_, .ToUInt⟩  => .ok (.app Op.ubv_to_int,  .int,            ctx)
-    | .bv ⟨_, .ToInt⟩   => .ok (.app Op.sbv_to_int,  .int,            ctx)
+    | .bv ⟨n, .ToInt⟩   =>
```
tautschnig: Split the SMT bv↔int cast changes into their own PR, or say why they ride along here.

<details>
<summary>Why</summary>

This `sbv_to_int` expansion, plus `ubv_to_int`→`bv2nat` (Op.lean), `int_to_bv`→`int2bv` (DDMTransform/Translate.lean), and the `BvIntCastTests` update, are an SMT-LIB conformance fix for bitvector↔int casts — independent of the `old()`/two-state Laurel work. This PR's only test, `T9_OldHeapTwoState`, is a translation/diagnostic test (`processLaurelFile`) and never invokes the SMT encoder, so these changes aren't exercised by anything in this PR. Splitting keeps both changes reviewable and bisectable.

The expansion itself looks correct: `ite(bvslt(x, 0), bv2nat(x) - 2^n, bv2nat(x))` is the standard two's-complement → int.
</details>

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -352,6 +381,35 @@ def throwStmtDiagnostic (d : DiagnosticModel): TranslateM (List Core.Statement)
...
+  let calleeOutputNames := calleeOutputs.map (·.name.text)
+  let calleeInoutNames := calleeInputNames.filter (calleeOutputNames.contains ·)
+  let inoutInputIndices := calleeInputNames.zipIdx.filterMap fun (name, i) =>
+    if calleeInoutNames.contains name then some i else none
+  let mut callArgs : List (Core.CallArg Core.Expression) := []
+  for (arg, i) in coreArgs.zipIdx do
+    if inoutInputIndices.contains i then
+      match arg with
+      | .fvar _ ident _ => callArgs := callArgs ++ [.inoutArg ident]
+      | _ => callArgs := callArgs ++ [.inArg arg]
```
tautschnig: Emit a `StrataBug` here instead of silently degrading a non-`fvar` inout argument to `.inArg`.

<details>
<summary>Why</summary>

At an inout parameter index, a non-`fvar` argument falls through to `.inArg arg`, dropping the output side of the inout — a silent mistranslation if it ever occurs. It's unreachable for `$heap` (always a local var), but the `.Old` case earlier in this file defends exactly this kind of "can't happen" with a `StrataBug` diagnostic; matching that here keeps the invariant visible rather than failing quietly.
</details>

tautschnig (commented):
- The PR description's "Tests: PushOldInwardTest …" paragraph is stale — that file was removed at review request; coverage now lives in T9.
- This branch predates #1222's pipeline refactor, so `LaurelCompilationPipeline.lean` (still inline pass records here vs. `LaurelPass` objects there) will conflict with #1222 — whoever merges second should expect it.
- The proof file build emits a `linter.unusedSimpArgs` warning (unused `StateT.modifyGet` in a `simp only`); worth removing.

Strata/Languages/Laurel/PushOldInwardInvariant.lean
```diff
@@ -0,0 +1,6047 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+public import Strata.Languages.Laurel.PushOldInward
```
keyboardDrummer: I haven't read this file, but I would guess that if someone refactors the implementation code in this PR, or the Laurel AST, that this proof then breaks. Can you say something about how much effort you expect it is to fix breakages in this proof? How much work was it to generate it ? Unless we're confident this proof won't hurt maintainability, then I feel now would not be the right time to merge it since our compiler is still very much in flux.

keyboardDrummer (dismissed):

Strata/Languages/Laurel/PushOldInwardInvariant.lean
julesmt: I took a lot a time. I don't see how I can make it more generic indeed, you want to remove it ? @keyboardDrummer

keyboardDrummer: Could you add a comment at the top of the file stating how it was generated, including the effort involved, that the code was intentionally not reviewed by anyone, and that if it turns out that it's difficult to maintain this file that we'll remove it?

julesmt: done ! @keyboardDrummer

julesmt:
@tautschnig @keyboardDrummer need a final review pls

keyboardDrummer (dismissed):

keyboardDrummer:
Thanks for adding the elaborate comment in the proof file, and sorry to see it already broke somewhat.

Still, great PR ! Thank you

keyboardDrummer (dismissed):

keyboardDrummer (dismissed):

fabiomadge:
`PushOldInwardInvariant.lean` is now `sorry`-stubbed (`#print axioms` → `sorryAx`), so it proves nothing — which meets the condition @keyboardDrummer set ("if it's hard to maintain we'll remove it") and the file's own "regenerate or drop, don't hand-repair" note. Suggest dropping it: 4229 inert lines (nothing imports it) no longer carrying their claimed guarantee.

Soundness is unaffected — the real net is the `StrataBug` check in `LaurelToCoreTranslator`'s `.Old` arm (confirmed it fires; `old()` also verifies end-to-end via SMT). Two spots still call the proof live and are worth fixing if it's dropped: the PR description ("is proved in PushOldInwardInvariant.lean") and the `.Old` arm comment ("The static guarantee is `pushOldInward_old_normalized`").

kadirayk (commented):
I agree with removing the proof if it is not used. Other than that LGTM. 
Very nice simplification: heap-writers now take `$heap` as a true inout and drop the `$heap_in` param + injected `$heap := $heap_in`. T9_OldHeapTwoState passes: positive `old()` cases verify via SMT and the wrong-body/wrong-relation cases correctly fail.

keyboardDrummer (dismissed):

julesmt:
@keyboardDrummer @fabiomadge @kadirayk @tautschnig Could I have an approval for merge pls?

keyboardDrummer (approved):

fabiomadge (approved):

PR #1205, opened 2026-05-21
Fix SMT encoding for polymorphic Sequence preconditions (#1201)

### LINKED ISSUE #1201: SMT encoding errors in #1197
PR #1197 fails CI with some SMT encoding errors. I asked Kiro for a diagnosis, and got the following.

## Root Cause

The error `🚨 SMT Encoding Error! Unimplemented encoding for type var $__ty7` occurs because the `PrecondElim` transformation generates precondition assertions for polymorphic `Sequence` operations (`select`, `update`, `take`, `drop`) without applying the type substitution from the call site.

The pipeline order is: `PrecondElim` → `typeCheck` → `symbolicEval` → SMT encoding.

When `PrecondElim` generates a bounds-check assertion for a call like `Sequence.select(state_out, 0)` where `state_out : Sequence bv32`:

1. It extracts the precondition from `seqSelectFunc`, which is defined with type variable `%a` (e.g., `0 <= i && i < Sequence.length(s)` where `Sequence.length` has type `Sequence %a → int`)
2. It substitutes the **value** arguments (`state_out` for `s`, `0` for `i`) via `substitutePrecondition`
3. But it does **not** substitute the **type** variable `%a` with the concrete type `bv32`

The resulting assertion still contains `%a` in the type annotations of operators like `Sequence.length`. When the type checker processes this, it creates fresh type variables (`$__ty7`) but can't always resolve them to concrete types — especially when:
- The argument expression itself has a polymorphic type annotation (e.g., `Sequence.empty` with type `Sequence %a`)
- The call is inside a polymorphic function body (e.g., `Seq_lib_insert<T>`)

When the SMT encoder encounters the unresolved `$__ty7`, it fails because `ctx.tySubst` has no binding for it.

## The Fix

The fix should be in `Strata/DL/Lambda/Preconditions.lean`, in the `collectWFObligations` function. After substituting value arguments into the precondition, it should also compute and apply the **type substitution** derived from the call site.

Specifically, in the `callObligations` branch where it processes a function call:

```lean
let callObligations := match Factory.callOfLFunc F e with
  | some (op, args, func) =>
    if func.preconditions.isEmpty then []
    else
      let md := e.metadata
      -- NEW: Compute type substitution from the call site
      let tySubst := func.computeTypeSubst op args |>.getD Subst.empty
      func.preconditions.map fun precond =>
        let substedPrecond := substitutePrecondition precond.expr func.inputs args
        -- NEW: Apply type substitution to resolve polymorphic type variables
        let substedPrecond := substedPrecond.applySubst tySubst
        { funcName := func.name.name
          obligation := wrapImplications implications substedPrecond
          callSiteMetadata := md
          precondMetadata := precond.md : WFObligation T }
  | none => []
```

The `computeTypeSubst` function already exists and handles both annotated operators (via `opTypeSubst`) and unannotated ones (via argument type matching). After type checking, the `.op` node at the call site will have a concrete type annotation (e.g., `Sequence bv32 → int → bv32` for `Sequence.select`), so `opTypeSubst` will produce the correct substitution `[%a → bv32]`.

**However**, there's a subtlety: `PrecondElim` runs **before** type checking, so the `.op` node might not have a fully instantiated type annotation yet. The `computeTypeSubst` fallback uses argument types (`arg.typeOf`), which also might not be available pre-type-checking.

This means the fix might need to be in a different place — either:

1. **Move precondition type resolution to the type checker**: After type checking resolves all type variables, have the type checker apply the resolved substitution to precondition assertions. This is complex.

2. **Move `PrecondElim` to run after type checking** (or add a post-type-check fixup pass): The comment in `PrecondElim.lean` says it must run before type checking "since in the presence of polymorphic preconditions, the resulting assertions have type variables that must be unified." But the current approach of relying on the type checker to resolve them isn't working for all cases.

3. **Make the SMT encoder handle unresolved type variables gracefully**: When encountering an unresolved type variable in a precondition obligation, treat it as an uninterpreted sort. This is the simplest fix but might produce weaker verification results.

4. **Have `PrecondElim` use the callee's type annotation** to derive the type substitution: Even before type checking, the callee `.op` node might have a type annotation from the parser/elaborator. If `Sequence.select` is called as `Sequence.select(state_out, 0)`, the Boole translator might have set the type annotation on the `Sequence.select` operator to `Sequence bv32 → int → bv32`. In that case, `opTypeSubst` would work.

The most likely correct fix is **option 4** — use `LFunc.computeTypeSubst` (which calls `opTypeSubst`) in `collectWFObligations` to derive the type substitution from the callee's type annotation, then apply it to the precondition. This works because:
- The Boole/Core parser typically annotates `.op` nodes with instantiated types
- `opTypeSubst` unifies the instantiated type against the generic type to get the substitution
- For cases where the annotation isn't available (pre-type-check), the fallback uses argument types

For the polymorphic function body case (`Seq_lib_insert<T>`), the type substitution would map `%a → T` (or `%a → $__ty7` after type checking), which still leaves a type variable. For this case, the SMT encoder would need to handle it — either by treating the WF procedure's type parameters as uninterpreted sorts, or by not generating bounds-check obligations for calls within polymorphic function bodies (deferring them to concrete call sites).

## Two Failing Tests

1. **`sha256_compact_indexed.lean`** — The `#guard_msgs` expected output doesn't include the new `out-of-bounds access check` obligations. The test needs its expected output updated, AND the SMT encoding error needs to be fixed.

2. **`seq_slicing.lean`** — Same issue: the `seqOutOfBoundsSeed` test's `#guard_msgs` doesn't account for the new bounds-check obligation that now appears (with the SMT encoding error).

## Summary of Required Changes

1. **`Strata/DL/Lambda/Preconditions.lean`** — In `collectWFObligations`, apply the type substitution (from `LFunc.computeTypeSubst`) to the precondition expression after value substitution. This resolves type variables for concrete call sites.

2. **`Strata/Languages/Core/SMTEncoder.lean`** — For the polymorphic function body case (where type variables genuinely can't be resolved), either:
   - Treat unresolved type variables as uninterpreted sorts (add them to `ctx.sorts` and use a `constr` TermType), or
   - Skip encoding obligations that contain unresolved type variables (report as "unknown" rather than error)

3. **Test files** — After the fix, update the `#guard_msgs` expected output in both `sha256_compact_indexed.lean` and `seq_slicing.lean` to reflect the correct behavior (either the obligations pass, or they're reported differently).

PROgram52bc:
## Summary

- Fixes #1201 — `PrecondElim` no longer leaves `%a` in obligations from polymorphic `Sequence.*` preconditions, and the SMT encoder no longer fails on residual unresolved type variables.
- `Preconditions.lean`: derives a call-site type substitution from the actuals' types (with the `.op` annotation as fallback) and applies it to the substituted precondition expression.
- `SMTEncoder.lean`: encodes a leftover `.ftvar` as a fresh uninterpreted sort instead of returning `Unimplemented encoding for type var`. Verification reports `unknown` for those cases (sound, just weaker).

## Why this branch

The two failing tests (`sha256_compact_indexed.lean`, `seq_slicing`'s `seqOutOfBoundsSeed`) were added in this branch's Boole language-extensions work (#1075) and are what surfaced the bug — main's existing Sequence tests use Core's `Sequence.empty<T>()` syntax which annotates the `.op` with a concrete type and masks the bug. A backport PR against `main` will follow with the same source-file fixes.

## Why two changes

Most call sites (e.g. `Sequence.select(state_out, 0)` with `state_out : Sequence bv32`) get a concrete substitution from arg types and produce well-typed obligations. The encoder fallback covers the remaining cases the issue calls out: calls inside a polymorphic function body (`Seq_lib_insert<T>`) and calls whose every argument is itself polymorphic (`Sequence.select(Sequence.empty_int, 0)`).

## Test plan

- [x] `lake build StrataTest` — all 614 targets pass.
- [x] `StrataTest/DL/Lambda/PreconditionsTests.lean` — added two cases pinning the new behaviour: annotated `.op` path and argument-type fallback.
- [x] `StrataTest/Languages/Boole/FeatureRequests/seq_slicing.lean` — `seqOutOfBoundsSeed` `#guard_msgs` updated; the SMT encoding error is gone, replaced by `unknown`.
- [x] `StrataTest/Languages/Boole/FeatureRequests/sha256_compact_indexed.lean` — `#guard_msgs` updated to include the new bounds-check obligations (`unknown` where the procedure has no length precondition, `pass` where it does).
- [x] `lake test` — only pre-existing failure is the `ion-java` jar fixture, unrelated.


PROgram52bc:
Seems like a similar fix is needed for main, but the problem is surfaced only on main2 (by tests added in PR #1075). 
I think either cherry-picking the specific commit or merging main2 back to main should do it.

MikaelMayer (commented):
🤖🔍 Clean fix for the polymorphic precondition encoding issue. The two-pronged approach (resolve at the precondition level + graceful fallback in the encoder) is sound. One design concern about code duplication.

Strata/DL/Lambda/Preconditions.lean
```diff
@@ -53,6 +53,33 @@ def substitutePrecondition
...
+    let argConstraints := (args.zip fn.inputs.values).filterMap
+      (fun (arg, formal) => arg.typeOf.map (·, formal))
+    let argSubst :=
+      if argConstraints.isEmpty then none
+      else match Constraints.unify argConstraints SubstInfo.empty with
+        | .ok substInfo => some substInfo.subst
+        | .error _ => none
+    match argSubst with
+    | some s => s
+    | none => fn.opTypeSubst callee |>.getD Subst.empty
```
MikaelMayer: 🤖🔍 `callSiteTypeSubst` duplicates most of `LFunc.computeTypeSubst` (in `Factory.lean`) with the priority reversed (args first, `.op` second). Could this reuse `computeTypeSubst` with a flag or be expressed as `computeTypeSubst` with swapped branches? If the reversed priority is intentional and important (the docstring says so), at minimum a cross-reference comment pointing to `computeTypeSubst` would help future maintainers understand why two near-identical unification helpers exist.

PROgram52bc: Thanks! I added a cross-reference. Refactoring might need adapting existing proofs and I figured that could be deferred to another PR.

MikaelMayer (commented):

MikaelMayer (commented):
🤖🔍 All previous comments appear addressed — the cross-reference docstring on `callSiteTypeSubst` clearly explains the relationship with `computeTypeSubst` and why they remain separate. Reviewer sign-off still needed.

MikaelMayer (commented):

StrataTest/DL/Lambda/PreconditionsTests.lean
```diff
@@ -105,4 +105,43 @@ info: [WFObligation(safeDiv, (∀ (bvar:int) ((~Bool.Implies : (arrow bool (arro
...
+-- `Sequence.select`'s `0 <= i && i < Sequence.length(s : Sequence %a)`
+-- bound check, which uses an annotated `Sequence.length` op.
+private def polySelFunc : LFunc TestParams :=
+  { name := "polySel"
+    typeArgs := ["a"]
+    inputs := [("s", mty[Sequence %a])]
+    output := mty[%a]
+    -- precondition: ((~lenOf : (Sequence %a) → int) s)
+    -- The op carries a `%a` annotation that must be substituted at the call site.
+    preconditions := [⟨esM[((~lenOf : (Sequence %a) → int) s)], ()⟩]
```
atomb: This precondition doesn't seem to have type `bool`. Am I missing something? It seems like it would exercise the treatment of polymorphism, but not otherwise be a sensible precondition.

PROgram52bc: I think it is testing specifically on type substitution under polymorphism, trying to show that the `%a` type variable in the precondition gets substituted successfully (into `int` in the two examples below)? Without doing type substitution on the pre-conditions, this test should fail.

atomb: It could just as easily use a bool-typed example, though. Maybe `((~Int.Lt : (int → int → bool)) #0 ((~lenOf : (Sequence %a) → int) s))`? I think that might avoid confusion in the future.

PROgram52bc: Ah, yes. Having `bool` as precondition return type makes much more sense.

PROgram52bc: Updated the precondition to `0 < lenOf(s)` so it has type `bool`, while still exercising `%a` substitution on the inner `lenOf` op annotation. Pushed as 4329bb5ae.

atomb (approved):

MikaelMayer (commented):
🤖🔍 All previous comments appear addressed — the test precondition is now `bool`-typed (`0 < lenOf(s)`) as requested, while still exercising `%a` substitution on the inner annotated op. Reviewer sign-off still needed.

Strata/DL/Lambda/Preconditions.lean
```diff
@@ -53,6 +53,42 @@ def substitutePrecondition
...
```
tautschnig: The cross-reference docstring is good, but the structural duplication with `LFunc.computeTypeSubst` (in `Factory.lean`, lines 653–665) is real — same unification, opposite priority. Please open a follow-up issue tracking unification (e.g. `LFunc.computeTypeSubstWith (preferOp : Bool)` plus a thin wrapper for each call site) and reference it here, so this isn't lost. Once the proofs in `Semantics.lean` (`computeTypeSubst_of_opTypeSubst` and friends) are touched for an unrelated reason, that's the moment to consolidate.

Minor: the inner `match argSubst with | some s => s | none => fn.opTypeSubst callee |>.getD Subst.empty` is slightly clearer as `argSubst.getD ((fn.opTypeSubst callee).getD Subst.empty)`.

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -236,7 +236,15 @@ def LMonoTy.toSMTType (E: Env) (ty : LMonoTy) (ctx : SMT.Context) (useArrayTheor
...
-                    | _ => .error f!"Unimplemented encoding for type var {tyv}"
+                    | _ =>
+                      -- Unresolved type variable from a polymorphic precondition
+                      -- whose call-site type couldn't be inferred (e.g. inside a
+                      -- polymorphic function body, or with polymorphic args).
+                      -- Treat as a fresh uninterpreted sort so the obligation can
+                      -- still be encoded; the solver will report `unknown` rather
+                      -- than the encoder failing outright (issue #1201).
+                      let ctx := ctx.addSort { name := tyv, arity := 0 }
+                      .ok ((.constr tyv []), ctx)
```
tautschnig: Two follow-ups:

1. Sort name leakage: `tyv` (the raw `TyIdentifier`, e.g. `"a"` or `"$__ty5"`) flows directly into the SMT output as a sort name. If two unrelated polymorphic preconditions in the same encoding both have an unresolved `%a` that *aren't* meant to be the same type, declaring them under one shared sort is silently unsound (the solver would conflate them). After type inference these typically have distinct `$__tyN` names so this is fine in practice, but please at minimum prefix the introduced sort (e.g. `s!"__poly_{tyv}"`) so it's visibly synthetic in `.smt2` dumps and harder to alias accidentally with a user-declared sort of the same name.

2. Add an `IO.println`/diagnostic emit when this branch fires, behind the existing verbose flag, so users and downstream pipelines can see *which* obligations were weakened to `unknown` because of unresolved type variables. Right now the only signal is a missing pass/fail.

StrataTest/DL/Lambda/PreconditionsTests.lean
```diff
@@ -105,4 +105,52 @@ info: [WFObligation(safeDiv, (∀ (bvar:int) ((~Bool.Implies : (arrow bool (arro
...
+  esM[((~polySel : (Sequence int) → int) myseq)]
+
+-- Same expectation when the operator is unannotated: the argument-types
+-- fallback unifies `Sequence int` against `Sequence %a` to derive `[%a → int]`.
+/--
+info: [WFObligation(polySel, ((~Int.Lt : (arrow int (arrow int bool)))
+ #0
+ ((~lenOf : (arrow (Sequence int) int)) (myseq : (Sequence int)))), ())]
+-/
+#guard_msgs in
```
tautschnig, line 154: Could you also add a `#guard_msgs` case where the call has neither an `.op` annotation nor argument types (the path documented at line 64–68: "Returns `Subst.empty` … the resulting unsubstituted precondition will be handled downstream by the SMT encoder")? Right now the `Subst.empty` arm of `callSiteTypeSubst` is unreachable from the existing tests, and the encoder-fallback claim in the PR description is verified only by the heavier `seq_slicing` / `sha256_compact_indexed` tests — a 5-line unit test here would lock the contract in at the unit level.

tautschnig (commented):
Proof / test suggestions (none currently exist for the new helper):
- `theorem callSiteTypeSubst_agrees_with_computeTypeSubst_when_args_unannotated (fn callee args) (h_unannot : ∀ a ∈ args, a.typeOf = none) : callSiteTypeSubst fn callee args = (fn.opTypeSubst callee).getD .empty` — pins the contract that the only behavioural divergence from `computeTypeSubst` is priority, not result.
- `theorem callSiteTypeSubst_typeArgs_eliminated (fn callee args precond) : (∀ x ∈ fn.typeArgs, x ∉ (precond.applySubst (callSiteTypeSubst fn callee args)).freeTyVars)` (under preconditions on `args`/`callee`) — the actual property the encoder is currently relying on; today only enforced at runtime by the `.error _` fallback inside `Constraints.unify`.
- The encoder fallback turns an unresolved `%a` into a fresh uninterpreted sort `tyv`. A small theorem `LMonoTy.toSMTType_ftvar_fallback_sound : … → smt model satisfies obligation → original semantics also satisfied` would formalise the soundness claim already in the PR description.

Test gaps (cheap to fill in this PR):
- Both new `#guard_msgs` cases instantiate `%a` to `int`. Add at least one `bv32`/`bool` instantiation so a future refactor that hard-codes `Sequence int` somewhere is caught.
- No test exercises the encoder's `LMonoTy.toSMTType` fallback path directly (the cases in the PR all resolve via `callSiteTypeSubst`). A small unit test invoking `LMonoTy.toSMTType` on a `(.ftvar _)` against an empty `tySubst` would lock in the new "fresh sort" behaviour.

MikaelMayer (dismissed):

Strata/DL/Lambda/Preconditions.lean
PROgram52bc: > Please open a follow-up issue tracking unification (e.g. LFunc.computeTypeSubstWith (preferOp : Bool) plus a thin wrapper for each call site) and reference it here, so this isn't lost.

Will do!

> the inner match argSubst with | some s => s | none => fn.opTypeSubst callee |>.getD Subst.empty is slightly clearer as argSubst.getD ((fn.opTypeSubst callee).getD Subst.empty)

Thanks! This is cleaner indeed. Will adapt it in a following commit.

StrataTest/DL/Lambda/PreconditionsTests.lean
PROgram52bc: Yes.

MikaelMayer (commented):
🤖🔍 All previous comments appear addressed:

- `callSiteTypeSubst` / `computeTypeSubst` unification tracked in #1298 ✅
- Sort name prefixed with `__poly_` to avoid aliasing ✅
- Nested match simplified to `Option.getD` ✅
- `#guard_msgs` test for the `Subst.empty` arm added ✅

Minor note: the verbose diagnostic suggestion (item 2 on SMTEncoder.lean — emit a message behind the verbose flag when the unresolved-type-variable branch fires) was not addressed in this commit. This seems fine as a follow-up rather than a blocker.

Strata/DL/Lambda/Preconditions.lean
PROgram52bc: Both addressed in 69ad473d7:

- Filed #1298 to track unifying `callSiteTypeSubst` with `LFunc.computeTypeSubst` once the `Semantics.lean` proofs are touched for an unrelated reason, per your suggestion.
- Inlined the `argSubst.getD ((fn.opTypeSubst callee).getD Subst.empty)` form.

Strata/Languages/Core/SMTEncoder.lean
PROgram52bc: Re: 1 (sort name leakage) — addressed in 69ad473d7. Synthetic sorts now declare as `__poly_{tyv}`, so `(declare-sort __poly_a 0)` rather than `(declare-sort a 0)`. Visibly distinct from any user-declared sort with a colliding identifier.

Re: 2 (verbose diagnostic emit) — pushing back on this one. The `unknown` outcome is already observable per-obligation in solver output, and there is no existing verbose-flag plumbing in `LMonoTy.toSMTType`'s call path that this could thread through cleanly. I'd rather not add either a new flag or a hardcoded `IO.println` in this PR; happy to revisit if a downstream consumer needs the signal.

StrataTest/DL/Lambda/PreconditionsTests.lean
PROgram52bc: Added in 69ad473d7. The new `#guard_msgs` case at the bottom of the file calls `polySel` with neither an `.op` annotation nor typed arguments, exercising the `Subst.empty` arm and locking in that the unsubstituted `%a` is preserved in the obligation for the encoder to handle.

PROgram52bc:
Thanks for the suggestions — recording our take so they don't get lost:

**Theorems**: deferring all three (`callSiteTypeSubst_agrees_with_computeTypeSubst_when_args_unannotated`, `callSiteTypeSubst_typeArgs_eliminated`, `LMonoTy.toSMTType_ftvar_fallback_sound`) for this PR. The first two will become natural once #1298 unifies the helpers; until then, proving them on a soon-to-be-removed surface adds maintenance cost. The soundness theorem on the encoder fallback is more interesting standalone — opening a separate issue for it would be the right call if reviewers want to track it.

**Test gaps**: also deferring. Both new `#guard_msgs` cases run through `LMonoTy.toSMTType`'s `int` branch only because the unification in `callSiteTypeSubst` resolves `%a` to whatever the call site provides; rerunning the same tests with `bv32` or `bool` exercises the same code path with different leaves, so the marginal coverage is small. A direct `LMonoTy.toSMTType .ftvar` unit test would lock in the new fresh-sort behaviour, but with the `__poly_` prefix added in 69ad473d7 the contract is already harder to break silently. Happy to revisit either in a follow-up PR if you'd prefer.

atomb (dismissed):

StrataTest/DL/Lambda/PreconditionsTests.lean
```diff
@@ -105,4 +105,63 @@ info: [WFObligation(safeDiv, (∀ (bvar:int) ((~Bool.Implies : (arrow bool (arro
...
+#eval collectWFObligations polyFactory
+  esM[(~polySel (myseq : (Sequence int)))]
+
+-- Subst.empty arm: neither `.op` annotation nor typed arguments. The
+-- `callSiteTypeSubst` documentation (lines 63–66) promises this case returns
```
joscoh: I don't think you need to mention the documentation for the function here. I would just give the expected result.

Strata/Languages/Core/SMTEncoder.lean
```diff
@@ -236,7 +236,18 @@ def LMonoTy.toSMTType (E: Env) (ty : LMonoTy) (ctx : SMT.Context) (useArrayTheor
   | .ftvar tyv => match ctx.tySubst.find? tyv with
                     | .some termTy =>
                       .ok (termTy, ctx)
-                    | _ => .error f!"Unimplemented encoding for type var {tyv}"
+                    | _ =>
+                      -- Unresolved type variable from a polymorphic precondition
+                      -- whose call-site type couldn't be inferred (e.g. inside a
+                      -- polymorphic function body, or with polymorphic args).
+                      -- Treat as a fresh uninterpreted sort so the obligation can
```
joscoh: What is the soundness argument for this? What happens if this is e.g. a polymorphic function with multiple callers at different types? Can the SMT solver ever raise an error because of this? I am not sure silently swallowing the error and reporting unknown (or maybe a solver error?) is better than an explicit error here.

PROgram52bc: Thanks for your insights and questions, Josh.

> What is the soundness argument for this? ... Can the SMT solver ever raise an error because of this?

Encoding the polymorphic type as an unknown sort (as done here) won't necessarily have soundness concerns under deductive mode (e.g., the solver issuing spurious `unsat` when satisfiable). But under bug finding mode, this may indeed lead to spurious satisfiability thus reporting "there is no bug" falsely. I will make another commit shortly making it raise a hard error for now, but I think below are some thoughts we can keep in mind beyond just this PR.

> What happens if this is e.g. a polymorphic function with multiple callers at different types?

I think in that scenario, each call should have their separate proof obligations, which won't interfere with each other. Please correct me if I'm wrong.

> I am not sure silently swallowing the error and reporting unknown (or maybe a solver error?) is better than an explicit error here.

The current encoding actually has one advantage over raising explicit errors: with polymorphic function bodies, where we do want an unknown sort on the SMT level to represent "any type". For example, in
```
  function first<T>(s: Sequence T) : T
    requires Sequence.length(s) >= 1;
  {
    Sequence.select(s, 0)
  }
```
`Sequence.length` here has a polymorphic type `Sequence.length(s : Sequence %a)`, and after subtitution, it gets another polymorphic type `T`, encoded as something like `__poly_T`, which effectively represents "for any type." If we raise an explicit error instead, the above example will be treated as invalid, because type `T` is still considered "unresolved" after type substitution. But there is no test case for that scenario right now, and if we want this feature, we should at least limit it to deductive mode only, still raising an error on bug finding mode.

StrataTest/DL/Lambda/PreconditionsTests.lean
PROgram52bc: Agreed. I have just removed that part.

MikaelMayer (commented):
🤖🔍 All previous comments appear addressed. The new commit replaces the `__poly_` uninterpreted-sort fallback with a hard error — this is a sound design change (the commit message explains why the fallback was unsound for bugFinding mode). The docstring in `Preconditions.lean` is updated to match the new behavior, and the cross-reference comment on `callSiteTypeSubst` is preserved. The verbose-diagnostic note from the prior review is now moot since the branch returns an explicit error. Reviewer sign-off still needed.

atomb (approved):

MikaelMayer (commented):
🤖🔍 All previous comments appear addressed — the latest commit is a merge from main2 bringing in unrelated changes (#1318 conjunction/precond fix, #1333 doc cleanups, etc.). The PR's own logic (callSiteTypeSubst with cross-reference docstring, Option.getD simplification, #guard_msgs test for the Subst.empty arm, and the hard error for unresolved type variables in SMTEncoder) remains intact. Reviewer sign-off still needed.

PROgram52bc:
Bug still present on mainline, not merged yet. Mirroring [CR](https://code.amazon.com/reviews/CR-296536028) on mainline.

PROgram52bc:
tracked by https://code.amazon.com/reviews/CR-296536028

PR #1206, opened 2026-05-21
pyspec: extend predicate bodies with strict comparisons, equality, and integer arithmetic

julesmt:
Extends the PySpec predicate-body recognizer (`transExpr` / `transCompare`)
so common assertion shapes translate to precise SMT obligations instead
of silently dropping to a placeholder.

1. **Strict comparisons and equality (`<` `>` `==` `!=`).** The recognizer
   previously enumerated only `<=` / `>=` for the general case (with a
   `subject == "literal"` short-circuit reused as enum membership). All
   other comparisons fell through to `unsupported comparison`. Also
   loosens `makeComparison` so var-to-var comparisons (`x != y`,
   `x < y` between two parameters) translate, not just literal-bound
   ones.

2. **Integer arithmetic (`+` `-` `*` `//` `%`).** `transExpr` previously
   hit `unsupported BinOp` for any arithmetic operator, dropping
   expressions like `kw["x"] + kw["y"] >= 0` to a placeholder.

After this PR, predicates like `assert x > 0`, `assert result != x`,
and `assert kw["x"] + kw["y"] >= 0` translate end-to-end (PySpec →
DDM → Laurel → Core).

### DIFF
--- a/Strata/Languages/Python/PythonLaurelTypedExpr.lean
+++ b/Strata/Languages/Python/PythonLaurelTypedExpr.lean
@@ -102,12 +102,21 @@ def intMul (x y : TypedStmtExpr .TInt)
     (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TInt :=
   .ofStmt (.PrimitiveOp .Mul [x.stmt, y.stmt]) source
 
-/-- Python `//` is floor (Euclidean) division → Laurel `Div`. -/
+/-- Lowers Python `//` to Laurel `Div`. Laurel `Div` is Euclidean
+    division (`Int.ediv`, non-negative remainder); Python `//` is floor
+    division (rounds toward `-∞`). The two coincide only when the
+    divisor is positive — e.g. `-7 // -2` is `3` in Python but `4` in
+    Euclidean. The recognizer in `Specs.lean` emits a warning when the
+    divisor's sign cannot be proven positive. -/
 def intFloorDiv (x y : TypedStmtExpr .TInt)
     (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TInt :=
   .ofStmt (.PrimitiveOp .Div [x.stmt, y.stmt]) source
 
-/-- Python `%` is Euclidean modulus → Laurel `Mod`. -/
+/-- Lowers Python `%` to Laurel `Mod`. Laurel `Mod` is Euclidean
+    modulus (always non-negative); Python `%` matches the sign of the
+    divisor. They coincide only when the divisor is positive. The
+    recognizer in `Specs.lean` emits a warning when the divisor's sign
+    cannot be proven positive. -/
 def intMod (x y : TypedStmtExpr .TInt)
     (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TInt :=
   .ofStmt (.PrimitiveOp .Mod [x.stmt, y.stmt]) source
--- a/Strata/Languages/Python/PythonLaurelTypedExpr.lean
+++ b/Strata/Languages/Python/PythonLaurelTypedExpr.lean
@@ -74,6 +74,22 @@ def intLeq (x y : TypedStmtExpr .TInt)
     (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
   .ofStmt (.PrimitiveOp .Leq [x.stmt, y.stmt]) source
 
+def intGt (x y : TypedStmtExpr .TInt)
+    (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
+  .ofStmt (.PrimitiveOp .Gt [x.stmt, y.stmt]) source
+
+def intLt (x y : TypedStmtExpr .TInt)
+    (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
+  .ofStmt (.PrimitiveOp .Lt [x.stmt, y.stmt]) source
+
+def intEq (x y : TypedStmtExpr .TInt)
+    (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
+  .ofStmt (.PrimitiveOp .Eq [x.stmt, y.stmt]) source
+
+def intNe (x y : TypedStmtExpr .TInt)
+    (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
+  .ofStmt (.PrimitiveOp .Neq [x.stmt, y.stmt]) source
+
 def realGeq (x y : TypedStmtExpr .TReal)
     (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TBool :=
   .ofStmt (.PrimitiveOp .Geq [x.stmt, y.stmt]) source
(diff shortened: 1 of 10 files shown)


Strata/Languages/Python/PythonLaurelTypedExpr.lean
```diff
@@ -74,6 +74,45 @@ def intLeq (x y : TypedStmtExpr .TInt)
...
+  .ofStmt (.PrimitiveOp .Add [x.stmt, y.stmt]) source
+
+def intSub (x y : TypedStmtExpr .TInt)
+    (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TInt :=
+  .ofStmt (.PrimitiveOp .Sub [x.stmt, y.stmt]) source
+
+def intMul (x y : TypedStmtExpr .TInt)
+    (source : Option FileRange := x.stmt.source) : TypedStmtExpr .TInt :=
+  .ofStmt (.PrimitiveOp .Mul [x.stmt, y.stmt]) source
+
```
tautschnig, line 104: The doc-comment is wrong: Python `//` is **floor** division (rounds toward −∞), but Laurel `PrimitiveOp.Div` resolves to `Int.ediv` (Euclidean — non-negative remainder). They coincide only when the divisor is positive.

Example: `-7 // -2` is `3` in Python (floor of `3.5`) but `4` in Euclidean (since `-7 = -2·4 + 1`). Same divergence for `intMod` at line 110: Python `%` matches the sign of the divisor, Euclidean `mod` is always non-negative.

This is pre-existing behavior inherited from `Strata/Languages/Python/PythonToCore.lean:180`, but predicate bodies will silently produce wrong VCs whenever a divisor can be negative. Two options:

1. Tighten the doc-comment to say "matches Python only when the divisor is positive" and file an issue to add a true floor-div op to Laurel/Core.
2. Emit a `specWarning` in `Specs.lean:952` when the divisor sign isn't a positive literal.

Minimum: don't claim the semantics is the same in the doc.

Strata/Languages/Python/Specs.lean
```diff
@@ -802,15 +802,23 @@ private def makeComparison
...
   else if boundType.isIntType then
     match bound with
     | .intLit .. => some (intCtor subj bound)
-    | _ => none
+    | _ =>
+      if subjType.isIntType || subjType.isAnyType then some (intCtor subj bound) else none
+  else if subjType.isAnyType && boundType.isAnyType then
+    -- Both sides Any-typed (e.g. two function parameters whose types
+    -- aren't seeded). Trust the user and use the int comparison.
+    some (intCtor subj bound)
```
tautschnig: Silently defaulting to `intCtor` when both operands are `Any`-typed (e.g., two unparameterized callees) is unsound if the actual runtime values are floats — the resulting SMT VC will have the wrong sort. Please add a `specWarning` here so users see that the recognizer is making a guess:

```lean
else if subjType.isAnyType && boundType.isAnyType then
  specWarning loc s!"comparison between Any-typed operands; assuming int"
  some (intCtor subj bound)
```

Or restrict the relaxation to only `==`/`!=` (where the cross-type case is more obviously a Python error).

Strata/Languages/Python/Specs.lean
```diff
@@ -916,6 +934,29 @@ partial def transExpr (e : expr SourceRange)
...
+  -- Integer arithmetic in predicate bodies. Both sides must translate
+  -- cleanly to int-typed (or Any-typed, e.g. function parameters whose
+  -- types weren't seeded into `localTypes`) SpecExprs; the result is
+  -- int-typed. Anything other than int / Any (float, str, …) falls
+  -- through to placeholder so transExpr doesn't claim a type it can't
+  -- justify.
+  | .BinOp _ left op right =>
+    let (lhs, lhsTp) ← transExpr left
+    let (rhs, rhsTp) ← transExpr right
+    let okType (tp : SpecType) : Bool := tp.isIntType || tp.isAnyType
```
tautschnig, line 959: Asymmetric coverage: this PR adds int `+`/`-`/`*`/`//`/`%` but no float arithmetic. A predicate `assert kw["x"] + kw["y"] >= 0.0` with float-typed operands still drops to `placeholder`. Either mirror the int branch for floats (the constructors are already in `PythonLaurelTypedExpr.lean`), or note in the PR description that float arithmetic is deliberately deferred and file a follow-up.

Strata/Languages/Python/Specs/Decls.lean
```diff
@@ -488,10 +490,25 @@ inductive SpecExpr where
...
+/-- Python `%` (modulus). -/
+| intMod (lhs rhs : SpecExpr) (loc : SourceRange)
 /-- A floating-point literal, stored as a string to preserve precision. -/
 | floatLit (value : String) (loc : SourceRange)
 | floatGe (subject : SpecExpr) (bound : SpecExpr) (loc : SourceRange)
 | floatLe (subject : SpecExpr) (bound : SpecExpr) (loc : SourceRange)
+| floatGt (subject : SpecExpr) (bound : SpecExpr) (loc : SourceRange)
+| floatLt (subject : SpecExpr) (bound : SpecExpr) (loc : SourceRange)
+| floatEq (lhs : SpecExpr) (rhs : SpecExpr) (loc : SourceRange)
+| floatNe (lhs : SpecExpr) (rhs : SpecExpr) (loc : SourceRange)
```
tautschnig, line 522: Refactor opportunity (out of scope for this PR): each new operator forces synchronized edits in 8 places — `SpecExpr` constructor, `softBEq`, DDM `op`, `toDDM`, `fromDDM`, `specExprToLaurel`, `PythonLaurelTypedExpr` def, and the recognizer. With 13 cases added here, the boilerplate is already painful and the next operator will be more so.

Worth filing a follow-up to collapse to two parameterized constructors:
```lean
| intBinOp   (op : IntBinOp) (lhs rhs : SpecExpr) (loc : SourceRange)   -- Add/Sub/Mul/Div/Mod
| intCompare (op : IntCmp)   (lhs rhs : SpecExpr) (loc : SourceRange)   -- Eq/Ne/Lt/Le/Gt/Ge
```
A round-trip property `(toDDM ∘ fromDDM) e = e` becomes a one-liner per family rather than a per-constructor case. The proof obligation pre-/post-refactor is itself a useful regression test.

StrataTestExtra/Languages/Python/SpecsTest.lean
```diff
@@ -292,6 +292,61 @@ meta def warningTestCase : IO Unit := withPython fun pythonCmd => do
...
+      match r with
+      | .ok (sigs, warnings) =>
+        if sigs.isEmpty then
+          throw <| IO.userError "Expected signatures from predicate_arith.py but got none"
+        if !warnings.isEmpty then
+          let warnStr := warnings.foldl (init := "") fun acc w => s!"{acc}\n  {w}"
+          throw <| IO.userError s!"Unexpected warnings from predicate_arith.py:{warnStr}"
+      | .error e =>
+        throw <| IO.userError e
+
```
tautschnig, line 349: Both new test cases assert only "no warnings produced". A regression in `specExprToLaurel` (e.g., wrong constructor, wrong type, swapped operand order) is invisible to these tests. Please add one fixture per family that runs the full pipeline (PySpec → DDM → Laurel → Core) and pins the resulting SMT VC shape — at minimum, an `assert kw["x"] + kw["y"] >= 0` whose generated obligation includes `(>= (+ x y) 0)` (or similar) verified via `#guard_msgs`. Without this, the round-trip soundness is being checked only by absence of warnings.

PR #1208, opened 2026-05-21
pyspec: add --reject-approximations CLI flag

julesmt:
Adds `--reject-approximations`. Today some constructs silently lower to a `Hole` or `.placeholder`, so the verifier can report `Analysis success` on code whose contract was dropped. The flag turns those sites into hard errors. Off by default.


tautschnig:

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -224,12 +255,18 @@ def manglePythonMethod (className : String) (methodName : String) : String :=
   className ++ "@" ++ methodName
 
 /-- Build a StaticCall for an instance method: ClassName@methodName(self, args...).
-    For Any-typed receivers (no model available), returns a Hole instead. -/
-def mkInstanceMethodCall (className : String) (methodName : String)
+    For Any-typed receivers (no model available), returns a Hole under lax
+    mode, or raises `unsupportedConstruct` under `--reject-approximations`. -/
+def mkInstanceMethodCall (rejectApproximations : Bool) (className : String) (methodName : String)
```
tautschnig: After this PR, `mkInstanceMethodCall` is called from exactly two sites (the `with`-statement `__enter__`/`__exit__` pair at lines 2059–2060), both of which pass `ctx.rejectApproximations`. Threading the bool explicitly is noise — please take the whole `ctx : TranslationContext` so the signature shrinks and future additions can read other fields too.

Also: the inline `|>.map (fun h => { h with source := source })` in the `Any` branch is a workaround because `rejectableHole` ignores `source`. Cleaner to add a `source : Option FileRange := none` argument to `rejectableHole` itself and let it stamp the location in one place — the rejection error then carries the same `source` you already have, which makes the diagnostic much more useful under `--reject-approximations`.

Strata/Languages/Python/PythonToLaurel.lean
```diff
@@ -207,6 +212,32 @@ def stmtExprToVar (e : StmtExprMd) : Except TranslationError VariableMd :=
...
+/-- Same as `rejectableHole` but for sites that silently drop a Python
+    statement (e.g., `raise`, `try…else`, `for…else`) under lax mode.
+    The error message names the dropped construct. -/
+def rejectableDrop (rejectApproximations : Bool) (construct astRepr : String)
+    : Except TranslationError Unit :=
+  if rejectApproximations then
+    throw (.unsupportedConstruct
+      s!"[approximation] {construct} is silently dropped today; not in the sound subset"
+      astRepr)
+  else
```
tautschnig, line 258: Two callers pass `(toString (repr e))` for `astRepr`. For deeply nested expressions this produces multi-kilobyte error strings — fine for unit tests, painful for users running on real specs. Consider taking `(source : Option FileRange)` instead and letting the diagnostic infrastructure render `file:line:col` plus a short construct name; the surface form is rarely useful when the location is known.

If you want to keep the AST repr as a fallback, capping it (`astRepr.take 200 ++ "…"`) would prevent the worst cases without losing the unit-test guarantees.

Strata/Languages/Python/Specs.lean
```diff
@@ -298,9 +303,15 @@ instance : PySpecMClass PySpecM where
...
-    let w : PipelineMessage := { file, loc, phase := pySpecParsingPhase, kind := .pySpecParsingWarning, message }
-    modify fun s => { s with warnings := s.warnings.push w }
+    let ctx ← read
+    -- Under strict mode (`--reject-approximations`), every warning is an
+    -- error: the parser only emits warnings on patterns it falls back to
+    -- `.placeholder` for, which silently drops the user's spec.
+    if ctx.rejectApproximations then
+      specErrorAt ctx.pythonFile loc s!"[approximation] {message}"
+    else
+      let w : PipelineMessage := { file := ctx.pythonFile, loc, phase := pySpecParsingPhase, kind := .pySpecParsingWarning, message }
```
tautschnig, line 314: Same logical change as `rejectableHole`/`rejectableDrop` in `PythonToLaurel.lean`, but expressed differently (warning→error rewrite vs. helper-with-throw). With `rejectApproximations` now in both `TranslationContext` and `PySpecContext`, please factor the message format into a single helper module (e.g. `Strata/Languages/Python/Approximation.lean`) so the prefix `[approximation] ` is only typed once and the three current call shapes can't drift.

No regression test currently fires this `specError` arm — please add one alongside the existing `Specs/...` tests so the warning-to-error promotion is exercised.

tautschnig (commented):
Proof / `#guard` opportunities (none currently exist):
- `theorem rejectableHole_strict (c r) : (rejectableHole true c r).isOk = false` and the dual `…false…isOk = true` — trivial `decide`-able lemmas (or `#guard`s) that lock in the boolean dispatch.
- `theorem rejectableHole_strict_message (c r) : ∃ msg ast, rejectableHole true c r = .error (.unsupportedConstruct msg ast) ∧ msg.startsWith "[approximation] "` — pins the message prefix so downstream tooling that filters by `[approximation]` can rely on it.

Minor stylistic nits:
- `"silently dropped today"` — the *today* in `rejectableDrop`'s message is informal and dates the wording. Either drop it or replace with `"by current translation"`.
- The two helper messages and `Specs.lean`'s `specWarning` rewrite each have slightly different shapes (`is approximated as Hole`, `is silently dropped today`, `[approximation] {message}`). Once you have an end-to-end test from point 1, please consider unifying via a small `approximationError : (kind : ApproxKind) → (construct : String) → String` helper.

PR #1209, opened 2026-05-21
Fix Boole tests after merge

joscoh:
This is a temporary PR to fix the Boole tests after the merge with `main`. This PR does 3 main things:
1. Makes the sequence functions in `sha256_compact_indexed` monomorphic, since the SMT backend does not support polymorphism (resulting from the newly produced preconditions for sequence indexing)
2. Adds invariants and preconditions to allow the SHA256 example to verify
3. Fixes a bug where the type of `Sequence.empty` was lost in the translation to Core, which also resulted in polymorphic precondition proof obligations that hit errors. For example, `v:=Sequence.select(Sequence.empty_int, 0);` generated `0 < Sequence.length (Sequence.empty)`, which lost the information that `Sequence.empty` was instantiated at type `int`.

### DIFF
--- a/Strata/Languages/Boole/Verify.lean
+++ b/Strata/Languages/Boole/Verify.lean
@@ -469,9 +469,11 @@ partial def toCoreExpr (e : Boole.Expr) : TranslateM Core.Expression.Expr := do
       let intSub : Core.Expression.Expr := .op () ⟨"Int.Sub", ()⟩ none
       return mkCoreApp Core.seqTakeOp
         [mkCoreApp Core.seqDropOp [s', lo'], mkCoreApp intSub [hi', lo']]
-  -- Typed empty-sequence constant (Sequence.empty for bv32; other types can be added when needed).
-  | .seq_empty_bv8 _ | .seq_empty_bv16 _ | .seq_empty_bv32 _
-  | .seq_empty_bv64 _ | .seq_empty_int _ => return Core.seqEmptyOp
+  | .seq_empty_bv8 _  => return Core.seqEmptyOp (some (.bitvec 8))
+  | .seq_empty_bv16 _ => return Core.seqEmptyOp (some (.bitvec 16))
+  | .seq_empty_bv32 _ => return Core.seqEmptyOp (some (.bitvec 32))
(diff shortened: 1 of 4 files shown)


atomb (approved):

shigoel (approved):

Strata/Languages/Boole/Verify.lean
PROgram52bc, line 482: I think here it still uses the default `Core.seqEmptyOp` as an initial value, and could lead to missing type information if `vals` is empty.

StrataTest/Languages/Boole/FeatureRequests/sha256_compact_indexed.lean
```diff
@@ -27,19 +27,21 @@ program Boole;
...
- function Seq_lib_sort_by<T> (s : Sequence T, less : T -> T -> bool) : Sequence T;
- function Seq_lib_to_set<T> (s : Sequence T) : Set T;
- function Set_finite<T> (s : Set T) : bool;
+ function Seq_new (len : nat, f : int -> bv32) : Sequence bv32;
+ function Seq_lib_map (s : Sequence bv32, f : int -> bv32 -> bv32) : Sequence bv32;
+ function Seq_lib_map_values (s : Sequence bv32, f : bv32 -> bv32) : Sequence bv32;
+ function Seq_lib_filter (s : Sequence bv32, p : bv32 -> bool) : Sequence bv32;
+ function Seq_lib_sort_by (s : Sequence bv32, less : bv32 -> bv32 -> bool) : Sequence bv32;
+ function Seq_lib_to_set (s : Sequence bv32) : Set bv32;
+ function Set_finite (s : Set bv32) : bool;
```
PROgram52bc, line 44: Since they are no longer polymorphic, maybe renaming to something like `Set_finite_bv32` (and similarly for the above) makes more sense?

PROgram52bc (commented):
I left some hand-written comments based on agent feedbacks.

atomb:
Since @joscoh is out today and owns this PR, we agreed to merge this one first and then address the two comments above either as an amendment to #1205 or a second PR right after it.

PR #1215, opened 2026-05-26
Support transparent procedures

keyboardDrummer:
Base PR for:
- https://github.com/keyboardDrummer/Strata/pull/28

### Changes
- Replace a panic in SMTUtils with a `.error`
- Change handling of free postconditions in Core so they do not generate trivial VCs. This is useful in this PR because this PR is adding new usages of free postconditions and this way existing tests don't get additional "passing" VCs. A side-effect of this change is that existing tests have fewer reported passing VCs.
- Improve usage of source locations in the lifting phase, which improves diagnostics
- For Laurel procedures without a source body, use free postconditions instead of assuming the non-free postconditions in a generated body.
- Add the transparency pass. for each non-functional Core procedure, generate a function with the same signature and name suffixed with `$asFunction`. If a Core procedure is marked as transparent, attempt to generate a functional body for the `$asFunction`, where assertions are erased and all calls are to functional versions. Tie the functional version to the procedure using a free postcondition.
- Refactored CoreGroupingAndOrdering to compute SCC ordering for functions and procedures separately. This is required because TransparencyPass creates a new AST type, `UnorderedCoreWithLaurelTypes`, where functions and procedures are already separated.

### Testing
- Updated T20_TransparentBody to account for new feature. In a follow-up we will have many more tests, because we'll convert all functions in tests to procedures.
- Added T20_TransparentBodyError to capture some not supported cases


Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -211,18 +221,21 @@ def translateWithLaurel (options : LaurelTranslateOptions) (program : Program)
...
+    emit "CoreWithLaurelTypes" "core.st" coreWithLaurelTypes
     let initState : TranslateState := { model := model, overflowChecks := options.overflowChecks }
     let (coreProgramOption, translateState) :=
-      runTranslateM initState (translateLaurelToCore options program ordered)
-    if let some coreProgram := coreProgramOption then
-      emit "CoreProgram" "core.st" coreProgram
-    let mut allDiagnostics := passDiags ++ translateState.diagnostics
+      runTranslateM initState (translateLaurelToCore options program coreWithLaurelTypes)
+    -- Because of the duplication between functions and procedures, this translation is liable to create duplicate diagnostics
+    -- User errors should be checked in an earlier phase, and all dumb translation errors are Strata bugs
```
joscoh: What does this mean?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -346,9 +347,10 @@ private def exprAsUnusedInit (expr : StmtExprMd) (md : Imperative.MetaData Core.
     : TranslateM (List Core.Statement) := do
   let coreExpr ← translateExpr expr
   let id ← freshId
+  let model := (← get).model
   let ident : Core.CoreIdent := ⟨s!"$unused_{id}", ()⟩
-  let tyVarName := s!"$__ty_unused_{id}"
-  let coreType := LTy.forAll [tyVarName] (.ftvar tyVarName)
+  let ty ← translateType (computeExprType model expr)
+  let coreType := LTy.forAll [] ty
```
joscoh, line 340: How do you know the type is never polymorphic?

Strata/Languages/Laurel/TransparencyPass.lean
```diff
@@ -0,0 +1,177 @@
...
+    | _ => e) expr
+
+/-- Adjust a datatype selector (destructor) name based on the `proof` flag.
+    Destructor names contain `..` (e.g. `IntList..head`, `IntList..head!`).
+    Tester names also contain `..` but start with `is` after the separator.
+    - `proof = true` → use safe selectors (strip `!` suffix)
+    - `proof = false` → use unsafe selectors (add `!` suffix) -/
+private def adjustSelectorName (name : String) : String :=
+  -- Only adjust destructor names (contain ".." but are not testers)
+  match name.splitOn ".." with
```
joscoh: Minor, but I would prefer having `isDestructorName` and `isTesterName` functions in `TypeFactory.lean` that do this check, to ensure that there aren't silent changes in Core naming that break this.

Strata/Languages/Laurel/TransparencyPass.lean
```diff
@@ -0,0 +1,177 @@
...
+        ⟨.StaticCall funcCallee args, e.source⟩
+      else
+        let newName := adjustSelectorName callee.text
+        ⟨ .StaticCall newName args, e.source⟩
+    | .PrimitiveOp operator arguments _ => ⟨ .PrimitiveOp operator arguments true, e.source⟩
+    | _ => e) expr
+
+/-- Build a free postcondition equating the procedure's output to its functional version.
+    For a procedure `foo(a, b) returns (r)`, produces:
+      `r == foo$asFunction(a, b)` -/
```
joscoh, line 89: IIUC, this pass constructs functions intended to be equivalent to the corresponding procedure and then assumes the equivalence. Is it possible to prove the equivalence instead?

joscoh (commented):
What happens with transparent recursive procedures? Does it produce a recursive function? If so, what arguments are given for the termination checker, and what happens if the generated function is nonterminating?

Strata/Languages/Laurel/TransparencyPass.lean
```diff
@@ -0,0 +1,177 @@
...
+private def mkFunctionCopy (asFunctionNames : List String) (proc : Procedure) : Procedure :=
+  let funcName := { proc.name with text := proc.name.text ++ "$asFunction", uniqueId := none }
+  let body := match proc.body with
+    | .Transparent b => .Transparent (rewriteCallsToFunctional asFunctionNames (stripAssertAssume b))
+    | .Opaque _ _ _ => .Opaque [] none []
+    | x => x
+  { proc with name := funcName, isFunctional := true, body := body, preconditions := [] }
+
+/-- Check whether a function copy has a body (i.e. the procedure was transparent). -/
+private def functionHasBody (proc : Procedure) : Bool :=
```
joscoh: Is this function ever called?

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -938,4 +942,50 @@ def resolve (program : Program) (existingModel: Option SemanticModel := none) :
...
+`UserDefined` type references. These additional types should not be necessary
+but they are because certain type references have incorrectly not been updated.
+-/
+def resolveUnorderedCore (uc : UnorderedCoreWithLaurelTypes)
+    (existingModel : Option SemanticModel := none)
+    (additionalTypes : List TypeDefinition := [])
+    : UnorderedCoreWithLaurelTypes × SemanticModel :=
+  let fnProgram := unorderedCoreToProgram uc additionalTypes
+  let fnResolveResult := resolve fnProgram existingModel
+  (fromResolvedProgram fnResolveResult.program, fnResolveResult.model)
```
joscoh, line 989: This silently drops any errors encountered from resolution. Is this intentional?

Strata/Languages/Laurel/TransparencyPass.lean
```diff
@@ -0,0 +1,177 @@
...
+private def rewriteCallsToFunctional (asFunctionNames : List String) (expr : StmtExprMd) : StmtExprMd :=
+  mapStmtExpr (fun e =>
+    match e.val with
+    | .StaticCall callee args =>
+      if asFunctionNames.contains callee.text then
+        let funcCallee := { callee with text := callee.text ++ "$asFunction", uniqueId := none }
+        ⟨.StaticCall funcCallee args, e.source⟩
+      else
+        let newName := adjustSelectorName callee.text
+        ⟨ .StaticCall newName args, e.source⟩
```
joscoh, line 83: This drops any metadata associated with the callee (e.g. source location).

Strata/Languages/Laurel/TransparencyPass.lean
```diff
@@ -0,0 +1,177 @@
...
+      if name.endsWith "!" then name else name ++ "!"
+  | _ => name  -- not a destructor name, leave unchanged
+
+/-- Rewrite StaticCall callees to their `$asFunction` versions,
+    but only for procedures whose names appear in `nonExternalNames`. -/
+private def rewriteCallsToFunctional (asFunctionNames : List String) (expr : StmtExprMd) : StmtExprMd :=
+  mapStmtExpr (fun e =>
+    match e.val with
+    | .StaticCall callee args =>
+      if asFunctionNames.contains callee.text then
```
joscoh, line 78: A `HashSet` would be more efficient, but for small programs this is OK

joscoh (commented):
It would also be good to see a test where a transparent procedure calls another transparent procedure, with the corresponding function calls

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -565,26 +568,27 @@ def translateProcedure (proc : Procedure) : TranslateM Core.Procedure := do
...
+      pure $ some r
+    | _ =>
+      pure none
 
   -- Translate postconditions for Opaque and Abstract bodies
   let postconditions : ListMap Core.CoreLabel Core.Procedure.Check ←
     match proc.body with
     | .Opaque postconds _ _ | .Abstract postconds =>
-        translateChecks postconds "postcondition"
+        translateChecks postconds s!"postcondition{(bodyStmts.getD []).length}" bodyStmts.isNone
```
MikaelMayer: 🤖🔍 The label `s!"postcondition{(bodyStmts.getD []).length}"` encodes the body statement count into the label base (e.g. `"postcondition5"`). This produces labels like `postcondition5_0`, `postcondition5_1` for multiple postconditions. This seems like a workaround to avoid label collisions — could you add a comment explaining why the body length is embedded in the label name?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -770,6 +770,7 @@ def translateLaurelToCore (options: LaurelTranslateOptions) (program : Program)
         emitDiagnostic $ diagnosticFromSource proc.name.source
```
MikaelMayer: 🤖🔍 This "Instance procedure ... is not yet supported" diagnostic is now emitted in two places: here and in `Resolution.lean` (the new loop at the top of `resolve`). This duplication is likely why `.eraseDups` was added to the pipeline. Consider removing one of the two emission sites rather than relying on deduplication downstream.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -211,18 +221,21 @@ def translateWithLaurel (options : LaurelTranslateOptions) (program : Program)
...
+    let mut allDiagnostics: List DiagnosticModel := passDiags ++ translateState.diagnostics.eraseDups;
```
MikaelMayer, line 237: 🤖🔍 Using `.eraseDups` to suppress duplicate diagnostics is a symptom, not a fix. The root cause is that the same diagnostic (e.g. "Instance procedure ... not yet supported") is emitted in both `Resolution.lean` and `LaurelToCoreTranslator.lean`. Fixing the duplication at the source would be cleaner than filtering after the fact.

Strata/Languages/Laurel/TransparencyPass.lean
```diff
@@ -0,0 +1,177 @@
...
+- Generate a function with the same signature, named `foo$asFunction`
+- If transparent, the function gets a functional body (assertions erased, calls to functional versions)
+- If the function has a body, add a free postcondition equating the procedure output to the function
+-/
+def transparencyPass (program : Program) : UnorderedCoreWithLaurelTypes :=
+  let (skipped, notSkipped) := program.staticProcedures.partition (fun p => p.body.isExternal ||
+    -- Skip functions until we introduce a contract pass,
+    -- which enables lifting procedure calls from contracts
+    p.isFunctional)
+  let asFunctionNames := notSkipped.map (fun p => p.name.text)
```
MikaelMayer: 🤖🔍 `mkFreePostcondition` returns `LiteralBool true` when `proc.outputs` is not a singleton. This means procedures with zero or multiple outputs silently get no free postcondition tying them to their functional version. The zero-output case is fine (no return value to equate), but the multi-output case could be supported with a conjunction of equalities. Is this intentional or a TODO?

MikaelMayer (commented):
🤖🔍 Overall this is a well-structured PR that introduces a clean intermediate representation (`UnorderedCoreWithLaurelTypes`) and a transparency pass that generates functional versions of procedures. The pipeline extension is logical and the test coverage is reasonable.

Positive notes:
- The `stripAssertAssume` / `rewriteCallsToFunctional` / `mkFreePostcondition` decomposition is clean and easy to follow.
- The change from `.assert` to `.assume` for free postconditions in `ProcedureEval.lean` is correct — free postconditions should not generate proof obligations.
- Replacing `panic!` with `.error` in `SMTUtils.lean` is a good improvement.

See inline comments for issues found.

StrataTest/Util/TestDiagnostics.lean
```diff
@@ -137,6 +137,10 @@ def testInputWithOffset (filename: String) (input : String) (lineOffset : Nat)
       IO.println s!"\nUnexpected diagnostics:"
       for diag in unmatchedDiagnostics do
         IO.println s!"  - Line {diag.start.line}, Col {diag.start.column}-{diag.ending.column}: {diag.message}"
+
+    if unmatchedExpectations.length == 0 && unmatchedDiagnostics.length == 0 then
```
MikaelMayer, line 144: 🤖🔍 This branch fires when `diagnostics.size != expectedErrors.length` but both `unmatchedExpectations` and `unmatchedDiagnostics` are empty — which means there are duplicate diagnostics matching the same expectations. The message `"Duplicate diagnostics: {repr diagnostics}"` dumps the full repr which could be very noisy. Consider printing just the count or the duplicate entries.

Strata/Languages/Laurel/TransparencyPass.lean
```diff
@@ -0,0 +1,177 @@
...
+      { proc with body := .Opaque (postconds ++ [freeCond]) impl modif }
+    | .Abstract postconds =>
+      { proc with body := .Abstract (postconds ++ [freeCond]) }
+    | .Transparent body =>
+      { proc with body := .Opaque [freeCond] (some body) [] }
+    | _ => proc
+
+/--
+Transparency pass: translate a Laurel program to the UnorderedCoreWithLaurelTypes IR.
+
```
MikaelMayer, line 131: 🤖🔍 `transparencyPass` always adds a free postcondition to transparent procedures (via `addFreePostcondition`), which promotes their `.Transparent` body to `.Opaque`. This means the original transparent body is now wrapped in an `Opaque` with an implementation — so the procedure is no longer "transparent" in the Laurel sense after this pass. This is fine architecturally (the function copy carries the transparent semantics), but the docstring on `addFreePostcondition` could note that this promotion is the expected behavior for transparent procedures.

atomb:
To optimize for reusability, I'd be happier to have a procedure -> function transformation exist at the Core level rather than the Laurel level. It will likely be useful for front-end languages that don't go through Laurel, too.

keyboardDrummer:
> What happens with transparent recursive procedures? 

It is handled the same way as recursive Laurel functions

> Does it produce a recursive function? 

Yes

> If so, what arguments are given for the termination checker, and what happens if the generated function is nonterminating?

It currently defaults to using the first argument that's a datatype as the decreasing argument. If that's not enough to prove the function terminates, then the user gets to see the Core error related to that. We will improve this when Laurel gets decreasing clauses. In a sense, this a not yet implemented topic that needs further design. It's not clear to me whether Laurel can rely on or use Core's termination checking features, or whether it needs to generate something separately. Can Core's function termination checking even be turned off ?

keyboardDrummer:
> To optimize for reusability, I'd be happier to have a procedure -> function transformation exist at the Core level rather than the Laurel level. It will likely be useful for front-end languages that don't go through Laurel, too.

That sounds good to me but it's not a trivial change. I suggest we leave it out of scope for this PR.

Right now the transparency pass benefits from sitting before the grouping and ordering pass. Moving it to after that pass would mean duplicating some of the logic from grouping and ordering. For example, grouping and ordering identifies which functions are recursive and puts them in a structure that Core accepts for that. Now the transparency pass doesn't need to think about that.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
keyboardDrummer: I will remove this line because it's about a desired state that is out of scope for this PR.

An approach to statically detectable program errors is to detect all of them during resolution, so none of the compilation passes would emit program error diagnostics. This approach makes sense from the standpoint of each compilation pass not changing the meaning of the program. It also prevents any error from appearing twice in case part of the program is duplicated. 

I think that approach to reporting program errors is good. However, usage of many unsupported features is currently only detected in the LaurelToCore translation, and this PR is not about changing that, so I'll leave it like it is for now.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
keyboardDrummer: I agree. I think the solution is not to report any user program errors in `LaurelToCoreTranslator.lean` and instead report all of them in Resolution.lean. I want to make that change, but I think it's better to do that in a separate PR since it's a rather orthogonal improvement.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: Right now Laurel doesn't have polymorphism. Would that prevent a polymorphic type from appearing here?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: I think this was a poor man's way of identifying which error came from where, when source locations weren't working. I will remove the weird part.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
keyboardDrummer: > This "Instance procedure ... is not yet supported" diagnostic is now emitted in two places: here and in Resolution.lean

Oops, sorry about that! I will remove it from here.

> This duplication is likely why .eraseDups was added to the pipeline.

No that relates to duplication of code from creating the equivalent functions.

Strata/Languages/Laurel/Resolution.lean
keyboardDrummer: It shouldn't create any errors that weren't there before, so from a user perspective this is OK. In a follow-up PR though I am using these currently ignored errors to detect bugs in this pipeline

Strata/Languages/Laurel/TransparencyPass.lean
keyboardDrummer: @keyboardDrummer-bot could you address this comment?

Strata/Languages/Laurel/TransparencyPass.lean
keyboardDrummer: @keyboardDrummer-bot could you use a HashSet ?

Strata/Languages/Laurel/TransparencyPass.lean
keyboardDrummer: Nice find. Fixed.

Strata/Languages/Laurel/TransparencyPass.lean
keyboardDrummer: Good question. With a semantics for both the functions and procedures, I guess we could prove the equivalence. Do you think such a proof has to be part of this PR? Laurel -> Core currently only has termination proofs.

Strata/Languages/Laurel/TransparencyPass.lean
keyboardDrummer: Nope, removed.

MikaelMayer (commented):

Specifically:
- The confusing `postcondition{bodyLength}` label was simplified to just `postcondition` ✅
- The duplicate "Instance procedure not yet supported" diagnostic emission was removed from `translateLaurelToCore` ✅
- The `.eraseDups` comment was updated to reflect the remaining (legitimate) reason for it ✅
- `adjustSelectorName` now operates on `Identifier` instead of `String`, preserving source locations ✅

The remaining comments (multi-output `mkFreePostcondition`, `addFreePostcondition` docstring, TestDiagnostics verbosity) were questions/suggestions that don't require code changes.

Strata/Languages/Laurel/TransparencyPass.lean
keyboardDrummer: Added an extra line to the comment

Strata/Languages/Laurel/TransparencyPass.lean
keyboardDrummer: Good question. I have left support for multi-out procedures out of scope for this PR, since it's a bit more work and it's hard to test since there's no support for assignment in transparent procedures yet.

I think the current implementation for `mkFreePostcondition` is OK for multi-out procedures in the sense that it's sound but incomplete.

MikaelMayer (commented):

The new commit adds the suggested docstring clarification to `addFreePostcondition` explaining that the opaqueness promotion is intentional since the function copy carries the transparent semantics. Also cleans up the unused `functionHasBody` helper.

MikaelMayer (commented):
🤖🔍 All previous comments appear addressed — reviewer sign-off still needed. No new commits since last review. The `asFunctionNames` List→HashSet optimization mentioned by the author is a self-identified improvement, not a blocker.

MikaelMayer (commented):

New commit `6f0d365` converts `asFunctionNames` from `List String` to `Std.HashSet String` (the optimization noted in the last review) and extracts `isSelectorName`/`isTesterName`/`isDestructorName` helpers into `TypeFactory.lean`, which `adjustSelectorName` now uses instead of inline string splitting. Clean refactoring.

MikaelMayer (commented):
🤖🔍 All previous comments appear addressed — reviewer sign-off still needed. New commits since last review are minor cleanup: removed unused `program` parameter from `translateLaurelToCore` and replaced `.length > 0` with `!.isEmpty`. No regressions to prior fixes.

atomb:
One more high-level comment I have is just about terminology: I see transparency as a proof tactic, not a property of a language construct. And so I see this PR as being more about producing functional representations of the semantics of procedures than about making procedure bodies transparent. How doing that transformation affects the visibility of their bodies is a question that different analysis implementations should be able to make different choices about.

keyboardDrummer:
> One more high-level comment I have is just about terminology: I see transparency as a proof tactic, not a property of a language construct. And so I see this PR as being more about producing functional representations of the semantics of procedures than about making procedure bodies transparent. How doing that transformation affects the visibility of their bodies is a question that different analysis implementations should be able to make different choices about.

Is this comment asking for something?

> I see transparency as a proof tactic, not a property of a language construct

The term "property of a language construct" is ambiguous to me. Whether a Laurel procedure's body is transparent or not does not affect how it behaves during concrete execution. Is that what you mean by that? In Laurel, verification is an advertised feature of the language so things like transparency get their own documentation, grammar and typed fields in the AST.

> And so I see this PR as being more about producing functional representations of the semantics of procedures than about making procedure bodies transparent.

The name "TransparencyPass" comes from this pass only creating functional bodies for procedures that have a body that's marked as being transparent. We could give it a more verbose name like `FunctionalizeTransparentBodies` if you like.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
joscoh: Makes sense, but maybe leave a comment to make sure this is updated if Laurel ever supports polymorphism?

Strata/Languages/Laurel/TransparencyPass.lean
joscoh: I don't think it necessarily has to be part of this PR, but I see two different ways to handle this:
1. At the meta-level, like you say, with a semantics for functions and procedures we could prove once-and-for-all that all functions/procedures constructed in this way satisfy the assumed postcondition
2. Per function and procedure generated, it may be possible to just construct the proof within Strata (though this adds additional proof obligations that you would not want to surface to the user) and validate the equivalence on a case-by-case basis.

(1) is the better long-term solution, but (2) is possible in principle today. The only reason I bring it up is that right now this essentially adds a bunch of axioms asserting that the translation truly produced equivalent functions/procedures, and it may not be obvious to the user what axioms are introduced or how trustworthy they should be considered.

joscoh (commented):
I'd still like to see the transparent-calling-transparent test I mentioned above, once that is added I will approve

MikaelMayer (commented):

New commits since last review:
- Added the polymorphism comment at `exprAsUnusedInit` as requested by the reviewer ✅
- Removed unused `program` parameter from `translateLaurelToCore` (warning fix) ✅
- Added `T20_RecursiveTransparent.lean` test file documenting a known limitation with recursive transparent procedures (test is commented out with a TODO explaining the remaining bug)

keyboardDrummer:
> I'd still like to see the transparent-calling-transparent test I mentioned above, once that is added I will approve

I've tried adding it in this PR but the support for transparent procedures in this PR is so small that it's not possible to define a terminating recursive transparent procedure. In the follow-up PR, https://github.com/keyboardDrummer/Strata/pull/28, it is possible and in there I've converted the existing test for recursive functions to use procedures, given that functions will be removed in the third PR of this series.

MikaelMayer (commented):

New commits since last review:
- Fixed pattern match for `PrimitiveOp` in `CoreGroupingAndOrdering.lean` to account for the new field added by the bindings commit ✅
- Deleted `T20_RecursiveTransparent.lean` — this was a commented-out test documenting a known limitation, not active coverage, so removal is fine.

All previously reviewed fixes remain intact (postcondition label, eraseDups comment, addFreePostcondition docstring, adjustSelectorName on Identifier, HashSet for asFunctionNames).

MikaelMayer (commented):

New commit since last review is a merge of `main2` bringing in unrelated bug fixes (Python for-loop, DDM circular imports, checkLeftRec, etc.). No changes to the transparency pass files themselves.

Note: joscoh's request for a transparent-procedure-calling-transparent-procedure test (from June 2) still appears outstanding — the existing `T20_TransparentBody.lean` only tests an opaque caller invoking a transparent procedure, not two transparent procedures in a call chain.

MikaelMayer (commented):

New commit since last review adds the transparent-procedure-calling-transparent-procedure test that joscoh requested: `tranparentCaller` (transparent) calls `transparentBody` (transparent), and `transparentCallerCaller` (opaque) verifies the result equals 3.

Minor nit: `tranparentCaller` has a typo (missing 's'), but this is just a test identifier and doesn't affect correctness.

joscoh (dismissed):

MikaelMayer (commented):

New commit since last review is a merge of `upstream/main2` bringing in import restructuring (Laurel→LaurelAST rename, Strata.Decimal→StrataDDM.Decimal, `module` declarations, `_root_.Core.verify` qualification). No changes to the transparency pass logic itself. All prior fixes remain intact (postcondition label, duplicate diagnostic removal, eraseDups comment, addFreePostcondition docstring, adjustSelectorName on Identifier, HashSet for asFunctionNames, polymorphism comment).

MikaelMayer (commented):

New commit since last review is a test fix (`T20_TransparentBody.lean`) adding the `module` declaration and `meta import all` / `meta section` pattern to align with the import restructuring from the main2 merge. No changes to transparency pass logic.

StrataTest/Languages/Laurel/Examples/Fundamentals/T20_TransparentBody.lean
```diff
@@ -3,9 +3,12 @@
 
   SPDX-License-Identifier: Apache-2.0 OR MIT
 -/
+module
 
-import StrataTest.Util.TestDiagnostics
-import StrataTest.Languages.Laurel.TestExamples
+meta import all StrataTest.Util.TestDiagnostics
```
joscoh, line 8: Why does this need a meta import?

keyboardDrummer: I don't know but all Laurel tests were changed to have this pattern as part of this PR: https://github.com/strata-org/Strata/pull/1010 and this test failed to work unless it was updated as well

joscoh (dismissed):

MikaelMayer (commented):

New commits since last review:
- `17c1d847` adds `module`/`meta import all`/`meta section` to T20_TransparentBody.lean (alignment with import restructuring from main2 merge)
- `68cec09b` removes `Foo_ensures_0_256` from the deterministic.lean StrataBoole test expectation (consequence of main2 merge changes to free postcondition obligation generation)

No changes to transparency pass logic. All prior fixes remain intact.

ssomayyajula (approved):

joscoh (approved):

Issue #1168, opened 2026-05-14
Split into multiple repositories


atomb:
To enforce clearer interfaces between components, decouple builds, and distribute code ownership, we would like to split the Strata codebase into multiple repositories. We can ease the transition process by first migrating code into separate packages within the current repository and then moving packages to separate repositories. We can use comments on this issue to discuss exactly how this split will occur.

One general philosophy we've decided on is to aim for language front ends and associated dialects to be as independent as possible from the main Strata codebase. It should be possible for them to be developed completely independently, possibly without the core Strata team even being aware of them. Currently, the Boole and Python dialects clearly fit into this category, as does the BoogieToStrata translator. All of them have a narrow interface with Strata: producing either the Laurel or Core dialect.

In addition, the Dialect Definition Mechanism (DDM) is an independent codebase that does not depend on anything else in Strata. In principle, it could be used in a completely unrelated context to define infrastructure for a new language.

Some code is very experimental, even if it may ultimately become production-ready. It can use a place to live while it matures (or to stay, if it remains an experiment).

And, finally, to avoid the `Strata` package depending on all the possible front ends, the top-level executables can live in a separate package that imports everything it needs.

Given those considerations, here's a concrete decomposition proposal.

* Move all code in [Strata/DDM](https://github.com/strata-org/Strata/tree/main/Strata/DDM) and its associated test code to a new `StrataDDM` package
  * This will require moving the tiny [Strata.Util.DecideProp](https://github.com/strata-org/Strata/tree/main/Strata/Util/DecideProp.lean) module
  * The main `Strata` package will depend on `StrataDDM`
* Move all code in [Strata/Languages/Python](https://github.com/strata-org/Strata/tree/main/Strata/Languages/Python) and its associated test code into `StrataPython`
* Move all code in [Strata/Languages/Boole](https://github.com/strata-org/Strata/tree/main/Strata/Languages/Boole) and its associated test code into `StrataBoole`
* Separate the code in [Tools/Python](https://github.com/strata-org/Strata/tree/main/Tools/Python) between the `StrataDDM` and `StrataPython` packages, for the Python DDM code generator and the Python dialect, respectively
* Move the code in [Tools/BoogieToStrata](https://github.com/strata-org/Strata/tree/main/Tools/BoogieToStrata) to a separate `StrataBoogie` package
* Move the top-level executables into `StrataCLI` (which already has a separate repository)
* Create a `StrataExperiments` package for experimental code

[...]

PR #1216, opened 2026-05-26
Split SimpleAPI.lean across package boundaries

aqjune-aws:
Prepare for the repository split by aligning API boundaries with package boundaries. Each future package (StrataDDM, Strata, StrataPython, etc.) will expose only the API surface it owns, without requiring downstream packages to import the full SimpleAPI and its dependencies.

SimpleAPI.lean is now a thin re-export of the modules in:
- Strata/DDM.lean
- Strata/Languages/Core/Core.lean
- Strata/Languages/Laurel/Laurel.lean
- Strata/Languages/Python/Python.lean

*Additional changes:*
- New LaurelAST.lean: Laurel AST types extracted from Laurel.lean so downstream files that needs only the AST (e.g., PythonLaurelTypedExpr.lean) can avoid pulling in the full Laurel API.
- `Core.runTransforms` extracted to PipelinePhase.lean — the inline transform loop from Verifier.lean is now a standalone function on `List PipelinePhase`, with optional telemetry and debug-file output.
- `Core.TransformPass` inductive removed — replaced by `Core.PipelinePhase` values via open `pass*` constructors.
- `readStrataProgram` in StrataMainLib now delegates to Strata.readStrataText (DDM) instead of manually assembling dialect contexts.
- `coreToGeneric` (noncomputable opaque) → coreToStrataProgram — now implemented via CST round-trip.


Strata/DDMAPI.lean
shigoel, line 1: I'd recommend putting these API files in their respective directories. E.g., Strata/DDM/API.lean. That'd allow the APIs to live in their respective packages.

Strata/Languages/Python/API.lean
```diff
@@ -0,0 +1,222 @@
...
+        | .ok info => acc := acc.push (info.moduleName, entry.path)
+        | .error msg =>
+          let _ ← IO.eprintln s!"warning: skipping {entry.path}: {msg}" |>.toBaseIO
+          continue
+    return acc
+  go sourceDir ⟨""⟩
+
+/-- Derive the output path for a Python file by mirroring the source directory
+    structure and replacing `.py` with `.pyspec.st.ion`. -/
+def pySpecOutputPath (sourceDir strataDir pythonFile : System.FilePath)
```
shigoel, line 85: Looks like this can be made private.

Strata/StrataAPI.lean
shigoel, line 1: "StrataAPI" is a confusing name, because this is Core-specific. I'd recommend moving this under Languages/Core (and analogously, LaurelAPI.lean under Languages/Laurel), and naming them as just API.lean.

Strata/Languages/Python/API.lean
```diff
@@ -0,0 +1,222 @@
...
+      failures := failures.push (toString mod, msg)
+    | .ok (sigs, warnings) =>
+      -- Write output
+      match ← Strata.Python.Specs.writeDDM outPath sigs |>.toBaseIO with
+      | .ok () => pure ()
+      | .error e =>
+        failures := failures.push (toString mod, s!"Could not write {outPath}: {e}")
+        continue
+      -- Report warnings per module
+      -- warningVerbosity: 0 = none, 1 = summary count, 2 = detailed list
```
shigoel: The old `WarningOutput` was an inductive API that described what the verbose level was better than numbers do. What's the rationale behind moving to `Nat` here?

Strata/StrataAPI.lean
```diff
@@ -0,0 +1,198 @@
...
+
+/-! ### Analysis of Core programs -/
+
+/--
+Power-user entry point for verifying a Core program. Exposes Lambda
+factories, external/prefix pipeline phases, a custom solver, a custom
+discharge function, and a shared `PipelineContext`. Most clients should
+prefer the basic `Core.verifyProgram`.
+-/
+def Core.verifyProgramAdvanced
```
shigoel: Not sure if we want a second function here, and also I'm a little iffy about using the terminology that calls out "power/expert" users. Perhaps `verifyProgram`'s defaults can be made to work out of the box, and a user is always free to provide their custom options.

Strata/StrataAPI.lean
```diff
@@ -0,0 +1,198 @@
...
+chain them with `Core.runTransforms`, or invoke a single pass via the
+per-pass entry points (`Core.inlineAllProcedures`, `Core.loopElimUsingContract`,
+…). -/
+
+/-- A single Core program transform pass.
+
+This type is intentionally opaque: build values with the smart constructors
+below (e.g., `Core.passLoopElim`, `Core.passInlineAll`). Clients should not
+pattern-match or inspect fields. -/
+structure Core.TransformPass where
```
shigoel: I like the new opaque structure! We should strive to make sure `validPasses` is in sync. with the available passes to aid discoverability.

Strata/DDMAPI.lean
shigoel: Revising this recommendation: let's move contents of Strata/DDMAPI.lean to Strata/Strata/DDM.lean (and analogously for other *API.lean files in this PR).

Strata/StrataAPI.lean
shigoel: Revised: Let's move contents of *API.lean to Languages/Core.lean and Languages/Laurel.lean

Strata/DDMAPI.lean
aqjune-aws: It seems Strata/Language/Laurel/Laurel.lean has its AST, but to follow the convention I will move it to Strata/Language/Laurel/LaurelAST.lean and place the APIs in Laurel.lean.

Strata/Languages/Python/API.lean
aqjune-aws: It had to be fixed, thanks for catching

Strata/StrataAPI.lean
aqjune-aws: Yeah, it was an artifact of the agent. I fixed it

Strata/StrataAPI.lean
aqjune-aws: Actually, I decided to use the existing `PipelinePhase` type and avoided introduction of a new type... Thanks for commenting

aqjune-aws:
Quite close to the final version - the types in APIs can be regarded as opaque types because APIs will provide the 'creators' of them. For example, users won't have to know what is the definition of Factory because `Core.defaultFactory` will return the default Factory. If they want more fine-grained control over this type, they can actually start to depend on the definition of Factory at the cost of being sensitive to the version of commit of Strata.

Strata/Util/IO.lean
```diff
@@ -34,113 +28,5 @@ def readBinInputSource (path : String) : IO ByteArray := do
 def displayName (path : String) : String :=
   if path == "-" then "<stdin>" else path
```
aqjune-aws: These are moved to DDM.lean

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -11,7 +11,7 @@ import Strata.Languages.Laurel.EliminateReturnsInExpression
 import Strata.Languages.Laurel.EliminateValueReturns
 import Strata.Languages.Laurel.ConstrainedTypeElim
 import Strata.Languages.Laurel.TypeAliasElim
-public import Strata.Languages.Core.Verifier
+public import Strata.Languages.Core.Core
```
atomb: Could we make the pattern be to have modules like `Strata/Languages/Core.lean`, so that this import is just `Strata.Languages.Core`, rather than `Strata.Languages.Core.Core`?

StrataBoole/StrataBoole/Verify.lean
```diff
@@ -968,7 +968,7 @@ def verify
     | .ok cp =>
       let runner tempPath :=
         EIO.toIO (fun dm => IO.Error.userError (toString (dm.format (some ictx.fileMap))))
-          (Core.verify cp tempPath proceduresToVerify options)
+          (_root_.Core.verify cp tempPath proceduresToVerify options)
```
atomb, line 972: It's kind of sad to need these prefixes, though I guess not a huge deal.

Strata/SimpleAPI.lean
shigoel, line 1: For the future split (#1168): Move SimpleAPI.lean to StrataCLI since its imports will span package boundaries, and rename it to Strata.lean so that it can serve as the "give me everything" convenience module.

atomb, line 1: In the long run I'd be in favor of getting rid of this file altogether, though it's probably worth keeping around for compatibility for a little while.

joehendrix:
I think `Strata/Languages/Core/Core.lean` should be renamed to `Strata/Languages/Core.lean` or just`Strata/Core.lean` given its prominance.  I do wonder whether the transforms in `Strata.Transform` should be moved under the Core namespace as well.  Are they all Core to Core transforms?

aqjune-aws:
> I do wonder whether the transforms in Strata.Transform should be moved under the Core namespace as well. Are they all Core to Core transforms?

Some of the transformations are on pure Imperative programs (e.g., StructuredToUnstructured). This was under a design assumption that other languages will have been written using Imperative and Lambda. It's hard to decide whether Strata.Transform should be moved to Strata.Languages.Core.Transform or not. :/

shigoel:
> I do wonder whether the transforms in `Strata.Transform` should be moved under the Core namespace as well. Are they all Core to Core transforms?

Most of them are, but there are some transforms that should be moved under DL (e.g., StructuredToUnstructured). I think we should find appropriate homes for them and this can be done in this PR or in the one that creates the StrataCLI package.

atomb (changes requested):
Just one key change to request: I'd like to avoid module names with duplication. I'd prefer `Strata.Languages.Core.Core` to be just `Strata.Languages.Core`

Strata/SimpleAPI.lean
shigoel: Agreed. My plan is to do this in the PR for StrataCLI.

shigoel: See https://github.com/strata-org/Strata/pull/1216#discussion_r3311918227

aqjune-aws:
Let me fix the conflicts after recent updates..!

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
aqjune-aws: Addressed in the latest commits. :)

StrataBoole/StrataBoole/Verify.lean
aqjune-aws: I think all files in Strata/ must start with `namespace Strata`, but there are a few files starting with e.g., `namespace Core`, and it is kind of making the collision between `Strata.Core.verify` and `Core.verify` hard to detect in declaration time. Like #1272 did, `namespace Strata` seems necessary for every .lean files in Strata/. This can be a follow-up patch.

atomb (approved):

shigoel (approved):

PR #1218, opened 2026-05-26
surface syntax casts for Boole and Core

kondylidou:
Adds Boole surface syntax for the three cast operators introduced in
pr/casts-core and extends them to Core as built-in functions.

## Surface syntax

| Boole syntax    | Core function  | Lowers to         | Semantics                       |
|-----------------|----------------|-------------------|---------------------------------|
| `e as_int`      | `as_uint(e)`   | `Bv{n}.ToUInt`    | unsigned widening (ubv_to_int)  |
| `e as_sint`     | `as_sint(e)`   | `Bv{n}.ToInt`     | signed widening (sbv_to_int)    |
| `e as_bv{n}`    | `as_bv{n}(e)`  | `Int.ToBv{n}`     | narrowing mod 2^n (int_to_bv n) |

All three are total. Source type is inferred; `as_int`/`as_sint` accept any
`bv{n}` source or `int` (identity); `as_bv{n}` accepts `int`.

## Changes

- **Boole/Grammar**: `cast_to_int`, `cast_to_sint`, `cast_to_bv1..bv128`
  DDM syntax rules with `@[prec(80)]` (above `*` at 30, `+` at 25, so
  `a * b as_bv8` → `a * (b as_bv8)`). Sync comment added.
- **Boole/Verify**: lowering to `Bv{n}.ToUInt` / `Bv{n}.ToInt` /
  `Int.ToBv{n}` Core ops. `as_int` and `as_sint` on an `int` source are
  now consistently a no-op (int is already signed). `bvWidth?` helper
  removes the repeated `bv1 _ | bv8 _ | … | bv128 _` match in every arm.
- **Core/DDMTransform/Grammar**: `as_uint`, `as_sint`, `as_bv1..bv128`
  as built-in function declarations usable in `#strata program Core;`.
- **Core/DDMTransform/Translate**: dispatch for the new grammar functions.

## Tests added / updated

- `cast_expr.lean` — all 3 operators × all 6 widths
- `cast_all_directions.lean` — cross-direction combinations
- `cast_errors.lean` — ill-typed casts; `as_sint`/`as_int` on `int` source
  pinned as passthrough (symmetric)
- `cast_nested.lean` — nested / chained cast expressions
- `widening_casts.lean` — widening patterns from dalek-lite `scalar.rs`
- `from_bytes_mod_order_wide_minimal.lean` — `u8_32_as_nat` seed
- `BvIntCastVerifyTests.lean` — rewritten as concrete `#strata program Core;`
  using the new `as_uint`/`as_sint`/`as_bv8` function syntax

## Reorganisation

- `FeatureRequests/widening_casts.lean` → `widening_casts.lean` (Gap 6
  closed, promoted to main test suite)
- `FeatureRequests/mutual_recursion.lean` → `mutual_recursion.lean`
- `FeatureRequests/seq_slicing.lean` → `seq_slicing.lean`
- `FeatureRequests/sha256_compact_indexed.lean`: doc note on remaining gaps
- `docs/BooleFeatureRequests.md`: Gap 6 (bv→int cast) marked closed


StrataBoole/StrataBoole/Verify.lean
```diff
@@ -492,6 +492,25 @@ partial def toCoreExpr (e : Boole.Expr) : TranslateM Core.Expression.Expr := do
...
+    | .bv1 _ | .bv8 _ | .bv16 _ | .bv32 _ | .bv64 _ | .bv128 _ => do
+      let n ← bvWidth m ty
+      return mkCoreApp (.op () (mkIdent s!"Bv{n}.ToUInt") none) [← toCoreExpr e]
+    | .int _ => toCoreExpr e
+    | _ => throwAt m s!"'as int' requires a bitvector source type, got: {repr ty}"
+  | .cast_to_sint m ty e =>
+    match ty with
+    | .bv1 _ | .bv8 _ | .bv16 _ | .bv32 _ | .bv64 _ | .bv128 _ => do
+      let n ← bvWidth m ty
+      return mkCoreApp (.op () (mkIdent s!"Bv{n}.ToInt") none) [← toCoreExpr e]
```
tautschnig: Asymmetry: `as_int` on an `int` source is silently passthrough (line 499), but `as_sint` on `int` is rejected. Either both should be a no-op (an `int` is already "signed"), or both should reject. As-is the user gets surprising behaviour depending on which surface keyword they typed. Please pick one and add a `#guard` test in `cast_errors.lean` (or `cast_expr.lean`) for the `int`-source case in both directions to lock the choice in — the existing tests don't exercise it.

Minor: the inner `match`es on `ty` repeat `bv1 _ | bv8 _ | … | bv128 _`. A small `Boole.Type.bvWidth?` helper (returns `Option Nat`) would let both arms read `match ty.bvWidth? with | some n => … | none => match ty with | .int _ => … | _ => err` and removes the duplication.

StrataBoole/StrataBoole/Grammar.lean
```diff
@@ -134,6 +134,22 @@ fn seq_drop_first (A : Type, s : Sequence A) : Sequence A =>
...
+fn cast_to_sint (T : Type, e : T) : int => @[prec(80)] e " as_sint";
+
+// Narrowing cast: `e as_bv{n}` converts an Int to a bitvector of width n (truncating mod 2^n).
+// Lowers to `Int.ToBv{n}` Core op → SMT-LIB 2.7 `(_ int_to_bv n)`.
+fn cast_to_bv1   (e : int) : bv1   => @[prec(80)] e " as_bv1";
+fn cast_to_bv8   (e : int) : bv8   => @[prec(80)] e " as_bv8";
+fn cast_to_bv16  (e : int) : bv16  => @[prec(80)] e " as_bv16";
+fn cast_to_bv32  (e : int) : bv32  => @[prec(80)] e " as_bv32";
+fn cast_to_bv64  (e : int) : bv64  => @[prec(80)] e " as_bv64";
+fn cast_to_bv128 (e : int) : bv128 => @[prec(80)] e " as_bv128";
```
tautschnig: Six near-identical `fn cast_to_bv{1,8,16,32,64,128}` declarations. The Core layer they lower to also carries the same shape (`Int.ToBv{n}`), so this duplication will recur every time bv widths are added (e.g. `bv256`). Worth a comment block here saying "keep widths in sync with `Bv{n}` types and Core's `Int.ToBv{n}` ops" so the next person adding a width updates all three.

Also, please document why `prec(80)` was chosen — neighbouring Boole `fn`s use prec 2/10/15 and the rest inherit Init defaults. The choice matters for parses like `a as_int + b` and is verified only by the tests, not by the grammar itself.

StrataTest/Languages/Boole/from_bytes_mod_order_wide_minimal.lean
```diff
@@ -0,0 +1,649 @@
...
+}
+ function nat.le (a : nat, b : nat) : bool {
+  nat.toInt(a) <= nat.toInt(b)
+}
+ function nat.gt (a : nat, b : nat) : bool {
+  nat.toInt(a) > nat.toInt(b)
+}
+ function nat.ge (a : nat, b : nat) : bool {
+  nat.toInt(a) >= nat.toInt(b)
+}
```
tautschnig: 649 lines, 108 declarations, exactly 2 occurrences of `as_int`. As a cast feature test this is overweight by two orders of magnitude — most of the file is a `nat` axiomatisation, sequence helpers, and a partial dalek-lite scalar reduction port. Please move it into a separate PR (or back to `FeatureRequests/` until the genuinely-required dependencies are in). At minimum, rename: "minimal" is misleading.

If you do want a cast-focused regression seed in this PR, a 30-line program with a single byte-sequence-to-int loop using `as_int` would cover the same ground with a tractable failure mode.

tautschnig (commented):
The cast feature itself (Grammar.lean +16, Verify.lean +24/-2) is small, correct, and well-targeted. The remaining +1400 lines fall into several other categories that should be split out before this lands — the diff is hard to review as one unit and a future bisect through this commit will be ambiguous.

Scope concerns (please split each into its own PR):
1. **Reorganisation moves** — `FeatureRequests/{mutual_recursion, widening_casts, seq_slicing}.lean` to `Boole/`. Independent of casts; only `widening_casts.lean` is even cast-related.
2. **`from_bytes_mod_order_wide_minimal.lean` (649 LOC, 2 occurrences of `as_int`)** — see inline. As a cast test it's vastly over-scoped; as an aspirational benchmark it doesn't belong here.
3. **Datatype let-binding refactor + comment-only edits in `Verify.lean`** (the unchanged `[.type (.data datatypes) .empty]` rewrite and the `per procedure` → `info.` comment) are stylistic and unrelated.
4. **`docs/BooleFeatureRequests.md` ±30 lines** is mostly a rewrite of unrelated entries (mutual recursion, sequences, opaque/reveal). Cast-only doc additions could land in a few lines.
5. **`ci: bump cvc5 1.2.1 → 1.3.4` (`c18eee51c`)** is a CI-wide change. Justified by the cast tests, but it changes the solver behaviour for *every* benchmark — needs its own scrutiny / separate PR.

Proof / `#guard` opportunities (none currently exist for the lowering):
- `theorem toCoreExpr_cast_to_int_bv (n : Nat) (e h_bv) : toCoreExpr (.cast_to_int m (bvN) e) = mkCoreApp (.op () ⟨"Bv{n}.ToUInt", ()⟩ none) [← toCoreExpr e]` — trivial but locks in the op-name string.
- `theorem toCoreExpr_cast_to_int_int (e) : toCoreExpr (.cast_to_int m .int e) = toCoreExpr e` — explicitly proves the identity passthrough.
- The op-name strings (`Bv{n}.ToUInt`, `Int.ToBvN`) live as `String` literals here and as factory keys elsewhere; a small `enum CastOp` plus `def Core.castOp.name : CastOp → String` would make typos a compile-time error and let the proofs above just be `rfl`.

StrataTest/Languages/Boole/from_bytes_mod_order_wide_minimal.lean
kondylidou: I will move it back to fr. But this is a benchmark we are targeting to verify so it's an important one hence the length.

atomb:
I have two general comments on this:
* We could add built-in functions instead of special syntax. To me `as_int(e)` is just as concise/readable/elegant as `e as_int`. (Though I can see an argument for the more general `e as int`, where `int` could be any type.)
* I think it'd be nice to have these in Core. In general, I see most features in Boole as potential additions to Core in the long run, but this is one where I don't see any reason to hesitate.

kondylidou:
> The cast feature itself (Grammar.lean +16, Verify.lean +24/-2) is small, correct, and well-targeted. The remaining +1400 lines fall into several other categories that should be split out before this lands — the diff is hard to review as one unit and a future bisect through this commit will be ambiguous.
> 
> Scope concerns (please split each into its own PR):
> 
> 1. **Reorganisation moves** — `FeatureRequests/{mutual_recursion, widening_casts, seq_slicing}.lean` to `Boole/`. Independent of casts; only `widening_casts.lean` is even cast-related.
> 2. **`from_bytes_mod_order_wide_minimal.lean` (649 LOC, 2 occurrences of `as_int`)** — see inline. As a cast test it's vastly over-scoped; as an aspirational benchmark it doesn't belong here.
> 3. **Datatype let-binding refactor + comment-only edits in `Verify.lean`** (the unchanged `[.type (.data datatypes) .empty]` rewrite and the `per procedure` → `info.` comment) are stylistic and unrelated.
> 4. **`docs/BooleFeatureRequests.md` ±30 lines** is mostly a rewrite of unrelated entries (mutual recursion, sequences, opaque/reveal). Cast-only doc additions could land in a few lines.
> 5. **`ci: bump cvc5 1.2.1 → 1.3.4` (`c18eee51c`)** is a CI-wide change. Justified by the cast tests, but it changes the solver behaviour for _every_ benchmark — needs its own scrutiny / separate PR.
> 
> Proof / `#guard` opportunities (none currently exist for the lowering):
> 
> * `theorem toCoreExpr_cast_to_int_bv (n : Nat) (e h_bv) : toCoreExpr (.cast_to_int m (bvN) e) = mkCoreApp (.op () ⟨"Bv{n}.ToUInt", ()⟩ none) [← toCoreExpr e]` — trivial but locks in the op-name string.
> * `theorem toCoreExpr_cast_to_int_int (e) : toCoreExpr (.cast_to_int m .int e) = toCoreExpr e` — explicitly proves the identity passthrough.
> * The op-name strings (`Bv{n}.ToUInt`, `Int.ToBvN`) live as `String` literals here and as factory keys elsewhere; a small `enum CastOp` plus `def Core.castOp.name : CastOp → String` would make typos a compile-time error and let the proofs above just be `rfl`.

I am ok to remove `from_bytes_mod_order_wide_minimal.lean` and make a new pr with this target benchmark. The rest which is just the casts for Boole and some reorganization I believe can stay in one pr. This would be too small to split. The reorg of benchmark also happened because casts were implemented. Some benchmark stopped being incomplete. Also the ci bump is already merged by the previous pr so not anymore in this one. The document `docs/BooleFeatureRequests.md` changes everytime I implement something since I'm documenting it there. I know the document has some more changes but these are cosmetic.

StrataBoole/StrataBoole/Verify.lean
kondylidou: done, thanks

kondylidou:
> I have two general comments on this:
> 
> * We could add built-in functions instead of special syntax. To me `as_int(e)` is just as concise/readable/elegant as `e as_int`. (Though I can see an argument for the more general `e as int`, where `int` could be any type.)
> * I think it'd be nice to have these in Core. In general, I see most features in Boole as potential additions to Core in the long run, but this is one where I don't see any reason to hesitate.

added to Core:) will update the pr description

StrataBoole/StrataBoole/Grammar.lean
kondylidou: done, thanks

StrataBoole/StrataBoole/Grammar.lean
```diff
@@ -134,6 +134,26 @@ fn seq_drop_first (A : Type, s : Sequence A) : Sequence A =>
 fn seq_subrange (A : Type, s : Sequence A, lo : int, hi : int) : Sequence A =>
   "Sequence.subrange" "(" s ", " lo ", " hi ")";
 
+// Unsigned widening cast: `e as_int` converts a bitvector to Int (unsigned).
+// Lowers to `Bv{n}.ToUInt` Core op → SMT-LIB 2.7 `ubv_to_int`.
+fn cast_to_int (T : Type, e : T) : int => @[prec(80)] e " as_int";
```
atomb: My thought was that by adding these as built-in functions in Core, they wouldn't need to exist as special syntax in Boole, because Boole extends Core.

kondylidou: well there is a different syntax that we want to keep in Boole

StrataBoole/StrataBooleTest/cast_nested.lean
```diff
@@ -0,0 +1,177 @@
...
+
+Exercises `e as_int` in contexts beyond the flat `x as_int` covered by
+`cast_expr.lean`:
+
+- Cast inside a ∀-quantifier body over a sequence
+- Cast of a compound BV expression (bitwise &)
+- Cast in an arithmetic sum bound
+- Cast inside a `let_in_expr` binding
+- Cast inside a ∃-quantifier body
+- Cast inside a `rec function` body with `decreases`, plus axiom-backed
```
joscoh, line 22: Not sure if the resulting proof obligations would be provable currently, but the more interesting test to me would be a cast inside the `decreases` clause. Also, one possible test to add is a cast in a function precondition (with a subsequent caller).

joscoh (dismissed):
Not blocking, but a suggested test case or two

shigoel (dismissed):
CI fails, but LGTM otherwise. @kondylidou -- mind fixing?

StrataBoole/StrataBooleTest/cast_nested.lean
kondylidou: good idea!

joscoh:
CI still broken, once it is fixed I will re-approve

joscoh (approved):

shigoel (approved):

shigoel:
As this PR touches cross-cutting code (following the Strata repo. split), we've split this into
https://github.com/strata-org/Strata/pull/1384
https://github.com/strata-org/Strata-Boole/pull/3
both of which have been merged.
@kondylidou Please close this PR. Thanks!

PR #1219, opened 2026-05-26
Examples using Lean to prove VCs of Core programs

atomb:
Previously, all of the examples of this functionality used the Boole dialect.

Almost all of the examples added in this PR can be verified with `grind`, but one requires a few extra steps to deal with division.

Note that this unfortunately only works if the files containing Lean theorems don't use `module`. There's no way to have `module`, `#strata` programs, and `gen_smt_vcs proofs` in the same file without `precompileModules`, and Strata doesn't build with `precompileModules` due to known issues with environment extensions being loaded multiple times. If we can resolve that issue, we can turn these tests back into proper `module`.


Strata/MetaVerifier.lean
```diff
@@ -162,15 +162,17 @@ def toSMTVCs (vcs : Core.coreVCs) : Option SMT.SMTVCs := do
 /--
 Generate SMT verification conditions for a `Strata.Program`.
 -/
-def genSMTVCs (program : Program) : Option SMT.SMTVCs := do
+@[expose]
+public def genSMTVCs (program : Program) : Option SMT.SMTVCs := do
```
joscoh: Wondering why these have to be public and exposed. IIUC, using the tactic should not involve unfolding these definitions, right?

shigoel (dismissed):

Strata/MetaVerifier.lean
atomb: You're right. They don't need to be. That's an artifact of a previous experiment.

joscoh (dismissed):

StrataTest/Languages/Core/Examples/Functions.lean
```diff
@@ -72,6 +73,10 @@ Result: ✅ pass
 #guard_msgs in
 #eval verify funcPgm
 
+theorem testPgm_correct : smtVCsCorrect funcPgm := by
```
tautschnig: Naming nit: theorem is `testPgm_correct` but the program is `funcPgm`. Rename to `funcPgm_correct` to match the program (and the convention used by the other theorems in this PR, including line 159's `quantBodyFuncPgm_correct`).

StrataTest/Languages/Core/Examples/Loops.lean
```diff
@@ -544,6 +553,20 @@ Result: ✅ pass
 #guard_msgs in
 #eval verify precondElimInMeasurePgm (options := .quiet)
 
+/-- This theorem requires a little bit of manual work, though most goals are
+solved partly or entirely by `grind`. I suspect a future version of `grind`
+might be able to do it unaided. -/
+theorem precondElimInMeasurePgm_correct : smtVCsCorrect precondElimInMeasurePgm := by
```
tautschnig, line 571: The two positional `intros … inonneg meas_def` and `intros … _ meas_def` blocks (lines 561, 565) are brittle: their correctness depends on the *number*, *order*, and *types* of the binders that `gen_smt_vcs` happens to produce after `all_goals (try grind)`. Any future change to VC generation, axiom ordering, or what `grind` happens to dispatch will silently misalign these and leave a confusing failure.

More stable alternatives:

```lean
all_goals (try grind)
case _ n d i meas _ dpos _ _ _ inonneg meas_def => …
case _ n d i meas _ dpos _ _ _ _ meas_def => …
```

or using `rename_i` to explicitly name only the binders you reference. Either way, please add a one-line comment per remaining goal stating the obligation (e.g. `-- loop_measure_end_calls_Int.SafeDiv_0`) so the manual proof is self-documenting if `grind` later subsumes one of them.

For stronger guarantees worth considering as a follow-up: extract the two arithmetic facts as named lemmas (`Int.ediv_nonneg_of_pos`, `Int.sub_div_pred`) so the proof reads as `case _ => exact int_lemma_1 …; case _ => exact int_lemma_2 …`. This also lets `grind` provers exercise just the lemma, not the surrounding VC plumbing.

joscoh: FWIW, I agree with using explicit `case` to ensure that goals are not silently appearing/getting solved, but I disagree that `intros` itself is unstable because of binder order, etc. All of these issues are also present with `case` and are unavoidable. `rename_i` is even more brittle since it interacts with other hypotheses in the context. I also think worrying about `intros` order brittleness is a little silly, since `all_goals (try grind)` is already an extremely unstable tactic script.

joscoh (dismissed):

StrataTest/Languages/Core/Examples/Loops.lean
atomb: Yep, I agree. But I do think the structure of this proof after responding to that feedback is better than it was before. :)

aqjune-aws:
Do you want to resolve the merge conflicts?

aqjune-aws (dismissed):

StrataTest/Languages/Core/Examples/Exit.lean
```diff
@@ -3,14 +3,12 @@
 
   SPDX-License-Identifier: Apache-2.0 OR MIT
 -/
-module
 
-meta import Strata.Languages.Core
-meta import Strata.Languages.Core
-meta import StrataTest.Languages.Core.Examples.Loops
+import Strata.Languages.Core
+import Strata.Languages.Core
```
joehendrix, line 7: Core doesn't need to be imported twice.

StrataTest/Languages/Core/Examples/Loops.lean
```diff
@@ -556,6 +564,36 @@ Result: ✅ pass
...
+  -- measure_lb_0: the loop measure i / d is non-negative
+  case measure_lb_0 =>
+    intro _ d i _ _ dpos _ _ _ inonneg meas_def
+    subst meas_def
+    exact ediv_nonneg_of_pos i d inonneg dpos
+  -- measure_decrease_0: the loop measure i / d strictly decreases
+  case measure_decrease_0 =>
+    intro _ d i _ _ dpos _ _ _ _ meas_def
+    subst meas_def
+    exact ediv_sub_lt i d dpos
```
joehendrix: Minor nit.  You could get rid of the extra lemma above with this:

```suggestion
    have p := Int.add_mul_ediv_left (a := i) (b := d) (c := -1)
    grind
```

StrataTest/Languages/Core/Examples/Loops.lean
```diff
@@ -556,6 +564,36 @@ Result: ✅ pass
...
+division, using the lemmas above, though most goals are solved by `grind`.
+-/
+theorem precondElimInMeasurePgm_correct : smtVCsCorrect precondElimInMeasurePgm := by
+  gen_smt_vcs
+  all_goals (try grind)
+  -- measure_lb_0: the loop measure i / d is non-negative
+  case measure_lb_0 =>
+    intro _ d i _ _ dpos _ _ _ inonneg meas_def
+    subst meas_def
+    exact ediv_nonneg_of_pos i d inonneg dpos
```
joehendrix: The extra lemma isn't needed with this (the have adds enough context that omega is able to figure out the inequalities).

```suggestion
    have p := Int.ediv_nonneg (a := i) (b := d)
    grind
```

joehendrix (dismissed):
I found some minor things to pick on, but this generally looks fine.

There does seem to be a Lean module bug when `Loops` is built with modules.  I have Claude working on reproducing and minimizing, but this seems fine for now.

joehendrix:
Claude and I found the bug.  It's a lake error fixed by [this PR](https://github.com/leanprover/lean4/pull/13600).

It should be fixed in 4.30.

atomb:
> Claude and I found the bug. It's a lake error fixed by [this PR](https://github.com/leanprover/lean4/pull/13600).
> 
> It should be fixed in 4.30.

Nice! Once we move to 4.30, I'll migrate these back to proper modules.

joehendrix (approved):

aqjune-aws (approved):

PR #1221, opened 2026-05-27
Extract DDM into standalone StrataDDM package

joehendrix:
Extracts the DDM (Dialect Definition Mechanism) into a standalone Lake package so it can be versioned and built independently of the main Strata package.

- DDM has no dependencies on the rest of Strata, making it a natural package boundary. Moved `Strata/DDM/` to `StrataDDM/` with its own `lakefile.toml` and the root package now depends on it via a path-based `require`.
- `#load_dialect` path resolution changed from file-relative to working-directory-relative (i.e., the package root). File-relative paths required fragile `../` escapes that broke when files moved between packages; CWD-relative is consistent with how Lake and Lean resolve other paths.
- `decideProp` made private (root namespace, `import all` access only) to avoid a duplicate-symbol conflict.  It's an internal utility, not part of the public API.
- Java codegen tests moved from `StrataTestExtra/DDM/Integration/Java/` to `StrataTestExtra/Languages/Java/` to reflect their role as a language integration test rather than a DDM-internal test.
- CI updated to build `StrataDDM` before the main package.


Strata/Util/IO.lean
```diff
@@ -5,9 +5,9 @@
 -/
 module
 
-import Strata.DDM.Elab
-public import Strata.DDM.Elab.LoadedDialects
-import Strata.DDM.Util.Ion
+import StrataDDM.Elab
+public import StrataDDM.Elab.LoadedDialects
```
shigoel, line 9: Note: PR #1216 (Split SimpleAPI) moves the `readStrataText`/`readStrataIon`/`readFile`/`DialectOrProgram` definitions out of `Strata/Util/IO.lean` and into `DDM.lean`. After #1216 rebases on this PR, that content will land in StrataDDM (its natural home). Just flagging so you're aware of the overlap — no action needed here.

shigoel (approved):

StrataDDM/StrataDDMTest/dialects/Bool.dialect.st
atomb, line 1: I just realized that it's a little weird to have this dialect and another one called "Boole". But that's not a change for this PR.

atomb (approved):

PR #1222, opened 2026-05-27
Document Laurel passes and their dependencies, and other documentation updates

keyboardDrummer:
### Changes
Update Laurel documentation based on inspiration from https://github.com/strata-org/Strata/pull/1144 and by making the dependencies between passes explicit

### Testing
Refactoring, no additional tests needed


Strata/Languages/Laurel/LaurelPass.lean
```diff
@@ -0,0 +1,37 @@
...
+public import Strata.Languages.Laurel.Resolution
+public import Strata.Util.Statistics
+
+namespace Strata.Laurel
+
+public section
+
+mutual
+structure ComesBefore where
+  name : LaurelPass
```
MikaelMayer: 🤖🔍 Nit: The field `name` of type `LaurelPass` is confusing since `LaurelPass` also has a `name : String` field. Consider renaming this to `pass` or `target` to avoid the double `.name.name` access pattern seen in `LaurelDoc.lean`.

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -571,6 +575,18 @@ def heapParameterization (model: SemanticModel) (program : Program) : Program :=
...
+public def heapParameterizationPass : LaurelPass where
+  name := "HeapParameterization"
+  documentation := "Transforms procedures that interact with the heap by adding explicit heap parameters. The heap is modeled as `Map Composite (Map Field Box)`. Procedures that write the heap receive both an input and output heap parameter; procedures that only read the heap receive an input heap parameter. Field reads and writes are rewritten to use `readField` and `updateField` functions."
+  needsResolves := true
+  run := fun p m =>
+    (heapParameterization m p, [], {})
+  comesBefore := [
+      ⟨ typeHierarchyTransformPass, "the type hierarchy pass modifies the 'Composite' datatype that is introduced by this pass." ⟩,
+      ⟨ modifiesClausesTransformPass, "the modifies pass refers to several types and variables introduced by heap parameterization: Composite, Field, $heap_in, $heap." ⟩,
+      ⟨ liftExpressionAssignmentsPass, "the heap paramterization pass introduces assignments (to the heap variables) that need to be lifted."⟩]
```
MikaelMayer: 🤖🔍 Nit: Typo — "paramterization" → "parameterization".

Strata/Languages/Laurel/Laurel.lean
```diff
@@ -322,12 +323,17 @@ inductive StmtExpr : Type where
   | Abstract
   /-- Refers to all objects in the heap. Used in reads or modifies clauses. -/
   | All
-  /-- A hole representing an unknown expression.
+  /-- A hole represents an unknown expression.
+      This can be used to represent programs that are still under development, for example the program `3 + `
+      The defining property of a hole is that interaction with it and other code should not produce any errors.
+      Besides representing partial user programs,
+      holes can also to handle under development parts of compilers that target Laurel.
```
MikaelMayer: 🤖🔍 Nit: Grammar — "holes can also to handle" → "holes can also be used to handle".

docs/verso/LaurelDoc.lean
```diff
@@ -22,8 +22,63 @@ open Verso.Genre Manual
 -- environment as Verso
 open Verso.Genre.Manual.InlineLean
 
+open Verso.Doc.Elab (DocElabM)
```
MikaelMayer: 🤖🔍 Nit: `DocElabM` is opened but never used in this file.

docs/verso/LaurelDoc.lean
```diff
@@ -146,38 +208,62 @@ A Laurel program consists of procedures, global variables, type definitions, and
...
+
+The static semantics of Laurel are defined by `Resolution.lean`. This is where Laurel references are resolved and where type checking, when implemented, will be done. Calling `resolve` will produce diagnostics and a `SemanticModel` that can be used to navigate between definitions and references.
+If new references or definitions are created during compilation, `resolve` must be called again to get a complete model.
+
+## Translation Pipeline
 
 Laurel programs are verified by translating them to Strata Core and then invoking the Core
-verification pipeline. The translation involves several passes, each transforming the Laurel
-program before the final conversion to Core.
+verification pipeline. The Laurel compilation pipelines consists of three parts:
```
MikaelMayer: 🤖🔍 Nit: "pipelines consists" → "pipeline consists".

docs/verso/LaurelDoc.lean
```diff
@@ -146,38 +208,62 @@ A Laurel program consists of procedures, global variables, type definitions, and
...
-## Lifting Expression Assignments
+# Differences between Laurel and Core
 
-The expression assignment lifting pass transforms assignments that appear in expression
-contexts into preceding statements. This is necessary because Strata Core does not support
-assignments within expressions.
+## Language design
+
+### Parameter lists
+Parameter lists. In Laurel, input and output parameters are defined in a separate list. Inout parameters are defined by repeated the parameter name in both lists. In Core, there is a single parameter list where each parameter defines its kind (in/out/inout).
```
MikaelMayer: 🤖🔍 Nit: "by repeated the parameter name" → "by repeating the parameter name".

MikaelMayer (commented):
🤖🔍 Good refactoring — extracting pass definitions to their respective modules and making the pipeline declarative is a clear improvement. The `comesBefore` dependency tracking is a nice addition for documentation generation. A few nits below.

Strata/Languages/Laurel/LaurelPass.lean
```diff
@@ -0,0 +1,37 @@
...
+  /-- Human-readable name, used for profiling and file emission. -/
+  name : String
+  /-- Whether `resolve` should be run after the pass. -/
+  needsResolves : Bool := false
+  /-- The pass action. -/
+  run : Program → SemanticModel → Program × List DiagnosticModel × Statistics
+  /-- A description of what this pass does, used for documentation generation. -/
+  documentation : String
+  comesBefore : List ComesBefore := []
+end
```
tautschnig, line 33: Two structural concerns with this design:

1. **`comesBefore` references the entire `LaurelPass` struct**, which forces the `mutual` block and means every pass file that wants to declare a dependency must `import` the dependency's file. Concretely: `HeapParameterization.lean` now imports `TypeHierarchy.lean`, `ModifiesClauses.lean`, and `LiftImperativeExpressions.lean` just to spell three constraints. That's tight coupling and creates spurious build edges. Storing dependencies by `String` (the pass name) and resolving names via `laurelPipeline` at the doc-generation site would eliminate the imports and let `LaurelPass` be a plain (non-mutual) structure:

   ```suggestion
   structure ComesBefore where
     /-- Name of the pass this one must precede. Resolved against
         `laurelPipeline` at use sites; an unknown name is a build error
         caught by the `laurelPipeline_comesBefore_resolved` theorem
         in `LaurelCompilationPipeline.lean`. -/
     target : String
     reason : String

   structure LaurelPass where
     name : String
     needsResolves : Bool := false
     run : Program → SemanticModel → Program × List DiagnosticModel × Statistics
     documentation : String
     comesBefore : List ComesBefore := []
   ```

2. The MikaelMayer nit about `name : LaurelPass` (still unresolved) is symptomatic of the same issue — once you go by string, the awkward `cb.name.name` access pattern in `LaurelDoc.lean` becomes `cb.target` and there's no shadowing.

After the rename the proof obligations in the top-level review become straightforward `decide`-able checks.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -76,63 +84,19 @@ public section
...
+  heapParameterizationPass,
+  typeHierarchyTransformPass,
+  modifiesClausesTransformPass,
+  inferHoleTypesPass,
+  eliminateDeterministicHolesPass,
+  desugarShortCircuitPass,
+  liftExpressionAssignmentsPass,
+  mergeAndLiftReturnsPass,
+  constrainedTypeElimPass
 ]
```
tautschnig, line 106: `typeAliasElim` (line 128 of this file), the initial `resolve`, and `validateDiamondFieldAccesses` all run *outside* this array. So `{laurelPipelineDocs}` in the manual will silently omit them — recreating the exact gap copilot flagged on #1144. Two ways to fix:

- Make all of them `LaurelPass` values (with `documentation` strings) and put the whole sequence into one array; then drop the special-casing in the loop above.
- Or split into a `prePipeline : Array LaurelPass` (resolve, diamond, typeAliasElim) and `mainPipeline : Array LaurelPass`, then have `laurelPipelineDocs` iterate both. Either way, the rendered manual should not be missing steps that are part of the actual pipeline.

While here: the trailing `CoreGroupingAndOrdering` and the final `LaurelToCore` translation are also pipeline phases worth surfacing in the manual — even as a one-line summary at the bottom of `Lowering Passes` — so the reader sees the complete journey from Laurel to Core.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -76,63 +84,19 @@ public section
...
-      let (p', diags) := constrainedTypeElim m p
-      (p', diags, {}) }
+def laurelPipeline : Array LaurelPass := #[
+  filterNonCompositeModifiesPass,
+  eliminateValueInReturnsPass,
+  heapParameterizationPass,
+  typeHierarchyTransformPass,
+  modifiesClausesTransformPass,
+  inferHoleTypesPass,
+  eliminateDeterministicHolesPass,
```
tautschnig, line 101: Of the eleven passes only five carry any `comesBefore`. The other six (`filterNonCompositeModifiesPass`, `typeHierarchyTransformPass`, `modifiesClausesTransformPass`, `eliminateDeterministicHolesPass`, `liftExpressionAssignmentsPass`, `mergeAndLiftReturnsPass`, `constrainedTypeElimPass`) are placed by current convention only — the dependency graph that `{laurelPipelineDependencyGraph}` renders will look much sparser than the actual constraint set is. Please go through each and either declare the real reasons (most of these almost certainly do have constraints — e.g. `constrainedTypeElim` clearly comes after several others) or add a one-line *“no ordering constraints — independent of other passes”* comment so the omission is intentional and not just unfilled.

With the proof obligation suggested in the top-level review, an empty `comesBefore` for a pass that actually has invisible constraints would be silent today but a future refactor breaking it would surface as a red CI.

docs/verso/LaurelDoc.lean
```diff
@@ -8,7 +8,7 @@ import VersoManual
 
 import Strata.Languages.Laurel.Laurel
 import Strata.Languages.Laurel.LaurelTypes
-import Strata.Languages.Laurel.LaurelToCoreTranslator
+import Strata.Languages.Laurel.LaurelCompilationPipeline
```
tautschnig, line 11: High-level review with respect to #1144, comment placed at arbitrary line:

Where this PR overlaps with #1144 — context audit:
- ✅ Pipeline-list maintenance burden: addressed by auto-generation here, in a way #1144 couldn't have.
- ✅ Hole-elimination terminology (#1144 LaurelDoc.lean:231/281; .kiro:88): renames make the doc accurate.
- ❌ `TypeAliasElim` omitted from the pipeline list (#1144 LaurelDoc.lean:287): still happens here; see inline. The auto-generated section will be incomplete in exactly the same way.
- ❌ The Design Principles / Conventions and Invariants / Error Correction via Holes / Python Front-End Relationship sections from #1144 are not in this PR. Those are independent additions; this PR is narrower in scope, which is fine, but worth either pulling them in here or stating clearly that they're left for a follow-up rebase of #1144 onto this work.
- ❌ olivier-aws's questions on #1144 (build-time check enforcing pipeline ordering, `$` collision with Java identifiers, error diagnostics carrying source location) are not addressed by this refactor either. They're orthogonal, so fair to leave for the #1144 follow-up.

keyboardDrummer:
@keyboardDrummer-bot please address the nits from @MikaelMayer

MikaelMayer (commented):

Strata/Languages/Laurel/LaurelPass.lean
keyboardDrummer: I think it's better to reference the pass instead of the name of the pass, if only to avoid the magic string. 

Note that instead of passes, I would like these files to become `LanguageFeatures`. The field `comesBefore` will turn into `dependsOn` and the features will reference more from each other than just the handle.

For example, `ModifiesClauses` currently contains a lot of magic strings that it should really reference from `HeapParameterization`. Currently `ModifiesClauses` is a pass that comes after `HeapParameterization`, but it should really be a language features that depends on the heap parameterization feature, and extend the pass added by heap parameterization instead of adding a new pass.

The current `comesBefore` field is chosen is chosen as a stepping stone towards that `dependsOn`. I considered having both a `comesBefore` and `comesAfter` field, since `ModifiesClauses` logically depends on `HeapParameterization`, not the other way around as this PR implies, but I decided on keeping this temporary state simple, despite its flaws.

The above refactoring is described here: https://github.com/strata-org/Strata/issues/1274

> let LaurelPass be a plain (non-mutual) structure:

Why do you mention this? Is there a reason to avoid mutual structures? Seems 100% fine to me.

> would eliminate the imports

My preference is not to be frugal with imports unless they are unused or we are trying to optimize build times of a particular part.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
keyboardDrummer: > typeAliasElim (line 128 of this file), the initial resolve, and validateDiamondFieldAccesses all run outside this array

I've moved `validateDiamondFieldAccesses` into Resolution so it's not a separate thing. And I've created a `LaurelPass` object for `typeAliasElim`. The initial resolve is a bit different. Like the parser, it's part of the "front-end" of the Laurel language, before it translates to anything. It is already mentioned though, here:

```
# Implementation

The static semantics of Laurel are defined by `Resolution.lean`. This is where Laurel references are resolved and where type checking, when implemented, will be done. Calling `resolve` will produce diagnostics and a `SemanticModel` that can be used to navigate between definitions and references.
If new references or definitions are created during compilation, `resolve` must be called again to get a complete model.

```

> While here: the trailing CoreGroupingAndOrdering and the final LaurelToCore translation are also pipeline phases worth surfacing in the manual 

They are mentioned:

```
## Translation Pipeline

Laurel programs are verified by translating them to Strata Core and then invoking the Core
verification pipeline. The Laurel compilation pipeline consists of three parts:
- Lowering, consisting of many phases. Maps Laurel to Laurel
- Ordering, consisting of a single pass. Maps Laurel to OrderedLaurel
- Translation, consisting of a single pass. Maps OrderedLaurel to Core.
```

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
keyboardDrummer: I have already gone through the passes and I've added the ordering constraints that I'm aware of. If we've missed certain constraints, I think that's fine. Whomever reorders things and runs into an issue will be prompted to add the missing constraints. If we decide to infer the ordering from the constraints in the future, then we might also find missing constraints.

Adding the comment seems misleading to me. The presence of such a comment does not guarantee that the ordering constraints are up to date.

> constrainedTypeElim clearly comes after several others

Like what?

docs/verso/LaurelDoc.lean
keyboardDrummer: > but worth either pulling them in here or stating clearly that they're left for a follow-up rebase

Let's consider this PR as independent of #1144. If there are other things that are added in #1144, that's fine. I've left some comments on #1144

MikaelMayer (commented):

MikaelMayer (commented):

keyboardDrummer:

MikaelMayer (commented):

MikaelMayer (commented):

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -861,9 +733,127 @@ def buildRefToDef (program : Program) : Std.HashMap Nat ResolvedNode :=
...
+      parentsWithField.length > 1
+  | _ => false
+
+/--
+Check whether accessing `fieldName` on `target` is a diamond-inherited field access,
+and if so return a diagnostic error using the given `source` range.
+-/
+private def checkDiamondFieldAccess (model : SemanticModel) (target : StmtExprMd)
+    (fieldName : Identifier) (source : Option FileRange) : List DiagnosticModel :=
+  let targetType := match target.val with
```
tautschnig: Restore the `computeExprType`-based target typing, or add a regression test that pins this narrowed coverage.

<details>
<summary>Why</summary>

This block replaces `computeExprType model target` (the prior implementation in `TypeHierarchy.lean`) with a 4-case approximation. `computeExprType` (`LaurelTypes.lean:39`) also resolves `.AsType` casts, `.StaticCall`/`.InstanceCall` results, `.PureFieldUpdate`, `.Block`, and `.IfThenElse`; those now fall to `.Unknown` and are skipped, so `(b as Bottom)#xValue` and `getBottom()#xValue` no longer error where they did before this PR.

The narrowing is forced by the move into `Resolution.lean`: `LaurelTypes.lean` publicly imports `Resolution`, so `Resolution` cannot call `computeExprType` without a cycle. Either keep `validateDiamondFieldAccesses` in a module downstream of `LaurelTypes` and run it as the first pipeline pass, or — if the narrowing is intended — pin it with a test, e.g. extend T5 with a cast target:

```
b2 := (b as Bottom)#xValue   // errored before this PR; silently accepted now
```

Either way "refactoring, no additional tests needed" doesn't hold — the set of rejected programs changed.
</details>

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -928,13 +918,15 @@ def resolve (program : Program) (existingModel: Option SemanticModel := none) :
...
   { program := program',
-    model := {
-      compositeCount := program.types.length,
-      refToDef := refToDef,
-      nextId := finalState.nextId
-    },
-    errors := finalState.errors
+    model := semanticModel,
+    errors := finalState.errors ++ diamondErrors
   }
```
tautschnig, line 930: Confirm diamond-access errors should now be suppressed when `emitResolutionErrors := false`.

<details>
<summary>Why</summary>

Before this PR `validateDiamondFieldAccesses` ran in `runLaurelPasses` and its result was added to `allDiags` unconditionally. Now the errors are folded into `ResolutionResult.errors`, which the pipeline keeps only via `resolutionErrors := if options.emitResolutionErrors then result.errors.toList else []` (`LaurelCompilationPipeline.lean:128`). The field defaults to `true`, so in-tree tests are unaffected, but any caller passing `false` loses diamond diagnostics it previously received. If diamond errors are intended as resolution-phase errors this is fine — worth a line in the PR description.
</details>

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -84,63 +87,20 @@ public section
...
-  { name := "EliminateReturns"
-    needsResolves := true
-    run := fun p _m =>
-      (eliminateReturnsInExpressionTransform p, [], {}) },
-  { name := "ConstrainedTypeElim"
-    needsResolves := true
-    run := fun p m =>
-      let (p', diags) := constrainedTypeElim m p
-      (p', diags, {}) }
+def laurelPipeline : Array LaurelPass := #[
```
tautschnig, line 93: Add a `#guard` that `laurelPipeline` satisfies every declared `comesBefore`, so reordering can't silently violate a constraint.

<details>
<summary>Why</summary>

`comesBefore` is consumed only by the doc generator; nothing checks it against the actual order. The line-99 resolution ("whoever reorders and runs into an issue will be prompted") only holds if something prompts them — this is that mechanism, and it also catches a `comesBefore` target that isn't in the pipeline at all.

```lean
/-- Every `comesBefore` constraint is respected by the pipeline order. -/
def comesBeforeRespected : Bool :=
  let names := laurelPipeline.toList.map (·.name)
  laurelPipeline.toList.enum.all fun (i, p) =>
    p.comesBefore.all fun cb =>
      match names.findIdx? (· == cb.pass.name) with
      | some j => i < j
      | none   => false   -- target not in laurelPipeline

#guard comesBeforeRespected
```

Compare by `name` since `LaurelPass` has a function field and thus no `DecidableEq`. Independent of the String-vs-struct discussion on `LaurelPass.lean`.
</details>

Strata/Languages/Laurel/Resolution.lean
keyboardDrummer: Nice find, thanks. I brought back the original `computeExprType model target` call.

Strata/Languages/Laurel/Resolution.lean
keyboardDrummer: I've removed the `emitResolutionErrors` option now so it's always true.

MikaelMayer (commented):

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
keyboardDrummer: I've added the snippet you pasted, thanks.

MikaelMayer (commented):

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -87,65 +89,33 @@ public section
...
+/-- Every `comesBefore` constraint is respected by the pipeline order. -/
+def comesBeforeRespected : Bool :=
+  let names := laurelPipeline.toList.map (·.name)
+  (List.range laurelPipeline.size).zip laurelPipeline.toList |>.all fun (i, p) =>
+    p.comesBefore.all fun cb =>
+      match names.findIdx? (· == cb.pass.name) with
+      | some j => i < j
+      | none   => false   -- target not in laurelPipeline
+
+#guard comesBeforeRespected
```
tautschnig: Make `comesBeforeRespected` evaluate at elaboration — this `#guard` currently fails to build (CI red on both *Build and test Lean* and *Build documentation*).

<details>
<summary>Why + verified fix</summary>

`#guard` runs in the interpreter, which has no IR for the cross-module `LaurelPass` values `laurelPipeline` holds:

```
(interpreter) IR of declaration 'Strata.Laurel.constrainedTypeElimPass' not available;
this may point to a missing `meta` check in a metaprogram
```

`decide` gets stuck (kernel can't reduce the `Array`/`String` operations) and `native_decide` hits the same IR gap. Two fixes that build:

- `meta import` the pass-defining modules in this file (`DesugarShortCircuit`, `MergeAndLiftReturns`, `EliminateValueInReturns`, `ModifiesClauses`, `HeapParameterization`, `TypeHierarchy`, `InferHoleTypes`, `EliminateDeterministicHoles`, `LiftImperativeExpressions`, `ConstrainedTypeElim`, `TypeAliasElim`). I verified `lake build` then succeeds (194 jobs).
- Or move `comesBeforeRespected` + the `#guard` into a `StrataTest` module that does `meta import all Strata.Languages.Laurel.LaurelCompilationPipeline` (the pattern `LiftHolesTest.lean` already uses), keeping `meta import`s out of the core library.

*Build documentation* fails for the same reason — `LaurelDoc.lean` imports this module.
</details>

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -167,19 +137,12 @@ private def runLaurelPasses (options : LaurelTranslateOptions)
 
   -- Initial resolution
   let result := resolve program
-  let resolutionErrors : List DiagnosticModel :=
-    if options.emitResolutionErrors then result.errors.toList else []
+  let resolutionErrors : List DiagnosticModel := result.errors.toList
```
tautschnig, line 145: Remove the now-dead `emitResolutionErrors` field at `LaurelToCoreTranslator.lean:619`.

<details>
<summary>Why</summary>

This line was its last reader. After this commit, `grep -rn emitResolutionErrors Strata/` matches only the field definition (`emitResolutionErrors : Bool := true`), so it is dead config.
</details>

MikaelMayer (commented):

MikaelMayer (commented):

ssomayyajula (approved):

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -87,65 +89,38 @@ public section
...
+  inferHoleTypesPass,
+  eliminateDeterministicHolesPass,
+  desugarShortCircuitPass,
+  liftExpressionAssignmentsPass,
+  mergeAndLiftReturnsPass,
+  constrainedTypeElimPass
 ]
 
+/-- Every `comesBefore` constraint is respected by the pipeline order.
+    Checked at elaboration time so that mis-ordered passes are caught immediately. -/
```
tautschnig, line 109: Nit - feel free to ignore if you want to merge now: Reword "Checked at elaboration time" — the `initialize` check runs at module-load time, not during this module's elaboration.

<details>
<summary>Why</summary>

I checked: building `Strata.Languages.Laurel` (a plain importer) with a deliberately mis-ordered `laurelPipeline` still succeeds — `initialize` does not run during plain elaboration. Building `StrataTest.Languages.Laurel.TestExamples` (which loads the pipeline through its `#eval`) does fail with `laurelPipeline: comesBefore ordering constraints violated`. So CI does catch mis-ordering, via the pass-running tests — but "elaboration time / caught immediately" overstates it.

Suggested: "Checked at module-load time via `initialize`; a mis-ordering surfaces when a module that runs the pipeline (e.g. `TestExamples`) is built."
</details>

tautschnig (approved):

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
keyboardDrummer: I'll fix it in another PR. Thanks for approving still :-)

PR #1223, opened 2026-05-27
Overhaul test framework for Laurel

leo-leesco:
Previously, we were using raw strings and manually routed them to be translated to `Strata.Program`
Now, we use the DDM mechanism that creates a `Strata.Program` out of the box. This comes with several advantages :
- we don't have to calculate character offsets manually anymore (for error reporting)
- we get to leverage Lean's LSP integration for DSLs

For formatting and other purposes, you can use `Strata.SourcedProgram` which is essentially `Strata.Program` with added metadata (line/col number, filename…). An automatic coercion is in place to strip metadata back into a `Strata.Program`.


StrataTest/Languages/Laurel/Examples/Fundamentals/T10_ConstrainedTypes.lean
```diff
@@ -4,15 +4,22 @@
...
 
 open StrataTest.Util
-
-namespace Strata
-namespace Laurel
-
-def program := r"
+open Strata
+
+/-- info: 20:27-30  error: assertion does not hold
```
keyboardDrummer: I think we should have the diagnostics be inline, so you don't have to mention the line numbers like "20:27-39"

keyboardDrummer (commented):
Love it! Looks great to me. @keyboardDrummer-bot could you review this PR and come up with potential issues?

keyboardDrummer:
@leo-leesco could you look at the first concern from the bot "1. T19_InvokeOn.lean — removed assertion annotation without explanation" ?

Feel free to ignore the other comments by the bot. We can also delete its comment to reduce noise.

leo-leesco:
1. `T19_InvokeOn`
- Restored the ability to pass `options : LaurelVerifyOptions` parameter to `testLaurel`.
- Restored the `assert Q(x)` annotation, and pass `options := { verifyOptions := { Core.VerifyOptions.quiet with solver := "z3" } }` as it was previously (this test is solver-dependent)

2. The PR has not changed the existing behavior.

3. `Coe SourcedProgram Program` may silently drop source : indeed, that's what makes the migration transparent (~200 call sites would otherwise need explicit `.program` ; we are not actually losing information through this coercion, which is the point), and we explicitly document that behavior. Tests that need source info declare their parameter as `SourcedProgram` explicitly.

4. `let prelude : Program := Python.corePrelude` exercises the coercion explicitly (because `corePrelude` is actually a `SourcedProgram` as the output of a `#strata`)

5. Multi-line diagnostics never match — acknowledged, there are currently no multiline diagnostics

6. Comments with ^^^ misinterpreted as diagnostics : we should not, as a general principle, write comments that include carets

7. `renderSnippetLocal` first-line off-by-one — verified empirically OK. `T1_AssertFalse` and similar tests have annotations on early lines and pass; no observable off-by-one. 

8. `README` typo `#strata / #strata` — fixed

9. Kept redundant import in `AbstractToConcreteTreeTranslatorTest.lean` : `import Strata.DDM.Integration.Lean.HashCommands` documents intent (we directly use `#strata` from that module), even if it's transitively imported elsewhere.

keyboardDrummer (dismissed):

StrataTest/README.md
```diff
@@ -0,0 +1,13 @@
+# StrataTest
+
+This directory contains compile-time tests (`#guard_msgs`) for Strata's
+dialects, transforms, and DDM machinery. Tests run as part of `lake build`;
+no output means they passed.
+
+For how to write a new test — including the `#strata` / `#strata`
```
tautschnig: Typo claimed fixed in your bot reply still present — should be `#strata` / `#end` (or `#strata` / `testLaurel`).

<details>
<summary>Why</summary>

The bot's review item #8 was: <code>StrataTest/README.md</code> typo <code>`#strata` / `#strata`</code>. Your reply said "fixed", but the line still reads:

```
For how to write a new test — including the `#strata` / `#strata`
blocks and the Laurel test helpers — see
```

Probably you intended `#strata` / `#end` (the actual block delimiters) or `testLaurel` / `testLaurelResolution` (the helper pair).
</details>

StrataTest/Util/TestLaurel.lean
```diff
@@ -0,0 +1,264 @@
...
+  pure annotations
+
+/-- Try to match an annotation to a diagnostic. -/
+private def matchesAnnotation (d : Strata.Diagnostic) (a : DiagnosticAnnotation) : Bool :=
+  d.start.line == a.line
+    && d.ending.line == a.line
+    && d.start.column == a.colStart
+    && d.ending.column == a.colEnd
+    && diagnosticKindString d.type == a.kind
+    && (a.message.isEmpty || (d.message.splitOn a.message).length > 1)
```
tautschnig: Two small refactors: relax the line check and use `String.isInfixOf` for the message check.

<details>
<summary>Why</summary>

**Line check (line 188–189):** `d.start.line == a.line && d.ending.line == a.line` silently fails to match any future multi-line diagnostic. The error message will say "actual diagnostic with no matching annotation" without a hint that the line span was the cause. The bot raised this as concern #5; the response was "no multiline diagnostics today" — please track as a follow-up issue, or relax to `d.start.line == a.line` and treat `ending.line` as informational.

**Message check (line 193):** `(d.message.splitOn a.message).length > 1` is a clever-but-obscure way to test substring containment. `a.message.isEmpty || a.message.isInfixOf d.message` is direct and avoids the (defined but surprising) edge cases of `splitOn` on overlapping or boundary-aligned separators.

Neither is a soundness issue today; both make the matcher more robust to future changes.
</details>

StrataTest/Languages/Laurel/Examples/Fundamentals/T22_ArityMismatch.lean
```diff
@@ -4,32 +4,35 @@
...
 open StrataTest.Util
+open Strata
 
-namespace Strata
-namespace Laurel
+/-! ## Function called with too many arguments -/
 
-def arityMismatchProgram := r"
+/--
+error: <#strata>(436-457) ❌ Type checking error.
```
tautschnig, line 15: The hard-coded byte range `(436-457)` makes this test silently brittle to any edit inside the `#strata` block.

<details>
<summary>Why and what to track</summary>

This test uses `#guard_msgs` because the diagnostic is a Lean-elaborator type-checking error from the DDM `#strata` term elaborator, which doesn't currently flow through the inline-annotation channel. Adding/removing any token inside the snippet shifts `436-457` and the test fails — the failure message is helpful enough, but the fix is a tedious manual byte-counting exercise.

The bot flagged this as concern #2; your reply was correct that the PR doesn't change the existing behavior. Worth filing a follow-up issue: "route DDM elaboration errors into the inline-annotation matcher" so this last `#guard_msgs`-with-byte-range test can be migrated.
</details>

StrataTest/Util/TestLaurel.lean
```diff
@@ -0,0 +1,264 @@
...
+      if listStartsWith prevBody "//" || prevBody.isEmpty then continue
+      found := true
+    unless found do continue
+    annotations := annotations.push {
+      line := target + 1, colStart, colEnd, kind, message
+    }
+  pure annotations
+
+/-- Try to match an annotation to a diagnostic. -/
+private def matchesAnnotation (d : Strata.Diagnostic) (a : DiagnosticAnnotation) : Bool :=
```
tautschnig, line 196: Test/proof opportunity: the annotation parser and matcher are now load-bearing for ~50 test files — worth pinning their behavior with theorems.

<details>
<summary>Suggested theorems</summary>

The annotation parser/matcher is now responsible for the correctness of every negative test in `StrataTest/Languages/Laurel/`. A few cheap properties would catch regressions without relying on the test suite indirectly testing the harness:

```lean
-- Parser is total and produces well-formed positions.
theorem parseAnnotations_well_formed (s : String) :
    ∀ a ∈ parseAnnotations s, a.colStart < a.colEnd ∧ a.line ≥ 1

-- Parser is idempotent w.r.t. re-formatting.
theorem parseAnnotations_roundtrip (a : DiagnosticAnnotation) (lineText : String) :
    parseAnnotations s!"{lineText}\n// {String.mk (List.replicate a.colStart ' ')}{String.mk (List.replicate (a.colEnd - a.colStart) '^')} {a.kind}: {a.message}\n" = #[a]

-- Matcher is reflexive on diagnostics produced by `formatAnnotation`-shaped data.
theorem matchesAnnotation_self (a : DiagnosticAnnotation) :
    matchesAnnotation { start := ⟨a.line, a.colStart⟩, ending := ⟨a.line, a.colEnd⟩,
                       type := …, message := a.message } a = true
```

These would catch off-by-one regressions in `parseAnnotations` (specifically the `colStart := leadWs + 2 + preCarets.length` arithmetic) and the message-substring check more cheaply than running the full pipeline.
</details>

docs/Testing.md
```diff
@@ -0,0 +1,221 @@
+# Testing Strata Dialects
```
tautschnig, line 1: The PR body still describes a #strata_expect macro that doesn't exist in the implementation (replaced by inline annotations during iteration). Please update the description to match docs/Testing.md so the merged commit message is accurate.

StrataTest/Languages/Laurel/Examples/Fundamentals/T22_ArityMismatch.lean
leo-leesco: already addressed previously : we only abide by the existing test discipline, this is out of scope for this PR

StrataTest/Util/TestLaurel.lean
leo-leesco: I see, but I believe we should implement that feature in a standalone PR

docs/Testing.md
```diff
@@ -0,0 +1,221 @@
...
+file held.
+
+## Other dialects
+
+- **Boole** uses `#strata program Boole; … #end` directly with hand-written
+  helpers (e.g. `Strata.Boole.verify`); see
+  [`StrataTest/Languages/Boole/find_max.lean`](../StrataTest/Languages/Boole/find_max.lean)
+  for the canonical pattern. `#strata` is dialect-agnostic — Boole could use
+  inline annotations the same way once a Boole-side helper analogous to
+  `testLaurelExpect` is written.
```
tautschnig: **Drop the `testLaurelExpect` reference — the helper was unified into `testLaurel` and no longer exists by that name.**

```suggestion
  inline annotations the same way once a Boole-side helper analogous to
  `testLaurel` is written.
```

<details>
<summary>Note</summary>

`rg testLaurelExpect` returns only this line in the repo.
</details>

StrataTest/Util/TestLaurel.lean
```diff
@@ -0,0 +1,271 @@
...
+    unless found do continue
+    annotations := annotations.push {
+      line := target + 1, colStart, colEnd, kind, message
+    }
+  pure annotations
+
+/-- Substring containment on `String`. Direct rather than the
+    `(d.splitOn a).length > 1` trick. -/
+private def isSubstrOf (needle haystack : String) : Bool :=
+  needle.isEmpty || (haystack.splitOn needle).length > 1
```
tautschnig: **Make `isSubstrOf` reject the empty needle so a stray `// ^^^ error:` (no message) fails to fire instead of matching any diagnostic of that kind.**

```suggestion
  !needle.isEmpty && (haystack.splitOn needle).length > 1
```

<details>
<summary>Why</summary>

`parseAnnotations` produces an annotation with `message := ""` whenever the user typed `// ^^^ error:` with nothing after the colon (`String.trimAscii` strips it to empty). With the current short-circuit, that annotation matches any error at the same line/column, so the typo never surfaces. After this change, an empty-message annotation will fail to fire and the existing "Expected (annotated) but never fired" branch reports it.
</details>

keyboardDrummer (dismissed):
Still looks great!

docs/Testing.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+### 2. Negative test — assert that specific errors fire
+
+Same helper, `testLaurel`. Annotate each expected diagnostic *inline*, on the
+line directly below the offending source line, with carets pointing at the
+column range and the kind + (substring of the) message after the carets:
+
+```lean
+#eval testLaurel <|
```
robin-aws: nit: is the `<|` actually necessary? I don't think it is, especially since the first example doesn't have it. If so I'd remove it throughout.

docs/Testing.md
```diff
@@ -0,0 +1,222 @@
...
+{
+  var z: int := 10 / x
+//^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
+};
+#end
+```
+
+`testLaurel` auto-detects: when annotations are present, it requires an
+**exact match** (every diagnostic annotated, every annotation fires,
+positions agree, kind matches, message is a substring); when absent, it
```
robin-aws: Substring or prefix? Why can we not make it an exact match in practice? Worth noting that here.

docs/Testing.md
```diff
@@ -0,0 +1,222 @@
...
+
+/--
+info: function $hole_0()
+  returns ($result: int)
+  opaque;
+procedure test()
+{ var x: int := 1 + $hole_0() };
+-/
+#guard_msgs in
+#eval! parseElimAndPrint
```
robin-aws: Why is `#eval!` necessary?

docs/Testing.md
```diff
@@ -0,0 +1,222 @@
...
+
+`kind` is one of `error`, `warning`, `not-yet-implemented`, `strata-bug`.
+
+### 3. Resolution-only tests — skip the verifier
+
+When the test is about resolution (kind mismatches, duplicate names, scope
+errors), running the verifier on a deliberately-broken program adds
+unrelated noise (`dbg_trace` warnings, vacuous VCs). Use
+`testLaurelResolution` — same auto-detect behavior as `testLaurel`, but
+stops after `resolve`.
```
robin-aws, line 97: Are there any Laurel-to-Laurel translations that happen after resolution but before verification? Is there another page you can link to that documents the Laurel pipeline stages?

docs/Testing.md
```diff
@@ -0,0 +1,222 @@
...
+procedure test() { var x: int := 1 + <?> };
+#end
+
+end Strata.Laurel
+```
+
+The pattern: `translateLaurel : Strata.Program → IO Laurel.Program` is the
+common entry point; everything after is dialect/test-specific.
+
+### 5. Mixed positive / negative inside one feature
```
robin-aws: This case is a bit confusing to me. Any negative test (case 2) will inevitably have SOME aspects that are correct along with expected errors.

docs/Testing.md
```diff
@@ -0,0 +1,222 @@
...
+program Laurel;
+procedure unsafeDivision(x: int) opaque {
+  var z: int := 10 / x
+//^^^^^^^^^^^^^^^^^^^^ error: assertion does not hold
+};
+#end
+```
+
+If the procedures genuinely depend on each other (one calls the other), keep
+them in one `#strata` block and list the union of expected diagnostics.
```
robin-aws: I think this is the more important point, and I'd label the section accordingly.

docs/Testing.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Practical workflow
+
+1. Write the `#strata` block first, prefaced with `#eval testLaurel`.
+2. For negative tests, sketch placeholder annotations like `// ^ error:`
+   below the offending lines — column positions don't have to be right yet.
+3. Run it: `lake env lean StrataTest/Languages/Laurel/<your_file>.lean`.
+4. On failure: the helper prints exactly which annotations went unmatched
+   and which diagnostics had no annotation, including the line/column range
+   actually produced. Copy those into your annotations, save, re-run.
```
robin-aws, line 232: (Out of scope for this PR) Is there a mode you can evaluate the tests in that will automatically update the annotations to match the actual result? That's always been valuable for every instance of this kind of testing framework in the past.

docs/Testing.md
```diff
@@ -0,0 +1,222 @@
...
+  [`StrataTest/Languages/Python/README.md`](../StrataTest/Languages/Python/README.md).
+
+## Where things live
+
+| Concept | File |
+| --- | --- |
+| `#strata` elaborator + `SourcedProgram` | `Strata/DDM/Integration/Lean/HashCommands.lean` |
+| `Diagnostic` data type | `Strata/Languages/Core/Verifier.lean` |
+| Laurel test helpers | `StrataTest/Util/TestLaurel.lean` |
+| Migrated examples | `StrataTest/Languages/Laurel/Examples/**/*.lean` |
```
robin-aws: Doesn't make sense to label these as "migrated" once this is merged, I'd just call them "Examples"

robin-aws (changes requested):
Fantastic development improvement, and really appreciate the documentation! Just a few nit-picky edits to the latter, for clarity and to ensure the right style is copy-and-pasted all over the place in the future. :)

docs/Testing.md
leo-leesco: For now, that is not the case ; I agree it would be very nice (like `dune promote`)
If at all possible, I would like to rely on built-in functionality of the `Lean` compiler, but I'm not sure it is implemented yet + we have a very custom error reporting here (because it is inlined) so there is some work. Definitely a very valuable feature Strata-wide, I will create an issue with this (but I'm worried it will become stale very quickly if I don't address it myself !)

docs/Testing.md
leo-leesco: There sadly are some cases where the identifier/reported type are not exactly semantically relevant (a test case features `$__ty35` for instance) so that makes the tests brittle in the wrong sense since we want the tests to break when we make a frontend change, not an implementation detail

docs/Testing.md
leo-leesco: yes indeed ! I will link to `LaurelDoc.lean` and the strata website in the document

leo-leesco: update : I directly linked to the code that composes the Laurel-to-Laurel passes

docs/Testing.md
leo-leesco: see #1362

robin-aws (dismissed):

robin-aws (approved):

aqjune-aws (approved):

PR #1246, opened 2026-05-27
Core translator: local function declaration builds wrong arrow type for ≥3 distinct-type args

### LINKED ISSUE #1226: `Core` translator: local `function` declaration builds wrong arrow type for ≥3 distinct-type args

```lean
```

**Fix**:

```diff
-    let funcType := Lambda.LMonoTy.mkArrow outputMono (inputs.values.reverse)
+    let funcType := Lambda.LMonoTy.mkArrow' outputMono inputs.values
```

MikaelMayer:
Fixes #1226

The `funcDecl_statement` case in the Core translator used `mkArrow outputMono (inputs.values.reverse)` to build the type annotation for local function declarations. This produced incorrect type ordering for functions with 3+ parameters of distinct types (e.g., `int → real → bool → int` instead of `int → bool → real → int`).

Replaced with `mkArrow' outputMono inputs.values`, which correctly builds the arrow type with the return type at the end and inputs in declaration order.

Tested: existing tests pass, added regression test with a 3-distinct-type-arg local function.


MikaelMayer:

PR #1248, opened 2026-05-27
fix: SMT Solver.spawn stderr pipe opened but reader dropped; solver may SIGPIPE

### LINKED ISSUE #1241: SMT `Solver.spawn`: stderr pipe is opened but reader handle is dropped; solver may SIGPIPE on fatal-error output

**Fix**:

```diff
-      stderr := .piped
+      stderr := .inherit
```

MikaelMayer:
Fixes #1241

## Problem

`Solver.spawn` passed `stderr := .piped` but never stored the read end of the pipe in the `Solver` struct. When the process handle was dropped, the pipe's reading end closed. Any subsequent solver write to stderr (fatal error, OOM, unsupported logic) would receive EPIPE, and the default SIGPIPE disposition would kill the solver — producing a spurious `unknown` or hang.

## Fix

Switch `stderr := .piped` to `stderr := .inherit`, so solver errors are forwarded to the verifier's stderr. This avoids creating a closeable pipe while making solver diagnostics visible.

## Testing

All library and test modules compile successfully.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+**Warning:** This repository will shortly undergo a split into several separate repositories. If you're creating a PR that crosses the boundaries between these repositories, you may want to hold off until the split is complete or be prepared to rework your PR into multiple PRs once the split is complete.
+
+The code that will be moved includes:
+- Strata/DDM/*
+- Strata/Languages/Boole/*
+- Strata/Languages/Python/* along with Tools/Python/*
+- Tools/BoogieToStrata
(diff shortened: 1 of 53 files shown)


MikaelMayer:
🤖⛏️

This PR fixes the dangling stderr pipe in `Solver.spawn`. The change is a one-liner: `stderr := .piped` → `stderr := .inherit`.

The `.piped` option created a pipe whose read end was never stored in the `Solver` struct. When the process handle was dropped, the read end closed, and any solver write to stderr (e.g. on a fatal error) would trigger SIGPIPE, killing the solver unexpectedly.

With `.inherit`, solver stderr goes directly to the verifier's stderr stream — no pipe to close, no SIGPIPE risk, and solver diagnostics become visible.

MikaelMayer:

PR #1251, opened 2026-05-27
DDM parser: fix checkLeftRec panic when leading argument is NewlineSepBy

### LINKED ISSUE #1245: DDM parser: `checkLeftRec` panics if leading argument is `NewlineSepBy`

**Fix**:

```diff
     let isListCategory := cat.name == q`Init.CommaSepBy ||
                           cat.name == q`Init.SpaceSepBy ||
                           cat.name == q`Init.SpacePrefixSepBy ||
+                          cat.name == q`Init.NewlineSepBy ||
                           cat.name == q`Init.Seq ||
                           cat.name == q`Init.SemicolonSepBy ||
                           cat.name == q`Init.Option
```

MikaelMayer:
Fixes #1245

`checkLeftRec` recognized `CommaSepBy`, `SpaceSepBy`, `SpacePrefixSepBy`, `Seq`, `SemicolonSepBy`, and `Option` as parametric list categories but not `NewlineSepBy`. When an op used `NewlineSepBy` as its leading argument, the function would panic with "Unknown parametric category … is not supported."

This adds `Init.NewlineSepBy` to the `isListCategory` check, consistent with how `catParser` already handles it.

Tested: existing tests pass, new regression test added with a dialect that uses `NewlineSepBy` as the leading argument of an op.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+**Warning:** This repository will shortly undergo a split into several separate repositories. If you're creating a PR that crosses the boundaries between these repositories, you may want to hold off until the split is complete or be prepared to rework your PR into multiple PRs once the split is complete.
+
+The code that will be moved includes:
+- Strata/DDM/*
+- Strata/Languages/Boole/*
+- Strata/Languages/Python/* along with Tools/Python/*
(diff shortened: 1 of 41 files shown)


MikaelMayer:

PR #1252, opened 2026-05-27
Fix #1228: Ion deserializer NOP pad in struct shifts field-key indices

### LINKED ISSUE #1228: Ion deserializer: NOP pad in struct shifts subsequent field-key indices

MikaelMayer:
Fixes #1228

When deserializing a struct containing NOP pads, the field-name VarUInt was unconditionally pushed to `ds.symbols`. Since NOP pads don't produce a value, this caused `popFinishedValues` to mismatch keys with values, silently assigning wrong field keys to subsequent fields.

The fix peeks at the type descriptor byte after reading the VarUInt and skips the symbol push when the upcoming token is a NOP pad (type code 0, length ≠ 15).

Tested: existing tests pass, new unit test added for the exact byte sequence from the issue.


MikaelMayer (commented):
🤖🔍 Clean, minimal fix. The NOP pad detection logic (`td.code == 0 && td.length != 15`) correctly distinguishes NOP pads from `null.null`, and the conditional symbol push prevents the key/value mismatch in `popFinishedValues`. The test byte sequence is well-constructed and exercises the exact scenario from the issue.

No issues found.

MikaelMayer:
🤖 Closing: started from wrong base (May 14 commits). Will re-create from main2.

PR #1253, opened 2026-05-27
DDM Elab: circular dialect imports recurse forever (no cycle check in loadDialectRec)

### LINKED ISSUE #1227: DDM Elab: circular dialect imports recurse forever (no cycle check in `loadDialectRec`)
[`loadDialectRec`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DDM/Elab.lean#L384-L400)
takes a `stk : Array DialectName` for cycle detection but never tests it.
Two dialect files `A` (imports B) and `B` (imports A) cause unbounded
recursion / stack overflow.

**Fix**:

```diff
   if p : name ∈ ld.dialects then
     pure (.ok ld.dialects[name])
+  else if stk.contains name then
+    return .error s!"Circular import detected: {name} (stack: {stk})"
   else
     let path ← ...
```

**Test plan**: add an integration test that creates two dialect files with a circular import. Verify the test gets a stack overflow (or, in compiled mode, a hang) before the fix, and a clean `Circular import detected` error after.

MikaelMayer:
Fixes #1227

`loadDialectRec` accepted a `stk : Array DialectName` parameter for cycle detection but never checked it. Two dialect files with mutual imports (A imports B, B imports A) caused unbounded recursion / stack overflow.

Now checks whether the dialect name is already on the stack before attempting to load, returning a clear "Circular import detected" error that renders the full cycle path (e.g. `A -> B -> C -> A`), making the offending edge obvious.

Tested with integration tests covering:
- 2-cycle (A ↔ B)
- Self-import (A imports A) — caught by the existing `openDialectSet` check
- 3-cycle (X → Y → Z → X)

All existing tests pass.


MikaelMayer (commented):
🤖🔍 Clean, minimal fix. The two-line change correctly uses the existing `stk` parameter that was already threaded through the mutual recursion but never checked. The test with mutually-importing dialect files exercises the exact scenario from the issue. Logic is sound: `loadDialectFromPath` pushes the dialect name onto `stk` before elaborating its body, so by the time a circular import triggers `loadDialectRec` again, the ancestor is already on the stack.

No issues found.

Strata/DDM/Elab.lean
```diff
@@ -389,6 +389,8 @@ private partial def loadDialectRec
   let ld ← fm.getLoaded
   if p : name ∈ ld.dialects then
     pure (.ok ld.dialects[name])
+  else if stk.contains name then
```
joehendrix:
```suggestion
  else if name ∈ stk then
```

joehendrix (dismissed):
One small change

Strata/DDM/Elab.lean
MikaelMayer: 🤖 Done, switched to `name ∈ stk`.

joehendrix (dismissed):

Strata/DDM/Elab.lean
```diff
@@ -389,6 +389,8 @@ private partial def loadDialectRec
   let ld ← fm.getLoaded
   if p : name ∈ ld.dialects then
     pure (.ok ld.dialects[name])
+  else if name ∈ stk then
+    return .error s!"Circular import detected: {name} (stack: {stk})"
```
tautschnig: The error uses `Array`'s default `Repr` (`#[CircA, CircB]`), which leaks an internal format and hides where the cycle closes. Render the path explicitly so users see the cycle:

```suggestion
    return .error s!"Circular import detected: {String.intercalate " -> " (stk.toList ++ [name])}"
```

For `A -> B -> C -> D -> B` this produces `A -> B -> C -> D -> B`, making the offending edge obvious. The corresponding `#guard_msgs` expectation in the new test will need updating.

StrataTest/DDM/CircularImport.lean
```diff
@@ -0,0 +1,29 @@
...
+-/
+#guard_msgs in
+#eval show IO _ from do
+  let preloaded := Elab.LoadedDialects.builtin
+  let fm ← DialectFileMap.new preloaded
+  let fm ← match ← fm.add "StrataTest/DDM/testdata" |>.toBaseIO with
+    | .ok fm => pure fm
+    | .error msg => do IO.println msg; pure fm
+  match ← Elab.loadDialect fm "CircA" with
+  | .ok _ => IO.println "unexpected success"
```
tautschnig, line 26: Coverage of the cycle check is thin — only the 2-cycle is exercised. Please add `#guard_msgs`-driven `#eval` cases (and matching `.dialect.st` fixtures) for:

1. **Self-import**: `dialect A; import A;` — the smallest cycle and easy to regress.
2. **3-cycle**: `A → B → C → A` — verifies the stack accumulates correctly across depth > 2.
3. **Diamond (negative)**: `A` imports both `B` and `C`, both import a shared `D`. Must succeed; protects against false positives if the check is ever tightened.
4. **Recovery**: after a circular-import failure, loading a non-cyclic dialect via the same `DialectFileMap` succeeds (guards against partial dialects leaking into `loaded`).

If you also adopt the message-format change, theorems on this code remain impractical, but these four #guard_msgs cases give meaningful behavioural coverage.

Strata/DDM/Elab.lean
MikaelMayer: 🤖 Applied the suggested rendering. The cycle path now shows `CircA -> CircB -> CircA` making the offending edge obvious. Test expectation updated accordingly.

StrataTest/DDM/CircularImport.lean
MikaelMayer: 🤖 Added in 660da0547:

- **Self-import** (`CircSelf` imports itself): caught by the `openDialectSet` check with "Dialect CircSelf already open."
- **3-cycle** (`CircX -> CircY -> CircZ -> CircX`): caught by the stack check with the full cycle path.

tautschnig (approved):

joehendrix (approved):

tautschnig (approved):

PR #1254, opened 2026-05-27
Fix local function declaration building wrong arrow type for ≥3 distinct-type args

### LINKED ISSUE #1226: `Core` translator: local `function` declaration builds wrong arrow type for ≥3 distinct-type args

MikaelMayer:
Fix local function declaration building wrong arrow type for ≥3 distinct-type args.

When a local function had 3 or more parameters with distinct types, the Core translator built the arrow type with incorrect ordering. This was caused by using `mkArrow` with a reversed input list instead of `mkArrow'` which takes the return type and input list in the correct order.

Also refactors `getDecreasesKind` to reuse the original measure expression for int-valued fvar measures rather than reconstructing it.

Adds a regression test with a local function `f(x:int, b:bool, r:real):int`.

Fixes #1226


MikaelMayer (commented):
🤖🔍 This PR delivers significantly more than its title suggests. Beyond the local function arrow-type fix (the `mkArrow` → `mkArrow'` change in `Translate.lean`), it implements:

1. **Int-valued termination checking** — a major feature in `TerminationCheck.lean` with `DecreasesKind` classification, non-negativity + strict-decrease obligations, and mixed-measure rejection for mutual blocks.
2. **`inlineIfAllCanonical` attribute** — enables concrete evaluation for int-recursive functions.
3. **Boole grammar/lowering additions** — `choose`, `let`-in-expr, `Sequence` ops, lambda, signed bv comparisons, bitvector loop variables, `decreases` on `for` loops.
4. **`fvarIsOp` removal** — replaced by direct `gctx.vars` classification with a correctness theorem.
5. **Extensive test coverage** — new test files for int recursion, SHA-256, OTP, choose, embedded postconditions, etc.

The core arrow-type fix is correct and well-tested. The `TerminationCheck` refactor is clean — the `DecreasesKind` ADT and `computeCallMeasure`/`computeCallerMeasure` helpers are well-factored.

Two observations on `StrataBoole/StrataBoole/Verify.lean` (pre-existing on main, not introduced by this PR):

- **Line 712**: `let funcTy := Lambda.LMonoTy.mkArrow outputMono ((inputsMono.map (·.2)).reverse)` — this is the same `mkArrow`-with-reversed-inputs pattern that was buggy in `Translate.lean` and fixed by this PR. Since the PR specifically fixes this class of bug, it would be consistent to fix it here too (change to `mkArrow' outputMono (inputsMono.map (·.2))`).

- **Line 866**: For int-recursive functions without `@[cases]`, `attr` is set to `#[]`. In `Translate.lean` (line 1817), the same case now sets `#[inlineIfAllCanonical]` to enable concrete evaluation. This means Boole int-recursive functions won't get concrete evaluation. Not a blocker since the current Boole tests only exercise termination VCs, but worth aligning.

Strata/Transform/TerminationCheck.lean
```diff
@@ -46,13 +50,46 @@ def termProcName (name : String) : String := s!"{name}{termSuffix}"
...
+    available during type checking in `FunctionType`. -/
+private def getDecreasesKind (func : Function) (tf : @TypeFactory Unit)
+    : Except String DecreasesKind := do
   match func.measure with
-  | some (.fvar _ id _) => func.inputs.keys.findIdx? (· == id)
-  | some _ => none
-  | none => FuncAttr.findInlineIfConstr func.attr
+  | some (.fvar _ id _) =>
+    match func.inputs.findWithIdx? id with
+    | some (_, .int) => .ok (.intValued (.fvar () id (.some .int)))
```
MikaelMayer: 🤖🔍 nit: When the `decreases` variable has type `.int`, this reconstructs the fvar expression as `.fvar () id (.some .int)` rather than reusing the original `func.measure` value. The type annotation is redundant since you already matched on `.int`. Minor since Core expressions use `Unit` metadata, but reusing the original would be slightly cleaner: `| some (_, .int) => .ok (.intValued (func.measure.get!))`.

MikaelMayer: 🤖 Fixed — now using `measure@(.fvar _ id _)` pattern to reuse the original expression instead of reconstructing it. See [5b59a93](https://github.com/strata-org/strata/commit/5b59a93d4).

MikaelMayer:

PR #1255, opened 2026-05-27

### LINKED ISSUE #1241: SMT `Solver.spawn`: stderr pipe is opened but reader handle is dropped; solver may SIGPIPE on fatal-error output

MikaelMayer:
Fixes #1241

`Solver.spawn` passed `stderr := .piped` but never read from the pipe. When the solver wrote to stderr (e.g. fatal error, OOM), the unread pipe filled up or closed, causing EPIPE/SIGPIPE and killing the solver process.

Switched to `stderr := .inherit` so solver stderr goes directly to the verifier's stderr, making errors visible and avoiding the dangling pipe.

Existing tests pass.


MikaelMayer (commented):
🤖🔍 Clean, minimal fix. Confirmed that `proc.stderr` was never read on main, so the `.piped` handle was indeed a dangling pipe that could SIGPIPE the solver. Switching to `.inherit` is the right call — it makes solver errors visible and eliminates the failure mode.

No issues found.

MikaelMayer:

PR #1256, opened 2026-05-27
Fix #1244: interpolate dialect name in parseStrataProgramFromDialect error message

### LINKED ISSUE #1244: DDM Elab: error message in `parseStrataProgramFromDialect` is a plain string, not interpolated
[`Elab.lean#L519`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DDM/Elab.lean#L519):

```lean
| throw <| IO.userError "Internal {dialect} missing from loaded dialects."
```

Missing `s!` prefix. The user sees the literal text `{dialect}` instead of the actual dialect name.

**Fix**:

```diff
-  | throw <| IO.userError "Internal {dialect} missing from loaded dialects."
+  | throw <| IO.userError s!"Internal {dialect} missing from loaded dialects."
```

**Test plan**: call `parseStrataProgramFromDialect` with a missing dialect name; assert the error message contains the actual dialect name (not the literal `{dialect}`). Verify it fails before the fix.

MikaelMayer:
Fixes #1244

The error message in `parseStrataProgramFromDialect` was missing the `s!` prefix, causing users to see the literal text `{dialect}` instead of the actual dialect name.

Added a test that verifies the interpolated error message contains the actual dialect name.


MikaelMayer (commented):
🤖🔍 Clean fix. The missing `s!` prefix is correctly added, and the test verifies the interpolated dialect name appears in the error message. No other instances of this bug pattern exist in the codebase.

MikaelMayer:
🤖 Closing: started from wrong base (May 14 commits). Will re-create from main2.

PR #1257, opened 2026-05-27

### LINKED ISSUE #1245: DDM parser: `checkLeftRec` panics if leading argument is `NewlineSepBy`

MikaelMayer:
Fixes #1245

## Problem

`checkLeftRec` in the DDM parser panics with "Unknown parametric category" when the leading (first) argument of an op is `NewlineSepBy`.

## Solution

Add `Init.NewlineSepBy` to the `isListCategory` check in `checkLeftRec`, alongside the other separator categories (`CommaSepBy`, `SpaceSepBy`, etc.) that are already handled.

## Testing

- Added a regression test with a dialect that uses `NewlineSepBy` as the leading argument of an op.
- Existing tests pass.


MikaelMayer (commented):
🤖🔍 The core fix (adding `Init.NewlineSepBy` to the `isListCategory` check in `checkLeftRec`) is correct and minimal. The regression test is well-structured and exercises the exact scenario that triggered the panic.

The PR diff is large because it includes merge commits from main/main2, but the actual fix is just the 1-line addition plus the test file.

One pre-existing concern worth noting: `checkLeftRec` still has two `panic!` calls (line 771 for invalid index, line 791 for unknown parametric categories). These are pre-existing and out of scope for this PR, but they represent the same class of issue this PR fixes — if a new parametric category is added in the future, it will hit the same panic. A follow-up to replace these with proper error returns (`.invalid`) would make the parser more robust.

MikaelMayer:

PR #1259, opened 2026-05-27
DDM Elab: fix failed import remaining in dialect.imports

### LINKED ISSUE #1243: DDM Elab: failed `import` is appended to `dialect.imports` before load attempt; remains on failure
[`elabDialectImportCommand`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DDM/Elab/DialectM.lean#L700-L724) pushes `name` into `imports` *before* attempting to load the dialect:

```lean
modifyDialect fun d => { d with imports := d.imports.push name }
let d ← match (← getLoadedDialects).dialects[name]? with
  | some d => pure d
  | none =>
    let r ← (← read).loadDialect name
    match r with
    | .ok d => pure d
    | .error msg =>
      logError identTree.loc msg
      modify fun s => { s with missingImport := true }
      return  -- ← name is already in imports!
```

When the load fails, the early `return` leaves `name` in the dialect's imports. Downstream `addDialect!` panics with "Unknown dialect" the next time this dialect is loaded.

**Fix**: move the `modifyDialect` push to after a successful load (right before `openLoadedDialect!`).

**Test plan**: elaborate a dialect source `dialect TestBugB; import NonExistent;`. Assert `d.imports.contains "NonExistent"` is `false` after the elaboration error.

MikaelMayer:
Fixes #1243

`elabDialectImportCommand` was pushing the import name into `dialect.imports` before attempting to load the dialect. When the load failed, the early `return` left the name in imports, causing downstream `addDialect!` panics with "Unknown dialect".

The fix extracts a `getOrLoadDialect` helper that encapsulates the lookup-or-load logic (including error reporting and `missingImport` flag). `elabDialectImportCommand` is now a linear sequence: resolve the dialect, push to imports, open it — making the invariant that `dialect.imports` only contains successfully resolved dialects locally evident.

Tested: existing tests pass, new test added that elaborates a dialect with a non-existent import, asserts the failed name is absent from `dialect.imports`, and exercises `openLoadedDialect!` on the result to confirm the previously-panicking path succeeds.


MikaelMayer (commented):
🤖🔍 Clean fix. Moving `modifyDialect` after the successful load path is the right approach — it ensures the invariant that `dialect.imports` only contains names of dialects that are actually loaded, which prevents the downstream `panic!` in `ensureLoaded!`. The test directly exercises the bug scenario.

MikaelMayer:
🤖 Done in 02df31e11 — converted the test to `#guard_msgs` style.

MikaelMayer:
🤖 Done in 02df31e11 — converted the test to `#guard_msgs` style.

Strata/DDM/Elab/DialectM.lean
```diff
@@ -719,6 +718,7 @@ def elabDialectImportCommand : DialectElab := fun tree => do
         logError identTree.loc msg
         modify fun s => { s with missingImport := true }
         return
+  modifyDialect fun d => { d with imports := d.imports.push name }
```
tautschnig, line 719: Fix is right; the underlying maintainability issue is that the failure path performs three coupled actions (`logError`, `missingImport := true`, `return`) and the success path performs two (push to `imports`, `openLoadedDialect!`), all interleaved via a `match` whose branches return early. Easy to break again. Worth a follow-up to extract:

```lean
def getOrLoadDialect (name : DialectName) (loc : SourceRange) : DialectM (Option Dialect) := do
  if let some d := (← getLoadedDialects).dialects[name]? then return some d
  match ← (← read).loadDialect name with
  | .ok d => return some d
  | .error msg =>
    logError loc msg
    modify fun s => { s with missingImport := true }
    return none
```
Then `elabDialectImportCommand` becomes a linear sequence: `let some d := … | return; modifyDialect …; openLoadedDialect!`, and the invariant "`dialect.imports` only contains names of successfully resolved dialects" is locally evident.

StrataTest/DDM/Import.lean
```diff
@@ -22,3 +24,23 @@ def testPgm :=
...
+    fileName := "<test>"
+    fileMap := Lean.FileMap.ofString src
+  }
+  let loaded := Strata.Elab.LoadedDialects.builtin
+  let fm ← (Strata.DialectFileMap.new loaded).toIO
+  let (d, _) ← (Strata.Elab.elabDialect fm inputCtx).toIO
+  if d.imports.contains "NonExistent" then
+    IO.println "BUG: failed import 'NonExistent' is still in dialect.imports"
+  else
+    IO.println "failed import 'NonExistent' was correctly excluded from dialect.imports"
```
tautschnig: Two improvements:

1. **Use `#guard` instead of `#guard_msgs`** — the test logic is a Bool predicate, no need to pin two branches of `IO.println`:
   ```lean
   #guard
     match (run the elaboration ...) with
     | .ok (d, _) => !d.imports.contains "NonExistent"
     | .error _   => false
   ```
   Cleaner, less brittle to message wording.
2. **Test the actual downstream consequence**. Issue #1243 says the bug caused `addDialect!` to panic with `"Unknown dialect"` later. The current test only checks the `imports` field; it doesn't exercise the panic path. Add a follow-up `#guard` (or similar) that takes the elaborated dialect, calls `LoadedDialects.addDialect!` (or any other code path that previously panicked), and confirms it succeeds.

For a stronger guarantee, the natural theorem is the invariant the bug violated:
```lean
theorem elabDialect_imports_loaded :
    ∀ fm ctx (d, s) ← elabDialect fm ctx,
    ∀ n ∈ d.imports, n ∈ (← fm.getLoaded).dialects
```
Any future refactor of `elabDialectImportCommand` would have to preserve this.

Strata/DDM/Elab/DialectM.lean
MikaelMayer: 🤖 Extracted `getOrLoadDialect` as suggested — `elabDialectImportCommand` is now a linear sequence: `let some d ← getOrLoadDialect … | return; modifyDialect …; openLoadedDialect!`.

47ecb12f4

StrataTest/DDM/Import.lean
MikaelMayer: 🤖 Replaced `#guard_msgs`/`IO.println` with `assert!` and added a downstream test that exercises `openLoadedDialect!` on the elaborated dialect — this is the path that previously panicked with "Unknown dialect".

47ecb12f4

tautschnig (approved):

joehendrix (approved):

PR #1260, opened 2026-05-27
fix: correct "Source rang" typo in SourceRange.fromIon error message

### LINKED ISSUE #1242: Ion deserializer: typo in `SourceRange.fromIon` error message (`"Source rang"`)
[`Ion.lean#L870`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DDM/Ion.lean#L870):

```lean
let ⟨exp, _⟩ ← .asSexp "Source rang" v
```

Missing trailing `e`. Affects the error message only.

**Fix**:

```diff
-      let ⟨exp, _⟩ ← .asSexp "Source rang" v
+      let ⟨exp, _⟩ ← .asSexp "Source range" v
```

**Test plan**: `#guard` on a malformed `SourceRange` Ion (a string instead of a sexp); assert the error starts with `Source range`. Verify it fails before the fix.

MikaelMayer:
Fixes #1242

Fix typo in `SourceRange.fromIon`: the `asSexp` call used `"Source rang"` instead of `"Source range"`, producing a truncated error message on malformed input.

The repeated `"Source range"` prefix is now hoisted into a local `let tag` binding to prevent the same class of typo from recurring.

Tests added:
- Error-path test asserting the error message starts with `"Source range"`
- Success-path round-trip tests for both the null case and a valid sexp with start/stop values


MikaelMayer (commented):
🤖🔍 Clean fix. The typo is confirmed on main (the `asSexp` call used `"Source rang"` which produced `"Error decoding Source rang expected sexpression."`), and the test correctly validates the fixed message prefix. LGTM.

joscoh (dismissed):

Strata/DDM/Ion.lean
```diff
@@ -867,7 +867,7 @@ private instance : FromIon SourceRange where
     | .null _ =>
       return .none
     | _ =>
-      let ⟨exp, _⟩ ← .asSexp "Source rang" v
+      let ⟨exp, _⟩ ← .asSexp "Source range" v
```
tautschnig: The literal `"Source range"` is now repeated four times within this instance (lines 870, 871, 873, 874 — though 873/874 add suffixes). To prevent the same typo class from recurring, hoist the prefix:

```lean
private instance : FromIon SourceRange where
  fromIon v := do
    let name := "Source range"
    match v.app with
    | .null _ => return .none
    | _ =>
      let ⟨exp, _⟩ ← .asSexp name v
      let ⟨p⟩ ← .checkArgCount name exp 2
      return {
        start := ⟨← .asNat s!"{name} start" exp[0]⟩
        stop  := ⟨← .asNat s!"{name} stop" exp[1]⟩
      }
```

StrataTest/DDM/Ion.lean
```diff
@@ -266,3 +266,10 @@ datatype Box(a: Type) { MkBox(val: a) };
 
 #guard testDialectRoundTrip TestIonDatatypes
 #guard testProgramRoundTrip testIonDatatypesPgm
+
+-- Test that SourceRange.fromIon error message says "Source range" (not "Source rang")
+#guard
+  let bs := Ion.serialize #[Ion.SymbolTable.system.localSymbolTableValue, Ion.string (Sym := Ion.SymbolId) "bad"]
+  match FromIon.deserialize (α := SourceRange) bs with
+  | .error msg => "Error decoding Source range".isPrefixOf msg
+  | .ok _ => false
```
tautschnig, line 290: Current coverage for `SourceRange` Ion is only the error path of `fromIon`. Please add success-path round-trip `#guard`s — these guard against any future regression in the (still-unverified) `toIon`/`fromIon` symmetry, including the `null ↔ {start:=0,stop:=0}` corner case:

```lean
-- Round-trip: zero range serializes as null and round-trips
#guard
  let r : SourceRange := { start := ⟨0⟩, stop := ⟨0⟩ }
  match FromIon.deserialize (α := SourceRange) (Ion.serialize #[Ion.SymbolTable.system.localSymbolTableValue, ToIon.toIon r]) with
  | .ok r' => r = r'
  | .error _ => false

-- Round-trip: non-zero range
#guard
  let r : SourceRange := { start := ⟨3⟩, stop := ⟨7⟩ }
  ...
```

If you want stronger coverage, a Plausible property `∀ r : SourceRange, deserialize (serialize r) = .ok r` would be ideal — and is the natural theorem statement to aim for if/when these `FromIon`/`ToIon` instances are eventually proved correct. Adjust the helper signatures if `Ion.serialize` needs different framing.

Strata/DDM/Ion.lean
MikaelMayer: 🤖 Hoisted the prefix into a `let tag := "Source range"` binding and used string interpolation for the suffixed variants. See 810217eaa.

StrataTest/DDM/Ion.lean
MikaelMayer: 🤖 Added two success-path `#guard` round-trip tests: one for the null case (decodes to 0,0) and one for a sexp with start/stop values. See 810217eaa.

tautschnig (approved):

joscoh (approved):

PR #1262, opened 2026-05-27
Report arity mismatch error for type aliases

### LINKED ISSUE #1239: Lambda type aliases: `aliasDef?` / `tconsAlias` silently return original type on arity mismatch

```lean
match matching with
| some alias =>
  else
    -- existing expansion
```

MikaelMayer:
Fixes #1239

`tconsAliasSimple` and `tconsAlias` previously matched aliases by both name and arity in a single `find?` predicate. When the name matched but arity didn't, `find?` returned `none` and the functions silently returned the original unresolved type, leading to confusing downstream type errors.

This PR splits the name-match from the arity-check: `find?` now matches by name only, and a separate arity check produces a clear error message ("Arity mismatch for alias 'X': expected N type argument(s), got M"). `tconsAliasSimple` now returns `Except Format LMonoTy` to propagate this error.

All existing proofs updated to handle the new control flow. Existing tests pass, new test added for arity mismatch detection.


StrataBoole/StrataBoole/Verify.lean
```diff
@@ -371,8 +436,56 @@ def toCoreExpr (e : Boole.Expr) : TranslateM Core.Expression.Expr := do
...
+  | .seq_empty_bv8 _  => return Core.seqEmptyOp (some (.bitvec 8))
+  | .seq_empty_bv16 _ => return Core.seqEmptyOp (some (.bitvec 16))
+  | .seq_empty_bv32 _ => return Core.seqEmptyOp (some (.bitvec 32))
+  | .seq_empty_bv64 _ => return Core.seqEmptyOp (some (.bitvec 64))
+  | .seq_empty_int _  => return Core.seqEmptyOp (some .int)
+  -- Sequence literals: Sequence.of_bv32[v0, v1, ..., vn]
+  -- Lowers to a left-fold of seq_build over seq_empty.
+  | .seq_of_bv8  _ ⟨_, vs⟩ | .seq_of_bv16 _ ⟨_, vs⟩ | .seq_of_bv32 _ ⟨_, vs⟩
+  | .seq_of_bv64 _ ⟨_, vs⟩ | .seq_of_int  _ ⟨_, vs⟩ => do
+    let vals ← vs.toList.mapM toCoreExpr
```
MikaelMayer: 🤖🔍 `seq_of_*` uses `Core.seqEmptyOp` (no type argument → polymorphic `Sequence.empty`), while the `seq_empty_*` variants pass an explicit element type. This means the initial empty sequence in a literal like `Sequence.of_bv32[1, 2]` is untyped, relying on type inference from the surrounding `seqBuildOp` calls. This works today but is fragile — if the literal is empty (`Sequence.of_bv32[]`), the result is an untyped `Sequence.empty` rather than the typed `Sequence.empty_bv32`. Consider passing the element type here too (you already know it from the match arm).

Strata/Transform/TerminationCheck.lean
```diff
@@ -46,13 +50,46 @@ def termProcName (name : String) : String := s!"{name}{termSuffix}"
...
+private def getDecreasesKind (func : Function) (tf : @TypeFactory Unit)
+    : Except String DecreasesKind := do
   match func.measure with
-  | some (.fvar _ id _) => func.inputs.keys.findIdx? (· == id)
-  | some _ => none
-  | none => FuncAttr.findInlineIfConstr func.attr
+  | some (.fvar _ id _) =>
+    match func.inputs.findWithIdx? id with
+    | some (_, .int) => .ok (.intValued (.fvar () id (.some .int)))
+    | some (idx, _) => validateStructuralParam func tf idx
```
MikaelMayer, line 88: 🤖🔍 When `func.inputs.findWithIdx?` returns `some (_, .int)`, the index is discarded and the measure is constructed as `.fvar () id (.some .int)`. But `findWithIdx?` returns the index for a reason — shouldn't `validateStructuralParam` be the only path that uses the index? This is fine logically (int-valued measures don't need the index), but the asymmetry with the next arm `some (idx, _)` is worth a brief comment explaining why the index is intentionally discarded for int-typed parameters.

Strata/Transform/TerminationCheck.lean
```diff
@@ -46,13 +50,46 @@ def termProcName (name : String) : String := s!"{name}{termSuffix}"
...
+/-- Classifies the termination measure for a recursive function. -/
+inductive DecreasesKind where
+  | structural (paramIdx : Nat) (dtName : String)
+  | intValued (measure : Expression.Expr)
+
+/-- Validate that a parameter index refers to a known ADT type, returning `.structural`. -/
+private def validateStructuralParam (func : Function) (tf : @TypeFactory Unit)
+    (idx : Nat)
+    : Except String DecreasesKind := do
+  if FuncAttr.findInlineIfConstr func.attr |>.isNone then
```
MikaelMayer, line 66: 🤖🔍 Nit: `validateStructuralParam` checks `FuncAttr.findInlineIfConstr func.attr |>.isNone` and errors if true. But the caller `getDecreasesKind` only reaches this path from two places: (1) `some (.fvar _ id _)` with a non-int type, and (2) `none` with `FuncAttr.findInlineIfConstr` returning `some idx`. In case (2), the check is redundant (it was just confirmed present). In case (1), the check is meaningful. Consider whether this defensive check is worth the potential confusion of erroring about `@[cases]` when the user wrote a `decreases` clause pointing to an ADT parameter.

StrataBoole/StrataBoole/Verify.lean
```diff
@@ -503,13 +632,34 @@ def toCoreStmt (s : BooleDDM.Statement SourceRange) : TranslateM Core.Statement
...
+    let md ← toCoreMetaData m
+    let label ← defaultLabel m "choose" none
+    -- Existence obligation: assert ∃ z : T . pred(z) before havocing.
+    -- Without this, `havoc w; assume pred(w)` silently becomes
+    -- `assume false` when pred is unsatisfiable, making every subsequent
+    -- obligation verify as a false positive.
+    let .mono_bind_mk _ _ vTy := v
+    let vMonoTy ← toCoreMonoType vTy
+    let existsBody ← withBVarExprs #[.bvar () 0] (toCoreExpr pred)
+    let existsExpr : Core.Expression.Expr :=
```
MikaelMayer, line 658: 🤖🔍 The `existsBody` is constructed with `withBVarExprs #[.bvar () 0] (toCoreExpr pred)`, which pushes a de Bruijn index 0 as the bound variable expression for `pred`. But `pred` was already translated with `lhsExpr` (an `.fvar`) as the bvar binding (line 636). This means `pred` references `lhs` as an fvar, while `existsBody` references it as a bvar — the two translations of `pred` use different binding strategies. Is this intentional? The `existsExpr` quantifier binds a fresh variable, so the body should use `.bvar () 0` to refer to it, but `toCoreExpr pred` is called again here with `.bvar () 0` in scope, which means the pred's free-variable reference to `lhs` will be translated as `.bvar () 0` (the quantified variable). This seems correct but subtle — a comment would help.

Strata/Languages/Core/CmdType.lean
```diff
@@ -58,15 +58,8 @@ bound variables.
...
-  -- We only allow free variables to appear in `init` statements. Any other
-  -- occurrence leads to an error.
-  let T ← match c with
-    | .init _ _ _ _ =>
-      let efv := LExpr.freeVars e
-      (Env.addInOldestContext efv).mapError DiagnosticModel.fromFormat
-    | _ =>
-      let _ ← Env.freeVarCheck e f!"[{c}]" |>.mapError DiagnosticModel.fromFormat
-      .ok Env
+  let _ ← Env.freeVarCheck e f!"[{c}]" |>.mapError DiagnosticModel.fromFormat
```
MikaelMayer, line 61: 🤖🔍 The `let T := Env` binding is now a no-op identity assignment. Consider just using `Env` directly in the subsequent code (or renaming the parameter) to avoid the dead binding.

MikaelMayer (commented):
🤖🔍 This PR delivers what it promises: splitting the name-match from the arity-check in `tconsAlias`/`tconsAliasSimple` so that arity mismatches produce a clear error instead of silently returning the unresolved type. The proof updates are mechanical and follow the new control flow correctly.

The PR also bundles a significant amount of additional work (int-valued termination measures, `inlineIfAllCanonical`, Boole grammar extensions for sequences/lambdas/choose/decreases, `getFVarIsOp` refactoring, new test seeds). This makes the PR quite large and harder to review in isolation, but the individual pieces are well-structured.

Positive notes:
- The `DecreasesKind` ADT in `TerminationCheck.lean` is a clean design that separates structural from int-valued termination and makes the code self-documenting.
- The `getFVarIsOp_spec` theorem is a nice addition that formally ties the implementation to its specification.
- The rejection of mixed structural/int-valued mutual blocks is a good defensive check.

StrataBoole/StrataBoole/Verify.lean
MikaelMayer: 🤖 Fixed — split the combined `seq_of_*` arms into individual arms and introduced a `seqOfToCore` helper that passes the explicit element type to `Core.seqEmptyOp`. Empty literals now produce typed `Sequence.empty` expressions. [298bed7](https://github.com/strata-org/Strata/commit/298bed778361e9ac59ce8808c24f11c9004aa332)

Strata/Transform/TerminationCheck.lean
MikaelMayer: 🤖 Added a comment explaining that int-valued measures use the expression directly and the parameter index is not needed. [298bed7](https://github.com/strata-org/Strata/commit/298bed778361e9ac59ce8808c24f11c9004aa332)

Strata/Transform/TerminationCheck.lean
MikaelMayer: 🤖 Expanded the docstring for `validateStructuralParam` to explain when the `@[cases]` check is redundant vs meaningful. [298bed7](https://github.com/strata-org/Strata/commit/298bed778361e9ac59ce8808c24f11c9004aa332)

StrataBoole/StrataBoole/Verify.lean
MikaelMayer: 🤖 Added a comment explaining that `existsBody` re-translates `pred` with a de Bruijn bvar so the quantifier body refers to the freshly quantified variable, not the fvar used for the assume statement. [298bed7](https://github.com/strata-org/Strata/commit/298bed778361e9ac59ce8808c24f11c9004aa332)

Strata/Languages/Core/CmdType.lean
MikaelMayer: 🤖 Removed the dead `let T := Env` binding and used `Env` directly in the `LExpr.resolve` call. [298bed7](https://github.com/strata-org/Strata/commit/298bed778361e9ac59ce8808c24f11c9004aa332)

MikaelMayer:

PR #1263, opened 2026-05-27
Ion deserializer: `SyntaxDefAtom.fromIon "indent"` crashes on malformed input

### LINKED ISSUE #1238: Ion deserializer: `SyntaxDefAtom.fromIon "indent"` crashes on malformed input
[`Ion.lean#L810-L814`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DDM/Ion.lean#L810-L814):

```lean
| "indent" => do
  .indent <$> .asNat "SyntaxDef indent value" args[1]!
```

No `checkArgCount` / `checkArgMin` precedes the `args[1]!` access. A malformed `(indent)` sexp (only the tag) crashes with `lean_array_get_panic` (interpreter) or `abort()` (compiled). The neighbouring `"ident"` case uses `checkArgCount` correctly.

**Fix**:

```diff
     | "indent" => do
+      let ⟨p⟩ ← .checkArgMin "indent" args 2
       .indent <$> .asNat "SyntaxDef indent value" args[1]
               <*> args.attach.mapM_off (start := 2) ...
```

(Drop the `!` after the bounds check — index becomes provably in-range.)

**Test plan**: deserialize a dialect Ion containing `(indent)` (no second arg). Verify it currently crashes the process and after the fix returns a graceful error.

MikaelMayer:
Fixes #1238

Adds a `checkArgMin` bounds check before accessing `args[1]` in the `"indent"` case of `SyntaxDefAtom.fromIon`. Previously, a malformed sexp like `(indent)` with no second argument would crash with a panic. Now it returns a graceful error message.

Also cleans up the neighbouring `"ident"` case by dropping panicking `!` indexing (the proof from `checkArgCount` already provides bounds), renames unused proof variables to `_`, and renames a shadowed `have p` to `have _h` for clarity.

Tested: existing tests pass, `#guard_msgs` regression tests added covering the malformed indent path, successful empty tail, successful with inner atoms, and wrong-type-at-args[1] error path.


MikaelMayer:
🤖⛏️ Added `checkArgMin "indent" args 2` before the `args[1]` access in the `"indent"` case of `SyntaxDefAtom.fromIon`. This follows the same pattern used by the neighbouring `"ident"` case and other deserializers in the file. The `!` is dropped since the proof `p : args.size ≥ 2` makes the index provably in-range.

Note: the linker error in CI (if it appears) is a pre-existing environment issue with `libc++abi` / `__tls_get_addr` — all Lean modules compile and type-check successfully.

MikaelMayer (commented):
🤖🔍 Clean, minimal fix. The `checkArgMin` guard replaces a potential `panic!` (from `args[1]!` on malformed input) with a graceful `throw`, and the proof witness `p : args.size ≥ 2` lets Lean verify the subsequent `args[1]` access is in bounds — no `!` needed. Consistent with how other cases in this file use `checkArgMin`.

MikaelMayer:
The fix is correct, but there's no regression test. Issue #1238 and the project's test conventions (per `AGENTS.md`) ask for a failing test added first, verified failing, then the fix applied. Could you add a `#guard_msgs` test that exercises the malformed `(indent)` path?

Suggested test (drop into `StrataTest/DDM/Ion.lean` or a similar Ion-deserializer test file):

```lean
-- Issue #1238: `SyntaxDefAtom.fromIon` should report a clean error on a
-- malformed `(indent)` sexp (only the tag, no value), not crash.
private def malformedIndentDialect : Ion String :=
  .list #[
    .sexp #[.symbol "dialect", .string "TestIndentBounds"],
    .struct #[
      ("type",   .symbol "op"),
      ("name",   .string "testOp"),
      ("result", .symbol "TestIndentBounds.Command"),
      ("syntax", .struct #[
        ("atoms", .list #[.sexp #[.symbol "indent"]]),  -- 1 element: was a crash
        ("prec", .int 0)
      ])
    ]
  ]

/-- info: error: indent expects at least 2 arguments has #[(symbol indent)] -/
#guard_msgs in
#eval do
  let bs := Ion.internAndSerialize [malformedIndentDialect]
  match FromIon.deserialize (α := Dialect) bs with
  | .error msg => IO.println s!"error: {msg}"
  | .ok _      => IO.println "unexpected ok"
```

Adjust the expected error string to whatever `checkArgMin "indent" args 2` actually emits in the framework's wrapping (probably with an `Error decoding` prefix). Verify the test fails on `main2` *before* this PR's fix is applied (would currently crash with `lean_array_get_panic`), then passes after.

MikaelMayer:
🤖 Added a `#guard_msgs` regression test that exercises the malformed `(indent)` path (sexp with only the tag, no value). The test verifies that deserialization returns a clean error message instead of crashing.

Commit: 4bc6a7740

atomb (dismissed):

Strata/DDM/Ion.lean
```diff
@@ -808,7 +808,8 @@ private protected def fromIon (v : Ion SymbolId) : FromIonM SyntaxDefAtom := do
       let prec ← .asNat "SyntaxDef ident prec" args[2]!
```
tautschnig: Drive-by cleanup opportunity: the neighbouring `"ident"` case (lines 807–808) already has `p : args.size = 3` in scope (from `checkArgCount` at 806), so `args[1]!` and `args[2]!` here can both drop the `!` — same idiom you applied below for the `"indent"` case. Same pattern also at line 381 (`entries[1]!` after `entries.size = 2`) and line 1626 (`ctx[1]!` after `ctx.size = 2`). A small follow-up to remove redundant `!`s would make the safety properties first-class everywhere.

Strata/DDM/Ion.lean
```diff
@@ -808,7 +808,8 @@ private protected def fromIon (v : Ion SymbolId) : FromIonM SyntaxDefAtom := do
...
       return .ident level prec
     | "indent" => do
-      .indent <$> .asNat "SyntaxDef indent value" args[1]!
+      let ⟨p⟩ ← .checkArgMin "indent" args 2
+      .indent <$> .asNat "SyntaxDef indent value" args[1]
               <*> args.attach.mapM_off (start := 2) fun ⟨u, _⟩ =>
                     have p : sizeOf u < sizeOf args := by decreasing_tactic
```
tautschnig: Minor: this `have p` shadows the outer `p : args.size ≥ 2` introduced on line 811. They serve different purposes (outer for the bounds proof on `args[1]`, inner for the recursive call's termination measure). Renaming the inner one to e.g. `hu` removes the confusion at no cost.

StrataTest/DDM/Ion.lean
```diff
@@ -266,3 +266,27 @@ datatype Box(a: Type) { MkBox(val: a) };
...
+    ]
+  ]
+
+/-- info: error: Error decoding indent expects at least 2 arguments -/
+#guard_msgs in
+#eval do
+  let bs := Ion.internAndSerialize [malformedIndentDialect]
+  match FromIon.deserialize (α := Dialect) bs with
+  | .error msg => IO.println s!"error: {msg}"
+  | .ok _      => IO.println "unexpected ok"
```
tautschnig, line 321: Regression test is good. Three cheap additions would tighten coverage of this fromIon branch:

1. **Successful empty tail**: `(indent 0)` — should round-trip to `.indent 0 #[]` (no atoms after the count).
2. **Successful with inner atoms**: `(indent 2 "x")` — exercises the variadic `mapM_off (start := 2)` branch.
3. **Wrong type at args[1]**: `(indent "notnum")` — should produce the `asNat` error, not a panic.

For stronger guarantees, consider a Plausible round-trip property:
```lean
example (a : SyntaxDefAtom) :
    FromIon.deserialize (Ion.serialize <| ToIon.toIon a) = .ok a := by plausible
```
This is the natural theorem to formalize and is the same pattern that would benefit other `private instance : FromIon …` instances in `Ion.lean`.

tautschnig (dismissed):

Strata/DDM/Ion.lean
MikaelMayer: 🤖 Done — dropped the `!` from `args[1]!` and `args[2]!` since `p : args.size = 3` is in scope. Commit: 79c6c4e49

Strata/DDM/Ion.lean
MikaelMayer: 🤖 Good catch — renamed the inner `have p` to `have _h` to avoid shadowing the outer `p : args.size ≥ 2`. Commit: 79c6c4e49

StrataTest/DDM/Ion.lean
MikaelMayer: 🤖 Added all three test cases: empty tail `(indent 0)`, inner atoms `(indent 2 "x")`, and wrong type `(indent "notnum")`. Refactored the boilerplate into a shared `indentTestDialect` helper.

Commit: d641119c3

Regarding the `Plausible` round-trip property — good idea for a follow-up but out of scope for this fix.

tautschnig (dismissed):

Strata/DDM/Ion.lean
```diff
@@ -804,14 +804,14 @@ private protected def fromIon (v : Ion SymbolId) : FromIonM SyntaxDefAtom := do
...
-      let prec ← .asNat "SyntaxDef ident prec" args[2]!
+      let level ← .asNat "SyntaxDef ident level" args[1]
+      let prec ← .asNat "SyntaxDef ident prec" args[2]
       return .ident level prec
     | "indent" => do
       let ⟨p⟩ ← .checkArgMin "indent" args 2
       .indent <$> .asNat "SyntaxDef indent value" args[1]
               <*> args.attach.mapM_off (start := 2) fun ⟨u, _⟩ =>
-                    have p : sizeOf u < sizeOf args := by decreasing_tactic
+                    have _h : sizeOf u < sizeOf args := by decreasing_tactic
```
atomb, line 814: Do you need the line if the variable it introduces is unused? If it's somehow used as an implicit argument later, it seems like that would be a bug in Lean's unused variable warnings.

atomb (dismissed):
Suggested one possible but non-blocking improvement, but otherwise looks good.

Strata/DDM/Ion.lean
MikaelMayer: 🤖 Good catch. The line is still needed for validation (produces a clear error message if args has fewer than 2 elements), but the proof variable `p` was indeed unused by name — `omega` finds it implicitly for the array bounds check. Renamed to `_` to suppress the unused variable warning.

tautschnig (dismissed):

tautschnig (approved):

Strata/DDM/Ion.lean
atomb: Huh. That seems like a bug in Lean to me. It should warn only about variables missing from the final proof term. Anyway, this is good for now.

atomb (approved):

Strata/DDM/Ion.lean
MikaelMayer: 🤖 Acknowledged — no code change needed. The reviewer confirmed the current approach is acceptable. The validation line is kept for its runtime error-reporting value even though the bound variable is unused by name in the proof term.

PR #1264, opened 2026-05-27
Fix: Ion deserializer `decimal` arg-count check reports wrong constructor name

### LINKED ISSUE #1237: Ion deserializer: `decimal` arg-count check reports wrong constructor name (`"num"`)
[`Ion.lean#L726-L731`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DDM/Ion.lean#L726-L731):

```lean
| "decimal" =>
  let ⟨p⟩ ← .checkArgCount "num" sexp 3
```

Copy-paste from the `"num"` case immediately above. A malformed `"decimal"` sexp produces an error message that says `num expects 3 arguments`.

**Fix**:

```diff
   | "decimal" =>
-    let ⟨p⟩ ← .checkArgCount "num" sexp 3
+    let ⟨p⟩ ← .checkArgCount "decimal" sexp 3
```

**Test plan**: add a `#guard` test that calls `ArgF.fromIon` on a 4-element `decimal` sexp and asserts the error message starts with `decimal`. Verify it fails before the fix.

MikaelMayer:
Fixes #1237

The `checkArgCount` call in the `"decimal"` case of `ArgF.fromIon` was copy-pasted from the `"num"` case and still passed `"num"` as the constructor name. A malformed `decimal` sexp would produce an error saying "num expects 3 arguments" instead of "decimal expects 3 arguments".

Fixed by passing `"decimal"` to `checkArgCount`. Added a test in `StrataTest/DDM/Ion.lean` that constructs a 4-element `decimal` sexp and verifies the error message starts with `"decimal"`.

Filed #1294 to track a broader refactoring that would eliminate this class of copy-paste bug.


MikaelMayer (commented):
🤖🔍 Clean, minimal fix for a copy-paste bug. The change is correct and the `#guard` test directly verifies the fix. Build passes.

Strata/DDM/Ion.lean
```diff
@@ -1663,3 +1663,19 @@ def filesFromIon (dialects : DialectMap) (bytes : ByteArray) : Except String (Li
     pure { filePath := filePath, program := program }
 
 end Strata.Program
+
+namespace Strata
+open _root_.Ion (SymbolTable Ion SymbolId)
+
+-- Verify that a malformed `decimal` sexp reports "decimal" in the error message.
+#guard
```
tautschnig: Test is in the source file but should live in `StrataTest/DDM/Ion.lean` per the repo convention (see `.kiro/steering/structure.md`: "Mirror test structure in `StrataTest/` to match `Strata/`"). That file already has `import all Strata.DDM.Ion`, which exposes private definitions including `ArgF.fromIon`, so the move is mechanical.

While moving, also tighten the assertion — `msg.startsWith "decimal"` would also pass on a hypothetical future `"decimalFoo expects …"` regression. Prefer exact equality or a substring check anchored on the full prefix:
```lean
msg = "decimal expects 3 arguments has #[(symbol $1), (null), (null), (null)]"
```
(or use `String.isInfixOf "decimal expects 3"` if the suffix is volatile).

Strata/DDM/Ion.lean
```diff
@@ -724,7 +724,7 @@ private protected def ArgF.fromIon {α} [FromIon α] (v : Ion SymbolId) : FromIo
       | throw s!"Arg num given {repr sexp}."
     pure <| .num ann x
   | "decimal" =>
-    let ⟨p⟩ ← .checkArgCount "num" sexp 3
+    let ⟨p⟩ ← .checkArgCount "decimal" sexp 3
```
tautschnig, line 727: This is the third bug in this file in this class (this PR; PR #1260 fixed `"Source rang"` typo; PR #1263 fixed missing bounds check). The shared root cause is that the case-arm string and the `checkArgCount/checkArgMin` name argument are duplicated literals with no static link.

Follow-up worth filing: introduce a tiny helper that takes the case string once, e.g.
```lean
let ⟨p⟩ ← .matchSexp sexp "decimal" 3 -- emits both the case string and the arity check from one source
```
or a macro `case_arm "decimal" (count := 3)` that desugars to the match arm + bounds check together. Either approach makes the duplication impossible. Out of scope for this PR but please file an issue.

MikaelMayer: Let's file an issue about that

Strata/DDM/Ion.lean
MikaelMayer: 🤖 Moved the test to `StrataTest/DDM/Ion.lean` per repo convention.

Strata/DDM/Ion.lean
MikaelMayer: 🤖 Filed #1294 to track the refactoring of `ArgF.fromIon` to eliminate this class of copy-paste bugs.

tautschnig (dismissed):

joehendrix (dismissed):

tautschnig (approved):

joehendrix (approved):

PR #1265, opened 2026-05-27
ProcedureInlining: remove dead `var_map.find?` lookup in `outputSetStmts`

### LINKED ISSUE #1234: ProcedureInlining: remove dead `var_map.find?` lookup in `outputSetStmts`
[`ProcedureInlining.lean#L261-L270`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/Transform/ProcedureInlining.lean#L261-L270):

```lean
let out_vars := sigOutputs.unzip.fst
let out_vars := out_vars.map
    (fun id => match var_map.find? id with
        | .none => id | .some x => x)
```

`out_vars` already contains renamed identifiers (since `proc.header.outputs` was updated by `renameAllLocalNames`), and `var_map`'s keys are the original (pre-rename) identifiers. So `var_map.find? id` always returns `.none` — the second `let out_vars` is a no-op. Result is correct by accident. Pure cleanup; no behaviour change.

**Fix**: drop the second `let out_vars := …` binding.

**Test plan**: existing inlining tests must continue to pass. No new test needed.

MikaelMayer:
Remove the dead `var_map.find?` lookup in `outputSetStmts`. The lookup always returned `.none` because `out_vars` already contains renamed identifiers (from `renameAllLocalNames`), while `var_map`'s keys are the original pre-rename identifiers. Also simplifies `renameAllLocalNames` to return `CoreTransformM Procedure` instead of a tuple, since `var_map` has no remaining consumers.

No behaviour change. Existing tests pass.

Fixes #1234


MikaelMayer (commented):
🤖🔍 Clean removal of dead code. The reasoning is correct: `renameAllLocalNames` returns `proc` with its header outputs already renamed via `var_map`, so `sigOutputs.unzip.fst` yields renamed identifiers. Looking those up in `var_map` (keyed by *original* identifiers) always returns `.none`, making the entire `.map` a no-op identity function.

No comments — the change is minimal and well-motivated.

Strata/Transform/ProcedureInlining.lean
```diff
@@ -227,7 +227,7 @@ def inlineCallCmd
 
         -- Create a copy of the procedure that has all input/output/local vars
         -- replaced with fresh ones
-        let (proc,var_map) <- renameAllLocalNames proc
+        let (proc, _var_map) <- renameAllLocalNames proc
```
joscoh: Nit: this could be an underscore if it's not used

joscoh (dismissed):

Strata/Transform/ProcedureInlining.lean
MikaelMayer: 🤖 Done, replaced `_var_map` with `_` in f010d8002.

joscoh (dismissed):

Strata/Transform/ProcedureInlining.lean
```diff
@@ -227,7 +227,7 @@ def inlineCallCmd
 
         -- Create a copy of the procedure that has all input/output/local vars
         -- replaced with fresh ones
-        let (proc,var_map) <- renameAllLocalNames proc
+        let (proc, _) <- renameAllLocalNames proc
```
tautschnig: Now that the only caller (this site) discards `var_map`, `renameAllLocalNames` has zero consumers of its second return component. Simplifying its signature to `: CoreTransformM Procedure` and updating the `return` at line 131 (drop the trailing `, var_map`) is a one-line follow-up that finishes the cleanup.

tautschnig (dismissed):

Strata/Transform/ProcedureInlining.lean
MikaelMayer: 🤖 Done — simplified `renameAllLocalNames` signature to `: CoreTransformM Procedure` and updated the call site. ea85f69ab

tautschnig (approved):

joscoh (approved):

PR #1266, opened 2026-05-27
SMT counter-example decoder: handle sc_decimal_neg

### LINKED ISSUE #1236: SMT counter-example decoder: missing case for `sc_decimal_neg`
[`translateFromDDMTermToUntyped`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DL/SMT/DDMTransform/Translate.lean#L324-L334) handles `sc_numeral`, `sc_numeral_neg`, `sc_decimal`, and `sc_str` but not `sc_decimal_neg`. Negative real-valued counter-examples (e.g. `(define-fun x () Real -2.0)`) fall through to the catch-all error and the variable is silently dropped from the parsed model.

**Fix**:

```diff
     | .sc_decimal _ d     => return .prim (.real d)
+    | .sc_decimal_neg _ d => return .prim (.real { d with mantissa := -d.mantissa })
     | .sc_str _ s         => return .prim (.string s)
```

**Test plan**: feed `sc_decimal_neg { mantissa := 2, exponent := 0 }` (representing `-2.0`) into `translateFromDDMTermToUntyped`. Verify it currently errors and after the fix returns `.prim (.real { mantissa := -2, exponent := 0 })`.

MikaelMayer:
Fixes #1236

`translateFromDDMTermToUntyped` was missing a case for `sc_decimal_neg`, causing negative real-valued counter-examples (e.g. `-2.0`) to fall through to the catch-all error and be silently dropped from the parsed model.

Added the missing case that negates the mantissa, mirroring the existing `sc_numeral_neg` pattern. Since all `SpecConstant` constructors are now exhaustively matched, the catch-all was removed and a comment documents the deliberate exhaustiveness so future reviewers don't re-add a defensive catch-all.

Tested: all library and test modules compile successfully.

### DIFF
--- a/Strata/DL/SMT/DDMTransform/Translate.lean
+++ b/Strata/DL/SMT/DDMTransform/Translate.lean
@@ -329,8 +329,8 @@ partial def translateFromDDMTermToUntyped (t : Strata.SMTResponseDDM.Term Strata
     | .sc_numeral _ n     => return .prim (.int n)
     | .sc_numeral_neg _ n => return .prim (.int (-(n : Int)))
     | .sc_decimal _ d     => return .prim (.real d)
+    | .sc_decimal_neg _ d => return .prim (.real { d with mantissa := -d.mantissa })
     | .sc_str _ s         => return .prim (.string s)
-    | _  => throw s!"translateFromDDMTermToUntyped: don't know how to convert {repr t}"
   | .qual_identifier _ qi =>
     match resolveQI qi with
     | some ("true", _)  => return .prim (.bool true)
--- a/Strata/DL/SMT/DDMTransform/Translate.lean
+++ b/Strata/DL/SMT/DDMTransform/Translate.lean
@@ -325,6 +325,7 @@ partial def translateFromDDMTermToUntyped (t : Strata.SMTResponseDDM.Term Strata
     : Except String Strata.SMT.Term := do
   match t with
   | .spec_constant_term _ sc =>
+    -- Exhaustive match over all SpecConstant variants; Lean will flag any missing case.
     match sc with
     | .sc_numeral _ n     => return .prim (.int n)
     | .sc_numeral_neg _ n => return .prim (.int (-(n : Int)))


MikaelMayer (commented):
🤖🔍 Clean, minimal fix. The `sc_decimal_neg` case correctly negates the mantissa (which the parser guarantees is positive), mirroring the `sc_numeral_neg` pattern. Removing the inner catch-all is safe since all 5 `SpecConstant` constructors are now exhaustively matched.

One minor observation: there's no regression test exercising a negative decimal counter-example end-to-end, but the fix is straightforward enough that existing compilation coverage is reasonable.

MikaelMayer:
The catch-all `| _ => throw …` removal is provably safe today — `SpecConstant` has exactly 5 variants per [`Strata/DL/SMT/DDMTransform/Parse.lean#L176-L190`](https://github.com/strata-org/Strata/blob/main2/Strata/DL/SMT/DDMTransform/Parse.lean#L176-L190) (`sc_numeral`, `sc_numeral_neg`, `sc_decimal`, `sc_decimal_neg`, `sc_str`), so the match is exhaustive and Lean's compiler will warn if a future variant is added.

Could you add a one-line comment above the `match sc with` to lock in that invariant for future maintainers?

```lean
| .spec_constant_term _ sc =>
  -- Exhaustive match over all 5 SpecConstant variants (see
  -- Strata/DL/SMT/DDMTransform/Parse.lean L176-L190). If a new variant is
  -- added there, Lean will flag the missing case here.
  match sc with
  | .sc_numeral _ n     => return .prim (.int n)
  ...
```

That makes the deliberate exhaustiveness explicit, so a future reviewer doesn't add back a defensive catch-all (which would silently swallow new variants instead of failing loudly).

MikaelMayer:
🤖 Added a concise comment documenting the deliberate exhaustiveness.

atomb (approved):

aqjune-aws (approved):

PR #1267, opened 2026-05-27
`Core` translator: unlabeled `cover` statements use `assert_def` counter and produce duplicate labels

### LINKED ISSUE #1235: `Core` translator: unlabeled `cover` statements use `assert_def` counter and produce duplicate labels
[`Translate.lean#L1369-L1372`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/Languages/Core/DDMTransform/Translate.lean#L1369-L1372):

```lean
let default_name := s!"cover_{bindings.gen.assert_def}"
let bindings := incrNum .cover_def bindings
```

The default label uses `assert_def` (the assert counter) instead of `cover_def`. The `cover_def` counter is incremented but never read. Two consecutive unlabeled `cover` statements get the same label `cover_0` when no asserts intervene.

**Fix**:

```diff
-    let default_name := s!"cover_{bindings.gen.assert_def}"
+    let default_name := s!"cover_{bindings.gen.cover_def}"
```

**Test plan**: add a Core program with two consecutive `cover (true);` statements. Verify the test fails before the fix (duplicate `cover_0`) and produces `cover_0`, `cover_1` after.

MikaelMayer:
Fixes #1235

Unlabeled `cover` statements were reading the `assert_def` counter instead of `cover_def` for their default label generation. This caused consecutive unlabeled `cover` statements to produce duplicate `cover_0` labels when no asserts intervened.

The fix uses the correct counter for each statement kind. The duplicated assert/cover translation pattern is extracted into a `translateLabeledCheck` helper.

Tested: existing tests pass, new regression tests verify distinct labels for consecutive covers and independent counters across assert/cover combinations.


MikaelMayer (commented):
🤖🔍 Clean, minimal fix for a clear copy-paste bug. The one-line change correctly reads `cover_def` instead of `assert_def`, and the new test directly demonstrates the fix by verifying distinct labels for consecutive unlabeled cover statements. No issues found.

atomb (dismissed):

Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -1369,7 +1369,7 @@ partial def translateStmt (p : Program) (bindings : TransBindings) (arg : Arg) :
     return ([.assert l c md], bindings)
   | q`Core.cover, #[rca, la, ca] =>
     let c ← translateExpr p bindings ca
-    let default_name := s!"cover_{bindings.gen.assert_def}"
+    let default_name := s!"cover_{bindings.gen.cover_def}"
```
tautschnig: Refactor opportunity: this same 5-line pattern is duplicated for `assert_def` (1361–1365), `cover_def` (1370–1374), `assume_def` (1379–1383), and `axiom_def` (1721–1723, 1731–1733). The pattern duplicates the `GenKind` constructor and the prefix string, which is exactly the inconsistency that produced this bug. Suggest a single source-of-truth helper:

```lean
def GenKind.prefix : GenKind → String
  | .var_def => "var" | .axiom_def => "axiom" | .assume_def => "assume"
  | .assert_def => "assert" | .cover_def => "cover"

def GenKind.counter (k : GenKind) (b : TransBindings) : Nat := ...

def freshDefaultLabel (k : GenKind) (b : TransBindings) : String × TransBindings :=
  (s!"{k.prefix}_{k.counter b}", incrNum k b)
```

Then each call site shrinks to `let (default_name, bindings) := freshDefaultLabel .cover_def bindings`, and the same typo cannot recur. Worth filing as a follow-up if not in this PR.

StrataTest/Languages/Core/Tests/CoverDefaultNames.lean
```diff
@@ -0,0 +1,53 @@
...
+Result: ✅ pass
+
+Obligation: cover_1
+Property: cover
+Result: ✅ pass
+-/
+#guard_msgs in
+#eval verify coverDefaultNames (options := .quiet)
+
+---------------------------------------------------------------------
```
tautschnig, line 53: Two extra cases would meaningfully tighten this regression test:

1. **`assert; cover;`** — the original bug also surfaces here (pre-fix, cover would have been labeled `cover_1` because it read `assert_def` after assert incremented it). This confirms the counters are independent.
2. **Mixed/long sequence**: e.g. `cover; assert; cover; assert;` — verifies the two counters advance independently and that ordering doesn't affect default-label assignment.

For stronger guarantees, the natural theorem to formalize is **default-label uniqueness**:
```lean
theorem translateStmts_distinct_default_labels :
    ∀ p stmts b, let (out, _) := translateStmts p b stmts
    out.collectDefaultLabels.Nodup := by ...
```
This would catch any future regression where two statements share a counter or a prefix without needing a hand-rolled #eval test for each combination.

Strata/Languages/Core/DDMTransform/Translate.lean
MikaelMayer: 🤖 Extracted `translateLabeledCheck` helper that deduplicates the assert/cover translation pattern. The assume case differs (no reachability check) so it stays inline. The axiom cases produce a different type (`Core.Decl`) so they also stay separate.

96557b339

StrataTest/Languages/Core/Tests/CoverDefaultNames.lean
MikaelMayer: 🤖 Added both test cases: `assertThenCover` (assert;cover;) and `coverThenAssert` (cover;assert;) verifying that counters are independent across statement kinds.

96557b339

tautschnig (approved):

atomb (approved):

PR #1270, opened 2026-05-27
SMT counter-example decoder: bitvec literals decoded as UF applications (indices discarded)

### LINKED ISSUE #1232: SMT counter-example decoder: bitvec literals decoded as UF applications (indices discarded)
[`resolveQI` at Translate.lean#L350-L364](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DL/SMT/DDMTransform/Translate.lean#L350-L364) collapses `iden_indexed _ sym _` to `(symbolToString sym, none)`, discarding the width index. A solver-returned `(_ bv5 32)` becomes `mkUFApp "bv5" []` rather than `.prim (.bitvec (BitVec.ofNat 32 5))`. Every bitvector variable in a counter-example is therefore reported with a meaningless UF name instead of its actual value.

**Fix**: extend `resolveQI` to return the indices, then in `translateFromDDMTermToUntyped` detect the `bv{N}` pattern with a numeral index and reconstruct the literal:

```lean
-- in resolveQI
| .iden_indexed _ sym idxs => some (symbolToString sym, none, idxs.val)

-- in translateFromDDMTermToUntyped after resolveQI returns (name, _, idxs):
if name.startsWith "bv" then
  if let some value := (name.drop 2).toNat? then
    if let #[.ind_numeral _ width] := idxs then
      return .prim (.bitvec (BitVec.ofNat width value))
```

**Test plan**: add a counter-example round-trip test for a bitvec-typed variable. Verify the decoded value is `BitVec.ofNat 32 5` (not a UF named `bv5`). The test should fail before the fix.

MikaelMayer:
Fixes #1232

`resolveQI` was discarding the indices from `iden_indexed` identifiers, so solver-returned bitvec literals like `(_ bv5 32)` were decoded as UF applications (`mkUFApp "bv5" []`) instead of proper bitvec values.

This extends `resolveQI` to preserve indices, then detects the `bv{N}` pattern with a numeral width index in `translateFromDDMTermToUntyped` to reconstruct the correct `.prim (.bitvec (BitVec.ofNat width value))`.

Tested: existing tests pass, new round-trip decode tests added for bitvec literals.


MikaelMayer:
🤖⛏️ 

This PR fixes the bitvec literal decoding in the SMT counter-example decoder.

**Changes:**
- `resolveQI` now returns indices from `iden_indexed` identifiers (previously discarded)
- `translateFromDDMTermToUntyped` detects the `bv{N}` pattern with a numeral width index and reconstructs `.prim (.bitvec (BitVec.ofNat width value))`

**Testing:**
- Added 3 round-trip decode tests (parse DDM term string → decode → verify bitvec value)
- All existing tests pass
- Note: executable linking fails due to a pre-existing environment issue (unrelated to this change)

MikaelMayer (commented):
🤖🔍 Clean, well-scoped fix. The approach of extending `resolveQI` to preserve indices and then pattern-matching the `bv{N}` + numeral-width pattern is straightforward. The fallthrough to `mkUFApp` when the pattern doesn't fully match is correct defensive behavior. Tests cover the key cases well.

One minor observation: the `qual_identifier_args` branch (line 347) also discards indices via `_`. If a solver ever returns an indexed identifier *with* arguments (e.g., `((_ extract 7 0) x)`), the indices would be lost there too. That's pre-existing behavior and out of scope for this PR, but worth noting for future work.

atomb (approved):

joscoh (approved):

PR #1271, opened 2026-05-27
Fix SMT translator to encode Term.none/Term.some instead of throwing

### LINKED ISSUE #1231: SMT translator throws on `Term.none`/`Term.some`; `panic!` swallows the error and emits an empty body
[`Translate.lean#L189`](https://github.com/strata-org/Strata/blob/d66fff99ae094ac1e0b8d093875387082f2e2d45/Strata/DL/SMT/DDMTransform/Translate.lean#L189):

```lean
| .none _ | .some _ => throw "don't know how to translate none and some"
```

The encoder's `encodeTerm` and the top-level `encode` both treat these terms as supported (they declare the `Option` datatype up front). `termToSMTString` swallows the throw via `panic!`, returning `""`, which produces a `(define-fun … () (Option Bool) )` with empty body — solvers reject it.

**Fix**: add proper SMT-LIB encoding cases to `translateFromTerm`:

```lean
| .none ty =>
  let retSort ← translateFromTermType (.option ty)
  let qi := QualIdentifier.qi_isort srnone (mkIdentifier "none") retSort
  return .qual_identifier srnone qi
| .some inner =>
  let innerTerm ← translateFromTerm inner
  let qi := QualIdentifier.qi_ident srnone (mkIdentifier "some")
  return .qual_identifier_args srnone qi (Ann.mk srnone #[innerTerm])
```

**Test plan**: write a test that calls `encodeTerm` on `Term.none .bool` and on `Term.some (.prim (.bool true))`, and verifies the resulting SMT-LIB validates (round-trip parseable, no empty body). Verify the test fails before the fix.

MikaelMayer:
Fixes #1231

`translateFromTerm` threw on `Term.none` and `Term.some`, which caused `termToSMTString` to produce an error (or previously, via `panic!`, an empty body that solvers reject).

This PR adds proper SMT-LIB encoding:
- `Term.none ty` → `(as none (Option T))`
- `Term.some inner` → `(some <inner>)`

Tested: existing tests pass, new `#guard_msgs` tests added for both cases at the `termToString` and `termToSMTString` levels.

### DIFF
--- a/Strata/DL/SMT/DDMTransform/Translate.lean
+++ b/Strata/DL/SMT/DDMTransform/Translate.lean
@@ -186,7 +186,14 @@ partial def translateFromTerm (t:SMT.Term): Except String (SMTDDM.Term Provenanc
   | .var v =>
     return .qual_identifier smtProv (.qi_ident smtProv (.iden_simple smtProv
       (.symbol smtProv (mkSimpleSymbol v.id))))
-  | .none _ | .some _ => throw "don't know how to translate none and some"
+  | .none ty =>
+    let retSort ← translateFromTermType (.option ty)
+    let qi := QualIdentifier.qi_isort smtProv (mkIdentifier "none") retSort
+    return .qual_identifier smtProv qi
+  | .some inner =>
+    let innerTerm ← translateFromTerm inner
+    let qi := QualIdentifier.qi_ident smtProv (mkIdentifier "some")
+    return .qual_identifier_args smtProv qi (smtAnn #[innerTerm])
   | .app op args retTy =>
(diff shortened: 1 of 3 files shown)


MikaelMayer (commented):
🤖🔍 Clean, well-scoped fix. The encoding is consistent with how the `Option` datatype is declared in `Encoder.lean` (`none` nullary + polymorphic → needs `(as none (Option T))`, `some` unary → type inferred from argument). Tests cover both the pure translation layer and the solver-level API. No issues found.

atomb (approved):

aqjune-aws (approved):

PR #1272, opened 2026-05-28
Rename StrataDDM-internal namespaces from Strata.* to StrataDDM.*

joehendrix:
Completes the DDM package extraction by renaming all StrataDDM-internal Lean namespaces from `Strata.*` to `StrataDDM.*`, giving the standalone package its own namespace root.

- Definitions that live in the `StrataDDM` package now use the `StrataDDM` namespace (e.g., `StrataDDM.Program`, `StrataDDM.Elab`, `StrataDDM.Parser`). External consumers add `open StrataDDM` or targeted `open StrataDDM (...)` to access them.
- `SourceRange` is re-exported from the `Strata` namespace via `Strata.Util.FileRange` since it is a general utility type not tied to DDM.
- DDM-specific types (`Program`, `Dialect`, `Arg`, `Ann`, etc.) require an explicit `open StrataDDM` rather than being silently available through the `Strata` namespace.
- Trace class names updated to `StrataDDM.*` (breaking change for anyone setting trace options).
- `#guard_msgs` expected output updated where byte offsets shifted or printed names changed.

### DIFF
--- a/StrataDDM/StrataDDM/AST.lean
+++ b/StrataDDM/StrataDDM/AST.lean
@@ -18,7 +18,7 @@ import all StrataDDM.Util.DecideProp
 set_option autoImplicit false
 
 public section
-namespace Strata
+namespace StrataDDM
 open Std (ToFormat Format format)
 
 abbrev DialectName := String
@@ -46,7 +46,7 @@ def ofString (fullname : String) : Option QualifiedIdent := do
     none
 
 section
-open _root_.Lean
+open Lean
 public protected def quote (i : QualifiedIdent) : Term :=
   Syntax.mkCApp ``QualifiedIdent.mk #[quote i.dialect, quote i.name]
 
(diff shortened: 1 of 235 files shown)


shigoel (approved):

aqjune-aws:
Hi, I think this pull request will introduce conflicts to #1216 - what would a good order be..?

joehendrix:
@aqjune-aws We discussed this first since it is largely mechanical then #1216.

StrataDDM/StrataDDM/AST.lean
```diff
@@ -46,7 +46,7 @@ def ofString (fullname : String) : Option QualifiedIdent := do
     none
 
 section
-open _root_.Lean
```
atomb, line 49: I'm happy to see these prefixes going away. :)

atomb (approved):

PR #1275, opened 2026-05-28
Fix addTypeAlias: resolve chained aliases and preserve typeArgs

### LINKED ISSUE #1229: Lambda type aliases: chained aliases not fully resolved; stored `typeArgs` use internal fresh names

MikaelMayer:
Fixes #1229

Chained type aliases (e.g. `BarAlias a := FooAlias a` where `FooAlias` is itself an alias) were not being resolved — `resolveAliases` would return the intermediate alias instead of the concrete type.

The fix de-aliases the RHS in `addTypeAlias` before storing, maintaining the invariant that stored alias types are never themselves aliases. This also switches from using `lhs.freeVars` (which produced internal names like `$__ty0`) to using the user-supplied `alias.typeArgs` directly.

Tested: existing tests pass, new tests added for chained resolution and preserved typeArgs.


MikaelMayer:

PR #1277, opened 2026-05-28

### LINKED ISSUE #1245: DDM parser: `checkLeftRec` panics if leading argument is `NewlineSepBy`

MikaelMayer:
Fixes #1245

`checkLeftRec` in the DDM parser panics with "Unknown parametric category" when an op's leading argument is `NewlineSepBy`, because that category was missing from the `isListCategory` check.

This adds `Init.NewlineSepBy` to the list of recognized parametric categories, alongside `CommaSepBy`, `SpaceSepBy`, etc.

Tested with a regression test dialect that uses `NewlineSepBy` as the leading argument of an op.


MikaelMayer:

PR #1278, opened 2026-05-28
Report arity mismatch error for type aliases in aliasDef/tconsAlias

### LINKED ISSUE #1239: Lambda type aliases: `aliasDef?` / `tconsAlias` silently return original type on arity mismatch

MikaelMayer:
Fixes #1239

## Problem

`tconsAliasSimple` combined name matching and arity checking in a single `find?` predicate. When a type alias name matched but the arity didn't, `find?` returned `none` and the function silently returned the original unresolved type, leading to confusing downstream type errors.

## Solution

Split the name-match from the arity-check in `tconsAliasSimple` and `tconsAlias`:
- `tconsAliasSimple` now returns `Except Format LMonoTy`
- When the name matches but arity doesn't, it returns a clear error message
- Callers (`aliasDef?`, `resolveAliases`) propagate the `Except`
- All proofs updated to handle the new control flow

## Testing

- Added test for arity mismatch error detection
- All existing tests pass


MikaelMayer:

PR #1279, opened 2026-05-28

### LINKED ISSUE #1241: SMT `Solver.spawn`: stderr pipe is opened but reader handle is dropped; solver may SIGPIPE on fatal-error output

MikaelMayer:
Fixes #1241

## Problem

`Solver.spawn` was opening stderr as `.piped`, but the pipe was never read. When an SMT solver wrote enough diagnostic output to fill the OS pipe buffer (~64KB), it would receive SIGPIPE and terminate unexpectedly.

## Solution

Changed `stderr := .piped` to `stderr := .inherit` so solver diagnostic output (fatal errors, OOM, unsupported logic messages) goes directly to the verifier's stderr. This prevents the pipe buffer from filling up and avoids SIGPIPE.

## Testing

Existing tests pass. The change is minimal and only affects how stderr is routed.


MikaelMayer:
🤖 Closing: wrong base (main). Will restart from main2.

PR #1280, opened 2026-05-28
Fix Core translator: local function declaration builds wrong arrow type for ≥3 distinct-type args

### LINKED ISSUE #1226: `Core` translator: local `function` declaration builds wrong arrow type for ≥3 distinct-type args

MikaelMayer:
Fix local function declaration building wrong arrow type for ≥3 distinct-type args.

The `funcDecl_statement` case in the Core translator used `mkArrow` with reversed inputs, producing incorrect type ordering (e.g. `int → real → bool → int` instead of `int → bool → real → int`) for local functions with 3+ parameters of distinct types. Replaced with `mkArrow'` which takes the input list in order, conforming to the Denote-layer pattern used in Assumptions.lean, CallOfLFuncDenote.lean, and Semantics.lean.

Tested with a regression test that declares a local function `f(x:int, b:bool, r:real):int` and calls it, plus a `#guard` that directly verifies the `mkArrow'`/`destructArrow` roundtrip invariant. Existing tests pass.

Fixes #1226


Strata/Languages/Core/DDMTransform/Translate.lean
```diff
@@ -1439,7 +1439,7 @@ partial def translateStmt (p : Program) (bindings : TransBindings) (arg : Arg) :
     -- The DDM parser's @[scope(b)] on the body adds only the parameters.
     -- The function name is NOT in scope inside the body (declareFn adds it
     -- for subsequent statements only). So body bindings = outer + parameters.
-    let funcType := Lambda.LMonoTy.mkArrow outputMono (inputs.values.reverse)
+    let funcType := Lambda.LMonoTy.mkArrow' outputMono inputs.values
```
tautschnig, line 1454: Same `mkArrow' output (inputs.map Prod.snd)` pattern is used at `Strata/DL/Lambda/Denote/Assumptions.lean:73` and `CallOfLFuncDenote.lean:673`. Worth extracting one helper (e.g. `Core.opTypeOf : LMonoTy → ListMap _ LMonoTy → LMonoTy` or even `LFunc.opMonoType`) and routing every site through it — the bug only surfaced because this site reinvented the formula and got it wrong. With a single helper, `LMonoTy.destructArrow_mkArrow'` (suggested in the top-level review) becomes a theorem about that helper and the contract is enforced by typechecking.

While here: please drop the now-stale `inputs.values.reverse` reasoning from the surrounding comment (the lines above still talk about parameters being added — fine — but if any explanation of the type construction remains, it should match the new form).

StrataTest/Languages/Core/Tests/FuncDeclStmtTest.lean
```diff
@@ -42,3 +42,33 @@ procedure test ()
...
+info: ok: program Core;
+
+procedure test ()
+{
+  function f (x : int, b : bool, r : real) : int { x }
+  var z : int := f(1, true, 0.0);
+};
+-/
+#guard_msgs in
+#eval (Std.format ((Core.typeCheck .default (translate localFuncDistinctTypesPgm).stripMetaData)))
```
tautschnig, line 74: Good regression — pre-fix the call `f(1, true, 0.0)` would fail unification against `int → real → bool → int`. Two improvements that would tighten this:

1. The current `#guard_msgs` only checks the pretty-printed *surface* form (`function f (x : int, b : bool, r : real) : int`) which doesn't depend on the internal `funcType`. Add an assertion that touches the constructed `funcType` directly — either by inspecting `decl.body` for the `.op f (some funcType)` node and `#guard`-ing its `LMonoTy.destructArrow`, or simpler, add a second program where the call uses arguments **in the parameter order** vs **rearranged** so a re-introduction of the reverse bug can't slip past with type-check still succeeding by coincidence.
2. Worth a one-liner `#guard` directly at the `LMonoTy` level: `#guard (LMonoTy.mkArrow' .int [.int, .bool, .real]).destructArrow = [.int, .bool, .real, .int]`. That nails the contract `mkArrow'` is supposed to satisfy and is independent of the whole translation pipeline.

tautschnig (commented):
The PR description / issue are slightly inaccurate when they say *"matching how top-level `translateFunction` already works"*: top-level `translateFunction` (around line 1756) does not build a `funcType` at all — it stores `inputs`/`output` directly on the `LFunc`. The pattern this PR is conforming to is the Denote-layer pattern. Worth mentioning so a future reader doesn't hunt for a non-existent twin.

Proof / test suggestions (none currently exist linking `mkArrow'` to `destructArrow`):
- `theorem LMonoTy.destructArrow_mkArrow' (ret : LMonoTy) (ins : LMonoTys) : LMonoTy.destructArrow (mkArrow' ret ins) = ins ++ [ret]` — locks in the very ordering invariant that was violated. With this `simp` lemma the original buggy `mkArrow outputMono (inputs.values.reverse)` would have left a goal that didn't close.
- `theorem LMonoTy.mkArrow_eq_mkArrow' (head : LMonoTy) (tail : LMonoTys) (h : tail ≠ []) : mkArrow head tail = mkArrow' (tail.getLast h) (head :: tail.dropLast)` — characterises `mkArrow` in terms of `mkArrow'`, making the *intended* shape explicit and discouraging future use of `mkArrow` for the input/output split.

Strata/Languages/Core/DDMTransform/Translate.lean
MikaelMayer: 🤖 Agreed that extracting a helper (e.g. `LFunc.opMonoType`) would centralize this pattern and prevent future drift. The pattern appears in 5+ files including proof files (Assumptions.lean, CallOfLFuncDenote.lean, LExprSemanticsConsistent.lean), so this is better as a follow-up refactoring. Filed as a TODO.

Re stale comments: the surrounding comment (lines 1439-1441) is about scope/bindings, not about the type construction — it remains accurate. There was no comment explaining the old `inputs.values.reverse` pattern.

StrataTest/Languages/Core/Tests/FuncDeclStmtTest.lean
MikaelMayer: 🤖 Added a `#guard` that directly tests the `mkArrow'`/`destructArrow` roundtrip invariant at the `LMonoTy` level:

```lean
#guard (Lambda.LMonoTy.mkArrow' .int [.int, .bool, .real]).destructArrow
    == [.int, .bool, .real, .int]
```

This nails the ordering contract independently of the translation pipeline.

137e52c8e

tautschnig (approved):

atomb (approved):

PR #1281, opened 2026-05-28

### LINKED ISSUE #1245: DDM parser: `checkLeftRec` panics if leading argument is `NewlineSepBy`

MikaelMayer:
Fixes #1245

## Problem

`checkLeftRec` in the DDM parser panics with "Unknown parametric category" when an op's leading (first) syntax argument is `NewlineSepBy`. This is because `NewlineSepBy` was missing from the `isListCategory` check.

## Solution

- Use the existing `SepFormat.isParametricCategory` helper in `checkLeftRec` instead of a manual disjunction of category names. This ensures any separator format recognized by `SepFormat` is automatically handled, preventing future omissions.
- Fix the passthrough case in `opSyntaxParser` to also reject left-recursive parametric categories (e.g. `op foo (xs : NewlineSepBy Item) : Item => xs`), which was previously silently accepted.

## Testing

- Added a regression test with a dialect that uses `NewlineSepBy` as the leading argument of an op, asserting both formatted output and parsed AST structure.
- Added a negative test verifying that left-recursive `NewlineSepBy` in passthrough syntax is rejected with an appropriate error.
- Existing tests pass.


Strata/DDM/Parser.lean
```diff
@@ -773,6 +773,7 @@ def checkLeftRec (thisCatName : QualifiedIdent) (argDecls : ArgDecls) (as : List
     let isListCategory := cat.name == q`Init.CommaSepBy ||
                           cat.name == q`Init.SpaceSepBy ||
                           cat.name == q`Init.SpacePrefixSepBy ||
+                          cat.name == q`Init.NewlineSepBy ||
```
tautschnig: Same membership test as `catParser` below (lines 911–923) and several `Gen.lean`/`Elab` files. Suggest replacing the disjunction with the existing `SepFormat.fromCategoryName?` helper plus an explicit `Option` check, e.g.

```suggestion
    let isListCategory := (SepFormat.fromCategoryName? cat.name).isSome ||
                          cat.name == q`Init.Option
```

Then the next time someone adds a separator category they only need to extend `SepFormat`.

StrataTest/DDM/NewlineSepByLeading.lean
```diff
@@ -0,0 +1,33 @@
...
+op items (xs : NewlineSepBy Item) : Command => xs ";";
+#end
+
+#guard_msgs in
+private def test_newline_sep_by_leading := #strata
+program NewlineSepByLeadingTest;
+1
+2
+3;
+#end
```
tautschnig: Two gaps in this regression:

1. Empty `#guard_msgs in` only certifies "no panic / no diagnostic". Please assert the parsed AST too (e.g. add an `#eval`/`#guard` on the value of `test_newline_sep_by_leading`, or capture the elaboration output via `#guard_msgs` with the expected info string), so a silent regression in the produced tree is also caught.
2. Add a negative case for the other `checkLeftRec` branch with `NewlineSepBy` — i.e. an op like `op items (xs : NewlineSepBy Items) : Items => xs;` should produce `Leading symbol cannot be recursive call to …`. This locks in the `c.name == thisCatName` arm for the new entry.

tautschnig (commented):
1. **Eliminate the duplicated enumeration that caused this bug.** The same list-of-list-categories appears in `Parser.lean` (`checkLeftRec` and `catParser`), `Integration/Java/Gen.lean` (3 places), `Integration/Lean/Gen.lean`, `Elab/Core.lean`, and `AST.lean`. `NewlineSepBy` was missed in exactly one of them. `Strata/DDM/AST.lean` already has `SepFormat.fromCategoryName?` enumerating these (minus `Option`); promoting that to a single boolean predicate (e.g. `SyntaxCat.isListLike`) and routing all six call sites through it removes the recurrence risk and is a small mechanical refactor. Defining it once also makes the proof obligation below trivial.

2. **Replace the surviving `panic!` with `.invalid`.** The fall-through at line 790 (`panic! "Unknown parametric category … is not supported."`) is the exact mechanism that turned this missing case into a crash rather than a typed error. It can become `.invalid mf!"…"` so future omissions surface as a parser-construction error rather than a process-killing panic.

Proof / `#guard` coverage suggestions (none currently exist for these helpers):
- `SepFormat.fromCategoryName?_args_size : ∀ c, SepFormat.fromCategoryName? c.name = some _ → c.args.size = 1` — guards the `assert! cat.args.size = 1` at the use sites.
- After the refactor: `SyntaxCat.isListLike_iff` characterising exactly which `q\`Init.*` names return true (`#guard` per constructor is enough).
- `checkLeftRec` doesn't panic on any well-formed `SyntaxCat`: a property test that enumerates all `SepFormat` constructors plus `Option` and asserts no panic.

Strata/DDM/Parser.lean
MikaelMayer: 🤖 Replaced the disjunction with `SepFormat.fromCategoryName?` plus an explicit `Init.Option` check: f815e4699

joehendrix (dismissed):

Strata/DDM/Parser.lean
```diff
@@ -770,11 +770,7 @@ def checkLeftRec (thisCatName : QualifiedIdent) (argDecls : ArgDecls) (as : List
     let .isTrue lt := decideProp (v < argDecls.size)
       | return panic! "Invalid index"
     let cat := argDecls[v].kind.categoryOf
-    let isListCategory := cat.name == q`Init.CommaSepBy ||
-                          cat.name == q`Init.SpaceSepBy ||
-                          cat.name == q`Init.SpacePrefixSepBy ||
-                          cat.name == q`Init.Seq ||
-                          cat.name == q`Init.SemicolonSepBy ||
+    let isListCategory := (SepFormat.fromCategoryName? cat.name).isSome ||
```
tautschnig: Reconciling my earlier review against this head:

- **Inline at `Parser.lean:776`** (refactor to `SepFormat.fromCategoryName?`): ✅ addressed in `f815e4699`. Already marked resolved.
- **Inline at `StrataTest/DDM/NewlineSepByLeading.lean:27-33`** (regression test gaps — assert parsed AST, add negative recursion case): ❌ unresolved, both asks still apply at the current head.
- **Top-level point 1** (eliminate the duplicated enumeration across all 6 call sites): partially done. Only the `checkLeftRec` site uses `SepFormat.fromCategoryName?`; manual disjunctions remain at `Strata/DDM/Parser.lean:913`, `Strata/DDM/Integration/Java/Gen.lean:128/145/182-183`, and `Strata/DDM/Integration/Lean/Gen.lean:1500-1509`. Worth a follow-up issue rather than blocking here.
- **Top-level point 2** (replace surviving `panic!` with `.invalid`): not addressed. The panic is now at `Strata/DDM/Parser.lean:785`. Worth a follow-up issue.

MikaelMayer: 🤖 Addressed both asks in 648449aa9:

1. **AST assertion**: Added `#guard` that verifies the parsed command contains an `ArgF.seq` with `.newline` separator and 3 items.
2. **Negative recursion case**: Found that the passthrough case in `opSyntaxParser` was missing the parametric-category left-recursion check (it only checked direct `cat.name == category`). Fixed it to also reject `NewlineSepBy Item : Item` and added a `#guard_msgs` test for the error.

tautschnig (approved):

joehendrix (approved):

PR #1282, opened 2026-05-28
Report arity mismatch error for type aliases

### LINKED ISSUE #1239: Lambda type aliases: `aliasDef?` / `tconsAlias` silently return original type on arity mismatch

MikaelMayer:
Report arity mismatch error for type aliases instead of silently returning original type.

Previously, `tconsAliasSimple` matched aliases by both name and arity simultaneously. When the name matched but arity didn't, `find?` returned `none` and the function silently returned the unresolved type, leading to confusing downstream type errors.

Now the name-match is split from the arity-check: if the name matches but the number of type arguments is wrong, a clear error message is returned (e.g., "Arity mismatch for alias 'X': expected N type argument(s), got M"). `tconsAliasSimple` returns `Except Format LMonoTy` to support this, and callers (`aliasDef?`, `resolveAliases`) propagate the error.

**Existing test rewritten**: the `LMonoTy.aliasDef? mty[myInt] …` test was previously asserting the buggy silent-fallthrough behaviour (no arity check). It has been rewritten to call with valid arity (`mty[myInt nat]`) and now exercises the success path. A new test was added below it to lock in the arity-mismatch error path that this PR introduces.

All existing tests pass. A new test verifies the arity mismatch error is reported.

---

*Note: `Fixes #1239` keyword removed after merge. This PR was unintentionally auto-merged into #1284's stack branch instead of landing on `main2`; the combined fix tracking #1239 is now in #1284. Issue #1239 has been reopened.*

### DIFF
--- a/Strata/DL/Lambda/LExprTypeEnv.lean
+++ b/Strata/DL/Lambda/LExprTypeEnv.lean
@@ -865,12 +865,15 @@ check whether the de-aliased types are registered/known.
 def LMonoTy.tconsAlias [ToFormat IDMeta] (name : String) (args : LMonoTys)
     (Env : TEnv IDMeta) : Except Format (LMonoTy × TEnv IDMeta) := do
   let inputMty := .tcons name args
-  -- Look for a matching alias with same name and arity.
-  let matchingAlias := Env.context.aliases.find?
-                        (fun a => a.name == name && a.typeArgs.length == args.length)
-  match matchingAlias with
+  -- Look for a matching alias by name.
+  let nameMatch := Env.context.aliases.find? (fun a => a.name == name)
+  match nameMatch with
   | none => return (inputMty, Env)
   | some alias =>
+    if alias.typeArgs.length != args.length then
+      .error f!"Arity mismatch for alias '{name}': \
+               expected {alias.typeArgs.length} type argument(s), got {args.length}"
+    else
     -- Create instantiation pair: [alias pattern, alias definition].
     -- The alias pattern and definition share the same type variables here.
     let aliasPattern := .tcons name (alias.typeArgs.map .ftvar)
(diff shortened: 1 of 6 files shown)


MikaelMayer:
Could you update the PR description to call out and justify the fact that an existing test was rewritten?

Specifically, in `StrataTest/DL/Lambda/LExprTypeEnvTests.lean`, the test that previously read

```lean
/-- info: ok: myInt -/
#guard_msgs in
#eval do let (ans, _) ← LMonoTy.aliasDef? mty[myInt] ...
```

was changed to use `mty[myInt nat]` (with the right arity) and now expects `info: ok: int`. A new separate test was added for the arity-mismatch error case.

Per `AGENTS.md` ("NEVER update the output of a test so that it matches an error that did not exist"), this is a borderline case — but defensible here because the original test was inadvertently codifying the buggy silent-fallthrough behaviour that this PR is fixing. The rewrite makes the original test exercise its **intended** assertion (success path with valid arity), and the new arity-mismatch test pins down the regression boundary for the bug fix.

Adding a paragraph along these lines to the PR description would save reviewers from having to re-derive the conclusion themselves and make the audit trail explicit. Suggested wording:

> **Existing test rewritten**: the `LMonoTy.aliasDef? mty[myInt] …` test was previously asserting the buggy silent-fallthrough behaviour (no arity check). It has been rewritten to call with valid arity (`mty[myInt nat]`) and now exercises the success path. A new test was added below it to lock in the arity-mismatch error path that this PR introduces.

MikaelMayer:
🤖 Updated the PR description with a paragraph explaining and justifying the rewritten test, using the suggested wording.

PR #1283, opened 2026-05-28

MikaelMayer:
## Problem

When the SMT solver process writes to stderr, the parent process's pipe buffer can fill up. If the parent never reads from the stderr pipe, the solver process receives SIGPIPE and terminates unexpectedly.

## Solution

Changed `Solver.spawn` to use `.inherit` for the stderr pipe instead of `.piped`, so solver stderr output goes directly to the parent's stderr stream. This prevents SIGPIPE by eliminating the unread pipe buffer.

## Testing

Existing tests pass. Added a regression test (`StrataTestExtra/DL/SMT/SolverSpawnStderrTest.lean`) that spawns a process writing 200 KB to stderr and verifies it responds without being killed by SIGPIPE.


tautschnig (commented):
1. **Stale struct-level doc.** `Strata/DL/SMT/Solver.lean` lines 47–50 (just outside this diff) still claim *"We don't have an error stream here, since we configure solvers to run in quiet mode and not print anything to the error stream"* — exactly the assumption #1241 invalidates. Please update it to match the new `spawn` doc, e.g. *"Solver stderr is inherited from the parent process and not captured here. Solvers configured with `--quiet` are usually silent on stderr, but fatal diagnostics (OOM, parse errors, unsupported logic) still go directly to the verifier's stderr."*

2. **Regression test missing.** The issue's "Test plan" sketches a reliable repro: spawn a fake "solver" (a tiny shell command suffices, e.g. `bash -c 'yes "diagnostic" | head -c 200000 >&2; cat >/dev/null'`) and drive it through `Solver.spawn` + a few large `assert`s. Under the old code this hangs or kills the child once the 64 KiB stderr pipe fills; under the new code it completes cleanly. This is a perfect candidate for `StrataTestExtra/DL/SMT/` so it actually exercises the OS pipe behaviour. Without a test, this same regression can recur silently.

3. **`.inherit` vs `.null` choice.** `.inherit` is the right default (users see OOM/parse errors), but it does change observable behaviour: solver chatter now reaches whatever stderr the parent has — including CI logs and `strata`'s own users. Worth a brief sentence in the PR description acknowledging the tradeoff vs. `.null`. No code change required if `.inherit` is the intended UX.

MikaelMayer:
🤖 Updated the struct-level doc comment for `SMTLibSolver` to reflect the current stderr handling (inherited by parent process via `spawnSolver`). Will be in the next commit.

aqjune-aws (dismissed):

MikaelMayer:
Something got wrong. This PR claims it modified 632 files. Investigate why and fix that

MikaelMayer:
🤖 Fixed. The branch was based on `main2` instead of `main`, which caused it to carry hundreds of unrelated commits. I rebased the branch onto `main` keeping only the 3 relevant commits, and updated the PR base to `main`. The PR now shows 1 file changed as expected (commit b25e20a95).

Strata/DL/SMT/Solver.lean
```diff
@@ -44,9 +45,9 @@ deriving DecidableEq, Repr
  the input stream. We assume that both the input and output streams conform to
  the SMTLib standard: the inputs are SMTLib script commands encoded as
  s-expressions, and the outputs are the s-expressions whose shape is determined
- by the standard for each command. We don't have an error stream here, since we
- configure solvers to run in quiet mode and not print anything to the error
- stream.
+ by the standard for each command. The solver's stderr is inherited by the parent process (see `spawnSolver`),
+ so diagnostic output goes directly to the verifier's stderr rather than
+ through a field on this structure.
```
tautschnig, line 49: **Fix `spawnSolver` (no such symbol; the function is `Solver.spawn`) and rewrap — line 48 is 110 chars vs ~78 for the rest of the doc.**

<details>
<summary>Why</summary>

Repo-wide grep for `spawnSolver` returns only this doc comment; the function is defined as `Solver.spawn` at `Strata/DL/SMT/Solver.lean:108`.

```suggestion
 by the standard for each command. The solver's stderr is inherited by the
 parent process (see `Solver.spawn`), so diagnostic output goes directly
 to the verifier's stderr rather than through a field on this structure.
```
</details>

Strata/DL/SMT/Solver.lean
```diff
@@ -44,9 +44,9 @@ deriving DecidableEq, Repr
  the input stream. We assume that both the input and output streams conform to
  the SMTLib standard: the inputs are SMTLib script commands encoded as
  s-expressions, and the outputs are the s-expressions whose shape is determined
- by the standard for each command. We don't have an error stream here, since we
- configure solvers to run in quiet mode and not print anything to the error
- stream.
+ by the standard for each command. The solver's stderr is inherited by the parent process (see `spawnSolver`),
```
tautschnig, line 47: **Regression test** (point 2 from my prior review): not added. Suggest a follow-up under `StrataTestExtra/DL/SMT/` driving `Solver.spawn` against `bash -c 'yes diag | head -c 200000 >&2; cat >/dev/null'` (or equivalent). Without it the same SIGPIPE regression can recur silently. Not blocking this PR.

MikaelMayer: 🤖 Added regression test in `StrataTestExtra/DL/SMT/SolverSpawnStderrTest.lean` — spawns a bash process that writes 200 KB to stderr then acts as a trivial solver. Verifies the solver responds without being killed by SIGPIPE.

Commit: 30343f9e2

joscoh (approved):

aqjune (approved):

tautschnig (approved):

Strata/DL/SMT/Solver.lean
MikaelMayer: 🤖 Fixed: renamed `spawnSolver` → `Solver.spawn` and rewrapped to ~78 chars per line.

PR #1285, opened 2026-05-28
Fix #1244: add missing s! prefix to error message in parseStrataProgramFromDialect

### LINKED ISSUE #1228: Ion deserializer: NOP pad in struct shifts subsequent field-key indices

### LINKED ISSUE #1244: DDM Elab: error message in `parseStrataProgramFromDialect` is a plain string, not interpolated

MikaelMayer:
Fixes #1244

`parseStrataProgramFromDialect` in `Strata/DDM/Elab.lean` threw `IO.userError "Internal {dialect} missing from loaded dialects."` as a plain string literal, missing the `s!` prefix. The user saw the literal text `{dialect}` in the error message instead of the actual dialect name.

This PR:
- Adds the missing `s!` prefix to the original error message
- Fixes 5 additional instances of the same bug across the codebase (in `panic!`, `throw`, and `dbg_trace` calls)
- Adds a CI lint check (`.github/scripts/checkMissingInterpolation.sh`) that detects this class of bug going forward

Tested with a new `#guard_msgs` test that triggers the error and asserts the interpolated value appears. Existing tests pass. The new CI lint check passes on the fixed codebase.


MikaelMayer:
Heads up — I edited the **body** of this PR. The **title** was already correct ("Fix #1244: …") and I left it unchanged.

**Why**: the previous body said `Fixes #1228` and described the Ion NOP-pad fix, which appears to be stale copy-paste from #1286. The actual code change in this PR is the one-character `s!` interpolation fix in `Strata/DDM/Elab.lean` for issue #1244, so the body needed to match.

**Effect on auto-close**: GitHub closes issues based on the body keyword (`Fixes #N`), so without this edit, merging this PR would have auto-closed #1228 (already covered by #1286) instead of #1244. After the edit, merging this PR will correctly auto-close #1244.

**No code or test changes** — only the PR description was rewritten. If you re-run any "verify body matches diff" tooling, it should now agree.

MikaelMayer:
🤖 Thanks for fixing the PR description — no code change needed on my end. The body now correctly reflects the actual one-character `s!` fix for #1244.

Strata/DDM/Elab.lean
```diff
@@ -516,7 +516,7 @@ def parseStrataProgramFromDialect (dialects : LoadedDialects) (dialect : Dialect
   let input := normalizeInputContext input
 
   let isTrue mem := (inferInstance : Decidable (dialect ∈ dialects.dialects))
-    | throw <| IO.userError "Internal {dialect} missing from loaded dialects."
+    | throw <| IO.userError s!"Internal {dialect} missing from loaded dialects."
```
tautschnig, line 519: The same missing-`s!` bug exists at four other sites, all printing literal `{…}` to users:

- `StrataTest/Util/Python.lean:107` — `throw <| .userError "{pythonCmd} not found."`
- `Strata/DDM/Elab/Core.lean:976` — `panic! "Invalid index {idx}"`
- `Strata/Languages/Python/PythonToCore.lean:246` — `panic! "Expected some constant string: {e}"`
- `Strata/Languages/Python/PythonToCore.lean:557` — `panic! "Unsupported UnaryOp: {repr e}"`

Please either fix them in this PR (each is a 1-char change) or open tracking issues. Also worth considering a lightweight CI check (e.g. a `grep` test in `lake test`) for `"…{[a-zA-Z_]"` inside `throw`/`panic!`/`userError` arguments without an `s!` prefix — this class of bug is invisible to the type checker and recurs easily.

For the corresponding fixes, the natural test pattern is the one you've already used: a `#guard_msgs` `#eval` that triggers the panic/throw and asserts the placeholder is interpolated.

MikaelMayer: I love the idea of a CI check too. Let's add this for more robustness

MikaelMayer: 🤖 Fixed all 5 missing `s!` prefixes (the 4 from the original comment + 1 more in `Strata/Util/Statistics.lean` found by the new lint check) and added a CI lint script `.github/scripts/checkMissingInterpolation.sh` that detects this class of bug going forward. See commit 9d510d8ea.

tautschnig (approved):

keyboardDrummer (approved):
Looks good, thanks!

PR #1286, opened 2026-05-28
Fix #1228: skip field-key symbol push for NOP pads in struct deserialization

### LINKED ISSUE #1228: Ion deserializer: NOP pad in struct shifts subsequent field-key indices

MikaelMayer:
Fixes #1228

When deserializing Ion binary structs, NOP pads (type code 0 with length != 15) between fields caused subsequent field-key symbol indices to shift, because the field-key symbol ID was unconditionally pushed to the symbol stack even for NOP pads.

The fix introduces a `TypeDesc.isNopPad` helper and uses it to skip pushing the field-key symbol for NOP pads, so subsequent fields retain their correct symbol indices.

Tested with `#guard` tests covering NOP pad between fields and NOP pad at start of struct. Existing tests pass.


Strata/DDM/Util/Ion/Deserialize.lean
```diff
@@ -527,7 +527,10 @@ def deserializeAux (bytes : ByteArray) (ds : DeserializeState bytes.size)
         let (symId, ⟨off, offp⟩) ← readVarUInt bytes ds.off
         let .isTrue offp := decideProp (off < bytes.size)
           | rfail off "Unexpected end of file"
-        let ds := { ds with symbols := ds.symbols.push ⟨symId⟩ }
+        let td := TypeDesc.ofByte bytes[off]
+        let isNop := td.code == 0 && td.length != 15
```
tautschnig: Two suggestions:

1. Extract a named helper for clarity and reuse:
   ```lean
   def TypeDesc.isNopPad (d : TypeDesc) : Bool := d.code == 0 && d.length != 15
   ```
   Then `let isNop := td.isNopPad`. Document the spec rationale (length 15 ⇒ `null.null`).

2. The bigger maintainability issue this fix exposes is the implicit invariant that **the symbols and values stacks must stay aligned in struct contexts**. It's currently the caller's job to push to `symbols` exactly when `readToken` pushes to `values`; NOPs broke that invariant silently for years. Worth a follow-up to either (a) merge the two stacks into a single `Array (SymbolId × Value)` for struct contexts, or (b) move the symbol push down inside the appropriate `read*Token` callees so the invariant is local. Out of scope for this PR but please file an issue.

StrataTest/DDM/Util/Ion.lean
```diff
@@ -60,3 +60,14 @@ def testRoundtrip (v : List (Ion SymbolId)) : Bool :=
...
+-- Issue #1228: NOP pad in struct must not shift subsequent field-key indices
+-- Input: struct with fields (key=1, int 1), NOP pad, (key=2, int 2)
+private def nopPadInStruct : ByteArray :=
+  ⟨#[0xE0, 0x01, 0x00, 0xEA, 0xD8, 0x81, 0x21, 0x01, 0x80, 0x00, 0x82, 0x21, 0x02]⟩
+
+#guard
+  match Ion.deserialize nopPadInStruct with
+  | .ok #[#[.mk (.struct fields)]] =>
+    fields == #[(.mk 1, .int 1), (.mk 2, .int 2)]
+  | _ => false
```
tautschnig, line 73: Single `#guard` on a hand-crafted byte string is brittle and undertests. Please add:

1. **NOP at start of struct**: `(field, NOP, field)` already covered; also test `(NOP, field)` to catch the symbol push at struct entry.
2. **NOP at end of struct**: ensures `popFinishedValues` after the NOP doesn't drop a phantom symbol.
3. **Consecutive NOPs**: two NOPs in a row inside a struct.
4. **Long NOP**: NOP with `td.length = 14` plus a VarUInt length (exercises non-zero NOP body).
5. **NOP in nested struct**: outer struct contains an inner struct that has a NOP — verifies the bug isn't masked by stack interactions across nesting.

For stronger coverage, the natural property is **NOP-pad invariance**:
```lean
-- For any valid Ion byte sequence bs and any insertion of a well-formed NOP
-- pad in any struct context, deserialization yields the same logical Ion value.
example : ∀ (vs : List (Ion SymbolId)),
    Ion.deserialize (Ion.serialize vs) =
      Ion.deserialize (insertNopPadsAtAllStructPositions (Ion.serialize vs)) := by plausible
```
A hand-rolled `insertNopPadsAtAllStructPositions` (or a generator that produces "equivalent" byte sequences with random NOPs) plus Plausible would give meaningful coverage and is the natural setup for a future formal proof of NOP-pad invariance.

Strata/DDM/Util/Ion/Deserialize.lean
MikaelMayer: Let's do the extraction and file an issue

MikaelMayer: 🤖 Extracted `TypeDesc.isNopPad` as a named helper and updated the usage site.

StrataTest/DDM/Util/Ion.lean
MikaelMayer: 🤖 Added a `nopPadAtStartOfStruct` test that verifies `(NOP, field)` deserializes correctly.

tautschnig (approved):

joscoh (approved):

Issue #1287, opened 2026-05-28
Free type variables in function declarations can be captured


joscoh:
## Summary

`Function.typeCheck` can silently merge semantically distinct type variables when a free (undeclared) type variable name collides with a generated fresh variable name (`$__tyN`). This causes the typechecker to accept ill-typed function bodies.

## Reproduction

```lean
import Strata.Languages.Core.FunctionType

open Lambda Imperative

private def C : Core.Expression.TyContext := LContext.default
private def Env : Core.Expression.TyEnv := default

-- f<T>(x : T, y : $__ty0) : T { y }
-- $__ty0 is NOT in typeArgs — semantically a different type from T.
-- The body `y` should be rejected: it has type $__ty0, not T.
private def unsoundFunc : Core.Function :=
  { name := ⟨"bad", ()⟩,
    typeArgs := ["T"],
    inputs := [(⟨"x", ()⟩, .ftvar "T"), (⟨"y", ()⟩, .ftvar "$__ty0")],
    output := .ftvar "T",
    body := some (LExpr.fvar () ⟨"y", ()⟩ none) }

-- Typechecks successfully (should fail):
-- Output: typeArgs: [$__ty0], inputs: (x, $__ty0) (y, $__ty0), output: $__ty0
#eval do
  let (func', _) ← Core.Function.typeCheck C Env unsoundFunc
  return format f!"typeArgs: {func'.typeArgs}\ninputs: {func'.inputs}\noutput: {func'.output}"
```

## Root Cause

In `Function.typeCheck` (FunctionType.lean):

1. `func.type` builds `.forAll ["T"] (T → $__ty0 → T)` — the free `$__ty0` appears inside the forAll body.
2. `LTy.instantiateWithCheck` calls `LTy.instantiate`, which generates a fresh var `$__ty0` (genState starts at 0) and substitutes it for `T`.
3. The generated `$__ty0` (for `T`) and the pre-existing free `$__ty0` are now the same string — they merge into a single type variable.
4. `monoty.freeVars.eraseDups` finds only `["$__ty0"]` — two intended type parameters collapsed into one.
5. The body `y` now has the same type as `x` and the return type, so it passes.

## Impact

- The concrete syntax parser rejects `$__ty` prefixed identifiers, so this cannot be triggered through normal surface programs.
- It could be triggered by a buggy translator or direct AST construction.
[...]

Issue #586, opened 2026-03-16
Soundness bug in a program-level type inference


aqjune-aws:
Found by @joscoh 

```
def fooPgm : Program := { decls := [
  .func { name := "foo",
          typeArgs := ["a"],
          inputs := [("x", .ftvar "a")],
          output := .ftvar "a",
          body := some eb[((~Int.Add x) #1)] },
  .proc { header := { name := "callFoo",
                      typeArgs := [],
                      inputs := [],
                      outputs := [] },
          spec := { modifies := [],
                    preconditions := [],
                    postconditions := [] },
          body := [
            Statement.init "s" (.forAll [] .string) (some eb[#hello]) .empty,
            Statement.init "r" (.forAll [] .string) (some eb[(~foo s)]) .empty
          ]
  }
]}

/-- info: [Strata.Core] Type checking succeeded.

---
info: ok: function foo<$__ty0> (x : $__ty0) : $__ty0 {
  x + 1
}
procedure callFoo () returns ()
{
  var s : string := "hello";
  var r : string := foo(s);
  };-/
#guard_msgs in
#eval (typeCheck .default fooPgm)
```

A copied message:

> Since type inference is only run at the expression level, the type of the overall function remains \forall a, a -> a even though the body has been inferred to have type int -> int. This means that we can subsequently call it on a string and the typechecker is fine with this.
> In OCaml, for example, the type of foo would be inferred as int -> int which prevents this. But to fix this we would need to modify type inference to be able to change the types of functions and procedures, which could be a very large refactor and extremely tricky for mutually recursive functions.

My suggestion was:

> I think another easy solution is to consider the $__ty0 as a generic of Java. In the case of Java generic, $__ty0 is more like an arbitrarily named type constructor. Since it is not a free type variable, Java cannot unify x 's type with int , instead considers $__ty0 as an opaque type that cannot be handled. Therefore, foo will simply fail type checking.

> The Java/Boogie style will still allow polymorphic type. It will simply state that if an expression like x + 1 is found, it will raise an error stating that "we only know that x has some polymorphic type, but we cannot ensure that it can be unified to int".

> Actually, this bug that Josh found is pretty connected to the definition of well-formedness that I had to specify in the soundness of type inference algorithm - it states that no input variable like x will have a free type variable like $__ty0. Without this LExpr.resolve had a counterexample... I was somewhat not fully confident that once we allow this LExpr.resolve can be fixed to be a bug-free.

[...]

PR #1288, opened 2026-05-28
Fix bugs in Function typing

joscoh:
Fixes three bugs in Function typing:
1. https://github.com/strata-org/Strata/issues/1287 
2. https://github.com/strata-org/Strata/issues/586 
3. https://github.com/strata-org/Strata/issues/1289

For the first, it adds a check to ensure that all type variables in the inputs/outputs are in the declared `typeArgs`. For the second, it checks that the inferred type is alpha-equivalent to the declared one, then applies the resulting type substitution to ensure that all annotations in the body are consistent with the (renamed) `typeArgs`. This PR includes tests for the new typechecker in `FunctionTypeTests.lean`. For the third, it checks to ensure there are no unbound type variables in the body.

This PR also changes the resulting function's type arguments to match the inputs rather than using fresh variables, resulting in more readable output.

## Design: exact polymorphism

 Unlike OCaml, where type annotations are lower bounds (the body may be more specific), we
 require exact polymorphism: if `f<a>(x: a): a` is declared, the body cannot force `a = int`.
 This is appropriate for an IR where the user can give annotations as needed.


Strata/DL/Lambda/LExpr.lean
```diff
@@ -340,6 +340,18 @@ def collectFvarNames {T : LExprParamsT} : LExpr T → List (Identifier T.base.ID
...
+  match e with
+  | .abs _ _ (some ty) body => LMonoTy.freeVars ty |>.foldl .insert (annotationTyVars body)
+  | .abs _ _ none body => annotationTyVars body
+  | .quant _ _ _ (some ty) tr body =>
+    LMonoTy.freeVars ty |>.foldl .insert (annotationTyVars tr |>.union (annotationTyVars body))
+  | .quant _ _ _ none tr body => (annotationTyVars tr).union (annotationTyVars body)
+  | .app _ e1 e2 => (annotationTyVars e1).union (annotationTyVars e2)
+  | .ite _ c t f => (annotationTyVars c).union ((annotationTyVars t).union (annotationTyVars f))
+  | .eq _ e1 e2 => (annotationTyVars e1).union (annotationTyVars e2)
+  | _ => {}
```
tautschnig: `annotationTyVars` ignores type annotations on `.fvar` and `.op`, even though both carry `Option LMonoTy`. The catch-all `_ => {}` lets a body like `.fvar _ x (some (.ftvar "a"))` (where `"a"` is captured from an outer scope, exactly the issue #1289 scenario but via fvar instead of abs) bypass the stray-var check in `Function.typeCheck`. Soundness happens to be saved downstream because `LMonoTy.instantiateWithCheck` re-freshens those vars, but the asymmetry with `.abs`/`.quant` is fragile and inconsistent with the docstring/intent.

Please either add the fvar/op cases or document explicitly why they're omitted. Suggested:

```suggestion
def annotationTyVars (e : LExpr ⟨⟨M, IDMeta⟩, LMonoTy⟩) : Std.HashSet TyIdentifier :=
  match e with
  | .fvar _ _ (some ty) => LMonoTy.freeVars ty |>.foldl .insert {}
  | .op _ _ (some ty)   => LMonoTy.freeVars ty |>.foldl .insert {}
  | .abs _ _ (some ty) body => LMonoTy.freeVars ty |>.foldl .insert (annotationTyVars body)
  | .abs _ _ none body => annotationTyVars body
  | .quant _ _ _ (some ty) tr body =>
    LMonoTy.freeVars ty |>.foldl .insert (annotationTyVars tr |>.union (annotationTyVars body))
  | .quant _ _ _ none tr body => (annotationTyVars tr).union (annotationTyVars body)
  | .app _ e1 e2 => (annotationTyVars e1).union (annotationTyVars e2)
  | .ite _ c t f => (annotationTyVars c).union ((annotationTyVars t).union (annotationTyVars f))
  | .eq _ e1 e2 => (annotationTyVars e1).union (annotationTyVars e2)
  | _ => {}
```

Also add a `#guard`/test that exercises a stray var via fvar/op annotation.

Strata/DL/Lambda/LTy.lean
```diff
@@ -373,6 +374,42 @@ theorem LMonoTys.freeVars_exists
...
+  goList (ts1 ts2 : List LMonoTy) (fwd : Std.HashMap TyIdentifier TyIdentifier)
+     (bwd : Std.HashMap TyIdentifier TyIdentifier) :
+     Option (Std.HashMap TyIdentifier TyIdentifier × Std.HashMap TyIdentifier TyIdentifier) :=
+    match ts1, ts2 with
+    | [], [] => some (fwd, bwd)
+    | t1 :: rest1, t2 :: rest2 =>
+      match go t1 t2 fwd bwd with
+      | some (fwd', bwd') => goList rest1 rest2 fwd' bwd'
+      | none => none
+    | _, _ => none
```
tautschnig, line 411: Two issues:

1. The function name `alphaEquiv` reads as a Boolean check, but it returns the witness map. Consider renaming to `alphaEquivMap` (and exposing a `Bool`-returning thin wrapper if useful), so call sites read more naturally.
2. The implementation creates `Std.HashMap`s in a tight recursive loop. Two arguments would be enough — keep `fwd` only and reject when an existing image collides — since `bwd` is purely defensive (under the invariant `fwd[a] = b ↔ bwd[b] = a`, the `bwd[y]? = some x` branch with `x' = x` is unreachable). Alternatively, since `LMonoTy.freeVars` is bounded, an `AssocList`/sorted list would suffice and avoid the hashing overhead.

This function is now on the soundness path. Please add a few `#guard`s for the obvious cases (reflexive, capture, distinct ftvars, mismatched constructors) and the proofs listed in the top-level review.

Strata/Languages/Core/FunctionType.lean
```diff
@@ -48,19 +53,35 @@ def typeCheck (C: Core.Expression.TyContext) (Env : Core.Expression.TyEnv) (func
...
+    let retty := func.output
     let S ← Constraints.unify [(retty, bodyty)] (TEnv.stateSubstInfo Env) |>.mapError format
+    -- The inferred type must be alpha-equivalent to the declared signature.
+    let inferredTy := LMonoTy.subst S.subst monoty
+    let bwdMap ← match LMonoTy.alphaEquiv monoty inferredTy with
+      | some m => pure m
+      | none =>
+        .error f!"Function '{func.name}': body constrains the type to '{inferredTy}', \
+                  incompatible with declared polymorphic signature '{monoty}'"
     let Env := TEnv.updateSubst Env S
```
tautschnig, line 91: Error message and the formatted result on success expose internal `$__tyN` names rather than the user's `origTypeArgs`. The issue #586 test prints `(arrow $__ty0 $__ty0)` instead of `(arrow a a)`, which is confusing in user-facing diagnostics.

You already have `origTypeArgs` and the bijection between fresh names and original ones from the instantiation step (and `bwdMap` here). Please format `inferredTy` and `monoty` using the user-facing names before producing the error, e.g. by mapping `$__tyi → origTypeArgs[i]` (or applying the inverse of the instantiation substitution). Update the corresponding `#guard_msgs` strings in `FunctionTypeTests.lean` accordingly.

Strata/Languages/Core/FunctionType.lean
```diff
@@ -48,19 +53,35 @@ def typeCheck (C: Core.Expression.TyContext) (Env : Core.Expression.TyEnv) (func
...
+      .error f!"Function '{func.name}': body contains undeclared type variables \
+                {strayVars.toList} (not in typeArgs {origTypeArgs})"
+    -- Add formals with monomorphic types (type parameters are fixed in the body).
     let Env := Env.pushEmptyContext
-    let Env := Env.addInNewestContext (LFunc.inputPolyTypes func)
-    -- Type check and annotate the body, and ensure that it unifies with the
-    -- return type.
+    let monoInputs : @LTySignature Unit :=
+      func.inputs.map (fun (id, mty) => (id, .forAll [] mty))
+    let Env := Env.addInNewestContext monoInputs
```
tautschnig: Worth promoting this idiom to a small helper alongside `LFunc.inputPolyTypes` in `Strata/DL/Lambda/Factory.lean` — e.g. `LFunc.inputMonoSignature : @LTySignature T.IDMeta`. It would document the contrast with `inputPolyTypes` (which is exactly the buggy form) and make the intent at this call site obvious.

While you're here: please also drop the leading `(FIXME) Very similar to Lambda.inferOp ...` comment or update it — the body now diverges meaningfully from `inferOp`.

Strata/Languages/Core/FunctionType.lean
```diff
@@ -48,19 +53,35 @@ def typeCheck (C: Core.Expression.TyContext) (Env : Core.Expression.TyEnv) (func
...
+                  incompatible with declared polymorphic signature '{monoty}'"
     let Env := TEnv.updateSubst Env S
-    -- Apply the outer unification substitution back to the body so that type
-    -- variables introduced inside it are resolved.
+    -- Apply S to the body, then rename type variables to match the
+    -- instantiated typeArgs so that body annotations are consistent.
     let bodya := LExpr.applySubstT bodya S.subst
+    let renameSubst : Subst := [bwdMap.toList.map (fun (k, v) => (k, .ftvar v))]
+    let bodya := LExpr.applySubstT bodya renameSubst
     -- Validate the measure expression type for int-recursive functions.
```
tautschnig, line 99: Two small things:

- Iteration order of `bwdMap.toList` is non-deterministic (HashMap). Correctness still holds because each key is unique and substitution is parallel, but it makes diagnostics non-reproducible. Sorting by key (or returning a sorted assoc list directly from `alphaEquiv`) is cheap and stabilises any future test/snapshot.
- Identity entries `(k, .ftvar k)` are no-ops; consider filtering them out to keep the substitution minimal.

StrataTest/Languages/Core/Tests/FunctionTypeTests.lean
```diff
@@ -0,0 +1,193 @@
...
+  } .empty
+]}
+
+/--
+info: error: Function 'foo': body constrains the type to '(arrow int int)', incompatible with declared polymorphic signature '(arrow $__ty0 $__ty0)'
+-/
+#guard_msgs in
+#eval (typeCheck .default issue586_pgm)
+
+end Core.FunctionTypeTests
```
tautschnig, line 335: Good regression coverage for the three issues. Please add the following missing cases:

1. A failing case using a quantifier annotation referencing an undeclared type var (exercises the `.quant _ _ _ (some ty) ...` branch of `annotationTyVars`).
2. A failing case using an fvar/op annotation referencing an undeclared type var (will surface the gap flagged on `LExpr.lean`).
3. A passing case where the body forces internal renaming so that `inferredTy ≠ monoty` syntactically but is alpha-equivalent (i.e., exercises the non-trivial `bwdMap` branch — the current passing tests all have identity bwd maps).
4. A passing case for a higher-order body, e.g. `apply<a,b>(f:a→b, x:a):b { f x }`, to lock in the multi-typeArg unification path.

Also, importing `Strata.Languages.Core.Verifier` for these small unit tests pulls in a heavy dependency. Importing only `Strata.Languages.Core.FunctionType` (plus `Program` for the issue #586 test) should be sufficient.

tautschnig (commented):
Proof coverage suggestions (none formally exist for this codepath beyond `typeCheck_inputs_nodup`):
- `LMonoTy.alphaEquiv_refl : alphaEquiv ty ty = some m` with `m` the identity on `freeVars ty`.
- `LMonoTy.alphaEquiv_sound`: when it returns `some bwd`, `LMonoTy.subst (bwd.toList ...) ty2 = ty1` (and `freeVars ty1 = freeVars ty2 |>.image bwd`).
- `LMonoTy.alphaEquiv_complete`: if `ty1` and `ty2` are alpha-equivalent (via any bijection), `alphaEquiv` returns some.
- `LExpr.annotationTyVars_correct`: `v ∈ annotationTyVars e ↔ v` appears in some `.abs`/`.quant` (and ideally `.fvar`/`.op`) type annotation in `e`.
- `Function.typeCheck_typeArgs_closed`: on success, every type variable appearing in `func'.body` annotations is in `func'.typeArgs`.
- `Function.typeCheck_polymorphism_preserved`: on success, the inferred type of the body is alpha-equivalent to `LFunc.type func'`.

Strata/DL/Lambda/LExpr.lean
shigoel: I'd steer away from the catch-all in pattern matching, as a rule.

StrataTest/Languages/Core/Tests/FunctionTypeTests.lean
shigoel, line 1: I think it's time to file an issue to create syntax (and also for Lambda using the DDM) so that these tests are easier to write and read.

Strata/DL/Lambda/LTy.lean
joscoh: I can change the name, but we need both maps. Otherwise we cannot enforce the bijection. Here is an example from Claude:
```
Concrete example:

  - ty1 = arrow (ftvar "a") (ftvar "b") — two distinct type vars
  - ty2 = arrow (ftvar "c") (ftvar "c") — one type var used twice

  With only fwd:
  1. ftvar "a" vs ftvar "c" — fwd["a"]? = none → insert fwd["a"] = "c" ✓
  2. ftvar "b" vs ftvar "c" — fwd["b"]? = none → insert fwd["b"] = "c" ✓

  Returns some — falsely claiming a → b is alpha-equivalent to c → c.
```

I don't think it is a good idea to use association lists instead. We want to get into the habit of using hash tables for sets/maps.

Also, this function will not be on the "soundness path"/TCB. I am in the process of writing a typing specification with against which we will prove the typechecker (including the alpha-equivalence check) sound. For that reason, I do not think the suggested proofs are immediately helpful. In a future PR when we prove the typechecker soundness we will need the more substantive proof (namely, that typing is invariant under alpha-equivalent types). The other properties are likely needed in that proof but are not immediately useful on their own at the moment.

aqjune-aws:
Instead of relying on alpha equivalence, what do you think of (1) instantiating the input parameters of func with type constructors, (2) calling LExpr.resolve, and (3) replacing the type constructors with type variables with same names?
I think this is clearer than checking alpha equivalence after type inference is finished, because it enforces that `LExpr.resolve` should not unify (previously) type variable `'a` with `int`. When it is given to `LExpr.resolve`, `'a` will have been converted into a type constructor `'a`, and two different type constructors cannot be unified. 

My Claude Code tried this and it also passed your tests. It introduces `LMonoTy.unrigidifyTyArgs` which converts type constructor with no argument into type variable.

https://github.com/strata-org/Strata/commit/6ac18083ce801a7bbe64377cada3a759c17d6b61

aqjune-aws:
Discussed separately and the current solution looks fine with me. Probably we will want to lift the `LExpr.checkContextTypesClosed Env →` assumption of `resolve_HasType` in `Strata/DL/Lambda/LExprTypeSpec.lean` to address this case, possibly by allowing that it is okay to have the Env have free variables with free type variables when substitution map is empty. A relevant proof (that appeared multiple times...) is https://gist.github.com/aqjune-aws/47713c51b5d43e121eb1651ba44c11fc.

StrataTest/Languages/Core/Tests/FunctionTypeTests.lean
joscoh: I can create an issue. Some of these are concrete syntax limitations (e.g. lambdas without type variables), but others illustrate bugs that are caught by the DDM typechecker (e.g. unbound type variables).

Strata/Languages/Core/FunctionType.lean
joscoh: 1. I don't think the overhead of sorting is worth it. As you note, substitution is parallel so the order doesn't matter.
2. Added filtering of identity entries.

Strata/DL/Lambda/LExpr.lean
joscoh: Update: this does not work, since operators may have free variables in their declaration that should get instantiated during typechecking (e.g. datatype destructors). I reverted this change

StrataTest/Languages/Core/Tests/FunctionTypeTests.lean
joscoh: https://github.com/strata-org/Strata/issues/1319

Strata/DL/Lambda/LExpr.lean
```diff
@@ -340,6 +340,26 @@ def collectFvarNames {T : LExprParamsT} : LExpr T → List (Identifier T.base.ID
   | .eq _ e1 e2 => collectFvarNames e1 ++ collectFvarNames e2
   | _ => []
 
+/-- Collects type variables from user-written binder annotations (abs/quant only).
+    fvar/op annotations are excluded because they come from the typing context
+    (e.g. datatype selectors) and will be instantiated during resolution. -/
+def annotationTyVars (e : LExpr ⟨⟨M, IDMeta⟩, LMonoTy⟩) : Std.HashSet TyIdentifier :=
```
aqjune-aws: I think the name of this function can be clearer by mentioning that it only collects from binders. What about `tyVarsOfBoundVars`?

But more importantly, why do we need to skip the type variables in `.fvar`/`.op`s? I kind of thought that since bodyVars.filter will remove those not included in `origTypeArgs`, it was safe to collect those from `.fvar`/`.op`s.
```
    let bodyVars := body.annotationTyVars
    let strayVars := bodyVars.filter (· ∉ origTypeArgs)
```

StrataTest/Languages/Core/Tests/FunctionTypeTests.lean
```diff
@@ -0,0 +1,335 @@
...
+  { name := ⟨"wrapId", ()⟩,
+    typeArgs := ["T"],
+    inputs := [(⟨"x", ()⟩, .ftvar "T")],
+    output := .ftvar "T",
+    body := some (.app () (.abs () "y" none (.bvar () 0)) (.fvar () ⟨"x", ()⟩ none)) }
+
+/--
+info: ok: typeArgs: [$__ty0]
+inputs: (x, $__ty0)
+output: $__ty0
```
aqjune-aws: I am wondering whether "T" could be kept instead of being replaced with "$__ty0" because from the pespective of user of `Function.typeCheck` it would be more natural - would using `Constraints.unify [(bodyty, retty)]` instead of `Constraints.unify [(retty, bodyty)]` work?

StrataTest/Languages/Core/Tests/FunctionTypeTests.lean
```diff
@@ -0,0 +1,335 @@
...
+/--
+info: [Strata.Core] Type checking succeeded.
+
+---
+info: ok: program Core;
+
+function fst<$__ty0, $__ty1> (x : $__ty0, y : $__ty1) : $__ty0 {
+  x
+}
+-/
```
aqjune-aws, line 70: Let's print the type of the body in this test case and below case as well, because it gives more information. The first case in this file already seems to achieve this.

Strata/DL/Lambda/LExpr.lean
joscoh: Sure, I can change the name.

The filter is not enough because the annotations on these nodes come from external declarations whose type variables are unrelated to the current function's typeArgs. Concrete example:
```
function g<U>(x: U): U { x }
function f<T>(y: T): T { g(y) }
```
After translation, `f`'s body contains `.op () "g" (some (arrow (ftvar "U") (ftvar "U")))`. If we collected `"U"` from that annotation, the filter would flag it, since "U" is not "T", and then `typeCheck` would incorrectly reject. The same applies to datatype destructors (e.g., `List.head` contains type `(arrow (list T) T)` where `T` is from the datatype declaration). These annotations are resolved later by LExpr.resolve, so we only check user-written binder annotations, which are the only place the user can introduce a stray type variable.

aqjune-aws: I see. Then why should the `fvar` case be omitted as well? It seems the function call cases are corresponding to `.op`.

StrataTest/Languages/Core/Tests/FunctionTypeTests.lean
joscoh: The difference is that these tests with concrete syntax are using the Program-wide type checking, which does not have that info. The ones that use more fine-grained info (e.g. test 1) use abstract syntax to construct a Function and then use Function.typeCheck directly. We could make more tests into abstract syntax to get this info, but this conflicts with Shilpi's comment above. I am fine with either but we should all agree.

Strata/DL/Lambda/LExpr.lean
joscoh: For example, consider the following:
```
procedure p<V>(x: V) (z: V) {
   function g<T>(y: T): T {
      if x == z then y else y
    }
   g(0);
  }
```
Here, `x` and `z` have type variables that are fixed from an outer context and not present in the inline function statement signature, but this is still a well-typed function/procedure.

shigoel: I'd vote for having the explanation in your comments in the code, Josh, so that we can recreate context later.

StrataTest/Languages/Core/Tests/FunctionTypeTests.lean
shigoel: I'm okay to have more tests with abstract syntax, as long as when we address https://github.com/strata-org/Strata/issues/1319, we fix them too. Will leave it up to you to determine whether adding more tests with abstract syntax is worth it right now or not.

shigoel (dismissed):
LGTM! Left a minor comment about including examples in the doc string, but that's not blocking.

aqjune-aws (approved):
LGTM :)

shigoel (approved):

PR #1290, opened 2026-05-28
Add README.md and AGENTS.md for StrataBoole

atomb:
Also adjust `lakefile.toml` so that `lake test` works as documented.


shigoel:
@kondylidou @abdoo8080 @ChengZ3 Please have a look!

StrataBoole/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+Programs can be embedded in Lean using the `#strata ... #end` syntax:
+
+```lean
+import StrataBoole.MetaVerifier
+
+open Strata
+
+def myProgram : Strata.Program :=
+#strata
```
shigoel, line 65: The type annotation `Strata.Program` doesn't exist after the DDM namespace rename (#1272). Should be `StrataDDM.Program` to match the actual codebase — existing tests all use `def loopSimple : StrataDDM.Program :=`.

shigoel (commented):
Documentation looks good overall. One type annotation issue in the README example.

StrataBoole/AGENTS.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+## Dependencies
+
+- **Strata** (parent package, via `path = ".."`) — provides Core dialect, VCG, SMT encoding, DDM infrastructure
+- **Lean 4** — version specified in `lean-toolchain`
+
+## Pitfalls
+
+- **Do not use Core's `out`/`inout` call argument syntax in Boole programs.** Boole enforces `call lhs := f(args)` for outputs. The translator will reject `out`/`inout` arguments with a clear error.
```
abdoo8080: @kondylidou, is there a reason for not supporting `out` and `inout` arguments?

kondylidou: Yes — `out` and `inout` don't fit Boole's calling convention. Boole cleanly separates
inputs (expressions in the arg list) from outputs (`returns` declarations, received via
`call lhs := f(args)`). The Core concept of a single parameter that is both input and
output (`inout`) has no equivalent in that model.

- **`out`**: `call lhs := f(args)` is Boole's call syntax. The translator converts the LHS
  directly into `Core.CallArg.outArg` entries. An explicit `out` argument would duplicate
  what the translator already inserts.

- **`inout`**: Boole has global variables (`command_var`) and each call site gets a prefix
  of global arguments automatically via `constructProcArgsPrefix`. Modified globals become
  `inoutArg`, read-only ones become `inArg`, based on the callee's `modifies` declarations.
  The user never sees or writes this; the translator handles it. Allowing explicit `inout`
  at the call site would conflict with that automatic prefix.

In short: both `out` and `inout` exist at the Core level, but Boole generates them
implicitly — from the call LHS and from `modifies` clauses respectively — so exposing them
in Boole syntax would duplicate or conflict with what the translator already inserts.
This was a deliberate design choice.

kondylidou:
One inaccuracy in the docs:

`gen_smt_vcs_boole` does not exist. The tactic is `gen_smt_vcs`, defined in
`MetaVerifier.lean` and used in every existing Boole test (`insertion_sort.lean`,
`datatype_selector_vcs.lean`, etc.). Any contributor following the docs will get an
"unknown tactic" error.

StrataBoole/README.md
kondylidou: So once #1272 (DDM namespace rename) lands, this will need to be updated to `StrataDDM.Program`. Worth tracking or updating now.

StrataBoole/AGENTS.md
tautschnig: Please fold @kondylidou's PR-thread explanation into this bullet so future readers don't have to dig through review history. Roughly:

> Boole's calling convention separates inputs (positional args) from outputs (`returns` declarations consumed by `call lhs := f(args)`). `out` is unnecessary because the translator already emits `Core.CallArg.outArg` from the call's LHS, and `inout` would conflict with the implicit global-variable prefix added by `constructProcArgsPrefix` based on `modifies` clauses. So both exist at the Core level but are generated implicitly from Boole syntax.

StrataBoole/README.md
```diff
@@ -0,0 +1,149 @@
...
+  ensures (r == x + y);
+}
+{
+  r := x + y;
+};
+#end
+```
+
+Programs can also be stored in `*.boole.st` source files. However, there is not
+currently a top-level CLI that will process these files. TODO: describe how
```
tautschnig: Please don't ship a `TODO` in user-facing docs. Either flesh out the loading instructions now (the relevant entry points are `Strata.Boole.verify` / `parseStrataProgramFromDialect`-based pipelines used in tests like `procedure_signatures.lean` and `format_program.lean`) or replace this paragraph with a single-line link to a tracking issue (`See #NNNN for library-mode loading.`).

StrataBoole/README.md
```diff
@@ -0,0 +1,149 @@
...
+3. **VCG**: Core's verifier generates verification conditions via symbolic execution
+4. **Encode**: VCs are encoded as SMT-LIB formulas
+5. **Solve**: Either dispatched to an external solver or proved within Lean
+
+### Boole vs. Core Syntax
+
+| Boole | Core equivalent |
+|-------|----------------|
+| `var x : int;` | inout parameter |
+| `modifies x;` | inout parameter |
```
tautschnig: These two rows are imprecise — `var` and `modifies` don't have a single Core counterpart, the translator *adds* parameters to procedures based on them. Suggest:

```suggestion
| `var x : int;` (global) | adds `x` as `inout` parameter to procedures whose `modifies` lists it; `in` parameter elsewhere |
| `modifies x;` (clause) | promotes the corresponding global to `inout` in this procedure's signature |
```

The other rows are already accurate.

tautschnig (commented):
### Note: this PR targets _main_

**About the `gen_smt_vcs_boole` PR comment from @kondylidou:** I checked `StrataBoole/StrataBoole/MetaVerifier.lean:106` — the tactic is in fact named `gen_smt_vcs_boole`, defined via `syntax (name := genSMTVCsBoole) "gen_smt_vcs_boole" : tactic` and used in *every* Boole test (including `insertion_sort.lean:57` and `demo.lean:40`). The doc usage is correct; no change needed there. (`gen_smt_vcs` without the suffix is the *non-Boole* variant in `Strata/MetaVerifier.lean:353`.)

kondylidou:
> ### Note: this PR targets _main_
> **About the `gen_smt_vcs_boole` PR comment from @kondylidou:** I checked `StrataBoole/StrataBoole/MetaVerifier.lean:106` — the tactic is in fact named `gen_smt_vcs_boole`, defined via `syntax (name := genSMTVCsBoole) "gen_smt_vcs_boole" : tactic` and used in _every_ Boole test (including `insertion_sort.lean:57` and `demo.lean:40`). The doc usage is correct; no change needed there. (`gen_smt_vcs` without the suffix is the _non-Boole_ variant in `Strata/MetaVerifier.lean:353`.)

oh yes you're right! sorry I was too fast commenting :p

shigoel:
> Note: this PR targets main

Confirming that this is intentional, in service of https://github.com/strata-org/Strata/issues/1168

StrataBoole/README.md
atomb: Good catch!

atomb:
> > ### Note: this PR targets _main_
> > **About the `gen_smt_vcs_boole` PR comment from @kondylidou:** I checked `StrataBoole/StrataBoole/MetaVerifier.lean:106` — the tactic is in fact named `gen_smt_vcs_boole`, defined via `syntax (name := genSMTVCsBoole) "gen_smt_vcs_boole" : tactic` and used in _every_ Boole test (including `insertion_sort.lean:57` and `demo.lean:40`). The doc usage is correct; no change needed there. (`gen_smt_vcs` without the suffix is the _non-Boole_ variant in `Strata/MetaVerifier.lean:353`.)
> 
> oh yes you're right! sorry I was too fast commenting :p

One thing that I'd like to confirm, though: are you on board with that new tactic with a new name? It was an agent's choice to create a new tactic in the `StrataBoole` package so that the one in the main `Strata` package doesn't depend on anything Boole-specific. However, there may be a way to have a single tactic that packages can extend.

kondylidou:
> > > ### Note: this PR targets _main_
> > > **About the `gen_smt_vcs_boole` PR comment from @kondylidou:** I checked `StrataBoole/StrataBoole/MetaVerifier.lean:106` — the tactic is in fact named `gen_smt_vcs_boole`, defined via `syntax (name := genSMTVCsBoole) "gen_smt_vcs_boole" : tactic` and used in _every_ Boole test (including `insertion_sort.lean:57` and `demo.lean:40`). The doc usage is correct; no change needed there. (`gen_smt_vcs` without the suffix is the _non-Boole_ variant in `Strata/MetaVerifier.lean:353`.)
> > 
> > 
> > oh yes you're right! sorry I was too fast commenting :p
> 
> One thing that I'd like to confirm, though: are you on board with that new tactic with a new name? It was an agent's choice to create a new tactic in the `StrataBoole` package so that the one in the main `Strata` package doesn't depend on anything Boole-specific. However, there may be a way to have a single tactic that packages can extend.

yes I believe this is ok if @abdoo8080 agrees too

shigoel (approved):

StrataBoole/README.md
```diff
@@ -0,0 +1,152 @@
...
+lake build
+```
+
+To run all tests (builds the test library, which includes `#eval` and `#guard_msgs` checks):
+
+```bash
+lake test
+```
+
+The package depends on the parent `Strata` package (via `path = ".."`), so ensure the main Strata package builds first.
```
joscoh, line 53: I assume this line will change when Boole becomes a separate repo?

joscoh (approved):

StrataBoole/README.md
atomb: Yes! Thanks for helping make sure I remember. :)

PR #1293, opened 2026-05-29
Python FE (Re)architecture

ssomayyajula:
TODO


keyboardDrummer:
If I understand correctly, this PR doesn't run existing Laurel tests using the new elaborator. Is that right? @ssomayyajula  Glancing over Elaborate.lean, it seems to have only part of the feature set of the existing Laurel -> Core pipeline. I guess it supports whatever Python needs.

If you want to see how well it supports the upcoming Laurel features, you should rebase this onto https://github.com/keyboardDrummer/Strata/pull/28, and then hookup all the tests to use the compilation pipeline this PR introduces.

A second thing, why does this PR modify Python->Laurel and Laurel->Core pipelines in one PR? Can't we scope it down to either of the two?

PR #1296, opened 2026-05-29
Change "unreachable" assertion message to "unreachable in this context"

### LINKED ISSUE #1295: Change `"unreachable"` assertion message to `"unreachable in this context"`
The `"unreachable"` outcome label is misleading — it implies the assertion is unreachable from the procedure entry, which is not what is actually proved. Consider:

```strata
axiom false
procedure f(i: int) {
  if i % 2 == 0 {
    assert true;
  }
}
```

Here, if we reached the procedure entry, the assertion is clearly reachable. What Strata actually proves is that the conjunction of axioms and path conditions is unsatisfiable — not that no execution path from declaration entry reaches the assertion.

**Discussion highlights**

* Wordings considered: `"in contradictory path"`, `"unreachable if axioms are satisfiable"`, `"unreachable in the context of global axioms"`.
* `"unreachable in the context of global axioms"` was noted as ambiguous and doesn't generalize to nested/inlined procedures, where the context includes local path conditions too.
* `"unreachable in this context"` was preferred as concise, accurate, and uniformly applicable regardless of procedure nesting.
* No change to proof obligations or SMT encoding — wording only.

**Proposed change**

Replace the `"unreachable"` outcome label with `"unreachable in this context"` in all user-facing output. Update tests accordingly.

MikaelMayer:
Fixes #1295

Replaces the "path unreachable" outcome label with "unreachable in this context" in all user-facing verification output (outcome labels, diagnostic messages, and SARIF messages). The SARIF message no longer includes "path condition is contradictory" since unreachability can stem from reasons beyond the path condition. This is a wording-only change with no impact on proof obligations or SMT encoding.

Tested: all existing tests pass with updated expected output.


Strata/Languages/Core/SarifOutput.lean
```diff
@@ -60,7 +60,7 @@ def outcomeToMessage (outcome : VCOutcome) : String :=
       else if !m2.isEmpty then s!" (false: {Std.format m2})"
       else ""
     s!"True or false depending on inputs{models}"
-  | .unsat, .unsat => "Unreachable: path condition is contradictory"
+  | .unsat, .unsat => "Unreachable in this context: path condition is contradictory"
```
atomb: One of the issues this PR is trying to address is that something may be unreachable for reasons other than the path condition.
```suggestion
  | .unsat, .unsat => "Unreachable in this context"
```

atomb (changes requested):
One tiny wording change suggestion.

Strata/Languages/Core/SarifOutput.lean
MikaelMayer: 🤖 Done in d5babc5c1 — simplified to just "Unreachable in this context" since unreachability can stem from reasons beyond the path condition.

MikaelMayer (commented):
🤖🔍 Clean wording-only change, applied consistently across all user-facing output. No old "path unreachable" or "path condition is contradictory" strings remain in Lean source. Capitalization is correct: SARIF messages use sentence-case ("Unreachable in this context") matching the convention of other SARIF messages, while labels use lowercase (mid-sentence context).

**Minor gap:** `docs/VerificationModes.md` (line 40) still says "The path condition is contradictory" — the exact phrasing this PR removes from code because unreachability can stem from reasons beyond the path condition. Consider updating it to match the new wording, e.g.:

> 3. **Unreachable** (unsat, unsat) — The code is unreachable in this context, indicating dead code. This is flagged as an error because unreachable code often signals a specification or implementation problem that should be investigated.

atomb (approved):

aqjune-aws (approved):

Issue #1067, opened 2026-04-28
Factor out proofs in semantics related files as `*Prop.lean`


aqjune-aws:
There are many files that define semantics and also has theorems proven about these. This is good for keeping small definitions private, but on the other hand it hides the key definitions under a pile of theorems.

Let's separate the theorems to `*Prop.lean`. The first target would be Strata/DL/Imperative/StmtSemantics.lean.

aqjune-aws:
#1327

PR #1306, opened 2026-06-01
CallElimCorrect: rewrite using small-step semantics

PROgram52bc:
## Summary

This PR **replaces the deprecated big-step `CallElimCorrect` proof with a small-step version**, plus reusable infrastructure (`SubstProps`, `CoreTransformProps`) and shared-file helpers. The previous file carried a `DEPRECATED` notice flagging it for a small-step rewrite; this is that rewrite.

Top-level public theorem `callElimStatementCorrect` is preserved (now a conjunction over a `_terminal` arm and an `_exiting` arm), the contract semantics for procedure calls (`EvalCommandContract.call_sem`) is unified into a single failure-flag–parameterized rule, and the `_terminal` arm is fully polymorphic over the failure flag (`{f : Bool}`) — a source success trace maps to a target success trace and a source failure trace maps to a target failure trace, in one statement.

Net repository delta vs `main2`: +9,277 / -4,861 LoC across 19 files (the bulk being reusable infrastructure rather than CallElim-specific). Compared to the original 14,972-LoC big-step proof, this represents a ~-8k LoC reduction overall, while strictly strengthening what is proved.

## What's added

The 19 files modified group into 4 layers:

### Top-level proof (~10,525 LoC churn)

- `Strata/Transform/CallElimCorrect.lean`: full rewrite against small-step semantics. Same public theorem name `callElimStatementCorrect`; structured as `_terminal ∧ ∀ lbl, _exiting`, mirroring `Specification.Overapproximates` and `TripleBlock`. The `_terminal` arm is polymorphic in the failure flag `{f : Bool}`; the `_exiting` arm reuses the original `Heval` for non-call cases and is vacuously discharged for the call case (a singleton call statement only ever reaches `.terminal`, never `.exiting`).

### Reusable infrastructure (~2,747 LoC, new files)

- `Strata/Transform/SubstProps.lean` (1,621 LoC, new): generic substitution-semantics infrastructure used by CallElim and reusable by future transforms. Includes `H_asserts_zip` / `H_assumes_zip` / `H_check_block_zip` walkers, `singletonAssertEval` / `singletonAssumeEval` and their `_poly` failure-flag variants, plus `H_asserts_zip_fail` (the L4 flag-flip walker that powers the polymorphic-f failure arm).
- `Strata/Transform/CoreTransformProps.lean` (1,126 LoC, new): reusable helpers for the Core small-step layer — `singleCmdToStmts` / `_poly`, `H_havocs` / `_poly`, `H_init`, `H_inits`, `genIdent*` correctness, fresh-name predicates, `EvalCallElim_glue` and `EvalCallElim_glue_fail` (composing the L1–L6 segments with the L4 success/failure flip).

### Core semantics adjustments (~30 LoC)

- `Strata/Languages/Core/StatementSemantics.lean`: `EvalCommandContract.call_sem` unified — previously two separate path shapes (success: havoc + write-back; failure: `σ' = σ` and the precondition iff in the wrong direction); now a single rule parameterized by a Bool failure flag indicator, with the precondition iff narrowed to non-`Free` preconditions to match `callElimCmd`'s actual emission rule. `EvalCommand.call_sem` propagates the callee body's terminal `hasFailure` flag to the caller (previously hardcoded to `false`, silently dropping any failures inside callee bodies).
- `Strata/Languages/Core/StatementSemanticsProps.lean` (+297 LoC): 8 new helpers around `updatedStates` / `readValues` / `evalExpressions` / `initStates`, plus the revived 5-theorem refinement-transit chain (`EvalCommandRefinesContract`, `StepStmt[Star]_refines_contract`, `EvalStatement[s]RefinesContract`) over the unified `call_sem` shape. `EvalCallBodyRefinesContract` is closed (was a long-standing `sorry` inside a comment block).
- `Strata/Languages/Core/WF.lean`: `WFCallSiteProp.preBoolTyped` and `WFPrePostProp.boolTyped` per-precondition Bool-totality clauses, used by the failure-arm to derive a failing-pre witness.
- `Strata/Transform/ProcBodyVerifyCorrect.lean`: takes `CalleesNoFailure π φ` as a new hypothesis (every procedure body, run from a non-failing init env, terminates with `hasFailure = false`), threaded through `core_noFailure_preserved`. The hypothesis is exactly the body-noFailure clause that clause 2 of the theorem's own conclusion (`ProcedureCorrect`) establishes per procedure, so callers discharge it inductively.

### Shared-file helpers (~250 LoC)

- `Strata/Languages/Core/StatementSemanticsProps.lean` (above): 8 new shared helpers consumed by both CallElim and its sibling Core proofs.
- `Strata/DL/Util/ListUtils.lean` (+150 LoC net): adds `notin_append4`, `unzip_snd_length`, `disjoint_of_nodup_append_three`, `zip_pair_split`, `nodup_3_decompose`, `notin_3_append_of`, `zip_zip_unzip_{snd,fst_unzip_fst}_of_lengths`, `zip_notin_{fst,snd}_pair`; promotes existing helpers to `public`.
- `Strata/DL/Imperative/CmdSemanticsProps.lean` (new, +33 LoC): generalized pure-Imperative lemmas `subst_defined_tail`, `subst_nodup_tail` (P : PureExpr).
- `@[expose]` attributes added where required by the new module-system reductions:
  - `Strata/DL/Imperative/CmdSemantics.lean`: `isNotDefined`, `substStores`, `substDefined`, `substNodup`, `invStores`, `substSwap`, `WellFormedSemanticEvalVal`.
  - `Strata/DL/Lambda/LExprWF.lean`: `substFvar`, `substFvars`.
  - `Strata/DL/Util/StringGen.lean`: `StringGenState.gen`.
  - `Strata/Languages/Core/CoreGen.lean`: `CoreGenState.WF`, `CoreGenState.gen`.
  - `Strata/Languages/Core/Procedure.lean`: `Procedure.Spec.getCheckExprs`, `Procedure.Spec.updateCheckExprs`.
  - `Strata/Languages/Core/StatementSemantics.lean`: `updatedState`, `updatedStates'`, `updatedStates`, `WellFormedCoreEvalTwoState`.
  - `Strata/Transform/CoreTransform.lean`: ~17 generator/transform defs.
  - `Strata/Transform/CallElim.lean`: `callElimCmd`, plus an extracted `oldTyLookupCallElim` helper required for the proof to share auxiliary `match` definitions across two `do`-blocks.

## Notable commits

The branch is structured so each landed commit can be reviewed independently:

- `286c6ce36` — initial small-step rewrite (preserves the original `_terminal`-only theorem at `f = false`).
- `3710bb339` — add the exit-arm to `callElimStatementCorrect` (review feedback from @aqjune-aws): mirrors `Specification.Overapproximates` shape; non-call cases reuse the original `Heval`, the call case is vacuously discharged via `seq_reaches_exiting`.
- `0eb0342ee` — review-feedback rename `*Semantics → *SemanticsProps` and relocate generic lemmas to `ListUtils` / `CmdSemanticsProps` / `SubstSemanticsProps`.
- `4bfccc31b` — integrate `hasFailure` flag into `EvalCommand.call_sem` (closes a soundness gap inherited from the original semantics).
- `e679bd560` — second review-feedback rename `*SemanticsProps → *Props` (the `Semantics` qualifier was misleading: these files hold transform machinery and substitution lemmas, not semantics defs).
- `44deb44b2` — close `EvalCallBodyRefinesContract` and revive the dormant Core refinement bridge.
- `ba122da08` — unify `EvalCommandContract.call_sem` (failure-flag-parameterized) and revive the 5-theorem refinement-transit chain.
- `3daec283b` — cleanup pass: 6 unused theorems in `CoreTransformProps`, 5-helper dead chain in `SubstProps`, dangling docstring references, TODO disposition. Net -327 LoC.
- `be987fddc` — lift `_terminal` to polymorphic `{f : Bool}` (PR #1340, merged into this branch).

## Build status

- `lake build`: clean (490 jobs)
- 0 sorries introduced (baseline `sorry` inventory in `ProgramWF` / `ProcedureWF` / `StatementWF` / `CmdEval` is unchanged; this PR closes one previously commented-out `sorry` in `EvalCallBodyRefinesContract`)
- 0 axioms introduced
- `#print axioms callElimStatementCorrect` → `[propext, Classical.choice, Quot.sound]` (Lean stdlib only)

## Test plan

- [x] `lake build` clean across all 490 jobs
- [x] `#print axioms callElimStatementCorrect` reports stdlib axioms only
- [x] No new sorries / admits / axioms introduced
- [x] Conjunctive shape of `callElimStatementCorrect` matches `Specification.Overapproximates` / `TripleBlock`
- [x] `_terminal` arm polymorphic in `{f : Bool}` (success and failure traces both covered)
- [x] Review-feedback follow-ups landed (exit arm, file renames, generic-lemma relocation)
- [ ] CI passes

- Strata/DDM/*
- Strata/Languages/Boole/*
- Tools/BoogieToStrata


Strata/Transform/CoreTransformProps.lean
```diff
@@ -0,0 +1,1151 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
```
aqjune-aws, line 5: The file name contains `Semantics`, however the file and `SubstSemantics.lean` don't seem to include formal definition of any style of semantics (denotational semantics, operational semantics, etc). I think we must adhere to the '*Prop.lean' suffix style for files that includes a collection of lemmas/theorems.

Strata/Transform/SubstProps.lean
```diff
@@ -0,0 +1,1721 @@
+/-
```
aqjune-aws, line 1: Can this file be moved to Strata/Languages/Core?

Strata/Transform/SubstSemantics.lean
```diff
@@ -0,0 +1,1721 @@
...
+    (σ x).isSome = true → ((updatedState σ k v) x).isSome = true := by
+  intros Hsome
+  simp [updatedState]
+  split <;> simp_all
+
+theorem subst_defined_tail
+    {σ σ' : CoreStore} {h : Expression.Ident × Expression.Ident}
+    {t : List (Expression.Ident × Expression.Ident)} :
+    Imperative.substDefined σ σ' (h :: t) →
+    Imperative.substDefined σ σ' t := by
```
aqjune-aws: This lemma doesn't seem to be specific to Core. Can we have a separate file that has a collection of lemmas which can be applied to Imperative? Let's check and apply similar migrations to other lemmas in this file if possible.

Strata/Transform/SubstSemantics.lean
```diff
@@ -0,0 +1,1721 @@
...
+  intros Hsubst k1 k2 Hin
+  refine ⟨?_, (Hsubst k1 k2 Hin).2⟩
+  exact updatedStateIsDefinedMono' (Hsubst k1 k2 Hin).1
+
+theorem zip_notin_fst_pair
+    {h : Expression.Ident}
+    {t : List Expression.Ident} {t' : List Expression.Ident} :
+    t.length = t'.length →
+    (∀ x, ¬(h, x) ∈ List.zip t t') →
+    ¬ h ∈ t := by
```
aqjune-aws: It seems this lemma can be moved to ListUtils.lean. Let's move general lemmas in this file (and CoreTransformationSemantics.lean) to there.

Strata/Transform/CallElimCorrect.lean
```diff
@@ -39,4558 +37,3777 @@ public section
...
+  exact ⟨(List.mem_append.mp Hmem.1).elim (Hσ_a _) (Hσ_b _),
+         (List.mem_append.mp Hmem.2).elim (Hσ'_a _) (Hσ'_b _)⟩
+
+/-- Decompose `(ks.zip ks').get n = (k1, k2)` into per-component equalities,
+    given explicit bounds for each list. -/
+private theorem zip_pair_split {α β} {ks : List α} {ks' : List β}
+    {n : Fin (ks.zip ks').length} {k1 : α} {k2 : β}
+    (hn : n.val < ks.length) (hn' : n.val < ks'.length)
+    (heq : (ks.zip ks').get n = (k1, k2)) :
+    k1 = ks[n.val]'hn ∧ k2 = ks'[n.val]'hn' := by
```
aqjune-aws: This lemma is general; let's move to ListUtils. Similarly to `nodup_3_decompose`; let's ask an agent to do search other possible candidates & do migration.

Strata/Transform/CallElimCorrect.lean
```diff
@@ -39,4558 +37,3777 @@ public section
...
+    (Hsome : (σ k).isSome) (Hndef : Imperative.isNotDefined σ ks) : k ∉ ks :=
+  fun h => σ_some_contradiction Hsome (Hndef k h)
+
+/-- Build `substDefined σ σ' ((a₁ ++ b₁).zip (a₂ ++ b₂))` from per-half
+    `isDefined` facts. -/
+private theorem substDefined_of_app
+    {σ σ' : CoreStore} {a₁ b₁ a₂ b₂ : List Expression.Ident}
+    (Hσ_a : Imperative.isDefined σ a₁) (Hσ_b : Imperative.isDefined σ b₁)
+    (Hσ'_a : Imperative.isDefined σ' a₂) (Hσ'_b : Imperative.isDefined σ' b₂) :
+    Imperative.substDefined σ σ' ((a₁ ++ b₁).zip (a₂ ++ b₂)) := by
```
aqjune-aws: Should this be moved to `SubstSemantics.lean`?

aqjune-aws (commented):
I left a few comments which can be applied orthogonally to our ongoing discussion in a private channel.
One extra question is that, I think the correctness of the top-level statement must also consider the case when a program may 'throw', through `Imperative.exit`. For example,
```
call f();
exit "some-block";
```
Note that the original program exits to "some-block", hence the final state after small-steps isn't ".terminal" (unlike what `EvalStmtsSmall` states, which is used transitively by `EvalStatementsContract`). If we apply call-elimination on this,
```
(omit temporary vars)
assume <pre>;
assert <post>;
exit "some-block";
```
it will have to exit to the right state.

Strata/Transform/CoreTransformProps.lean
aqjune-aws: #1327 is doing the similar refactoring, in order to resolve #1067.

PROgram52bc:
Fixed in 3710bb339. The theorem now has a conjunctive shape, refactored into three theorems: `_terminal` (private, unchanged), `_exit` (private, new for the call case), and a public conjunction. No external callers needed updating.

Strata/Transform/CoreTransformProps.lean
PROgram52bc: Good point. I renamed to `CoreTransformSemanticsProps.lean` (in 0eb0342ee), matching the existing `*SemanticsProps.lean` convention used by `Strata/Languages/Core/StatementSemanticsProps.lean` — which is the same convention #1327 is adopting (`*Props.lean`, plural).

One coordination note: 0eb0342ee also adds `Strata/DL/Imperative/CmdSemanticsProps.lean` for two Imperative-level lemmas migrated from `SubstSemantics.lean`. #1327 also creates that exact file path. Happy to rebase onto #1327 once it merges into main.

Strata/Transform/SubstProps.lean
PROgram52bc: It appears that the file needs `Core.Transform.createFvar(s)` (in many lemma signatures) and `singletonAssertEval` / `singletonAssumeEval` from `CoreTransformSemanticsProps`, so I left it under `Strata/Transform/`. Renamed to `SubstSemanticsProps.lean` to match the `*SemanticsProps.lean` convention. Perhaps a separate refactor PR can take place to move those to `Core` if needed.

Strata/Transform/SubstSemantics.lean
PROgram52bc: Agreed. Moved to a new `Strata/DL/Imperative/CmdSemanticsProps.lean` and generalized to `P : PureExpr` (Done in 0eb0342ee). Now lives as `Imperative.subst_defined_tail`. `subst_nodup_tail` is also migrated.

Strata/Transform/SubstSemantics.lean
PROgram52bc: Agreed. Done in 0eb0342ee along with a sweep across the three files. `zip_notin_fst_pair` and its sibling `zip_notin_snd_pair` are now in `Strata/DL/Util/ListUtils.lean` under the `List` namespace, generalized to arbitrary `α β`.

Strata/Transform/CallElimCorrect.lean
PROgram52bc: Done in 0eb0342ee. Swept all three files (`CallElimCorrect`, `SubstSemantics`, `CoreTransformSemantics`); seven pure-list lemmas now live in `Strata/DL/Util/ListUtils.lean` under the `List` namespace:

- `zip_pair_split`
- `nodup_3_decompose`
- `notin_3_append_of`
- `zip_zip_unzip_snd_of_lengths`
- `zip_zip_unzip_fst_unzip_fst_of_lengths`
- `zip_notin_fst_pair` / `zip_notin_snd_pair`

Two pure-Imperative lemmas (`subst_defined_tail`, `subst_nodup_tail`) went to the new `Strata/DL/Imperative/CmdSemanticsProps.lean`.

Strata/Transform/CallElimCorrect.lean
PROgram52bc: Yes. Moved to `SubstSemanticsProps.lean` (renamed from `SubstSemantics.lean`) in 0eb0342ee. Now lives in the `Core.Transform` namespace and resolves unqualified at the call sites via the existing `open Core.Transform`.

Strata/Transform/CoreTransformProps.lean
aqjune-aws: Umm, I think the name has to be `CoreTransformProps.lean`, because as I said above there is no _formal semantics_ defined in `CoreTransform.lean` and *Props.lean is the suffix, not *SemanticsProps.lean. Could you fix `SubstSemanticsProps.lean` accordingly as well?

aqjune-aws:
Thanks for adding the exiting cases. It is good to see them proven :)

I have two high-level important comments which will more be related to what is to be proven.

1. The `call_sem` of `EvalCommandContract`

You would have already figured out the issue in `call_sem`. To talk about its proper fix, let me introduce the difference between the small-step semantics of assume and assert first:
* For assume, if the program state doesn't satisfy its condition, it is simply stuck; it doesn't proceed.
* For assert, even if the program state doesn't satisfy the condition, it *proceeds*; it only updates the hasFailure flag.
You can confirm this from `EvalCmd` which has `eval_assert_pass`, `eval_assert_fail` and `eval_assume`. This is designed in this way so that even if a program includes multiple `assert`s and the first `assert` fails, we can still talk about the second `assert` individually (without using `hasFailure`, because `hasFailure` will already have been marked as true in this case).

Now coming to `call_sem`... When we are calling a procedure and the procedure has pre-post condition, from the caller's side I think a natural flow would be (1) we assert the precondition, (2) execute the body, and (3) assume the postcondition. It means that actually this part is too strong:

```
    (∀ pre, (Procedure.Spec.getCheckExprs p.spec.preconditions).contains pre →
      isDefinedOver (HasVarsPure.getVars) σAO pre ∧
      δ σAO pre = .some HasBool.tt) →
```

This has to be something like this:
```
    preconditionPassed <-> (∀ pre, (Procedure.Spec.getCheckExprs p.spec.preconditions).contains pre →
      isDefinedOver (HasVarsPure.getVars) σAO pre ∧
      δ σAO pre = .some HasBool.tt) →
...
    ----
    EvalCommandContract π δ σ (.call n callArgs md) σ' (\neg preconditionPassed)
```


2. Connection between `EvalCommand` and `EvalCommandContract`

I thought `EvalCommand` should be the ground-truth operational semantics for procedure call in a closed program, because when we do a procedure call, its interpreter will actually get into the body of the procedure call and concretely execute stuffs. How are `EvalCommand` and `EvalCommandContract`s related? Is there a theorem stating equivalence or refinement relation betweeen these two?

Strata/Transform/CoreTransformProps.lean
PROgram52bc: Ah, I see. Will do that. Thanks for pointing it out!

PROgram52bc:
> The `call_sem` of `EvalCommandContract`

Thanks for pointing it out. That makes sense. I will weaken the precondition requirement in `call_sem`, and adapt the proof accordingly.

>  Connection between `EvalCommand` and `EvalCommandContract`

[Here](https://github.com/strata-org/Strata/blob/4bfccc31bf4695f0c6178963d1a4cdb8cb3d3867/Strata/Languages/Core/StatementSemanticsProps.lean#L2221-L2229) is a potential theorem we could prove to connect the two. It is currently commented out because the contract version doesn't create the extra variables in the store, which makes the proof a little more complicated than it needs to be. But thanks for the question -- I will try to look into that next.

aqjune-aws:
Hi David, thanks for working on this topic - I think that including the refinement proof might be a too big work & interfere with the unstructured CFG. I will be glad to work on the task.
In parallel, I am trying to port a few updates that fixes problems in small-step semantics and extending the definition of correctness of transformation.
Could you wait for a while until the updates from my end are delivered please? :)

PROgram52bc:
Hi June, Yes. I am still working on adapting the call elimination proof to work with a generic failure flag (the case where preconditions fail). So I can definitely wait for your updates while iterating on that. I think it would be good to reuse some of the definitions in your updates in the current call-elimination proof.

As for the refinement proof, I do think it is quite complicated with the failure flag integrated -- if we constrain it to `failed = false`, I believe the proof should be straightforward, but that's just a conjecture.

PROgram52bc:
Superseded by a refactored proof using the instantiable `WellFormedSemanticEval` instance on Brazil mainline implemented by @aqjune-aws.

PR #1308, opened 2026-06-01
Extract Python into standalone StrataPython package

joehendrix:
Moves all Python-related Lean code into a new StrataPython/ package so the root Strata library no longer depends on language-specific frontends.

- New StrataPython package depends on Strata (not the reverse), keeping the dependency graph acyclic: StrataDDM ← Strata ← StrataPython ← StrataCLI
- DiffTestCore executable moves to StrataCLI since it's a CLI tool and StrataCLI already depends on StrataPython
- Generalize Scripts/StrataTestMain.lean with `--dir` flag so downstream packages can reuse the same test runner with different test directories
- StrataPythonTest uses `globs = ["StrataPythonTest.+"]` to avoid needing a root aggregator file
- CI workflow will need follow-up updates to run Python tests from StrataPython/ instead of filtering by prefix in the root package


StrataCLI/lake-manifest.json
```diff
@@ -30,7 +37,7 @@
    "name": "StrataDDM",
    "manifestFile": "lake-manifest.json",
    "inherited": true,
-   "dir": "../StrataDDM",
+   "dir": "../StrataPython/../StrataDDM",
```
atomb, line 40: Is this intentional?

joehendrix: It's what lake auto-generates.  I could fix it by adding an explicit StrataDDM dependency to StrataCLI, but that seems like a workaround for an auto-generated file.

StrataPython/AGENTS.md
```diff
@@ -0,0 +1,96 @@
...
+1. **Direct to Core** (`PythonToCore.lean`) — bypasses Laurel. Used by
+   `pyInterpret` and `pyAnalyzeToGoto`.
+2. **Through Laurel** (`PythonToLaurel.lean` + `PySpecPipeline.lean`) — main
+   pipeline. Combines Python source with PySpec type specifications, resolves
+   overloads, and produces typed Laurel that compiles to Core. Used by
+   `pyAnalyzeLaurel`.
+
+When adding a feature, decide which path it belongs to first. Direct-to-Core is
+simpler but lacks PySpec / overload support; the Laurel path is the strategic
+target for future work.
```
atomb: Maybe this should say that direct-to-core is deprecated and shouldn't be used?

StrataPython/README.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+This builds the `StrataPython` library, `DiffTestCore` executable, and the `StrataPythonTest` compile-time tests.
+
+## Package Purpose
+
+StrataPython provides:
+
+- **Python AST** - Types generated from the Python dialect DDM definition
+- **Python-to-Core translation** - Direct translation from Python to Core IR
```
atomb: Maybe also mention here that it's deprecated?

atomb (changes requested):
I think this mostly looks great, assuming some of the large chunks of code moved between files are indeed moves without substantial changes. There are two tiny changes I suggested to the new overview documentation that you could make alongside any CI fixes you need.

atomb (approved):

shigoel (approved):

PR #1309, opened 2026-06-01
Prove `LExpr.resolve` produces well-annotated terms

joscoh:
Proves that type inference for `LExpr` (`LExpr.resolve`) produces terms that satisfy the well-annotated typing relation `HasTypeA` (e.g. all annotations on variables and operators are consistent, giving each a uniquely determined type). This proof involves fewer preconditions than the proof of `resolve_HasType` since it deals with a simpler relation (e.g. nothing about the context or polymorphism). The only precondition needed beyond general well-formedness is the assumption that all aliases have been resolved in the context; this is guaranteed by `TEnv.addTypeAlias`


Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean
```diff
@@ -0,0 +1,1118 @@
...
+This module proves that when `LExpr.resolve` succeeds, the resulting expression
+satisfies `HasTypeA` — i.e., the type annotations placed by resolution are
+internally consistent.
+
+### Comparison with `resolve_HasType`
+
+`resolve_HasType` (in `LExprTypeSpec`) proves full polymorphic typing
+(`HasType`) but requires `WellScoped`, `allKeysFresh`, and
+`checkContextTypesClosed` assumptions .`resolve_HasTypeA` here proves
+annotation-consistency (`HasTypeA`) under weaker assumptions: only `TEnvWF`,
```
aqjune-aws: It seems it is slightly misleading to use weaker assumptions here - `AliasesResolved` isn't in `resolve_HasType`.

Strata/DL/Lambda/LExprTypeEnv.lean
```diff
@@ -143,6 +143,31 @@ structure TContext (IDMeta : Type) where
 def TContext.AliasesWF (Γ : TContext IDMeta) : Prop :=
   ∀ a, a ∈ Γ.aliases → a.WF
 
+mutual
+/-- A monotype is alias-free w.r.t. an alias list: none of its `.tcons` names
+    match any alias name in the list. -/
```
aqjune-aws: Let's clarify in the doc string that LMonoTy.aliasFree also checks the length of the parameter list because it assumes type aliases' names don't overlap with those of non-alias types.

Strata/DL/Lambda/LExprAliasFree.lean
```diff
@@ -0,0 +1,327 @@
...
+  | cons hd tl ih =>
+    simp [LMonoTys.aliasFree] at h_vals
+    cases List.mem_cons.mp h_mem with
+    | inl h_eq => subst h_eq; exact h_vals.1
+    | inr h_tl => exact ih h_vals.2 h_tl
+
+/-- `openVars` preserves list length. -/
+private theorem openVars_length (vars : List TyIdentifier) (vals : LMonoTys)
+    (mtys : LMonoTys) :
+    (LMonoTys.openVars vars vals mtys).length = mtys.length := by
```
aqjune-aws: I think this theorem can be moved to Strata/DL/Lambda/LExprTypeEnv.lean.

Strata/DL/Lambda/LExprAliasFree.lean
```diff
@@ -0,0 +1,327 @@
...
+module
+
+import Strata.DL.Lambda.LExprTypeSpec
+import all Strata.DL.Lambda.LExprTypeSpec
+
+/-! ## Alias-free infrastructure
+
+This module proves properties of `aliasFree` and `AliasesResolved`:
+
+1. Alias-free types are fixpoints of `resolveAliases`.
```
aqjune-aws: From the word 'fixpoint' I wondered whether there will be a new definition of partially ordered set for LExpr types, but it seems the proven theorems are closer to simple statements proving that `LMonoTys.resolveAliases` just returns the input type list when they are alias-free. Do you have a preference for an alternative wording that doesn't have a connotation of partially ordered set..?

Strata/DL/Lambda/LExprT.lean
```diff
@@ -71,23 +71,25 @@ Remove any type annotation stored in metadata for all
 expressions, except the `.op`s, free variables
 `.fvar`s, and bound variables in `.abs` and `.quant`.
 -/
```
aqjune-aws, line 75: Orthogonal to this pull request, but I thought the comment above this line was very confusing. I think it must be clarified as follows. Do you want to update it in this pull request as well?
* Remove any type annotation stored in metadata for all
expressions
* The optional type annotation (which is not in metadata) is reset to the one previously stored in metadata's type field
* It doesn't check if metadata.type conflicts with the stored optional type; it just overwrites them.

Strata/DL/Lambda/LExprAliasFree.lean
```diff
@@ -0,0 +1,327 @@
...
+2. The output of `resolveAliases` is alias-free when `AliasesResolved` holds.
+3. `instantiateWithCheck` produces alias-free output.
+4. `typeBoundVar` preserves the alias list and produces alias-free types.
```
aqjune-aws, line 23: This isn't blocking, but I think it is good to additionally clarify in this comment that this file won't have a top-level theorem (unlike LExprResolveAnnotated.lean), instead it proves useful theorems proving the preservation of alias-free property across utility functions used in LExpr type checking.

Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean
```diff
@@ -0,0 +1,1118 @@
...
+    · elim_errs h
+      cases h; rfl
+  · simp at heq_iwc
+
+
+/-- Bundled invariants that hold for the output environment of `resolveAux`:
+    context is preserved, well-formedness holds, types are non-empty, and
+    aliases remain resolved. -/
+private structure ResolveAuxInvariants [DecidableEq T.IDMeta]
+  (Env' : TEnv T.IDMeta) (Env : TEnv T.IDMeta) where
```
aqjune-aws: It wasn't clear which one from `Env'` or `Env` was the output type environment (from `ResolveAuxInvariants.mk_from_resolveAux` I could identify that it was `Env'` which is the output env) - can we make it explicit here?

aqjune-aws: Or what about renaming `ResolveAuxInvariants` to `ResolveAuxWF` and keeping `context : Env'.context = Env.context` as a separate conjunction whennever `ResolveAuxInvariants` is used..? I think this will make the statement of `ResolveAuxInvariants.mk_from_resolveAux` very clear because it will be almost "If ResolveAuxWF Env holds, ResolveAuxWF Env' holds".

Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean
```diff
@@ -0,0 +1,1118 @@
...
+    — only `AliasesResolved`. The trade-off is that it proves annotation-consistency
+    (`HasTypeA`) rather than full polymorphic typing (`HasType`). -/
+theorem resolve_HasTypeA [DecidableEq T.IDMeta] [HasGen T.IDMeta]
+    (e : LExpr T.mono) (e_typed : LExprT T.mono) (C : LContext T)
+    (Env : TEnv T.IDMeta) (Env' : TEnv T.IDMeta)
+    (h : e.resolve C Env = Except.ok (e_typed, Env'))
+    (h_envwf : TEnvWF Env)
+    (h_fwf : FactoryWF C.functions)
+    (h_resolved : TContext.AliasesResolved Env.context) :
+    LExpr.HasTypeA [] e_typed.unresolved (e_typed.toLMonoTy) := by
```
aqjune-aws, line 1082: As discusssed in a separate channel, `resolve_HasTypeA` states typing on the output term (`e_typed.unresolved`). `e_typed` and `e` are equivalent modulo type annotations (as you mentioned in a separate thread), and the diverging parts are only twofolds: (1) the type annotations in metadata, and (2) type annotations in the optional field.
Item (1) doesn't matter I believe - HasType and HasTypeA are agnostic to the metadata. A concern is item (2); since `e` might have much less type annotations, I wonder whether `HasTypeA` could accept `e`. Or if we want to use `HasType e`, it might work, but still a bridge seems necessary. How can we fill in the gap?
Not blocking for this PR.

aqjune-aws (commented):
Could you update Strata/DL/Lambda/Denote/README.md with explanation of the new files LExprResolveAnnotated.lean?

Strata/DL/Lambda/LExprTypeEnv.lean
aqjune-aws: It seems https://github.com/strata-org/Strata/issues/1239 suggests we can make this function throw when the size of arguments doesn't match - but wonder whether this function will be used in an actual executable environment.

Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean
joscoh: Yes, this is not an easy question, and I've only thought about it a little bit so far. One possibility could be a proof that, in addition to the fact that the input and output expressions are equivalent modulo metadata and annotations, we prove additionally that the output annotations refine the input annotations (e.g. by instantiating type variables). But this doesn't quite get us all the way there: intuitively this sort of bounds the original annotations from below, and we would probably need a bound from above (e.g. something like: if `resolve` succeeds, then the original expression had `.op` and `.fvar` annotations consistent with the context). My intuition is that this + refinement + a relationship between HasTypeA and HasType might be enough to get us there without needing `resolve_HasType` directly. Not completely sure though.

Strata/DL/Lambda/LExprTypeEnv.lean
joscoh: I think #1239 is orthogonal to this PR: `aliasFree` is a `Prop` so it can't throw, while that issue is about the executable code.

Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean
joscoh: Good point, fixed

Strata/DL/Lambda/LExprAliasFree.lean
joscoh: Good point. It really means that this function is the identity for alias-free types. I changed the wording.

Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean
```diff
@@ -0,0 +1,1119 @@
...
+/-- If `ResolveAuxWF Env` holds and `resolveAux` succeeds, then
+    `ResolveAuxWF Env'` holds for the output environment. -/
+private theorem ResolveAuxWF.mk_from_resolveAux
+  [DecidableEq T.IDMeta] [HasGen T.IDMeta]
+    (e : LExpr T.mono) (et : LExprT T.mono) (C : LContext T)
+    (Env Env' : TEnv T.IDMeta)
+    (h_res : resolveAux C Env e = Except.ok (et, Env'))
+    (h_envwf : TEnvWF Env) (h_ne : Env.context.types ≠ [])
+    (h_fwf : FactoryWF C.functions)
+    (h_resolved : TContext.AliasesResolved Env.context) :
```
aqjune-aws: I think `h_resolved`, `h_ne` and `h_envwf` can be combined into one `ResolveAuxWF Env`. :)

aqjune-aws (dismissed):

shigoel:
General comment: I'd recommend updating https://strata-org.github.io/Strata/langdef/html-single/#The-Strata-Core-Language-Definition--Semantics--Lambda-Denotational-Semantics as well.

Strata/DL/Lambda/LTyUnify.lean
```diff
@@ -2329,6 +2339,18 @@ theorem unify_absorbs (constraints : Constraints) (S_old S_new : SubstInfo)
     simp only [Except.ok.injEq] at h; subst h
     exact (Constraints.unifyCore_sound constraints S_old relS h_core).absorbs
 
+/-- Unification produces a substitution that makes every constraint pair equal. -/
+theorem unify_sound (constraints : Constraints) (S_old S_new : SubstInfo)
```
shigoel: Beautiful theorem! 
Nit: how about renaming this to `Constraints.unify_sound`, given this is about `Constraints.unify`? Looks like `unify_absorbs` would need to be similarly changed.

Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean
```diff
@@ -0,0 +1,1114 @@
...
+### Comparison with `resolve_HasType`
+
+`resolve_HasType` (in `LExprTypeSpec`) proves full polymorphic typing
+(`HasType`) but requires `WellScoped`, `allKeysFresh`, and
+`checkContextTypesClosed` assumptions. `resolve_HasTypeA` here proves
+annotation-consistency (`HasTypeA`) under different assumptions: `TEnvWF`,
+`FactoryWF`, and `AliasesResolved`. It drops the scoping/freshness/closure
+conditions but adds `AliasesResolved`, which `resolve_HasType` does not need.
+
+### Proof structure
```
shigoel, line 28: Thanks for this really nice write-up!

Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean
```diff
@@ -0,0 +1,1114 @@
...
+      · exact List.cons_ne_nil _ _
+      · rename_i h_not_empty; intro h_abs; simp_all; contradiction
+    exact resolveAux_TEnvWF e et C Env0 Env_out h_res h_envwf0 h_ne0 h_fwf
+
+/-! ### Layer 3: Main induction -/
+
+/-- Core soundness lemma: for any substitution `S` absorbing the output substitution,
+    the resolved and substituted expression satisfies `HasTypeA`. Proved by
+    well-founded induction on expression size. -/
+private theorem resolveAux_HasTypeA_aux [DecidableEq T.IDMeta] [HasGen T.IDMeta] :
```
shigoel, line 592: Wonder if inversion lemmas about `resolveAux` would've decreased the size of this proof. Not a blocker for this PR, but if we find ourselves wanting more proofs that are closely tied to `resolveAux`'s structure, it'd be worth proving those lemmas.

joscoh: Good idea, I think I will leave this to a future PR but I will investigate

aqjune-aws (dismissed):

shigoel (approved):

aqjune-aws (approved):

PR #1318, opened 2026-06-03
Fix/precond and conjunction main2

kondylidou:
### Summary

- In `collectWFObligations`, `Bool.And` is now handled like `Bool.Implies` and `ite`: the LHS of `&&` is added as a hypothesis before collecting WF obligations from the RHS.
- This fixes spurious `❓ unknown` out-of-bounds checks when the index bound is established by an earlier conjunct, e.g. `0 <= j && j < Sequence.length(s) && ret == Sequence.select(s, j)`.
- Adds a regression test (`seq_oob_in_conjunction.lean`) verified by both cvc5 and `grind`.

### Motivation

In a postcondition like:
```boole
ensures ∃ j :: 0 <= j && j < Sequence.length(nums) && ret == Sequence.select(nums, j)

### CODE REVIEW

**Comment by kondylidou** [comment]
@joscoh fyi:)

**Review by atomb — DISMISSED** [review]
Nice. This is something I've found very ergonomic about Dafny, and it's good to have it in Strata, too. :)

**Comment by joscoh** [comment]
This is a partial duplicate of https://github.com/strata-org/Strata/pull/711, which had larger design questions. I am OK with this PR, but do you also need "or" short circuiting (e.g. something like `i == 0 || x / i > 1` passes)?

**Comment by atomb** [comment]
> This is a partial duplicate of #711, which had larger design questions. I am OK with this PR, but do you also need "or" short circuiting (e.g. something like `i == 0 || x / i > 1` passes)?

I feel similarly in the broader sense. If we have this behavior for implication, it makes sense to have it for conjunction, too (and disjunction). But we likely should have a deeper design discussion about partial expressions in general.

File: StrataTest/Languages/Boole/seq_oob_in_conjunction.lean [hunk 1]
```diff
@@ -0,0 +1,55 @@
...
+-/
+
+import Strata.MetaVerifier
+
+open Strata
+
+/-!
+Regression test: out-of-bounds check inside a conjunction.
+
+`Bool.And` is now treated like `Bool.Implies` in `collectWFObligations`:
```
joscoh: Comments should be stateless. This should describe the test, not mention a "fix" or refer to current vs past behavior

joscoh (commented):
I would rather have the "or" case as well to make it consistent.

atomb (approved):

joscoh (approved):

PR #1320, opened 2026-06-03
feat(laurel): add Java-style ++ and -- operators

olivier-aws:
Adds prefix and postfix increment/decrement operators to the Laurel dialect, usable both as statements (`x++;`) and inside expressions (`int y := (x++) + (x++);`).

Surface tokens `++` and `--` are reserved for the new operators by renaming Laurel's string-concatenation operator from `++` to `^` (OCaml-style). Existing tests using the old token are migrated.

Implementation:
* New `.IncrDecr` constructor on `StmtExpr` carrying mode (Pre/Post), op (Incr/Decr) and an lvalue target (Local or Field).
* New `EliminateIncrDecr` pass, run first in the Laurel pipeline, lowers IncrDecr to existing constructs:
    Pre  -> (target := target ± 1)            (yields new value)
    Post -> ((target := target ± 1) ∓ 1)      (yields old value)
  Statement-position IncrDecr lowers to a clean assignment without a
  dead snapshot.
* Lift fix: `transformStmt` now lifts side effects out of expression-
  statements (only when they contain an assignment / imperative call),
  so postfix-as-statement reaches Core correctly.
* All exhaustive StmtExpr matches updated: MapStmtExpr, computeExprType,
  Resolution.resolveStmtExpr / collectStmtExpr, FilterPrelude, lift
  helpers, and both grammar tree translators.

Tests:
* StrataTest/Languages/Laurel/IncrDecrLiftTest.lean — exact lowering of stmt form, prefix, and postfix.
* StrataTest/Languages/Laurel/Examples/Fundamentals/T23_IncrDecr.lean — 10 verification procedures covering Java semantics: prefix yields new, postfix yields old, (x++)+(x++), (++x)+(++x), mixed pre/post, pre/post decrement, postfix in for-loop step, intentional failing assert.

A design plan accompanying this change is at
docs/LaurelIncrDecrPlan.md.

- Strata/DDM/*
- Strata/Languages/Boole/*
- Tools/BoogieToStrata


docs/LaurelIncrDecrPlan.md
```diff
@@ -0,0 +1,231 @@
...
+|------|---------|---------------|-------------|
+| Prefix increment | `++x` | new value (`x+1`) | `x := x+1` |
+| Postfix increment | `x++` | old value (`x`) | `x := x+1` |
+| Prefix decrement | `--x` | new value (`x-1`) | `x := x-1` |
+| Postfix decrement | `x--` | old value (`x`) | `x := x-1` |
+
+As a statement the yielded value is discarded, so prefix and postfix are
+equivalent there. The distinction only matters inside a larger expression.
+
+**Terminology note:** the request said "infix and suffix." `++`/`--` are unary,
```
keyboardDrummer: Please remove this meta comment

docs/LaurelIncrDecrPlan.md
```diff
@@ -0,0 +1,231 @@
...
+so there is no true infix form. This plan reads **infix = prefix (`++x`)** and
+**suffix = postfix (`x++`)**. *Please confirm.*
+
+---
+
+## 2. Key finding — the `++` token already exists
```
keyboardDrummer: This is also something that shouldn't be committed.

docs/LaurelIncrDecrPlan.md
```diff
@@ -0,0 +1,231 @@
+# Design Plan: Increment/Decrement Operators for Laurel
```
keyboardDrummer: Should this file be committed? Parts of it clearly shouldn't because they're tracking changes or referring to some request that's not being checked in. I will stop reviewing this file. Please manually review it and improve it or delete it before the second review.

My instinct is that this file is not necessary. All the relevant info should be in the documentation attached to the code.

Strata/Languages/Laurel/EliminateIncrDecr.lean
```diff
@@ -0,0 +1,209 @@
...
+the concrete-to-abstract translator already enforces this. A stray
+`.Declare` falls back to the expression form (which Lift will then handle
+gracefully).
+-/
+
+namespace Strata.Laurel
+
+public section
+
+private def bare (v : StmtExpr) (source : Option FileRange) : StmtExprMd :=
```
keyboardDrummer: `bare` is a strange name. Why not inline this method? I think this was inspired by another pass but it's not a useful pattern.

Strata/Languages/Laurel/EliminateIncrDecr.lean
```diff
@@ -0,0 +1,209 @@
...
+      | .Decr => .Add
+    let one := bare (.LiteralInt 1) source
+    bare (.PrimitiveOp inverseOp [assign, one]) source
+
+mutual
+
+/-- Lower `.IncrDecr` nodes appearing in expression-yielding positions.
+    Recurses into all StmtExpr children with `lowerExpr` except the bodies of
+    blocks/while-loops, whose top-level statements use `lowerStmt`. -/
+partial def lowerExpr (expr : StmtExprMd) : StmtExprMd :=
```
keyboardDrummer: Please use a utility function to traverse AST nodes that are unrelated to this transformation. Does `mapExprStmt` not work here?

Strata/Languages/Laurel/EliminateIncrDecr.lean
```diff
@@ -0,0 +1,209 @@
...
+  | .ContractOf ty f => ⟨.ContractOf ty (lowerExpr f), source⟩
+  -- Leaves with no StmtExpr children
+  | .Exit _ | .LiteralInt _ | .LiteralBool _ | .LiteralString _
+  | .LiteralDecimal _ | .New _ | .This | .Abstract | .All | .Hole .. => expr
+
+/-- Lower a statement-position `StmtExprMd`. A top-level `.IncrDecr` becomes
+    a clean `.Assign` (no value yielded). Everything else delegates to
+    `lowerExpr`, except control-flow constructors whose children remain
+    statements. -/
+partial def lowerStmt (stmt : StmtExprMd) : StmtExprMd :=
```
keyboardDrummer: I think we don't need to differentiate between expression and statement positions. If you want to optimize the cases where -- or ++ occurs in a position where the value is dropped, then I suggest enabling `Resoltion.lean` to compute for each stmtExpr whether its value is dropped, so it becomes available to all transformations. This information would also be useful to improve the lifting pass verification performance

keyboardDrummer (dismissed):
Amazing!

fabiomadge (commented):
Nice feature. The design (keep `.IncrDecr` in the AST, lower it in dedicated pass #1) is clean, the value-semantics tests are thorough, and it builds green. I ran the branch and probed the edges — **no blockers**, findings inline.

**Verified sound:** int `++`/`--` in all value positions; left-to-right eval order (`(x++)-(x++) == -1`); postfix-yields-old / prefix-yields-new. Constrained types compose correctly — `x--` on a `nat` at 0 is rejected, and `var y:T := x--` checks the *written* value against `x`'s type and the *yielded* value against `y`'s, independently (verified for lower- and upper-bounded constraints). Falls out of existing checks with no `++`-specific logic — a genuine strength.

**Relationship to #1356:** overlaps 10 files (same `main2` base); both add a new `StmtExpr` constructor through the same matches/grammar, so the second to merge rebases. Separately, #1356 fixes a pre-existing constrained/bv composite-**field** boxing bug — orthogonal to this PR, but relevant since both touch field handling; it does **not** fix `bv` increment here.

Nits: `T23_IncrDecr`/`T23b` collide with the in-flight `T23_DoWhile` (renumber one); no positive test for constrained-int `++` nor rejection test for `bv`/`real`. Also, the `.IncrDecr` arm in `FilterPrelude.lean:107` is the one genuinely live-but-untested path (PySpec, before lowering) — worth a test or a note.

Strata/Languages/Laurel/EliminateIncrDecr.lean
```diff
@@ -0,0 +1,108 @@
...
+  | .Declare param => ⟨.Var (.Local param.name), source⟩
+
+/-- Build `.Assign [target] (target ⊕ 1)` where `⊕` is `Add` for `Incr` and
+    `Sub` for `Decr`. The resulting assignment expression yields the new value. -/
+private def lowerToAssign (op : IncrDecrOp) (target : VariableMd)
+    (source : Option FileRange) : StmtExprMd :=
+  let primOp : Operation := match op with
+    | .Incr => .Add
+    | .Decr => .Sub
+  let one : StmtExprMd := ⟨.LiteralInt 1, source⟩
```
fabiomadge, line 64: **`bv`/`real`/`float64` silently unsupported.** The delta is hardcoded `.LiteralInt 1` + `.Add`/`.Sub`, so the lowering only type-checks for `int` (and constrained-over-int). Verified: `x++` on a `bv 32` → `Impossible to unify bv32 with int`; on a `real` → similar — raw Core-unifier errors that never mention `++`. Make it deliberate: either support them (bv via a width-matched literal — arguably the most Java-like `++`), or reject early. The clean reject site is the `.IncrDecr` arm in `Resolution.lean:449`: it's reachable (initial `resolve` precedes pass #1) and `getType` there distinguishes int / constrained-over-int / `bv` / `real` without duplicating `ConstrainedTypeElim` — so you get a Laurel diagnostic with a source range instead of unifier jargon.

StrataTest/Languages/Laurel/Examples/Fundamentals/T23_IncrDecr.lean
```diff
@@ -0,0 +1,274 @@
...
+};
+
+procedure mixedIncrInArithmetic()
+  opaque
+{
+  var x: int := 1;
+  // `x++ * 2 + ++x - x--` parses as `((x++ * 2) + ++x) - x--`.
+  // Lift evaluates right-to-left:
+  //   x-- → snapshot $x_0 (= 1), x := 0;     (yielded value: $x_0 = 1)
+  // wait, let me re-trace... actually with our pass, the values seen are:
```
fabiomadge: **Leftover artifact + wrong comment.** `// wait, let me re-trace...` should go, and the nearby "Lift evaluates right-to-left" is wrong — the order is left-to-right (verified).

CLAUDE.md
```diff
@@ -0,0 +1,145 @@
+# CLAUDE.md
```
fabiomadge: **Split into its own PR.** This is a net-new 145-line repo-wide guidance file, unrelated to `++`/`--`, and the repo is about to split — better reviewed/landed separately.

Strata/Languages/Laurel/EliminateIncrDecr.lean
olivier-aws: Added a specific error message when using ++/-- on real/float/bv types

CLAUDE.md
olivier-aws: Removed by commit ddd7c3056a25e5b63a9f8e1f4873b0ab8ccf5292

olivier-aws:
Thanks @fabiomadge for your comment, took them into account.

fabiomadge (dismissed):
Re-reviewed on `ddd7c3056`. My three earlier comments are all addressed — the `bv`/`real`/`float64` rejection is now a clean resolution-time `UserError` (`Resolution.lean:307`), and the test artifact and `CLAUDE.md` are gone. Built `T23_IncrDecr` + `T23b_IncrDecrField` + `IncrDecrTypeRejectionTest`: green. The lowering is correct (re-checked Pre/Post and the `x++ + x++` double-eval semantics). LGTM.

One optional, non-blocking note:

**Ergonomics.** Field `++`/`--` currently needs parens (`(c#n)++`), as documented in `T23b`. `fieldAccess` and `postIncr` are both trailing ops at `prec(90)`, so `c#n++` is a precedence tie. Raising `fieldAccess` to `prec(95)` lets it parse paren-free; I verified it (parses, increments once, full suite green). Only covers single-level fields — chained access (`a#b#c`) is a separate, deeper gap (filed as #1371). Fine to defer.

fabiomadge:
Since `main2` is locked for the restructuring (likely another week or so), this PR can't merge there in the meantime. If you'd rather not wait, you can retarget it to `reviewed-kbd-will-merge-to-main` — that branch is open and @keyboardDrummer merges it onward to main. The retarget is clean (the two bases differ by only a commit either way, so no diff blowup). My approval carries over either way.

keyboardDrummer (approved):

fabiomadge (approved):

PR #1322, opened 2026-06-03
Add composite GitHub Actions for lint scripts

shigoel:
## Summary
- Wrap the four lint scripts (`check_copyright_headers.py`, `lintWhitespace.sh`, `checkLeanImport.sh`, `checkNoPanic.sh`) as reusable composite actions under `.github/actions/`.
- Enables other repos in `strata-org` to reference them cross-repo (e.g. `uses: strata-org/Strata/.github/actions/check-copyright-headers@main`), matching the existing pattern used by `install-cvc5`.
- Each action references the canonical script in `.github/scripts/` via a relative path from `github.action_path`, avoiding duplication and drift.
- Updates `ci.yml` to use the new composite actions in the `lint_checks` job.
- Parameterizes `lintWhitespace.sh` to accept an optional directory argument (defaults to `Strata`).

## Cross-repo usage example
```yaml
- uses: actions/checkout@v6
  with:
    fetch-depth: 0  # required by check-no-panic (needs git history)
- uses: strata-org/Strata/.github/actions/check-copyright-headers@main
- uses: strata-org/Strata/.github/actions/lint-whitespace@main
  with:
    directory: MyLeanLib
- uses: strata-org/Strata/.github/actions/check-lean-import@main
- uses: strata-org/Strata/.github/actions/check-no-panic@main
  with:
    base-ref: "origin/main"
```

> **Note:** `check-no-panic` requires `fetch-depth: 0` in the checkout step because it computes a merge base via `git merge-base` to diff only newly-added lines. The other three actions only scan files on disk and work fine with a shallow clone.

## Test plan
- [ ] CI `lint_checks` job passes on this PR (validates the composite actions work with local `./` paths)
- [ ] Verify no behavioral change: same scripts, same defaults, same exit codes


atomb (approved):
This looks great.

.github/actions/check-lean-import/action.yml
```diff
@@ -0,0 +1,22 @@
...
+    default: "."
+
+runs:
+  using: composite
+  steps:
+    - name: Check for bare import Lean
+      shell: bash
+      env:
+        LINT_DIR: ${{ inputs.directory }}
+      run: |
```
atomb, line 21: **Nit (non-blocking):** This action inlines the logic instead of delegating to the canonical `checkLeanImport.sh` script, unlike the other three actions which all use `github.action_path` to call their script. This means the script and action can drift independently.

Consider parameterizing `checkLeanImport.sh` similarly to `lintWhitespace.sh` and calling it from the action.

joehendrix (approved):

.github/actions/check-copyright-headers/action.yml
```diff
@@ -0,0 +1,19 @@
...
+    required: false
+    default: "."
+
+runs:
+  using: composite
+  steps:
+    - name: Check copyright headers
+      shell: bash
+      run: |
+        python3 "${{ github.action_path }}/../../scripts/check_copyright_headers.py" "${{ inputs.directory }}"
```
atomb, line 19: **Nit (non-blocking):** The underlying Python script hardcodes a `Tools/` subdirectory for `.py` header checks (`check_all_source_files(directory_path / 'Tools', ext=".py", ...)`). When cross-repo consumers pass a custom `directory` value, the script will look for `<directory>/Tools/*.py` which may not exist. It silently passes in that case so it won't break, but worth documenting this assumption in the action description for downstream clarity.

.github/actions/lint-whitespace/action.yml
```diff
@@ -0,0 +1,19 @@
...
+name: Lint whitespace
+description: >
+  Check .lean files for trailing whitespace and missing final newlines.
+
+inputs:
+  directory:
+    description: Directory to scan for .lean files (relative to the workspace).
+    required: false
+    default: "Strata"
+
```
atomb, line 12: **Nit (non-blocking):** Minor inconsistency in default `directory` values across the new actions: `lint-whitespace` defaults to `Strata` while `check-copyright-headers` and `check-lean-import` default to `.`. This matches the original behavior (the old script hardcoded `Strata`), so not a bug, but could be surprising for cross-repo consumers who expect uniform defaults. Consider either documenting the discrepancy or normalizing to `.` and having `ci.yml` pass `directory: Strata` explicitly.

.github/workflows/ci.yml
```diff
@@ -184,16 +184,17 @@ jobs:
...
-        run: .github/scripts/lintWhitespace.sh
+        uses: ./.github/actions/lint-whitespace
       - name: Check for import Lean
-        run: .github/scripts/checkLeanImport.sh
+        uses: ./.github/actions/check-lean-import
       - name: Check Lean version consistncy
         run: .github/scripts/check_lean_consistency.sh
       - name: Check for new panic! usage
-        run: .github/scripts/checkNoPanic.sh "origin/${{ github.base_ref || 'main' }}"
+        uses: ./.github/actions/check-no-panic
```
atomb, line 195: **Nit (non-blocking):** Pre-existing typo: "consistncy" should be "consistency". Since the surrounding lines are being modified, this would be a good opportunity to fix it.

.github/scripts/lintWhitespace.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
```
atomb, line 1: **Nit (non-blocking):** The `lintWhitespace.sh` script does not use `set -euo pipefail`. The `check-no-panic` script already has it, but for cross-repo robustness you might consider adding it here as well (or at least `set -e`) so that unexpected errors in `find`/`grep` don't get silently swallowed.

atomb:
I had my agent post some non-blocking comments, but they're more for potential future improvements.

PR #1325, opened 2026-06-04
Add OverapproxiatesUpto/OverapproximatesWhen/OverapproximatesAggressivelyWhen

aqjune-aws:
Updates in specification of transformations to support several missing features needed by loop elimination correctness.

* A new `OverapproximatesAggressively` that allows spurious assertion failures even when executions started from the exactly same initial program state. :)
  * If transformation `T` satisfies `OverapproximatesAggressively`, it also satisfies `Overapproximates`.

* A new `OverapproximatesUpto(When)`: an extension of `Overapproximates` in the sense that it takes the mapping relation between program states of the input and output program.
  * `OverapproxiamtesUptoWhen` additionally takes the precondition on the input program.

* New `OverapproximatesWhen` / `OverapproximatesAggressivelyWhen`:
  * Precondition-bearing variants: take a `pre : Stmt → Prop` that the input program (which is an imperative statement) must satisfy.
  * Original `Overapproximates` / `OverapproximatesAggressively` are now `*When`-wrappers with `pre := fun _ => True`.
  * `*.toWhen` and `*When.strengthen` provided in SpecificationProps.

* `InitEnvWF`: a structure capturing well-formedness of the initial program state (`Env`).

  * `defUseWellFormed` / `Block.defUseWellFormed`: a static check that every use of a variable references a previously-defined variable, and every use of an operator references a previously-`funcDecl`'d name.

  * Reflection-based `factoryDeclared` predicate replacing the prior hard-coded `factoryFuncNames` list (derived from `WFFactoryArray`).


aqjune-aws:
Hey, this PR is too large. Can we split it into smaller ones?

aqjune-aws:
#1327

aqjune-aws:
#1337 will factor out (1) Updates / fixes in Imperative small-step semantics

aqjune-aws:
This pull request will be moved to the internal repository of Strata after migration.

PR #1327, opened 2026-06-04
Factor out theorems and lemmas into dedicated Props files

### LINKED ISSUE #1067: Factor out proofs in semantics related files as `*Prop.lean`

aqjune-aws:
Resolves #1067 
NOTE: There are two large files that don't follow the convention - Strata/DL/Lambda/LExprTypeSpec.lean and Strata/DL/Lambda/Semantics.lean , but a concern is that in these cases if theorems are all moved to say LExprTypeSpecProps.lean and SemanticsProps.lean then the original files literally have only one inductive predicate definition. :/ So I chose to unsplit them...

Split definitions from their associated theorems by moving the latter
into companion `*Props.lean` files. This separates the definitional
content from proofs, making each file's responsibility clearer and
easier to navigate.

- StmtSemantics.lean → StmtSemanticsProps.lean (multi-step properties,
  inversion lemmas, hasFailure monotonicity/irrelevance, exit-coverage
  preservation, noFuncDecl preservation, assert helpers, set commutation)
- CmdSemantics.lean → CmdSemanticsProps.lean (isDefined/isNotDefined
  lemmas, substitution flips, InitState/UpdateState properties,
  assert/set commutation)
- Specification.lean → SpecificationProps.lean (Hoare rules and the
  bidirectional connection between Triple and AssertValid, Sound and
  Overapproximates theorems, statement-list overapproximation)
- LTyUnify.lean → LTyUnifyProps.lean (absorption lemmas, unifyOne_sound,
  unifyCore_sound, unify_absorbs, plus their supporting private helpers
  and the UnifyOneProperties/UnifyCoreProperties bundles)
- Stmt.lean's exitsCoveredByBlocks lemmas: unused; delete

The freestanding SemanticsProps.lean is removed; its theorems are
distributed between CmdSemanticsProps.lean and StmtSemanticsProps.lean
based on whether they reason about commands or statement-level
configurations.

---

Also, this PR moves Strata/DL/Lambda/Lambda.lean to Strata/DL/Lambda.lean, to make it consistent with Strata/DL/{Imperative,SMT}.lean


joscoh (dismissed):

Strata/Transform/Specification.lean
```diff
@@ -177,26 +176,6 @@ def Triple
...
-    {Pre Pre' : Env P → Prop} {Post Post' : Env P → Prop} {s : L.StmtT}
-    (h : Triple L Pre s Post)
-    (hpre : ∀ ρ, Pre' ρ → Pre ρ) (hpost : ∀ ρ, Post ρ → Post' ρ) :
-    Triple L Pre' s Post' := by
-  intro ρ₀ ρ' hpre' hwfb hf₀ hstar
-  have ⟨hp, hf⟩ := h ρ₀ ρ' (hpre ρ₀ hpre') hwfb hf₀ hstar
-  exact ⟨hpost ρ' hp, hf⟩
-
-
 /-! ## Structural Hoare rules (Imperative-specific) -/
```
PROgram52bc: Perhaps this comment is now outdated?

Strata/Transform/Specification.lean
PROgram52bc: This comment might be outdated, too.

PROgram52bc (dismissed):

shigoel:
Agent said: Reviewed the full diff. Every theorem body I compared between the deletion and addition is identical — no subtle changes to proof terms, hypotheses, or conclusions. The refactor is purely mechanical.

One potential concern: three lemmas are deleted from `Stmt.lean` without being moved elsewhere — `block_exitsCoveredByBlocks_append` (trivial), `all_cmd_exitsCoveredByBlocks` (simple), and `exitsCoveredByBlocks_weaken` (moderately involved — ~50 lines of mutual induction over `Stmt`/`List Stmt` with `motive_2`). The build passing confirms they're unused today, but `exitsCoveredByBlocks_weaken` would be the most tedious to re-derive if needed later due to its mutual recursion scaffolding. Worth noting in case future work on exit-coverage reasoning needs it.

shigoel (approved):

joscoh (approved):

PR #1328, opened 2026-06-04
feat(laurel): Adding surface syntax for calling instance procedures & lowering to Core

### LINKED ISSUE #1321: Laurel resolution pass should allow instance procedures on different composite types with the same name
The following Laurel program will cause an (unexpected error) during the resolution pass:
```
composite Counter {
  var count: int
  procedure tick(self: Counter) opaque modifies self { self#count := self#count + 1 };
}

composite Clock {
  var time: int
  procedure tick(self: Clock) opaque modifies self {
    self#time := self#time + 1
  };
}
```
Error message is: `error: Duplicate definition 'tick' is already defined in this scope`

This roots from line 248 in https://github.com/strata-org/Strata/blob/8ff810199ca7aa95b9c26f37b9e4a972bf0c8e2f/Strata/Languages/Laurel/Resolution.lean#L245-L250

Full reproduction:
```lean
import StrataTest.Util.TestDiagnostics
import Strata.DDM.Elab
import Strata.DDM.BuiltinDialects.Init
import Strata.Languages.Laurel.Grammar.LaurelGrammar
import Strata.Languages.Laurel.Grammar.ConcreteToAbstractTreeTranslator
import Strata.Languages.Laurel.Resolution

open StrataTest.Util
open Strata
open Strata.Elab (parseStrataProgramFromDialect)

namespace Strata.Laurel

private def processResolution (input : Lean.Parser.InputContext) : IO (Array Diagnostic) := do
  let dialects := Strata.Elab.LoadedDialects.ofDialects! #[initDialect, Laurel]
  let strataProgram ← parseStrataProgramFromDialect dialects Laurel.name input
  let uri := Strata.Uri.file input.fileName
  match Laurel.TransM.run uri (Laurel.parseProgram strataProgram) with
  | .error e => throw (IO.userError s!"Translation errors: {e}")
  | .ok program =>
    let result := resolve program
    let files := Map.insert Map.empty uri input.fileMap
    return result.errors.toList.map (fun dm => dm.toDiagnostic files) |>.toArray

/-- Same method name `tick` on two different composites. Today this fires
    `Duplicate definition 'tick' is already defined in this scope`. -/
def crossCompositeSameName := r"
composite Counter {
  var count: int
  procedure tick(self: Counter) opaque modifies self { self#count := self#count + 1 };
}

composite Clock {
  var time: int
  procedure tick(self: Clock) opaque modifies self {
//          ^^^^ error: Duplicate definition 'tick' is already defined in this scope
    self#time := self#time + 1
  };
}
"

#guard_msgs (error, drop all) in
#eval testInputWithOffset "CrossCompositeSameName" crossCompositeSameName 33 processResolution

/-- Distinct method names on two different composites. Resolves cleanly. -/
def crossCompositeDifferentName := r"
composite Counter {
  var count: int
  procedure tickCounter(self: Counter) opaque modifies self { self#count := self#count + 1 };
}

composite Clock {
  var time: int
  procedure tickClock(self: Clock) opaque modifies self { self#time := self#time + 1 };
}
"

#guard_msgs (error, drop all) in
#eval testInputWithOffset "CrossCompositeDifferentName" crossCompositeDifferentName 49 processResolution

end Laurel
```

AD1024:
## Summary

This PR addresses the problem that Laurel composite types could *declare* instance procedures inside `composite { ... }` blocks, but they couldn't be compiled or called. The Laurel→Core translator unconditionally rejected every instance procedure with a `NotYetImplemented` diagnostic, and even without that block they would have been silently dropped: the SCC ordering in [`CoreGroupingAndOrdering.lean`](Strata/Languages/Laurel/CoreGroupingAndOrdering.lean) only enumerates `program.staticProcedures`, so anything on `CompositeType.instanceProcedures` never reached Strata Core.

### 1. Surface syntax of instance procedure calls: `obj#method(args)`

- **Parser** ([`ConcreteToAbstractTreeTranslator.lean`](Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean)): when the callee of `call(...)` is a `fieldAccess` node, emit `InstanceCall target method args` instead of dropping the receiver into an empty-string static call.
- **Grammar** ([`LaurelGrammar.st`](Strata/Languages/Laurel/Grammar/LaurelGrammar.st)): adjust `call`'s precedence so `c#m(args)` parses as `call(fieldAccess(c, m), args)`.
- **Resolution** ([`Resolution.lean`](Strata/Languages/Laurel/Resolution.lean)): pre-register instance procedures in the global scope under their lifted key (`<CompositeName>$<methodName>`). Two composites can now share a method name without colliding (Note: this change seems to resolve #1321 but the the issue of the Resolver still exists - variable scope should be carefully refactored for the resolution pass). `InstanceCall` resolution looks up the receiver's composite type, builds the lifted key, and stamps the resolved `uniqueId` on the original callee identifier.

### 2. Laurel-to-Laurel pass: `LiftInstanceProcedures`

A new pass under [`Strata/Languages/Laurel/LiftInstanceProcedures.lean`](Strata/Languages/Laurel/LiftInstanceProcedures.lean), wired into [`LaurelCompilationPipeline.lean`](Strata/Languages/Laurel/LaurelCompilationPipeline.lean) with `needsResolves := true` between `EliminateValueReturns` and `HeapParameterization`. The pass:

- Walks every `.Composite ct` in `program.types` and clones each `proc ∈ ct.instanceProcedures` into a fresh top-level static procedure named `<CompositeName>$<methodName>`. Body, parameters, contracts, and `invokeOn` are copied verbatim.
- Walks the entire program and rewrites every `InstanceCall` whose callee so it points at the lifted name (with the receiver prepended as the first argument to match the lifted procedure's `self : <CompositeName>` parameter).
- Clears `ct.instanceProcedures := []` on every composite and appends the lifted procedures to `program.staticProcedures`.

Downstream simplifications: [`HeapParameterization.lean`](Strata/Languages/Laurel/HeapParameterization.lean) drops its secondary traversal of `instanceProcedures` (now always empty), and the `NotYetImplemented` block in [`LaurelToCoreTranslator.lean`](Strata/Languages/Laurel/LaurelToCoreTranslator.lean) is replaced by a defensive `StrataBug` assertion that fires only on a pass-ordering regression (all instance procedures should already be lifted and rewritten at this point).


Strata/Languages/Laurel/LiftInstanceProcedures.lean
```diff
@@ -0,0 +1,213 @@
...
+
+The pipeline runs `resolve` again after this pass (`needsResolves := true`)
+so the fresh static procedures and the rewritten call sites get bound.
+-/
+
+namespace Strata.Laurel
+
+/-! ## uniqueId clearing on cloned procedures
+
+The cloned procedure inherits `Identifier.uniqueId` values from the prior
```
keyboardDrummer: I don't understand the issue that requires uuid clearing. The moved procedure is given a new name, but I would think that uuid wise everything can stay exactly the same. 

"two lifted procs both with a `self` parameter sharing a uid from the original composite scope" what is shared? All the references in two separate procedures will have their own uuid. They may resolve to things with the same uuid, but that's fine.

Strata/Languages/Laurel/LiftInstanceProcedures.lean
```diff
@@ -0,0 +1,213 @@
...
+      match td with
+      | .Composite ct =>
+        acc ++ ct.instanceProcedures.map fun proc =>
+          let renamed := { proc with name := liftedProcName ct.name proc.name }
+          clearProcUniqueIds renamed
+      | _ => acc
+
+  if liftedProcs.isEmpty then program else
+
+  -- Step 2: rewrite call sites everywhere they could appear.
```
keyboardDrummer: There's quite a lot of boilerplate in rewriting the calls across the program. Could we introduce a `mapStmtExpr` that works on entire programs?

keyboardDrummer (commented):
Really nice, thanks! Just A FYI: what should be done still, in another PR, by whomever is up to it, is add support for dynamic dispatch when calling abstract instance procedures. That feature is the main reason Laurel has instance procedures, since otherwise you could easily use static procedures instead of instance ones.

Strata/Languages/Laurel/LiftInstanceProcedures.lean
AD1024: I see. That makes sense to me. I think on the parse path, uids are always unique so it is fine to skip this. We will maintain the invariant of the resolution pass in this transformation.

Strata/Languages/Laurel/LiftInstanceProcedures.lean
AD1024: Yes - and in general I think we should have `AstTransformer` and `AstVisitor` at some point. There are many passes that does the structural recursion manually, and each of them needs a (very similar) termination proof.

AD1024:
Thanks for your feedback Remi!
I addressed your first comment - the UID scraping is indeed unnecessary (if we always take the program from a resolution pass); we only need to maintain the invariant given by the resolution pass on uids. 

For your second point, I definitely agree that the rewrite passes need refactor. Maybe we can factor this out to another PR? What do you think.

Strata/Languages/Laurel/LaurelAST.lean
```diff
@@ -544,6 +544,12 @@ def DatatypeDefinition.destructorName (dt : DatatypeDefinition) (field : Paramet
 def DatatypeDefinition.unsafeDestructorName (dt : DatatypeDefinition) (field : Parameter) : String :=
   s!"{dt.name.text}..{field.name.text}!"
 
+/-- Canonical lifted name for an instance procedure, used by both Resolution
+    pre-registration and the `LiftInstanceProcedures` pass.
+    Format: `"<CompositeName>$<methodName>"`. -/
+def liftedProcName (typeName methodName : Identifier) : Identifier :=
```
keyboardDrummer: Would prefer this living in `LiftInstanceProcedures.lean`

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -883,7 +898,13 @@ private def preRegisterTopLevel (program : Program) : ResolveM Unit := do
         let qualifiedName := ct.name.text ++ "." ++ field.name.text
         let _ ← defineNameCheckDup field.name (.field ct.name field) (some qualifiedName)
       for proc in ct.instanceProcedures do
+        -- Register the instance procedure under its lifted name
+        -- (`<CompositeName>$<methodName>`) so two composites can share a method
+        -- name without colliding in the global scope. `obj#method(args)` call
+        -- sites are resolved by `resolveStmtExpr` looking up this lifted key.
+        let liftedKey := (liftedProcName ct.name proc.name).text
```
keyboardDrummer: I think the only reason we need `liftedProcName` in `Resolution.lean` is because we don't have a concept of nested scopes yet. The `liftedProcName` does not actually need to be the same as the one used in the instance procedure pass. I think it would be good to clarify that, by duplicating the function `liftedProcName`, and renaming the one used in `Resolution.lean` to something like `containerScopedName`

keyboardDrummer (commented):
Really nice! Left one more comment.

keyboardDrummer:
For the build feature, I think I have a fix here: https://github.com/strata-org/Strata/pull/1313

keyboardDrummer:
For the build issue. I don't think it's caused by this PR. I hope https://github.com/strata-org/Strata/pull/1313 will fix that.

AD1024:
Thanks! I found one more issue (calls on nested objects) when using this PR'a feature. Trying to fix it.

keyboardDrummer (dismissed):

AD1024:
FIxed the conflicts and redirect the merge to the backlog branch. 
@keyboardDrummer @ssomayyajula

keyboardDrummer (dismissed):

ssomayyajula (approved):

keyboardDrummer (approved):

PR #1332, opened 2026-06-05
Add induction principle for `resolveAux` and clean up type inference proofs

joscoh:
This PR cleans up and refactors the `resolve` proofs in 4 ways:
1. It defines an induction principle over `resolveAux` (`resolveAux_ind` on line 5453) which proves most of the boilerplate once-and-for-all rather than repeatedly in each proof. Proofs about `resolveAux` can now use this induction principle instead of needing strong induction on size and then threading through all of the assumptions about well-formedness, substitution absorption, etc. This has the concrete benefit of making the proof of `resolveAux_hasType` about 350 lines shorter. This addresses a TODO from https://github.com/strata-org/Strata/pull/1309.
2. It uses the `elim_err/elim_errs` tactics to greatly reduce the amount of nesting in proofs for case split -> 1 trivial goal, making the proofs more readable.
3. It splits `resolve_HasType` into `resolve_HasType_core`, generalized over type substitutions and requiring fewer preconditions (and needed for commands in a future PR) and the main theorem.
4. It removes uses of Lean auto-generated names (e.g. `this`) from proofs in `LExprTypeSpec.lean`.


shigoel (approved):
A much nicer solution to have an induction principle than individual inversion lemmas! Thanks!

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -5497,6 +5452,347 @@ theorem TEnvWF.of_resolveAux
...
+      P c ct C Env Env1 → P th tht C Env1 Env2 → P el elt C Env2 Env3 →
+      P (.ite m c th el) et C Env Env')
+    -- Main statement
+    (e : LExpr T.mono) (et : LExprT T.mono) (C : LContext T)
+    (Env Env' : TEnv T.IDMeta)
+    (h_res : resolveAux C Env e = .ok (et, Env'))
+    (h_envwf : TEnvWF Env)
+    (h_ne : Env.context.types ≠ [])
+    (h_fwf : FactoryWF C.functions) :
+    P e et C Env Env' := by
```
aqjune-aws, line 5585: Wonder how AI could find this...

Strata/DL/Lambda/LExprTypeSpec.lean
```diff
@@ -311,18 +323,27 @@ theorem applySubstT_toLMonoTy {T : LExprParamsT}
...
-2. **`resolve_HasType`**: The top-level theorem. Since `resolve` is just
-   `resolveAux` followed by `applySubstT`, we decompose the hypothesis,
-   apply `resolveAux_HasType` (instantiating `S` with the final substitution),
-   and use `applySubstT_toLMonoTy`.
+2. **`resolve_HasType_core`**: Lifts `resolveAux_HasType` through `resolve` (which is
+   `resolveAux` followed by `applySubstT`, plus the empty-context guard). It exposes the
+   *universally-quantified-over-`S`* typing conclusion together with idempotence and
+   `TEnvWF Env'`, under the **minimal** preconditions (`TEnvWF`, `FactoryWF`,
+   `WellScoped` — no `checkContextTypesClosed`/`allKeysFresh`). This is the form
+   consumed by callers that compose substitutions themselves (e.g. `CmdType.inferType_HasType`).
```
aqjune-aws, line 340: Looking forward to it

aqjune-aws (approved):

PR #1337, opened 2026-06-05
Fix evaluator scoping in Imperative small-step semantics

aqjune-aws:
Reworks the small-step semantics of Imperative to dix a scoping bug regarding internal function declaration.

Subset of the changes from PR #1325 — the `OverapproximatesWhen`/`OverapproximatesAggressively`/`InitEnvWF` work is deferred to a follow-up PR.

* Evaluator scoping (FIX of a scoping bug — function declarations leaked out of their enclosing block): Inner function declarations (`step_funcDecl`) extend the evaluator (`Env.eval`) with the new function, and previously, that extension persisted past the enclosing block — i.e. a `funcDecl` inside a block was visible to code AFTER the block exited. This is wrong: by the language's variable-scoping rules, a `funcDecl`'s scope is its enclosing block, exactly parallel to `init`. The fix:
  * `Config.block` now carries a snapshot of the parent evaluator (`e_parent`) in addition to the parent store (`σ_parent`): `Config.block label σ_parent e_parent inner` (was 3-arg).
  * On `step_block_done` / `step_block_exit_match` / `step_block_exit_mismatch` the result config's `eval` is RESTORED to `e_parent`, so a `funcDecl` introduced inside the block is NOT visible after exit.

* New HasFvars / HasOps / HasOpsImp typeclasses:
  * `HasFvars` returns ALL free variables in an expression.
  * `HasOps` (`P.Expr → List P.Ident`) returns operator/function names referenced in an expression, parallel to `HasFvars` for variables.
  * `HasOpsImp` lifts that to commands (`Cmd.getOps`, `Cmds.getOps`).

* Refactoring of HasBool / HasBoolOps:
  * `HasBool` now extends `HasVal` and bundles `boolIsVal` (former `HasBoolVal` is folded in).
  * Boolean operations split out into `HasBoolOps` (extends `HasBool`): not, and, imp.


Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -246,23 +303,37 @@ inductive StepStmt
       labeled pairs `(String × P.Expr)`; only the expression part is
       evaluated.
 
+      Similarly, when a measure is present, it must evaluate to some integer
+      `v` and `v < 0` must evaluate to a boolean.  If `v < 0` evaluates to
+      `tt`, the cumulative `hasFailure` flag is set via `hasMeasureFailure`,
+      matching the invariant-failure pattern.  The measure is a labeled pair
+      `(String × P.Expr)` whose label can join the assertion-label collection
+      via `isAtAssert`.
+
```
atomb: I'm unconvinced that the measure should be part of the semantics. (I'm not even fully convinced that it should be in the formal AST -- I think it would make more sense as metadata.)

Is there a reason that it needs to be? Are there things you're trying to prove that require it?

aqjune-aws: It was motivated from a few thoughts:

Firstly the distinction between whether a component must be explicitly in AST vs. metadata depends on whether it should (or is desirable to?) appear in the formal semantics, and measure was in AST, so I thought it was a bit odd to not be described in the formal semantics. 

The second one (that is especially about adding a label to measure) is about the provenance of new assertions created by LoopElim (not related to proof as well). Currently loop elimination creates an assert that checks whether the measure truly decreases after loop elimination - but there was no way to find an assertion or any label in the source program that can be linked with the new assertion. A String label field was added next to the measure field in this patch for this reason (it will be Provenance once it is used for distinguishing labels), and once we have a way of tracking a chain of provenance of assertion, this will nicely point that 'hey, it is this measure that made this assertion'. :)

From the proof of LoopElim, absence of this check isn't a blocker because overapproximation must allow introduction of assertion failures that didn't happen in the source program.
But there can be other ways of checking the decreasing measure without relying on LoopElim, and I thought that if this is enforced in the semantics, it won't be missed in the different pipeline.

Actually I wonder how it works in other verification languages that provide a syntax for measure. Will it be used as a hint for termination, or does it strictly check whether the measure is truly decreasing...?

aqjune-aws: Switching the measure decrease check to optional, as discussed in a separate thread

joscoh: If we wanted to reason about total correctness, would we need the measure to be part of the semantics?

aqjune-aws: Our tentative conclusion was that having it in the semantics might be too restrictive if Strata has a pipeline that lowers to an analysis that doesn't support reasoning about termination by default (e.g., a simple implementation of abstract interpretation). It will make proving soundness of the pipeline challenging.
We will decouple measure check with state change & keep it in metadata, and it will be used as a hint when proving termination.

atomb: In response to @joscoh's question, I think it would still be possible to prove that, if a measure exists and decreases, evaluation will terminate, even if it's not part of the evaluation rules.

aqjune-aws:
Removed the measure check from the small-step semantics. The well-formedness of the measure (must evaluate to an integer numeral) is kept - it enforces introduction of 'assert (measure >= 0)' be well-typed, otherwise the evaluation of the assert (introduced by LoopElim) will stuck.

Strata/DL/Imperative/StmtSemantics.lean
```diff
@@ -289,17 +319,17 @@ inductive StepStmt
     StepStmt EvalCmd extendEval
       (.stmt (.loop .nondet m inv body md) ρ)
       (.seq
-        (.block .none ρ.store (.stmts body
+        (.block .none ρ.store ρ.eval (.stmts body
           { ρ with hasFailure := ρ.hasFailure || hasInvFailure }))
         [.loop .nondet m inv body md])
 
-  /-- Non-deterministic loop: exit the loop. -/
+  /-- Non-deterministic loop: exit the loop.  Measure must be `.none`. -/
```
joscoh: This seems more like a typing rule than an evaluation rule; is there a reason to have the `none` restriction in the step relation?

Strata/DL/Imperative/Stmt.lean
```diff
@@ -107,7 +111,7 @@ def Stmt.inductionOn {P : PureExpr} {Cmd : Type}
 /-! ### sizeOf -/
 
 mutual
-@[simp]
+@[simp, expose]
```
joscoh: I'm a bit surprised that these need to be exposed when so far we have been able to reason about statement/block termination without exposing them.

Strata/DL/Imperative/StmtSemanticsProps.lean
```diff
@@ -713,121 +782,35 @@ theorem block_noFuncDecl_preserves_eval
...
+theorem block_noFuncDecl_preserves_eval_exiting
+    (ss : List (Stmt P CmdT)) (ρ ρ' : Env P) (lbl : String)
+    (hnofd : Block.noFuncDecl ss = true)
+    (hexit : StepStmtStar P EvalCmd extendEval (.stmts ss ρ) (.exiting lbl ρ')) :
+    ρ'.eval = ρ.eval := by
+  suffices ∀ c₁ c₂,
+      Config.noFuncDecl c₁ →
+      StepStmtStar P EvalCmd extendEval c₁ c₂ →
+      c₂.getEnv.eval = c₁.getEnv.eval by
+    exact this _ _ (show Config.noFuncDecl (.stmts ss ρ) from hnofd) hexit
```
joscoh: Nit: we should avoid auto-generated names

Strata/DL/Imperative/StmtSemanticsProps.lean
```diff
@@ -1369,5 +1478,322 @@ theorem eval_stmts_set_comm
...
+  {EvalCmd : EvalCmdParam P CmdT} {extendEval : ExtendEval P}
+
+omit [HasOps P] in
+/-- Invert a `.seq` execution reaching terminal in `ReflTransT`: the inner
+    terminates first, then the tail stmts run to terminal.  Length bound
+    is strict so callers can recurse on `hstar.len`. -/
+theorem seqT_reaches_terminal
+    {inner : Config P CmdT} {ss : List (Stmt P CmdT)} {ρ' : Env P}
+    (hstar : ReflTransT (StepStmt P EvalCmd extendEval) (.seq inner ss) (.terminal ρ')) :
+    ∃ (ρ₁ : Env P), ∃ (h1 : ReflTransT (StepStmt P EvalCmd extendEval) inner (.terminal ρ₁)),
```
joscoh, line 1560: I could be missing something here, but it seems strange to me to have a relation in `Type` so that it can be used to define a function (via termination/size functions/etc) and yet the decomposition lemmas involve `exists` which lives in `Prop` (and so cannot be used for the same purpose). Could you briefly explain why the `Prop`-valued decomposition lemmas are sufficient?

Strata/DL/Imperative/PureExpr.lean
```diff
@@ -76,11 +51,51 @@ class HasFvar (P : PureExpr) where
...
+  numeralIsValue : ∀ n, isNumeral n = Bool.true → (@HasVal.value P) n
+  zeroIsNumeral : isNumeral zero = Bool.true
+  numeralHasNoFvars : ∀ (n : P.Expr), isNumeral n = Bool.true →
+    HasFvars.getFvars (P := P) n = []
+
+/-- Integer arithmetic / comparison primitives. -/
+class HasIntOps (P : PureExpr) [HasBool P] [HasFvars P] [HasInt P] where
+  eq    : P.Expr → P.Expr → P.Expr
+  lt    : P.Expr → P.Expr → P.Expr
+  decr  : P.Expr → P.Expr
```
joscoh: Is `decr` subtraction? If so, does it need that `lt (decr x) x` holds for all `x`? I ask only because this could plausibly be implemented by a non-well-behaved version (e.g. `Nat` subtraction) if a user is not careful. Maybe worth a comment what properties are needed of these? Eventually I think we would want `Lawful` versions.

joscoh (commented):
Can you add a test that demonstrates the negative case for the semantics: trying to call a function after it should be out of scope?

Strata/DL/Imperative/StmtSemanticsProps.lean
aqjune-aws: The benefit of `h1 : ReflTransT` defined as `Type` (and `h2` as well) is that the length of trace `.len` function can be defined. So conceptually it is really closer to a data structure that represents a trace of states.

Decomposing `seqT_reaches_terminal` will indeed return a trace `h1` (and `h2`) but the actual value of `h1` won't be visible because otherwise proof irrelevance won't hold, but when it is destructed inside a proof, the different values of `h1` can be behaved as observed because the proof will not depend on the actual value of `h1` anyway - did I address your concern..?

Strata/DL/Imperative/PureExpr.lean
aqjune-aws: decr wasn't necessary anymore - I erased it, thanks

Strata/DL/Imperative/StmtSemantics.lean
aqjune-aws: It doesn't help writing proofs, but the deterministic case which states that the measure has to evaluate to a numeric constant was necessary for LoopElim (otherwise the `assert (measure >= 0)` introduced by LoopElim will stuck). I thought about its corresponding update in the nondeterministic loop's semantics, and confirming that the measure being `.none` seemed like a good choice. :)

joscoh: Just to confirm, would evaluation get stuck if we allowed arbitrary `m` (instead of `.none`) here?

Strata/DL/Imperative/StmtSemanticsProps.lean
joscoh: I'm not sure I completely understand your explanation, but we can discuss later; it's not blocking.

joscoh (dismissed):

Strata/DL/Imperative/StmtSemantics.lean
aqjune-aws: Yes, it will stuck - it is analogous to the case when a measure that doesn't reduce to a numeric constant is given in the deterministic case.

Strata/DL/Imperative/StmtSemanticsProps.lean
aqjune-aws: Comparing `ReflTransT` with `ReflTrans`, the latter is truly an inductive predicate, but since it is `Prop` we can't define the 'length' of the specific relation chain. `ReflTransT` allows this (in Strata/DL/Util/Relations.lean):
```
@[simp] def ReflTransT.len : @ReflTransT A r a b → Nat
  | .refl _ => 0
  | .step _ _ _ _ rest => 1 + rest.len
```
The benefit of allowing the notion of size is that now we can use a strong induction. For example, if we want to write a proof that uses strong induction on `ReflTrans`, it is tricky because we can't define "what is less than what". For `ReflTransT`, this is now possible because we have a size of relation.

Strata/DL/Imperative/CmdSemantics.lean
```diff
@@ -166,6 +171,7 @@ inductive EvalCmd [HasFvar P] [HasBool P] [HasNot P] :
     ---
     EvalCmd δ σ (.init x _ .nondet _) σ' false
 
+
```
atomb: Was this added line intentional?

Strata/DL/Imperative/CmdSemanticsProps.lean
```diff
@@ -254,7 +254,7 @@ theorem InitStateUniqueResult
 /-! ### Assert / set commutation -/
 
 theorem eval_assert_store_cst
-  [HasFvar P] [HasBool P] [HasNot P]:
+  [HasFvar P] [HasBool P] [HasBoolOps P] [HasFvars P] [HasOps P] [HasInt P]:
```
atomb: Is `HasInt` still necessary in most of these, given the removal of measures from the semantics?

Strata/DL/Imperative/HasVars.lean
```diff
@@ -19,10 +19,16 @@ class HasVarsPure (P : PureExpr) (α : Type) where
 /-! # Imperative Variable Lookup : HasVarsImp -/
 
 class HasVarsImp (P : PureExpr) (α : Type) where
-  definedVars : α → List P.Ident
+  definedVars : α → Bool/-exclude vars not visible from outside?-/ → List P.Ident
```
atomb: I'm not sure I understand this comment. Is it a question that still needs to be resolved?

Strata/DL/Imperative/StmtSemantics.lean
atomb: This is a complex issue, and I think one that might benefit from a more detailed design discussion. Ultimately, we're interleaving the question of what the program does what what we want to prove about it in a way that I don't think should be necessary. But avoiding it could require rethinking a few things. At the moment I'm thinking I'd like us to do one of two things:
* Start with a contract -> assertion pass that turns things like contracts, invariants, and measures into explicit assertions, and follow that up with a pass that eliminates loops (or does other transformations) with that already in place. Notably, the first pass would actually change semantics -- it could lead a program to fail more often -- but I think that's ultimately what we want if we take this path.
* Have invariants and measures live in metadata and keep them in metadata after the loop elimination process. The verifier would then interpret the metadata when generating VCs.

aqjune-aws: Splitting a pass into two steps sounds like a good idea - it will make an non-overapproximating version of LoopElim available as well.

Looking at this diff again, the measure update in the `step_loop_nondet_exit` case looked slightly weird because another case `step_loop_nondet_enter` didn't need a corresponding update. I could confirm that this was a truly redundant update & reverted it.

Strata/DL/Imperative/CmdSemantics.lean
aqjune-aws: No, I removed the new line. :) Thanks

Strata/DL/Imperative/CmdSemanticsProps.lean
aqjune-aws: It was not necessary, and I removed `HasInt` and `HasFvars` from a few theorems.

Strata/DL/Imperative/HasVars.lean
aqjune-aws: I clarified the comment - it explains what the `Bool` argument is.

aqjune-aws:
Fixed the merge conflict and ready to get comments :)

atomb (approved):

joscoh (approved):

PR #1352, opened 2026-06-10
Enable calling procedures in contracts

keyboardDrummer:
## Functional changes
1. [Debugging] Improve the printing of Laurel if-then-else expressions
1. `EliminateReturnsInExpression` now runs for procedures as well, which enables more types of transparent bodies for procedures. To make it work for both functions and procedures, it was also necessary for the body of functions to be immediately wrapped in a return statement during parsing.
1. Allow calling procedures from contracts. Combined with the previous change this makes procedures strictly more powerful than functions
1. Let the transparency pass rewrite the bodies of assume statements so they don't assert anything.
1. Improve diagnostics related to contracts, using the correct verbiage "precondition" and "postcondition" instead of "assertion"
1. Generalized the `LaurelPass` concept so it works for all transformation between Laurel source and Core, not just the Laurel->Laurel transformation. This helps make the documentation more complete.

### Why let the transparency pass rewrite the bodies of assume statements so they don't assert anything?
After the contract pass, a call will look like `assert <preconditions>; call(..); assume <postconditions>`, where the body of the callee looks like `assume <preconditions>; <body>; assert <postconditions>`. If we now do either concrete execution, or we do inlining, then any assertions that occur inside the pre or postconditions will be asserted twice, because they occur once in an assert and once in an assume. By ignoring the assertions inside the assume, we prevent the duplication.

Whether you also want this behavior for assumptions that were created by users is something I'm not sure about. However, if we want we can let those behave differently. Right now I think we don't have enough data to decide what we want for user created assumptions, and they are AFAIK not yet used, so I think it's OK to change their behavior.

## Implementation
Add these passes:
- [New] EliminateReturnStatements: rewrite `return` to `exit` statements, needed for the next pass.
- [New] ContractPass: translate away pre and postconditions entirely by introducing assertion and assumptions at call sites and at procedure starts and ends
- [Updated] Lift assertions, assumptions and procedure calls when they occur in expressions. Note: the changes in this pass could have been extracted to a different PR to reduce the scope of this one, but I think that keeping them in this PR is most efficient from a developer time perspective.

## Follow-up work
- Remove the now obsolete functions from Laurel
- Create WF proofs for quantifier bodies
- Lift assumptions in expressions to axioms.
- In the transparency phase, if something has no asserts and only calls functions, only create a function and no procedure


Strata/Languages/Laurel/ContractPass.lean
```diff
@@ -0,0 +1,312 @@
...
+            let (tempDecls, tempRefs) := mkTempAssignments args' callee.text info.inputParams counter src
+            let callWithTemps : StmtExprMd := ⟨.Assign targets ⟨.StaticCall callee tempRefs, callSrc⟩, src⟩
+            let preCheck := mkPreChecks info isFunctional tempRefs src
+            let outputArgs := targets.filterMap fun t =>
+              match t.val with
+              | .Local name => some (mkMd (.Var (.Local name)))
+              | .Declare param => some (mkMd (.Var (.Local param.name)))
+              | _ => none
+            let postAssume := mkPostAssumes info tempRefs outputArgs src
+            set (counter + 1)
```
joscoh: The counter should be threaded through the recursion (line 207) or else there can be name collisions.

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
```diff
@@ -6,114 +6,75 @@
...
-  | ⟨.IfThenElse cond thenBr (some elseBr), source⟩ =>
-    ⟨.IfThenElse cond (lastStmtToExpr thenBr) (some (lastStmtToExpr elseBr)), source⟩
-  | _ => stmt
-  termination_by (sizeOf stmt, 0)
-  decreasing_by
-    all_goals (simp_all; term_by_mem)
+/-- Recurse into a statement, applying the guard-return rewrite inside blocks and branches.
+    Returns `Except DiagnosticModel StmtExprMd` so that unsupported statement forms produce
+    a diagnostic instead of panicking. -/
+partial def removeReturns (stmt : StmtExprMd) : Except DiagnosticModel StmtExprMd :=
```
joscoh: Any reason that this is `partial`?

Strata/Languages/Laurel/ContractPass.lean
```diff
@@ -0,0 +1,312 @@
...
+              | .StaticCall callee' args' =>
+                match contractInfoMap.get? callee'.text with
+                | some info' =>
+                  let counter' ← get
+                  let (stmts, counter'') := rewriteStaticCall counter' callee' args' info' e'.source
+                  set counter''
+                  return ⟨.Block stmts none, e'.source⟩
+                | none => return e'
+              | _ => return e'))
+            let (tempDecls, tempRefs) := mkTempAssignments args' callee.text info.inputParams counter src
```
fabiomadge: **Must-fix (correctness): nested contracted calls collide on temp names.** `counter` is read (L204) *before* recursing into the args, but nested contracted calls advance it (L214); L218 then builds this call's temps with the stale value and L227's `set (counter + 1)` discards the nested increments. `out := bar(bar(5))` (contracted `bar`) emits two `$bar$0$arg0` → `Duplicate definition`, valid program rejected (reproduced). Fix: read the counter *after* the arg recursion.

```suggestion
            let src := e.source
            -- Recurse into arguments first; nested contracted calls advance the
            -- shared counter, so we must read it *after* this loop (not before)
            -- to give this call site a distinct index.
            let args' ← args.mapM (mapStmtExprM (m := StateM Nat) (fun e' => do
              match e'.val with
              | .StaticCall callee' args' =>
                match contractInfoMap.get? callee'.text with
                | some info' =>
                  let counter' ← get
                  let (stmts, counter'') := rewriteStaticCall counter' callee' args' info' e'.source
                  set counter''
                  return ⟨.Block stmts none, e'.source⟩
                | none => return e'
              | _ => return e'))
            let counter ← get
            let (tempDecls, tempRefs) := mkTempAssignments args' callee.text info.inputParams counter src
```

<details><summary>Self-contained reproducer (valid program, rejected on this branch)</summary>

```
procedure bar(x: int) returns (r: int)
  requires x > 0
  opaque
  ensures r > 0
{
  r := x + 1
};

procedure caller() returns (out: int)
  opaque
{
  out := bar(bar(5))
};
```
On this branch: `Internal error: resolution after 'LiftImperativeExpressionsInCore' ... Duplicate definition '$bar$0$arg0' is already defined in this scope`. With the fix above: translates clean (verified, full StrataTest + StrataPythonTest green).
</details>

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
```diff
@@ -6,114 +6,75 @@
...
+  match stmt.val with
+  | .Return (some expr) => .ok expr
```
fabiomadge: **Must-fix (correctness): explicit `return` in a `function` is rejected.** The parser wraps function bodies in `Return` (ConcreteToAbstract L523), so `function f(x){ return x }` becomes `Return(Block[Return e])`; only the outer layer is peeled, and `EliminateValuesInReturns` turns the survivor into a destructive assignment → "destructive assignments are not supported in transparent bodies or contracts" (reproduced; no existing test covers `return` in a `function`). Fix: recurse only when the wrapped value actually contains a `return` (a blind recurse would break pure `if`-expression branches).

```suggestion
/-- Whether `stmt` contains a `return` statement in a statement position
    (block element or if-branch). Used to decide whether the value wrapped by
    the parser's implicit function-body `Return` still needs return-merging. -/
partial def containsReturnStmt (stmt : StmtExprMd) : Bool :=
  match stmt.val with
  | .Return _ => true
  | .Block stmts _ => stmts.attach.any fun ⟨s, _⟩ => containsReturnStmt s
  | .IfThenElse _ thenBr (some elseBr) => containsReturnStmt thenBr || containsReturnStmt elseBr
  | .IfThenElse _ thenBr none => containsReturnStmt thenBr
  | _ => false

mutual

/-- Recurse into a statement, applying the guard-return rewrite inside blocks and branches.
    Returns `Except DiagnosticModel StmtExprMd` so that unsupported statement forms produce
    a diagnostic instead of panicking. -/
partial def removeReturns (stmt : StmtExprMd) : Except DiagnosticModel StmtExprMd :=
  match stmt.val with
  -- A function body is wrapped in a `Return` during parsing (ConcreteToAbstract).
  -- If the wrapped value still contains explicit `return` statements (because the
  -- user wrote them), recurse to merge them into the single outer expression.
  -- Otherwise the value is already a pure expression and is returned directly,
  -- matching the original behaviour for bodies without explicit returns.
  | .Return (some expr) =>
    if containsReturnStmt expr then removeReturns expr else .ok expr
```

<details><summary>Self-contained reproducer (+ control isolating the cause)</summary>

Rejected on this branch with `destructive assignments are not supported in transparent bodies or contracts`:
```
function inc(x: int): int
{
  return x + 1
};

procedure useInc() opaque
{
  var a: int := inc(41);
  assert a == 42
};
```
Control — identical but bare body (no `return`) — verifies clean, isolating the cause to the explicit `return`:
```
function inc(x: int): int
{
  x + 1
};
```
With the fix above, the `return` form also verifies clean (verified, both suites green).
</details>

Strata/Languages/Laurel/ContractPass.lean
```diff
@@ -0,0 +1,312 @@
...
+    (inputParams : List Parameter) (outputParams : List Parameter)
+    (condition : Condition) : Procedure :=
+  let allParams := inputParams ++ outputParams
+  { name := mkId name
+    inputs := allParams
+    outputs := [⟨mkId "$result", { val := .TBool, source := none }⟩]
+    preconditions := []
+    decreases := none
+    isFunctional := true
+    body := .Transparent condition.condition }
```
fabiomadge: De-dup: `mkPostConditionProc` is `mkConditionProc` with `params := inputParams ++ outputParams`. ⚠ This suggestion deletes `mkPostConditionProc`, so you must ALSO update its call site (HEAD L283-284) — otherwise the build breaks — to:
```lean
      mkConditionProc (postCondProcName proc.name.text i) (proc.inputs ++ proc.outputs) c
```
The merged definition:

```suggestion
/-- Build a helper function for a single condition over the given parameters.
    Preconditions pass `proc.inputs`; postconditions pass `proc.inputs ++ proc.outputs`. -/
private def mkConditionProc (name : String) (params : List Parameter)
    (condition : Condition) : Procedure :=
  { name := mkId name
    inputs := params
    outputs := [⟨mkId "$result", { val := .TBool, source := none }⟩]
    preconditions := []
    decreases := none
    isFunctional := true
    body := .Transparent condition.condition }
```

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -148,22 +142,53 @@ private def computeType (expr : StmtExprMd) : LiftM HighTypeMd := do
...
   termination_by expr
   decreasing_by
-    all_goals ((try cases x); simp_all; try term_by_mem)
+    all_goals (try cases x)
+    all_goals (try simp_all)
+    all_goals (try have := Condition.sizeOf_condition_lt ‹_›)
+    all_goals (try term_by_mem)
+    all_goals omega
 
 /-- Like containsAssignment but does NOT recurse into Blocks (treats them as opaque).
```
fabiomadge: Dead code (cleanup): `containsBareAssignment` (here) and `containsImperativeCall` (just below, through ~L230) are unreachable on this branch (grep-confirmed, repo-wide). `containsBareAssignment` was orphaned by this PR — it removed its only callers (the old assert/assume guards at main2 L411/L420); `containsImperativeCall` was already dead on `origin/main2`. Safe to delete both.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -212,25 +269,40 @@ def translateWithLaurel (options : LaurelTranslateOptions) (program : Program)
...
+      unorderedCore := uc'
+      fnModel := m'
+    emit pass.name "unorderedCoreWithLaurelTypes.st" unorderedCore
+
+  let coreWithLaurelTypes := orderFunctionsAndProcedures unorderedCore
+  -- This early return is a simple way to protect against duplicative errors. Without this return,
+  -- resolution errors reported by Laurel would also be reported by Core.
+  -- There might be better solution that allows getting some resolution errors from Laurel and some verification errors from Core,
+  -- but that would need more consideration.
+  if ! passDiags.isEmpty then
```
fabiomadge: Cleanup: `passDiags` is bound by `let` (L272) and never reassigned, and the check right after `runLaurelPasses` already returned when it was non-empty — so this second `if ! passDiags.isEmpty` is always false and its `else` is unconditional. The `if/else` can be dropped and the body de-indented one level.

Strata/Languages/Laurel/EliminateValuesInReturns.lean
```diff
@@ -55,11 +56,19 @@ def hasValuedReturn (stmt : StmtExprMd) : Bool :=
 /-- Apply value-return elimination to a single procedure. Only applies to
     non-functional procedures with exactly one output parameter.
     Emits an error if a valued return is used with multiple output parameters. -/
```
fabiomadge: Stale doc: says "Only applies to non-functional procedures" but there is no `isFunctional` guard — it transforms any single-output procedure (behaviour is fine; the comment is wrong).

```suggestion
/-- Apply value-return elimination to a single procedure. Rewrites `return expr`
    into `outParam := expr; return` for any procedure with exactly one output
    parameter (functional and non-functional alike).
    Emits an error if a valued return is used with zero or multiple output parameters. -/
```

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -263,7 +284,7 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
           if hasSubst then
             pure (⟨.Var (.Local (← getSubst param.name)), source⟩)
           else
-            return expr
+            pure (⟨.Var (.Local param.name), source⟩)
         | _ =>
           dbg_trace "Strata bug: non-identifier targets should have been removed before the lift expression phase";
```
fabiomadge, line 270: Pre-existing (not introduced here, exists on `origin/main2`): this `dbg_trace` + silent `return expr` reports a "should be impossible" state to stderr rather than as a diagnostic. Hard to fix cleanly because `LiftM` is a pure `StateM` with no diagnostic channel — noting for a future pass, not asking for it in this PR.

fabiomadge (changes requested):
Reviewed for correctness + soundness; ran candidate programs through the pipeline (cvc5) and diffed behaviour against `origin/main2`. Genuine improvement, and the new contract encoding **held up under every soundness probe I could construct** — but two valid programs are rejected (must-fix), plus a systemic-but-pre-existing namespace issue and some quality debt.

**Soundness (sampled, not proven):** I ran 16 programs through the verifier (cvc5). Every should-fail case that exercises the new encoding was correctly rejected — precondition checks, postcondition (re)establishment, partial-function `$asFunction` use outside its precondition domain, chained calls — and the should-pass cases (including a correct modular proof) stayed clean. The lone should-fail case that verified anyway is a *bodiless* procedure with a self-contradictory `ensures` (caller assumes `false`); I confirmed it **also reproduces on `origin/main2`**, so it's the standard "unsound spec ⇒ garbage in/out" of modular verifiers — pre-existing, out of scope.

**Must-fix (reject valid programs) — fixes proposed as inline suggestions:**
1. Nested contracted calls → colliding temps (`bar(bar(5))`).
2. Explicit `return` in a `function` fails — breaks the second item in the PR's "Functional changes" list (functions wrapped in a return during parsing).
I verified each fix locally (applying it makes the failing program translate, and the full StrataTest + StrataPythonTest suites stay green; reverting it reproduces the failure). The two trigger programs are quoted in the inline comments so you can drop them into the `Fundamentals` test suite as regression tests.

**Low (pre-existing class, not a PR blocker):** compiler `$`-names can collide with same-named user identifiers — reproduced for `foo$pre0`, `$cndtn_0` (hard `Duplicate definition`) and `$return` (silent, no diagnostic). But this is a **codebase-wide convention, not a regression**: every generated name is `$`-prefixed (`$tmp`, `$hole_`, `$cndtn_`, …), uniqueness is by counter with no guard anywhere, and the Python front-end itself emits `$`-names (`$body`, `$if_cond_`, …) into the same namespace — so a naive "reject `$`" guard would break the front-end. Worth a dedicated hardening pass, orthogonal to this PR. Flagging, not blocking.

**Quality (non-blocking):** a mix of inline code-suggestions (dead code, de-dups, stale comments) and — where GitHub won't let me anchor to unchanged lines — a checklist below. I verified these locally — applied, they build with both suites green (526 + 227).

---

**Quality follow-ups (couldn't inline — they target unchanged lines GitHub won't let me anchor a comment to; each verified locally; applied, they build with both suites green):**
- `LaurelToCoreTranslator.lean` ~L580-614 — **dead code:** `translateInvokeOnAxiom` + its local `buildQuants` are unreachable (axioms now translated ~L739-741; grep-confirmed no callers). Delete the block.
- `LaurelToCoreTranslator.lean` ~L666-680 — **de-dup:** the `{ result := <e>; exit "$return" } $return → <e>` unwrap is copy-pasted in the Transparent and Opaque arms. Extract `private def unwrapReturnBlock (b : StmtExprMd) : StmtExprMd` and call it in both (also gives one place to harden the brittle exact-shape match).
- `LaurelToCoreTranslator.lean` ~L741 — **nit:** multiple `axioms` on one proc would all be emitted as `Core.Decl.ax` named `invokeOn_{name}` (duplicate). Moot today (ContractPass produces a single-element list), worth uniquifying when touched.
- `MapStmtExpr.lean` — **largest duplication:** `mapStmtExprM` is exactly `mapStmtExprPrePostM` with no pre-filter; it can become `def mapStmtExprM f e := mapStmtExprPrePostM (fun _ => pure none) f e` (placed after `mapStmtExprPrePostM`), deleting the ~65-line duplicate match + its termination proof with no new obligations. `mapStmtExprFlattenM` is genuinely different (list-producing, Block-flattening) — leave it.
- `MapStmtExpr.lean` — **proof duplication:** the three byte-identical `decreasing_by` blocks can be one `local syntax`+`macro_rules` tactic (`map_stmt_expr_decreasing <var>`), invoked as `decreasing_by map_stmt_expr_decreasing expr` (Flatten site uses `e`). NB: the inline `macro "…" x:ident` form fails to elaborate in `decreasing_by`; the `syntax`+`macro_rules` form builds (matches the existing `term_by_mem` precedent).

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -192,13 +203,59 @@ private def runLaurelPasses (options : LaurelTranslateOptions)
...
+  /-- Human-readable name, used for profiling and file emission. -/
+  name : String
+  /-- Whether `resolveUnorderedCore` should be run after the pass. -/
+  needsResolves : Bool := false
+  /-- The pass action. -/
+  run : UnorderedCoreWithLaurelTypes → SemanticModel → UnorderedCoreWithLaurelTypes
+
+/-- The ordered sequence of passes on the unordered Core representation. -/
+private def corePipeline : Array CorePass := #[
+  -- { name := "EliminateMultipleOutputs"
```
joscoh: Nit: should not add commented out code

Strata/Languages/Laurel/ContractPass.lean
keyboardDrummer: Thanks! Addressed

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
keyboardDrummer: Not being able to have returns in a function is intended behavior, although the error message is bad and should be improved. However, we will remove functions entirely in the PR following this one, so I suggest we only improve error messages for procedures. After that PR, and merging the type checking PR, I intend to have a PR for improving the clarity of diagnostics.

Strata/Languages/Laurel/EliminateReturnsInExpression.lean
keyboardDrummer: Nope, fixed.

Strata/Languages/Laurel/DesugarShortCircuit.lean
```diff
@@ -24,32 +24,33 @@ namespace Strata.Laurel
...
+        ⟨.IfThenElse a (wrap (.LiteralBool true)) (some b), source⟩
       else expr
     | _, _ => expr
   | _ => expr
 
 /-- Desugar short-circuit operators in a program. -/
-def desugarShortCircuit (model : SemanticModel) (program : Program) : Program :=
-  mapProgram (mapStmtExpr (desugarShortCircuitNode model)) program
+def desugarShortCircuit (program : Program) : Program :=
+  let imperativeCallees := program.staticProcedures.map (fun p => p.name.text)
```
fabiomadge: **Now treats *all* calls as imperative, contradicting the module doc.** This populates `imperativeCallees` from every `staticProcedures` name with no `isFunctional` filter, but the module spec (line 7) still says "non-functional procedure calls", and the old `containsImperativeCall` filtered on `!proc.isFunctional`. Net effect is benign — the generated `if c then func(1) else false` short-circuits identically, and the lift pass operates on `coreProcedures` (the non-functional set) so the functional call isn't hoisted; VCs unchanged, T15 green. But the code now contradicts its own doc. Either restore the filter or update the comment:
```lean
let imperativeCallees := (program.staticProcedures.filter (!·.isFunctional)).map (·.name.text)
```

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -300,49 +287,61 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
         prepend (⟨.IfThenElse seqCond thenBlock seqElse, source⟩)
-        prepend ⟨.Var (.Declare ⟨condVar, condType⟩), source⟩
-        return ⟨.Var (.Local condVar), source⟩
+        if needsCondVar then
+          prepend ⟨.Var (.Declare ⟨condVar, condType⟩), source ⟩
+          modify fun s => { s with prependedStmts := condPrepends ++ s.prependedStmts }
+          return ⟨.Var (.Local condVar), source⟩
+        else
+          modify fun s => { s with prependedStmts := condPrepends ++ s.prependedStmts }
+          return default
```
fabiomadge: **`return default` for the void if-then-else branch.** `default` for `StmtExprMd` is a `.Hole`. It's unused today (the if-then-else is lifted to statements via prepend, so this return value is discarded), but returning a hole as a sentinel is easy to misread later. A one-line comment that the value is unused in the TVoid case — or an explicit node — would make the intent clear.

Strata/Languages/Laurel/EliminateReturnStatements.lean
```diff
@@ -0,0 +1,73 @@
...
+  | .Transparent body =>
+    let body' := replaceReturn proc.outputs body
+    let wrapped := match body'.val with
+      | .Block stmts none => ⟨.Block stmts (some returnLabel), body'.source⟩
+      | _ => mkMd (.Block [body'] (some returnLabel))
+    { proc with body := .Transparent wrapped }
+  | _ => proc
+where
+
+  mkMd (e : StmtExpr) : StmtExprMd := { val := e, source := proc.name.source }
```
fabiomadge: **Synthetic nodes stamped with `proc.name.source`.** `mkMd`/`mkVarMd` assign `proc.name.source` to every generated exit/assign/block, so a diagnostic on lifted return-handling points at the procedure declaration rather than the originating `return`. Where a synthetic node derives from a specific return, prefer that statement's own `e.source`.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -130,20 +135,26 @@ private def laurelPipeline : Array LaurelPass := #[
...
-  { name := "LiftExpressionAssignments"
-    run := fun p m =>
-      (liftExpressionAssignments m p, [], {}) },
-  { name := "EliminateReturns"
-    needsResolves := true
-    run := fun p _m =>
-      (eliminateReturnsInExpressionTransform p, [], {}) },
+    run := fun p _ =>
+      (desugarShortCircuit p, [], {}) },
+  -- { name := "LiftExpressionAssignments"
```
fabiomadge: **Stale commented-out pass.** The commented `LiftExpressionAssignments` registration is dead — the function is live at line 227. Delete it.

fabiomadge (dismissed):
**Approving** — supersedes my earlier `CHANGES_REQUESTED`: the must-fix (contract-pass counter collision) is fixed and re-verified, and the function-return concern is an accepted design choice with a tracked diagnostics follow-up.

This is a second pass over the files my first review didn't cover (the lifting/transparency/return reworks — ~half the PR's own diff). **No blockers, no majors** — the correctness core is solid. A few minors/nits inline, none gating; I built the branch and ran probes, including for items I then refuted (no `$asFunction` name collision, no prepend-ordering bug, no `$cp_` scope escape — all checked empirically).

One body-level question: `errorVoid` (added to `Core/Factory.lean` `KnownLTys`) has no producer or consumer — the only other reference is the `TypeDecl` snapshot echoing the registry. Intended placeholder for void contract-calls, or leftover? Harmless either way, but worth resolving.

keyboardDrummer:
@keyboardDrummer-bot can you merge in the changes from main2?

Strata/Languages/Laurel/ContractPass.lean
```diff
@@ -0,0 +1,317 @@
...
+    let posts' ← posts.mapM (·.mapM rw)
+    let impl' ← impl.mapM rw
+    let mods' ← mods.mapM rw
+    return { proc with body := Body.Opaque posts' impl' mods' }
+  | _ => return proc
+
+/-- Build an axiom expression from `invokeOn` trigger and ensures clauses.
+    Produces `∀ p1, ∀ p2, ..., ∀ pn :: { trigger } (ensures1 && ensures2 && ...)`.
+    The trigger controls when the SMT solver instantiates the axiom. -/
+private def mkInvokeOnAxiom (params : List Parameter) (trigger : StmtExprMd)
```
fabiomadge, line 344: **Two fixes here (Blocker 1 + Issue 3).** This axiom is built from `ensures` only and quantified over `params` (= `proc.inputs`) only — so it's both unsound (no precondition guard) and crashes on an output-referencing ensures (`r` unbound → `Cannot find this fvar`).

Guard the body as `preconds => ensures`, and pass `proc.preconditions` at the call site (`:293`):

```
private def conjoin (conds : List Condition) : Option StmtExprMd :=
  match conds.map (·.condition) with
  | [] => none
  | e :: rest => some (rest.foldl (fun acc x => mkMd (.PrimitiveOp .And [acc, x])) e)

private def mkInvokeOnAxiom (params : List Parameter) (trigger : StmtExprMd)
    (preconds : List Condition) (postconds : List Condition) : StmtExprMd :=
  let ensures := (conjoin postconds).getD (mkMd (.LiteralBool true))
  let body := match conjoin preconds with
    | some pre => mkMd (.PrimitiveOp .Implies [pre, ensures])
    | none => ensures
  params.foldr (init := (body, true)) (fun p (acc, isInnermost) =>
    let trig := if isInnermost then some trigger else none
    (mkMd (.Quantifier .Forall p trig acc), false)) |>.1
```

For the crash (Issue 3): reject output-referencing ensures in `lowerContracts` rather than emitting an unbound axiom — return `Program × List DiagnosticModel` and surface it from `contractPass.run`:

```
private def invokeOnOutputRefError (proc : Procedure) : Option DiagnosticModel :=
  if proc.invokeOn.isNone then none else
    let postconds := getPostconditions proc.body
    let referenced := proc.outputs.filterMap (fun out =>
      if postconds.any (fun c => exprMentions out.name.text c.condition)
      then some out.name.text else none)
    match referenced with
    | [] => none
    | names => some (diagnosticFromSource proc.name.source
        s!"'invokeOn' procedure '{proc.name.text}' has an ensures referencing its output(s) ({String.intercalate ", " names}); the auto-invocation axiom is quantified over inputs only."
        DiagnosticType.UserError)
```

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -301,49 +289,62 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
+
         -- Lift the entire if-then-else. Introduce a fresh variable for the result.
         let condVar ← freshCondVar
-        let seqCond ← transformExpr cond
         -- Save outer state
         let savedSubst := (← get).subst
         let savedPrepends := (← get).prependedStmts
+
+        let seqCond ← transformExpr cond
+        let condPrepends := (← get).prependedStmts
```
fabiomadge: **Duplicate prepends: clear before lifting the condition.** Snapshotting `prependedStmts` without clearing means `transformExpr cond` accumulates onto the outer prepends, so `condPrepends` already contains `savedPrepends` and they get emitted twice below (including a duplicate `$cp_*` decl → `Duplicate definition` resolution failure). The `StaticCall` handler at `:274` avoids this with `takePrepends`. Mirror it:

```suggestion
        let savedSubst := (← get).subst
        let savedPrepends ← takePrepends

        let seqCond ← transformExpr cond
        let condPrepends ← takePrepends
```

fabiomadge (changes requested):
Re-reviewed after the rework (head `10779790c`): **one soundness bug, one that rejects valid programs, one crash**, all in the new contract machinery, all verified by built repros (suite stays green — 534 jobs — with my fixes). Requesting changes.

Honest correction: all three were already present at the commit I approved earlier — not regressions. I missed them by not probing `requires`+`invokeOn`, output-referencing `invokeOn` ensures, or a call in an if-condition. Apologies. And on intent: B2 is an internal error, clearly not by-design; B1 and Issue 3 match the `invokeOn` doc on `LaurelAST.lean` ("the ensures clause universally quantified over the procedure's inputs"), so you may read them as intended — my point is that *documented* behavior is itself unsound once `requires` is present, so the doc needs fixing alongside the code.

## 🔴 Blocker 1 — `invokeOn` axiom drops the precondition guard (unsound)

`mkInvokeOnAxiom` (`ContractPass.lean:293`) builds the axiom from `ensures` only — preconditions are never threaded in. But the body is verified as `assume preconditions; … ; assert postconditions`, so `ensures` only holds *under* the preconditions. Verified (z3): `requires x>0 invokeOn T(x) ensures x>0` emits `∀x :: {T(x)} x>0`; a caller doing `assume T(y); assert y>0` then passes — invalid program accepted. (Controls: drop the proc → `assert y>0` fails; `assert false` still fails.)

Fix (your call): guard as `preconds => ensures` (inline — matches what the body proves), or reject `requires`+`invokeOn`. Either way, update the doc.

## 🔴 Blocker 2 — if-then-else condition lifting duplicates prepends (rejects valid programs)

`LiftImperativeExpressions.lean:316` snapshots `savedPrepends` without clearing before transforming the condition, so `condPrepends` re-includes them and they're emitted twice (incl. a duplicate `$cp_N` decl). `StaticCall` (`:274`) gets this right with `takePrepends`. Verified: a call in both the condition and branches of an if-expression aborts with `Internal error: … Duplicate definition '$cp_N'`. Fix inline: `takePrepends` instead of snapshot.

## 🟠 Issue 3 (crash, not a merge-blocker) — output-referencing `invokeOn` ensures

Same axiom, quantified over inputs only: an `invokeOn` proc returning a value with `ensures r > 0` leaves `r` unbound → crashes with `Cannot find this fvar in the context! r`. The doc scopes `invokeOn` to inputs, so the limitation may be intended — but it should reject with a `UserError`, not crash (fix inline, with Blocker 1). Supporting output-ensures (bind to `$asFunction`) hits termination-checking and multi-output issues — a design call for its own issue.

fabiomadge:
Confirmed on `b30f4cbac` (built + ran repros):

- **Blocker 1** — fixed. The exploit (`assume T(y); assert y>0`) now fails, and a caller that establishes the precondition still proves the postcondition.
- **Issue 3** — fixed. Output-referencing ensures now gives the `UserError` instead of crashing; skipping axiom generation is a nice touch.

**Blocker 2** (if-then-else prepend duplication, `LiftImperativeExpressions.lean:316`) is still open — that push only touched `ContractPass`. Still repros: a call in both the condition and branches of an if-expression aborts with `Internal error … Duplicate definition '$cp_N'`. The inline `suggestion` is ready to apply. Keeping the review at request-changes until that lands.

fabiomadge (dismissed):
All three findings are fixed on `6dd545a` — re-verified by building the repros:

- **Blocker 1** (unsound `invokeOn` axiom): `exploit`'s `assert y > 0` now fails — the precondition guard holds.
- **Blocker 2** (if-then-else prepend duplication): the call-in-both-condition-and-branches repro compiles cleanly, no `Duplicate definition '$cp_N'`. (Nice — you also applied `takePrepends` to the second if-then-else site.)
- **Issue 3** (output-referencing `invokeOn`): clean `UserError`, no `Cannot find fvar` crash.

Full suite green (534 jobs). Thanks for the quick turnaround — LGTM.

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
```diff
@@ -34,7 +34,7 @@ info: procedure assertInBlockExpr()
   opaque
 {
   var x: int := 0;
-  assert x == 0;
+  assert $x_0 == 0;
```
joscoh: Why is this referenced before it is declared?

Strata/Languages/Laurel/LaurelToCoreSchemaPass.lean
```diff
@@ -574,63 +571,27 @@ def translateProcedure (proc : Procedure) : TranslateM Core.Procedure := do
...
   verifyOptions : Core.VerifyOptions := .default
 
 instance : Inhabited LaurelVerifyOptions where
   default := {}
 
+/-- Unwrap the pattern produced by EliminateValuesInReturns + EliminateReturnStatements:
+    `{ result := <expr>; exit "$return" } $return` → `<expr>` -/
+private def unwrapReturnBlock (b : StmtExprMd) : StmtExprMd :=
+  match b.val with
+  | .Block [⟨.Assign [⟨.Local _, _⟩] value, _⟩, ⟨.Exit "$return", _⟩] (some "$return") => value
```
joscoh, line 678: 1. Is it possible for procedures to end with any other sequence of statements? (e.g. does Laurel allow multi-output procedures?)
2. It would be better to use `returnLabel` here.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -180,38 +186,44 @@ def translateWithLaurel (options : LaurelTranslateOptions) (program : Program)
...
-      emit "Core" "core.st" coreProgramOption.get!
-    let coreProgramOption :=
-      if translateState.coreDiagnostics.isEmpty then coreProgramOption else none
-    return (coreProgramOption, allDiagnostics, program, stats)
+  let (program, model, passDiags, stats) ← runLaurelPasses options pctx program
+  -- This early return is a simple way to protect against duplicative errors. Without this return,
+  -- resolution errors reported by Laurel would also be reported by Core.
+  -- There might be better solution that allows getting some resolution errors from Laurel and some verification errors from Core,
+  -- but that would need more consideration.
+  if ! passDiags.isEmpty then
```
joscoh: Just wondering: does `passDiags` always consist only of errors that should block verification, or could there be non-blocking warnings as well?

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -103,24 +106,16 @@ private def onlyKeepSideEffectStmtsAndLast (stmts : List StmtExprMd) : LiftM (Li
...
-          -- Otherwise we could keep them in the block
-          prepend s
-          pure []
-      | .Assume _ => do
-          -- Hack to work around Core not supporting assume expressions
-          -- Otherwise we could keep them in the block
-          prepend s
-          pure []
+          pure [s]
+      -- | .Assert _ => do
```
joscoh: Should remove commented-out code

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -282,11 +265,16 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
-    if model.isFunction callee then
+    let imperativeCallees := (← get).imperativeCallees
+    if !imperativeCallees.contains callee.text then
+      let seqArgs ← args.reverse.mapM transformExpr
+      let seqCall := ⟨.StaticCall callee seqArgs.reverse, source⟩
       return seqCall
     else
+      let startingPrepend ← takePrepends
+      let seqArgs ← args.reverse.mapM transformExpr
+      let argsPepends ← takePrepends
```
joscoh: Nit: `argsPrepends`

joscoh (commented):
AI-suggested test cases to add: recursive procedures with contracts, multiple postconditions with multiple return paths, transparent procedure bodies with ensures clauses

Strata/Languages/Laurel/LaurelToCoreSchemaPass.lean
keyboardDrummer: > It would be better to use returnLabel here.

I haven't found a way to refer to variables in pattern matching positions without creating a fresh variable instead of a reference.

keyboardDrummer: > Is it possible for procedures to end with any other sequence of statements? (e.g. does Laurel allow multi-output procedures?)

Yes a procedure does not need to return and yes there are multi-output procedures.

I think this `unwrapReturnBlock` code is temporary. Ideally we find a way to translate arbitrary control flow to a functional program and have that as a pass on functions before running LaurelToCoreSchemaPass.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
keyboardDrummer: Right now in theory there can be non-blocking warnings, although I'm not sure any exist yet. Updated this to `if passDiags.any (·.type != .Warning) then`

StrataTest/Languages/Laurel/LiftExpressionAssignmentsTest.lean
keyboardDrummer: That's a bug. Lifted statements should not refer to snapshots, so the assert should refer to `x` not `$x_0`. Fixed it and updated the test. 

After I finish the work on merging function and procedures (1 more PR after this one), I will have a PR just to add more tests. In particular the lifting pass has turned out to be rather buggy. I will make sure that all the passes only refer to a limited set of AST nodes, using the generic traversal functions to handle nodes that don't relate to that pass.

keyboardDrummer:
> AI-suggested test cases to add: 

In the follow-up PR, all the `functions` in Laurel tests are converted to procedures, so there will a few more test-cases coming in.

> recursive procedures with contracts

Since this PR is already huge, is it okay if I add a TODO item on my list to add this?

> multiple postconditions with multiple return paths

Since this PR is already huge, is it okay if I add a TODO item on my list to add this?

> transparent procedure bodies with ensures clauses

That's syntactically not allowed

Strata/Languages/Laurel/LaurelToCoreSchemaPass.lean
joscoh: > I haven't found a way to refer to variables in pattern matching positions without creating a fresh variable instead of a reference.

You can use `@[match_pattern]` I believe.

> I think this unwrapReturnBlock code is temporary. Ideally we find a way to translate arbitrary control flow to a functional program and have that as a pass on functions before running LaurelToCoreSchemaPass.

So right now, are multi-output procedures expected to fail with an error or silently succeed with perhaps unexpected behavior? I would add a comment either way.

keyboardDrummer: > You can use @[match_pattern] I believe.

I will apply it in the follow-up PR.

> So right now, are multi-output procedures expected to fail with an error or silently succeed with perhaps unexpected behavior? I would add a comment either way.

Fail with an error. There's only one test for it in the suite now:

```
procedure transparentBodyMultipleOuts() returns (q: int, r: int)
{
  assert true;
  q := 3;
  r := 2
//^^^^^^ error: ending a transparent body with a Assign statement is not supported
}
```

There should be more tests. 

> I would add a comment either way.

I'll add a note to add it in the follow-up PR for this one

keyboardDrummer: image

"unexpected token '@['; expected '⟩'"
Any other suggestions? I have asked an LLM in the past on how to do this and didn't find any method.

joscoh: If this is referring to the `@[match_pattern]`, that is an annotation that goes before the definition of `returnLabel`. You can find many such examples throughout Strata, e.g. `LMonoTy.bool` in `LTy.lean`.

joscoh (dismissed):

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -393,10 +389,89 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
...
+      let argPrepends ← takePrepends
+      modify fun s => { s with prependedStmts := argPrepends ++ [expr] ++ prepends }
+      default
+
+  | .Assert cond =>
+      let prepends ← takePrepends
+      _ ← transformExpr cond.condition
+      let argPrepends ← takePrepends
+      modify fun s => { s with prependedStmts := argPrepends ++ [expr] ++ prepends }
+      default
```
fabiomadge, line 419: **Regression (incompleteness):** this arm discards the transformed condition and re-emits the original `expr`, so a liftable subterm in an expression-position `assert` is emitted twice and Core then rejects valid code:

```laurel
// foo() opaque ensures r == 5
var y := { assert foo() == 5; 1 };
//   => "calls to procedures are not supported in functions or contracts"
//    + "assumes are not YET supported in functions or contracts"
var y := { assert <??>; 1 };
//   => "holes should have been eliminated before translation"
```

A plain revert fixes both but re-breaks your golden `assert x == 0` (the use-before-def of `$x_0`). The fix below clears `subst` around the lift instead — the assert sits before any right-snapshot, so the original variable name is the correct one, and the nested call/hole still gets lifted. Verified on a clean build: case 1 -> `ok`; case 2 -> `assertion does not hold` (hole now correctly havoc'd, matching statement position); golden + LiftHoles/MapStmtExpr/IncrDecrLift + Fundamentals (T10/T15/T16/T8b/T9) stay green. The two cases above make good regression tests.

```suggestion
  | .Assert cond =>
      let prepends ← takePrepends
      let savedSubst := (← get).subst
      modify fun s => { s with subst := [] }
      let seqCond ← transformExpr cond.condition
      let argPrepends ← takePrepends
      modify fun s => { s with subst := savedSubst, prependedStmts := argPrepends ++ [⟨.Assert { cond with condition := seqCond }, source⟩] ++ prepends }
      default
```

Strata/Languages/Laurel/LiftImperativeExpressions.lean
```diff
@@ -393,10 +389,89 @@ def transformExpr (expr : StmtExprMd) : LiftM StmtExprMd := do
       else
         return expr
 
+  | .Assume cond =>
+      let prepends ← takePrepends
+      _ ← transformExpr cond
+      let argPrepends ← takePrepends
+      modify fun s => { s with prependedStmts := argPrepends ++ [expr] ++ prepends }
+      default
```
fabiomadge, line 413: Same bug as the `.Assert` arm below — clear `subst` around the lift so the nested call/hole is lifted but the condition isn't mis-substituted:

```suggestion
  | .Assume cond =>
      let prepends ← takePrepends
      let savedSubst := (← get).subst
      modify fun s => { s with subst := [] }
      let seqCond ← transformExpr cond
      let argPrepends ← takePrepends
      modify fun s => { s with subst := savedSubst, prependedStmts := argPrepends ++ [⟨.Assume seqCond, source⟩] ++ prepends }
      default
```

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -183,47 +190,55 @@ def translateWithLaurel (options : LaurelTranslateOptions) (program : Program)
...
+      for proc in ct.instanceProcedures do
+        passDiags := passDiags ++ [diagnosticFromSource proc.name.source
+          s!"Instance procedure '{proc.name.text}' on composite type '{ct.name.text}' was not lifted before Core translation (pipeline-ordering bug)"
+          DiagnosticType.StrataBug]
+
+  -- This early return is a simple way to protect against duplicative errors. Without this return,
+  -- resolution errors reported by Laurel would also be reported by Core.
+  -- There might be better solution that allows getting some resolution errors from Laurel and some verification errors from Core,
+  -- but that would need more consideration.
+  if passDiags.any (·.type != .Warning) then
```
fabiomadge, line 207: Non-blocking: this is a no-op today. No Laurel pass or resolution emits `.Warning` (they emit only `UserError`/`StrataBug`/`NotYetImplemented`; the sole `.Warning` producer is `Core/Verifier.lean`, downstream of this gate), so `passDiags.any (·.type != .Warning)` is currently equivalent to the old `!passDiags.isEmpty`. Sound as a forward-looking guard, just doesn't change behavior yet.

fabiomadge (commented):
Re-reviewed the delta since my approval at `6dd545a` (head now `799f16789`). The author commits since then are small; the rest of the range is unrelated merged PRs (#1328/#1361/#1320).

One regression worth fixing before merge (rejects valid programs — incompleteness, not unsound), plus a no-op note on the pipeline gate. Both my earlier open threads are resolved by the current code. Each finding was reproduced through the full pipeline with controls on a clean rebuild (revert flips it; the revert also re-breaks the golden test, confirming the commit fixed a real bug too).

Strata/Languages/Laurel/LiftImperativeExpressions.lean
keyboardDrummer: I ran into this in the follow-up PR. I've pulled the fix and some others and some additional tests from that one.

Strata/Languages/Laurel/LiftImperativeExpressions.lean
keyboardDrummer: I ran into this in the follow-up PR. I've pulled the fix and some others and some additional tests from that one.

Strata/Languages/Laurel/DesugarShortCircuit.lean
fabiomadge: Addressed — now filters `(!·.isFunctional)` (`:53`), doc matches (`:17`).

Strata/Languages/Laurel/LiftImperativeExpressions.lean
fabiomadge: Still present (`.Hole` at the void branch) but unused — the if-then-else lifts to a prepend and `onlyKeepSideEffectStmtsAndLast` drops the `.Hole`. Pre-existing nit, not gating.

fabiomadge:
Re-checked at `4efaebd45`: assert/assume expr-position regression fixed via `transformLiftedExpr` (clears `subst`, still lifts calls/holes). Verified (546 green): `assert foo() == 5` translates clean; `assert <??>` → `assertion does not hold`, not the old crash. Older threads addressed/non-gating (inline).

Note: PR is `CONFLICTING` (7 Laurel files vs base, incl. `LiftImperativeExpressions.lean`) — worth confirming the fix survives the merge.

fabiomadge (dismissed):

StrataTest/Languages/Laurel/Examples/Objects/T1_MutableFields.lean
```diff
@@ -9,7 +9,7 @@ import StrataTest.Util.TestLaurel
 open StrataTest.Util
 open Strata
 
-#eval testLaurel
+#eval testLaurelKeepIntermediates
```
keyboardDrummer, line 12: Will undo in next PR

joscoh (approved):

robin-aws (approved):

PR #1355, opened 2026-06-10
feat(laurel): Support bitvector and constrained types as composite fields

olivier-aws:
Add support for bitvector types and constrained types (type aliases with constraints) as composite fields in the Laurel heap parameterization.
  
  1. Bitvector fields – Add TBv n cases to boxConstructorName, boxDestructorName, and boxConstructorDef so composites can have bv-typed fields. The Box variant is named BoxBv{n}, generated
  dynamically per width.
 
  2. Constrained type fields – Add resolveConstrainedType helper that chases constrained type chains to their base type before choosing a Box variant.
 
 Tests: T10_CompositeBvField.lean, T11_ConstrainedField.lean


olivier-aws:
Closing as targetted main instead of main2

PR #1356, opened 2026-06-10
feat(laurel): Support bitvector and constrained types as composite fields

olivier-aws:
Add support for bitvector types, bitvector literals, and constrained types as composite fields in the Laurel heap parameterization. (Retargeted from #1355 onto main2.)
  
** Changes
 1. Bitvector composite fields – TBv n cases in boxConstructorName/boxDestructorName/boxConstructorDef; BoxBv{n} generated per width.
 2. Constrained composite fields – constrained-typed fields box as their resolved base type (e.g. BoxInt) instead of BoxComposite.
 3. Bitvector literals – new grammar syntax <value> bv <width> (e.g. 100 bv 16), wired through AST, parser, formatter, type inference, and Core translation (bitvecConst).
 4. Unified constrained-type resolution – single lookup-agnostic resolveConstrainedTypeWith in LaurelAST; both HeapParameterization and ConstrainedTypeElim delegate to it so they always agree
  on the base type.
 
Tests: T10_CompositeBvField, T11_ConstrainedField (incl. write-side constraint check and a documented read-side completeness gap).
 
 Known completeness gap (follow-up): reading a constrained composite field doesn't recover its constraint (heap stores the unconstrained base type; no assume on read). Writes are sound; locals unaffected. Documented as readCountCompletenessGap; to be fixed in a follow-up.


Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -170,11 +170,20 @@ private def isDatatype (model : SemanticModel) (name : Identifier) : Bool :=
...
+    (fun name => match model.get name with
+      | .constrainedType ct => some ct
+      | _ => none) ty
+
 /-- Get the Box destructor name for a given Laurel HighType.
     For UserDefined datatypes, uses "Box..<datatypeName>Val!";
     for Composite types, uses "Box..compositeVal!". -/
 def boxDestructorName (model : SemanticModel) (ty : HighType) : Identifier :=
-  match ty with
+  match resolveConstrainedType model ty with
```
keyboardDrummer, line 186: The heap param pass having to think about constrained types is concerning. Constrained types is a high level feature and it should not complicate other features. I think it would be better if the ConstrainedTypeElim pass would replace constrained types with their base types in the AST

Is it possible to have  separate PRs for the changes to Bitvectors and to constrained types? I ask because the bitvector changes already look good.

keyboardDrummer (changes requested):

Strata/Languages/Laurel/HeapParameterization.lean
olivier-aws: Thanks @keyboardDrummer . Created a dedicated PR for BV types: https://github.com/strata-org/Strata/pull/1357

olivier-aws:
Split into two pull requests

PR #1357, opened 2026-06-11
feat(laurel): Support bitvector types as composite fields

olivier-aws:
## Summary
Split out of #1356, this PR contains **only the bitvector-related changes**. The constrained-types work from #1356 is being handled separately.

Adds support for bitvector-typed fields in composite types, plus bitvector literal syntax in the Laurel grammar.

### Changes
- **HeapParameterization**: add `.TBv n` cases to `boxConstructorName`, `boxDestructorName`, and `boxConstructorDef` so composite types can have bv-typed fields (boxed as `BoxBv{n}`, generated per width).
- **Grammar**: add `bvLiteral(value, width)` operation with syntax `<value> bv <width>` (e.g. `100 bv 16`).
- **AST**: add `LiteralBv (value width : Nat)` variant to `StmtExpr`.
- **Parsing/formatting/translation/typing**: handle bv literals in `ConcreteToAbstractTreeTranslator`, `AbstractToConcreteTreeTranslator`, `LaurelToCoreTranslator` (-> Lambda `bitvecConst`), and type inference (`LiteralBv` has type `TBv width`).

### Tests
- `T10_CompositeBvField`: covers writing/reading bv fields, bv literals, and an error case (postcondition violation when a bv field is not written / wrong literal value).
- Verified with `lake build` of the affected Laurel modules and the T10 test (`#guard_msgs` passes).


keyboardDrummer (approved):
Excellent!

joehendrix (approved):

Issue #1350, opened 2026-06-09
Laurel: body-tested loop encoding for do-while (and friends)


fabiomadge:
## Problem

Laurel's only loop primitive is head-tested `While(cond, invariants, decreases, body)` (`Languages/Laurel/LaurelAST.lean:270`). Front-ends with body-tested loops (Java do-while; JVerify https://github.com/aws/jverify/pull/427) must bridge this themselves, and **both available front-end encodings are bad**:

- **`while(true)` + `if(!cond) exit(break)`** — the guard becomes constant `true`, so the exit `assume(¬G)` (`Transform/LoopElim.lean:173`) degenerates to `assume(false)`. The negated real guard never reaches the post-loop `assert(Q)` → **incomplete**: `do { inv; x++; } while (x < 3)` cannot prove `x == 3` (confirmed empirically against the verifier; the identical head-tested `while` proves it).
- **Peel first iteration + head-tested `while`** — sound and complete, but emits the body twice per loop → **2ⁿ blowup** in VC size / solve time at nesting depth n.

No sound + complete + linear front-end encoding exists: the head→body-tested bridge inherently forces either dropping the exit-guard fact (incomplete) or duplicating the body (exponential).

## Proposal

Add a body-tested loop at the Laurel level — a `doWhile` grammar op, or a body-first variant of `While` — lowered once in `LoopElim`:

1. Assert the invariant before the first iteration (preserves the entry-invariant soundness check, `LoopElim.lean:144`).
2. Run the body once.
3. Reuse the head-tested machinery for the remainder, threading the **real** guard so `I ∧ ¬cond` holds at exit.

Result: single body in the IR (linear), real guard (complete), `decreases`/termination works. Precedent: `forLoop` is already lowered centrally (`Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean:325`).

Any imperative front-end with body-tested loops (Java now; C-like later) gets correct, scalable do-while for free.

## Open design question

Body-tested invariant semantics: invariant checked *before every iteration including the first* (what jverify#427 picks) vs. *after each body* (JML `maintaining`-style). This is the call flagged as unresolved in aws/jverify@bf85240f ("require more design"); deciding it once in the verified core is preferable to re-litigating per front-end.

## References

Pinned to Strata @ `e115b8ec` (paths/lines moved between versions):

- `Languages/Laurel/LaurelAST.lean:270` — `While` (head-tested; has `decreases`)
[...]

PR #1358, opened 2026-06-11
feat(laurel): add body-tested do-while loop

fabiomadge:
Laurel had only a pre-test `while`. This adds a `doWhile` grammar op and a body-tested `do … while` loop. Rather than a separate AST node, a do-while is represented as a post-test `While` — the existing `While` constructor gains a `postTest : Bool := false` field — and is lowered in a new `EliminateDoWhile` Laurel-to-Laurel pass into the existing pre-test loop:

```
  do S while(G) invariant I
```
becomes
```
  { while(true) invariant I { S; if (!G) { exit L } } } L
```

`L` is a fresh, collision-free label (`$dowhile_exit_{n}`, `$`-prefixed so it can't collide with user identifiers), so nested do-whiles and user `break`/`continue` don't capture each other's exits. The body runs once per iteration with the guard re-checked after it, and the real guard reaches post-loop code via the structured `exit` — so the encoding is sound, complete, and linear (single body in the IR, no peeling/duplication).

Invariant placement is head-tested (checked before each body), matching `while`. **Gotcha (documented in T23):** because the guard is re-tested only after the body, the invariant must hold of the *pre-body* state, so `do { x:=x+1 } while(x<3)` needs the bound `x <= 2`, not `x <= 3`.

**Why a flag + a pass:** `postTest` captures the while/do-while distinction at the type level (per review feedback) rather than duplicating the loop as a second constructor — so most passes match `While` once and carry the flag through, with no `.DoWhile` arms. `postTest := false` (pre-test) is the default, so every existing `.While` construction is unchanged. The desugaring lives in `EliminateDoWhile` rather than in `ConcreteToAbstractTreeTranslator`, so it runs for every program reaching the pipeline — including one supplied directly as abstract AST, bypassing the concrete translator. The parser builds a post-test `While`; the serializer emits `while`/`doWhile` by the flag; the pass (modeled on `EliminateIncrDecr`, with its own fresh-label counter) does the desugar and runs first, so no later pass observes `postTest = true`. No new Core construct or `LoopElim` change — the desugar reuses the existing verified pre-test loop machinery. Resolution handles `postTest` in `Check.while` directly (it runs before the passes, the same arrangement as `IncrDecr`).

The per-program traversal over static and composite-instance procedures, which both `EliminateDoWhile` and `EliminateIncrDecr` had hand-rolled, is now `mapProgramProceduresM` in `MapStmtExpr` (per review feedback) — so neither pass refers to unrelated features like instance procedures.

Tests in `T23_DoWhile` (end-to-end): basic, runs-at-least-once (guard false at entry), nested, break-out via labelled block, no-invariant (zero `invariant` clauses — asserts the negated guard, the only fact provable without an invariant), and `falsePostRejected` (a false postcondition is rejected, confirming the `while(true)` desugar isn't vacuous). Plus `EliminateDoWhileTest` (pass-output shape, incl. distinct fresh labels for nested loops) and a round-trip case in `AbstractToConcreteTreeTranslatorTest` (serialization arg order).

**Earlier label-clash warning, now resolved.** Multi-procedure `T23` used to surface a latent label clash (`⚠️ [addPathCondition] Label clash detected for assume_invariant_0_0, …`) — `do-while` was just the first construct to trip it, because path-condition labels weren't reset between procedures. This branch has merged the base fix #1391 (path-condition isolation at the procedure boundary), so the warning is gone (verified: zero clashes on the T23 build). Not caused by this PR.

**Strata half only.** To use do-while end-to-end, the jverify front-end must emit the new op (regenerate the `laurel/` bindings and add a `JCDoWhileLoop` case to `JavaToLaurelCompiler`, which currently rejects do-while) — a separate follow-up. Implements the front-end desugar from #1350.

### DIFF
--- a/Strata/Languages/Laurel/EliminateDoWhile.lean
+++ b/Strata/Languages/Laurel/EliminateDoWhile.lean
@@ -1,0 +1,91 @@
+/-
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+public import Strata.Languages.Laurel.LaurelAST
+public import Strata.Languages.Laurel.LaurelPass
+import Strata.Languages.Laurel.MapStmtExpr
+import Strata.Util.Tactics
+
+/-!
+# Eliminate Do-While
+
+Lowers post-test `While` loops (`postTest = true`, the `do … while` form) into
+the pre-test `while` machinery. Runs early so that no later pass (or the Core
+translator) observes `postTest = true`.
+
+The desugaring of `do BODY while(COND) invariant I` is:
+```
+  { while(true) invariant I { BODY; if (!COND) exit L } } L
+```
+`BODY` runs once per iteration with `COND` re-checked after it; the real guard
+reaches post-loop code via the structured `exit L`, so the encoding is sound,
+complete, and linear (a single body in the IR — no peeling/duplication). The
+invariant `I` is checked at the loop head (before each body), matching `while`.
+
+The fresh exit label `L` is `$dowhile_exit_{n}` for a per-program monotonic
+counter `n`. The leading `$` keeps it out of the user-name space (no source
+identifier can contain `$`), so it can never capture, or be captured by, a user
+`break`/`continue` label, and the counter keeps nested do-whiles distinct.
+
+The traversal is the generic bottom-up `mapStmtExprM`, so inner do-whiles are
+eliminated before their enclosing ones.
+-/
+
+namespace Strata.Laurel
+
+/-- Monotonic counter feeding fresh exit labels (scheme described in the module header). -/
+private structure ElimState where
+  freshCounter : Nat := 0
+
+private abbrev ElimM := StateM ElimState
+
+private def freshExitLabel : ElimM String :=
+  modifyGet fun s => (s!"$dowhile_exit_{s.freshCounter}", { s with freshCounter := s.freshCounter + 1 })
+
+/-- Rewrites a post-test `While` to its pre-test desugaring; all other nodes pass through. -/
+private def rewriteNode (node : StmtExprMd) : ElimM StmtExprMd := do
+  match node.val with
+  | .While cond invs dec body true =>
+    let source := node.source
+    let exitLabel ← freshExitLabel
+    let notCond : StmtExprMd := ⟨.PrimitiveOp .Not [cond], source⟩
+    let exitStmt : StmtExprMd := ⟨.Exit exitLabel, source⟩
+    let guardCheck : StmtExprMd := ⟨.IfThenElse notCond exitStmt none, source⟩
+    let loopBody : StmtExprMd := ⟨.Block [body, guardCheck] none, source⟩
+    let trueCond : StmtExprMd := ⟨.LiteralBool true, source⟩
+    -- Thread `dec` onto the `while(true)` rather than dropping it: each desugared
+    -- iteration is one pass through that loop's head, so the measure decreases
+    -- across exactly those iterations. The emitted loop is pre-test
+    -- (`postTest := false`), so it is not rewritten again.
+    let whileStmt : StmtExprMd := ⟨.While trueCond invs dec loopBody false, source⟩
+    pure ⟨.Block [whileStmt] (some exitLabel), source⟩
+  | _ => pure node
+
+public section
+
+/-- Eliminate every post-test `While` in a Laurel program; afterward every `While` has `postTest = false`. -/
+def eliminateDoWhile (program : Program) : Program :=
+  let rewrite : Procedure → ElimM Procedure := mapProcedureM (mapStmtExprM rewriteNode)
+  let go : ElimM Program := do
+    let staticProcs ← program.staticProcedures.mapM rewrite
+    let types ← program.types.mapM fun td =>
+      match td with
+      | .Composite ct => do
+        let procs ← ct.instanceProcedures.mapM rewrite
+        pure (.Composite { ct with instanceProcedures := procs })
+      | other => pure other
+    pure { program with staticProcedures := staticProcs, types := types }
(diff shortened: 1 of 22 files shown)


kadirayk (commented):
Nice work!  Clean and well-scoped change. The desugaring to existing`while` strategy reuses machinery that's already verified. 

## Design
In general the design is sound.
- The desugar `{ while(true) invariant I { S; if (!G) exit L } } L` is correct do-while semantics: body runs once before the guard is tested, and the structured `exit L` is the only escape from `while(true)`. 
- The guard is translated once (`ConcreteToAbstractTreeTranslator.lean`: a single `let cond ← translateStmtExpr condArg` feeding one `.PrimitiveOp .Not [cond]`). No peeling/duplication, so a side-effecting guard won't be evaluated twice. 
- Fresh labels are `$`-prefixed (`freshLabel` returns `s!"${pfx}_{s.freshCounter}"`) and backed by a monotonic counter on `TransState`. Since `$` isn't a legal user identifier character, collision with user labels is structurally impossible, and the counter handles nesting.
- Extracting `translateInvariantClauses` and reusing it across `while`/`forLoop`/`doWhile` is a nice dedup, and the extracted body is character-identical to the old inline code, so it's behavior-preserving.


## Test coverage

The five original `T23` cases (basic, runs-at-least-once, nested, break-out, false-postcondition-rejected) are a good and the negative `falsePostRejected` case is important. One gap:

1. **No-invariant case.** The grammar is `invariants: Seq InvariantClause`
   (zero-or-more), and `translateInvariantClauses` returns `[]` on the empty case, so `do { ... } while(G)` with no `invariant` should parse and verify. This path was untested. 


## Nitpicking
- In `freshLabel`, the counter is read before the `set`, so the *returned* label uses the pre-increment value while the state stores the post-increment value. Could be kept like this, but feels rather unusual.

kadirayk (dismissed):

kadirayk (dismissed):
Thanks for the update. Both my earlier points are addressed:

- `freshLabel` simplified to a single atomic `modifyGet` (pre/post-increment nit gone).
- `noInvariant` test added, covering the empty-invariant path.

kadirayk (dismissed):
Reapproving

Strata/Languages/Laurel/Grammar/ConcreteToAbstractTreeTranslator.lean
```diff
@@ -338,30 +359,37 @@ partial def translateStmtExpr (arg : Arg) : TransM StmtExprMd := do
...
-              | q`Laurel.invariantClause, #[exprArg] => translateStmtExpr exprArg
-              | _, _ => TransM.error "Expected invariantClause"
-            | _ => TransM.error "Expected operation"
-        | _ => pure []
+      let invariants ← translateInvariantClauses translateStmtExpr invSeqArg
       let body ← translateStmtExpr bodyArg
       let whileBody := mkStmtExprMd (.Block [body, step] none) src
       let whileStmt := mkStmtExprMd (.While cond invariants none whileBody) src
       return mkStmtExprMd (.Block [init, whileStmt] none) src
+    | q`Laurel.doWhile, #[bodyArg, condArg, invSeqArg] =>
```
keyboardDrummer, line 367: Can you do the desugaring after parsing? I'm planning to update the serialization so `ConcreteToAbstractTreeTranslator.lean` won't even run when calling Strata from JVerify.

keyboardDrummer (changes requested):

Strata/Languages/Laurel/LaurelAST.lean
```diff
@@ -307,6 +307,14 @@ inductive StmtExpr : Type where
   | While (cond : AstNode StmtExpr) (invariants : List (AstNode StmtExpr))
     (decreases : Option (AstNode StmtExpr))
     (body : AstNode StmtExpr)
+  /-- A body-tested `do BODY while(COND)` loop: BODY runs once before COND is
+      first tested. Invariants are checked at the loop head (before each body),
+      matching `while`. Only allowed in impure contexts. Desugared to a
+      head-tested loop by the `EliminateDoWhile` pass. The `decreases` measure
+      mirrors `While`; no surface syntax sets it, so it is `none` as parsed. -/
+  | DoWhile (cond : AstNode StmtExpr) (invariants : List (AstNode StmtExpr))
```
keyboardDrummer: I think I'd prefer a property `doBeforeWhile: Bool` as part of the regular `While`, to capture the repetition at the type level.

fabiomadge: I also prefer merging them. Good call!

Strata/Languages/Laurel/EliminateDoWhile.lean
```diff
@@ -0,0 +1,91 @@
...
+    let whileStmt : StmtExprMd := ⟨.While trueCond invs dec loopBody false, source⟩
+    pure ⟨.Block [whileStmt] (some exitLabel), source⟩
+  | _ => pure node
+
+public section
+
+/-- Eliminate every post-test `While` in a Laurel program; afterward every `While` has `postTest = false`. -/
+def eliminateDoWhile (program : Program) : Program :=
+  let rewrite : Procedure → ElimM Procedure := mapProcedureM (mapStmtExprM rewriteNode)
+  let go : ElimM Program := do
```
keyboardDrummer: Can we extract this mapping over static and instance procedures into `MapStmtExpr.lean` as well? I would rather not refer to unrelated features like instance procedures from the EliminateDoWhile file.

keyboardDrummer (commented):
Really nice. Left one more comment

keyboardDrummer (approved):

kadirayk (approved):
The `postTest` flag + `EliminateDoWhile` pass improves over the desugaring in translator, and reuses the same verified loop encoding. LGTM 👍

PR #1361, opened 2026-06-11
Per-invariant source ranges for loop-invariant diagnostics

kadirayk:
## Summary

Loop-invariant verification diagnostics (a failing `invariant(...)` in a `while`/`for` loop) previously pointed at the **whole loop** instead of the specific invariant that failed. This change threads each invariant's source range through the loop's metadata so loop elimination can attribute each invariant's verification condition to that invariant's own source location.

This support is required for [JVerify#437](https://github.com/strata-org/jverify/issues/437)

## Problem

In Strata, loop-invariant proof obligations are synthesized by `LoopElim` and were tagged with the loop-wide metadata `md`. The per-invariant source range was lost earlier in the pipeline: Core loop invariants are bare `(label, expr)` pairs and Core expressions are `Unit`-annotated, so they carry no source range.

A front-end-only change does not help: the diagnostic location is driven by the synthesized assert's metadata, not by anything attached to the invariant expression. The fix therefore has to live in Strata.

## Solution

Thread each invariant's source range through the loop's existing `MetaData` array and use it per-invariant in `LoopElim`. When a per-invariant provenance is present, each invariant's generated assert/assume is attributed to that invariant's own source location; otherwise we fall back to the loop metadata `md`, so loops not originating from Laurel (Core `.st`, C_Simp) are unchanged.

## Testing

`StrataTest/.../Fundamentals/T13_WhileLoopsError.lean` adds two caret-annotated regression tests using the existing diagnostic harness (`TestDiagnostics`, `matchesDiagnostic`), which checks the exact start/end line and column of each diagnostic:

1. **`badInitialInvariant`** — a single `invariant i >= 0` that fails on entry. The caret asserts the diagnostic lands on the invariant expression, not the `while` loop.
2. **`secondInvariantFails`** — two invariants where the first holds on entry but the second (`invariant j >= 0`) does not. The caret asserts the diagnostic points specifically at the failing second invariant. This is the case that distinguishes per-invariant attribution from loop-wide attribution: before the fix the range would resolve to the loop, so the carets would not match.

Verification:

- `lake build Strata` — compiles, no proof breakage.
- `lake build StrataTest` — all `#guard_msgs` tests pass, including the new ones.
- The fallback to the loop `md` keeps existing Core `.st` and C_Simp loop diagnostics unchanged.

### DIFF
--- a/Strata/DL/Imperative/MetaData.lean
+++ b/Strata/DL/Imperative/MetaData.lean
@@ -293,6 +293,27 @@ def MetaData.eraseAllElems {P : PureExpr} [BEq P.Ident]
     (md : MetaData P) (fld : MetaDataElem.Field P) : MetaData P :=
   md.filter (fun e => !(e.fld == fld))
 
+/-- Label for a per-loop-invariant source provenance. Loop lowering stores one
+    such element per invariant, in invariant order, so that loop elimination can
+    attribute each invariant's verification condition to the specific invariant's
+    source location rather than to the whole loop. -/
+def MetaData.invariantProvenanceLabel : String := "invariantProvenance"
+
+/-- Append a loop invariant's provenance to metadata. These are appended in
+    invariant order. Note this matches on the `.label` constructor directly, so
+    it needs no `BEq P.Ident` instance. -/
+def MetaData.pushInvariantProvenance {P : PureExpr} (md : MetaData P) (p : Provenance) : MetaData P :=
+  md.push { fld := .label MetaData.invariantProvenanceLabel, value := .provenance p }
+
+/-- Get all per-invariant provenances from metadata, in the order they were
+    pushed (matching the loop's invariant order). -/
+def getInvariantProvenances {P : PureExpr} (md : MetaData P) : Array Provenance :=
+  md.filterMap fun elem =>
+    match elem.fld, elem.value with
+    | .label l, .provenance p =>
+      if l == MetaData.invariantProvenanceLabel then some p else none
+    | _, _ => none
+
 /-- Replace the primary provenance with a new one, shifting existing related
     provenances and prepending the old primary provenance. -/
 def MetaData.setCallSiteFileRange {P : PureExpr} [BEq P.Ident]
--- a/Strata/DL/Imperative/MetaData.lean
+++ b/Strata/DL/Imperative/MetaData.lean
@@ -307,7 +307,7 @@ def MetaData.pushInvariantProvenance {P : PureExpr} (md : MetaData P) (p : Prove
 
 /-- Get all per-invariant provenances from metadata, in the order they were
     pushed (matching the loop's invariant order). -/
-def getInvariantProvenances {P : PureExpr} (md : MetaData P) : Array Provenance :=
+def MetaData.getInvariantProvenances {P : PureExpr} (md : MetaData P) : Array Provenance :=
   md.filterMap fun elem =>
     match elem.fld, elem.value with
     | .label l, .provenance p =>
--- a/Strata/Languages/Laurel/LaurelAST.lean
+++ b/Strata/Languages/Laurel/LaurelAST.lean
@@ -393,6 +393,13 @@ def fileRangeToCoreMd (source : Option FileRange) : Imperative.MetaData Core.Exp
   | some fr => Imperative.MetaData.ofSourceRange fr.file fr.range
   | none => Imperative.MetaData.ofProvenance (.synthesized .laurel)
 
+/-- Build a provenance from an optional source location (mirrors
+    `fileRangeToCoreMd`, but returns the bare `Provenance`). -/
+def fileRangeToProvenance (source : Option FileRange) : Provenance :=
+  match source with
+  | some fr => Provenance.ofSourceRange fr.file fr.range
+  | none => .synthesized .laurel
+
 /-- Build Core metadata from an AstNode's source location. -/
 def astNodeToCoreMd (node : AstNode α) : Imperative.MetaData Core.Expression :=
   fileRangeToCoreMd node.source
--- a/Strata/Languages/Laurel/LaurelAST.lean
+++ b/Strata/Languages/Laurel/LaurelAST.lean
@@ -387,18 +387,15 @@ def Condition.mapM [Monad m] (f : AstNode StmtExpr → m (AstNode StmtExpr)) (c 
 def Condition.mapCondition (f : AstNode StmtExpr → AstNode StmtExpr) (c : Condition) : Condition :=
   { c with condition := f c.condition }
 
-/-- Build Core metadata from an optional source location. -/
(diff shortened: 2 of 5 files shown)


Strata/Transform/LoopElim.lean
```diff
@@ -140,16 +140,26 @@ def Stmt.removeLoopsM
...
+    -- Per-invariant source provenance, threaded through the loop metadata by the
+    -- Laurel→Core translation (in invariant order). When present, each
+    -- invariant's generated assert/assume is attributed to that invariant's own
+    -- source location instead of the whole loop; otherwise we fall back to the
+    -- loop metadata `md` (e.g. for loops not originating from Laurel).
+    let invProvs := getInvariantProvenances md
+    let invMd : Nat → MetaData P := fun i =>
+      match invProvs[i]? with
+      | some p => MetaData.ofProvenance p
+      | none => md
```
fabiomadge: **Latent: sourceless invariant loses its range.** `fileRangeToProvenance none = .synthesized .laurel` (never `none`), so `invProvs[i]?` returns `some (.synthesized …)`, this `| none => md` never fires, and `toDiagnostic` drops the range — worse than the pre-PR loop range. Unreachable today, so latent. Gate on `.loc`:
```lean
match invProvs[i]? with
| some (p@(.loc ..)) => MetaData.ofProvenance p
| _ => md
```

Strata/Transform/LoopElim.lean
```diff
@@ -140,16 +140,26 @@ def Stmt.removeLoopsM
     -- reference to the source invariant.
     let invSuffix : Nat → String → String := fun i lbl =>
       if lbl.isEmpty then toString i else s!"{i}_{lbl}"
+    -- Per-invariant source provenance, threaded through the loop metadata by the
+    -- Laurel→Core translation (in invariant order). When present, each
+    -- invariant's generated assert/assume is attributed to that invariant's own
+    -- source location instead of the whole loop; otherwise we fall back to the
+    -- loop metadata `md` (e.g. for loops not originating from Laurel).
+    let invProvs := getInvariantProvenances md
```
fabiomadge: **Implicit cross-file contract.** `invProvs[i]` ↔ i-th invariant spans push (`LaurelToCoreTranslator`) and retrieval (here). It holds and `secondInvariantFails` pins it, but nothing enforces it and `invProvs[i]?` silently masks a desync. A note or `invProvs.size == invariants.length` check would harden it.

Strata/Languages/Laurel/LaurelAST.lean
```diff
@@ -393,6 +393,13 @@ def fileRangeToCoreMd (source : Option FileRange) : Imperative.MetaData Core.Exp
   | some fr => Imperative.MetaData.ofSourceRange fr.file fr.range
   | none => Imperative.MetaData.ofProvenance (.synthesized .laurel)
 
+/-- Build a provenance from an optional source location (mirrors
+    `fileRangeToCoreMd`, but returns the bare `Provenance`). -/
+def fileRangeToProvenance (source : Option FileRange) : Provenance :=
```
fabiomadge, line 422: **Duplication.** Same `match` as `fileRangeToCoreMd` just above. Since `ofSourceRange = ofProvenance ∘ Provenance.ofSourceRange`, make `fileRangeToCoreMd := MetaData.ofProvenance ∘ fileRangeToProvenance` so this is the only primitive.

Strata/DL/Imperative/MetaData.lean
```diff
@@ -293,6 +293,27 @@ def MetaData.eraseAllElems {P : PureExpr} [BEq P.Ident]
...
+
+/-- Append a loop invariant's provenance to metadata. These are appended in
+    invariant order. Note this matches on the `.label` constructor directly, so
+    it needs no `BEq P.Ident` instance. -/
+def MetaData.pushInvariantProvenance {P : PureExpr} (md : MetaData P) (p : Provenance) : MetaData P :=
+  md.push { fld := .label MetaData.invariantProvenanceLabel, value := .provenance p }
+
+/-- Get all per-invariant provenances from metadata, in the order they were
+    pushed (matching the loop's invariant order). -/
+def getInvariantProvenances {P : PureExpr} (md : MetaData P) : Array Provenance :=
```
fabiomadge: **Naming.** Namespace it (`MetaData.getInvariantProvenances`) to match `MetaData.pushInvariantProvenance`. (The bare, `BEq`-free form is otherwise fine and mirrors the `relatedFileRange` getter.)

fabiomadge (commented):
**LGTM, non-blocking comments** (leaving as a comment, not a formal approval). Clean, well-scoped fix for the `LoopElim` whole-loop-range issue (jverify#437). Built and verified: build green, loop suite green (T13/T17/Core Loops), VCs unchanged — each invariant's expr, label, and property-type are untouched; only the per-invariant `MetaData` differs, and metadata drives diagnostics, not the VC.

**Design is the right tradeoff.** The Core loop IR stores invariants as `(String × Expr)` with no per-invariant metadata slot, and Core exprs are `Unit`-annotated. Threading provenances through the loop `md` is the correct localized choice (4 files, no IR/proof/backend churn):
- Metadata on the invariant tuple is cleaner but changes `Stmt.loop` + its induction principle → ripples into semantics proofs and the CBMC backend. Too invasive.
- Metadata on Core exprs needs a slot the `Unit`-annotated type lacks — the "front-end-only fix doesn't help" point, confirmed.
- It mirrors the existing `relatedFileRange`/`getRelatedFileRanges` pattern, so it's house idiom.

**Main ask: add a for-loop test.** For-loops work (verified: a failing 2nd for-loop invariant lands on the invariant expr, since `for` desugars to the same `.While`), but only while-loops are tested. Verified drop-in for `T13_WhileLoopsError.lean`:

```lean
def forSecondInvFailsProgram := r"
procedure forSecondInvFails()
  opaque
{
    var j: int := -1;
    for(var i: int := 0; i < 10; i := i + 1)
      invariant i >= 0
      invariant j >= 0
//              ^^^^^^ error: assertion does not hold
    {
        j := j + 1
    }
};
"

#guard_msgs (drop info, error) in
#eval testInputWithOffset "ForSecondInvFails" forSecondInvFailsProgram 60 processLaurelFile
```

Good to merge once that's in. Remaining points inline.

fabiomadge (commented):
_(duplicate of the review above — posted twice by mistake; see the other review for the full content)_

fabiomadge (approved):
Thanks for the follow-ups — all four points from my earlier comments are addressed: the `.loc` gating (so a sourceless invariant falls back to the loop range instead of losing it), the `fileRangeToCoreMd`/`fileRangeToProvenance` de-dup, the documented `invProvs[i]` contract with graceful fallback, and the `MetaData.getInvariantProvenances` naming. The added for-loop test is a nice extra. Built `T13_WhileLoopsError` locally and the per-invariant caret annotations all match. LGTM.

aqjune-aws (approved):
Great patch :)

PR #1364, opened 2026-06-11
[Laurel] Support constrained types as composite fields

olivier-aws:
## Summary

Adds support for **constrained types as composite fields** in Laurel — e.g. a composite with a field of a refinement type:

```
constrained nat = x: int where x >= 0 witness 0
composite Counter { var count: nat }
```

Field writes now check the constraint (`c#count := -1` fails), and the field is boxed as its base type on the heap.

## ConstrainedTypeElim changes made along the way

Supporting this cleanly required running `ConstrainedTypeElim` **earlier** in the Laurel pipeline (right after `typeAliasElim`, before `HeapParameterization`) so that constrained field types are lowered to their base types before the heap-boxing pass needs them. Moving the pass forward exposed a few places where it had implicitly relied on running last, which this PR fixes:

1. **Composite field types** — `ConstrainedTypeElim` removed the constrained type definitions but left `UserDefined "nat"` references inside composite fields, which then failed to resolve in Core translation. `elimCompositeType` now resolves constrained field types to their base types (and runs elimination on instance procedures) before the definitions are removed. The field-write constraint check (previously enabled by `HeapParameterization` via a `Declare` temporary) is now emitted directly from the `.Field` assignment target.

2. **Constrained-typed assignments in expression position** — `elimStmt` only checked assignments that appear as statements, so `y := (x := -1) + 1` with `x : nat` was unchecked. `wrapExprAssigns` now traverses expression positions and wraps such assignments as `{ x := v; assert T$constraint(x); x }`, asserting on a read-back so the RHS is evaluated exactly once (semantics-preserving).

3. **`LiftExpressionAssignments` assert ordering** — for the read-back form to work, the assert must stay *after* the assignment through lowering. `transformExpr` was prepending `.Assert`/`.Assume` in expression-position blocks only after the assignment had already been prepended, moving them ahead of it (so they checked the stale value). `transformExpr` now lifts `.Assert`/`.Assume` during its traversal so they keep their position relative to assignments lifted from the same block. This is a general correctness fix for the lifting pass.

4. **Cleanup** — once `ConstrainedTypeElim` runs first, `HeapParameterization` no longer needs its own constrained-type resolution (the field types reaching it are already base types), so the dead `resolveConstrainedType` helper and the shared `resolveConstrainedTypeWith` helper in `LaurelAST` were removed. Also included the offending type name in the `LaurelToCoreTranslator` "could not be resolved" diagnostic.

## Pipeline placement

`constrainedTypeElimPass` is moved to the second position (after `typeAliasElimPass`). It has no `comesBefore` ordering constraints and nothing depends on it running late, so the load-time `comesBeforeRespected` check still holds.

## Notes

- Bitvector-as-composite-field changes that previously sat on this branch have been removed; they live in a separate branch/PR. This PR is scoped to constrained types.
- Rebased onto the latest `main2` (which includes the #1222 pipeline-framework refactor).

## Testing
- Full `lake build` passes (495 jobs, including all `#guard_msgs` elaboration-time tests and the load-time pipeline-ordering check).
- `T11_ConstrainedField` covers constrained composite fields (valid/invalid writes + a documented read-side completeness gap).
- `T10_ConstrainedTypes` gains a `sideEffect` case exercising an assignment-in-expression to a constrained local.


Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -126,7 +126,58 @@ private def inScope (action : ElimM α) : ElimM α := do
...
+  | _ => node
+
+/-- Insert constraint assertions for every assignment to a constrained-typed
+    target that appears within an expression. A no-op on expressions that
+    contain no such assignment. -/
+def wrapExprAssigns (ptMap : ConstrainedTypeMap) (model : SemanticModel)
+    (expr : StmtExprMd) : StmtExprMd :=
+  mapStmtExpr (wrapAssignNode ptMap model) expr
+
+def elimStmt (ptMap : ConstrainedTypeMap) (model : SemanticModel)
```
keyboardDrummer: If you want to fix more bugs in the ConstrainedTypeElim pass, `elimStmt` should use `mapStmtExpr` or one of its forms so it does not need a catch all case `| _ => pure [stmt]` at the end which is sure to cause incomplete behavior.

keyboardDrummer (dismissed):
Nice!

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -154,23 +209,39 @@ def elimStmt (ptMap : ConstrainedTypeMap)
...
+        -- Writing to a constrained-typed composite field: assert the
+        -- constraint holds for the value being stored. The field's declared
+        -- type comes from the semantic model. (When this pass ran after
+        -- HeapParameterization, this check was produced via the `Declare`
+        -- temporary that lowering introduced; running first, we must emit it
+        -- directly from the field write, asserting on the assigned value.)
+        match (model.get fieldName).getType.val with
+        | .UserDefined name =>
+          if ptMap.contains name.text then
+            let c : StmtExprMd := ⟨.StaticCall (mkId s!"{name.text}$constraint") [value], source⟩
```
fabiomadge: **Assert on a field read-back, not `value`.** `value` is the full RHS expression, already emitted as the field-write statement above — re-using it here emits it a second time, so any side effect in it runs twice. Verified (with `x` starting at `0`): `c#count := (x:=x+1)+1` ends with `x == 2` instead of `1` — the output is `c#count := (x:=x+1)+1; assert nat$constraint((x:=x+1)+1)`, so `x:=x+1` runs twice (0→2). Reading the field back evaluates the RHS once. This mirrors the `constrainedTargetReadback` helper at `:133`, which already builds this exact read-back for the expression-position case. Gives `x == 1` non-vacuously; suite + proofs green after the two T11 annotation updates (see body).

```suggestion
            let readback : StmtExprMd := ⟨.Var (.Field _fieldTarget fieldName), source⟩
            let c : StmtExprMd := ⟨.StaticCall (mkId s!"{name.text}$constraint") [readback], source⟩
```

fabiomadge (changes requested):
Already approved, but flagging a **verified correctness bug** — requesting changes. The feature is sound and deferring the read-side gap is fine (verified: genuine incompleteness, not unsoundness).

**Field-write constraint check double-evaluates the RHS.** The `.Field` arm at `:222` asserts the constraint on `value` — the whole RHS expression — which has already been emitted as the field-write statement just above (`:197`), so any side effect in the RHS runs twice. Verified on this head, with `count : nat` and `x` starting at `0`: `c#count := (x := x + 1) + 1` leaves `x == 2`, not `1`; the `ConstrainedTypeElim` output is literally `c#count := (x:=x+1)+1; assert nat$constraint((x:=x+1)+1)` — `x:=x+1` appears twice, so `x` is incremented 0→2 instead of 0→1. A legal program gets the wrong state. Fix inline at `:222`: assert on a read-back of the field, not on `value`. With it, `x == 1` holds (non-vacuous) and suite + proofs stay green **once T11's two `does not hold` carets become `could not be proved`** (the read-back reasons through a heap read). This is in the hand-written `elimStmt` traversal you flagged at `:180` — though it's a double-emission rather than the catch-all incompleteness you mentioned. Notably the pass already has a `constrainedTargetReadback` helper (`:133`) that builds exactly this field read-back for expression-position assignments; the statement-level arm just doesn't use it.

**Non-blocking — a smaller alternative.** The bug, the new `wrapExprAssigns`, and the edit to the shared `LiftImperativeExpressions` all stem from moving `ConstrainedTypeElim` ahead of lifting. Keeping it last (as on `main2`) and instead resolving constrained field types to base inside `HeapParameterization` (box selection + `Eq`/`Neq`) reaches the same goal: I prototyped it at **−79 lines**, no shared-pass change, cleaner "does not hold" diagnostic, double-eval can't arise. I know #1356 argued heap-param shouldn't know about constrained types — resurfacing it only because that was decided before this bug and before I'd verified the alternative end-to-end. Your call, not a blocker.

**Minor aside (no action).** `Eq`/`Neq` ref-compares any `.UserDefined _` operand without resolving constrained types. The reorder makes that safe here (verified: `c#count == d#count` lowers to an int compare), so it's an ordering invariant, not a live bug — but a one-line `resolveBoxType` there would make it order-independent.

Strata/Languages/Laurel/ConstrainedTypeElim.lean
olivier-aws: Thanks for noticing the soundness issue. Fixed in [c647619](https://github.com/strata-org/Strata/pull/1364/commits/c64761976d4386592c750b7385ec3c3020c7b122). I prefer to keep the pass early in the pipeline to make the assert/assume explicit ASAP, so that other passes do not need to worry about constrained types at all.

fabiomadge (dismissed):

Strata/Languages/Laurel/ConstrainedTypeElim.lean
```diff
@@ -154,23 +209,49 @@ def elimStmt (ptMap : ConstrainedTypeMap)
             fun c => ⟨.Assert { condition := c }, source⟩
           pure (acc ++ assert)
         | none => pure acc
-      | _ => pure acc
-    pure ([stmt] ++ declareChecks)
+      | .Field fieldTarget fieldName => do
```
keyboardDrummer: I think it's wrong for this pass to be aware of the targets, for example the `.Field` target. Constrained types is a feature that can logically be defined independently of fields.

I think the right solution here is to only operate on the RHS. If you want to prevent evaluating the RHS twice, you will have to introduce a temporary variable: 
```
var x: nat := <expr>
```
==>
```
var x: nat := (var temp := <expr>; assert constraint_nat(temp); temp)
```

keyboardDrummer (changes requested):

Strata/Languages/Laurel/ConstrainedTypeElim.lean
olivier-aws: The pass needs at least to know the target type to know if a check must be inserted. Given that types are resolved differently if the target is a local variable or a field, some distinction has to be made. I tried to refactor the code to keep this case very local and outside of the traversal done by `elimStmt`.

fabiomadge: Read-back vs RHS precision, measured on `a6e0fbc5c`:

| Write | Result |
|---|---|
| `y := -1` (local, direct) | `does not hold` |
| `c#count := -1` (field, read-back) | `could not be proved` |
| `c#count := 5` (valid field) | passes |

The field read-back goes through the heap box/unbox, so an invalid field write comes back `unknown`, not a definite `fail` — sound but weaker than the local case. @keyboardDrummer's temp-var-on-RHS would check `constraint(-1)` directly → `does not hold`. So RHS-vs-read-back is a real completeness (not soundness) tradeoff. Just data, not a design vote.

olivier-aws: Thanks for the data. I assume that we will at some point improve our encoding/precision so that something like:
```
c#count := 3;
assert(c#count == 3);
```
will be proven, despite the heap box/unbox. Then the read-back case that you showed should return "does not hold" as well. So I would prefer to keep the encoding as it is now (with assert on the read-back) to avoid the cost of introducing a new temp variable.

keyboardDrummer: Makes sense @olivier-aws, let's keep the read-back encoding. I do think the read-back code is re-usable for other passes as well. I extracted that code and did some other refactoring to the constrained type pass and placed it in this PR: https://github.com/strata-org/Strata/pull/1400

@olivier-aws could you have a look to see if you like those changes, and if so merge them into this PR?

keyboardDrummer (dismissed):

fabiomadge (dismissed):

olivier-aws:
@keyboardDrummer-bot please fix conflicts

fabiomadge (approved):

keyboardDrummer (approved):

PR #1366, opened 2026-06-12
fix(core): thread LoopElim loop counter across procedures

fabiomadge:
## Problem

`loopElim` ran `removeLoopsM` with a fresh `LoopElimState {}` per procedure, so `loopCounter` reset to 0 each time and the first loop of every procedure emitted the same labels (`assume_invariant_0_0`, `loop_havoc_0`, …).

The verifier (`CmdEval.addPathCondition`) accumulates path conditions across all procedures in one run. When collided labels stay live there, it auto-renames and emits a noisy (but harmless) trace:

```
⚠️ [addPathCondition] Label clash detected for assume_invariant_0_0, using unique label assume_invariant_0_0_1.
```

Verification is correct either way; this just removes the noise. It showed up via the desugared `do-while` from #1358.

## Fix

Thread the `LoopElimState` (and so `loopCounter`) through the per-procedure fold instead of resetting it, making loop numbers monotonic program-wide. Statistics accumulate identically (`merge`/`increment` are both per-key sums).

## Blast radius

Only `StrataBooleTest/grammar_extensions` changes — a 5-loop multi-procedure program now numbered `0..4` instead of all-`0`; its expected labels are updated. Every other pinned-label snapshot uses `loop_num = 0`, unchanged.

## Verification

- `lake build StrataTest` — green (529)
- `lake build StrataBooleTest` — green (227)
- `lake test -- --exclude Languages.Python` — loop tests pass; only failure is the pre-existing `Languages.Java.TestGen` (missing optional `ion-java` jar, unrelated)

### DIFF
--- a/Strata/Transform/LoopElim.lean
+++ b/Strata/Transform/LoopElim.lean
@@ -241,20 +241,27 @@ def Stmt.removeLoops
 
 /-- Eliminate loops in all procedures of a Core program by replacing each loop
     with assertions and assumptions about its invariants.
-    Returns the transformed program together with accumulated statistics. -/
+    Returns the transformed program together with accumulated statistics.
+
+    The `LoopElimState` (and so `loopCounter`) is threaded across procedures
+    rather than reset per procedure, so the generated labels
+    (`assume_invariant_{loop_num}_…`, `loop_havoc_{loop_num}`, etc.) are unique
+    program-wide. The downstream verifier accumulates path conditions across all
+    procedures in a single run, so a per-procedure reset would make the first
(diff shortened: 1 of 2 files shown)


fabiomadge:
The label clash here is a symptom of #1390 — a soundness leak where a procedure's path-conditions escape past a structured `exit` into later procedures. This PR silences the warning but not the leak (#1390's `assert false` repro still passes with it applied); once #1390's fix lands the clash disappears on its own. Flagging so it isn't read as fixing the root issue — cosmetic label change at most.

PR #1367, opened 2026-06-12
File-relative error reporting

leo-leesco:
The last testing framework update regressed the error reporting to be tested-snippet-relative, but that is inconvenient. So whenever we flag an error now, it is relative to the entire file (easier to jump to)

### DIFF
--- a/Strata/Languages/Laurel/LaurelCompilationPipeline.lean
+++ b/Strata/Languages/Laurel/LaurelCompilationPipeline.lean
@@ -232,33 +232,8 @@ def translate (options : LaurelTranslateOptions) (program : Program) : IO Transl
   let (core, diags, _, _) ← translateWithLaurel options program
   return (core, diags)
 
-/-- Run `Core.verify` on a translated Core program, returning the verify-phase
-    failure as a **structured** `DiagnosticModel` value (via `.toBaseIO`) rather
-    than throwing it.
-
-    `Core.verify : EIO DiagnosticModel VCResults` carries its error as a
-    `DiagnosticModel` (with byte-offset `fileRange`). Capturing it as an
-    `Except` here is the single point where that structure is preserved; callers
-    decide whether to re-throw it (production) or render it (tests). Mirrors the
-    pre-existing `removeIrrelevantAxioms := .Precise` and `vcDirectory` handling. -/
-private def runVerify (coreProgram : Core.Program) (options : LaurelVerifyOptions)
-    : IO (Except DiagnosticModel VCResults) := do
-  let verifyOptions := { options.verifyOptions with removeIrrelevantAxioms := .Precise }
-  let runner tempDir : IO (Except DiagnosticModel VCResults) :=
-    (_root_.Core.verify coreProgram tempDir (proceduresToVerify := none) verifyOptions).toBaseIO
-  match verifyOptions.vcDirectory with
-  | .none => IO.FS.withTempDir runner
-  | .some p => IO.FS.createDirAll ⟨p.toString⟩; runner ⟨p.toString⟩
-
 /--
 Verify a Laurel program using an SMT solver.
-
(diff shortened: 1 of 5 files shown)


keyboardDrummer (dismissed):
Very nice!

kadirayk (changes requested):
Nice work! Two points:
- It seems previously, a `Core.verify` failure was converted to an `IO.userError` and **thrown**, propagating out of `verifyToVcResults` as an exception. After this change, the error is captured and returned as a diagnostic, while the CLI caller returns success (exit 0). Is this the desired behavior?
- PR should be created against the branch: `eviewed-kbd-will-merge-to-main` instead of `main2`

Scripts/JavaGenTestData.lean
```diff
@@ -0,0 +1,79 @@
+/-
```
keyboardDrummer, line 1: This file shouldn't be here right? Could you rebase the commits from this PR on the target branch?

leo-leesco: Good catch — fixed. Those files (`Scripts/JavaGenTestData.lean`, `LaurelToCBMC.lean`, the CBMC patch, README, etc.) were never touched by this PR's commits; they leaked into the `base...head` diff via a `Merge main into main2` commit (b85c4801b) that rode in on the branch, since this PR's target isn't descended from that `main2` lineage.

I've replayed the branch on top of `reviewed-kbd-will-merge-to-main`, dropping the merge commits. The diff is now exactly the 3 intended files (the file-relative error reporting + opt-in `showLocations`).

leo-leesco:
@kadirayk thanks, indeed you were right, so :
- I reverted the behavior of `verifyToVcResults`
- I added a new `verifyToDiagnosticModelsCapturing` that returns data instead of throwing so we can incorporate the file-relative location; this is only used in tests

The verification pipeline has been extracted into a `runVerify` helper shared by both execution paths

kadirayk (approved):
LGTM, thanks!

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -242,14 +270,11 @@ def verifyToVcResults (program : Program)
...
-      EIO.toIO (fun f => IO.Error.userError (toString f))
-          (_root_.Core.verify coreProgram tempDir .none options)
-    let ioResult ← match options.vcDirectory with
-      | .none => IO.FS.withTempDir runner
-      | .some p => IO.FS.createDirAll ⟨p.toString⟩; runner ⟨p.toString⟩
-    return (some ioResult, translateDiags)
+    match ← runVerify coreProgram options with
+    | .ok ioResult => return (some ioResult, translateDiags)
+    -- Reconstruct the throwing path: stringify the structured error exactly as
+    -- the previous `EIO.toIO (fun f => .userError (toString f))` did.
```
keyboardDrummer, line 276: I don't care for comments about what the code did before :D that's in the past.

Strata/Languages/Laurel/LaurelCompilationPipeline.lean
```diff
@@ -232,8 +232,36 @@ def translate (options : LaurelTranslateOptions) (program : Program) : IO Transl
...
+    (_root_.Core.verify coreProgram tempDir (proceduresToVerify := none) verifyOptions).toBaseIO
+  match verifyOptions.vcDirectory with
+  | .none => IO.FS.withTempDir runner
+  | .some p => IO.FS.createDirAll ⟨p.toString⟩; runner ⟨p.toString⟩
+
 /--
 Verify a Laurel program using an SMT solver.
+
+A verify-phase failure (a type-checking / symbolic-evaluation error) is
+**thrown** as an `IO` exception, exactly as before file-relative reporting was
```
keyboardDrummer, line 260: I don't care for comments about what the code did before :D that's in the past.

keyboardDrummer (approved):

PR #1370, opened 2026-06-12
Port main2 core commits to main

shigoel:
## Summary

Ports all main2-only core commits to Strata `main` as part of the main2 → main sync (post PR #1363).

**35 commits** covering:
- Type system fixes (function typing, type aliases, seq literal)
- Core translator fixes (func decl, cover labels, bv casts, int recursion)
- SMT/verification pipeline (findUnique registry, encoder, decoder, array axiom, stack overflow)
- Laurel/parser (docs, grammar, checkLeftRec, formatting, valueless return)
- Procedures/transforms (Body sum type, transparent procs, inlining, preconditions, Props)
- Misc (cache hash, Maps cleanup, docs, examples)

All commits preserve original authorship and dates.

**Depends on:** [Strata-DDM #2](https://github.com/strata-org/Strata-DDM/pull/2) (StrataDDM dependency temporarily pinned to `port/main2-ddm-fixes` branch — will be changed to `main` after DDM PR merges).

**⚠️ Merge with merge commit (not squash) to preserve individual commit attribution.**

## Test plan

- [x] `lake build` passes (494 jobs)
- [x] `lake test` passes (only unrelated Java jar test fails)
- [x] CI passes
- [x] After DDM PR merges: update `lakefile.toml` to point StrataDDM at `main`


aqjune-aws:
I compared the contents of the files in 1370 with the latest main2 branch (just did diff -r ... ) and it seems files Strata/DL/Lambda/LExprTypeSpec.lean, Strata/DL/Lambda/Denote/LExprResolveAnnotated.lean diverged a bit. Claude said the reason is #1332 ([f2f0fdd](https://github.com/strata-org/Strata/pull/1370/commits/f2f0fdd44d0e21018b029c29dfee349393b7b5ad) in this pull request) which is a clean-up pull request was supposed to be added after #1309 and #1327 , but it was added before them, making the updates lost.

One small question is that if we retain the commit order from main2, probably it will be useful in bisecting which commit is the culprit when our internal benchmark shows regression. It seems the order is somewhat arbitrary - is keeping the original commit order causing conflicts?

aqjune-aws (approved):
I compared the file contents of this PR with the contents in the latest main2 branch and confirmed that there was no suspicious change.

- I confirmed that the updates in EncoderTests.lean of https://github.com/strata-org/Strata/pull/1250 was dropped in main2, then readded in this PR, so it is good!
- Found that this PR doesn't have docs/BooleBenchmarks.md and docs/BooleFeatureRequests.md , but I guess these are for Strata-Boole?
- There were a few updates in StrataTestExtra/Languages/Java/TestGen.lean to reduce usage of StrataDDM and it looked fine.

joehendrix (approved):

PR #1374, opened 2026-06-16
Laurel: avoid quantified modifies frames when using array theory

julesmt:
Under `--use-array-theory`, a `modifies` clause naming only individual object
references is now encoded as a single **quantifier-free heap equation** instead of
a `∀`-quantified frame, so the solver discharges it by array rewriting instead of
instantiating a quantifier per object/field.

    ∀ obj, fld. obj ≠ c ⇒ readField(old,obj,fld) == readField(new,obj,fld)   -- before
    data(new) == store(data(old), c, select(data(new), c))                   -- now

The `∀` frame is unchanged for set-valued/empty `modifies` and when array theory is
off, so default behavior is untouched.

### DIFF
--- a/Strata/Languages/Laurel/HeapParameterization.lean
+++ b/Strata/Languages/Laurel/HeapParameterization.lean
@@ -580,7 +580,7 @@ public def heapParameterizationPass : LaurelPass where
   name := "HeapParameterization"
   documentation := "Transforms procedures that interact with the heap by adding explicit heap parameters. The heap is modeled as `Map Composite (Map Field Box)`. Procedures that write the heap receive both an input and output heap parameter; procedures that only read the heap receive an input heap parameter. Field reads and writes are rewritten to use `readField` and `updateField` functions."
   needsResolves := true
-  run := fun p m =>
+  run := fun _ p m =>
     (heapParameterization m p, [], {})
   comesBefore := [
       ⟨ typeHierarchyTransformPass, "the type hierarchy pass modifies the 'Composite' datatype that is introduced by this pass." ⟩,
--- a/Strata/Languages/Laurel/LaurelPass.lean
+++ b/Strata/Languages/Laurel/LaurelPass.lean
@@ -17,9 +17,11 @@ public section
 structure LaurelTranslateOptions where
   inlineFunctionsWhenPossible : Bool := false
   overflowChecks : Core.OverflowChecks := {}
-  /-- Encode the heap `Map` as the SMT-LIB theory of arrays. Inside the modifies pass
-      this selects the enumerated (quantifier-free) frame for single-reference clauses. -/
-  useArrayTheory : Bool := false
+  /-- Select the enumerated (quantifier-free) modifies frame for clauses that name
+      only single references. This is only beneficial when the heap `Map` is encoded
+      with SMT-LIB array theory, so it is set together with the verifier's
+      `useArrayTheory`. -/
+  enumeratedModifiesClauses : Bool := false
   keepAllFilesPrefix : Option String := none
 
 instance : Inhabited LaurelTranslateOptions where
--- a/Strata/Languages/Laurel/LaurelPass.lean
+++ b/Strata/Languages/Laurel/LaurelPass.lean
@@ -17,10 +17,8 @@ public section
 structure LaurelTranslateOptions where
   inlineFunctionsWhenPossible : Bool := false
   overflowChecks : Core.OverflowChecks := {}
-  /-- Select the enumerated (quantifier-free) modifies frame for clauses that name
-      only single references. This is only beneficial when the heap `Map` is encoded
-      with SMT-LIB array theory, so it is set together with the verifier's
-      `useArrayTheory`. -/
+  /-- Quantifier-free modifies frame for single-reference clauses. Set-valued entries
+      keep the quantified frame. Use with the verifier's `useArrayTheory`. -/
   enumeratedModifiesClauses : Bool := false
   keepAllFilesPrefix : Option String := none
 
--- a/Strata/Languages/Laurel/LaurelPass.lean
+++ b/Strata/Languages/Laurel/LaurelPass.lean
@@ -7,10 +7,23 @@ module
 
 public import Strata.Languages.Laurel.SemanticModel
 public import Strata.Util.Statistics
+public import Strata.Languages.Core.Options
 
 namespace Strata.Laurel
 
 public section
+
+/-- Options controlling Laurel-to-Core translation, threaded into every pass's `run`. -/
+structure LaurelTranslateOptions where
+  inlineFunctionsWhenPossible : Bool := false
+  overflowChecks : Core.OverflowChecks := {}
+  /-- Encode the heap `Map` as the SMT-LIB theory of arrays. Inside the modifies pass
+      this selects the enumerated (quantifier-free) frame for single-reference clauses. -/
+  useArrayTheory : Bool := false
+  keepAllFilesPrefix : Option String := none
+
+instance : Inhabited LaurelTranslateOptions where
+  default := {}
 
 mutual
 structure ComesBefore where
@@ -25,8 +38,8 @@ structure LaurelPass where
   name : String
   /-- Whether `resolve` should be run after the pass. -/
   needsResolves : Bool := false
-  /-- The pass action. -/
-  run : Program → SemanticModel → Program × List DiagnosticModel × Statistics
+  /-- The pass action; `options` carries the translate settings (e.g. the modifies-frame choice). -/
+  run : LaurelTranslateOptions → Program → SemanticModel → Program × List DiagnosticModel × Statistics
   /-- A description of what this pass does, used for documentation generation. -/
   documentation : String
   comesBefore : List ComesBefore := []
--- a/Strata/Languages/Laurel/ModifiesClauses.lean
+++ b/Strata/Languages/Laurel/ModifiesClauses.lean
@@ -90,45 +90,75 @@ def conjoinAll (exprs : List StmtExprMd) : StmtExprMd :=
   | first :: rest => rest.foldl (fun acc e => mkMd <| .PrimitiveOp .And [acc, e]) first
 
 /--
-Build the modifies frame condition as a Laurel StmtExpr.
-
-Generates a single quantified formula:
+Pointwise frame: every allocated object the `modifies` clause does not name keeps
+all of its field values across the call.
 
   forall $obj: Composite, $fld: Field =>
-    notModified($obj) && $obj < $heap_in.nextReference ==> readField($heap_in, $obj, $fld) == readField($heap, $obj, $fld)
-
-Returns `none` if there are no entries.
+    $obj < $heap_in.nextReference && notModified($obj)
+      ==> readField($heap_in, $obj, $fld) == readField($heap, $obj, $fld)
 -/
-def buildModifiesEnsures (proc: Procedure) (model: SemanticModel) (modifiesExprs : List StmtExprMd)
-    (heapInName heapOutName : Identifier) : Option StmtExprMd :=
-  let entries := extractModifiesEntries model modifiesExprs
+def buildPointwiseFrame (proc : Procedure) (entries : List ModifiesEntry)
+    (heapIn heapOut : StmtExprMd) : StmtExprMd :=
   let objName : Identifier := "$modifies_obj"
   let fldName : Identifier := "$modifies_fld"
   let obj := mkMd <| .Var (.Local objName)
   let fld := mkMd <| .Var (.Local fldName)
-  let heapIn := mkMd <| .Var (.Local heapInName)
-  let heapOut := mkMd <| .Var (.Local heapOutName)
-      -- Build the "obj is allocated" condition: Composite..ref($obj) < $heap_in.nextReference
   let heapCounter := mkMd <| .StaticCall "Heap..nextReference!" [heapIn]
   let objRef := mkMd <| .StaticCall "Composite..ref!" [obj]
   let objAllocated := mkMd <| .PrimitiveOp .Lt [objRef, heapCounter]
   let antecedent := if entries.isEmpty
     then objAllocated
     else
-      -- Build the "not modified" precondition from all entries
-      -- Combine: $obj < $heap_in.nextReference && notModified($obj)
       let notModified := conjoinAll (entries.map (buildNotModifiedForEntry obj))
       mkMd <| .PrimitiveOp .And [objAllocated, notModified]
-  -- Build: readField($heap_in, $obj, $fld) == readField($heap, $obj, $fld)
   let readIn := mkMd <| .StaticCall "readField" [heapIn, obj, fld]
   let readOut := mkMd <| .StaticCall "readField" [heapOut, obj, fld]
-  let heapUnchanged := mkMd <| .PrimitiveOp .Eq [readIn, readOut]
-  -- Build: antecedent ==> heapUnchanged
-  let implBody := mkMd <| .PrimitiveOp .Implies [antecedent, heapUnchanged]
-  -- Build: forall $obj: Composite, $fld: Field => ...
+  let fieldUnchanged := mkMd <| .PrimitiveOp .Eq [readIn, readOut]
+  let implBody := mkMd <| .PrimitiveOp .Implies [antecedent, fieldUnchanged]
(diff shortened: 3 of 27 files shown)


Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -584,7 +584,7 @@ public def heapParameterizationPass : LaurelPass where
     (heapParameterization m p, [], {})
   comesBefore := [
       ⟨ typeHierarchyTransformPass, "the type hierarchy pass modifies the 'Composite' datatype that is introduced by this pass." ⟩,
-      ⟨ modifiesClausesTransformPass, "the modifies pass refers to several types and variables introduced by heap parameterization: Composite, Field, $heap_in, $heap." ⟩,
+      ⟨ modifiesClausesTransformPass false, "the modifies pass refers to several types and variables introduced by heap parameterization: Composite, Field, $heap_in, $heap." ⟩,
```
keyboardDrummer: The `false` value here seems arbitrary. Can you refactor the passes so they take the LaurelTranslateOptions in their run method? That way, you won't need to parameterize the passes themselves.

keyboardDrummer:
I found the term "array-theory" in this PR misleading because the PR does not directly use any SMT array theory. Instead, the PR relies on Core Map's which are not necessarily implemented using SMT array theory. In fact, I think Core is or at least was not using array theory for their maps, simply because it is or was not implemented yet.

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -90,45 +90,71 @@ def conjoinAll (exprs : List StmtExprMd) : StmtExprMd :=
...
+`buildHeapEquationFrame` when array theory is on; every other case (set-valued or
+empty `modifies`, or array theory off) uses `buildPointwiseFrame`.
+-/
+def buildModifiesEnsures (proc : Procedure) (model : SemanticModel) (modifiesExprs : List StmtExprMd)
+    (heapInName heapOutName : Identifier) (useArrayTheory : Bool := false) : Option StmtExprMd :=
+  let entries := extractModifiesEntries model modifiesExprs
+  let heapIn := mkMd <| .Var (.Local heapInName)
+  let heapOut := mkMd <| .Var (.Local heapOutName)
+  let onlySingleRefs := !entries.isEmpty && entries.all fun e => match e with | .single _ => true | _ => false
+  if useArrayTheory && onlySingleRefs then
```
keyboardDrummer: Interesting that it only works for single refs. Why can't you make this work for sets as well? I would think you can do that without introducing quantifiers. I think we would only need something like a reduce function on the set.

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -90,45 +90,71 @@ def conjoinAll (exprs : List StmtExprMd) : StmtExprMd :=
...
+procedure may allocate fresh references but never reuses them.
+-/
+def buildHeapEquationFrame (proc : Procedure) (entries : List ModifiesEntry)
+    (heapIn heapOut : StmtExprMd) : StmtExprMd :=
+  let data h := mkMd <| .StaticCall "Heap..data!" [h]
+  let nextRef h := mkMd <| .StaticCall "Heap..nextReference!" [h]
+  let dataOut := data heapOut
+  let modifiedRefs := entries.filterMap fun e => match e with | .single r => some r | _ => none
+  let framedData := modifiedRefs.foldr
+    (fun ref acc => mkMd <| .StaticCall "update" [acc, ref, mkMd <| .StaticCall "select" [dataOut, ref]])
```
keyboardDrummer, line 114: So for a single updated ref, `heapOut := update heapIn ref (select heapOut ref)`. It's feels silly that heapOut occurs on both sides of the equation. Can't we already fill in the new value of the object on the RHS ? Suppose the user also has a postcondition that states what the new field values of some reference are, then the resulting final postcondition will look like:
```
&& heapOut := update heapIn ref (select heapOut ref)
&& readField heapOut ref <f1> := <someExpr1>
&& readField heapOut ref <f2> := <someExpr1>
..
```

That doesn't look very optimal. I think we could enable a different modifies syntax that states how each ref is modified, like:

```
composite SomeObject {
  x: int
}

procedure foo(o: SomeObject)
  modifies o -> o with { x := 3 }
```

and then the resulting postcondition would be:

```
heapOut := update heapIn ref (update (select heapIn ref) xField 3)
```

If you think this is interesting, could you create a GH issue to investigate this?

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -242,12 +268,12 @@ public def filterNonCompositeModifiesPass : LaurelPass where
     (p', diags, {})
 
 /-- Pipeline pass: translate modifies clauses into ensures clauses. -/
-public def modifiesClausesTransformPass : LaurelPass where
+public def modifiesClausesTransformPass (useArrayTheory : Bool := false) : LaurelPass where
```
keyboardDrummer: I think the term `useArrayTheory` is not very accurate. Core Maps, which I think is what you're referring to with array theory here, are used in both methods. Let's try to come up with a better name.

In the existing method, we're iterating over the unchanged objects using a quantifier, and in the new one we're iterating over the changed objects, allowing us to avoid a quantifier since the elements are known.

Can we base the option name on that description of the two methods?

Also, is the new method not just better? Should it be the default? Do we even want to keep this option?

keyboardDrummer (commented):
Very cool!

Strata/Languages/Laurel/ModifiesClauses.lean
julesmt: I agree, the name is off. Both frames use the same Core `Map`; the difference is that the old one quantifies over the unchanged objects and the new one enumerates the changed ones, so I'll rename it that way.

The new frame only wins when maps are arrays, and I don't have a case where it helps with array theory off, so rather than a separate toggle I'll let `--use-array-theory` pick it when a modifies clause names only single references.

Array theory is definitely better on my benchmarks but it affects every VC so I don't think it's up to me to decide.

Strata/Languages/Laurel/ModifiesClauses.lean
julesmt: From what I undestood we can can't make it work with sets.
For modifies a, b, we know exactly which objects may change, so we can encode the frame as a fixed sequence of heap updates.

For `modifies S`, S is symbolic, so we don't know its elements. Expressing that everything outside S is unchanged requires a quantifier, which is exactly what we're trying to avoid. We could open an issue ? @keyboardDrummer

Strata/Languages/Laurel/ModifiesClauses.lean
julesmt: I do find it interesting ! I'll create an issue.

Strata/Languages/Laurel/HeapParameterization.lean
julesmt: agree

Strata/Languages/Laurel/ModifiesClauses.lean
julesmt: https://github.com/strata-org/Strata/issues/1388

keyboardDrummer:

keyboardDrummer:
How come you can't make the enumerated encoding work with procedures that have a body? Suppose I have:

```
composite C { var x: int }
procedure creator() returns (c: C)
  ensures c#x == 3
  ensures fresh(c)
  modifies {}
  opaque
{
  c := new C;
  c#x := 3;
}
```

then that would translate to something like:

```
procedure creator() returns (c: C)
// c#x == 3
  ensures read($heap, c, C.x) == 3 
  
// fresh(c)
  ensures c.ref >= old($heap).nextRef

// modifies {}, enumerated encoding
  ensures $heap.data == old($heap).data 

// implicit postcondition stating that every output object is allocated in the new heap
  ensures c.ref < $heap.nextRef
```

which has the curious implication that `read(old($heap), c, C.x) == 3`, so the old heap contains data for an object that did not yet exist at the time of the old heap. However, I think that might not cause any issues. What issues did you run into?

julesmt:
Yeah, I had pretty much the same example.
**data == old data** fails because the body writes the fresh row, even though fresh objects shouldn't count for modifies.

The pointwise frame dodges this with the `obj < nextReference($heap_in) `guard, which I can't keep without a quantifier, so I this is why I wanted to restrict the enumerated frame to bodiless procedures.

However to get it for bodies too, I split the jobs: callers assume the enumerated frame, and the body is checked separately by asserting the pointwise frame at every exit.

what do you think **(see next commit)** ?  @keyboardDrummer

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -166,39 +140,28 @@ indicating it mutates the heap.
...
-        let enumerate := useEnumeratedFrame && impl.isNone
-        let frameCondition := buildModifiesEnsures proc model modifiesExprs heapName enumerate
-        let postconds' := match frameCondition with
-          | some frame => postconds ++ [{ condition := frame, summary := "modifies clause" }]
-          | none => postconds
-        .ok { proc with body := .Opaque postconds' impl [] }
+        let entries := extractModifiesEntries model modifiesExprs
+        let heapIn := mkMd <| .Old (mkMd (.Var (.Local heapVarName)))
+        let heapOut := mkMd <| .Var (.Local heapVarName)
+        let pointwise := buildPointwiseFrame proc entries heapIn heapOut
```
keyboardDrummer: Consider moving `pointwise` to its usage branch

keyboardDrummer:
> However to get it for bodies too, I split the jobs: callers assume the enumerated frame, and the body is checked separately by asserting the pointwise frame at every exit.
> what do you think **(see next commit)** ? @keyboardDrummer

Brilliant!


>

Strata/Languages/Laurel/LaurelPass.lean
```diff
@@ -7,11 +7,24 @@ module
...
 
 public section
 
+/-- Options controlling Laurel-to-Core translation, threaded into every pass's `run`. -/
+structure LaurelTranslateOptions where
+  inlineFunctionsWhenPossible : Bool := false
+  overflowChecks : Core.OverflowChecks := {}
+  /-- Encode the heap `Map` as the SMT-LIB theory of arrays. Inside the modifies pass
+      this selects the enumerated (quantifier-free) frame for single-reference clauses. -/
+  useArrayTheory : Bool := false
```
keyboardDrummer: Can we change the name `arrayTheory`? Maybe `quantifierFreeModifiesClauses` or `enumeratedModifiesClauses` or `executableModifiesClauses`

Strata/Languages/Laurel/LaurelPass.lean
```diff
@@ -7,11 +7,24 @@ module
...
 namespace Strata.Laurel
 
 public section
 
+/-- Options controlling Laurel-to-Core translation, threaded into every pass's `run`. -/
+structure LaurelTranslateOptions where
+  inlineFunctionsWhenPossible : Bool := false
+  overflowChecks : Core.OverflowChecks := {}
+  /-- Encode the heap `Map` as the SMT-LIB theory of arrays. Inside the modifies pass
+      this selects the enumerated (quantifier-free) frame for single-reference clauses. -/
```
keyboardDrummer: Can you more explicitly state that if a set modifies clause occurs in the procedure, then the entire modifies clause condition stays in the quantified form?

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -90,45 +74,64 @@ def conjoinAll (exprs : List StmtExprMd) : StmtExprMd :=
...
 
   forall $obj: Composite, $fld: Field =>
     notModified($obj) && $obj < old($heap).nextReference ==> readField(old($heap), $obj, $fld) == readField($heap, $obj, $fld)
-
-Returns `none` if there are no entries.
 -/
-def buildModifiesEnsures (proc: Procedure) (model: SemanticModel) (modifiesExprs : List StmtExprMd)
-    (heapName : Identifier) : Option StmtExprMd :=
-  let entries := extractModifiesEntries model modifiesExprs
+def buildPointwiseFrame (proc : Procedure) (entries : List ModifiesEntry)
```
keyboardDrummer: `buildQuantifiedFrame` ?

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -137,35 +140,28 @@ indicating it mutates the heap.
...
-that all allocated objects are preserved between heaps:
-  `forall $obj: Composite, $fld: Field => $obj < old($heap).nextReference ==> readField(old($heap), $obj, $fld) == readField($heap, $obj, $fld)`
-
-If the modifies clause uses a wildcard (`*`), the frame condition is skipped
-entirely — the procedure may modify anything.
--/
+/-- Build and attach `proc`'s modifies frame, then clear the clause. -/
 def transformModifiesClauses (model: SemanticModel)
-    (proc : Procedure) : Except (Array DiagnosticModel) Procedure :=
+    (proc : Procedure) (useEnumeratedFrame : Bool := false) : Except (Array DiagnosticModel) Procedure :=
```
keyboardDrummer: Please remove the default values in the private methods. I think the only default should be in LaurelTranslateOptions

julesmt: Oh yeah sorry, I thought I had modified that, will do

keyboardDrummer:
This will have a few merge conflicts after https://github.com/strata-org/Strata/pull/1352 merges. I'll try to resolve these before you get to it @julesmt

Strata/Languages/Laurel/LaurelPass.lean
```diff
@@ -7,11 +7,24 @@ module
...
 
 namespace Strata.Laurel
 
 public section
 
+/-- Options controlling Laurel-to-Core translation, threaded into every pass's `run`. -/
+structure LaurelTranslateOptions where
+  inlineFunctionsWhenPossible : Bool := false
+  overflowChecks : Core.OverflowChecks := {}
+  /-- Quantifier-free modifies frame for single-reference clauses. Set-valued entries
```
keyboardDrummer, line 19: I think this comment should mention that if the procedure modifies clause contains sets, then this option has no effect.

keyboardDrummer (approved):

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -91,45 +75,64 @@ def conjoinAll (exprs : List StmtExprMd) : StmtExprMd :=
...
+  let refsMonotone := mkMd <| .PrimitiveOp .Leq [nextRef heapIn, nextRef heapOut]
+  { val := .PrimitiveOp .And [dataPreserved, refsMonotone], source := proc.name.source }
+
+/-- True when the `modifies` clause is non-empty and names only individual references
+(no set-valued entries), so the enumerated frame applies. -/
+def onlyIndividualRefs (entries : List ModifiesEntry) : Bool :=
+  !entries.isEmpty && entries.all fun e => match e with | .single _ => true | _ => false
+
+/-- Assert `frame` before every exit of `body` (`return` and body-block `exit`) and at its end. -/
+def insertFrameChecks (proc : Procedure) (frame : StmtExprMd) (body : StmtExprMd) : StmtExprMd :=
```
keyboardDrummer, line 126: If we had postconditions that were checked but not assumed at the call-site, we could use one of those instead of `insertFrameChecks`, which multiplies the modifies clause for each exit branch. I'll make a PR to add this and use it here.

fabiomadge (commented):
Independent soundness pass (built at `5caabd852`, suite green). Sound: callers assume the enumerated frame (free) while the body proves only pointwise; enumerated is abstractly stronger (constrains *unallocated* rows) but the gap is confined to fresh objects the caller never observed (opaque refs + monotone, never-reused `nextReference`) — imprecision, not a derivable falsehood. Verified: unmodified objects preserved, modified ones not over-preserved, wildcard effects still propagate into a caller's frame.

One non-blocking note — **loop + early `return` rejects legal code.** A body writing only its framed object then returning inside a loop fails (`modifies clause does not hold`); the loopless analog verifies — the loop lacks a heap-frame invariant to discharge the `old()` assert. Sound, just imprecise.

fabiomadge (approved):

kadirayk (approved):
IIUC body proves the quantified frame, callers assume the enumerated (quantifier-free) frame, which should be faster. The default case is untouched. Tests seem to cover the changes and tricky paths. LGTM!

PR #1378, opened 2026-06-16
feat(laurel): chained field access and paren-free field increment

### LINKED ISSUE #1371: Support chained field access (`o#inner#count`)
# Support chained field access (`o#inner#count`)

Chained field access on nested composites is unimplemented. `o#inner#count` fails — single-level (`o#inner`) is all that works today. It's a capability gap, not a deliberate restriction (the grammar and `Variable.Field` AST are already recursive; no design doc forbids it).

It breaks across three layers, each with a small fix:

| Layer | Cause | Fix |
|---|---|---|
| Parse | `fieldAccess` (`LaurelGrammar.st:42`) lacks `leftassoc`, so `a#b#c` can't left-recurse | add `leftassoc` (keep `prec(90)`) — matches every other chainable op |
| Resolution | `targetTypeName` (`Resolution.lean:162`) only types `.Var (.Local _)` targets; returns `none` for a nested `.Field` target | add a `.Var (.Field …)` case that resolves the inner target's type, then looks up the field in that type's scope. Terminate structurally via the `Variable.Field` target lemma (`LaurelAST.lean:388`) or fuel (cf. `underlyingBaseType`), not `partial` |
| Heap | the field-**read** arm (`HeapParameterization.lean:280`) embeds `selectTarget` raw, leaking an un-eliminated `FieldSelect`; the **write** arm already recurses | `let selectTarget ← recurseOne selectTarget` before building `readField` (symmetric with the write arm) |

With all three, chained **reads and writes** parse, resolve, and verify; full suite stays green.

## Verified soundness (prototype, cvc5/z3)

- Negative: `assert (o#inner#count) == 7` on an unconstrained field **fails** (no spurious value).
- Aliasing: after `(a#inner)#count := 5`, `assert (b#inner)#count == 0` **fails** (`a#inner`/`b#inner` may alias — frame stays sound).
- Modifies negative: `modifies (o)` does **not** license a write to `o#inner` — frame correctly distinguishes inner vs outer object.
- Depth-3 (`a#b#c#count`) read+write verifies.

## Scope

- Surface syntax needs parens for field refs in a `modifies` clause (`modifies (o#inner)`) — the clause's `refs` is an unbracketed `CommaSepBy StmtExpr`. Pre-existing quirk; out of scope here.
- Composite instance methods are `not-yet-implemented`, so the static-only modifies-frame generation is not a concern for this work.

fabiomadge:
Closes #1371.

Chained field access (`a#b#c`) and paren-free field increment (`c#n++`) were both blocked by the single `fieldAccess` grammar op, so they are fixed together — splitting them would only produce a conflict on that one line.

**Grammar.** `fieldAccess` becomes `@[prec(95), leftassoc]`. Raising the precedence to 95 puts it above the postfix `++`/`--` ops (prec 90), so `c#n++` parses paren-free as `(c#n)++` instead of being a precedence tie; `leftassoc` lets `a#b#c` left-recurse. The two changes are independent and compose on the one op.

**Resolution.** `targetTypeName` is generalized to resolve a chained `.Var (.Field tgt f)` target by recursing on the inner target and looking the field up in that type's scope. The recursion is kept total with a fuel parameter (mirroring the existing `underlyingBaseType`) rather than `partial`. The shared type-scope field lookup is factored into a new `fieldTypeInScope` helper that `incrDecrTargetType` now reuses instead of duplicating.

**Heap parameterization.** The field-**read** arm now recurses into its target, so a nested `FieldSelect` is eliminated rather than reaching a downstream guard and aborting with a `strata-bug`. This single line unblocks both chained reads and chained writes: the write arm already recursed its target, but that recursion bottoms out in the read arm, so neither path worked beforehand (confirmed by ablation).

**Tests.** `T9_ChainedFieldAccess` covers chained read/write, read-after-inner-assign, must-alias, depth-3 (`a#mid#inner#count`), chained reads on both sides of a comparison, and paren-free chained `o#inner#count++`, plus three negative tests (unconstrained read, two-object may-alias, write isolation) that pin the encoding as non-vacuous and frame-sound. `T23b_IncrDecrField` adds paren-free field `++`/`--` and updates its header comment. Full `lake build` and `lake test` pass.

### DIFF
--- a/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
+++ b/Strata/Languages/Laurel/Grammar/LaurelGrammar.lean
@@ -9,7 +9,7 @@ module
 -- Laurel dialect definition, loaded from LaurelGrammar.st
 -- NOTE: Changes to LaurelGrammar.st are not automatically tracked by the build system.
 -- Update this file (e.g. this comment) to trigger a recompile after modifying LaurelGrammar.st.
--- Last grammar change: renamed strConcat token to `^`; added preIncr/preDecr/postIncr/postDecr; `return` value is now Option StmtExpr (supports a valueless return).
+-- Last grammar change: fieldAccess now @[prec(95), leftassoc] (paren-free `c#n++` and chained `a#b#c`).
 public import StrataDDM.AST
 import StrataDDM.BuiltinDialects.Init
 import StrataDDM.Integration.Lean.HashCommands
--- a/Strata/Languages/Laurel/Grammar/LaurelGrammar.st
+++ b/Strata/Languages/Laurel/Grammar/LaurelGrammar.st
@@ -39,8 +39,11 @@ op call(callee: StmtExpr, args: CommaSepBy StmtExpr): StmtExpr => callee "(" arg
 // Instantiation
 op new(name: Ident): StmtExpr => "new " name;
 
-// Field access
-op fieldAccess (obj: StmtExpr, field: Ident): StmtExpr => @[prec(90)] obj "#" field;
+// Field access.
+// prec(95) binds tighter than the postfix incr/decr ops (prec 90) so `c#n++`
+// parses paren-free as `(c#n)++`. leftassoc lets `a#b#c` left-recurse for
+// chained field access.
+op fieldAccess (obj: StmtExpr, field: Ident): StmtExpr => @[prec(95), leftassoc] obj "#" field;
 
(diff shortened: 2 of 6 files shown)


fabiomadge:
Recreating with the branch hosted on strata-org (not the fork) so the stacked compound-assignment PR can base on it within the same repo.

PR #1381, opened 2026-06-17
fix(laurel): Run LiftInstanceProcedure before EliminateValueInReturns

### LINKED ISSUE #1385: LiftInstanceProcedures drops source provenance of lifted instance methods (skips verification)
## Summary

`liftInstanceProcedures` drops the **source provenance** of an instance procedure's name when it lifts the method to a top-level `<Composite>$<method>` static procedure. The lifted procedure ends up with `name.source = none`, even though the original instance procedure had a real source location.

This is a soundness gap for any consumer that classifies procedures by their originating file in order to decide *which procedures to verify*: lifted instance-method bodies get misclassified and are silently **never verified**.

- File: `Strata/Languages/Laurel/LiftInstanceProcedures.lean`
- Branch: `reviewed-kbd-will-merge-to-main` (the pass is new on this branch; bug present at `0c993543c12d7735156d61556161fcb1ffda9c08`)

## Root cause

```lean
def liftedProcName (typeName methodName : Identifier) : Identifier :=
  mkId s!"{typeName.text}${methodName.text}"
```

`mkId` builds `{ text := name }` with `source := none`, so the lifted name discards `methodName.source`. The lifted `Procedure`'s only provenance lives on its `name : Identifier` (there is no separate metadata field), and `LaurelToCoreTranslator` derives the Core `Decl`'s `FileRange` from `proc.name` (`identifierToCoreMd proc.name`). So once lifted, the procedure has no `FileRange`.

## Impact

A downstream consumer that partitions "user" vs "prelude" procedures by `FileRange` (e.g. the Python frontend's `splitProcNames`, used to pick verification targets) treats every lifted instance method as prelude. The method body is then never selected for verification.

Concretely, in the Python→Laurel pipeline a method like:

```python
class Greeter:
    def greet(self) -> str:
        assert self.name != "", "name must not be empty"
        return self.name
```

translates to an instance procedure with an in-body `assert`. After lifting, that `assert` (and the method's `ensures`) silently disappear from the verification results — the analysis reports success without ever checking them.

## Reproduction (Laurel only)

Standalone regression test (no Python frontend needed). Construct a composite with one instance procedure whose name carries a known source, run `resolve` + `liftInstanceProcedures`, and check the lifted name's source:

```lean
import Strata.Languages.Laurel.LiftInstanceProcedures

open Strata
open Strata.Laurel

private def methodSource : FileRange :=
  { file := .file "Counter.lr", range := SourceRange.none }

private def resetProc : Procedure :=
  { name := { text := "reset", source := some methodSource }
    inputs := [{ name := mkId "self",
                 type := { val := .UserDefined (mkId "Counter"), source := none } }]
    outputs := [], preconditions := [], decreases := none,
    isFunctional := false, body := .Opaque [] none [] }

private def counterComposite : CompositeType :=
  { name := mkId "Counter", extending := [],
    fields := [{ name := mkId "count", isMutable := true,
                 type := { val := .TInt, source := none } }],
    instanceProcedures := [resetProc] }

private def program : Program :=
  { staticProcedures := [], staticFields := [],
    types := [.Composite counterComposite], constants := [] }

private def liftedReset : Option Procedure :=
  (liftInstanceProcedures (resolve program).model program).staticProcedures.find?
    (·.name.text == "Counter$reset")

#guard resetProc.name.source == some methodSource                 -- original has a source
#guard liftedReset.isSome                                         -- the method was lifted
#guard (liftedReset.bind (·.name.source)) == some methodSource    -- FAILS: lifted source is `none`
```

**Observed (current branch):** the third `#guard` fails; the lifted name's source is `none`.
**Expected:** the lifted procedure keeps the originating file (`Counter.lr`).

A polished version following the existing `*Test.lean` conventions is included in the accompanying PR as `StrataTest/Languages/Laurel/LiftInstanceProceduresProvenanceTest.lean`.

## Suggested fix

Preserve the member's source when building the lifted name:

```lean
def liftedProcName (typeName methodName : Identifier) : Identifier :=
  { mkId s!"{typeName.text}${methodName.text}" with source := methodName.source }
```

With this change the reproduction passes and lifted instance-method bodies are again selected for verification.

## Environment

- Branch `reviewed-kbd-will-merge-to-main` @ `0c993543c12d7735156d61556161fcb1ffda9c08`
- Lean toolchain `leanprover/lean4:v4.29.1`

olivier-aws:
EliminateValueInReturns only folded over program.staticProcedures, so a value-returning instance method with a body (using `return expr`) was never rewritten into `outParam := expr; return`.  Updated the order of passes to run the `LiftInstanceProcedure` pass before `EliminateValueInReturns`.

Add T9_ValueReturningInstanceMethods.lean: 9 positive cases (field/ computed/parameterized returns, conditional/early returns, modifies + return, bool return, shared method names, chained receiver, local var) and 2 negative cases pinning the no-output and multiple-output valued return diagnostics, proving the pass now reaches instance procedures.

- Strata/DDM/*
- Strata/Languages/Boole/*
- Tools/BoogieToStrata


Strata/Languages/Laurel/EliminateValueInReturns.lean
```diff
@@ -83,13 +83,29 @@ def eliminateValueReturnsInProc (proc : Procedure) : Procedure × Array Diagnost
 
 public section
 
-/-- Transform a program by eliminating value returns in all imperative procedures. -/
+/-- Transform a program by eliminating value returns in all imperative procedures,
```
keyboardDrummer: Can't we lift the instance procedure before running this pass?

keyboardDrummer (changes requested):

Strata/Languages/Laurel/EliminateValueInReturns.lean
olivier-aws: We could, the added test passes with that (except some errors on localization). But this creates a dependency order between these two passes that is not really required. And the generalization of EliminateValueInReturns to all methods seems legit to me.

keyboardDrummer: > But this creates a dependency order between these two passes that is not really required.

Changing the pass order is cheap, so why not do that? Unless there is a cycle, I don't mind having the pass dependency. I think the less code we add the less we have to maintain and eventually prove.

IMO, we should strive for encapsulating features when possible. With this PR, the instance procedure feature becomes part of 2 compilation components instead of one. Think of instance procedures as an extension to a smaller language rather than being part of the base.

fabiomadge (commented):
Reviewed for correctness; reproduced the StrataBug on base and verified the fix + tests through the pipeline. The fix is correct and the test file is thorough. One comment-accuracy fix below, and I added a note on the open pass-ordering thread.

Strata/Languages/Laurel/EliminateValueInReturns.lean
```diff
@@ -83,13 +83,29 @@ def eliminateValueReturnsInProc (proc : Procedure) : Procedure × Array Diagnost
...
+    including instance procedures declared inside `composite` blocks.
+
+    Instance procedures are processed here (rather than after they are lifted to
+    static scope by `LiftInstanceProcedures`) because this pass must run before
+    `HeapParameterization`, which itself runs before lifting. -/
```
fabiomadge: Rationale is backwards — order is `eliminateValueInReturns → lift → heapParameterization`, so lifting precedes `HeapParameterization`. Real reason: this pass runs before `LiftInstanceProcedures` (as the test docstring says).
```suggestion
    Instance procedures are processed here, while they still live inside their
    composite, because this pass runs before `LiftInstanceProcedures` lifts
    them to static scope. -/
```

Strata/Languages/Laurel/EliminateValueInReturns.lean
fabiomadge: Tested the reorder (worktree): fixes the StrataBug — 9 positives pass, but the 2 negatives regress to `1:0-0`. That regression is shallow, though: `liftedProcName` (`LiftInstanceProcedures.lean:40`) drops source via `mkId`. One-liner:
```lean
{ mkId s!"{typeName.text}${methodName.text}" with source := methodName.source }
```
With that + reorder: 11/11 at the right location, `StrataTest` green, no traversal code. Worth doing regardless — it also fixes two live `1:0-0` mis-localizations today (instance `function` with a postcondition, and with a constrained return), both verified by building.

So: "in-place traversal" vs "reorder + 1-line fix (also fixes 2 live bugs)" — your call.

olivier-aws: Agree, this fix is in https://github.com/strata-org/Strata/pull/1386 as well but would need to be included in this PR if we decide to go with the reordering

keyboardDrummer (dismissed):

keyboardDrummer (approved):

fabiomadge (approved):

PR #1383, opened 2026-06-17
feat(laurel): add C-style compound assignment (+= -= *= /= %= ^=)

fabiomadge:
Adds C-style compound assignment to the Laurel dialect (`+=` `-=` `*=` `/=` `%=`, plus `^=` for string concatenation — Laurel's `^`, not bitwise XOR). `x op= e` lowers to `x := x op e`, yields the new value, and is usable both as a statement and in expression position. It generalizes the `++`/`--` machinery from #1320 — compound assignment is the general case of which `++`/`--` is the degenerate `x op= 1` form.

**AST.** A new `StmtExpr.CompoundAssign (op) (target) (rhs)` constructor, distinct from `.IncrDecr` (which carries a mode and no RHS).

**Grammar.** The six operators are added at prec 10 with `rightassoc`, so `a += b += c` groups right. Mixed `:=`/`op=` chaining is asymmetric: `x := y += z` does not parse (matching plain `x := y := z`), while `x += y := z` does — use parentheses to be explicit.

**Lowering.** `lowerToAssign` is generalized to take an arbitrary operator and RHS, shared with the `++`/`--` path. The lowered `.PrimitiveOp` keeps `skipProof := false`, so `/=` and `%=` carry the same division-by-zero obligation as a hand-written `x := x / e`. It runs first in the pipeline, so no later pass observes `.CompoundAssign`.

**Resolution / type checking.** Resolution is handled by a new `Synth.compoundAssign` rule in the bidirectional `Synth`/`Check` framework, modelled on `Synth.incrDecr`/`Synth.assign`: it resolves the target, checks the target's *element type* against the operator via `checkCompoundAssignTargetType`, then checks the RHS against the target type with `Check.resolveStmtExpr` (so a mismatched RHS is reported by the standard `expected 'T', got 'U'` rule — no bespoke RHS pass needed).

The target-type check is construct-keyed by design: `++`/`--` reject `real` (their synthesized int `1` cannot type-check against a real under Laurel's strict numeric typing, which has no int→real coercion), but `+= -= *= /=` accept `real` because the RHS is user-written; `%=` is int-only (`.Mod` has no real lowering); `^=` is string-only.

**Tests.**
- `T24_CompoundAssign` — end-to-end verification: every operator on int; `real` for `+= -= *= /=`; `string` for `^=`; constrained-int (`nat`); integer truncation (`7 /= 2`, `7 %= 3`); expression-position yielded value for `+=` and non-`+=` (`-=`, `^=`); right-associative `a += b += c`; and the same on composite-field and chained-field (`o#inner#count`) targets, including real/string/`/=` on fields.
- `CompoundAssignTypeRejectionTest` — per-operator target rejections (`^=` on int, `+=` on string/bv/float64, `%=` on real); RHS-type mismatches (`expected 'T', got 'U'`); a real-based constrained type rejected by `%=` (constraint peels to `real`); the `r += 1.0`-accepts / `r++`-rejects asymmetry; and `/= 0` / `%= 0` division-by-zero verification failures.
- `CompoundAssignLiftTest` — golden (`#guard_msgs`) lowering for statement, expression, and one-level nested-RHS forms.

Full `lake build` and `lake test` pass.

### DIFF
--- a/Strata/Languages/Laurel/EliminateIncrDecr.lean
+++ b/Strata/Languages/Laurel/EliminateIncrDecr.lean
@@ -11,11 +11,12 @@ import Strata.Languages.Laurel.MapStmtExpr
 import Strata.Util.Tactics
 
 /-!
-# Eliminate Increment / Decrement
+# Eliminate Increment / Decrement and Compound Assignment
 
-Lowers `.IncrDecr` nodes (Java-style `++x`, `x++`, `--x`, `x--`) into
+Lowers `.IncrDecr` nodes (Java-style `++x`, `x++`, `--x`, `x--`) and
+`.CompoundAssign` nodes (C-style `x += e`, `-=`, `*=`, `/=`, `%=`, `^=`) into
 existing Laurel constructs. Runs before `LiftImperativeExpressions` so that
-later passes never see `.IncrDecr`.
+later passes never see either node.
 
 The lowering, parameterised on `op` (`Incr | Decr`) and `mode` (`Pre | Post`):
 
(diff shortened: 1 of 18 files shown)


Strata/Languages/Laurel/EliminateIncrDecr.lean
keyboardDrummer: Could you update the pass documentation and name?

Also, could you add a comment that says that while we could first eliminate IncrDecr by mapping to compound assignment, and then eliminating compound assignment, we're doing both in one pass for the sake of performance?

keyboardDrummer (commented):
Nice!

Strata/Languages/Laurel/EliminateIncrDecr.lean
fabiomadge: Updated the documentation and renamed some things

keyboardDrummer: Hmm. Wouldn't you say that if we'd fully prioritize readability over performance, then we'd make this two passes? First translate `++x` to `x += 1` and then `x += 1` to `x = x + 1` ?

keyboardDrummer (approved):

Strata/Languages/Laurel/EliminateIncrDecr.lean
fabiomadge: One aspect of readability is also having less to read, but I agree, it's a balance

PR #1387, opened 2026-06-18
Add downstream-check CI and shared composite actions

shigoel:
Advisory cross-repo CI: when a Strata PR is marked ready for review (and on every subsequent push), build each repo that depends on Strata — **Strata-Boole**, **Strata-Python**, **Strata-CLI** — against the PR's code, so we catch breakage before it lands on `main`. Results appear as per-downstream jobs on the PR (`Downstream / Strata-Boole`, …). **Non-blocking** — advisory signal, not a merge gate.

## How it works

For each downstream, one matrix job:
1. Checks out the PR's Strata and clones the downstream repo.
2. Rewrites the downstream's `require "Strata"` to a local `path` pointing at the PR checkout (via the `rewrite-require` composite action). This is **fork-safe** — it builds the PR's actual code without needing the head SHA to be fetchable from the remote, so it works for fork PRs too.
3. `lake update Strata` + build + test (Boole/Python run their `lake test` suites; CLI has no test driver, so it mirrors its own CI: build, then `lake exe strata --help` + `run_examples.sh`).

Runs only on `pull_request` for non-draft PRs. Building PR code happens in GitHub's isolated, unprivileged `pull_request` context (read-only token, PR-scoped cache), which is the safe way to execute untrusted contributions.

This PR also adds the `rewrite-require` composite action (+ `scripts/rewrite_require.py`), reused by the sibling downstream-check workflows in the other upstream repos.

## Dependency graph

```
StrataDDM → Strata → {Strata-Boole, Strata-Python, Strata-CLI} → Strata-Boogie
                Strata-Python → Strata-CLI
```

This PR covers the **Strata →** edges. Sibling PRs add the checks for the other upstream repos (Strata-DDM builds Strata; Strata-Python builds Strata-CLI; Strata-CLI builds Strata-Boogie). They reference the composite action added here, so **this PR should land first.**

Strata-Boole and Strata-Boogie get no workflow — they're leaves (nothing depends on them).

## Known follow-ups (feedback welcome)

- **CI-recipe drift:** each downstream's build+test recipe is currently inlined in the check. **Proposed next step:** extract each downstream's build+test core into a composite action *owned by that downstream repo*, so its own `ci.yml` and this check share one source of truth and can't drift. Holding off pending alignment on this approach — **thoughts welcome.**
- **Pin downstreams to SHAs:** downstreams currently track `rev = "main"`. Pinning + an automated bump PR would make the advisory-only model fully honest. Deferred.
- **Redundant upstream builds:** each downstream job rebuilds Strata (+ StrataDDM) from scratch. Should build the upstream once and share it across jobs. Not done yet.

Draft for early feedback.


.github/scripts/rewrite_require.py
robin-aws, line 1: Definitely don't love this approach since a) it's potentially-brittle grepping through config files, and b) it would need to be expanded to support lakefile.lean files as well.

Here are two alternatives (which I should have thought of earlier!)

1. Generate a [package overrides file](https://lean-lang.org/doc/reference/latest/Build-Tools-and-Distribution/Lake/#package-overrides) to point references to the package to the local path.
2. Create a local reservoir-style image of the modified package as if it was deployed there, and set the `RESERVOIR_API_URL` environment variable to `file:///...`

1 is probably the way to go. I mention 2 for completeness (and because I think we can use it for other purposes).

PR #1389, opened 2026-06-18
Fix resolution errors during phases

keyboardDrummer:
### Changes
- Report resolution errors that occur after transformation passes during the compilation of Laurel to Core. This change triggers errors in existing tests that lead to the remainder of the changes in this PR.
- Remove `.THeap` and `.TTypedField`. The usages have been replaced by `.UserDefinedType "Heap"` and `.UserDefinedType "Field"`, so that all references to the heap type use that form, preventing type equality errors.
- Only re-resolve after HeapParam, TypeInference and ModifiesClauses passes all complete, since they are part of the same logical pass. In the future, we should refactor so they're actually one pass, but still three components, since TypeHierarchy and ModifiesClauses are features that built on top of Composite types.
- Replace all reference types with `Composite` during the TypeInference pass. This also enables a slight simplification of LaurelToCoreTranslator.
- Correct a source location usage in the lifting pass. Without this change, the reported location for an error would change, and this is detected during re-resolution.
- In `Resolution.lean`, synth instead of void-check non-last block elements. This is required because some passes create non-void non-last block elements. In particular the `EliminateIncrDecr` pass creates them. This change also means that `Resolution.lean` no longer needs to special-case checking `| .StaticCall .. | .InstanceCall .. | .IncrDecr `.
- Because of the previous change, `Resolution.lean` needs to support synthing for any StmtExpr, which it now does.
- In the resolver, add special casing for calls to select/update/const. These are polymorphic procedures which we can't type check without special casing yet, since we don't support polymorphism yet.
- Remove expecting the body of a function to correspond to its result type. This check failed for more complicated functions from `T3_ControlFlow` after they were lowered by some passes. Note that function will be removed entirely. In the test related to this removed check, I have changed the function into a procedure and the same diagnostic is reported then, although through a different mechanism.
- Refactor TypeAliasElim so it uses a generic traversal to update types. This generic traversal was added a part of changes to TypeHierarchy

### Testing

Added one test, but mostly this PR relies on the existing tests.


Strata/Languages/Laurel/Resolution.lean
```diff
@@ -725,12 +714,24 @@ def Synth.resolveStmtExpr (exprMd : StmtExprMd) : ResolveM (StmtExprMd × HighTy
...
+    return (r, ⟨ .TVoid, source ⟩)
+  | .Return val => do
+    let r ← Check.return exprMd val source (by rw [h_node])
+    return (r, ⟨ .TVoid, source ⟩)
+  | .Assert ⟨condExpr, summary, free⟩ => do
+    let r ← Check.assert exprMd condExpr summary free source (by rw [h_node])
+    return (r, ⟨ .TVoid, source ⟩)
+  | .Assume cond => do
+    let r ← Check.assume exprMd cond source (by rw [h_node])
+    return (r, ⟨ .TVoid, source ⟩)
```
leo-leesco, line 734: Can you give an example where we would need to synthesize the type of these constructs ?

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -1127,48 +1104,41 @@ def Check.return (exprMd : StmtExprMd)
...
+    which is then discarded. This means any expression is accepted in
+    statement position — the `f(x);` idiom works regardless of `f`'s
+    return type, and `x++;` is admitted even though `++` synthesizes the
+    target's type.
 
     This is the single definition of "what counts as a statement". It is
     used by `Check.block` for every non-last statement, and for the last
     statement when the block itself sits in statement position
     (`expected = TVoid`). -/
 def Check.statement (s : StmtExprMd) : ResolveM StmtExprMd := do
```
fabiomadge, line 1140: Non-blocking: non-last block elements are now synthed, not checked at `TVoid`, so a stray value like `5;` is silently accepted (verified) — drops the dead-code lint. Needed for EliminateIncrDecr, but intentional that it's this broad?

fabiomadge (approved):

Strata/Languages/Laurel/Resolution.lean
keyboardDrummer: Yes it's intentional

kadirayk (approved):
LGTM. Reporting errors that were silently dropping after transformation steps. Giving each type a single consistent name (and grouping the related steps) is a nice clean up.

Strata/Languages/Laurel/Resolution.lean
keyboardDrummer: If they occur in a non-last block statement. From the PR details:
1. In Resolution.lean, synth instead of void-check non-last block elements. This is required because some passes create non-void non-last block elements. In particular the EliminateIncrDecr pass creates them. This change also means that Resolution.lean no longer needs to special-case checking | .StaticCall .. | .InstanceCall .. | .IncrDecr .
2. Because of the previous change, Resolution.lean needs to support synthing for any StmtExpr, which it now does.

Example:
Input program:
```
{
  x--;
  var x: int := 3;
  x
}
```

After the incrDecr pass:

```
{
  (x := x - 1) + 1; // this was not accepted before, since all non-last block members has to be void, so we had to make change 1.
  var x: int := 3; // this was not accepted after change 1, so we had to make change 2
  x
}
```

leo-leesco: Thank you for this example ! Semantically, I don't think the problem lies with the typechecker, rather with the translation : in this position, `x++` really means increment `x` rather than return the increased value.
The thing I am more surprised about : why is `var x: int := 3` having any type at all ?

Can we take a look at the `EliminateIncrDecr` pass, and figure out if there is a more idiomatic way to resolve this conflict ? Here is a possible solution in this particular case : enforce `++x`/`--x` in statement position

leo-leesco: We can probably find a middle ground that allows us to not have to synthesize constructs as much as possible ; what seems to be the biggest issue here is mostly that we are trying to typecheck within a block in a very generic manner.
This is not suitable because we lack the necessary precision. Proposed structural changes are in PR #1398

PR #1396, opened 2026-06-19
fix(laurel): synthesizing heap well-formedness conditions in HeapParameterization

AD1024:
A `modifies` clause on Laurel procedure generates a frame condition that preserves every heap allocated objects besides the modifies set, but there is no condition (before this PR) asserting that the objects (passed by input parameters that are of some composite types) are allocated on the heap. This makes frame conditions cannot be discharged under certain cases (e.g., calling a procedure with a `modifies` clause consecutively). 

This PR adds 3 synthesized well-formedness conditions of heap:
1. input-allocation precondition: this asserts each composite parameters are allocated on the input heap
2. monotonicity of heap counter: the top-of-heap allocation counter does not decrease
3. output-allocation post-conditions: each composite output must be allocated on the output heap

Added a new regression test (`T11_HeapWfConditions`).


Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -471,6 +471,55 @@ where
...
+  | .UserDefined name => !isDatatype model name
+  | _ => false
+
+/-- Wrap a `StmtExpr` with a source location for downstream VC labeling. -/
+private def mkMdAt (source : Option FileRange) (e : StmtExpr) : StmtExprMd :=
+  { val := e, source }
+
+/-- For each composite parameter `p`, the precondition
+    `Composite..ref!(p) < Heap..nextReference!(heapVar)` (`p` is allocated) -/
+private def heapWellFormednessPreconds (model : SemanticModel)
```
keyboardDrummer, line 491: I think this only works for composite values that are not wrapped in a container. I think a more general approach would be a precondition like `forall c: Composite :: c.ref < heap.nextRef`.

However, my gut feeling is that if we can avoid something like that, our system can remain simpler. An alternative would be that we require opaque procedures to declare what references they create, so that the modifies clause does not depend on a timestamp.

If I revert all the changes to HeapParameterization, and remove the timestamp check from the modifies postcondition generates by ModifiesClasses.lean, then the tests you added all pass.

I want to investigate whether we can resolve the issue this PR solves purely through changes in ModifiesClauses.lean. I will see if https://github.com/strata-org/Strata/pull/1374 can help with that.

keyboardDrummer (dismissed):

Strata/Languages/Laurel/HeapParameterization.lean
keyboardDrummer: I've confirmed that the T11 test from this PR passes with the new modifies encoding from https://github.com/strata-org/Strata/pull/1374. I will push the T11 test to that PR.

keyboardDrummer: https://github.com/strata-org/Strata/pull/1374 currently has the limitation that it's encoding can only be applied to procedures with a modifies clause but no body, since the body can allocate fresh objects and it fails to prove the modifies clause in those cases.

I'm afraid I would like to take a little more time to come up with an ideal solution. How blocked are you currently?

keyboardDrummer:
@AD1024 I've added a test-case `wrappedExistingObjectsAreAllocated` and some changes to enable it to pass. Could you check if you're OK with these changes?

Strata/Languages/Laurel/HeapParameterization.lean
keyboardDrummer: After more consideration, I think we can't get away with not having the precondition, and still get the free property that existing objects are disjunct from newly created objects.

AD1024: Hi Remy. Thanks for your review. For immediate blocking cases: I'm currently working on verification of coroutine procedures, which can mutate the shared heap. I think in that case the procedure contract will have a modifies clause with a transparent body (since we need to reason about the yields and resumes).

AD1024:
> @AD1024 I've added a test-case `wrappedExistingObjectsAreAllocated` and some changes to enable it to pass. Could you check if you're OK with these changes?

Yes, looking good to me. Thanks!

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -471,6 +471,38 @@ where
       simp_all
       omega)
 
+/-- Wrap a `StmtExpr` with a source location for downstream VC labeling. -/
+private def mkMdAt (source : Option FileRange) (e : StmtExpr) : StmtExprMd :=
```
keyboardDrummer: Do you prefer having this method over having it inlined? It's unclear to me what's it's adding

StrataTest/Languages/Laurel/Examples/Objects/T7_InstanceProcedures.lean
```diff
@@ -219,6 +219,7 @@ procedure useOuter()
   opaque
 {
   var o: Outer := new Outer;
+  o#inner := new Inner;
```
keyboardDrummer, line 222: What happens without the change? If that produces an error, should we test the presence of that error somewhere as well?

Strata/Languages/Laurel/HeapParameterization.lean
AD1024: Added this for generating expressions with tracking of source locations.

StrataTest/Languages/Laurel/Examples/Objects/T7_InstanceProcedures.lean
AD1024: Without allocation the well-formed mess would fail. We want a projection into an allocated composite object here.

Strata/Languages/Laurel/HeapParameterization.lean
keyboardDrummer: Right, but is `mkMdAt source (<expr>)` easier to read than `⟨ <expr>, source ⟩` ?

StrataTest/Languages/Laurel/Examples/Objects/T7_InstanceProcedures.lean
keyboardDrummer: Could you add a test like that to your new T11 file?

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -471,6 +471,38 @@ where
...
+  { condition := valid, summary := some "input heap is valid", free := true }
+
+/-- The postcondition
+    `Heap..nextReference!($heap_in) <= Heap..nextReference!($heap)` -
+    the top of heap pointer never decreases. -/
+private def heapMonotonicityPostcond (source : Option FileRange)
+    (heapVar : Identifier) : Condition :=
+  let inCounter  := ⟨ .Old ⟨ .StaticCall "Heap..nextReference!" [mkMdAt source (.Var (.Local heapVar))], source⟩,  source ⟩
+  let outCounter := ⟨ .StaticCall "Heap..nextReference!" [mkMdAt source (.Var (.Local heapVar))], source ⟩
+  { condition := mkMdAt source (.PrimitiveOp .Leq [inCounter, outCounter]),
```
keyboardDrummer: Can you make this a free postcondition? I think it is guaranteed to hold by construction. That may also resolve the Python test failures

StrataTest/Languages/Laurel/Examples/Objects/T7_InstanceProcedures.lean
AD1024: Sounds good

keyboardDrummer (changes requested):
Actually I think this [245a76b](https://github.com/strata-org/Strata/pull/1396/commits/245a76b6224a5c53d53d705d79990bf4e7ac27e2) change is unsound.

We can go back to the solution you had at first, although it is incomplete. We could make it more complete by defining axioms for custom types, that state that if the container is allocated in a heap, then so are its elements.

Could we take your initial solution, make the added pre and postcondition free, and leave a comment stating that it currently is not as complete as we'd like it to be: the implicit "is allocated" property is only available for reference values that occur directly as parameter values, but not for indirectly reachable reference values.

AD1024:
> Actually I think this [245a76b](https://github.com/strata-org/Strata/pull/1396/commits/245a76b6224a5c53d53d705d79990bf4e7ac27e2) change is unsound.
> 
> We can go back to the solution you had at first, although it is incomplete. We could make it more complete by defining axioms for custom types, that state that if the container is allocated in a heap, then so are its elements.
> 
> Could we take your initial solution, make the added pre and postcondition free, and leave a comment stating that it currently is not as complete as we'd like it to be: the implicit "is allocated" property is only available for reference values that occur directly as parameter values, but not for indirectly reachable reference values.

"*defining axioms for custom types, that state that if the container is allocated in a heap, then so are its elements*" - how would this operate on recursive composites? Are we going to rejecting a definition that looks like `composite ListNode { var v: int, next: ListNode }`? I think even without those axioms, we need to have an mechanism to handle (reject or convert to a recursive type) it.

keyboardDrummer:
> how would this operate on recursive composites?

Can you help me understand the additional complexity with those? Suppose we have an axiom of the form 

```
// member allocatedness axiom
axiom forall heap :: Heap, c: Composite :: isType(c, ListNode_Type) && isAllocated(heap, c) ==> 
  isAllocated(heap, read(heap, c, Composite#next))

function isAllocated(heap: Heap, c: Composite) => c.ref < heap.nextRef
```

What's the issue?

MikaelMayer (commented):
🤖🔍 The synthesized heap well-formedness conditions are a reasonable approach to making frame conditions discharge across sequential opaque calls. The architecture of free pre/postconditions for `heapIsValid` is sound — it establishes the allocation invariant without imposing proof obligations at boundaries where the invariant is maintained by construction.

One observations below on diagnostic regression and style.

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -471,6 +471,38 @@ where
...
+  let heapRead := mkMdAt source (.Var (.Local heapVar))
+  let valid := mkMdAt source (.StaticCall "heapIsValid" [heapRead])
+  { condition := valid, summary := some "input heap is valid", free := true }
+
+/-- The postcondition
+    `Heap..nextReference!($heap_in) <= Heap..nextReference!($heap)` -
+    the top of heap pointer never decreases. -/
+private def heapMonotonicityPostcond (source : Option FileRange)
+    (heapVar : Identifier) : Condition :=
+  let inCounter  := ⟨ .Old ⟨ .StaticCall "Heap..nextReference!" [mkMdAt source (.Var (.Local heapVar))], source⟩,  source ⟩
```
MikaelMayer: 🤖🔍 Nit: This function mixes raw anonymous constructors `⟨ .Old ⟨...⟩, source ⟩` with calls to `mkMdAt` (for the inner argument), making it harder to read than the sibling helpers. Consider using `mkMdAt` consistently:

```lean
  let heapNextRef := mkMdAt source (.StaticCall "Heap..nextReference!" [mkMdAt source (.Var (.Local heapVar))])
  let inCounter  := mkMdAt source (.Old heapNextRef)
  let outCounter := heapNextRef
```

Also: extra whitespace before `:=` and between `source⟩,  source`.

AD1024:
> > how would this operate on recursive composites?
> 
> Can you help me understand the additional complexity with those? Suppose we have an axiom of the form
> 
> ```
> // member allocatedness axiom
> axiom forall heap :: Heap, c: Composite :: isType(c, ListNode_Type) && isAllocated(heap, c) ==> 
>   isAllocated(heap, read(heap, c, Composite#next))
> 
> function isAllocated(heap: Heap, c: Composite) => c.ref < heap.nextRef
> ```
> 
> What's the issue?

I think the main issue is that suppose a recursive composite is defined as such, a `new` clause on `var n: ListNode = new ListNode` will add the fact that `n` is allocated. This fact will make the "allocatedness" of its `next` (and all the nested nodes)` also provable, even though this cannot happen in practice (which would require an infinite heap). In general I think we should reject the definitions that violates the positivity property.

MikaelMayer (commented):

keyboardDrummer (dismissed):

keyboardDrummer:
> I think the main issue is that suppose a recursive composite is defined as such, a `new` clause on `var n: ListNode = new ListNode` will add the fact that `n` is allocated. This fact will make the "allocatedness" of its `next` (and all the nested nodes)` also provable, even though this cannot happen in practice (which would require an infinite heap).

If there's a cycle then it wouldn't require an infinite heap right? We shouldn't block anything that could be a valid program.

MikaelMayer (commented):

keyboardDrummer (dismissed):

MikaelMayer (commented):

StrataTest/Languages/Laurel/Examples/Objects/T11_HeapWfConditions.lean
```diff
@@ -0,0 +1,91 @@
...
+`setIt(c, x) modifies c` is called once and twice from a caller that
+also passes a disjoint `d` and asserts `d` is unchanged. The frame
+condition of `setIt` only quantifies over objects allocated in the
+input heap and distinct from `c`, so two facts must reach the caller
+for the assertions to discharge:
+
+  * `Composite..ref!(d) < Heap..nextReference!($heap_in)` — every
+    composite is allocated in the input heap. This follows from the
+    synthesized `heapIsValid($heap)` precondition supplied by
+    `HeapParameterization` (a `forall` over composites).
```
MikaelMayer: **Stale comment.** Describes a `heapIsValid($heap)` precondition implemented as `a `forall` over composites` — that was the earlier design. The current PR synthesizes a per-parameter `Composite..ref!(p) < Heap..nextReference!(heap)` clause for each composite input. Suggest updating L16-19 to describe the current mechanism.

AD1024:
> If there's a cycle then it wouldn't require an infinite heap right? We shouldn't block anything that could be a valid program.

Yeah, I think you're right - we can assume the allocatedness of inner composite.
I have some notes on that: I was thinking around the semantics of unallocated composite. For verification, we do `havoc` for them, but it assumes that the type is inhabited. In a case where a recursive type violates the positivity rule (you can think it as `datatype BadType = Loop of BadType -> BadType`, which is logically equivalent to `False`), we should reject it. In the Laurel land, it is exactly `composite BadType { var inner: BadType }`. What would fix it is turning the inner `BadType` to `Option BadType`.

keyboardDrummer:
> we should reject it. In the Laurel land, it is exactly
> `composite BadType { var inner: BadType }`

In a way the Laurel composite values are integers pointing to a place on the heap, so I don't think it's the same case as the `datatype BadType = Loop of BadType -> BadType`

I think we can come up with a mechanism to allow constructing valid values of `composite BadType`:

```
var newBadType: BadType = new BadType;

var inner: BadType := newBadType.inner;
// error: newBadType has not been constructed yet, so it may not be dereferenced

newBadType.inner := newBadType; // cycle is created
construct newBadType; // checks if all fields have been assigned
assert newBadType.inner == newBadType; // pass, dereference is allowed
```

AD1024:
> > we should reject it. In the Laurel land, it is exactly
> > `composite BadType { var inner: BadType }`
> 
> In a way the Laurel composite values are integers pointing to a place on the heap, so I don't think it's the same case as the `datatype BadType = Loop of BadType -> BadType`
> 
> I think we can come up with a mechanism to allow constructing valid values of `composite BadType`:
> 
> ```
> var newBadType: BadType = new BadType;
> 
> var inner: BadType := newBadType.inner;
> // error: newBadType has not been constructed yet, so it may not be dereferenced
> 
> newBadType.inner := newBadType; // cycle is created
> construct newBadType; // checks if all fields have been assigned
> assert newBadType.inner == newBadType; // pass, dereference is allowed
> ```

That sounds good! Let me think more carefully and get back to you. I just resolved the conflicts.

ssomayyajula (dismissed):

AD1024:
> > > we should reject it. In the Laurel land, it is exactly
> > > `composite BadType { var inner: BadType }`
> > 
> > 
> > In a way the Laurel composite values are integers pointing to a place on the heap, so I don't think it's the same case as the `datatype BadType = Loop of BadType -> BadType`
> > I think we can come up with a mechanism to allow constructing valid values of `composite BadType`:
> > ```
> > var newBadType: BadType = new BadType;
> > 
> > var inner: BadType := newBadType.inner;
> > // error: newBadType has not been constructed yet, so it may not be dereferenced
> > 
> > newBadType.inner := newBadType; // cycle is created
> > construct newBadType; // checks if all fields have been assigned
> > assert newBadType.inner == newBadType; // pass, dereference is allowed
> > ```
> 
> That sounds good! Let me think more carefully and get back to you. I just resolved the conflicts.

@leo-leesco Looping Leo in, since this issue may also cut in for type checking.

Strata/Languages/Laurel/HeapParameterization.lean
keyboardDrummer: To be clear: I'd prefer to inline this

AD1024: That's fine for me. I think it would be similar to the following helpers (should they also be inlined?):
https://github.com/strata-org/Strata/blob/6a074b1eac4bd496ecb7948129ab84d73bb3639d/Strata/Languages/Laurel/HeapParameterization.lean#L256-L258

AD1024:
> > we should reject it. In the Laurel land, it is exactly
> > `composite BadType { var inner: BadType }`
> 
> In a way the Laurel composite values are integers pointing to a place on the heap, so I don't think it's the same case as the `datatype BadType = Loop of BadType -> BadType`
> 
> I think we can come up with a mechanism to allow constructing valid values of `composite BadType`:
> 
> ```
> var newBadType: BadType = new BadType;
> 
> var inner: BadType := newBadType.inner;
> // error: newBadType has not been constructed yet, so it may not be dereferenced
> 
> newBadType.inner := newBadType; // cycle is created
> construct newBadType; // checks if all fields have been assigned
> assert newBadType.inner == newBadType; // pass, dereference is allowed
> ```

I think this contradicts with the allocatedness axiom. The axiom says the inner is allocated on the heap after line 1, but here we are raising an error. Or, is it the case that the memory is allocated but cannot be dereferenced? Did I get anything wrong here?

Strata/Languages/Laurel/HeapParameterization.lean
keyboardDrummer: Those helpers are awful because they use a none `source`. I plan to change `source` so it's not nullable and then those helpers will be deleted.

keyboardDrummer:
> > > we should reject it. In the Laurel land, it is exactly
> > > `composite BadType { var inner: BadType }`
> > 
> > 
> > In a way the Laurel composite values are integers pointing to a place on the heap, so I don't think it's the same case as the `datatype BadType = Loop of BadType -> BadType`
> > I think we can come up with a mechanism to allow constructing valid values of `composite BadType`:
> > ```
> > var newBadType: BadType = new BadType;
> > 
> > var inner: BadType := newBadType.inner;
> > // error: newBadType has not been constructed yet, so it may not be dereferenced
> > 
> > newBadType.inner := newBadType; // cycle is created
> > construct newBadType; // checks if all fields have been assigned
> > assert newBadType.inner == newBadType; // pass, dereference is allowed
> > ```
> 
> I think this contradicts with the allocatedness axiom. The axiom says the inner is allocated on the heap after line 1, but here we are raising an error. Or, is it the case that the memory is allocated but cannot be dereferenced? Did I get anything wrong here?

I think from that perspective, the allocation only happens at the `construct` call, where the cyclic object is allocated atomically.

Strata/Languages/Laurel/HeapParameterization.lean
AD1024: great! I'm pushing my changes that inline the `mkMdAt`.

MikaelMayer (commented):
🤖🔍 Style comment (raw constructors → `mkMdAt`) was addressed in 24f6131. The stale doc-comment in `T11_HeapWfConditions.lean` (L17-19 still references `heapIsValid($heap)` / `forall` over composites, which is the old design) was **not** updated. The current mechanism synthesizes per-parameter `Composite..ref!(p) < Heap..nextReference!(heap)` clauses, so the doc should reflect that. Minor nit — not blocking.

Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -498,11 +494,11 @@ private def heapWellFormednessPreconds (model : SemanticModel)
   inputs.filterMap fun p =>
     if isCompositeParam model p then
       let src := p.name.source
-      let pRead   := mkMdAt src (.Var (.Local p.name))
-      let pRef    := mkMdAt src (.StaticCall "Composite..ref!" [pRead])
-      let heapRead := mkMdAt src (.Var (.Local heapVar))
-      let counter := mkMdAt src (.StaticCall "Heap..nextReference!" [heapRead])
-      let allocated := mkMdAt src (.PrimitiveOp .Lt [pRef, counter])
+      let pRead     : StmtExprMd := { val := .Var (.Local p.name), source := src }
```
keyboardDrummer: Can I ask why'd you added these type annotations? They're not required.

AD1024: Oh, I asked claude did the inlining and it annotated them. Seems unnecessary, I'll remove these.

MikaelMayer (commented):
🤖🔍 All reviewer comments addressed: `mkMdAt` helper inlined, type annotations removed. My previous doc-comment nit (T11 still references `heapIsValid($heap)` / forall-over-composites phrasing) remains but is non-blocking. Reviewer sign-off still needed.

MikaelMayer (commented):
🤖🔍 All previous comments addressed: `mkMdAt` used consistently (no raw constructors), whitespace fixed, conditions marked `free`. The T11 doc-comment (L17-19 still says `heapIsValid($heap)` / forall-over-composites) remains stale but is non-blocking as noted before. Reviewer sign-off still needed.

keyboardDrummer (dismissed):

MikaelMayer (commented):

MikaelMayer (commented):
🤖🔍 All previous comments addressed: `mkMdAt` helper inlined with consistent `{ val := ..., source }` style across all three helpers, T11 docstring updated to reflect per-parameter allocation clauses, conditions marked `free := true`, type annotations removed. Reviewer sign-off still needed.

StrataPython/StrataPythonTest/expected_laurel/test_class_empty.expected
```diff
@@ -1,4 +1,5 @@
+test_class_empty.py(5, 4): ✅ pass - input is allocated on the heap
```
keyboardDrummer, line 1: That this is here means it actually wasn't free: the precondition was asserted at the call-site. I think Laurel currently is not respecting the `free` property correctly. I'm fixing that in https://github.com/strata-org/Strata/pull/1406, but I'm not exactly sure when that will land, since it depends on another PR being merged first.

Two options that looks good to me:
- Wait for 1406 to merge
- Include the modifications from 1406, except those to the modifies clause pass, in this PR.

AD1024:
I checked out the failed test cases. The failures are due to the synthesized preconditions and the `LaureToCoreSchema` pass (shown below): https://github.com/strata-org/Strata/blob/0a4a29b36b64a0e6f97b0402977a52a487b8ea75/Strata/Languages/Laurel/LaurelToCoreSchemaPass.lean#L433-L435

The label clashes because the pre-condition involving the same input heap reference share the same source location (same as the location at the heap reference, so that the diagnostic can point back to that variable). To resolve this, we can either (1) deduplicate the tag (I have a fix for this, haven't pushed); (2) deduplicate the source location (will lose the  location connection to the heap reference). 

edit: there are other subtle cases where assume/asserts are generated that can share this issue. The pass can generate statements sharing the same call-site location. I think going for (1) is a better choice.

@keyboardDrummer @ssomayyajula which one do you think makes more sense?

keyboardDrummer:
> I checked out the failed test cases. The failures are due to the synthesized preconditions and the `LaureToCoreSchema` pass (shown below):
> 
> https://github.com/strata-org/Strata/blob/0a4a29b36b64a0e6f97b0402977a52a487b8ea75/Strata/Languages/Laurel/LaurelToCoreSchemaPass.lean#L433-L435
> 
> The label clashes because the pre-condition involving the same input heap reference share the same source location (same as the location at the heap reference, so that the diagnostic can point back to that variable). To resolve this, we can either (1) deduplicate the tag (I have a fix for this, haven't pushed); (2) deduplicate the source location (will lose the location connection to the heap reference).
> 
> edit: there are other subtle cases where assume/asserts are generated that can share this issue. The pass can generate statements sharing the same call-site location. I think going for (1) is a better choice.
> 
> @keyboardDrummer @ssomayyajula which one do you think makes more sense?

Do we understand where the duplication comes from? Looking at `ContractPass.lean`, I don't immediately see where source locations are duplicated.

I guess it's the assertions/assumptions created on the call sites that are duplicated. It uses the same source location, the location of the call-site, for each of the assertions/assumptions generated there.

I think for now, let's try to work around the problem "(1) deduplicate the tag (I have a fix for this, haven't pushed)" that sounds good.

The real solution will be recording both the source location of the call and of the callee contract, in the generated assertion/assumptions, but right now there's no system for that.

AD1024:
Sounds good - yes, the duplication is because of the call-site location sharing (The duplication of source location is on my synthesized well-formedness conditions). I'll add a regression test and push together with the fix.

MikaelMayer (commented):

MikaelMayer (commented):
🤖🔍 All previous comments addressed (T11 docstring updated, helper inlined, annotations removed). New commits add label disambiguation per procedure — the `freshStmtLabel` simplification using set size as suffix is clean and correct. CI still pending. Reviewer sign-off still needed.

StrataPython/StrataPythonTest/expected_laurel/test_class_empty.expected
AD1024: I think the `free` made the pre-conditions to be an `assume`. I can wait for PR1406. Meanwhile, I can merge these changes into my local working branch

AD1024:
I have locally merged these changes - will open CR with other features together later!

PR #1400, opened 2026-06-22
Constrained composite fields refac

keyboardDrummer:
Constrained composite fields refactoring


fabiomadge (approved):
Behavior-preserving refactor, verified end-to-end (544 green): the unified `resultUsed`-aware traversal reproduces the old `elimStmt`/`wrapExprAssigns` split — statement and expression-position assigns both correct, field double-eval fix intact (RHS evaluated once), field constraint type detected via `computeExprType`. Nice simplification; resolves the `mapStmtExpr`/catch-all point. (Supplementary to @olivier-aws's review.)

olivier-aws:
LGTM, thanks for the refactor.

PR #1401, opened 2026-06-22
fix(laurel) : incorrect reporting of mismatched annotation location

leo-leesco:
With the new testing framework, we were correctly locating file-wide any uncaught diagnostic, but if an annotation was not matched (when we were over-expecting errors, or the error has changed), the annotation was located snippet-relative instead of file-relative.


keyboardDrummer (dismissed):
Thank you!

keyboardDrummer (approved):

kadirayk (approved):
Nice Fix!

Issue #1402, opened 2026-06-23
Support field-granular modifies clauses in Laurel (modifies o.f)


julesmt:
Laurel's `modifies` frame is object-granular: you can name a whole `Composite`, but not "only `o.balance` changes, `o.count` preserved."

## Current behavior
- `classifyModifiesType` drops field targets as non-composite.
- The frame (`buildModifiesEnsures`) excludes whole objects: `$obj ≠ o`.
- Native grammar doesn't parse `modifies o.f`.

julesmt:
https://github.com/strata-org/Strata/pull/1404

PR #1404, opened 2026-06-23
Laurel: field-granular modifies frame

julesmt:
Makes `modifies o#f` frame only the (o, f) pair instead of the whole owner object: that field may change while all other fields and every other object stay fixed.

https://github.com/strata-org/Strata/issues/1402

### DIFF
--- a/Strata/Languages/Laurel/ModifiesClauses.lean
+++ b/Strata/Languages/Laurel/ModifiesClauses.lean
@@ -56,25 +56,6 @@ inductive ModifiesEntry where
   | field (objExpr : StmtExprMd) (fieldConst : StmtExprMd)
 
 /--
-Recognize a heap-parameterized field read `readField($heap, obj, fieldConst)`
-(optionally under one Box destructor), as emitted for a field target like `o#f`.
-`fieldConst` is the per-declaring-type `Field` constructor (`A.x` ≠ `B.x`), the
-same one `readField` uses — so the soundness gate holds by construction.
--/
-def matchModifiesFieldRead (e : StmtExpr) : Option (StmtExprMd × StmtExprMd) :=
-  let tryReadField : StmtExpr → Option (StmtExprMd × StmtExprMd) := fun
-    | .StaticCall callee [_heap, objExpr, fieldConst] =>
-        if callee.text == "readField" then some (objExpr, fieldConst) else none
-    | _ => none
-  match tryReadField e with
-  | some r => some r
-  | none =>
-    -- Unwrap one Box destructor; gate on `Box` so unrelated unary calls aren't matched.
-    match e with
-    | .StaticCall callee [arg] => if callee.text.startsWith "Box" then tryReadField arg.val else none
-    | _ => none
-
-/--
 Classify a heap-relevant type into a `ModifiesEntry`, or `none` for
 non-heap-relevant types. Delegates to `classifyModifiesHighType` for the
 type classification.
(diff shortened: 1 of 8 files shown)


Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -184,6 +220,12 @@ def filterBodyNonCompositeModifies (model : SemanticModel) (body : Body)
     let (kept, diags) := mods.foldl (fun (acc, ds) e =>
       match e.val with
       | .All => (acc ++ [e], ds)  -- wildcard is always kept
+      | .Var (.Field target _) =>
+        -- Field target `o#f`: keep when the owner is a heap object (the field's own type is irrelevant).
+        let ownerTy := (computeExprType model target).val
+        if isHeapRelevantType ownerTy then (acc ++ [e], ds)
+        else
+          (acc, ds ++ [diagnosticFromSource e.source s!"modifies clause field target has non-composite owner type '{formatHighTypeVal ownerTy}' and will be ignored"])
```
keyboardDrummer: Let's dedup this line: `(acc, ds ++ [diagnosticFromSource e.source s!"modifies clause field target has non-composite owner type '{formatHighTypeVal ownerTy}' and will be ignored"])`

keyboardDrummer (commented):
We need to add more documentation to the `modifies` field in LaurelAST, explicitly stating the allowed modes:
- single object
- object set
- field of single object

I think we could also support a "field of object set", but given that Laurel doesn't have sets yet it seems hard to test that.

Strata/Languages/Laurel/ModifiesClauses.lean
keyboardDrummer: A bit unrelated to this PR, but in case you have the energy for it: this method and the accompanying pass should be removed and replaced by a check in Resolution.lean.

Strata/Languages/Laurel/ModifiesClauses.lean
```diff
@@ -42,12 +46,33 @@ public section
...
+  -- field-granular: only `fieldConst` of `objExpr` may change
+  | field (objExpr : StmtExprMd) (fieldConst : StmtExprMd)
+
+/--
+Recognize a heap-parameterized field read `readField($heap, obj, fieldConst)`
+(optionally under one Box destructor), as emitted for a field target like `o#f`.
+`fieldConst` is the per-declaring-type `Field` constructor (`A.x` ≠ `B.x`), the
+same one `readField` uses — so the soundness gate holds by construction.
+-/
+def matchModifiesFieldRead (e : StmtExpr) : Option (StmtExprMd × StmtExprMd) :=
```
keyboardDrummer: I think you should prevent HeapParameterization from translating outer field reads that occur in modifies clauses, to simplify this code. After that you can probably also inline `matchModifiesFieldRead`

We could even change `modifies : List StmtExprMd` in LaurelAST.lean to a more structured `modifies : List ModifiesTarget`, with:
```
inductive ModifiesTarget where
  | All -- remove StmtExpr.All since it's weird anyways
  | Object (objects : StmtExprMd) -- either a single ref or a set
  | Field (field : String) (objects : StmtExprMd)
```

fabiomadge (commented):
Reviewed for soundness; built at `3e52bce1d` (suite green). The field-granular frame is sound: it's a frame *tightening* (`.field` adds `!($obj==o && $fld==f)` to `notModified`, preserving strictly more), the gate is safe since `fieldConst` is the per-type `DeclaringType.fieldName` constant (`a#x` can't free `b#x`), and aliasing holds — `T9_ChainedFieldAccess`'s `chainedMayAliasFails` confirms a write through a possible alias isn't framed away. Frame violations caught, valid writes pass.

Non-blocking, reinforcing @keyboardDrummer's [suggestion to stop HeapParameterization translating field reads in modifies clauses](https://github.com/strata-org/Strata/pull/1404#discussion_r3466144932): `matchModifiesFieldRead` matches by callee text (`"readField"`, `startsWith "Box"`), coupling to HeapParameterization's names and single-Box wrapping depth. If that pass changes, an `o#f` target silently falls through with no diagnostic. That suggestion (or the structured `ModifiesTarget` it proposes) removes this; short of that, share those names as constants.

julesmt:
@keyboardDrummer @fabiomadge updated with your suggestions; one commit per suggestion

Strata/Languages/Laurel/ModifiesClauses.lean
julesmt: did part 1, for part 2 : sure maybe for another PR though, should I open an issue ?

keyboardDrummer:
Beautiful! Thanks so much

keyboardDrummer (dismissed):

Strata/Languages/Laurel/ModifiesClauses.lean
keyboardDrummer: Thanks! Up to you whether to follow-up.

fabiomadge (dismissed):

fabiomadge (dismissed):

fabiomadge (approved):

keyboardDrummer (approved):

PR #1408, opened 2026-06-24
Remove functions

keyboardDrummer:
### Changes

1. Remove Laurel functions.
1. Support let expressions in transparent bodies. This is necessary to implement the above change, because the contract pass can produce let expressions when calling procedures with preconditions, and some of the updated tests have transparent bodies that call procedures with preconditions.
1. Fix several bugs in the lifting pass. Again this is now necessary because with more things being procedures, there were more cases where lifting had to be done correctly.
1. Improve printing of if-then-else
1. Fix contract pass so it eliminates all pre and postconditions
1. Verification performance is worse because calling transparent procedures has a more complicated encoding than calling functions had. Calling transparent procedures should get an encoding optimization but that's out of scope for this PR.
1. Added `#guard_msgs (drop info) in` to several tests to prevent getting `info` output when running `lake test`. These guards previously already existed but they were accidentally removed in a test refactoring.

### Tests
1. Replaced usages of `function` in tests with `procedure`. Tests that had duplicated cases for functions and procedures had the function test-cases removed.
3. The T3_ControlFlow test was updated to use let expressions in a transparent body
4. The T2_ImpureExpressions test was updated to check for the new fixed bugs in the lifting pass
5. Some tests checked against printed if-then-else and they've been updated
6. No explicit test for this.
7. Updated tests. Some tests now check for "could not prove" instead of "does not hold"


keyboardDrummer:
@keyboardDrummer-bot can you investigate the build failures? If you have trouble resolving them, just post what you found

keyboardDrummer:
> I investigated the CI failure ("Build and test Lean (stable)"). Here's what I found:
> 
> **Local reproduction results (with cvc5 1.3.4, matching CI):**
> 
> * `lake build` (default targets): ✅ passes
> * `lake test` (StrataTest build + test runner): ✅ passes (only failure is the missing `ion-java-1.11.11.jar` which CI downloads separately)
> * StrataPython build + tests: ✅ passes
> * StrataCLI build: ✅ passes
> * Editor syntax verification: ✅ passes
> * `strata verify` smoke test: ✅ passes
> 
> All of these pass from a **completely clean build** (deleted `.lake/build` first).
> 
> **Most likely cause of CI failure:**
> 
> The CI uses a "Restore lake cache" step that restores cached `.lake` directories from previous builds on the base branch (`reviewed-kbd-will-merge-to-main`). This PR removes the `isFunctional` field from the `Procedure` structure, which changes the serialized representation of `Procedure` in `.olean` files. If the cache restores old `.olean` files that contain the old `Procedure` layout, they would be incompatible with the new code, potentially causing deserialization failures or type mismatches during incremental builds.
> 
> **Recommendation:** Re-run the CI workflow. If the cache is the culprit, the CI should either:
> 
> 1. Pass after the cache is invalidated by the structural change, or
> 2. A manual cache clear might be needed if Lake's cache key doesn't account for this kind of structural change
> 
> **Note:** I initially saw a `BvIntCastVerifyTests` failure locally, but that was because our local cvc5 was 1.2.1 which doesn't support `ubv_to_int`. After upgrading to cvc5 1.3.4 (matching CI), that test passes fine.

If you look at the CI results (https://github.com/strata-org/Strata/actions/runs/28098917360/job/83195378967?pr=1408), you can see that running `./run_py_interpret.sh` fails. I have reproduced that locally. Please investigate these failures

keyboardDrummer:
@keyboardDrummer-bot can you try to reproduce the SARIF failure that CI has now when running `run_py_analyze_sarif.py` ? https://github.com/strata-org/Strata/actions/runs/28114111319/job/83248951269?pr=1408

keyboardDrummer:
@keyboardDrummer-bot currently the analysis if a Python test is failing. You can reproduce with `ubuntu@ip-172-31-13-142:~/repos/Strata2/StrataPython/StrataPythonTest$ ./run_py_analyze.sh --filter test_bubble_sort_step`. Running this command fails because it hangs. It seems like it's hanging in Core's symbolic execution step. Can you determine how the generated Core is different for this test on this PR, compared to the target branch of this PR, and think about why this causes Core to not terminate?

keyboardDrummer:
@keyboardDrummer-bot can you investigate the failures in ./run_py_interpret.sh ?

keyboardDrummer:
@keyboardDrummer-bot please investigate the CMBC failures that occur when doing `./StrataPython/StrataPythonTest/run_py_cbmc_tests.sh`

keyboardDrummer:
@keyboardDrummer-bot please cherry-pick https://github.com/strata-org/Strata/pull/1430/changes/8a3bbe4368dc7551208023da7da30fdd5e49afde into this PR

keyboardDrummer:
@keyboardDrummer-bot please debug the failures in this CI run: https://github.com/strata-org/Strata/actions/runs/28402459992/job/84162551771?pr=1408

PR #1417, opened 2026-06-26
Generalize OverapproximatesUptoWhen to distinct input/output relations

PROgram52bc:
## What

Split the single state relation `R` of `OverapproximatesUptoWhen` into two: an **input** relation `Rpre` (the hypothesis relating initial envs) and an **output** relation `Rpost` (the existential relating final envs), as a new `OverapproximatesUptoWhen₂ Rpre Rpost`.

`OverapproximatesUptoWhen R` is then the **diagonal** `Rpre = Rpost = R`, defined as `OverapproximatesUptoWhen₂ R R`.

## Why

`R` sits in two structurally opposite positions:

```
Rpre :  R ρ₀ ρ₀'              -- hypothesis on the INPUT (antitone)
Rpost:  ∃ ρ'', R ρ' ρ'' ∧ …   -- existential on the OUTPUT (monotone)
```

The existing `OverapproximatesUptoWhen.mono` already documents this — it needs `R' ⊆ R` for the input *and* `R ⊆ R'` for the output. Forcing one relation into both positions cannot model a transform whose input discipline differs from its output discipline: e.g. one that runs from a **fixed** initial store yet **extends** the store on output (call elimination is exactly this — input pinned to the source store by its freshness invariant, output extended with fresh temp/old vars). No single `R` is simultaneously "forced-equal on input" and "store-extension on output".

This also fits the algebra already present: `OverapproximatesUptoWhen.comp` composes `R₁ ∘ R₂`, which is naturally an `Rpre ⇒ Rpost` morphism threaded through a middle relation.

## Compatibility

The redefinition is **definitionally equal** to the previous body, so every existing consumer is untouched: `OverapproximatesUpto`, `OverapproximatesWhen`, `Overapproximates`, `OverapproximatesUptoWhen.mono` / `.strengthen` / `.comp` / `.comp_preorder`, and the `overapproximates*_iff_upto_eq` `Iff.rfl` bridges.

## Verification

`lake build` green for `Strata.Transform.SpecificationProps`, `Strata.Transform.DetToKleeneCorrect`, and `Strata.Transform.CoreSpecification`. One file changed (+33/−20).

### DIFF
--- a/Strata/Transform/Specification.lean
+++ b/Strata/Transform/Specification.lean
@@ -291,30 +291,33 @@ There are more generic predicates such as `OverapproximatesWhen`,
   ∃ cfg : Config P CmdT, cfg.getEnv.hasFailure = true ∧
     StepStmtStar P evalCmd extendEval (.stmts ss ρ₀) cfg
 
-/-! ## Overapproximation up to a relation on states
+/-! ## Overapproximation up to relations on states
 
-`OverapproximatesUptoWhen R` relates the source and target executions up to a
-single state relation `R`: initial environments are related by `R` and final
-environments by `R`.  This is useful for transformations that
-rename/restructure the state.
+`OverapproximatesUptoWhen₂ Rpre Rpost` relates the source and target executions
+up to two state relations: initial environments are related by the *input*
+relation `Rpre` (a hypothesis) and final environments by the *output* relation
+`Rpost` (an existential).  This separation matters because a relation used as an
+input hypothesis is antitone while one used as an output witness is monotone, so
+a single relation cannot model a transform whose input discipline differs from
+its output discipline — e.g. one that runs from a *fixed* initial store yet
+*extends* the store on output.
 
-`OverapproximatesWhen` (the same-environment version below) is the special case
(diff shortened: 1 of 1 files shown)


aqjune-aws:
Hi, I would like to understand the background of this generalization further - the original `R` is an equivalence relation between two states (one from the input program and one from the output program). If we have to use two different equivalence relations for input states and output states, the situation is somewhat not intuitive; every program transformation must define a unique state equivalence relation it would like to preserve.

What is the motivation  behind generalizing it into taking two different relations?

PROgram52bc:
> Hi, I would like to understand the background of this generalization further - the original `R` is an equivalence relation between two states (one from the input program and one from the output program). If we have to use two different equivalence relations for input states and output states, the situation is somewhat not intuitive; every program transformation must define a unique state equivalence relation it would like to preserve.
> 
> What is the motivation behind generalizing it into taking two different relations?

Thanks for the question. After taking a deeper look, I think the current `OverapproximatesUptoWhen₂ R R` is too general and not needed for any of the transformations we have right now. Closing this PR.

PR #1419, opened 2026-06-26
Move laurel tests

keyboardDrummer:
### Changes
- Reorganized the Laurel tests folder and add documentation about its structure
- Add eager boolean operator test
- Add non-deterministic hole test
- Add for loop test
- Combine several error and non-error tests that relate to the same feature
- Merged control flow tests into transparentBody tests

### Next steps
- Add a Laurel interpreter so the tests in the EndToEndTests/Execution folder can be ran both using the interpreter and using verification

### Testing
- All existing tests are still there but some have been moved between files.


keyboardDrummer:
Merged internally. Will push to GH from there.

PR #1421, opened 2026-06-26
Laurel: field access on Any-typed values

julesmt:
Adds the read-side bridge that lets Laurel resolve and read a field on a value
whose static type is `Any` (a class instance carried by reference), by unboxing
it to its heap `Composite`. Gated on the program defining an `Any` datatype, so
existing Laurel programs are unaffected.

Example:
Python values often reach Laurel typed `Any` (unannotated parameters, spec
`self`, heterogeneous containers). Reading an attribute then has to recover the
underlying object:

```python
class Account:
    balance: int

def is_rich(a):                # `a` is unannotated → `Any` in Laurel
    return a.balance > 1000    # attribute read on an `Any`-typed value
```

### DIFF
--- a/Strata/Languages/Laurel/HeapParameterization.lean
+++ b/Strata/Languages/Laurel/HeapParameterization.lean
@@ -291,7 +291,18 @@ where
 
         let valTy := (model.get fieldName).getType
         let selectTarget' ← recurseOne selectTarget
-        let readExpr := ⟨ .StaticCall "readField" [mkMd (.Var (.Local heapVar)), selectTarget', mkMd (.StaticCall qualifiedName [])], source ⟩
+        -- Unbox an `Any`-typed target to a `Composite` using the field's
+        -- owning-class tag; other targets pass through unchanged.
+        let targetTy := (computeExprType model selectTarget).val
+        let selectComposite : StmtExprMd :=
+          match targetTy with
+          | .UserDefined n =>
+            if n.text == "Any" then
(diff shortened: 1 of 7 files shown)


Strata/Languages/Laurel/HeapParameterization.lean
```diff
@@ -291,7 +291,18 @@ where
...
-        let readExpr := ⟨ .StaticCall "readField" [mkMd (.Var (.Local heapVar)), selectTarget', mkMd (.StaticCall qualifiedName [])], source ⟩
+        -- Unbox an `Any`-typed target to a `Composite` using the field's
+        -- owning-class tag; other targets pass through unchanged.
+        let targetTy := (computeExprType model selectTarget).val
+        let selectComposite : StmtExprMd :=
+          match targetTy with
+          | .UserDefined n =>
+            if n.text == "Any" then
+              let owner := (qualifiedName.splitOn ".").head!
+              mkMd (.StaticCall "anyToComposite" [selectTarget', mkMd (.StaticCall (owner ++ "_TypeTag") [])])
```
keyboardDrummer, line 302: `any` is not a Laurel concept. It should really be renamed to `PythonDynamicValue`. We shouldn't refer to it from any Laurel code.

The correct way to get field reads/writes on Any typed objects is to introduce a wrapper method around field read/write that unpacks the Any to get a composite out of it.

keyboardDrummer (changes requested):

PR #1422, opened 2026-06-26
Add gradual-typing elaboration to Laurel resolution

julesmt:
Adds an optional `GradualConfig` so resolution keeps native operators on native operands and routes `Any` operands through prelude operators, inserting box/unbox/`toBool` casts. Inert without a config; strict typing
  unchanged. Foundation for the stacked native typed-Python PR.

PR #1424, opened 2026-06-27
Add gradual-typing elaboration to Laurel resolution

julesmt:
Adds an optional `GradualConfig` so resolution keeps native operators on native operands and routes `Any` operands through prelude operators, inserting box/unbox/`toBool` casts. Inert without a config; strict typing unchanged. Foundation for the stacked native typed-Python PR.

---
_Continuation of #1422. That PR was auto-closed by GitHub when its head branch was renamed to `julesmt/feature/native-types/laurel` (GitHub closes any open PR whose head branch is renamed). Same commits, same base._

### DIFF
--- a/Strata/Languages/Laurel/Resolution.lean
+++ b/Strata/Languages/Laurel/Resolution.lean
@@ -560,6 +560,101 @@ private def checkIncrDecrTargetType (op : IncrDecrOp) (target : VariableMd)
            Use an explicit assignment instead, e.g. 'x := x + 1'."
       modify fun s => { s with errors := s.errors.push diag }
 
+/-- Test whether `op` is in the arithmetic family. -/
+private def isArithmeticOp : Operation → Bool
+  | .Neg | .Add | .Sub | .Mul | .Div | .Mod | .DivT | .ModT => true
+  | _ => false
+
+/-- LUB of the operand types, or `none` if inconsistent. -/
+private def joinAll (ctx : TypeLattice) (argTypes : List HighTypeMd)
+    (source : Option FileRange) : Option HighTypeMd :=
+  argTypes.foldl
+    (fun acc t => match acc with | some l => join ctx l t | none => none)
+    (some { val := .Unknown, source := source })
+
+/-- A `MultiValuedExpr` operand is a multi-output call (`multi(x)` declared
+    `returns (a, b)`) used in value position. It is an internal pseudo-type with
+    no Core lowering, so it must never reach an operator slot — letting it
+    through crashes a later pass as a `StrataBug`. Emit the position-oriented
+    diagnostic for the offending operand and return `true` so the caller
+    short-circuits to the operator's natural result type, suppressing the
+    per-family check (and its cascading error) on that operand. -/
+private def reportMultiValued (a : StmtExprMd) (aTy : HighTypeMd) : ResolveM Bool := do
+  match aTy.val with
+  | .MultiValuedExpr _ =>
+    let diag := diagnosticFromSource a.source
+      "multi-output call cannot be used as a value here; it returns multiple values. Unpack it into separate variables first"
+    modify fun s => { s with errors := s.errors.push diag }
+    pure true
+  | _ => pure false
+
+/-- Native arithmetic typing: every operand must be numeric, and the result is
+    the LUB of the operand types under the consistency relation. A multi-valued
+    operand short-circuits to `Unknown` (arithmetic's natural
+    cascade-suppression type). -/
+private def nativeArith (op : Operation) (expr : StmtExpr) (args' : List StmtExprMd)
+    (argTypes : List HighTypeMd) (skipProof : Bool) (source : Option FileRange) :
+    ResolveM (StmtExpr × HighTypeMd) := do
+  let unknownTy : HighTypeMd := { val := .Unknown, source := source }
+  let mut hasMulti := false
+  for (a, aTy) in args'.zip argTypes do
+    if (← reportMultiValued a aTy) then hasMulti := true
+  if hasMulti then
+    return (.PrimitiveOp op args' skipProof, unknownTy)
+  let ctx := (← get).typeLattice
+  for (a, aTy) in args'.zip argTypes do
+    unless isNumeric ctx aTy do
+      typeMismatch a.source (some expr) "expected a numeric type" aTy
+  match joinAll ctx argTypes source with
+  | some ty => pure (.PrimitiveOp op args' skipProof, ty)
+  | none =>
+    let formatted := ", ".intercalate (argTypes.map (fun t => s!"'{formatType t}'"))
+    let diag := diagnosticFromSource source s!"cannot apply '{op}' to operands of types {formatted}"
+    modify fun s => { s with errors := s.errors.push diag }
+    pure (.PrimitiveOp op args' skipProof, unknownTy)
+
+/-- Native typing for the boolean / comparison / equality / string-concat
+    families: a fixed result type (`TBool` / `TString`) with the family's operand
+    constraint (booleans subtype `TBool`, comparisons are numeric, equality is
+    consistent, concat operands subtype `TString`). A multi-valued operand
+    short-circuits to the family's natural result type. -/
+private def nativeOther (op : Operation) (expr : StmtExpr) (args' : List StmtExprMd)
+    (argTypes : List HighTypeMd) (skipProof : Bool) (source : Option FileRange) :
+    ResolveM (StmtExpr × HighTypeMd) := do
+  let resultTy := match op with
+    | .Eq | .Neq | .And | .Or | .AndThen | .OrElse | .Not | .Implies
+    | .Lt | .Leq | .Gt | .Geq => HighType.TBool
+    | .StrConcat => HighType.TString
+    | _ => HighType.Unknown
+  let mut hasMulti := false
+  for (a, aTy) in args'.zip argTypes do
+    if (← reportMultiValued a aTy) then hasMulti := true
+  if hasMulti then
+    return (.PrimitiveOp op args' skipProof, { val := resultTy, source := source })
+  match op with
+  | .And | .Or | .AndThen | .OrElse | .Not | .Implies =>
+    for (a, aTy) in args'.zip argTypes do
+      checkSubtype a.source { val := .TBool, source := a.source } aTy
+  | .Lt | .Leq | .Gt | .Geq =>
+    let ctx := (← get).typeLattice
+    for (a, aTy) in args'.zip argTypes do
+      unless isNumeric ctx aTy do
+        typeMismatch a.source (some expr) "expected a numeric type" aTy
+  | .Eq | .Neq =>
+    match argTypes with
+    | [lhsTy, rhsTy] =>
+      let ctx := (← get).typeLattice
+      unless isConsistent ctx lhsTy rhsTy do
+        let diag := diagnosticFromSource source
+          s!"cannot compare '{formatType lhsTy}' with '{formatType rhsTy}' using '{op}'"
+        modify fun s => { s with errors := s.errors.push diag }
+    | _ => pure ()
+  | .StrConcat =>
+    for (a, aTy) in args'.zip argTypes do
+      checkSubtype a.source { val := .TString, source := a.source } aTy
+  | _ => pure ()
+  pure (.PrimitiveOp op args' skipProof, { val := resultTy, source := source })
+
 /-! ## Typing rules
 
 The judgment is bidirectional:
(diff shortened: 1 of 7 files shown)


julesmt:
_↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#pullrequestreview-4581894267))

**Review (COMMENTED):**

Great PR ! This goes really helps moving forward to lower `Dyn` type into Laurel
I wrote a few design considerations that we can discuss and asked for some modifications

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -136,8 +136,35 @@ abbrev Scope := Std.HashMap String ScopeEntry
...
 
+/-- Frontend-supplied gradual-typing prelude. When present, resolution rewrites
+    operations on `dynamic` operands to the named prelude operators and
+    materializes the implicit box/unbox/`toBool` casts at boundaries. All names
+    must already be in scope; `none` ⇒ no elaboration. -/
+structure GradualConfig where
+  /-- The dynamic type whose values are boxed (`TCore "Any"` for Python). -/
+  dynamic : HighType
+  /-- Whether a type is the dynamic type (may have several spellings). -/
+  isDynamic : HighType → Bool
```
julesmt, line 147: _↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3483426111))

Design consideration : instead of threading this `Dyn` type in the resolution, can we imagine referencing it throughout the program?
In other words, do we expect `Dyn` to be source-language dependent ? This looks like a dependency inversion to me, if we were to add a new type to express the `Dyn` type, we could probably defer translating the source-language `Dyn` to Laurel's `Dyn` to the corresponding frontend

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -136,8 +136,34 @@ abbrev Scope := Std.HashMap String ScopeEntry
...
+      to frontend (user) code, leaving hand-written prelude bodies untouched. -/
+  shouldElaborate : String → Bool
+
 /-- State threaded through the resolution pass. -/
 structure ResolveState where
+  /-- Gradual-typing elaboration config, if the frontend requested it. -/
+  gradual : Option GradualConfig := none
+  /-- Whether gradual elaboration is active for the procedure being resolved.
+      Set per-procedure by `resolveProcedure` so only user code is elaborated. -/
+  gradualActive : Bool := false
```
julesmt: _↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3483436371))

is there a benefit to having two fields as opposed to destructing gradual to `Some _ | None` ?
In other words, do we gain anything from not defining `gradualActive = gradual.isSome()` ?

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -629,6 +657,290 @@ pattern-match on the constructor and delegate to the corresponding helper. -/
...
+    ResolveM (Option (StmtExpr × HighTypeMd)) := do
+  unless argTypes.any (fun t => g.isDynamic t.val) do return none
+  forcePrelude g op args' argTypes source
+
+/-- LUB of the operand types, or `none` if inconsistent. -/
+private def joinAll (ctx : TypeLattice) (argTypes : List HighTypeMd)
+    (source : Option FileRange) : Option HighTypeMd :=
+  argTypes.foldl
+    (fun acc t => match acc with | some l => join ctx l t | none => none)
+    (some { val := .Unknown, source := source })
```
julesmt, line 776: _↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3483486261))

just a warning : this is not yet an actual LUB (please update doc) as it does not walk any inheritance tree

`join` simply erases any `Unknown` if tried to merge with another static type. Can you look at `join` and see if it needs updating using your `GradualConfig` ?

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -629,6 +657,290 @@ pattern-match on the constructor and delegate to the corresponding helper. -/
...
+private def reportMultiValued (a : StmtExprMd) (aTy : HighTypeMd) : ResolveM Bool := do
+  match aTy.val with
+  | .MultiValuedExpr _ =>
+    let diag := diagnosticFromSource a.source
+      "multi-output call cannot be used as a value here; it returns multiple values. Unpack it into separate variables first"
+    modify fun s => { s with errors := s.errors.push diag }
+    pure true
+  | _ => pure false
+
+/-- A candidate lowering: `some` if it applies, else `none`. -/
```
julesmt: _↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3483492280))

can you expand this docstring (eg providing an example) or provide an example in response to this thread ?

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -629,6 +669,295 @@ pattern-match on the constructor and delegate to the corresponding helper. -/
...
+    Emit the position-oriented diagnostic and return `true` so the caller
+    short-circuits to the operator's result type, suppressing cascades. -/
+private def reportMultiValued (a : StmtExprMd) (aTy : HighTypeMd) : ResolveM Bool := do
+  match aTy.val with
+  | .MultiValuedExpr _ =>
+    let diag := diagnosticFromSource a.source
+      "multi-output call cannot be used as a value here; it returns multiple values. Unpack it into separate variables first"
+    modify fun s => { s with errors := s.errors.push diag }
+    pure true
+  | _ => pure false
```
julesmt, line 805: _↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3483561857))

I might be wrong, but I seem to remember the typechecker was already making the same kind of check ; if not we should move this helper function with the rest of the typechecker

StrataTest/Languages/Laurel/GradualResolutionTests.lean
```diff
@@ -0,0 +1,263 @@
...
+import StrataTest.Util.TestLaurel
+
+open Strata
+open StrataTest.Util
+
+namespace Strata.Laurel
+
+/-- Substring containment. -/
+private def containsSubstr (haystack needle : String) : Bool :=
+  (haystack.splitOn needle).length > 1
```
julesmt, line 29: _↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3483779465))

[out of scope of this PR] : considering we have a similar helper function used to match annotations in tests, maybe we could factor both of them out ?

Strata/Languages/Laurel/Resolution.lean
julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484290381))

sure, like a canonical `HighType.Dyn `constructor in `LaurelAST.lean`?

Strata/Languages/Laurel/Resolution.lean
julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484309821))

I don't think so, at least without some refactoring.
So `gradual` tells us if gradual typing enabled globally and `gradualActive` tells us if the current procedure be elaborated. The latter is `gradual.isSome && shouldElaborate(currentProc)`, so they are not equivalent. However, I agree it's not super elegant, let me see if we have something neeter.

Strata/Languages/Laurel/Resolution.lean
julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484310477))

good catch

Strata/Languages/Laurel/Resolution.lean
julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484310814))

sure

Strata/Languages/Laurel/Resolution.lean
julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484314596))

I'll ask kiro to check if you have something similar to extract a helper !

StrataTest/Languages/Laurel/GradualResolutionTests.lean
julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484317840))

true

Strata/Languages/Laurel/Resolution.lean
julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484391434))

maybe another PR no ?

Strata/Languages/Laurel/Resolution.lean
julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484503755))

at least I didn't find anything close, nor did Kiro, could you look ?

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -1850,106 +2189,16 @@ def Synth.primitiveOp (exprMd : StmtExprMd) (expr : StmtExpr)
...
   let _ := h_expr  -- carries the constructor identity for `expr` in diagnostics
-  -- Guard (all operator families): a `MultiValuedExpr` operand is a
-  -- multi-output call (`multi(x)` declared `returns (a, b)`) used in value
-  -- position. It is an internal pseudo-type with no Core lowering, so it must
-  -- never reach an operator slot — letting it through crashes a later pass as
-  -- a `StrataBug`. Emit the position-oriented diagnostic per offending operand
-  -- and return `true` so the caller short-circuits to the operator's natural
-  -- result type, suppressing the per-family check (and its cascading error)
-  -- on that operand.
-  let reportMultiValued (a : StmtExprMd) (aTy : HighTypeMd) : ResolveM Bool := do
```
julesmt, line 1861: _↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3483750005))

see comment https://github.com/strata-org/Strata/pull/1422/changes#r3483561857

julesmt: _↪ Imported from #1422 — originally by @leo-leesco on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3483760112))

maybe we can discuss the location of this ? If it makes sense to move it, that's perfectly okay

julesmt: _↪ Imported from #1422 — originally by @julesmt on 2026-06-26_ ([original](https://github.com/strata-org/Strata/pull/1422#discussion_r3484317354))

yeah let's discuss it monday

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -136,8 +136,35 @@ abbrev Scope := Std.HashMap String ScopeEntry
 /-- Per-composite-type scope mapping field names to their scope entries. -/
 abbrev TypeScopes := Std.HashMap String Scope
 
+/-- Frontend-supplied gradual-typing prelude. When present, resolution rewrites
+    operations on `dynamic` operands to the named prelude operators and
+    materializes the implicit box/unbox/`toBool` casts at boundaries. All names
+    must already be in scope; `none` ⇒ no elaboration. -/
+structure GradualConfig where
+  /-- The dynamic type whose values are boxed (`TCore "Any"` for Python). -/
+  dynamic : HighType
```
keyboardDrummer, line 145: I had a different design in mind to support performant Python that I would hoping would be more powerful, and so would enable supporting different use-cases as well, related to nullable reference types and exceptions.

The design I had in mind for performant Python is:
- Do not insert any `Any` types in PythonToLaurel, except for possibly as type annotations in library functions. Leave local variables without a type annotation.
- Configure Laurel with a set of implicit value conversions from specific types to the `Any` type.

Input Laurel from Python:
```
procedure foo(b) {
  var x := 3;
  var y := x + x;
  var z := if b then y else "hello";
  var a := z + z
};
```

After Laurel type inference:

```
procedure foo(b: bool) {
  var x: int := 3;
  var y: int := x +int x;
  var z: Any := Any..int(y) if b else Any..string("hello");
  var a := z +Any z;
};
```

The `+` operator in Laurel would be overloaded, meaning there are several definitions with different signatures, and type inference would decide which one to pick. 

In the first implementation I would say that the implicit conversions related to Any would be one directional, only going from specific types to Any but not the other way around: once you're in Any land you're not getting out. We could enable going both ways but this would change the verification output and we should be careful to keep the verification and runtime behavior in sync, especially if we view not verifying everything as Any as purely a verification performance optimization.

@leo-leesco @ssomayyajula does that match with your thoughts?

What I like about the above system is that I think it would also resolve issues other than Python's dynamic value. For example, Java has nullable reference variables, and they need to be wrapped in an Option in Laurel.

Input Laurel from Java:
```
procedure foo() {
  var o := new Object();
  var p := o;
  p := null;
  p.toString();
}
```

After Laurel type inference:
```
procedure foo() {
  var o: Object := new Object();
  var p: Object := Some(o); 
  o := None();
  Option..fromSome(p).toString(); // precondition adds an implicit assertion
}
```

Java programs would have the option to use annotations to specify that a particular dereference can throw, which would not generate an assertion but will possibly throw a `NullPointerException`, which will then propagate up the call-chain until caught.

Java program:
```java
void foo() {
  var p: Object := null;
  JVerify.allowNullPointerException();
  p.toString();
}
```

Before type inference:

```
void foo() {
  var p: Object := null;
  dereferenceOrThrow(p).toString();
}

procedure derefenceOrThrow<T>(nullable: Option<T>): T throws NullPointerException
  .. ensures clauses..
```

After type inference:

```
void foo() throws NullPointerException // throws has been added, ie: procedure type has been lifted
{
  var p: Option<Object> := null; // local variable type has been lifted!
  dereferenceOrThrow(p).toString();
}
```

Strata/Languages/Laurel/Resolution.lean
```diff
@@ -136,8 +136,35 @@ abbrev Scope := Std.HashMap String ScopeEntry
 /-- Per-composite-type scope mapping field names to their scope entries. -/
 abbrev TypeScopes := Std.HashMap String Scope
 
+/-- Frontend-supplied gradual-typing prelude. When present, resolution rewrites
+    operations on `dynamic` operands to the named prelude operators and
+    materializes the implicit box/unbox/`toBool` casts at boundaries. All names
+    must already be in scope; `none` ⇒ no elaboration. -/
+structure GradualConfig where
```
keyboardDrummer, line 143: Can we try a GradualConfig that is just `coercions: List Coercion` with
```
structure Coercion where
   from : HighType
   to : HighType
   cast: StmtExprMd -> StmtExprMd 
```

and then fill this with coercions from all the constructors of PythonAny to PythonAny?

We might need to remove some (or all?) of the PythonAny boxing that is happening in PythonToLaurel, so Resolution can ensure to only insert it where necessary.

PR #1428, opened 2026-06-27
fix(semantics): make WellFormedSemanticEval{Val,Var} jointly satisfiable

PROgram52bc:
## Problem

`EvalCommand.call_sem` (and `EvalCommandContract.call_sem`) require their evaluator `δ : CoreEval` to satisfy both:

- `WellFormedSemanticEvalVal δ` — every output of `δ` is a value, and
- `WellFormedSemanticEvalVar δ` — on a free-variable atom, `δ` returns the store binding (`δ σ (fvar x) = σ x`).

As stated, **no `δ` can satisfy both at once.** Both predicates quantify their store argument over *all* stores internally, so a store binding a variable to a non-value (e.g. `.app …`, which is not a `Core.Value`) forces the var rule to return a non-value, contradicting the value rule:

```lean
theorem no_wf_eval : ¬ ∃ δ : Core.CoreEval, WellFormedSemanticEvalVar δ ∧ WellFormedSemanticEvalVal δ
```

is provable. The `call_sem` constructors are therefore **vacuous** — the operational semantics can never take a procedure-call step, and no concrete evaluator can ever witness them.

## Fix

Introduce `WellFormedStore` (a store holds only values) and **relativize the inner store quantifier** of `WellFormedSemanticEvalVal`'s value conjunct and of `WellFormedSemanticEvalVar` to well-formed stores:

```lean
@[expose] def WellFormedStore [HasVal P] (σ : SemanticStore P) : Prop :=
    ∀ x v, σ x = some v → HasVal.value v
```

This is sound: reachable stores are always value-only (they are populated by evaluator outputs and copies thereof), so the restriction only excludes the pathological non-value stores that made the conjunction unsatisfiable — it does not weaken the semantics. A `WellFormedStore σ` premise is added to the two `call_sem` constructors and to `ProcEnvWF`, and threaded through the affected proofs (`EvalExpressionIsDefined`, `eval_projectStore_to_full`, `PrefixStepsOK_det_init_*`, plus a small `prefixInitEnv_store_wf` preservation lemma).

Note: the fix necessarily targets the predicates' **inner** quantifier — adding a premise about `call_sem`'s parameter store would *not* work, since `WellFormedSemanticEvalVal δ ∧ WellFormedSemanticEvalVar δ` is contradictory on its own, independent of any parameter.

`WellFormedSemanticEvalVal` is now a `structure` whose two clauses are named (`outputsAreValues`, `identityOnValues`) rather than an anonymous conjunction, so use sites refer to the clauses by name.

## A concrete evaluator now satisfies the conditions

Test file `StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean` defines a concrete `coreWitnessEval : CoreEval` (look up fvars in the store; identity on values; reduce other expressions to a canonical value) and proves, with no `sorry`:

```lean
theorem exists_wf_coreEval :
    ∃ δ : CoreEval, WellFormedSemanticEvalVar δ ∧ WellFormedSemanticEvalVal δ
```

so the conjunction `call_sem` requires is now demonstrably inhabitable.

## Verification

- All affected proofs compile (`CallElimCorrect`, `ProcBodyVerifyCorrect`, `StatementSemanticsProps`, `DetToKleeneCorrect`, the Kleene/Spec files, `StepStmtTest`, and the witness test file — full closure green).
- **No new `sorry` or axioms introduced.** The pre-existing `sorry` in `EvalCallBodyRefinesContract` is untouched. The witness theorems depend only on `propext` and `Quot.sound`.

## Scope / follow-ups

- The witness is a purpose-built `CoreEval`, not the production `LExpr.eval`. Wiring `LExpr.eval` in would additionally require an `LState ↔ SemanticStore` adapter and reconciling `isCanonicalValue` with `Core.Value` (a saturated `.app` is `isCanonicalValue` but not a `Core.Value`); that reconciliation is left as a follow-up.
- `EvalCallBodyRefinesContract`'s existing `sorry` is out of scope here, but the added `WellFormedStore` premise is now available to whoever discharges it.

### DIFF
--- a/Strata/DL/Imperative/CmdSemantics.lean
+++ b/Strata/DL/Imperative/CmdSemantics.lean
@@ -96,12 +96,14 @@ def substSwap {P : PureExpr} (substs : List (P.Ident × P.Ident))
 @[expose] def WellFormedStore {P : PureExpr} [HasVal P] (σ : SemanticStore P) : Prop :=
     ∀ x v, σ x = some v → HasVal.value v
 
-@[expose] def WellFormedSemanticEvalVal {P : PureExpr} [HasVal P]
-    (δ : SemanticEval P) : Prop :=
-  -- evaluator only evaluates to values (on well-formed stores)
-    (∀ v v' σ, WellFormedStore σ → δ σ v = some v' → HasVal.value v') ∧
-  -- evaluator is identity on values
-    (∀ v' σ, HasVal.value v' → δ σ v' = some v')
+/-- Well-formedness of a `SemanticEval`'s value behavior, split into named
+    clauses. -/
+structure WellFormedSemanticEvalVal {P : PureExpr} [HasVal P]
+    (δ : SemanticEval P) : Prop where
+  /-- The evaluator produces only values (on well-formed stores). -/
+  outputsAreValues : ∀ v v' σ, WellFormedStore σ → δ σ v = some v' → HasVal.value v'
+  /-- The evaluator is the identity on values. -/
+  identityOnValues : ∀ v' σ, HasVal.value v' → δ σ v' = some v'
 
 @[expose] def WellFormedSemanticEvalVar {P : PureExpr} [HasVal P] [HasFvar P] (δ : SemanticEval P)
     : Prop := (∀ e v σ, WellFormedStore σ → HasFvar.getFvar e = some v → δ σ e = σ v)
--- a/Strata/DL/Imperative/CmdSemantics.lean
+++ b/Strata/DL/Imperative/CmdSemantics.lean
@@ -89,15 +89,22 @@ def substSwap {P : PureExpr} (substs : List (P.Ident × P.Ident))
       (δ σ e = some Imperative.HasBool.tt ↔ δ σ (Imperative.HasBoolOps.not e) = (some HasBool.ff)) ∧
       (δ σ e = some Imperative.HasBool.ff ↔ δ σ (Imperative.HasBoolOps.not e) = (some HasBool.tt))
 
-def WellFormedSemanticEvalVal {P : PureExpr} [HasVal P]
+/-- A store is well-formed when every binding it holds is a value.  The
+    predicates below only constrain such stores, so the value rule and the var
+    rule stay jointly satisfiable (an arbitrary store could bind a variable to a
+    non-value). -/
+@[expose] def WellFormedStore {P : PureExpr} [HasVal P] (σ : SemanticStore P) : Prop :=
+    ∀ x v, σ x = some v → HasVal.value v
+
+@[expose] def WellFormedSemanticEvalVal {P : PureExpr} [HasVal P]
     (δ : SemanticEval P) : Prop :=
-  -- evaluator only evaluates to values
-    (∀ v v' σ, δ σ v = some v' → HasVal.value v') ∧
+  -- evaluator only evaluates to values (on well-formed stores)
+    (∀ v v' σ, WellFormedStore σ → δ σ v = some v' → HasVal.value v') ∧
   -- evaluator is identity on values
     (∀ v' σ, HasVal.value v' → δ σ v' = some v')
 
-@[expose] def WellFormedSemanticEvalVar {P : PureExpr} [HasFvar P] (δ : SemanticEval P)
-    : Prop := (∀ e v σ, HasFvar.getFvar e = some v → δ σ e = σ v)
+@[expose] def WellFormedSemanticEvalVar {P : PureExpr} [HasVal P] [HasFvar P] (δ : SemanticEval P)
+    : Prop := (∀ e v σ, WellFormedStore σ → HasFvar.getFvar e = some v → δ σ e = σ v)
 
 @[expose] def WellFormedSemanticEvalExprCongr {P : PureExpr} [HasFvars P] (δ : SemanticEval P)
     : Prop := ∀ e σ σ', (∀ x ∈ HasFvars.getFvars e, σ x = σ' x) → δ σ e = δ σ' e
--- a/StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
+++ b/StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
@@ -23,6 +23,10 @@ The evaluator follows the shape of a real expression evaluator:
   identity-on-values half of `WellFormedSemanticEvalVal`);
 * on any other (reducible) expression it normalizes to a canonical value.
 
+This witness preserves value-ness but not typing — e.g. a boolean expression
+normalizes to the integer `canonicalValue`. That suffices here: the
+well-formedness predicates constrain value-ness, not types.
+
 The two conditions are jointly satisfiable because they only constrain
 value-only stores (`WellFormedStore`): on such a store the free-variable lookup
 returns a value, so it agrees with the "every output is a value" requirement.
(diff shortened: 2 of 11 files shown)


Strata/DL/Imperative/CmdSemantics.lean
```diff
@@ -89,15 +89,22 @@ def substSwap {P : PureExpr} (substs : List (P.Ident × P.Ident))
...
 
-def WellFormedSemanticEvalVal {P : PureExpr} [HasVal P]
+/-- A store is well-formed when every binding it holds is a value.  The
+    predicates below only constrain such stores, so the value rule and the var
+    rule stay jointly satisfiable (an arbitrary store could bind a variable to a
+    non-value). -/
+@[expose] def WellFormedStore {P : PureExpr} [HasVal P] (σ : SemanticStore P) : Prop :=
+    ∀ x v, σ x = some v → HasVal.value v
+
+@[expose] def WellFormedSemanticEvalVal {P : PureExpr} [HasVal P]
```
aqjune-aws: Can we make `WellFormedSemanticEvalVal` a structure so that each clause has its name?

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
```diff
@@ -0,0 +1,116 @@
+/-
```
aqjune-aws, line 1: I know this directly exhibits what we wanted to fix, but on the other hand I don't think this will address the concern of "what about other pairs of wellformednesses?". I think the right way is to define Core-version instances of every wellformedness typeclasses, then let the proof not take those as variables, but infer from existing instances. I want to do that.
Can we remove this file in the CR?

Strata/DL/Imperative/CmdSemantics.lean
PROgram52bc: Yes. For sure.

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
PROgram52bc: > I don't think this will address the concern of "what about other pairs of wellformednesses?"

I know that this evaluator is not what we want to keep, but the idea is that, if we can instantiate all wellformednesses using a single evaluator, then by definition it satisfies "all pairs of wellformednesses". Are we on the same page here?

> Can we remove this file in the CR?

Absolutely. For now though, do we want some evidence that at least some evaluator _can_ satisfy the well-formedness notion, so we don't run into the same `¬ ∃ δ : Core.CoreEval, WellFormedSemanticEvalVar δ ∧ WellFormedSemanticEvalVal δ` problem as Timothy described? Since Core evaluator is not terminating right now, do we want to adjust the existing well-formedness definitions as well, in order to instantiate them with Core?

aqjune-aws: > I know that this evaluator is not what we want to keep, but the idea is that, if we can instantiate all wellformednesses using a single evaluator, then by definition it satisfies "all pairs of wellformednesses". Are we on the same page here?

No, I think we need to build instances for all WellFormed* type classes and make the proofs that consume these use them. With this, we will be able to remove lots of `variable [WellFormed*]` and `theorem ... [WellFormed*]`.
An instantiation on some simple instantiation won't be enough.

> Absolutely. For now though, do we want some evidence that at least some evaluator can satisfy the well-formedness notion, so we don't run into the same ¬ ∃ δ : Core.CoreEval, WellFormedSemanticEvalVar δ ∧ WellFormedSemanticEvalVal δ problem as Timothy described? 

Then let's put this at StrataTest/Languages/Core/Tests. There are a few test files that only consist of proofs (e.g., StrataTest/DL/Imperative/StepStmtTest.lean).

> Since Core evaluator is not terminating right now, do we want to adjust the existing well-formedness definitions as well, in order to instantiate them with Core?

I will take this work and try to write a CR by today.

PROgram52bc: Okay. Will do soon!

Strata/DL/Imperative/CmdSemantics.lean
PROgram52bc: Done in 63780d8da. `WellFormedSemanticEvalVal` is now a structure with named clauses `outputsAreValues` and `identityOnValues`; the projection sites in `CallElimCorrect` and `KleeneSemanticsProps` use the named fields.

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
PROgram52bc: Moved to `StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean` in 63780d8da, following the `StepStmtTest.lean` proof-only test convention (`meta import`/`meta section`), and dropped its import from `Strata.lean`. It's out of the main library now but still witnesses joint satisfiability so we don't regress into the vacuity problem. The broader "instances for all WellFormed* typeclasses" rework is yours to take in a follow-up CR.

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
```diff
@@ -0,0 +1,116 @@
...
+  Copyright Strata Contributors
+
+  SPDX-License-Identifier: Apache-2.0 OR MIT
+-/
+module
+
+meta import Strata.Languages.Core.StatementSemantics
+
+/-!
+# A concrete evaluator satisfying the semantic well-formedness conditions
```
joscoh, line 11: But not type safety :)

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
```diff
@@ -0,0 +1,116 @@
...
+  | .const _ c  => some (.const () c)
+  | .bvar _ i   => some (.bvar () i)
+  | .op _ o ty  => some (.op () o ty)
+  | .abs _ n ty b => some (.abs () n ty b)
+  | _ => some canonicalValue
+
+/-- On a free-variable atom the witness returns the store binding. -/
+theorem coreWitnessEval_wfVar : WellFormedSemanticEvalVar coreWitnessEval := by
+  intro e v σ _hwfs hget
+  -- `getFvar e = some v` forces `e` to be `.fvar _ v _`; every other case is
```
joscoh: I don't think this comment is needed, but not blocking.

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
```diff
@@ -0,0 +1,116 @@
...
+instance : Inhabited WFCoreEval :=
+  ⟨coreWitnessEval, coreWitnessEval_wfVar, coreWitnessEval_wfVal⟩
+
+/-- Corollary: the conjunction is inhabitable by a genuine evaluator. -/
+theorem exists_wf_coreEval :
+    ∃ δ : CoreEval, WellFormedSemanticEvalVar δ ∧ WellFormedSemanticEvalVal δ :=
+  ⟨(default : WFCoreEval).eval, (default : WFCoreEval).wfVar, (default : WFCoreEval).wfVal⟩
+
+/-- Non-degeneracy: the witness genuinely returns the store binding on a free
+    variable (it is not a constant-`none` evaluator). -/
```
joscoh: Minor, but I'm not sure why this property is important, since this is a degenerate evaluator anyway. A constant `none` evaluator would be fine if the goal is to show satisfiability of the predicates.

joscoh (approved):

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
PROgram52bc: Fair — added a docstring note in 8f63593ca that the witness preserves value-ness but not typing (a boolean expr normalizes to the integer `canonicalValue`), which is all the well-formedness predicates constrain.

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
PROgram52bc: Removed in 8f63593ca.

StrataTest/Languages/Core/Tests/SemanticEvalWitness.lean
PROgram52bc: Right, it's not needed for satisfiability — removed the lemma in 8f63593ca.

PR #1432, opened 2026-07-01
feat(Core): emit recursive functions as define-fun-rec (SMT-LIB §4.2.5/§4.2.6)

kondylidou:
## Summary

Recursive functions declared with `rec function` (either `@[cases]` for
structural recursion or `decreases` for int-valued recursion) were previously
emitted as opaque uninterpreted functions with axioms. This PR emits them as
`define-fun-rec` / `define-funs-rec` instead, giving solvers complete
definitions for both proof (UNSAT) and counterexample (SAT) directions.

## Changes

- **`Solver.lean`** — add `defineFunRec` emitting `(define-fun-rec ...)` (§4.2.5);
  existing `defineFunsRec` handles mutual groups (§4.2.6)
- **`AbstractSolver.lean` / `IncrementalSolver.lean`** — add `defineFunRec` to
  the abstract interface and implement it for the incremental backend
- **`SMTEncoder.lean`** — add `SMT.Context.recifs` to collect recursive function
  bodies; `toSMTOp` merges the two identical `isRecursive` branches and registers
  any `rec function` body as a RecIF; add `encodeFunctionsRec` for batch emission
- **`Verifier.lean`** — add `encodeFunctionRec` to `AbstractEncoder`; patch both
  the batch path (`encodeCore`) and incremental path (`encodeDeclarationsAbstract`)
  to filter RecIF UFs from `declare-fun` and emit them as `define-fun-rec` instead
- **`MetaVerifier.lean`** — carry `recifs` through `SanitizedContext` so the
  meta-verifier tactic path has the recursive definitions available
- **`SMTEncoderTests.lean`** — fix positional `SMT.Context.mk` calls broken by
  the new `recifs` field

### DIFF
--- a/Strata/DL/SMT/AbstractSolver.lean
+++ b/Strata/DL/SMT/AbstractSolver.lean
@@ -138,6 +138,9 @@ structure AbstractSolver (τ : Type) (σ : Type) (m : Type → Type) [Monad m] [
   /-- Define an interpreted function with a body term. -/
   defineFun : String → List (String × σ) → σ → τ → m Unit
 
+  /-- Define a self-recursive function as `define-fun-rec` (SMT-LIB §4.2.5). -/
+  defineFunRec : String → List (String × σ) → σ → τ → m Unit
+
   /-- Declare a new sort with the given arity. Returns the declared sort. -/
   declareSort : String → Nat → m σ
 
--- a/Strata/DL/SMT/IncrementalSolver.lean
+++ b/Strata/DL/SMT/IncrementalSolver.lean
@@ -237,6 +237,16 @@ def mkIncrementalSolver : AbstractSolver Term TermType IncrementalSolverM where
     let bodyStr ← termToStr body
     emitln s!"(define-fun {quoteIdent name} ({inline}) {retStr} {bodyStr})"
 
+  defineFunRec name args retTy body := do
+    let retStr ← typeToStr retTy
+    let mut typedArgs := []
+    for (n, ty) in args.reverse do
+      let tyStr ← typeToStr ty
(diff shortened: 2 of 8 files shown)


shigoel:
@kondylidou Thanks for this PR! However, we do not want to generate `define-fun-rec` (at least, not by default) -- recursive functions at the SMTLib level are axioms anyway (no termination checking) and our current encoding makes that explicit.  Also, IIUC, z3 and cvc5 have slight variations in handling them. 

May I ask why you prefer `define-fun-rec`?

kondylidou:
> @kondylidou Thanks for this PR! However, we do not want to generate `define-fun-rec` (at least, not by default) -- recursive functions at the SMTLib level are axioms anyway (no termination checking) and our current encoding makes that explicit. Also, IIUC, z3 and cvc5 have slight variations in handling them.
> 
> May I ask why you prefer `define-fun-rec`?

Thanks for the feedback, you're right, and we ran the experiment to confirm it.

We tested both approaches (opaque UF + 3 axioms vs `define-fun-rec` for `pos.fromInt`, with `pos`/`nat` as algebraic datatypes in both cases):

- UNSAT direction: both pass identically
- SAT direction: both return `unknown` through Strata's verify interface

The concrete counterexample model benefit we had in mind for `define-fun-rec` would only matter if Strata implemented a `get-model` + validation workflow. Since it doesn't currently, there is no practical difference.

We'll withdraw the Strata PR. The StrataBoole PR will revert `pos.fromInt` to opaque UF + 3 axioms — the algebraic datatype is the actual fix, and that lives entirely in StrataBoole.

Sorry for the noise.

PR #1435, opened 2026-07-06
Exception Support Prototype

kadirayk:


docs/design/laurel_extensions.md
```diff
@@ -0,0 +1,449 @@
...
+
+This document proposes a **single, coherent set of Laurel extensions** intended to
+host *all* the exception-handling features `F1`–`F36` — F1–F30 for the three initial
+front-ends (Java/Python/JavaScript) and F31–F36 for the Kotlin/C# candidates (each
+`F<n>` is named in the coverage table below). The features are interdependent, so we
+design the extension *areas* together rather than feature-by-feature, then show how
+each area serves the relevant `F<n>`.
+
+It is anchored in a design baseline: a single optional `throws` type per procedure, a
+separate `onThrow` exceptional postcondition, a value-carrying `throw`, and
```
keyboardDrummer, line 13: Do we not also need to specify when the throws occurs? We could have syntax like `on <conditionForThrowing> throws <postconditionWhenThrowing>`

Strata/Languages/Laurel/LaurelAST.lean
```diff
@@ -229,9 +229,14 @@ structure Procedure : Type where
...
       with this expression as the SMT trigger. -/
   invokeOn : Option (AstNode StmtExpr) := none
-  /-- Axioms to emit alongside this procedure. Populated by the contract pass from
-      `invokeOn` and ensures clauses. -/
-  axioms : List (AstNode StmtExpr) := []
+  /-- Optional declared exception type (E4): the single type this procedure may
+      throw, bounded by the channel root `BaseException`. Recorded as
+      exceptional-contract data; not enforced (declare-or-catch is front-end
+      policy) and not yet lowered (E7). -/
+  throwsType : Option (AstNode HighType) := none
```
keyboardDrummer, line 236: Small design decision: whether to bind the exception variable here or in the onThrow clauses.

Binding it in the onThrow might be good since the variable is not available in the body, unlike the return variables.

Strata/Languages/Laurel/LaurelAST.lean
```diff
@@ -456,12 +489,29 @@ def StmtExpr.constrName : StmtExpr → String
...
+/-- Name of the E1/E2 exceptional-channel root composite (`BaseException`),
+    defined in the exception prelude (`exceptionDefinitionsForLaurel`).
+
+    Shared here (like `bodyLabel`) so the resolver, the prelude gating, and
+    FilterPrelude's dependency tracking all agree on the exact name:
+    - `throw`'s operand is type-checked against this type (resolution);
+    - any program containing a `throw` (or naming this type) pulls the
+      exception prelude in (gating), since the channel root must be in scope
+      wherever the exceptional channel is used. -/
+def baseExceptionTypeName : String := "BaseException"
```
keyboardDrummer, line 535: Doesn't this mechanism imply that we can't throw arbitrary types? Is there a benefit to having that restriction?

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
```diff
@@ -582,27 +693,85 @@ def translateStmt (stmt : StmtExprMd)
       -- Hole in statement position: treat as havoc (no-op).
       -- This can occur when an unmodeled call's Block is flattened.
       return []
+  | .Throw value =>
```
keyboardDrummer: If you extract the code in LaurelToCoreTranslator to a separate pass, and then add a test that shows the Laurel before and after that pass, then it'll be much easier to review the quality of the transformation.

docs/design/laurel_extensions.md
kadirayk: Right! I didn't think about this case, this is added now.

Strata/Languages/Laurel/LaurelAST.lean
kadirayk: Exception variable binding is per clause, like `onThrow(e) P`, `catch e when ...`, `when C throws (e) P`, etc.

Strata/Languages/Laurel/LaurelAST.lean
kadirayk: It is still possible to throw arbitrary types. Currently the idea is that we require exceptions to be rooted on `BaseException`, so that everything lowers to a single composite `Err`and we can reuse composite subtyping for `catch`/`is` dispatch. The frontends of languages that throw arbitrary values can box such values in a composite extending `BaseExceotion`, a demo of this is in`ThrowArbitraryValue.lean`.

Strata/Languages/Laurel/LaurelToCoreTranslator.lean
kadirayk: Done! `EliminateExceptionPass.lean` shows the before/after.

PR #1436, opened 2026-07-08
fix(Core): complete bv128 literal support in FormatCore and Translate

kondylidou:
FormatCore's bitvec catch-all arm discarded the actual constant value and emitted bv64(128) for any 128-bit literal — wrong type and wrong value. Translate had no bv128Lit case, causing Core→DDM translation to fail for programs with 128-bit literals.

Add explicit bv128 arms to both, mirroring the existing bv64 pattern.

- Strata/DDM/*
- Strata/Languages/Boole/*
- Tools/BoogieToStrata


joscoh (approved):

shigoel:
@kondylidou I think bv of width 128 made it to the Core Grammar independently. Also note that Core syntax has changed so we don't see `bv128` but `bv W128` now.

In any case, this PR looks like it can be closed?

